PR #182, opened 2011-09-28
Use external downloaders

rbrito:
Hi.

> I'm filing this so that it doesn't get lost and so that I also motivate myself to not let this feature slip and become out-of-sync with the rest of the program. This *is* very preliminary *and* unfinished *and* dirty, but I'm kind of trying to adopt some "agile" practices of being pragmatic and giving early what I have, so that it can be changed soon.
> As a corollary of the "dirty" in the paragraph above, I would like to have anybody give opinions on this branch of mine. I expect to be rebasing often so that we can have something clean when this gets merged.

I am working on using other, external, more specialized downloaders with `youtube-dl`. The main intention of this is to be able to leverage what those tools already have (e.g., segmenting the downloads, making multiple connections etc.).

My main motivation is to use `aria2c` so that I can download large videos from youtube without having to wait an entire life to download something that is big (e.g., a long video from a talk or lectures).

Of course, this can, in principle, be generalized to work with any site---not only youtube, but I first needed a prototype.

The code is not clean and has many issues, but I would appreciate reviews, comments even at this early stage.

Regards.


HPP8140:
This is what I was looking for. I currently use pyLoad download manager, but their youtube plugin is dated and does not support DASH. 

Here's my current usecase:
1. Select youtube video on android app
2. Share video to PyLoad app which will begin the download on pyLoad running on a server.
3. Use pyLoad Android app to check progress.

I've modified the pyLoad youtube plugin to call youtube-dl. Youtube-dl is awesome, but I've lost the capability to check the progress like described above through the app.

So I wanted to find a way to read the download links from youtube-dl from a file or send them to pyLoad (external download manager) instead.

HPP8140:
Actually, I just discovered the  -g, --get-url option, which might work also.

phihag:
Thank you for your contribution! I've added aria2c with the same options you proposed to the supported external downloaders and hence think this PR can be closed.

PR #464, opened 2012-10-09
avconv output

danut007ro:
This pull request is capturing avconv output (at least, don't know about ffmpeg) and is showing how much of the encoding it has done, as percentage.

Also, on my system, the aac codec is marked as experimental, so I used libfaac for aac encoding (if there is another option, that code can be removed)

### DIFF
--- a/youtube_dl/PostProcessor.py
+++ b/youtube_dl/PostProcessor.py
@@ -71,14 +71,13 @@ class FFmpegExtractAudioPP(PostProcessor):
 
 	@staticmethod
 	def detect_executables():
-		def executable(exe):
-			try:
-				subprocess.check_output([exe, '-version'])
-			except OSError:
-				return False
-			return exe
-		programs = ['avprobe', 'avconv', 'ffmpeg', 'ffprobe']
-		return dict((program, executable(program)) for program in programs)
+		available = {'avprobe' : False, 'avconv' : False, 'ffmpeg' : False, 'ffprobe' : False}
+		for path in os.environ["PATH"].split(os.pathsep):
+			for program in available.keys():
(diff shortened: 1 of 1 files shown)


youtube_dl/PostProcessor.py
```diff
@@ -108,16 +117,55 @@ def run_ffmpeg(self, path, out_path, codec, more_opts):
...
+		start = time.time()
+		# open process redirecting stderr to stdout
+		p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, close_fds=True)
+		import fcntl
+		import errno
+		import select
+		# entire captured output
+		p_output = ''
+		# size=     765kB time=243.67 bitrate=  25.7kbits/s
+		reo = re.compile("""size=\s*(?P<size>\S+)				# size
```
phihag: The formatting looks weird here. Did you use tabs?

youtube_dl/PostProcessor.py
```diff
@@ -108,16 +117,55 @@ def run_ffmpeg(self, path, out_path, codec, more_opts):
...
+							\stime=(?P<time>\S+)				# time
+							\sbitrate=\s*(?P<bitrate>[\d\.]+)	# bitrate
+							""", re.X)
+		# make stdout non-blocking
+		fcntl.fcntl(p.stdout.fileno(), fcntl.F_SETFL, fcntl.fcntl(p.stdout.fileno(), fcntl.F_GETFL) | os.O_NONBLOCK)
```
phihag: Sorry for the delay, I'm looking into merging this (at least the avconv status part) just now. Why is blocking I/O insufficient here?

PR #507, opened 2012-10-31
Submitting Infinite Retries on 0 option

Sepero:
Code and Documentation were modified accordingly. Anything neglected is unintentional.


phihag:
We can simply set retries to `inf` ;) - fixed in baeaeffce550b23848983cd281f173a7906cd7f9

PR #672, opened 2013-02-06
removed data_files param

viniciuscainelli:
always returning:
error: can't copy 'youtube-dl.bash-completion': doesn't exist or not a regular file


phihag:
This should be fixed by now. I apologize for the delay.

ru2nuts:
Thanks!

PR #1283, opened 2013-08-20
Added ability to specify audio channels and sample rate

schmoofer:
Added option to specify number of channels for audio extraction. Also added support for specifying sample rate as part of the "audio quality" parameter. Integers greater than 8000 will be interpreted as the sampling rate, thus 8000 is the minimal.


phihag:
I'm still not sure what this option would be used for, and there has been no update since my inquiry well over a year ago. I must therefore sadly reject this PR. Feel free to comment and provide [context](https://github.com/rg3/youtube-dl/blob/master/README.md#is-there-enough-context-in-your-bug-report) for this PR and we will re-discuss it. Thanks!

PR #1348, opened 2013-08-29
Filesize as xattribute

brimston3:
Rebase old pull request to master.

Requires the xattr package.

When used, the xattribute user.ytdl.filesize will be set with the expected filesize. This can be useful for re-streaming the download somewhere else.


phihag:
Adapted in master. Thank you for your contribution!

PR #1601, opened 2013-10-14
Travis test also rtmpdump videos

FiloSottile:
Is there a reason we are not doing this? Maybe we needed #1565 

If this works, I'll convert all the `skip_download` tests


phihag:
Thanks! Applied to current youtube-dl master.

Issue #1626, opened 2013-10-18
Please, provide a changelog


rbrito:
Hi.

Wearing my downstream hat, it would be nice to have a changelog that I can ship to users.

I have one that asked me how I provide the Debian changelogs on the Debian packaging and I replied "by reading the source code", which after thinking a bit, is some duplicated effort for other distributions that don't benefit from my work.

Can you provide something as a changelog? Even something like a `git shortlog --no-merges` would be helpful and (hopefully, not too noisy), which is, BTW, what Linus Torvalds himself uses to announce newer kernel releases.

Thanks in advance,

Rogério.

shirishag75:
oh goody, this is/would be good, exactly what I was gonna ask for, thanx rbrito :)

dntt1:
please create it or assist me creating it

yan12125:
Added in 35aa6c538f593eb47169e6b06ef383ad12b45930.

PR #1715, opened 2013-11-03
Changelog updates

jaimeMF:
There've been some requests (like #1626) for having a CHANGELOG. While we have one, it has only been used for one version. I personally think that it would be useful to maintain a changelog; for users, packages maintainers and developers that use the module.
For that reason I've done some modifications so that we can edit the changes for the version in development in the `@DEV@` section of the `CHANGELOG.md`, and the `realease.sh` script will set the correct version when run.
Additionally the changes will be added to the `versions.json` file and to the atom feed (using pandoc).


phihag:
Sorry, but I do not think a separate CHANGELOG file would be useful enough to warrant updating it with basically every commit. I am open to providing a semi-automatically-generated CHANGELOG, provided that someone suggests the code that drops me from the release script into an editor prefilled with all commits (plus some intelligence, for example removing merge commits and summarizing new extractors since the last release).

PR #2216, opened 2014-01-23
temporary fix for the time jumps

amq:
I have added -t 72000 (20 hours) to force ffmpeg to mux the file when a "time jump" happens.

For some reason every jump I've seen so far (at least 6 in total) was to 20 hours or more.

I think It's a safe value, which won't cause any troubles. There is no chance recording a live stream for 20 hours without interruptions anyway (my personal record is 3 hours).

https://github.com/rg3/youtube-dl/issues/2213

### DIFF
--- a/youtube_dl/downloader/hls.py
+++ b/youtube_dl/downloader/hls.py
@@ -13,7 +13,7 @@ class HlsFD(FileDownloader):
         self.report_destination(filename)
         tmpfilename = self.temp_name(filename)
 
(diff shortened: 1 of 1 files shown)


phihag:
This would break HLS videos of over 20 hours or more. I do not see any reason for this limit. If you are sure this occurs with a certain service, please propose a new field where the deficiency is documented, upon which the fix may be applied.

PR #2266, opened 2014-01-28
Retrieve the URLs of all availale thumbnails

MikeCol:
The clips on Tumblr pages are usually accompanied by more than one thumbnail. Their URL is stored in an array in  the source code of the page.

This patch decodes this array in its entirety and returns their URL under the *thumbnails* key.
It also returns the first URL under the *thumbnail* key.


phihag:
Sorry, but the multiple thumbnails do not seem to be present anymore, at least not in [our tumblr test page](http://tatianamaslanydaily.tumblr.com/post/54196191430/orphan-black-dvd-extra-behind-the-scenes).

PR #2269, opened 2014-01-29
Write all thumbnails to disk

MikeCol:
Many extractors return a thumbnail in `thumbnail` or a list of them in `thumbnails` (in form of a list of dicts each with an `url` key).

The proposed change uses both sources to write one or all of them to disk, depending how many the user has asked for: one with `--write-thumbnail` or all with the new `--write-all-thumbnails`.

If the user only requests one, the content of `thumbnail` is returned, or if not present, the first entry from the list.

If the user requests all, the content of `thumbnail` (if present) is prepended to the list, if it is not already in there.

If only one file is to be written, the normal naming schema is used.
If more than one files are being wirtten, a counter is appended to the file name (e.g. _1, _2 or _01, _02, depending on the number of files in the list).

The download part is the same as before, it only now iterates over the array.

I hope I found all place in `youtube-dl` where additional command line parameters have to be documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ which means you can modify it, redistribute it or use it however you like.
     --write-annotations              write video annotations to a .annotation
                                      file
     --write-thumbnail                write thumbnail image to disk
+    --write-all-thumbnails           write all thumbnail images to disk
 
 ## Verbosity / Simulation Options:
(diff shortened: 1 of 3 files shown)


phihag:
Thank you very much! I've simplified and updated the implementation in ec82d85.

PR #2398, opened 2014-02-16
Add support for 4tube.com.

patheticpat:


chillin-:
I've never seen youtube-dl work for 4tube. Either remove support or fix.

jaimeMF:
@chillin- You'll have to open a new issue [with the output of `youtube-dl URL --verbose`](https://github.com/rg3/youtube-dl/blob/master/README.md#bugs).

PR #2442, opened 2014-02-23
YoutubeDL: Make the decision about removing the original file after each...

### LINKED ISSUE #2261: --add-metadata -x / --xattrs -x does not discard source video
Follow-up to Issue #1570  

When running ```youtube-dl --add-metadata -x --audio-format mp3 BaW_jenozKc``` the video file is not erased. The same occurs with using ```--xattrs``` in conjunction with -x

jaimeMF:
... postprocessor is run (fixes #2261)

If one of the processors said the file should be kept, it wouldn't pay
attention to the response from the following processors. This was wrong if the
'keep_video' option was False, if the first extractor modifies the original file
and then we extract its audio we don't want to keep the original video file.

I think that's how we should handle the `keep_video` option, but I want that someone else reviews this change.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1006,14 +1006,15 @@ class YoutubeDL(object):
         """Run all the postprocessors on the given file."""
         info = dict(ie_info)
         info['filepath'] = filename
-        keep_video = None
         pps_chain = []
         if ie_info.get('__postprocessors') is not None:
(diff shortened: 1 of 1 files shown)


phihag:
LGTM, feel free to merge. @jaimeMF, can you also have a look at the somewhat related #3829?

jaimeMF:
Merged in f3ff1a3696c4080468e2cc5810c34273b148bd3e

PR #2723, opened 2014-04-07
Add a postprocessor to concatenate files in a playlist

AGSPhoenix:
It may not be pretty, but it friggin' works. I did some initial testing with The Daily Show, but I have no idea how it handles other sites. Of particular concern is non-MP4 formats. Currently, the output file is always an MP4 container (this should be fixed), since I have concerns about it breaking with obscure containers. I've tested AVI, FLV, MP4, and MKV (they work), but there are dozens of others I haven't tested. For the time being, MP4 seems a reasonable default, since just about everything can play it.

This patch changes a few interfaces, most notably adding a return value to the process_info() function in YoutubeDL.py. Also, a 'saved-filename' key is added to the dictionary returned by process_video_result(). Also, run_ffmpeg_multiple_files() now accepts an optional list argument called preopts, for arguments that need to go before the input files. (Oh, ffmpeg. You so silly.)

The command line option is `--join`. Also, joining silently fails if the IE doesn't return "playlist" as the `_type`. Also, I have no idea what happens if you specify multiple URLs that are each a playlist. Or if you specify some that are and some that aren't. Or if the items in the playlist have different formats. Or something else I forgot in the time it took me to write out those other three.

I can't think of anything else to add to this wall of text, and getting acquainted with the youtube-dl internals has left my brain a pile of mush, so beware of mistakes both large and small. Now if you'll excuse me, I've got about a month's worth of Daily Show to catch up on.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -58,7 +58,7 @@ from .utils import (
 )
 from .extractor import get_info_extractor, gen_extractors
 from .downloader import get_suitable_downloader
-from .postprocessor import FFmpegMergerPP
+from .postprocessor import FFmpegMergerPP, FFmpegConcatPP
 from .version import __version__
 
 
@@ -640,6 +640,18 @@ class YoutubeDL(object):
                                                       extra_info=extra)
                 playlist_results.append(entry_result)
             ie_result['entries'] = playlist_results
+            
+            #Run concat PP
+            if self.params['concat']:
+                pp = FFmpegConcatPP(self)
+                
+                mergelist = []
+                for video in playlist_results:
+                    mergelist.append(video['saved_filename'])
+                ie_result['__files_to_append'] = mergelist
+                
+                pp.run(ie_result)
+            
             return ie_result
         elif result_type == 'compat_list':
             def _fixup(r):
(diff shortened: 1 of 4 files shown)


yan12125:
Is this PR still in consideration?

AGSPhoenix:
In a word, no. It was never merged because it was never completed. I managed to get it working for what I needed, but it wasn't good enough to release even as an experimental feature in the main branch. After getting it to it's present kinda-working state, RL issues took me away from it, and by the time I was able to work on it again, I had forgotten how half of it worked, and another PR to do that same thing was submitted (#2844).

If it's a feature you need and you can't implement yourself, I might be able to take another shot at it, though I no longer have hours of free time nowadays, so no guarantees.

Wowfunhappy:
I would like to add my support for this feature!

If edge cases are a concern for certain formats, would it be worth simply restricting the flag to being compatible with only certain formats (eg mp4)?

yan12125:
It's OK to support MP4 first, while the implementation should be easy to extend support to other formats.

Wowfunhappy:
After playing extensively today with concatenating Youtube downloads with ffmpeg, I believe the primary cause of failures are _FPS differences_!

It is of course not the only cause of failure—resolution or codec differences will do it too—but it's the least obvious one. And, the failure state for frame-rate differences is bizarre—ffmpeg will happily write the concatenated file, and the first part may even play fine, but as soon as you reach the section whose source video had a different frame-rate... well, the effects vary, but suffice to say really weird things start happening.

If the reason this feature hasn't been implemented is because devs are worried about edge cases, try adding a framerate check! I bet that will solve most of the problems. (And if not, I'd still like to have this as an experimental feature. 😀)

PR #2778, opened 2014-04-20
Add stop/pause/resume functionality

MrS0m30n3:
Add stop/pause/resume funcitonality on FileDownloader

Example: https://github.com/MrS0m30n3/youtube-dl/blob/master/test_handlers.py


phihag:
There is now a far easier way to implement this: Simply set one object as one of the `progress_hooks` and do in that what you want (stopping / pausing / etc.).

As a further advantage, this is also supported by the f4m downloader and support may be extended to more downloaders in the future.

MrS0m30n3:
You mean something like this:

```python
import thread
from time import sleep

from youtube_dl import YoutubeDL

class MyObject(object):
    def __init__(self):
        self.pause = False
        self.stop = False

    def data_hook(self, data_dictionary):
        print data_dictionary

        # Do whatever i want here
        while self.pause:
            print 'Paused'
            sleep(1)
            
        if self.stop:
            raise KeyboardInterrupt


if __name__ == '__main__':
    test_object = MyObject()

    test_urls = ["https://www.youtube.com/watch?v=zSAJ0l4OBHM"]

    ydl_opts = dict()
    
    with YoutubeDL(ydl_opts) as ydl:
        ydl.add_default_info_extractors()
        ydl.add_progress_hook(test_object.data_hook)
        
        thread.start_new_thread(ydl.download, (test_urls, ))
        
        # Wait 2 seconds and pause
        sleep(2)
        test_object.pause = True
        
        # Wait 10 seconds then continue the download process
        sleep(10)
        test_object.pause = False
        
        # Wait 4 seconds then kill the download process
        sleep(4)
        test_object.stop = True
        
        # Wait for the thread to finish
        # 20 seconds should be enough
        sleep(20)
```

Also, which way would be more suitable to terminate the download process? Is it ok to raise an exception like the example above?

phihag:
Yup, precisely. Note that you most likely need more stringent synchronization - at the moment you're relying on race conditions in your favor (which will happen with cpython thanks to the GIL).

You can just raise exceptions in the progress hooks.

MrS0m30n3:
Yup, i know this is just an example. Thanks for your time.

PR #2845, opened 2014-05-01
Don’t simulate when a quiet printing is asked

galex-713:
It would be really usefull to be able to print arbitrary information without having to run youtube-dl a second time to download the video. This way we still can quietly print and simulate using both --simulate and --get-*.


phihag:
Sorry for the long delay. I've added `--print-json` now (this should be the preferred method when you're running youtube-dl programatically, which I strongly suspect).

PR #2951, opened 2014-05-19
[youtube_dl] Added cancelable metadata / playlist downlod progress hook

disfasia:
I added cancelable hooks to playlist metadata downloads.

Here an example showing how to skip downloads after the 3rd entry of a playlist:

``` python
import youtube_dl

url = 'http://www.youtube.com/watch?v=QonQ_8qnAbY&list=PLC3CC1956BA739976&index=8'

ydl = youtube_dl.YoutubeDL({
    'ignoreerrors': True,
    'quiet': True,
    'socket_timeout': 10
})

ydl.add_default_info_extractors()

# dummy hook
def hook(status):
    if status['current'] > 3:
        print "Skipping..."
        return False # returning False from any of the added hooks makes the next request skipped
    else:
        print(status)

# Adding the hook
ydl.add_metadata_hook(hook)

result = ydl.extract_info(url, download = False)
```

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -640,9 +640,10 @@ class YoutubeDL(object):
                     'extractor_key': ie_result['extractor_key'],
                 }
                 playlistProgress = {
-                    'current': i,
+                    'current': i-1,
                     'total':   n_entries,
                     'playlist': ie_result['webpage_url'],
+                    'currentData': playlist_results[-1] if len(playlist_results) > 0 else None,
                     'next': entry
                 }
                 skipNext = False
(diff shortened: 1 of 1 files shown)


phihag:
Sorry, without insufficient [context](https://github.com/rg3/youtube-dl/blob/master/README.md#is-there-enough-context-in-your-bug-report) my imagination fails me why this would be useful - why not simply pass in `--playlist-end 3`? What does this option enable that was missing previously?

youtube_dl/YoutubeDL.py
```diff
@@ -114,6 +114,7 @@ class YoutubeDL(object):
     nooverwrites:      Prevent overwriting files.
     playliststart:     Playlist item to start at.
     playlistend:       Playlist item to end at.
+    playlistonly:      Stops after getting the playlist entries
```
phihag: I believe this can be achieved already by setting `extract_flat`.

timofonic:
@phihag @me@disf.asia @disfasia @dstftw  Any news?

PR #3000, opened 2014-05-28
Filter DRM protected media in f4m downloader

rzhxeo:


phihag:
Can you name a URL that's currently available? It's fine if it's only available from Germany.

rzhxeo:
http://rtl-now.rtl.de/ahornallee/folge-1.php?film_id=90419&player=1&season=1

phihag:
That URL works fine for me already. What error do you get before applying this PR?

rzhxeo:
The download works but the downloaded video is encrypted.
Adobe added an extra layer of encryption to rtmp that rtmpdump doesn't know about.
With this PR applied, the encrypted videos get filtered out and no useless garbage is downloaded.

phihag:
I apologize; I may be misunderstanding, but are you saying that this will simply lower the file size, but not result in a playable video either? Additionally, would you mind updating this PR so that it can be merged? Maybe my misunderstanding is caused by a fault of my own when merging the PR.

rzhxeo:
I've updated the PR.
Encrypted media just gets filtered out, the encrypted (useless) videos are not downloaded at all.
I don't think that Adobe's new encryption scheme is broken yet.

phihag:
Thanks, merged.

PR #3426, opened 2014-08-03
Add --sleep-interval option to sleep between downloads during a batch process

rupertbaxter2:
Adds the option --sleep-interval <seconds> (default 0)

Waits the specified number of seconds before downloading the next file in a batch.  This is helpful for dealing with "503: Service Unavailable" errors that sometimes occur if one download is quickly started after the previous.  I've seen this happen sometimes with vimeo channel downloads.


phihag:
Thanks, merged!

PR #3618, opened 2014-08-29
Add --bind-ip

aivarsk:
When making client TCP/IP connections, bind to IP on the local machine. This option can be useful if your machine is bound to multiple IPs.

Should fix issue#721 and others that are marked as duplicates. Other tools have it as well: wget has --bind-address option and I think curl --interface does the same.
Issue#2379 with proposal to change socket.socket() had a comment that "We would prefer not to meddle with the global Python socket's settings." This approach (hopefully) changes behaviour of TCP/IP client sockets that call connect() and does not call bind for UDP and server sockets.
Once support for Python 2.6 is dropped, it can be replaced with HTTPConnection's source_address (since 2.7)


phihag:
Starting from youtube-dl 2015.01.10.1, there is a new [`--source-address` option](https://github.com/rg3/youtube-dl/blob/master/README.md#network-options). See [our FAQ](https://yt-dl.org/update) if you need help updating.

PR #3829, opened 2014-09-25
YoutubeDL: check for overwriting files in the downloader (fixes #3916)

### LINKED ISSUE #3916: The progress hooks are not passed the file info when it already exists
I opened this issue to describe what I'm trying to solve in this pull request #3829.

When we add a progress hook using `YoutubeDL.add_progress_hook`, to be able to get the download info `dict` from the `HttpFD` class which looks like this:

```python
self._hook_progress({
    'downloaded_bytes': byte_counter,
    'total_bytes': data_len,
    'tmpfilename': tmpfilename,
    'filename': filename,
    'status': 'downloading',
    'eta': eta,
    'speed': speed,
})
```

This info `dict` is passed to the progress hooks only in these cases:

- Download in progress
- Download finished
- The file exists and `continuedl` is `True` and we're using `.part` files

But we don't get this `dict` when `nooverwrites` is `True` and the file exists, which I think is the same as when `continuedl` is `True`, so what's preventing the hooks from getting the info `dict` is this check inside `YoutubeDL.process_info`:

```python
if self.params.get('nooverwrites', False) and os.path.exists(encodeFilename(filename)):
    success = True
```

If we let `FileDownloader.download` handle this check, this problem will be solved:

```python
def download(self, filename, info_dict):
    """Download to a filename using the info from info_dict
    Return True on success and False otherwise
    """
    # TODO: Check here for nooverwrites and if the file exists
    if self.params.get('continuedl', False) and os.path.isfile(encodeFilename(filename)) and not self.params.get('nopart', False):
        self.report_file_already_downloaded(filename)
        self._hook_progress({
            'filename': filename,
            'status': 'finished',
            'total_bytes': os.path.getsize(encodeFilename(filename)),
        })
        return True

    return self.real_download(filename, info_dict)
```

So this pull request #3829 fixes this problem.

mdawar:
Right now `YoutubeDL.process_info` is checking if `nooverwrites` is `True` and if the file exists and setting `success = True`, which is preventing this to be reported and the progress hooks added using `YoutubeDL.add_progress_hook` from being called to access the status `dict`.

```python
self._hook_progress({
    'filename': filename,
    'status': 'finished',
    'total_bytes': os.path.getsize(encodeFilename(filename)),
})
```

So if we let `FileDownloader.download` handle this check the same way it's doing for `continuedl` then the progress hooks will be called and it will be reported that the file already exists, and `success` will be set to `True` which will be returned by `dl()`, so no major changes were made to the class.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1000,49 +1000,46 @@ class YoutubeDL(object):
                             (info_dict['thumbnail'], compat_str(err)))
 
         if not self.params.get('skip_download', False):
-            if self.params.get('nooverwrites', False) and os.path.exists(encodeFilename(filename)):
-                success = True
-            else:
-                try:
-                    def dl(name, info):
-                        fd = get_suitable_downloader(info)(self, self.params)
-                        for ph in self._progress_hooks:
-                            fd.add_progress_hook(ph)
-                        if self.params.get('verbose'):
-                            self.to_stdout('[debug] Invoking downloader on %r' % info.get('url'))
-                        return fd.download(name, info)
-                    if info_dict.get('requested_formats') is not None:
-                        downloaded = []
-                        success = True
-                        merger = FFmpegMergerPP(self, not self.params.get('keepvideo'))
-                        if not merger._get_executable():
-                            postprocessors = []
-                            self.report_warning('You have requested multiple '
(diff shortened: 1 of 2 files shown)


jaimeMF:
Merged, thanks (and sorry for the delay).

PR #3954, opened 2014-10-15
Fix #2310. Play by the 8tracks rules

### LINKED ISSUE #2310: [8tracks] throws 403 error after first 4 songs. Nothing downloads.
8tracks recently updated their API to version 3. They probably updated their security on this (because, according to their license, you can only skip 3 songs within a certain time frame).

lenaten:
because we cant to predict the real download links without the server, we must play by the 8tracks rules:
* max 3 skip in hour. but playlist contains min 8 tracks, so 3 isn't enough
* continue to the next song after the current track done. but we don't know the exact length of track

so, some is better than nothing, and currently the exiting plugin is useless.
I calculate the average track length from playlist length and track count, then sleep after download failed until that new request be considered as legitimate.

ugly, takes a long time, but works!

### DIFF
--- a/youtube_dl/extractor/eighttracks.py
+++ b/youtube_dl/extractor/eighttracks.py
@@ -135,7 +135,7 @@ class EightTracksIE(InfoExtractor):
                     if download_tries > 3:
                         raise
                     else:
-                        ++download_tries
+                        download_tries += 1
                         time.sleep(avg_song_duration)
 
             api_data = json.loads(api_json)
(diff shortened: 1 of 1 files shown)


phihag:
Okay, I now understand why this is a very good idea. However, a couple of fine points:

- Shouldn't we download the current song before waiting?
- Can't we wait for the precise song length instead of an average?
- We should explain to the user what's happening
- We should provide a generic bailout option (`--dont-sleep` or so)

lenaten:
- I thought to download all songs together. One by one is like buffering. I don't like buffering.
- you cant get the song length from webpage. I tried.
- yap
- yap

phihag:
I've merged this now. Any further improvements can be made in-tree. Thank you very much, and sorry for the long wait!

PR #4015, opened 2014-10-23
[extractor/common] Added support for f4m manifest Version 2.0 and Rte.ie IE

Dineshs91:


phihag:
Thank you very much! Merged.

phihag:

Dineshs91:
Would love that. Please use this name: Dinesh S

Thanks

PR #4554, opened 2014-12-24
Add various anime download sites

MiJyn:
Soulanime, Gogoanime, and various sites that host anime (videofun, play44, byzoo, video44, videowing, playpanda, videozoo, playbb, easyvideo)

Let me know if anything needs to be changed if you wish to merge :)


youtube_dl/extractor/soulanime.py
```diff
@@ -0,0 +1,90 @@
...
+        domain = mobj.group('domain')
+
+        page = self._download_webpage(url, video_id, "Downloading video page")
+
+        video_url_encoded = self._html_search_regex(self._VIDEO_URL_REGEX, page, 'url', fatal=True)
+        video_url = "http://www.soul-anime." + domain + video_url_encoded
+
+        vid = self._request_webpage(video_url, video_id)
+        #ext = vid.getheader("Content-Type").split("/")[1]
+        ext = vid.info().gettype().split("/")[1]
```
phihag: Fixed in 2ccd1b1.

dstftw: >2ccd1b1

No such commit.

phihag: Pushed now; 2ccd1b10e58cc8e5173dc1aeedc2b3f0ef9b55bf

phihag:
As pointed out by @JaimeMF, these sites may not have the legal rights to distribute their content.

# gogoanime

According to http://www.gogoanime.com/private-policy , they did not acquire anything, and indeed, they only redirect.

However, the test URL refers to content that is - [according to Wikipedia](http://en.wikipedia.org/wiki/Puella_Magi_Madoka_Magica#Broadcast_and_distribution) copyrighted way too recently to have become part of the public domain (2011). I cannot find any licensing that would allow public distribution.

For the above reasons, I have removed the test. Since the gogoanime extractor is now without a valid test case, I have removed it as well.

# play44

play44 [claims](http://www.goodanime.net/private-policy) that they don't host the videos in a link on their main sites. This does not seem to be true:

```
$ youtube-dl -g test:play44
http://gateway.play44.net/videos/M/mahou-shoujo-madoka-magica-07.flv?st=5myTWm-VXyXTNz0XLvHWTA&e=1420401773&server=play44
```agica

The example URL refers to the same copyrighted series. The example URLs for the other play44-derived extractors seem to refer to [another series](http://myanimelist.net/anime/10719/Boku_wa_Tomodachi_ga_Sukunai) copyrighted in 2011, of which I cannot find any licensing that would allow public distribution.

Therefore, I have removed the test cases and hence the extractors related to play44.

# soulanime

soulanime [claims that they don't the content as well](http://www.soul-anime.tv/dmca-policy.html), but that does not seem to be true:

```
$ youtube-dl test:soulanime:watching -g
http://www.soul-anime.net/downloads/seirei-tsukai-no-blade-dance-episode-9/
```

The test URL seems to refer to [content that has been copyrighted in 2014](http://myanimelist.net/anime/22877/Seirei_Tsukai_no_Blade_Dance). Again, I fail to find any public domain or even a free dedication.

Therefore, I have removed the test case. Since the soulanime extractor is now without a valid test case, I have removed the extractor as well.

# videofun

videofun.me is somewhat surprising in that they don't seem to have any content on their [front page](http://videofun.me/). Looking at the [code](view-source:http://videofun.me/embed/8267659be070860af600fee7deadbcdb?w=600&h=438), I see references to anime44 though, which is probably related to play44. Our test URL seems to refer to the same content as the one for play44.

Therefore, I have removed the test case. Since the videofun extractor is now without a valid test case, I have removed the extractor as well.


If you do think that these sites do have permission to distribute their content, please provide hints that they do, and example URLs where it's obvious that the uploader had rights to distribute the content, and has transferred those to the video service. As it stands, it looks like all of these sites infringe copyright, and are therefore not fit to be included in youtube-dl.

PR #4612, opened 2015-01-02
Push API updates to simplify code, better error handling, add support for albums

xavierbeynon:
* Updated the api call to return more JSON info (as opposed to page scraping)
* Caught additional inconsistent error messages
* Added cases to fill in missing metadata with mediocre values, as opposed to failing all together
* Added audiomack album/playlist class to support album downloads

### DIFF
--- a/youtube_dl/extractor/audiomack.py
+++ b/youtube_dl/extractor/audiomack.py
@@ -9,21 +9,21 @@ import time
 
 
 class AudiomackIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?audiomack\.com/(song)/(?P<id>[\w/-]+)'
+    _VALID_URL = r'https?://(?:www\.)?audiomack\.com/song/(?P<id>[\w/-]+)'
     IE_NAME = 'audiomack'
     _TESTS = [
-        # audiomack
+        # hosted on audiomack
         {
             'url': 'http://www.audiomack.com/song/roosh-williams/extraordinary',
             'info_dict':
             {
                 'id': '310086',
-                "ext": "mp3",
-                "artist": "Roosh Williams",
+                'ext': 'mp3',
+                'artist': 'Roosh Williams',
                 'title': 'Extraordinary'
             }
         },
-        # audiomack through soundcloud
+        # audiomack wrapper around soundcloud song
         {
             'add_ie': ['Soundcloud'],
             'url': 'http://www.audiomack.com/song/xclusiveszone/take-kare',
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/audiomack.py
```diff
@@ -9,20 +9,21 @@
 
 
 class AudiomackIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?audiomack\.com/song/(?P<id>[\w/-]+)'
+    _VALID_URL = r'https?://(?:www\.)?audiomack\.com/(song)/(?P<id>[\w/-]+)'
```
phihag: Why is this change correct? This looks really strange and just adds a useless group, doesn't it?

youtube_dl/extractor/audiomack.py
```diff
@@ -9,20 +9,21 @@
...
     IE_NAME = 'audiomack'
     _TESTS = [
-        # hosted on audiomack
+        # audiomack
```
phihag: I very much prefer the older comment, which was much clearer.

youtube_dl/extractor/audiomack.py
```diff
@@ -9,20 +9,21 @@
...
         {
             'url': 'http://www.audiomack.com/song/roosh-williams/extraordinary',
             'info_dict':
             {
-                'id': 'roosh-williams/extraordinary',
-                'ext': 'mp3',
-                'title': 'Roosh Williams - Extraordinary'
+                'id': '310086',
+                "ext": "mp3",
```
phihag: Why have you changed the quotes here? It's not that important, but we strive to use `'` when possible, and have when possible a consistent quote character in a file.

youtube_dl/extractor/audiomack.py
```diff
@@ -37,33 +38,88 @@ class AudiomackIE(InfoExtractor):
         },
     ]
 
+    @staticmethod
+    def create_song_dictionary(api_response, album_url_tag, track_no=0):
+        # All keys are the same in audiomack api and InfoExtractor format
```
phihag: This seems to be a really complicated way of writing
```
{
    'title': api_response.get('title', album_url_tag),
    'uploader': api_response.get('artist'),
    'url': api_response['url'],
}
```
Notably, `id` looks really strange - shouldn't that include the `video_id`?

youtube_dl/extractor/audiomack.py
xavierbeynon: Sorry, that was a typo.  Originally I was detecting whether it was song or album, before I broke album off into its own extractor.

youtube_dl/extractor/audiomack.py
xavierbeynon: Agreed.

youtube_dl/extractor/audiomack.py
xavierbeynon: I made them consistent.

youtube_dl/extractor/audiomack.py
xavierbeynon: I didn't change this.  Yes, the dictionary comprehension could be unraveled.  But I think the one liner followed by the if statements makes clear what is expected, followed by what must be done with the unexpected.  Both are clear and equivalent; mine is just a bit more fun.

I don't know what you mean by the id's.  I understand that an id field must be present in the dictionary - the API response from AM conveniently contains a corresponding field that I copy over.

phihag:
Thanks! Merged.

PR #4615, opened 2015-01-02
[xhamster] Add HTTPS support

dwemthy:
xHamster can be browsed in HTTPS even though the downloads are HTTP. This patch adds support for the HTTPS URLs.


dstftw:
Thanks.

PR #4617, opened 2015-01-02
[vier] Add new extractor

lovebug356:
Add new extractor for vier.be


dstftw:
Thanks, merged with some changes.

PR #4625, opened 2015-01-03
[WatchTvSeries] Add new extractor for watch-tv-series.to

Mortal:
No tests -- sorry.

Optionally uses the [microdata](https://github.com/edsu/microdata) module (can be installed via pip) for more robust page parsing -- otherwise falls back on parsing HTML with regex.

With microdata, it should be possible to fetch entire series and not just seasons, but this is left for future work.

watch-tv-series.to only hosts links to other sites, so this depends on the movshare.net extractor -- other video host sites are not attempted.


jaimeMF:
Since the webpage seems to be exclusively used for copyrighted material, we can't add support for it.

PR #4639, opened 2015-01-05
Update rtlnl.py

bartkappenburg:
Added support for the non-www version of rtlxl.nl by making "www." optional.
E.g.:
http://www.rtlxl.nl/#!/divorce-277291/291d70e8-8bd2-39ef-b849-c4ec94fc46c2
and
http://rtlxl.nl/#!/divorce-277291/291d70e8-8bd2-39ef-b849-c4ec94fc46c2
are two urls for the same content.


jaimeMF:
Merged, thanks.

PR #4643, opened 2015-01-05
[Auengine] corrected extractions logic

Oteng:
The way the video download url was been extracted was
not working well so i change it for it to extract the
correct url

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@ updates_key.pem
 test/testdata
 .tox
 youtube-dl.zsh
+.idea
+.idea/*
--- a/youtube_dl/extractor/auengine.py
+++ b/youtube_dl/extractor/auengine.py
@@ -29,17 +29,12 @@ class AUEngineIE(InfoExtractor):
         webpage = self._download_webpage(url, video_id)
         title = self._html_search_regex(r'<title>(?P<title>.+?)</title>', webpage, 'title')
         title = title.strip()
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -31,3 +31,5 @@ updates_key.pem
 test/testdata
 .tox
 youtube-dl.zsh
+.idea
```
jaimeMF: This is probably specific to your editor, you [should add them to your global gitignore file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) instead.

jaimeMF:
The test video is an episode [from an anime](http://en.wikipedia.org/wiki/The_Legend_of_the_Legendary_Heroes#Anime), I doubt they have the rights to distribute it.
We must find one video that doesn't infringe copyright, otherwise we'll have to remove the extractor.

.gitignore
phihag: Oh well, doesn't hurt.

phihag:
Thanks, merged.

As @jaimeMF mentioned, the extractor is danger of removal and will be removed within 10 days if a valid example URL cannot be found.

Oteng:
about my editor sorry. About the extractor i am acctually working on a project based on this wonderfull project and in an attempted to get a passing build i decided to work on all the failing extractors and send you guys a pull request as a way of contributing back to the project that is helping me. Thanks guys

PR #4647, opened 2015-01-05
[downloader/f4m] Improved metadata handling

aajanki:
Metadata may be missing, and correctly write the metadata tag length

### DIFF
--- a/youtube_dl/downloader/f4m.py
+++ b/youtube_dl/downloader/f4m.py
@@ -187,6 +187,10 @@ def build_fragments_list(boot_info):
     return res
 
 
+def write_unsigned_int(stream, val):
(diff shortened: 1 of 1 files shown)


aajanki:
I don't have any samples, but Section 11.17 of F4M manifest specification (available at https://www.adobe.com/devnet/hds.html) defines <metadata> as optional ("This element MAY be present"). So I think youtube-dl should handle a missing metadata gracefully.

phihag:
This should indeed improve youtube-dl without any downside I'm seeing. Merged, thanks!

jaimeMF:
I don't know how I could miss all that info when I read the spec. Thanks!

PR #4673, opened 2015-01-10
Merge output format

aft90:
As explained in https://github.com/rg3/youtube-dl/issues/3610, https://github.com/rg3/youtube-dl/issues/2765 there are cases when downloading separate video and audio and then merging when the video container (usually webm containing VP8/VP9) is incompatible with the audio codec (aac/mp3). By default, youtube-dl attempts to create a merge in the container format of the video file.

This is a redo of pull request https://github.com/rg3/youtube-dl/pull/4577. This time around I've added a video option called "merge-ouput-format" that specifies what container format the merge output should be put into. It is only used if merging is required (something like "-f best" will not use it for instance). If you provide nothing for it, the video part's container format will be used, just like before.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -912,7 +912,7 @@ class YoutubeDL(object):
                             selected_format = {
                                 'requested_formats': formats_info,
                                 'format': rf,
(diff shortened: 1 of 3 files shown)


phihag:
Thanks! Merged.

PR #4682, opened 2015-01-11
Update readme.md configuration paragraph

tordans:
- for me the file path is `~/.config/youtube-dl.conf` which is more consistent with the second URL so I thing this is a good update
- also updated the paragraph for readabilty by separating win and mac
- also extracted the example line to show implicitly that multiple lines in the .conf file are ok.

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,7 +348,16 @@ which means you can modify it, redistribute it or use it however you like.
 
 # CONFIGURATION
 
-You can configure youtube-dl by placing default arguments (such as `--extract-audio --no-mtime` to always extract the audio and not copy the mtime) into `/etc/youtube-dl.conf` and/or `~/.config/youtube-dl/config`. On Windows, the configuration file locations are `%APPDATA%\youtube-dl\config.txt` and `C:\Users\<user name>\youtube-dl.conf`.
+**On Mac:** You can configure youtube-dl by placing default arguments into `/etc/youtube-dl.conf` and/or `~/.config/youtube-dl.conf`.
+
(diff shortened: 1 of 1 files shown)


phihag:
Having a subdirectory seems to be the preferred version in the [XDG specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html), so it should come first. On the other hand, on Windows you'd rather have a nicely-looking file name.

README.md
```diff
@@ -348,7 +348,16 @@ which means you can modify it, redistribute it or use it however you like.
 
 # CONFIGURATION
 
-You can configure youtube-dl by placing default arguments (such as `--extract-audio --no-mtime` to always extract the audio and not copy the mtime) into `/etc/youtube-dl.conf` and/or `~/.config/youtube-dl/config`. On Windows, the configuration file locations are `%APPDATA%\youtube-dl\config.txt` and `C:\Users\<user name>\youtube-dl.conf`.
+**On Mac:** You can configure youtube-dl by placing default arguments into `/etc/youtube-dl.conf` and/or `~/.config/youtube-dl.conf`.
```
phihag: Umm, I think the correct term would be OSX. In any case, this is of course missing FreeBSD, OpenBSD, NetBSD, Solaris, OpenSolaris, and of course the primary platform, Linux. That's why it was written as "everyone uses ~, except Windows".

tordans:
@phihag if the subdirectory way is better, lets at least explain more detailed how that works. For `~/.config/youtube-dl/config`, what would the filename be?

phihag:
Umm, that is the filename, literally `~/.config/youtube-dl/config`.

tordans:
OK, I see. For me this is confusing but since both work, lets just kill the PR …

PR #4690, opened 2015-01-12
[utils] Prevent override of custom headers.

derrotebaron:
The dict of headers of request objects in urllib has its keys always
capitalized.

This causes the lookup to fail and overwrite the header. If for example
a Extractor tries to add a "User-Agent" header the internal
representation in the request object is "User-agent". The header is
therefore clobbered by the "User-Agent" in std_headers, because the
strings are not equal.

### DIFF
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -611,7 +611,9 @@ class YoutubeDLHandler(compat_urllib_request.HTTPHandler):
 
     def http_request(self, req):
         for h, v in std_headers.items():
-            if h not in req.headers:
+            # Capitalize is needed because of Python bug 2275: http://bugs.python.org/issue2275
+            # The dict keys are capitalized because of this bug by urllib
+            if h.capitalize() not in req.headers:
                 req.add_header(h, v)
         if 'Youtubedl-no-compression' in req.headers:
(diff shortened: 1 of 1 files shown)


phihag:
Sorry, I fail to reproduce this issue; everything works fine for me no matter the capitalization. Can you add a test case for this issue?

derrotebaron:
Have a look at https://gist.github.com/derrotebaron/e0712d0d6cfe9ed5843a
If run "nc -l 3000"  and after that "youtube-dl http://localhost:3000" I get the following output:

```
GET / HTTP/1.1
Host: localhost:3000
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Charset: ISO-8859-1,utf-8;q=0.7,*;q=0.7
Accept-Encoding: gzip, deflate
Accept-Language: en-us,en;q=0.5
X-Test: shit
Connection: close
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:10.0) Gecko/20100101 Firefox/10.0 (Chrome)
```

The custom User-Agent is missing without my patch.
I'm currently running Python 3.4.2

phihag:
Thanks! Merged.

PR #4725, opened 2015-01-16
Add megavideoz.eu support.

jbuchbinder:


youtube_dl/extractor/megavideozeu.py
```diff
@@ -0,0 +1,39 @@
...
+from ..utils import (
+    int_or_none,
+    parse_filesize,
+    unified_strdate,
+)
+
+
+class MegavideozeuIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?megavideoz\.eu/video/(?P<id>.*)(?:.*)'
+
```
jaimeMF: Add a test video: https://github.com/rg3/youtube-dl/blob/master/README.md#adding-support-for-a-new-site

phihag:
At http://megavideoz.eu/videos/, it looks like this site specializes in copyright-protected videos, yet I cannot find any sign that they have the rights to distribute this. Since we are missing an obvious non-infringing example video, I regrettably have to close this PR.

PR #4727, opened 2015-01-16
Add travis CI build status badge in readme

PeterDaveHello:


phihag:
I do not think the build status adds anything to the project aside from the page taking longer to load. 

Since youtube-dl includes a lot of extractors, extractors are also bound to go out of date (or the test video gets deleted/updated), so the travis status is red basically all the time, and one needs to read the actual log to understand what is broken.

This does not mean that new commits are affected in any way by the build "failure"; we regularly test all the core parts. Our tree is always open.

PeterDaveHello:
Okay

PR #4745, opened 2015-01-18
Embed description and URL as MP4 tags

BitLooter:
Description is one of the MP4 tags honored by ffmpeg, looks like a good place to store the video description when embedding metadata. I believe the source URL is useful information to keep as well, the comment tag seems like a good place to record it.

While I don't see the description becoming long enough to significantly affect the size of the output video, it can be long and unwieldy, for example while reading mediainfo dumps. Should including it be dependant on another option in addition to --add-metadata, perhaps --add-description?


phihag:
Thanks! Merged.

PR #4758, opened 2015-01-21
[StreetVoice] Add new extractor

yan12125:
tw.streetvoice.com is an online music sharing platform for independent bands in Taiwan.


dstftw:
Thanks, merged.

PR #4768, opened 2015-01-23
download raw videos from LiveLeak (without branding)

jomo:
Removing `.h264_base.mp4` from the URL gives us the raw video, which is essentially the same video without the LiveLeak logo at the top:

before-after image
Actually the video quality seems to be slightly better now as well.

Some of the raw videos are `flv` instead of `mp4`, so they won't always be `mp4` (as it was before). I hope that is okay.

All tests are passing after I changed the expected result to flv for one of the videos.

I have never worked with youtube-dl before, please forgive me if this is the wrong place for this kind of modification.


phihag:
youtube-dl supports multiple video formats, so we can make use of that functionality. That allows users to still pick the mp4 file format.

phihag:
Thanks for pointing this out, by the way! We appreciate your contribution!

PR #4791, opened 2015-01-27
[xuite] Add new extractor

yan12125:
vlog.xuite.net, or Xuite影音, is a video and audio sharing platform hosted by Hinet, largest ISP in Taiwan.


dstftw:
Thanks, merged.

PR #4796, opened 2015-01-28
added extractor for dctp.tv

bastik:
Needs --realtime flag for rtmpdump, otherwise the downloaded video is broken.

The fetched json file has more meta data. Not sure what it is used for, other than bookkeeping, so I leave this as a TODO.

### DIFF
--- a/youtube_dl/downloader/rtmp.py
+++ b/youtube_dl/downloader/rtmp.py
@@ -106,6 +106,7 @@ class RtmpFD(FileDownloader):
         protocol = info_dict.get('rtmp_protocol', None)
         no_resume = info_dict.get('no_resume', False)
         continue_dl = info_dict.get('continuedl', False)
+        real_time = info_dict.get('real_time', False)
 
         self.report_destination(filename)
         tmpfilename = self.temp_name(filename)
@@ -143,6 +144,8 @@ class RtmpFD(FileDownloader):
             basic_args += ['--conn', conn]
         if protocol is not None:
             basic_args += ['--protocol', protocol]
+        if real_time:
+            basic_args += ['--realtime']
 
         args = basic_args
         if not no_resume and continue_dl and not live:
--- a/youtube_dl/downloader/rtmp.py
+++ b/youtube_dl/downloader/rtmp.py
@@ -104,9 +104,9 @@ class RtmpFD(FileDownloader):
         live = info_dict.get('rtmp_live', False)
         conn = info_dict.get('rtmp_conn', None)
         protocol = info_dict.get('rtmp_protocol', None)
+        real_time = info_dict.get('rtmp_real_time', False)
         no_resume = info_dict.get('no_resume', False)
         continue_dl = info_dict.get('continuedl', False)
-        real_time = info_dict.get('real_time', False)
 
         self.report_destination(filename)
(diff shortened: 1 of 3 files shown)


bastik:
There are two ids: One from the URL, which is sort of the title in lower case and with dashes instead of spaces. Then there is a numerical id (e.g. 2418).

Which one would be better for the 'id' parameter in the return value?

youtube_dl/downloader/rtmp.py
```diff
@@ -143,6 +144,8 @@ def run_rtmpdump(args):
             basic_args += ['--conn', conn]
         if protocol is not None:
             basic_args += ['--protocol', protocol]
+        if real_time:
+            basic_args += ['--realtime']
```
dstftw: Works perfectly without `--realtime` for me, downloaded video is fine. Is it really required?

youtube_dl/extractor/dctp.py
```diff
@@ -0,0 +1,48 @@
...
+            'title': 'Videoinstallation für eine Kaufhausfassade'}
+        }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        base_url = 'http://dctp-ivms2-restapi.s3.amazonaws.com/'
+        version_json = self._download_json(base_url + 'version.json', video_id)
+        version = version_json['version_name']
+        info_json = self._download_json(
+            '{}{}/restapi/slugs/{}.json'.format(base_url, version, video_id), video_id)
```
dstftw: This does not work in python 2.6.

dstftw:
The idiomatic way is to put a numeric id into `id` and "sort of the title in lower case and with dashes" into `display_id`.

youtube_dl/downloader/rtmp.py
bastik: > Works perfectly without --realtime for me, downloaded video is fine. Is it really required?

It is like a scratched vinyl record: at some point you see block artifacts and it jumps back in time. At first glance the video looks fine, you have to watch a few seconds to notice this. I would expect this problem to be more pronounced for longer videos, but the one from the test case is also affected.

youtube_dl/downloader/rtmp.py
```diff
@@ -106,6 +106,7 @@ def run_rtmpdump(args):
         protocol = info_dict.get('rtmp_protocol', None)
         no_resume = info_dict.get('no_resume', False)
         continue_dl = info_dict.get('continuedl', False)
+        real_time = info_dict.get('real_time', False)
```
phihag: Please prefix this parameter with `rtmp_` so it's clear it's for RTMP only, unless it makes sense to return a result with `real_time` that does not use RTMP.

bastik:
The test aborts at about 1%:
```
[rtmpdump] DEBUG: Got Play.Complete or Play.Stop from server. Assuming stream is complete
[download]   1.3% of ~3.47MiB at     ---b/s ETA Unknown ETA
```
Nevertheless, it reports status OK.
(No problems when starting the download normally.)

youtube_dl/downloader/rtmp.py
bastik: Done.

phihag:
Thanks, merged!

PR #4813, opened 2015-01-29
UX: Add a note about the help if no parameters are passed

valioDOTch:
When I tried to display the help, the necessary command wasn't intuitively clear to me.
I would suggest to add a small note how one can display the help if no parameters are passed.


phihag:
Thanks for the notice! The proposed wording is somewhat nonstandard, so I changed it slightly.

valioDOTch:
Thanks!

Issue #4807, opened 2015-01-28
ssl certificate verify failed, on freebsd 10.1


satan1st:
Hello, on FreeBSD youtube-dl fails, if you request a https-site:
```
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['-v', '--no-check-certificate', 'http://youtu.be/fNGho63QYT0']
[debug] Encodings: locale UTF-8, fs UTF-8, out None, pref UTF-8
[debug] youtube-dl version 2015.01.25
[debug] Python version 2.7.9 - FreeBSD-10.1-RELEASE-amd64-64bit-ELF
[debug] exe versions: ffmpeg 2.3.6, ffprobe 2.3.6, rtmpdump 2.4
[debug] Proxy map: {}
[youtube] fNGho63QYT0: Downloading webpage
ERROR: Unable to download webpage: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:581)> (caused by URLError(SSLError(1, u'[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:581)'),))
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 299, in _request_webpage
    return self._downloader.urlopen(url_or_request)
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 1571, in urlopen
    return self._opener.open(req, timeout=self._socket_timeout)
  File "/usr/local/lib/python2.7/urllib2.py", line 431, in open
    response = self._open(req, data)
  File "/usr/local/lib/python2.7/urllib2.py", line 449, in _open
    '_open', req)
  File "/usr/local/lib/python2.7/urllib2.py", line 409, in _call_chain
    result = func(*args)
  File "/usr/local/bin/youtube-dl/youtube_dl/utils.py", line 659, in https_open
    req)
  File "/usr/local/lib/python2.7/urllib2.py", line 1197, in do_open
    raise URLError(err)
```
I have run git bisect and found out the commit be4a824d74add1a3b78b8244dff12f4f078f168a is the problem.

k0ral:
I confirm I have the exact same issue here on Linux.

jaimeMF:
The '--no-check-certificate' should work again with the fix in #4815. But I wouldn't recommend using it.

[...]

PR #4815, opened 2015-01-29
[utils] YoutubeDLHTTPSHandler.https_open: pass all required arguments to to do_open

jaimeMF:
With this change the '--no-check-certificate' works again (#4807).

/cc @phihag.


phihag:
I've added a test in 83fda3c000a680317cfc9bb6fec899beb8bca773 for this.

PR #4831, opened 2015-01-31
Handling Connection Reset by Peer Error

light94:
"Connection Reset By Peer" error is frequent in when connecting through a firewall and retrying download often is a workaround for this.


phihag:
Thanks! Merged.

Issue #4346, opened 2014-12-01
Globo.com video download not working


tbs:
> C:\files>youtube-dl.exe -v http://globotv.globo.com/rede-globo/paralimpiadas/t/e
dicoes/v/mulheres-espetaculares-terezinha-guilhermina-a-atleta-paralimpica-mais-
rapida-do-mundo/3799985/
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['-v', 'http://globotv.globo.com/rede-globo/paralimpi
adas/t/edicoes/v/mulheres-espetaculares-terezinha-guilhermina-a-atleta-paralimpi
ca-mais-rapida-do-mundo/3799985/']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2014.12.01
[debug] Python version 2.7.8 - Windows-8-6.2.9200
[debug] exe versions: none
[debug] Proxy map: {}
[Globo] rede-globo/paralimpiadas/t/edicoes/v/mulheres-espetaculares-terezinha-gu
ilhermina-a-atleta-paralimpica-mais-rapida-do-mundo/3799985/: Downloading webpag
e
[Globo] 3799985: Downloading JSON metadata
[Globo] 3799985: Downloading security hash for 547b3f859214474800001b12
[Globo] 3799985: Downloading security hash for 547b3f8c92144733f2002934
[Globo] 3799985: Downloading security hash for 547b3f9a9214474800001b14
[Globo] 3799985: Downloading security hash for 547b3f8b9214472b1c000001_m3u8
Traceback (most recent call last):
  File "__main__.py", line 19, in <module>
  File "youtube_dl\__init__.pyo", line 355, in main
  File "youtube_dl\__init__.pyo", line 345, in _real_main
  File "youtube_dl\YoutubeDL.pyo", line 1117, in download
  File "youtube_dl\YoutubeDL.pyo", line 553, in extract_info
  File "youtube_dl\extractor\common.pyo", line 241, in extract
  File "youtube_dl\extractor\globo.pyo", line 387, in _real_extract
KeyError: u'height'

Tested this video http://globotv.globo.com/rede-globo/paralimpiadas/t/edicoes/v/mulheres-espetaculares-terezinha-guilhermina-a-atleta-paralimpica-mais-rapida-do-mundo/3799985/

Thanks

kinetoskombi:
[...]

PR #4832, opened 2015-01-31
[globo] Fix error on some globo videos #4346

kinetoskombi:
I've fixed an error (#4346) on some globo videos. The problem was that some videos have a m3u playlist listed on the format's list. I'm new to python and youtube-dl, so I don't know if my fix is the best possible.


dstftw:
Thanks, merged.

PR #4845, opened 2015-02-02
Passing source address option to external downloaders

vijayanandnandam:
Currently Youtube-dl is not passing the `source-address` to external downloaders like wget, curl and aria2c.
This pull request adds that support


dstftw:
Thanks, merged.

Issue #4860, opened 2015-02-04
progress hook not called on some videos


xantares:
hello,
I'm writing a GUI to browse/download some videos (https://github.com/xantares/maxitube) based on yt-dl.
I'm trying to display the progress of the download, but on some videos (from canalplus for example) the progress hook is only called at the end of download (although I can see the curses progress percentage in the console) with the 'finished' state:
http://www.canalplus.fr/c-divertissement/c-le-petit-journal/pid6515-le-petit-journal.html?vid=1206051
whereas on some other sites the progress hook is called with the percentage of progress along with the 'downloading' state:
http://www.goldenmoustache.com/le-lab-la-drogue-arthur-laloux-56283/
any clue ?

phihag:
This is a bug / missing feature. I'll leave this issue open as a feature request (for hook support in the f4m/hls/rtmp downloader).

xantares:
seems f4m has code related to progress hook.
is it hls or rtmp ?
do you think it would be easy to implement ?

xantares:
fixed by https://github.com/rg3/youtube-dl/pull/4973

PR #4875, opened 2015-02-05
[f4m] call progress hook

xantares:
the actual progress hooks were not called (#4860)
however, I'd like the fields downloaded_bytes and total_bytes be computed correctly from the progress of each frag.
the sum ot total bytes downloaded seems recorded in state['downloaded_bytes'],
but how would you account for total_bytes ? it seems fragments don't have the same size, do we know it in advance ?
otherwise we could add a 'progress' field in the status with the ratio downloaded_bytes/total_bytes when total_bytes is known or else the estimate of f4m.
should I add a get_progress_hooks() to return _progress_hooks ?

### DIFF
--- a/youtube_dl/downloader/f4m.py
+++ b/youtube_dl/downloader/f4m.py
@@ -335,8 +335,10 @@ class F4mFD(FileDownloader):
             # report totals to actual hooks
             toplevel_status = {'status': 'downloading', 'filename': filename, 'tmpfilename': tmpfilename}
             toplevel_status['downloaded_bytes'] = state['downloaded_bytes']
-            if (state['downloaded_bytes'] > 0) and (progress > 0):
-                toplevel_status['total_bytes'] = int(100.0 * state['downloaded_bytes'] / progress)
+            total_bytes = 0
+            if progress > 0:
+                total_bytes = int((100.0 * state['downloaded_bytes']) / progress)
+            toplevel_status['total_bytes'] = total_bytes
             if 'speed' in status:
                 toplevel_status['speed'] = status['speed']
             if 'eta' in status:
--- a/youtube_dl/downloader/f4m.py
+++ b/youtube_dl/downloader/f4m.py
@@ -333,11 +333,12 @@ class F4mFD(FileDownloader):
                                  status.get('speed'), eta)
 
             # report totals to actual hooks
-            status['status'] = 'downloading'
-            #status['downloaded_bytes'] = state['downloaded_bytes'] ?
-            #status['total_bytes'] = ?
+            toplevel_status = {'status': 'downloading', 'filename': filename, 'tmpfilename': tmpfilename}
+            toplevel_status['downloaded_bytes'] = state['downloaded_bytes']
+            if (state['downloaded_bytes'] > 0) and (progress > 0):
+                toplevel_status['total_bytes'] = int(100.0 * state['downloaded_bytes'] / progress)
             for hook in self._progress_hooks:
-                hook(status)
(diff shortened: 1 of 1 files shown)


xantares:
I get the total from the progress ratio.
Works ok now!

xantares:
Someone for a review ?

youtube_dl/downloader/f4m.py
```diff
@@ -331,6 +331,21 @@ def frag_progress_hook(status):
             eta = self.calc_eta(start, time.time(), estimated_size, byte_counter)
             self.report_progress(progress, format_bytes(estimated_size),
                                  status.get('speed'), eta)
+
+            # report totals to actual hooks
+            toplevel_status = {'status': 'downloading', 'filename': filename, 'tmpfilename': tmpfilename}
+            toplevel_status['downloaded_bytes'] = state['downloaded_bytes']
+            total_bytes = 0
+            if progress > 0:
```
phihag: `progress` is a value for human consumption. I'm pretty sure you meant `state['downloaded_bytes']` here.

youtube_dl/downloader/f4m.py
```diff
@@ -331,6 +331,21 @@ def frag_progress_hook(status):
...
+                total_bytes = int((100.0 * state['downloaded_bytes']) / progress)
+            toplevel_status['total_bytes'] = total_bytes
+            if 'speed' in status:
+                toplevel_status['speed'] = status['speed']
+            if 'eta' in status:
+                toplevel_status['eta'] = status['eta']
```
phihag: This doesn't look correct. While is the ETA only dependent on the current status?

youtube_dl/downloader/f4m.py
```diff
@@ -331,6 +331,21 @@ def frag_progress_hook(status):
...
             self.report_progress(progress, format_bytes(estimated_size),
                                  status.get('speed'), eta)
+
+            # report totals to actual hooks
+            toplevel_status = {'status': 'downloading', 'filename': filename, 'tmpfilename': tmpfilename}
+            toplevel_status['downloaded_bytes'] = state['downloaded_bytes']
+            total_bytes = 0
+            if progress > 0:
+                total_bytes = int((100.0 * state['downloaded_bytes']) / progress)
+            toplevel_status['total_bytes'] = total_bytes
```
phihag: We should indicate that this is only a guess - the value may be smaller or larger than the actual size.

youtube_dl/downloader/f4m.py
jaimeMF: You should actually use the `eta` calculated in line 331, the eta in `status` is for the fragment.

youtube_dl/downloader/f4m.py
```diff
@@ -331,6 +331,21 @@ def frag_progress_hook(status):
...
+            if 'speed' in status:
+                toplevel_status['speed'] = status['speed']
+            if 'eta' in status:
+                toplevel_status['eta'] = status['eta']
+            for hook in self._progress_hooks:
```
jaimeMF: Simply call `self._hook_progress(...)`, it will loop over the hooks for you.

xantares:
ok.

xantares:
ok, fixed

phihag:
Thanks, I've adapted this in my rework of the general progress reporting framework.

xantares:
That's way better, I got the percentage directly, but with a leading whitespace :/
'_percent_str': ' 64.1%'
Thanks a lot!

PR #4881, opened 2015-02-05
Tweakers

robin007bond:
Added support for Dutch IT site Tweakers.net

Example: https://tweakers.net/video/9926/new-nintendo-3ds-xl-op-alle-fronten-beter.html

I don't know about the _VALID_URL regex, and I could done the URL parsing cleaner by using regex instead of split.

### DIFF
--- a/youtube_dl/extractor/tweakers.py
+++ b/youtube_dl/extractor/tweakers.py
@@ -25,10 +25,10 @@ class TweakersIE(InfoExtractor):
 
     def _real_extract(self, url):
         splitted_url = re.split('.html|/', url)
-        del splitted_url[-1] # To remove extra '/' at the end
+        del splitted_url[-1]  # To remove extra '/' at the end
         video_id = splitted_url[4]
-        title = splitted_url[5].title() # Retrieve title for URL and capitalize
-        splitted_url[3] = splitted_url[3] + '/player' # Add /player to get the player page
+        title = splitted_url[5].title()  # Retrieve title for URL and capitalize
+        splitted_url[3] = splitted_url[3] + '/player'  # Add /player to get the player page
         player_url = '/'.join(splitted_url) + '.html'
         player_page = self._download_webpage(player_url, video_id)
 
@@ -36,6 +36,6 @@ class TweakersIE(InfoExtractor):
             'id': video_id,
             'ext': 'mp4',
             'title': title,
-            'url':  re.findall('http.*mp4', player_page)[0],
+            'url': re.findall('http.*mp4', player_page)[0],
             'player_url': player_url
         }
--- a/youtube_dl/extractor/tweakers.py
+++ b/youtube_dl/extractor/tweakers.py
@@ -37,5 +37,4 @@ class TweakersIE(InfoExtractor):
             'ext': 'mp4',
             'title': title,
             'url': re.findall('http.*mp4', player_page)[0],
-            'player_url': player_url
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/tweakers.py
```diff
@@ -0,0 +1,41 @@
...
+class TweakersIE(InfoExtractor):
+    _VALID_URL = r'https?://tweakers\.net/video/(?P<id>[0-9]+).*'
+    _TEST = {
+        'url': 'https://tweakers.net/video/9926/new-nintendo-3ds-xl-op-alle-fronten-beter.html',
+        'md5': 'f7f7f3027166a7f32f024b4ae6571ced',
+        'info_dict': {
+            'id': '9926',
+            'ext': 'mp4',
+            'title': 'New-Nintendo-3Ds-Xl-Op-Alle-Fronten-Beter',
+            # TODO more properties, either as:
```
dstftw: Do or remove otherwise.

youtube_dl/extractor/tweakers.py
```diff
@@ -0,0 +1,41 @@
...
+        title = splitted_url[5].title()  # Retrieve title for URL and capitalize
+        splitted_url[3] = splitted_url[3] + '/player'  # Add /player to get the player page
+        player_url = '/'.join(splitted_url) + '.html'
+        player_page = self._download_webpage(player_url, video_id)
+
+        return {
+            'id': video_id,
+            'ext': 'mp4',
+            'title': title,
+            'url': re.findall('http.*mp4', player_page)[0],
```
dstftw: There are more video formats available that should be extracted as well.

youtube_dl/extractor/tweakers.py
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    def _real_extract(self, url):
+        splitted_url = re.split('.html|/', url)
+        del splitted_url[-1]  # To remove extra '/' at the end
+        video_id = splitted_url[4]
+        title = splitted_url[5].title()  # Retrieve title for URL and capitalize
+        splitted_url[3] = splitted_url[3] + '/player'  # Add /player to get the player page
+        player_url = '/'.join(splitted_url) + '.html'
+        player_page = self._download_webpage(player_url, video_id)
```
dstftw: Why not just use
```
player_page = self._download_webpage(url, video_id)
```

robin007bond: Oh man, I feel so stupid now. I thought it was necessary to get the exact player page, because I didn't notice thje og:video tags.. It seemed the only way to get the exact video url...

youtube_dl/extractor/tweakers.py
```diff
@@ -0,0 +1,41 @@
...
+
+        return {
+            'id': video_id,
+            'ext': 'mp4',
+            'title': title,
+            'url': re.findall('http.*mp4', player_page)[0],
+            'player_url': player_url
```
dstftw: This should be removed. `player_url` is used for RTMP.

robin007bond:
I commited some changes, but there are still things a bit unclear for me.

Downloading the regular page: https://tweakers.net/video/9926/new-nintendo-3ds-xl-op-alle-fronten-beter.html, with:
`player_page = self._download_webpage(url, video_id)`

dstftw:
Thanks, merged.

PR #4907, opened 2015-02-08
[nationalarchivesuk] Add nationalarchives.gov.uk

robin007bond:
Added the nationalarchives.gov.uk extractor


phihag:
The implementation looks to generic to not be included in the generic extractor.

phihag:
Thank you for the pull request; I've simply amended the generic extractor to catch this video as well.

robin007bond:
That's a great solution and adds support for more sites that initialize JWPlayer this way.

PR #4938, opened 2015-02-12
Add support for camdemy.com

yan12125:
Camdemy is a teaching video sharing platform in Taiwan.


phihag:
Thank you very much! Merged with minor adaptions.

PR #4943, opened 2015-02-13
[Yam] Add new extractor

yan12125:
Yam.com is an audio and video sharing platform in Taiwan [http://mymedia.yam.com]


phihag:
Thank you very much! Merged.

PR #4946, opened 2015-02-13
Change status message 'video' to 'media'

robin007bond:
Changed occurences of 'video' to 'media', since youtube-dl supports audio files too, not only videos.

Sorry for the branch name, it's because of using GitHub's builtin editor.


phihag:
While it's technically correct, it feels awfully stilted. Why not instead add a property to the playlist object and motdify the few extractors that don't return videos to set that property?

robin007bond:
Agreed!

Issue #4019, opened 2014-10-24
Automatically extract ASS subs when available


ChrisK2:
(See #4003 for context)

We currently have to know beforehand if we want to have ASS subs, which requires a site-specific check, which we'd like to avoid. It would be good if ytdl would output ASS subs by default when available (in the case of CR, the SRT subs are completely broken anyway), or if it would be possible to specify an order of prefered formarts, just like for video:
``--sub-format=ass/srt``

jaimeMF:
Thanks for the suggestion, it will work with the syntax you proposed in the next version (and in the case of crunchyroll it will default to ass).

Issue #3804, opened 2014-09-20
--sub-lang doesn't seem to work


foresto:
When I use the `--sub-lang en` option, youtube-dl says it is downloading subtitles, but no subtitle file is created. When I combine that option with `--write-sub`, all the subtitle files are downloaded, not just the language I selected.

I'm testing with this page:
http://www.crunchyroll.com/yamishibai-japanese-ghost-stories/episode-2-zanbai-641889

This is on version 2014.08.05.

jaimeMF:
Thanks for the report, in the next version only the requested language file will be written.

phihag:
I believe this issue to be fixed in youtube-dl version 2015.02.24 and newer. See [our FAQ](https://yt-dl.org/update) if you need help updating.

PR #4964, opened 2015-02-16
Subtitles improvements

jaimeMF:
Basically: for each language we give a list containing all the formats available, and `YoutubeDL` is now in charge of selecting the language and formats to download.

This tries to address #4019 and #3804.

Here's a list of the extractor that need to be converted:
* [x] atresplayer
* [x] bbccouk
* [x] bliptv
* [x] ceskatelevize
* [x] crunchyroll
* [x] dailymotion
* [x] drtv
* [x] lynda
* [x] mit
* [x] mtv
* [x] npo
* [x] nrk
* [x] rai
* [x] ted
* [x] theplatform
* [x] viki
* [x] vimeo
* [x] walla
* [x] youtube


jaimeMF:
In case you don't have any objection/suggestion, don't merge this. I would like to migrate all the extractors first, and since it changes how the subtitles are stored in the json output I want to talk to the mpv team so that they can update the youtube-dl support (if you know of any project that uses it, we may also want to contact them).

youtube_dl/YoutubeDL.py
```diff
@@ -1586,6 +1655,18 @@ def list_thumbnails(self, info_dict):
             ['ID', 'width', 'height', 'URL'],
             [[t['id'], t.get('width', 'unknown'), t.get('height', 'unknown'), t['url']] for t in thumbnails]))
 
+    def list_subtitles(self, video_id, subtitles, name='subtitles'):
+        if not subtitles:
+            self.to_screen('%s has no %s' % (video_id, name))
+            return
+        header_line = 'Language    formats'
+        sub_lines = [
```
phihag: Can we use `render_table` from utils here?

phihag:
LGTM, feel free to merge when you think it's ready.

youtube_dl/YoutubeDL.py
jaimeMF: Yes, I forgot to change it, I wrote this part before you added the function.

jaimeMF: Changed in edab9dbf4d00a7f76fbfd2df9ef4b205c88e47a8

jaimeMF:
I have converted all the extractors, I'll probably add/clean some test and merge it this weekend.

PR #4973, opened 2015-02-17
[YoutubeDL] Use a progress hook for progress reporting

phihag:
Instead of every downloader calling two helper functions, let our progress report be an ordinary progress hook like everyone else's. Amend the documentation and add a new "error" case.

Closes #4875.


jaimeMF:
LGTM.

PR #4998, opened 2015-02-19
[imgur] Support regular URL

jbboehr:
Just one minor improvement to 3bf5705. It should support extracting a video from imgur's normal URL format: e.g. https://imgur.com/A61SaA1

### DIFF
--- a/youtube_dl/extractor/imgur.py
+++ b/youtube_dl/extractor/imgur.py
@@ -11,7 +11,7 @@ from ..utils import (
 )
 
 class ImgurIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:i\.)?imgur\.com/(?P<id>[a-zA-Z0-9]+)(?:\.)?(?:mp4|gifv)?'
+    _VALID_URL = r'https?://(?:i\.)?imgur\.com/(?P<id>[a-zA-Z0-9]+)(?:\.mp4|\.gifv)?'
 
     _TESTS = [{
         'url': 'https://i.imgur.com/A61SaA1.gifv',
--- a/youtube_dl/extractor/imgur.py
+++ b/youtube_dl/extractor/imgur.py
@@ -5,6 +5,7 @@ import re
 from .common import InfoExtractor
 from ..utils import (
     int_or_none,
+    str_or_none,
     js_to_json,
     mimetype2ext,
     ExtractorError,
@@ -35,11 +36,6 @@ class ImgurIE(InfoExtractor):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
 
-        sources = re.findall(r'<source src="([^"]+)" type="([^"]+)"', webpage)
-        if not sources:
-            raise ExtractorError(
-                'No sources found for video %s' % video_id, expected=True)
-
         width = int_or_none(self._search_regex(
             r'<param name="width" value="([0-9]+)"',
             webpage, 'width', fatal=False))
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/imgur.py
```diff
@@ -21,12 +21,25 @@ class ImgurIE(InfoExtractor):
...
+            'title': 'MRW gifv is up and running without any bugs',
+            'description': 'The Internet\'s visual storytelling community. Explore, share, and discuss the best visual stories the Internet has to offer.',
+        },
     }]
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
 
+        sources = re.findall(r'<source src="([^"]+)" type="([^"]+)"', webpage)
```
phihag: Why is this code necessary or useful? What test case does fail if it gets removed?

youtube_dl/extractor/imgur.py
```diff
@@ -7,11 +7,11 @@
     int_or_none,
     js_to_json,
     mimetype2ext,
+    ExtractorError,
 )
 
-
 class ImgurIE(InfoExtractor):
-    _VALID_URL = r'https?://i\.imgur\.com/(?P<id>[a-zA-Z0-9]+)\.(?:mp4|gifv)'
+    _VALID_URL = r'https?://(?:i\.)?imgur\.com/(?P<id>[a-zA-Z0-9]+)(?:\.)?(?:mp4|gifv)?'
```
phihag: It looks like you want to make the extension optional. If so, please move the dot into the group of extensions - otherwise, it's not obvious what is getting matched here.

youtube_dl/extractor/imgur.py
jbboehr: I borrowed it from theonion extractor, which it appears you just refactored: https://github.com/rg3/youtube-dl/commit/e880c66bd800e4769ffdff41fb9dc88cf3206c12

jbboehr: The idea was to display a reasonable error message if the user attempted to download a non-image.

jbboehr: This is what happens if you try to download an image from imgur instead of a video:

```
python -m youtube_dl https://imgur.com/hL8R5YG
[Imgur] hL8R5YG: Downloading webpage
WARNING: unable to extract width; please report this issue on http://yt-dl.org/bug
WARNING: unable to extract height; please report this issue on http://yt-dl.org/bug
ERROR: Unable to extract video elements; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

phihag: Alright, isn't it possible to just catch this case a little bit nicer? I.e. look at `formats` at an opportune moment, and give out a better error message.

jbboehr: I don't really have a strong preference on the failure mode; it's whatever you want.

If I remove this block, you get this:

```
  File "youtube_dl/YoutubeDL.py", line 643, in extract_info
    ie_result = ie.extract(url)
  File "youtube_dl/extractor/common.py", line 269, in extract
  File "youtube_dl/extractor/imgur.py", line 48, in _real_extract
    webpage, 'video elements')
  File "youtube_dl/extractor/common.py", line 531, in _search_regex
```

If I replace it with this:

```
        formats = []
        video_elements = str_or_none(self._search_regex(
            r'(?s)<div class="video-elements">(.*?)</div>',
            webpage, 'video elements', fatal=False))
        if not video_elements:
            raise ExtractorError(
                'No sources found for video %s' % video_id, expected=True)
```

Then you get:
```
python -m youtube_dl https://imgur.com/hL8R5YG
[Imgur] hL8R5YG: Downloading webpage
WARNING: unable to extract width; please report this issue on http://yt-dl.org/bug
WARNING: unable to extract height; please report this issue on http://yt-dl.org/bug
WARNING: unable to extract video elements; please report this issue on http://yt-dl.org/bug
ERROR: No sources found for video hL8R5YG
```

phihag:
Thanks, merged.

PR #5019, opened 2015-02-20
Change example URLs in readme (fixes #5018)

### LINKED ISSUE #5018: Example of URL change in the readme does not change
Line 574 of README.md gives an example of a site's URL structure changing from http://example.com/v/1234567 to http://example.com/v/1234567 however both URLs are identical.

cyberjacob:


phihag:
Great catch! Merged.

PR #5023, opened 2015-02-21
Removing Non Ascii for Windows filenames

meshr-net:
windows can't save file names with Non Ascii characters


dstftw:
We already have a generic solution for that namely `--restrict-filenames`.
Nevertheless, Windows can handle non-ASCII. Could you name an example URL that results in failure due to this?

phihag:
In addition or instead of an example URL, the output of youtube-dl when called with `-v` would also be quite helpful, since it will tell us about the system configuration, namely the filesystem charsets.

PR #5030, opened 2015-02-23
[airmozilla] Add new extractor

duncankl:
Air Mozilla is where Mozilla's live videos are saved. This extractor doesn't attempt to grab the live streams yet, just the archived videos.


phihag:
Thank you very much! Merged.

duncankl:
Great! Thanks for cleaning up the patch, I learned lots from your changes.

PR #5055, opened 2015-02-24
[Letv] Add new extractor

yan12125:
Letv.com is a video publishing platform in China


phihag:
Thank you very much! Merged.

PR #5056, opened 2015-02-24
Eporner Fix

logon84:
Hi. I updated the eporner extractor to work. That are my first steps at github, I hope that I'm doing this well...


dstftw:
Thanks, merged.

PR #5057, opened 2015-02-24
Live F4M streams and support for WDR live TV

aajanki:
This adds basic support for live streams to the F4M downloader.

Specifically, this is needed for downloading the WDR live stream, but I think other sites will benefit as well. WDR live example:

> python -m youtube_dl --no-part http://www1.wdr.de/mediathek/video/livestream/index.html


phihag:
Thank you very much! Merged.

PR #5076, opened 2015-02-26
[lynda] Fixed subtitles broken file

Ftornik:
Ajax subtitle request returns text captions with new line characters in the beginning and the end.

    '\nIn the next several chapters of this\ncourse, we're going to be going over the \n'

Result .srt file has bad format and ignored by VLC and some other players.
Example:

    1
    00:00:00.025 --> 00:00:04.939
    
    In the next several chapters of this 
    course, we're going to be going over the 

To fix this bug, we need to remove empty line after timecode

    1
    00:00:00.025 --> 00:00:04.939    
    In the next several chapters of this 
    course, we're going to be going over the 

This is done with **strip** function

### DIFF
--- a/youtube_dl/extractor/lynda.py
+++ b/youtube_dl/extractor/lynda.py
@@ -152,7 +152,7 @@ class LyndaIE(InfoExtractor):
                 continue
             appear_time = m_current.group('timecode')
             disappear_time = m_next.group('timecode')
-            text = seq_current['Caption']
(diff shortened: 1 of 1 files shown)


jaimeMF:
Can you give an example url?, for testing it.

dstftw:
Thanks, merged.
Example URL is https://www.lynda.com/player/embed/133770?tr=foo=1;bar=g;fizz=rt&fs=0.

PR #5077, opened 2015-02-27
Multiple proxies

### LINKED ISSUE #3391: Separate extractor proxy and download proxy
Some china video sharing sites such as Youku and Sohu limit some of their vidoes available only in Mainland China. In such cases, I need a proxy in China to get the video. For instance, proxy.uku.im:8888, the proxy used by a popular browser plugin Unblock Youku. However, this proxy allow only addresses for checking China IP. Take Sohu as example, hot.vrs.letv.com is allowed while tv.sohu.com not. Is it possible to use two different proxies, one for extracting video info and the other for real downloading? A similar project named you-get [1] has this ability.
[1] https://github.com/soimort/you-get

yan12125:
This patch closes #3391 I opened long time ago. Such a feature is used in letv.com extractor.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -301,7 +301,7 @@ class InfoExtractor(object):
     def IE_NAME(self):
         return type(self).__name__[:-2]
 
-    def _request_webpage(self, url_or_request, video_id, note=None, errnote=None, fatal=True):
+    def _request_webpage(self, url_or_request, video_id, note=None, errnote=None, fatal=True, opener_name=None):
         """ Returns the response handle """
         if note is None:
             self.report_download_webpage(video_id)
@@ -311,7 +311,7 @@ class InfoExtractor(object):
             else:
                 self.to_screen('%s: %s' % (video_id, note))
         try:
-            return self._downloader.urlopen(url_or_request)
+            return self._downloader.urlopen(url_or_request, opener_name=opener_name)
         except (compat_urllib_error.URLError, compat_http_client.HTTPException, socket.error) as err:
             if errnote is False:
                 return False
@@ -324,13 +324,13 @@ class InfoExtractor(object):
                 self._downloader.report_warning(errmsg)
                 return False
 
-    def _download_webpage_handle(self, url_or_request, video_id, note=None, errnote=None, fatal=True):
+    def _download_webpage_handle(self, url_or_request, video_id, note=None, errnote=None, fatal=True, opener_name=None):
         """ Returns a tuple (page content as string, URL handle) """
         # Strip hashes from the URL (#1038)
         if isinstance(url_or_request, (compat_str, str)):
             url_or_request = url_or_request.partition('#')[0]
 
-        urlh = self._request_webpage(url_or_request, video_id, note, errnote, fatal)
+        urlh = self._request_webpage(url_or_request, video_id, note, errnote, fatal, opener_name=opener_name)
         if urlh is False:
             assert not fatal
             return False
@@ -410,13 +410,13 @@ class InfoExtractor(object):
 
         return content
 
-    def _download_webpage(self, url_or_request, video_id, note=None, errnote=None, fatal=True, tries=1, timeout=5):
+    def _download_webpage(self, url_or_request, video_id, note=None, errnote=None, fatal=True, tries=1, timeout=5, opener_name=None):
         """ Returns the data of the page as a string """
         success = False
         try_count = 0
         while success is False:
             try:
-                res = self._download_webpage_handle(url_or_request, video_id, note, errnote, fatal)
+                res = self._download_webpage_handle(url_or_request, video_id, note, errnote, fatal, opener_name=opener_name)
                 success = True
             except compat_http_client.IncompleteRead as e:
                 try_count += 1
@@ -431,10 +431,10 @@ class InfoExtractor(object):
 
     def _download_xml(self, url_or_request, video_id,
                       note='Downloading XML', errnote='Unable to download XML',
-                      transform_source=None, fatal=True):
+                      transform_source=None, fatal=True, opener_name=None):
         """Return the xml as an xml.etree.ElementTree.Element"""
         xml_string = self._download_webpage(
-            url_or_request, video_id, note, errnote, fatal=fatal)
+            url_or_request, video_id, note, errnote, fatal=fatal, opener_name=opener_name)
         if xml_string is False:
             return xml_string
         if transform_source:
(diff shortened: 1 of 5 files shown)


jaimeMF:
Are cookies shared by the two openers? If not, that can be problematic in some cases.

youtube_dl/options.py
```diff
@@ -177,6 +177,11 @@ def _hide_login_info(opts):
         default=None, metavar='URL',
         help='Use the specified HTTP/HTTPS proxy. Pass in an empty string (--proxy "") for direct connection')
     network.add_option(
+        '--alternative-proxy', dest='alternative_proxy',
```
jaimeMF: Maybe rename it to `--extraction-proxy` and update the help message to something like `Proxy used for extracting the video info`, to make it clearer what it does.

youtube_dl/extractor/common.py
```diff
@@ -1053,6 +1053,15 @@ def extract_automatic_captions(self, *args, **kwargs):
     def _get_automatic_captions(self, *args, **kwargs):
         raise NotImplementedError("This method must be implemented by subclasses")
 
+    def _multiple_opener_supported(self):
+        return hasattr(self._downloader, 'use_opener')
```
jaimeMF: This can probably be removed, nobody is going to use their own class for `_downloader`.

yan12125:
Is there a site requiring consistent cookies for correct downloading?

yan12125:
According to [1], cookies are consistent in my approach. All openers use the same cookie jar ```self.cookiejar```.

[1] http://stackoverflow.com/questions/11131067/share-the-cookies-between-two-openers-with-urllib2-and-cookielib

youtube_dl/options.py
yan12125: There is something tricky. Take sohu.py (at yan12125/youtube-dl@f7b8b0c39e8fd6dfdddb85b5bed2812d3fd68d69) for example, not all traffic for info extraction need China IP. Only the one with hostname hot.vrs.sohu.com needs it. So the name extraction proxy may be misleading.

youtube_dl/extractor/common.py
yan12125: Thanks. Removed in yan12125/youtube-dl@603251c34d6421ff7c9069630c4abcfbde7b7e16

youtube_dl/options.py
jaimeMF: Instead of doing that on each extractor, I think it should be done for all of them (for example in the `InfoExtractor.extract` method). It would be more consistent.

jaimeMF:
But you are creating a new cookiejar for each opener (at [this line](https://github.com/yan12125/youtube-dl/blob/00b621a8906791d251dd784982fa40db79ac36eb/youtube_dl/YoutubeDL.py#L1754)), I don't think they will have the same cookies.

It will probably also break [saving the cookies in a file](https://github.com/yan12125/youtube-dl/blob/00b621a8906791d251dd784982fa40db79ac36eb/youtube_dl/YoutubeDL.py#L462).

> Is there a site requiring consistent cookies for correct downloading?

niconico (I think)

phihag:
I do not believe this implementation solves the actual issue at hand. If I understand [the rationale](https://github.com/rg3/youtube-dl/issues/3391#issuecomment-76411032), correctly, the proxy is just needed for one step.

Therefore, there should be an option `--cn-verification-proxy` (or `--sohu-verification-proxy`), and that proxy should be used at the particular step and nowhere else. In particular, I see absolutely no reason to slow down the default download steps.

It is with a heavy heart that I must reject this pull request in the current form. However, you are very welcome to suggest a similar one that leaves the default download code untouched, documents that this proxy is just needed for one step, and uses the proxy in that step and nowhere else.

youtube_dl/extractor/letv.py
```diff
@@ -76,10 +91,15 @@ def _real_extract(self, url):
             'tkey': self.calc_time_key(int(time.time())),
             'domain': 'www.letv.com'
         }
+
+        self._use_opener('alternative')
```
phihag: This introduces an implicit global state. Please do not use implicit global state. Instead, a parameter for _download_* or another helper function seems way more reasonable.

yan12125:
Thanks for the suggestions. I managed to fix them in https://github.com/yan12125/youtube-dl/tree/multiple_proxies. Documentations are not updated yet, however.

@phihag Sorry I don't quite understand why this approach slows down normal downloads. I believe the original download helpers work semantically the same as the past except some tiny overhead arising from parameter passing.

phihag:
Oh, you are right, there will no be significant slowdown. Nevertheless, the global state quarry remains, this is an absolute no-go right as we are considering multi-threading/-processing.

yan12125:
In yan12125/youtube-dl@63367c442bb1cd0a9f90b4e8a18051ac26ee315f, there's no more global affects. Web requests, either json, xml, or generic web pages, use their own openers specified as the function parameter. Also, I replace ```self._opener``` with ```self._opener_pools[opener_name]``` in the YoutubeDL class so that different web requests doesn't interfere with others with respect to the opener they use (blame me if I miss something still overlapping). I hope this modification solves the problem.

yan12125:
Documentations updated. Additionally, I confirm that cookies are consistent among openers with the following test: (Is it proper to put the test here?) Tests for thread-safe seems more complicated and I have no idea how.

```Diff
diff --git a/test/test_InfoExtractor.py b/test/test_InfoExtractor.py
index be8d129..860f944 100644
--- a/test/test_InfoExtractor.py
+++ b/test/test_InfoExtractor.py
@@ -58,5 +58,21 @@ class TestInfoExtractor(unittest.TestCase):
         self.assertEqual(ie._html_search_meta('e', html), '5')
         self.assertEqual(ie._html_search_meta('f', html), '6')
 
+    def test_consistent_cookies_among_openers(self):
+        video_id = 'fake_id'
+        httpbin_cookies_url = 'http://httpbin.org/cookies'
+        cookie_name = 'cookie_name'
+        cookie_value = 'cookie_value'
+
+        self.ie._download_webpage(
+            '%s/set?%s=%s' % (httpbin_cookies_url, cookie_name, cookie_value),
+            video_id, opener_name='default')
+        result_cookies = self.ie._download_json(
+            httpbin_cookies_url, video_id,
+            opener_name='cn_verification')['cookies']
+
+        self.assertIn(cookie_name, result_cookies)
+        self.assertEqual(result_cookies[cookie_name], cookie_value)
+
 if __name__ == '__main__':
     unittest.main()
```

phihag:
By just adding a per-request proxy handler, lots of the implementation troubles can be averted entirely. I've integrated exactly that into mainline youtube-dl. Thanks for the report and code!

PR #5080, opened 2015-02-27
[puls4] Add new extractor

HanYOLO:
New Extractor for Puls4.com


dstftw:
Thanks, merged.
Would you like to be credited as one of the authors of youtube-dl? If so, under what name?

PR #5083, opened 2015-02-27
[oppetarkiv] Add new extractor

thc202:
Some, if not all, of the videos appear to be geo-blocked (Sweden).
Test might fail (403 Forbidden) if not run through a Swedish connection.

### DIFF
--- a/youtube_dl/extractor/__init__.py
+++ b/youtube_dl/extractor/__init__.py
@@ -350,6 +350,7 @@ from .odnoklassniki import OdnoklassnikiIE
 from .oktoberfesttv import OktoberfestTVIE
 from .ooyala import OoyalaIE
 from .openfilm import OpenFilmIE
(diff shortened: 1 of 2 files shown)


dstftw:
Thanks, merged.
Would you like to be credited as one of the authors of youtube-dl? If so, under what name?

dstftw:
Looks like you've just copy pasted another extractor's code, namely [SVTPlay](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/svtplay.py). Eschew copy pasting code when possible and try to reuse and generalize existing, like I did in df5ae3eb16effec4dc2609ed3520ddc068305efe.

thc202:
No need to, thanks!

thc202:
Right, I should have checked better how similar cases are handled, or 
asked before PR. Sorry about that!

PR #5087, opened 2015-02-28
strip the leading dots of a file name in case of hidden file

chaos33:
fix for the issue "Strip leading dots from file names"
https://github.com/rg3/youtube-dl/issues/2865


phihag:
Thanks! I've kept the old behavior for IDs because those should be verbatim and added a test.

PR #5105, opened 2015-03-02
[lynda] Check for the empty subtitles

Ftornik:
We can't use **lstrip** for subtitles. Not every subtitle line in the response contains two "new line" characters. And in some cases .srt files can be invalid.

Example: http://www.lynda.com/ajax/player?videoId=159736&courseId=123563&type=transcript
There are lines with only one "new line" and there are empty lines.
We should skip empty lines and add \r\n\r\n after each subtitle line.
Also sequence counter was updated.


dstftw:
Thanks.

PR #5108, opened 2015-03-02
[letv] Add --cn-verification-proxy (Closes #5077)

phihag:
Needs to be tested; example proxy does not work for me.


yan12125:
Thanks a lot for the excellent work! There's still something need to be fixed. I leave them in  #5116

PR #5123, opened 2015-03-04
[Primesharetv] Add primeshare.tv support (with public domain example)

jbuchbinder:


dstftw:
Thanks, cherry picked relevant commits.

PR #5125, opened 2015-03-04
Add metadata from video title parser

phiresky:
This allows the addition of actual metadata from music to the output file.

Example:
`youtube_dl -f bestaudio https://www.youtube.com/watch?v=066CyIM-PDk --add-metadata --metadata-from-title '%(artist)s - %(title)s'` adds the Artist name 'Feint' and the title 'Pastels' to the output file.

A currently missing addition would be to allow the added tags to be integrated into the filename - for which this postprocessor would probably need to be run before writing the video. I wasn't sure how to do that so I left it out for now.

(checked with flake8)

### DIFF
--- a/youtube_dl/postprocessor/metadatafromtitle.py
+++ b/youtube_dl/postprocessor/metadatafromtitle.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+
+import re
+
+from .common import PostProcessor
+from ..utils import PostProcessingError
+
+
+class MetadataFromTitlePPError(PostProcessingError):
+    pass
+
+
+class MetadataFromTitlePP(PostProcessor):
+    def __init__(self, downloader, titleformat):
+        self._titleformat = titleformat
+        self._titleregex, self._attributes = self.fmtToRegex(titleformat)
+
+    def fmtToRegex(self, fmt):
+        """
+        Converts a string like
+           '%(title)s - %(artist)s'
+        to a regex like
+           '(?P<title>.+)\ \-\ (?P<artist>.+)'
+        and a list of the named groups [title, artist]
+        """
+        lastpos = 0
+        regex = ""
+        groups = []
+        # replace %(..)s with regex group and escape other string parts
+        for match in re.finditer(r'%\((\w+)\)s', fmt):
+            regex += re.escape(fmt[lastpos:match.start()])
+            groupname = match.group(1)
+            groups.append(groupname)
+            regex += r'(?P<' + groupname + '>.+)'
+            lastpos = match.end()
+        if lastpos < len(fmt):
+            regex += re.escape(fmt[lastpos:len(fmt)])
+        return regex, groups
+
+    def run(self, info):
+        title = info['title']
+        match = re.match(self._titleregex, title)
+        if match is None:
+            raise MetadataFromTitlePPError('Could not interpret title of video as "%s"' % self._titleformat)
+        for attribute in self._attributes:
+            value = match.group(attribute)
+            info[attribute] = value
+            self._downloader.to_screen('[fromtitle] parsed ' + attribute + ': ' + value)
+
+        return True, info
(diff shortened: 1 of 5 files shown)


phiresky:
any thoughts, @phihag / @dstftw / @jaimeMF ? this is very useful for music

youtube_dl/postprocessor/metadatafromtitle.py
```diff
@@ -0,0 +1,50 @@
...
+        if lastpos < len(fmt):
+            regex += re.escape(fmt[lastpos:len(fmt)])
+        return regex, groups
+
+    def run(self, info):
+        title = info['title']
+        match = re.match(self._titleregex, title)
+        if match is None:
+            raise MetadataFromTitlePPError('Could not interpret title of video as "%s"' % self._titleformat)
+        for attribute in self._attributes:
```
jaimeMF: You don't need to store the attributes, you can get them via `match.groupdict`:
```
for attribute, value in match.groupdict().items():
```

jaimeMF:
I don't know why you need a special `songtitle`, you can just override `title`.

The `fmtToRegex` function seems to work, but I need to read it again.

I would prefer to use the new python format syntax (`{title} - {artist}'`), which [offers a nice function](https://docs.python.org/3/library/string.html#string.Formatter.parse) for splitting the format string. But for consistency with the `--output` option, I think it's ok to use the older.

youtube_dl/postprocessor/metadatafromtitle.py
```diff
@@ -0,0 +1,50 @@
...
+class MetadataFromTitlePPError(PostProcessingError):
+    pass
+
+
+class MetadataFromTitlePP(PostProcessor):
+    def __init__(self, downloader, titleformat):
+        self._titleformat = titleformat
+        self._titleregex, self._attributes = self.fmtToRegex(titleformat)
+
+    def fmtToRegex(self, fmt):
```
jaimeMF: We should probably add some tests for the method, otherwise it's harder to edit it.

youtube_dl/postprocessor/metadatafromtitle.py
phiresky: Oh, right. fixed in https://github.com/phiresky/youtube-dl/commit/6a0a6bbb6eaefb5623900955ecfef10eda267019

phiresky:
I wasn't sure if title is used for anything else / if any other part assumes title is the complete video title.

> I would prefer to use the new python format syntax ({title} - {artist}')

me too, but I thought having it the same as the other parts was better.

(I also admittedly did not really understand the meaning of the output of that function, literal_text seems to be the text before each mach)

jaimeMF:
> I wasn't sure if title is used for anything else / if any other part assumes title is the complete video title.

I think the only one that uses after that is the metadata postprocessor.

> (I also admittedly did not really understand the meaning of the output of that function, literal_text seems to be the text before each mach)

I also had some problems understanding it, I ended up just running it with some format strings. I think this is how the function would be implemented:
```python
import re
import string

def fmtToRegex(fmt):
    regex = ''
    for literal, field, _, _ in string.Formatter().parse(fmt):
        if literal:
            regex += re.escape(literal)
        if field:
            regex += '(?P<' + field + '>.+)'
    return regex

fmts = [
    '{title} - {artist}',
    '{title}, by {artist}',
]

for f in fmts:
    print(repr(f), '->', repr(fmtToRegex(f)))

```

phiresky:
Okay, that makes sense. Also pretty similar to how I implemented it for the other syntax.
Which syntax shall we use?

And do you know how it would be possible to run this before the file saving? (That might also influence the necessity of `songtitle`)

jaimeMF:
> Which syntax should we use?

Keep it as it is now. If we change the output template to the new format we'll update it.

> And do you know how it would be possible to run this before the file saving? (That might also influence the necessity of songtitle)

It could be run inside a YoutubeDL method, but I think it currently has enough code. I don't know if it would make sense to add something like a preprocessor.

If nobady else complains, I'll merge it (I'll delete the `songtitle` usage`).

phiresky:
thanks!

jaimeMF:
Merged, thanks.

PR #5134, opened 2015-03-05
Sohu fix

yan12125:
Sohu info extractor currently fails from time to time. I fix it and add some test cases.

### DIFF
--- a/youtube_dl/extractor/sohu.py
+++ b/youtube_dl/extractor/sohu.py
@@ -4,8 +4,10 @@ from __future__ import unicode_literals
 import re
 
 from .common import InfoExtractor
-from ..utils import compat_str
-from ..compat import compat_urllib_request
+from ..compat import (
+    compat_str,
+    compat_urllib_request
+)
 
 
 class SohuIE(InfoExtractor):
--- a/youtube_dl/extractor/sohu.py
+++ b/youtube_dl/extractor/sohu.py
@@ -73,6 +73,17 @@ class SohuIE(InfoExtractor):
                 'title': '【神探苍实战秘籍】第13期 战争之影 赫卡里姆',
             }
         }]
+    }, {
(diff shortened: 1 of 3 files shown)


youtube_dl/extractor/sohu.py
```diff
@@ -77,6 +104,11 @@ def _fetch_data(vid_id, mytv=False):
                     % (format_id, i + 1, part_count))
 
                 part_info = part_str.split('|')
+
+                # Sanitize URL to prevent download failure
+                if part_info[0][-1] == '/' and su[i][0] == '/':
+                    su[i] = su[i][1:]
+
                 video_url = '%s%s?key=%s' % (part_info[0], su[i], part_info[3])
```
jaimeMF: I think the proper way to do that is `'%s?key=%s'  % (compat_urlparse.urljoin(part_info[0], su[i]), part_info[3])`.

yan12125: It does not work. For example, when ```part_info[0]``` is ```http://70.39.188.10/sohu/s26h23eab6/2/``` and su[i] is ```/0/191/bymuey1CRhqd9706LeW3YD.mp4```, I need ```http://70.39.188.10/sohu/s26h23eab6/2/0/191/bymuey1CRhqd9706LeW3YD.mp4``` but I get ```http://70.39.188.10/0/191/bymuey1CRhqd9706LeW3YD.mp4```. The result is 404 not found.

yan12125:
Add a more generic function to youtube_dl/utils.py based on the implementation at http://stackoverflow.com/a/8926421/3786245

yan12125:
I've fixed title extraction (the original approach was unable to handle titles like "youtube-dl testing video") and add some test cases. Please review it.

dstftw:
Thanks.

PR #5150, opened 2015-03-06
[Yam] Add an error detection and update test cases

yan12125:
The original test case _1 does not exist anymore.

### DIFF
--- a/youtube_dl/extractor/yam.py
+++ b/youtube_dl/extractor/yam.py
@@ -8,6 +8,7 @@ from ..compat import compat_urlparse
 from ..utils import (
     float_or_none,
     month_by_abbreviation,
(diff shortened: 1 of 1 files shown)


dstftw:
Thanks.

dstftw:
Looks like `http://mymedia.yam.com/m/3599430` died as well.

yan12125:
According to [a Yahoo Answer post](https://tw.knowledge.yahoo.com/question/question?qid=1509030402473) (in Chinese), the error message indicates that the video does not exist anymore. Videos may be deleted by the original user, or be deleted by site administrators periodically if any violation against copyrights found. The conclusion is, test cases require frequent update. It's not practical, and just let it go...

PR #5153, opened 2015-03-07
[douyutv] Add new extractor

bonfy:
douyutv is a video broadcast platform in China


dstftw:
Thanks.

bonfy:
that's awesome,you can use "bonfy chen" as my name.

PR #5163, opened 2015-03-08
[walla] Fixed extractor and added site-wide video extraction

guyveis:
The previous extractor allowed downloading videos only from the VOD section.
Now when there is a video in an article, you can download it as well, using the article's URL.

I had to add the strip() to the _download_xml method, since some XML coming from walla can contain spaces in the beginning. If this is not the correct place, please notify me and I will change it.


yan12125:
It might be better to use the ```transform_source``` parameter of ```_download_xml()```?

youtube_dl/extractor/walla.py
```diff
@@ -57,16 +80,23 @@ def _real_extract(self, url):
...
+                'url': playlist_url % default_file,
+                'ext': 'mp4',
+                'format_id': 40,
+            }
+        )
+
         for quality in item.findall('./qualities/quality'):
             format_id = xpath_text(quality, './title')
             fmt = {
-                'url': 'rtmp://wafla.walla.co.il/vod',
```
dstftw: Why did you remove rtmp source? It was a nice way to workaround georestriction.

guyveis: I checked this from a german and US IP on contents I know to be blocked, can you please check it or let me know a specific url and geo-location to check?

PR #5174, opened 2015-03-10
Add subtitles for Funny Or Die

pishposhmcgee:
See, for example, http://www.funnyordie.com/videos/224829ff6d/judd-apatow-will-direct-your-vine
So far I have only seen WebVTT format subtitles, and the pattern is somewhat ambiguous, as they are usually added via the <track> element and lack MIME/type info separate of the filename.


dstftw:
Thanks, merged.

PR #5178, opened 2015-03-10
Update setup.py

dslackw:
fix imported version

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,7 @@ from __future__ import print_function
 import os.path
 import warnings
 import sys
+from youtube_dl.version import __version__
 
 try:
     from setuptools import setup


jaimeMF:
Could you explain what does it fix? The version is imported at [this line](https://github.com/dslackw/youtube-dl/blob/a03aaaed2e0f4653a6b40e92bb91eb848e041de3/setup.py#L74)

dslackw:
Yes but defined [here!](https://github.com/dslackw/youtube-dl/blob/a03aaaed2e0f4653a6b40e92bb91eb848e041de3/setup.py#L79)

jaimeMF:
`__version__` is assigned in youtube_dl.version, which gets "included" in setup.py, so when doing `version=__version__,`, `__version__` is defined. What's the problem? does it raise some error?

phihag:
This is intentional, see [line 73](https://github.com/dslackw/youtube-dl/blob/a03aaaed2e0f4653a6b40e92bb91eb848e041de3/setup.py#L73) and [this stackoverflow answer](http://stackoverflow.com/a/2073599/35070) on why one shouldn't import the package. I'd be fine with automatically bumping the version in the release scripts though, even if that means duplicating it to `version.py` and `setup.py`.

dslackw:
No any problem all good but PEP8 style unused variable

PR #5189, opened 2015-03-12
[BeatportPro] Add new extractor

djpohly:
This extractor is for Beatport's 2-minute, low-quality track previews
only.  To obtain an entire track, you obviously have to purchase and
download it normally through the Beatport store!

Possible future improvements:
- Playlists for albums or other track-list pages
- User login to play from My Beatport, Hold Bin, or Cart


dstftw:
Thanks.

PR #5198, opened 2015-03-14
Be able to download videos directly from a Livesteam JSON feed.

nmrugg:
Tweaked livestream.py to be able to download directly from a JSON feed.

### DIFF
--- a/youtube_dl/extractor/livestream.py
+++ b/youtube_dl/extractor/livestream.py
@@ -134,9 +134,18 @@ class LivestreamIE(InfoExtractor):
                     api_url, video_id, 'Downloading video info'))
                 return self._extract_video_info(info)
 
-        config_json = self._search_regex(
-            r'window.config = ({.*?});', webpage, 'window config')
-        info = json.loads(config_json)['event']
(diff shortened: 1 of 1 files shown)


jaimeMF:
Why do you need to pass the feed url?

nmrugg:
Well, I was using this because if you give it an event URL (e.g., https://new.livestream.com/chess24/tatasteelchess), it will only download the videos that initially appear on the page. It does not download the videos that load when you scroll down. You can get those by manually passing it the feed URL's. Granted, the ideal behavior would be for it to find all of the videos in the event, but I couldn't figure out how to do that. Not yet anyway.

jaimeMF:
I have updated the extractor in 082b1155a36dc9b51424151f80860e52ee30b55e to extract all the videos in an event.

nmrugg:
Marvelous.

dstftw:
Fixed in youtube-dl 2015.03.15. See [our FAQ](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) on how to update.

PR #5204, opened 2015-03-14
[YoutubeDL] extract_info: Default 'download' to False

jaimeMF:
That's what I would expect from the method name


dstftw:
This may break behavior of some third party code that uses `extract_info`.

jaimeMF:
I know, but I think that default makes more sense and it should be easy to detect the break.

PR #5206, opened 2015-03-14
Add moviestorm InfoExtractor

orangeoval:
The moviestorm InfoExtractor is unique when compared to the other extractors in that it is designed to handle a link farm page, moviestorm.eu pages to be specific.  The handler performs some basic web scraping to pull a list of up to 50 external links ('direct_urls') from the given moviestorm page.  It is currently limited to pulling 50 links, as moviestorm throttles user access after making 50+ rapid succession requests, at which point it starts returning 503 responses.

After scraping the direct_urls, MovieStormIE returns the list back to its parent YoutubeDL object.  I have added a check to the YoutubeDL object that, before proceeding with any IE's response, checks to see if that IE is a link farm IE by checking if said IE has a _LINK_FARM variable defined and set to True.  Right now moviestorm is the only link farm IE, but hopefully there will be more soon that can use this basic framework, which we can identify with _LINK_FARM = True.

When YoutubeDL knows its dealing with a response from a link farm extractor, it passes the extractor's result to a new 'process_farmed_links' method that I added to the YoutubeDL object.  This method basically combs through the list and weeds out any urls that cannot be handled by one of youtube-dl's various extractors.  Then, the method loops over that culled list and feeds the urls to their corresponding extractor.  If the extractor succeeds in downloading the resource, the process ends when the download is complete.  If, however, the extractor fails on the provided link (sometimes bad links are submitted it moviestorm, or sometimes they are old), the process_farmed_links method reports an error about the failure, then moves on to the next link in the culled list, passing it off to its corresponding handler, and so on.

I've outlined some info in the MovieStormIE docstring.  But basically, there are two types of moviestorm links that can work here, a movie page, or an episode page.  Since the generic extractor will always fail on any moviestorm link, I've made MovieStormIE._VALID_URL extremely greedy.  It will pull in any http://moviestorm.eu links, as falling back to generic on them will fail, and it does some checking afterwards.  If the extractor notices that the link is not an episode or movie page, it raises an error informing the user what went wrong and what type of moviestorm links are valid to process.  Here are some example links:

    [GOOD: movie page]     http://moviestorm.eu/view/5269-watch-taken-3-online.html
    [GOOD: episode page]  http://moviestorm.eu/view/5821-watch-portlandia/season-1/episode-1
    [BAD: series page]         http://moviestorm.eu/view/5821-watch-portlandia.html

The extractor does not currently attempt to handle a series page because there are many episodes therein, and there is no way to know which one the user wants.  If it seems like a good idea to everyone else, there could be future development done to allow the extractor to handle to series page and attempt to download all episodes available, kind of like a youtube playlist.  This might be a bit more complicated, but I don't see why it wouldn't be be do-able.

I think it would be nice to add similar link farm handlers for other sites.  Moviestorm.eu is nice because of its structure where all links are listed on a single page, and the content is quote-un-quote guaranteed.  alluc.to is also a nice link farm, but scraping their search result page could suck in results that aren't what the user is looking for.  If other's know of sites similar in format to moviestorm where links are cleanly structured by series/episode, I'm all ears.

One last note.  It seems like the folks at moviestorm might have a funky drupal configuration, as every now and then when you try to load a series page or episode page, you will get a blank page with 'MySQL server has gone away' as the only response.  The MovieStormIE extractor currently detects this and will wait 5 seconds, then try again.  It will attempt 5 retries.  If all of those fail, it will give up and inform the user of what's up.  That MySQL response is a bit flakey and sometimes moviestorm will return that, but simply refreshing the page a few times in a browser (or having the extractor retry) gets around the error.  Some pages seem to exhibit the issue more frequently than others.  I've noticed it as particularly prevent on the Simpsons series page (http://moviestorm.eu/view/218-watch-the-simpsons.html) and its subsequent episode pages.  The issue is worse on some days/times than others, but over the past few weeks regular business hours EST seem to be the worst.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -638,7 +638,7 @@ class YoutubeDL(object):
 
                 # ignore non-familiar links
                 if c != 'GenericIE' and c != 'MovieStormIE' and ie.suitable(farmed_url):
-                    familiar_farmed_urls.append( [ie, farmed_url] )
+                    familiar_farmed_urls.append([ie, farmed_url])
 
         for tuple in familiar_farmed_urls:
             ie = tuple[0]
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -629,6 +629,28 @@ class YoutubeDL(object):
         for key, value in extra_info.items():
             info_dict.setdefault(key, value)
 
+    def process_farmed_links(self, direct_urls):
+        familiar_farmed_urls = []
+        for farmed_url in direct_urls:
+            for ie in self._ies:
+                # not all extractors have IE_NAME set, using class name for fuller coverage
+                c = ie.__class__.__name__
+
+                # ignore non-familiar links
+                if c != 'GenericIE' and c != 'MovieStormIE' and ie.suitable(farmed_url):
+                    familiar_farmed_urls.append( [ie, farmed_url] )
+
+        for tuple in familiar_farmed_urls:
+            ie = tuple[0]
+            familiar_farmed_url = tuple[1]
+
+            try:
+                ie_result = ie.extract(familiar_farmed_url)
+                return ie_result, ie
+            except:
+                # Failed extract, move on to next url in list
+                ie.to_screen("\033[0;33mWARNING:\033[0m failed attempt, trying next farmed link")
+
     def extract_info(self, url, download=True, ie_key=None, extra_info={},
                      process=True):
         '''
@@ -652,6 +674,11 @@ class YoutubeDL(object):
 
             try:
                 ie_result = ie.extract(url)
+
+                # handle link farm extractors
+                if hasattr(ie, '_LINK_FARM') and ie._LINK_FARM:
+                	ie_result, ie = self.process_farmed_links(ie_result)
+
                 if ie_result is None:  # Finished already (backwards compatibility; listformats and friends should be moved here)
                     break
                 if isinstance(ie_result, list):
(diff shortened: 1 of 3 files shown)


orangeoval:
Oh, one last thing.  I also did not include any tests in MovieStormIE._TEST, as the links on any given page can change dynamically.  Also, the moviestorm IE doesn't do any actual downloading, but rather hands the downloading off to the other handlers.  Given these two realities, I couldn't think of a consistent test for this.  If someone has a suggestion, let me know, but I don't think it is necessarily a problem that there is no test here.  As long as the other handlers pass their tests, this one should work.

orangeoval:
Seeing as this i my first go-round Travis-CI, and with submitting a pull request in general, might anyone have some words of wisdom to offer regarding this build failure?

I checked the logs here (https://travis-ci.org/rg3/youtube-dl/builds/54406499) and see a bunch of failures, but, unless I am missing something (which is entirely possible, as the output is quite large), the failures have to do with other extractors/code that I didn't touch.  I cloned the primary youtube-dl repository before doing work, so I figured any code there must have passed the build tests in order to get into the main repository.  

Furthermore, there isn't even a unit test (_TEST=False) in the extractor that I added, for reasons explained above.  Again, total noob here, so be gentle, but some insight would be greatly appreciated. 

Thanks, youtube-dl is awesome!

dstftw:
youtube-dl does not support sites dedicated to [copyright infringement](https://github.com/rg3/youtube-dl#can-you-add-support-for-this-anime-video-site-or-site-which-shows-current-movies-for-free).

orangeoval:
alrighty then.  Well I guess that indirectly solves the build problem.  Thanks for pointing me to that link, I overlooked it.

PR #5215, opened 2015-03-15
[mixcloud] Fix extraction of URL and like-count

Roman2K:
When I went to download Mixcloud songs, youtube-dl tried with a .mp3 URL first, then only m4a (which works once it finds a server from stream1 to stream30). This doesn't seem like a valid strategy anymore. From when I've seen in the Chrome console, only .m4a are streamed now.

So I updated the way URLs are extracted from the current DOM structure (like_count as well) and removed the .mp3 => .m4a fallback, now .m4a only.

Also, trying stream1..30 was slow. I made it so it tries the preview's server first (like stream18), then further and further out: stream19, 17, 20, 16, 21, ...

Needs testing, but should be stable from my tests.

### DIFF
--- a/test/test_mixcloud.py
+++ b/test/test_mixcloud.py
@@ -1,0 +1,16 @@
+import unittest
+from youtube_dl.extractor.mixcloud import server_numbers
+
+class TestMixcloud(unittest.TestCase):
+    def test_server_numbers(self):
+        self.assertEqual([n for n in server_numbers(2, (1, 5))],
+            [2, 3, 1, 4, 5])
+        self.assertEqual([n for n in server_numbers(1, (1, 5))],
+            [1, 2, 3, 4, 5])
+        self.assertEqual([n for n in server_numbers(5, (1, 5))],
+            [5, 4, 3, 2, 1])
+        self.assertEqual([n for n in server_numbers(-1, (1, 5))],
+            [-1, 1, 2, 3, 4, 5])
+
+if __name__ == '__main__':
+    unittest.main()
--- a/youtube_dl/extractor/mixcloud.py
+++ b/youtube_dl/extractor/mixcloud.py
@@ -73,14 +73,10 @@ class MixcloudIE(InfoExtractor):
         webpage = self._download_webpage(url, track_id)
 
         preview_url = self._search_regex(
-            r'\s(?:data-preview-url|m-preview)="([^"]+)"', webpage, 'preview url')
-        song_url = preview_url.replace('/previews/', '/c/originals/')
-        template_url = re.sub(r'(stream\d*)', 'stream%d', song_url)
+            r'\bm-play-on-spacebar\b.*\n?.*\bm-preview="([^"]+)"', webpage, 'preview url')
+        song_url = re.sub(r'\.mp3$', '.m4a', preview_url.replace('/previews/', '/c/m4a/64/'))
+        template_url = re.sub(r'(stream\d+)', 'stream%d', song_url)
         final_song_url = self._get_url(track_id, template_url)
-        if final_song_url is None:
-            self.to_screen('Trying with m4a extension')
-            template_url = template_url.replace('.mp3', '.m4a').replace('originals/', 'm4a/64/')
-            final_song_url = self._get_url(track_id, template_url)
         if final_song_url is None:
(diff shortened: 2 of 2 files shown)


jaimeMF:
It doesn't work for our test urls (http://www.mixcloud.com/dholbach/cryptkeeper/ and http://www.mixcloud.com/gillespeterson/caribou-7-inch-vinyl-mix-chat/)

youtube_dl/extractor/mixcloud.py
```diff
@@ -122,3 +120,31 @@ def _real_extract(self, url):
             'view_count': view_count,
             'like_count': like_count,
         }
+
+class ServerAttemptSequence:
```
jaimeMF: The idea is cool, but it's too complicated. If I have read it correctly, you can use something simpler:
```python
def server_sequence(start, max_value):
     low = high = start
     yield start
     while low > 1 and high < max_value:
         low -= 1
         high += 1
         yield high
         yield low
```
Or even shorter:
```python
def server_sequence(start, max_value):
    yield start
    for low, high in zip(range(start - 1, 0, -1), range(start + 1, max_value)):
        yield high
        yield low
```
The last one may be a bit harder to understand, but I like it.

Roman2K: @jaimeMF I agree, my way is not very elegant but I couldn't make it any simpler. My algorithmic skills are a bit rusty also :laughing: (where's the WD40 when you need it?). Your alternatives do look nicer but they don't work as expected.

With `start = 2` and `max_value = 5`, both generate:
```
2, 3, 1
```

Where I expect:

```
2, 3, 1, 4, 5
```

Do you have other suggestions?

jaimeMF:
```python
def server_sequence(start, max_value):
     low = high = start
     yield start
     while low > 1 or high < max_value:
         low -= 1
         high += 1
         if high <= max_value:
             yield high
         if low >= 1:
             yield low
```
In the second form, you would need to use `itertools.izip_longest`.

(By the way, you should probably add a docstring explaining what it does so that someone reading it understand what it's supposed to do).

Roman2K: @jaimeMF Thanks for your suggestions. I prefer your second solution and implemented it here: https://github.com/Roman2K/youtube-dl/commit/f4fd09ffd5d1e9f218267b2206a90aa998cbcd9d#diff-a845099a3da4a695771782a299402291R125 -- what do you think?

test/test_mixcloud.py
```diff
@@ -0,0 +1,16 @@
+import unittest
+from youtube_dl.extractor.mixcloud import server_numbers
+
+class TestMixcloud(unittest.TestCase):
```
jaimeMF: It's an internal function, I don't think it's worth a test.

youtube_dl/extractor/mixcloud.py
```diff
@@ -122,3 +121,30 @@ def _real_extract(self, url):
...
+    that of the preview; to be confirmed.
+    """
+
+    if len(boundaries) != 2:
+        raise ValueError("boundaries should be a two-element tuple")
+    min, max = boundaries
+    highs = range(first + 1, max + 1)
+    lows = range(first - 1, min - 1, -1)
+    rest = filter(None,
+        itertools.chain.from_iterable(itertools.izip_longest(highs, lows)))
```
jaimeMF: It's `zip_longest` in python3

youtube_dl/extractor/mixcloud.py
jaimeMF: It's ok, but it still doesn't work with the test urls. Which urls are you downloading?

jaimeMF:
I have cherry picked the fix for the like count in fa8ce2690447fa06a10398efd5b6dfa96c2ed1b4 and the change in the server order in aae53774f247edcca7671434c16f7931d5c7b9da.

If you have a way for detecting if the track is available in m4a or mp3, we'd review it.

Thanks!

Roman2K:
@jaimeMF OK, thanks for your comments, I had never coded in Python before, that was very helpful! Don't yet know what's the deal with mp3 vs m4a though.

PR #995, opened 2013-07-05
Support download of multipart videos

jaimeMF:
This is my solution for the issue I posted in #991

A `parts` field can be set in the info_dict containing the urls of the different parts, when all the parts have been downloaded they are joined using the `FFmpegJoinVideos` postprocessor

I have tested it with a Tudou video: http://www.tudou.com/programs/view/J-P3o-chBp4/ (It's the only IE I  have found with multiple parts videos)

PR #5217, opened 2015-03-16
Support download of multipart videos, take 2

malept:
This is a continuation of #995.
 The only notable changes from that PR, apart from being updated to master, is that the comedycentral and adultswim extractors have been modified to support the `--join-parts` flag. Those extractors have been consistently tested.


yan12125:
I appreciate your excellent work! This is a great feature I need desperately. However, I think it's better to start the modifications based on a more recent version instead of such an old commit. In this PR, lots of unrelated commits are involved (for example [[vice] Remove extractor](https://github.com/malept/youtube-dl/commit/87a26277000b425972ec427f86b68fef2097c1ac)), and something unexpected may happen after the merge.

malept:
Rebasing might take a while. It was hard enough merging master in (given the churn that this repository goes through), and I would want to make sure that the original commits were kept as much as possible.

PR #5230, opened 2015-03-18
[niconico] fixed bug for nm URLs when logged in

ndac-todoroki:
You couldn't download *nm* URL videos when you authenticated (you were able to when *not* authenticated)
I changed the download URL a bit; It shouldn't affect other than *nm* URLs.

for test
- *sm* URLs
http://www.nicovideo.jp/watch/sm25798024

- *nm* URLs
http://www.nicovideo.jp/watch/nm14296458


dstftw:
Thanks.

Issue #2050, opened 2013-12-27
Support for channel4.com OnDemand


canavan:
Please add support for channel 4 OnDemand, e.g. http://www.channel4.com/programmes/the-it-crowd/4od

There's already a separate tool (in ruby, uses rtmpdump) that supports downloading from that service: https://github.com/robwatkins/4od-dl

youtube-dl --verbose http://www.channel4.com/programmes/the-it-crowd/4od
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['--verbose', 'http://www.channel4.com/programmes/the-it-crowd/4od']
[debug] Encodings: locale 'ISO8859-1', fs 'ISO8859-1', out 'ISO8859-1', pref: 'ISO8859-1'
[debug] youtube-dl version 2013.12.26
[debug] Python version 2.7.3 - IRIX64-6.5-IP35-mips-n32bit-ELF
[debug] Proxy map: {}
WARNING: Falling back on generic information extractor.
[generic] 4od: Downloading webpage
[generic] 4od: Extracting information
ERROR: Unsupported URL: http://www.channel4.com/programmes/the-it-crowd/4od; please report this issue on https://yt-dl.org/bug . Be sure to call youtube-dl with the --verbose flag and include its complete output. Make sure you are using the latest version; type  youtube-dl -U  to update.
Traceback (most recent call last):
  File "/usr/nekoware/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 484, in extract_info
    ie_result = ie.extract(url)
  File "/usr/nekoware/bin/youtube-dl/youtube_dl/extractor/common.py", line 149, in extract
    return self._real_extract(url)
  File "/usr/nekoware/bin/youtube-dl/youtube_dl/extractor/generic.py", line 330, in _real_extract
    raise ExtractorError(u'Unsupported URL: %s' % url)
ExtractorError: Unsupported URL: http://www.channel4.com/programmes/the-it-crowd/4od; please report this issue on https://yt-dl.org/bug . Be sure to call youtube-dl with the --verbose flag and include its complete output. Make sure you are using the latest version; type  youtube-dl -U  to update.

fstirlitz:
Downloading itself seems easy; the stream is a regular f4m, and finding the stream's URL is not difficult. The only problem is that the stream is encrypted by Flash Access DRM.

Some resources:
* F4V spec: http://download.macromedia.com/f4v/video_file_format_spec_v10_1.pdf
[...]

PR #5246, opened 2015-03-20
4oD extractor, with a testcase

fstirlitz:
mostly unusable, since FlashAccess DRM is not implemented, here or in ffmpeg.

youtube-dl can at least download subtitles and metadata, and make the stream URL available through `-j` for external decryption.

see also: #2050


dstftw:
If I get it right, all channel4 videos are DRM protected. So, despite subtitles and metadata extraction, I see little sense in extractor that will grab unplayable videos, as downloading playable media is the primary purpose of youtube-dl.

PR #5262, opened 2015-03-22
Add safaribooksonline extractor

zx8:
Following on from https://github.com/rg3/youtube-dl/issues/5253, I decided to give implementing this a go myself.

### DIFF
--- a/youtube_dl/extractor/__init__.py
+++ b/youtube_dl/extractor/__init__.py
@@ -420,6 +420,10 @@ from .rutube import (
 )
 from .rutv import RUTVIE
 from .sandia import SandiaIE
+from .safari import (
+    SafariIE,
+    SafariCourseIE,
+)
 from .sapo import SapoIE
 from .savefrom import SaveFromIE
 from .sbs import SBSIE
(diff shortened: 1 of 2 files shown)


dstftw:
Thanks, merged.

zx8:
No problem. Keep up the good work.

It's ok, it was only a small change, maybe if I contribute something more substantial in future. :-)

evliu:
Do you know if there is a difference between www.safaribooksonline.com and techbus.safaribooksonline.com? I have an enterprise account and I found this is not able to download via my account. Thx

zx8:
@evliu Not sure I'm afraid, I don't have an Enterprise account so can't test/fix it myself.

evliu:
Ahh okay. I think it used to work before the change on their site. I’ll see if I can figure something out

evliu:
I think the mobile site might be the same, do you know how I could implement for something like http://m.safaribooksonline.com/hd/video?isbn=9781771371179&linkid=1-2013-09-19 ?

jcdauchy:
I have the same type of problem with this type of URL : https://my.safaribooksonline.com/video/databases/hadoop/9781771372374

Don't understand the difference with the default www.safaribooksonline.com.

Anyone with the same issue ?
Thanks

PR #5265, opened 2015-03-23
[MiomioTv] Add new extractor

tiktok7:
added entry for high lag video site http://www.miomio.tv
(contains a range of content mostly in asian languages including instructional and promotional videos e.g. http://www.miomio.tv/teach/draw/)

### DIFF
--- a/docs/supportedsites.md
+++ b/docs/supportedsites.md
@@ -244,7 +244,6 @@
  - **Mgoon**
  - **Minhateca**
  - **MinistryGrid**
- - **Miomio.tv**
  - **mitele.es**
  - **mixcloud**
  - **MLB**
--- a/docs/supportedsites.md
+++ b/docs/supportedsites.md
@@ -244,6 +244,7 @@
  - **Mgoon**
  - **Minhateca**
  - **MinistryGrid**
+ - **Miomio.tv**
  - **mitele.es**
  - **mixcloud**
  - **MLB**
--- a/youtube_dl/extractor/miomio_tv.py
+++ b/youtube_dl/extractor/miomio_tv.py
@@ -1,0 +1,70 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class MiomioTvIE(InfoExtractor):
+    IE_NAME = 'miomio.tv'
+    _VALID_URL = r'https?://(?:www\.)?miomio\.tv/watch/cc(?P<id>[0-9]+)'
+    _TEST = {
+        'url': 'http://www.miomio.tv/watch/cc179734/',
+        'md5': '48de02137d0739c15b440a224ad364b9',
+        'info_dict': {
+            'id': '179734',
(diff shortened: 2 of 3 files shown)


docs/supportedsites.md
```diff
@@ -244,6 +244,7 @@
  - **Mgoon**
  - **Minhateca**
  - **MinistryGrid**
+ - **Miomio.tv**
```

youtube_dl/extractor/miomio_tv.py
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._html_search_regex(r'<meta\s+name="description"\s+content="\s*([^"]*)\s*"', webpage, 'title')
+        ref_path = self._search_regex(r'src="(/mioplayer/.*?)"', webpage, 'ref_path')
+        referer = 'http://www.miomio.tv{}'.format(ref_path)
```
dstftw: `{}` doesn't work in python 2.6.

youtube_dl/extractor/miomio_tv.py
```diff
@@ -0,0 +1,70 @@
...
+        xml_config = self._search_regex(r'flashvars="type=sina&amp;(.*?)&amp;cid=', webpage, 'xml config')
+        self._request_webpage("http://www.miomio.tv/mioplayer/mioplayerconfigfiles/xml.php?id={}&r=cc{}".format(id, 945), video_id)
```
jaimeMF: Do you really need to request that webpage? You don't use its contents.

If it's required, please add a comment.

youtube_dl/extractor/miomio_tv.py
```diff
@@ -0,0 +1,70 @@
...
+            entries.append({
+                'id': segment_id,
+                'title': segment_title,
+                'duration': segment_duration,
+                'url': segment_url
+            })
+
+        http_headers = {
+            'Referer': referer,
+            'Accept-Language': 'en,en-US;q=0.7,de;q=0.3',
```
jaimeMF: Are all headers required?

docs/supportedsites.md
tiktok7: got rid of the change -> checked out old version

youtube_dl/extractor/miomio_tv.py
tiktok7: changed to {0} format

youtube_dl/extractor/miomio_tv.py
tiktok7: The server starts dropping connections when it's not requested. I added a comment

youtube_dl/extractor/miomio_tv.py
tiktok7: The language header is not. I've removed it

dstftw:
Thanks, merged.

tiktok7:
That's very generous, thanks! My employer is very conservative so I'd prefer to omit my last name. 
How about: "Will W."

PR #5276, opened 2015-03-24
[22Tracks] Add new extractor 

amishb:
Added ability to download sing tracks and entire genres from 22Tracks.com

Note: No tests can be added due to the sites rolling content.


dstftw:
Cherry-picked relevant commits, thanks.

amishb:
Thanks for that. I'll review what you have cleaned up and ensure any future commits or extracts follow the same style. Saves you refactoring! :)

PR #5291, opened 2015-03-26
[Varzesh3] Add new extractor

mtp1376:
Adds support for video.varzesh3.com, a famous Iranian sports video sharing website.


mtp1376:
Tell me what's the problem or shoot me :8ball:

dstftw:
Thanks.

mtp1376:
Sure, :D, "Mohammad Teimori Pabandi"

PR #5305, opened 2015-03-28
Add dhm extractor

ossi96:
Add a new extractor for the video collection of the German historical museum

### DIFF
--- a/youtube_dl/extractor/__init__.py
+++ b/youtube_dl/extractor/__init__.py
@@ -106,6 +106,7 @@ from .dbtv import DBTVIE
 from .dctp import DctpTvIE
 from .deezer import DeezerPlaylistIE
 from .dfb import DFBIE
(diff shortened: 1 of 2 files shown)


dstftw:
Thanks, merged.

ossi96:
There is a problem with the changes you've done in _VALID_URL. When you try to download a video from the "Mapping the wall" collection it doesn't work because it has a different root URL.

dstftw:
@ossi96 post an URL.

ossi96:
http://www.dhm.de/filmarchiv/02-mapping-the-wall/peter-g/rolle-1/

dstftw:
Fixed in 5a3b315b5fcc1b1e153baeaed669e64d137d4047.

ossi96:
And now my name appears 2 times in the credits. :D

dstftw:
Oops.

PR #5311, opened 2015-03-29
[douyutv] Fix extractor and improve error handling

yan12125:
douyutv.com (斗鱼) has recently changed how urls are calculated. I've posted a working version here. Note that the key ('1231') may change in the future.

I've also improved error handling so that comprensive error messages are shown when API requests failed (due to wrong key, etc.)

### DIFF
--- a/youtube_dl/extractor/douyutv.py
+++ b/youtube_dl/extractor/douyutv.py
@@ -1,19 +1,23 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
+import hashlib
+import time
 from .common import InfoExtractor
-from ..utils import ExtractorError
+from ..utils import (ExtractorError, unescapeHTML)
+from ..compat import (compat_str, compat_basestring)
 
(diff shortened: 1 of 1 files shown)


yan12125:
The test ```TestDownload.test_DouyuTV_1``` fails now because the channel is offline. I keep it in ```_TESTS``` because the title has HTML entities and I'd like to test the correctness of [this line](https://github.com/rg3/youtube-dl/pull/5311/files#diff-4892f8f0dc4d5abeaf923057f7af0ba5R83)

dstftw:
Thanks.
Is there any advantage of current api over using `http://www.douyutv.com/swf_api/room`?

yan12125:
The algorithm for calculating swf_api url is inside a swf file, which cracks all my flash decompilers, so I can't reverse it. The algorithm in this PR is based on Douyu's Android App, which is much easier for reverse enginnering.

PR #5313, opened 2015-03-29
[Xuite] Fix extraction on python 3.2

yan12125:
base64.b64decode() accept only binary types in Python 3.2

This patch should fix failed tests with Python 3.2 on Travis-CI


dstftw:
Thanks.

PR #5319, opened 2015-03-29
[Dumpert] Add new extractor

jorams:
Add support for the Dutch video site Dumpert. http://www.dumpert.nl/


dstftw:
Thanks.

Niek:
@jorams any chance you can update it for the redesign? See also: https://github.com/ytdl-org/youtube-dl/issues/22428

PR #5352, opened 2015-04-03
[Gamersyde] Add new extractor

snipem:
Added support for Gamersyde a resource of unedited high quality (1080p60) video game footage


dstftw:
Thanks.

snipem:
Thank you. That would be Matthias Küch

PR #5355, opened 2015-04-03
[RadioJavan] Add new extractor

mtp1376:
Adds support for Iranian famous music-video sharing website, RadioJavan.


dstftw:
Thanks.

PR #5365, opened 2015-04-05
[SpankBang] Add new extractor

newtonelectron:
Add an extractor for spankbang.com.

### DIFF
--- a/youtube_dl/extractor/spankbang.py
+++ b/youtube_dl/extractor/spankbang.py
@@ -8,6 +8,20 @@ class SpankBangIE(InfoExtractor):
     """Extractor for http://spankbang.com"""
     
     _VALID_URL = r"https?://(?:www\.)?spankbang\.com/(?P<id>\w+)/video/.*"
+    
+    _TEST = {
+        "url": "http://spankbang.com/3vvn/video/fantasy+solo",
+        "md5": "1cc433e1d6aa14bc376535b8679302f7",
+        "info_dict": {
+            "id": "3vvn",
+            "title": "fantasy solo",
+            "description": "Watch fantasy solo free HD porn video - 05 minutes - dillion harper masturbates on a bed free adult movies.",
+            "format": "720p",
+            "format_id": "720p",
+            "ext": "mp4",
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/spankbang.py
```diff
@@ -0,0 +1,38 @@
...
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+import re
+
+class SpankBangIE(InfoExtractor):
+    """Extractor for http://spankbang.com"""
+    
+    _VALID_URL = r"https?://(?:www\.)?spankbang\.com/(?P<id>\w+)/video/.*"
+
```
dstftw: Any tests?

youtube_dl/extractor/spankbang.py
```diff
@@ -0,0 +1,38 @@
...
+        
+        qualities = re.findall(r"<span.+?>([0-9]+p).*?</span>", webpage)
+        
+        formats = []
+        for q in sorted(qualities):
+            formats.append({
+                "format_id": q,
+                "format": q,
+                "ext": "mp4",
+                "url": "http://spankbang.com/_{}/{}/title/{}__mp4".format(video_id, stream_key, q)
```
dstftw: `{}` does not work in python 2.6

youtube_dl/extractor/spankbang.py
newtonelectron: No, sorry.
I think the  ``stream_key`` is dynamic. I'm not sure how to test for that or if it would be worth it.

youtube_dl/extractor/spankbang.py
newtonelectron: I just need to number them, right? i.e ``{0}/{1}/title/{2}__mp4``

youtube_dl/extractor/spankbang.py
dstftw: Just add a test video as described in [developer instructions](https://github.com/rg3/youtube-dl#developer-instructions) or look at any other extractor.

youtube_dl/extractor/spankbang.py
dstftw: Yes, that will work.

dstftw:
Thanks, merged.

PR #5368, opened 2015-04-06
crooksandliars.com extractor

fstirlitz:
And by the way, I wish the generic extractor would simply download the webpage and match all extractors against all the iframe and embed tags in the webpage. Could have saved me a bit of trouble.


yan12125:
All tests failed due to missing 'ext' attribute in 'info_dict'. Please fill in the field.

fstirlitz:
What is it supposed to be set to? This particular video is available as both WebM and mp4. I don't see the point in specifying 'ext' in the test at all.

yan12125:
It should be the extension of the best quality video file.

dstftw:
Thanks.

PR #5374, opened 2015-04-07
[mtv/comedycentral] Add alternate way to retrieve mgid

malept:
Uses the embed URL to retrieve the mgid for a given episode, if all other methods fail.

Fixes rg3/youtube-dl#5369.


dstftw:
Turns out it was fixed meanwhile.

malept:
Please reopen, the mgid extractor has failed again, and I have confirmed that my patch fixes the problem.

PR #5377, opened 2015-04-08
[nrk] Fixed subtitles

einarhb:
Added detection for subtitles placed in child nodes (sometimes used for styling), and subtitles should now be entirely extracted

### DIFF
--- a/youtube_dl/extractor/cnn.py
+++ b/youtube_dl/extractor/cnn.py
@@ -12,7 +12,7 @@ from ..utils import (
 
 class CNNIE(InfoExtractor):
     _VALID_URL = r'''(?x)https?://(?:(?:edition|www)\.)?cnn\.com/video/(?:data/.+?|\?)/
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/nrk.py
```diff
@@ -210,7 +210,14 @@ def _get_subtitles(self, subtitlesurl, video_id, baseurl):
             duration = parse_duration(p.get('dur'))
             starttime = self._subtitles_timecode(begin)
             endtime = self._subtitles_timecode(begin + duration)
-            srt += '%s\r\n%s --> %s\r\n%s\r\n\r\n' % (compat_str(pos), starttime, endtime, p.text)
+
+            subs = []
+            subs.append(unicode(p.text).strip())
```
jaimeMF: `unicode` is not defined in python 3.x, you'll have to use `compat_str`. But are you sure it's needed? I think `p.text`  is always an unicode string.

yan12125:
Hello @einarhb I've moved subtitle conversion codes to utils.py in 1c7e2e64f6328024711d5fa999d4498396f4cb5c. I think NRK subtitles are correctly extracted now. If you still have a problem. Feel free to open a new PR/issue.

PR #5407, opened 2015-04-12
[FootyTube] Fixed missing HTTP prefixes

snipem:
Some videos like the ones from the recent European Qualifiers where missing the HTTP prefix


dstftw:
Cherry-picked relevant commits. Thanks.

PR #5434, opened 2015-04-15
Add ability to embed subtitles in mkv files

Nikoli:


youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -504,9 +504,6 @@ def _conver_lang_code(cls, code):
         return cls._lang_map.get(code[:2])
 
     def run(self, information):
-        if information['ext'] != 'mp4':
```
jaimeMF: Can ffmpeg/libav embed subtitles in other formats apart from mkv and mp4?

Nikoli:
About other containers: https://en.wikipedia.org/wiki/Comparison_of_container_formats#Caption_.28Subtitle.29_formats_supported mentions only mkv, mp4 and ogg, but i doubt if using ogg container for video makes any sense.

jaimeMF:
Merged, thanks.

Issue #2885, opened 2014-05-10
Remove trackers by default


Nikoli:
For tracking purposes google adds unique fields at container level to mp4 files downloaded from youtube:

```
$ mediainfo -f _video.mp4 |grep '^g'
gsst                                     : 0
gstd                                     : xxxxxx
gssd                                     : yyyyyyyyyyyyyyyyyyyyyyyyyyy
gshh                                     : zzzzzzzzzzzzzzzz.googlevideo.com
```

https://superuser.com/questions/503868/how-i-can-remove-all-google-data-from-youtube-video-downloads

I think youtube-dl should by default create reproducible downloads: if several people from different countries using different platforms execute at different time same youtube-dl commands for same urls they should get _exactly_ the same files. Now md5sum for mp4 files from youtube changes after simple redownloading file from same system and location, because gs\* entries are changed.

The most obvious and trivial workaround is to simply remux:
 `avconv -i _video.mp4 -c copy -flags bitexact video.mp4`. 

`-flags bitexact` is required, because else info about ffmpeg/libav versions will be added to file

```
$ mediainfo -f video.mp4 |grep -i lavf
Writing application                      : Lavf55.12.0
```

P.S. With matroska files bitexact option works fine too.

NulAsh:
Wow, 6 years ago. So, no one did anything about it yet? I had the same problem - downloaded audio track from youtube twice and files had differences. After diving in ffmpeg's code I found -bitexact parameter, but I had to manually patch my youtube-dl as there are no other ways to pass arbitrary parameters to postprocessing stage.

PR #5438, opened 2015-04-15
Create reproducible non unique files when muxing or encoding

Nikoli:
Also prevents adding Unique ID and ffmpeg/libav version to files.

P.S. It seems libav has bug:
https://bugzilla.libav.org/show_bug.cgi?id=845

### DIFF
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -131,6 +131,7 @@ class FFmpegPostProcessor(PostProcessor):
             os.stat(encodeFilename(path)).st_mtime for path in input_paths)
 
         files_cmd = []
+        opts += ['-fflags', '+bitexact', '-flags', '+bitexact']
         for path in input_paths:
             files_cmd.extend([encodeArgument('-i'), encodeFilename(path, True)])
         cmd = ([encodeFilename(self.executable, True), encodeArgument('-y')] +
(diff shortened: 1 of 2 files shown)


Nikoli:
Since ```--format='bestvideo+bestaudio/best'``` is default this patch should mostly fix issue #2885.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -233,7 +233,7 @@ def run(self, information):
                 # MP3 otherwise.
                 acodec = 'libmp3lame'
                 extension = 'mp3'
-                more_opts = []
+                more_opts = ['-flags', '+bitexact',]
```
jaimeMF: Aren't you adding the same args in `run_ffmpeg`?:
```python
opts = ['-vn', '-fflags', '+bitexact'] + acodec_opts + more_opts
```

Nikoli:
@jaimeMF, ```-fflags +bitexact``` affects only muxer and ```-flags +bitexact``` affects only encoder, so all ffmpeg/avconv commands should have ```-fflags +bitexact```. If encoding is done, then also ```-flags +bitexact`` should be added.

https://libav.org/avconv.html
https://git.libav.org/?p=libav.git;a=blob;f=tests/fate-run.sh;h=a0a69ad0a7ca716e008446eea67f69b19fb14840;hb=HEAD#l122

https://www.ffmpeg.org/ffmpeg-formats.html#toc-Format-Options
https://www.ffmpeg.org/ffmpeg-codecs.html#toc-Codec-Options
http://git.videolan.org/?p=ffmpeg.git;a=blob;f=tests/fate-run.sh;h=1a8ff940bd92f4e301435c49d0dc9331bda61700;hb=HEAD#l143

Nikoli:
Simpler patch, now hls downloads are not unique too.

PR #5439, opened 2015-04-15
[gfycat] Add new extractor

mandb:


mandb:
Build failed, but not on anything related to these additions?

dstftw:
Thanks.

Issue #5394, opened 2015-04-10
Add official support for gfycat


woodenphone:
While gfycat.com works at the moment, it is not listed in the supported sites.
Can it be officially maintained as a supported site, with detection of the highest quality version if that is possible?
Example video link:
http://gfycat.com/VainRichCub

mandb:
does anybody have documentation on how to build an extractor? Can possibly do a PR.

mandb:
added, PR #5439

julianrichen:
I was working on this as well, guess you beat me to it! Regarding your pull it might be better to use the built in json api they have as it will cut down on the code.:
```
http://gfycat.com/cajax/get/<video_id>
```

The url `http://gfycat.com/cajax/get/RequiredUnkemptBuzzard` returns:
```json
{
  "gfyItem": {
    "gfyId": "requiredunkemptbuzzard",
    "gfyName": "RequiredUnkemptBuzzard",
    "gfyNumber": "607648367",
    "userName": "anonymous",
    "width": 600,
    "height": 338,
    "frameRate": 29,
    "numFrames": 180,
    "mp4Url": "http://fat.gfycat.com/RequiredUnkemptBuzzard.mp4",
    "webmUrl": "http://zippy.gfycat.com/RequiredUnkemptBuzzard.webm",
    "gifUrl": "http://giant.gfycat.com/RequiredUnkemptBuzzard.gif",
    "gifSize": 5935649,
    "mp4Size": 1740123,
    "webmSize": 749951,
    "createDate": "1422584031",
    "views": 492324,
    "title": "Headshot!",
    "extraLemmas": null,
    "md5": "5dd3ed3ff36e37839cb11fb2faeff3a5",
    "tags": null,
    "nsfw": "0",
    "likes": "6",
    "dislikes": 0,
    "published": 1,
    "description": null,
    "copyrightClaimaint": null,
    "sar": "1",
    "url": "https://i.4cdn.org/wsg/1422071452954.webm",
    "source": "1",
    "dynamo": null,
    "subreddit": "all",
    "redditId": "32ol2t",
    "redditIdText": "a_skillful_headshot",
    "uploadGifName": null
  }
}
```
Might still make a pr if the api json version is prefered, not finished coding it.

mandb:
if there's json readily available probably best to use that, you can recycle the url match regex and extension preference.

dstftw:
Gfycat will be supported in the next version. Thanks for the report.

[...]

PR #5440, opened 2015-04-16
[gfycat] Add new extractor

julianrichen:
Gfycat support for issue #5394
Pull request is similar to #5439 but uses gfycats' built in json api, we discussed this in #5394.

First time coding python so any pointers on what I did wrong (if I did) is appreciated.

### DIFF
--- a/youtube_dl/extractor/gfycat.py
+++ b/youtube_dl/extractor/gfycat.py
@@ -1,0 +1,107 @@
+# coding: utf-8
+
+from __future__ import unicode_literals
+
+import datetime
+
+from .common import InfoExtractor
+
+class GfycatIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?gfycat\.com/(?P<id>[^/?#]+)'
+    _TESTS = [{
+        'url': 'http://gfycat.com/DeadlyDecisiveGermanpinscher',
+        'info_dict': {
+            'id':          'DeadlyDecisiveGermanpinscher',
+            'title':       'Ghost in the Shell',
+            'ext':         'mp4',
+            'upload_date': '20140913'
+        }
+    },{
+        'url': 'http://gfycat.com/pleasinghilariouskusimanse',
+        'info_dict': {
+            'id':          'pleasinghilariouskusimanse',
+            'title':       'PleasingHilariousKusimanse',
+            'ext':         'webm',
+            'upload_date': '20150412'
+        }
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/gfycat.py
```diff
@@ -0,0 +1,107 @@
...
+        # Pref: mp4, webm, gif
+        formats = [{
+            'format_id':  'mp4',
+            'ext':        'mp4',
+            'url':        json['mp4Url'],
+            'width':      json['width'],
+            'height':     json['height'],
+            'fps':        json['frameRate'],
+            'filesize':   json['mp4Size'],
+            'preference': '-1'
```
yan12125: Why strings here? It causes GfycatIE broken in Python 3.

julianrichen: I was not aware of that, I switched `-1` to an integer however rerunning the test I get
```
$ python test/test_download.py TestDownload.test_Gfycat
[Gfycat] DeadlyDecisiveGermanpinscher: Downloading video info
[info] Writing video description metadata as JSON to: DeadlyDecisiveGermanpinscher.info.json
[debug] Invoking downloader on u'http://giant.gfycat.com/DeadlyDecisiveGermanpinscher.gif'
[download] DeadlyDecisiveGermanpinscher.gif has already been downloaded
[download] 100% of 10.00KiB
F
FAIL: test_Gfycat (__main__.TestDownload)
  File "test/test_download.py", line 185, in test_template
    self.assertTrue(os.path.exists(tc_filename), msg='Missing file ' + tc_filename)
AssertionError: Missing file DeadlyDecisiveGermanpinscher.mp4

Ran 1 test in 0.415s

FAILED (failures=1)
```

It says it's downloading a gif but then says it can't find the mp4. What did I do incorrectly?

yan12125: FYI: formats with larger preference value is more preferable.

mandb:
you can take the preference code from my commit, if the user submits the link with a 'gif' or 'mp4' on the end (from the file host itself) then you should inflate the preference to that format if not specified.

youtube_dl/extractor/gfycat.py
julianrichen: Ahh thank you, guess I read common.py wrong. I have updated the PR.

dstftw:
Thanks, merged.

PR #5454, opened 2015-04-17
[megavideozeu] Add megavideoz.eu support with public domain video test

jbuchbinder:


jaimeMF:
Apart from that public domain, is the rest of the content uploaded with author permission? If you go to http://megavideoz.eu/videos/, you see that all the videos listed seem to be uploaded for http://watchseries-online.ch/ and I really doubt that they have the rights for distributing that material.

dstftw:
Isn't it enough according to [FAQ](https://github.com/rg3/youtube-dl#can-you-add-support-for-this-anime-video-site-or-site-which-shows-current-movies-for-free)?
>As a rule of thumb, if you cannot easily find a video that the service is quite obviously allowed to distribute (i.e. that has been uploaded by the creator, the creator's distributor, or is published under a free license), the service is probably unfit for inclusion to youtube-dl.

Otherwise we have to drop support for about 20 similar extractors.

jaimeMF:
I would personally just remove most of those, I don't think that having an old video in the public domain or one of the films produced by the Blender Institute means that the site is not used mostly for hosting copyrighted material (and it doesn't help that they don't usually provide a list with the more popular videos). I don't care if I'm the only one with that opinion, I just wanted to brought that up.

jbuchbinder:
It's up to you guys -- I just added the support for the site.

I've personally used the site as an easy web host for videos *I* share (hence the public domain video that I uploaded as a test video), but if you guys have ideological reasons not to support it, that's fine too.

jaimeMF:
> I've personally used the site as an easy web host for videos I share 

If people use it for hosting personal videos, it's ok. It's just that for, that kind of sites, it doesn't seem to be their main use.

Issue #5298, opened 2015-03-27
-f bestvideo+bestaudio tries mixing codecs that don't match


RickyRomero:
YouTube started adding support for 4K video at 60 FPS today, which youtube-dl actually downloads if you choose that format.  Here's their testing playlist:  https://www.youtube.com/playlist?list=PLyqf6gJt7KuHCoYaGNWPx8vkHlClmbO9u

Here's an example format list for one of the videos:

    [youtube] fEAyDNWaCto: Downloading webpage
    [youtube] fEAyDNWaCto: Extracting video information
    [youtube] fEAyDNWaCto: Downloading DASH manifest
    [info] Available formats for fEAyDNWaCto:
    format code  extension  resolution note
    171          webm       audio only DASH audio  128k , audio@128k (44100Hz), 2.57MiB
    140          m4a        audio only DASH audio  129k , m4a_dash container, aac  @128k (44100Hz), 3.12MiB
    141          m4a        audio only DASH audio  255k , m4a_dash container, aac  @256k (44100Hz), 6.19MiB
    160          mp4        256x144    DASH video  111k , 15fps, video only, 2.61MiB
    278          webm       256x144    DASH video  204k , webm container, VP9, 15fps, video only, 2.36MiB
    133          mp4        426x240    DASH video  246k , 30fps, video only, 5.87MiB
    242          webm       426x240    DASH video  293k , 30fps, video only, 4.42MiB
    243          webm       640x360    DASH video  541k , 30fps, video only, 8.19MiB
    134          mp4        640x360    DASH video  607k , 30fps, video only, 9.19MiB
    244          webm       854x480    DASH video  975k , 30fps, video only, 14.63MiB
    135          mp4        854x480    DASH video 1112k , 30fps, video only, 17.96MiB
    247          webm       1280x720   DASH video 1961k , 30fps, video only, 29.53MiB
    136          mp4        1280x720   DASH video 2224k , 30fps, video only, 36.12MiB
    302          webm       1280x720   DASH video 3123k , VP9, 60fps, video only, 47.73MiB
    248          webm       1920x1080  DASH video 3212k , 30fps, video only, 51.73MiB
    298          mp4        1280x720   DASH video 3332k , h264, 60fps, video only, 63.46MiB
    137          mp4        1920x1080  DASH video 4168k , 30fps, video only, 75.08MiB
    303          webm       1920x1080  DASH video 5200k , VP9, 60fps, video only, 81.94MiB
[...]

PR #5456, opened 2015-04-17
[YoutubeDL] Make `bestvideo+bestaudio/best` default format when merger is available

### LINKED ISSUE #5447: Document and make default --format='bestvideo+bestaudio/best'
Please document ```--format='bestvideo+bestaudio/best'``` in man page and consider making it default.

dstftw:
Closes #5447.

Pros:
+ Results in best overall quality.

Cons:
- Potential merge issues (#5298).
- Unable to watch video while downloading (by piping to video player).

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1390,9 +1390,15 @@ class YoutubeDL(object):
                         return False
 
                     requested_formats = info_dict['requested_formats']
-                    # Merge incompatible formats into mkv
+                    uncompatible_merge_warn = False
+                    # Incompatible formats will be merged into mkv
                     if not compatible_formats(requested_formats):
                         filename = os.path.splitext(filename)[0] + '.mkv'
+                        uncompatible_merge_warn = True
+                    if os.path.exists(encodeFilename(filename)):
+                        self.report_file_already_downloaded(filename)
+                        return
+                    if uncompatible_merge_warn:
                         self.report_warning('You have requested formats uncompatible for merge. '
                                             'The formats will be merged into mkv')
                     for f in requested_formats:
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1091,7 +1091,11 @@ class YoutubeDL(object):
 
         req_format = self.params.get('format')
         if req_format is None:
-            req_format = 'best'
+            req_format_list = []
+            if info_dict['extractor'] == 'youtube' and FFmpegMergerPP(self).available:
+                req_format_list.append('bestvideo+bestaudio')
+            req_format_list.append('best')
+            req_format = '/'.join(req_format_list)
         formats_to_download = []
         if req_format == 'all':
             formats_to_download = formats
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1373,7 +1373,29 @@ class YoutubeDL(object):
                                             ' The formats won\'t be merged')
                     else:
                         postprocessors = [merger]
-                    for f in info_dict['requested_formats']:
+
+                    def compatible_formats(formats):
+                        video, audio = formats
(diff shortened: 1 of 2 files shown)


jaimeMF:
In some issue report someone commented saying that for videos in low resolution it's more convenient to just get the normal formats, since they have the same or better quality. Checking these cases would probably complicate the code too much and I don't think it's worth the effort.

> Potential merge issues

We should make sure that the default formats work (otherwise users would get confused), I propose two alternatives:
* If bestaudio and bestvideo are not compatible, we merge into a mkv file
* Use `bestvideo[mp4]+bestaudio[m4a]/bestvideo[webm]+bestaudio[webm]/mp4` instead, with this option we may not get the best quality (I think in some cases there's either a webm or mp4 with better resolution, but I'm not really sure).

With this change the YoutubeIE test fails, we have to change the `format` field in `test/parameters.json`.

Apart from the previous comments, I like the change and agree that `bestvideo+bestaudio/best` should be the default

dstftw:
I think merging in mkv sounds reasonable since theoretically there may be a situation when there are no video/audio of the compatible formats available. Moreover, general video+audio format selection will benefit from it as well.

jaimeMF:
I also prefer defaulting to mkv. Some players may not support it, I don't think it's a big problem (it happens the same if some website only gives flv files), but some users may not like it.

We could also merge into mkv and use `bestvideo[mp4]+bestaudio[m4a]/bestvideo[webm]+bestaudio[webm]/bestvideo+bestaudio/mp4`, but that's quite ugly, I'd prefer to keep it as simple as possible.

youtube_dl/YoutubeDL.py
```diff
@@ -1369,7 +1373,29 @@ def dl(name, info):
...
+                                if video_ext in exts and audio_ext in exts:
+                                    return True
+                        # TODO: Check acodec/vcodec
+                        return False
+
+                    requested_formats = info_dict['requested_formats']
+                    # Merge incompatible formats into mkv
+                    if not compatible_formats(requested_formats):
+                        filename = os.path.splitext(filename)[0] + '.mkv'
+                        self.report_warning('You have requested formats uncompatible for merge. '
```
jaimeMF: Since `bestvideo+bestaudio/best` will be the default, maybe it's better to use `to_screen`.

jaimeMF:
Before merging this I think we should detect if the video has already been downloaded and merged. Currently if you don't use --keep-video, the files are redownloaded and merged again (if you prefer, I can try to implement it myself).

One issue with merging is that for long videos (like fEAyDNWaCto from #5298), ffmpeg takes a relatively long time to merge them. I don't know if we can do something to fix it.

youtube_dl/YoutubeDL.py
dstftw: Why? Uncompatible formats scenario is still expected to happen less often that compatible and can be treated as *reasonable-to-warn-about*.

dstftw:
I've added basic check for already merged file. Feel free to tweak it if that's not what you expected.

youtube_dl/YoutubeDL.py
jaimeMF: You are right, it's fine.

jaimeMF:
I have written an alternative in #5469, we have to use a temporary file with ffmpeg, otherwise we can end with a partial video if we cancel the merging.
(If you like it we can merge it to master directly or you cherry pick it in your branch, as you prefer)

dstftw:
Ok, lets just merge both to master.

2015-04-19 21:03 GMT+06:00 Jaime Marquínez Ferrándiz <
notifications@github.com>:

jaimeMF:
Cherry picked everything except the last commit (feccf29c876869f44a9c983977371073b9801a51, 81cd954a512e48fb5cccc3159cd0581088bff0b3 and 70947ea7b13d22a55756dc2b6d086e058edd91b1). Nice work!

PR #5469, opened 2015-04-19
Detect already merged videos

jaimeMF:
Without the '--keep-video' option the two files would be downloaded again and even using the option, ffmpeg would be run again, which for some videos can take a long time.
We use a temporary file with ffmpeg so that the final file only exists if it success


jaimeMF:
Cherry picked in 5b5fbc0867f0eb73416c10c9d692fceee92b5766

ldmldmldm:
Thank you for this feature!

PR #5473, opened 2015-04-20
[bandcamp] fix video_id parsing

quinq:
Hi,
this should fix https://github.com/rg3/youtube-dl/issues/4861
The problem was that the regex stopped at the first occurrence of id:, which could be art_id: for example.


jaimeMF:
Cherry picked in b524a001d68767298272ef59274c826a6ea6a1b7, thanks!

Issue #5043, opened 2015-02-23
FIle system: Specify a destination directory for the video


keybounce:
Right now, youtube-dl will always save the video where I am at.

I'd like to be able to put a default download directory into my config (for example, /Volumes/Kleiman-Movies/Videos\ to\ Watch), and on the command line, specify something like "etho" -- which would result in a final directory of
/Volumes/Kleiman-Movies/Videos\ to\ Watch/etho/

phihag:
Right now, you can [create a configuration file](https://github.com/rg3/youtube-dl/blob/master/README.md#configuration) with the contents
```
-o  "/Volumes/Kleiman-Movies/Videos to Watch/%(title)s-%(id)s.%(ext)s"
```
and all downloads will go to that folder by default.

I think adding another option `--subdirectory` that maps to a template `%(subdirectory)s` would be possible.

mitchellmebane:
I think more user-friendly behavior would be if the value of `-o` evaluates, pre-expansion, to an existing directory, use that directory but append the default template.

PR #5495, opened 2015-04-21
Allow specifying a directory as the output template

mitchellmebane:
Possible solution to #5043.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -16,10 +16,12 @@ import json
 import xml.etree.ElementTree
 
 from youtube_dl.utils import (
+    add_defaults_if_directory,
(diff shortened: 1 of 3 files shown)


jaimeMF:
I see one problem with that approach: currently youtube-dl creates directories from the output template that doesn't exist (for example for `'foo/bar/%(title)s.%(ext)s'` it will create `foo/bar`), I think that being inconsistent would be bad. Maybe it would be better to use a new option (probably just to the command line, if you are using youtube-dl from a python program you can manually join the directory and the default template).

mitchellmebane:
Hm, alright.  I'll try to find the time to work it over.

PR #5513, opened 2015-04-23
Added support for mpv if mplayer is not available.

mrkrossxdx:
This commit simply try to use mpv if mplayer is not found.

### DIFF
--- a/youtube_dl/downloader/mplayer.py
+++ b/youtube_dl/downloader/mplayer.py
@@ -16,11 +16,18 @@ class MplayerFD(FileDownloader):
         self.report_destination(filename)
         tmpfilename = self.temp_name(filename)
 
-        args = [
-            'mplayer', '-really-quiet', '-vo', 'null', '-vc', 'dummy',
-            '-dumpstream', '-dumpfile', tmpfilename, url]
+        args = []
         # Check for mplayer first
-        if not check_executable('mplayer', ['-h']):
+        if check_executable('mplayer', ['-h']):
+            args = [
+                'mplayer', '-really-quiet', '-vo', 'null', '-vc', 'dummy',
+                '-dumpstream', '-dumpfile', tmpfilename, url]
+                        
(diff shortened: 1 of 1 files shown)


youtube_dl/downloader/mplayer.py
```diff
@@ -16,11 +16,18 @@ def real_download(self, filename, info_dict):
...
+            args = [
+                'mplayer', '-really-quiet', '-vo', 'null', '-vc', 'dummy',
+                '-dumpstream', '-dumpfile', tmpfilename, url]
+                        
+        # Check for mpv
+        elif check_executable('mpv', ['-h']):
+            args = [
+                'mpv', '-really-quiet', '--vo=null', '--stream-dump=' + tmpfilename, url]
+        else:
             self.report_error('MMS or RTSP download detected but "%s" could not be run' % args[0])
```
dstftw: This will fail if neither mplayer nor mpv is available.

mrkrossxdx:
I'm sorry, but I don't understand what you mean, is the original code supposed to work also if mplayer is not available?

I tried to download a MMS stream and youtube-dl fails complaining about the absence of mplayer. I have writes a small shell script that calls mpv with the correct parameters and everything worked fine, so I had assumed that mplayer is required for download MMS stream, but given my poor knowledge of the codebase I can be wrong.

Thank you for taking the time to review this.

Best regards,

Marco

dstftw:
[`args` is empty list](https://github.com/rg3/youtube-dl/pull/5513/files#diff-ce2afa610d98a8227657cea63af36563R19), [access to 0 item of empty list](https://github.com/rg3/youtube-dl/pull/5513#discussion-diff-29000180R31) => `IndexError`

mrkrossxdx:
D'oh, I don't have noted that the error message use the args list, thank you very much for this clarification. I will create another pull request with the correct error message.

PR #5521, opened 2015-04-24
Added support for mpv if mplayer is not available (new version)

mrkrossxdx:


dstftw:
Download is not interruptible by `Ctrl+C` when using mpv.

mrkrossxdx:
What OS? ctrl+c works fine here on Arch Linux (tested on both mpv 0.8.3 and 0.9.0), maybe is a bug that affect only some platforms.

dstftw:
Windows-2003Server-5.2.3790-SP2 and Windows-7-6.1.7601-SP1 with [these binaries](http://mpv.srsfckn.biz/).

dstftw:
Works fine under ubuntu. Turns out to be a windows/these particular binaries issue.

dstftw:
Thanks.

Issue #5508, opened 2015-04-23
--max-quality ignored when used without -f


Tailszefox:
Example:

```
$ youtube-dl -F https://www.youtube.com/watch\?v\=lYePxF7CJ6I                                                                                                                                                                           
[youtube] lYePxF7CJ6I: Downloading webpage
[youtube] lYePxF7CJ6I: Extracting video information
[youtube] lYePxF7CJ6I: Downloading DASH manifest
[info] Available formats for lYePxF7CJ6I:
format code  extension  resolution note
171          webm       audio only DASH audio  126k , audio@128k (44100Hz), 5.42MiB
140          m4a        audio only DASH audio  129k , m4a_dash container, aac  @128k (44100Hz), 6.14MiB
141          m4a        audio only DASH audio  255k , m4a_dash container, aac  @256k (44100Hz), 12.19MiB
160          mp4        256x144    DASH video  120k , 15fps, video only, 4.61MiB
278          webm       256x144    DASH video  134k , webm container, VP9, 15fps, video only, 3.03MiB
133          mp4        426x240    DASH video  274k , 30fps, video only, 9.92MiB
242          webm       426x240    DASH video  289k , 30fps, video only, 5.67MiB
243          webm       640x360    DASH video  485k , 30fps, video only, 10.02MiB
134          mp4        640x360    DASH video  603k , 30fps, video only, 11.21MiB
244          webm       854x480    DASH video  851k , 30fps, video only, 16.15MiB
135          mp4        854x480    DASH video 1109k , 30fps, video only, 20.30MiB
302          webm       1280x720   DASH video 2678k , VP9, 60fps, video only, 41.82MiB
298          mp4        1280x720   DASH video 3310k , h264, 60fps, video only, 61.45MiB
303          webm       1920x1080  DASH video 4966k , VP9, 60fps, video only, 79.56MiB
299          mp4        1920x1080  DASH video 5523k , h264, 60fps, video only, 107.74MiB
17           3gp        176x144    
36           3gp        320x240    
5            flv        400x240    
43           webm       640x360    
18           mp4        640x360    
22           mp4        1280x720   (best)
```

```
$ youtube-dl --max-quality 99 https://www.youtube.com/watch\?v\=lYePxF7CJ6I                                                                                                                                                             
[youtube] lYePxF7CJ6I: Downloading webpage
[youtube] lYePxF7CJ6I: Extracting video information
[youtube] lYePxF7CJ6I: Downloading DASH manifest
[...]

PR #5523, opened 2015-04-25
Remove the --max-quality option

jaimeMF:
It doesn't work well with 'bestvideo' and 'bestaudio' because they are usually before the max quality.
Format filters should be used instead, they are more flexible and don't require the requested quality to exist for each video.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,6 @@ which means you can modify it, redistribute it or use it however you like.
     -f, --format FORMAT              Video format code, see the "FORMAT SELECTION" for all the info
     --all-formats                    Download all available video formats
     --prefer-free-formats            Prefer free video formats unless a specific one is requested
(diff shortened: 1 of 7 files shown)


dstftw:
LGTM. [README.md](https://github.com/rg3/youtube-dl/blob/master/README.md) needs to be cleaned up as well.

jaimeMF:
Updated the README, for some reason I only searched for `format_limit`.
(For reference, the issue reporting that `--max-quality` didn't work is #5508)

LouisDK1:
I'm pretty sad this fuction was removed. I need the best quality for 720p and that open was great for that.

jaimeMF:
Use `-f best[height<=?720]` instead, which works better.

LouisDK1:
Thanks jaimeMF

PR #5575, opened 2015-05-02
Get audio first when downloading DASH media

fstirlitz:
The reason is twofold: audio data is usually smaller (more "dense", i.e. has a higher bitrate), and often more important than video data (e.g. when downloading talk show episodes). A partial audio download is therefore usually more useful, as in the same amount of time youtube-dl could capture a larger timeslice of a more valuable part of the media stream.

All of this matters little for a completed download, of course.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1116,7 +1116,7 @@ class YoutubeDL(object):
                                 if self.params.get('merge_output_format') is None
                                 else self.params['merge_output_format'])
                             selected_format = {
-                                'requested_formats': formats_info,
+                                'requested_formats': (formats_info[1], formats_info[0]),
                                 'format': '%s+%s' % (formats_info[0].get('format'),
                                                      formats_info[1].get('format')),
                                 'format_id': '%s+%s' % (formats_info[0].get('format_id'),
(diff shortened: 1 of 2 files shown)


jaimeMF:
> A partial audio download is therefore usually more useful

I disagree, apart from talk show and similar, I think the video is usually as important or more than the audio. I would see little benefit in having the audio downloaded if I can't watch the actual video.
And for talk shows you can just use `-f bestaudio`, which will download only the audio.

Also: downloading the audio first can be surprising if you use `-f bestvideo+bestaudio`.

fstirlitz:
But I don't want *only* the audio downloaded, but audio *first*.

Don't forget that audio is smaller. Even for the case when video is more important, the amount of time before the content you want *more* starts downloading is smaller than in the opposite case (where you care more about audio but have to wait for video first). And if you don't care that much which downloads first, you still get a larger slice of the stream in the same amount of time.

Two other alternatives would be to implement parallel downloads, or to have `-f bestaudio+bestvideo` download audio first (it seems it isn't currently allowed at all).

dstftw:
As you correctly noticed, currently video+audio format selection is restricted video first and audio second. As we process format entries from left to right that's reasonable to download files in order they go in format selection string. So, to allow audio to be downloaded first we will need to allow audio+video order.

PR #5581, opened 2015-05-02
Allow downloading DASH streams in any order

fstirlitz:
In response to the discussion in #5575. Please note that the `'format'` and `'format_id'` fields in the `info_json` are normalised back to video+audio order; this may or may not be desirable.

I still think downloading audio first should be the default.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1105,32 +1105,31 @@ class YoutubeDL(object):
                         formats_info = (self.select_format(format_1, formats),
                                         self.select_format(format_2, formats))
                         if all(formats_info):
-                            # The first format must contain the video and the
-                            # second the audio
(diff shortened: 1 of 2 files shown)


fstirlitz:
Any particular reasons for not merging this or is it just general overload?

remitamine:
i don't think that normalizing it is good idea, the options you point at here https://github.com/rg3/youtube-dl/pull/5575#issuecomment-98386403 are better.

fstirlitz:
I don't understand your comment. This patch actually implements one of them.

remitamine:
> to have -f bestaudio+bestvideo download audio first

here you said you want this to download audio before the video.
> are normalised back to video+audio order

as i understand this mean if i put -f bestaudio+bestvideo it will change to look like -f bestvideo+bestaudio.

PR #5585, opened 2015-05-03
Fix for unicode filenames

maleficarium:
Fix for the issue of unicode file names discussed here: https://github.com/rg3/youtube-dl/issues/5527 https://github.com/rg3/youtube-dl/issues/4787 by hashing the input and output filenames before processing them and then restoring them.

### DIFF
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -4,7 +4,7 @@ import io
 import os
 import subprocess
 import time
-
+import hashNames
 
 from .common import AudioConversionError, PostProcessor
 
@@ -131,12 +131,17 @@ class FFmpegPostProcessor(PostProcessor):
             os.stat(encodeFilename(path)).st_mtime for path in input_paths)
 
         files_cmd = []
+        original_names = []
         for path in input_paths:
-            files_cmd.extend([encodeArgument('-i'), encodeFilename(path, True)])
+            hashed_path = hashNames.hashRename(path)
+            os.rename(path, hashed_path)
+            original_names.append([hashed_path, path])
+            files_cmd.extend([encodeArgument('-i'), encodeFilename(hashed_path, True)])
+        hashed_out_path = hashNames.hashRename(out_path)
         cmd = ([encodeFilename(self.executable, True), encodeArgument('-y')] +
                files_cmd +
                [encodeArgument(o) for o in opts] +
-               [encodeFilename(self._ffmpeg_filename_argument(out_path), True)])
+               [encodeFilename(self._ffmpeg_filename_argument(hashed_out_path), True)])
 
         if self._downloader.params.get('verbose', False):
             self._downloader.to_screen('[debug] ffmpeg command line: %s' % shell_quote(cmd))
@@ -146,6 +151,9 @@ class FFmpegPostProcessor(PostProcessor):
             stderr = stderr.decode('utf-8', 'replace')
             msg = stderr.strip().split('\n')[-1]
             raise FFmpegPostProcessorError(msg)
+        for o_name in original_names:
+            os.rename(o_name[0], o_name[1])
+        os.rename(hashed_out_path, out_path)
         self.try_utime(out_path, oldest_mtime, oldest_mtime)
 
     def run_ffmpeg(self, path, out_path, opts):
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -570,8 +570,8 @@ class YoutubeDL(object):
             # Temporary fix for #4787
             # 'Treat' all problem characters by passing filename through preferredencoding
             # to workaround encoding issues with subprocess on python2 @ Windows
-            if sys.version_info < (3, 0) and sys.platform == 'win32':
-                filename = encodeFilename(filename, True).decode(preferredencoding())
+            #if sys.version_info < (3, 0) and sys.platform == 'win32':
+            #    filename = encodeFilename(filename, True).decode(preferredencoding())
             return filename
         except ValueError as err:
             self.report_error('Error in output template: ' + str(err) + ' (encoding: ' + repr(preferredencoding()) + ')')
--- a/youtube_dl/postprocessor/hashNames.py
+++ b/youtube_dl/postprocessor/hashNames.py
@@ -1,0 +1,9 @@
+import os
+import hashlib
+
+#Hash the file names to drop unicode characters for FFMPEG
+def hashRename(fileN):
+    ext = os.path.splitext(fileN)
+    fileHash = hashlib.sha224(fileN.encode('utf-8')).hexdigest()
+    hash_name = fileHash+ext[1]
+    return hash_name


youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -131,12 +131,17 @@ def run_ffmpeg_multiple_files(self, input_paths, out_path, opts):
...
 
         files_cmd = []
+        original_names = []
         for path in input_paths:
-            files_cmd.extend([encodeArgument('-i'), encodeFilename(path, True)])
+            hashed_path = hashNames.hashRename(path)
+            os.rename(path, hashed_path)
+            original_names.append([hashed_path, path])
+            files_cmd.extend([encodeArgument('-i'), encodeFilename(hashed_path, True)])
+        hashed_out_path = hashNames.hashRename(out_path)
```
dstftw: Why bother with hashing and not just use `NamedTemporaryFile`?

dstftw:
1. Python3 is broken. Generally, this tempnames workaround has to be triggered on python2 @ Windows only.
2. You miss [lots](https://github.com/maleficarium/youtube-dl/blob/master/youtube_dl/postprocessor/ffmpeg.py#L189) [of](https://github.com/maleficarium/youtube-dl/blob/master/youtube_dl/postprocessor/embedthumbnail.py#L62) [other](https://github.com/maleficarium/youtube-dl/blob/master/youtube_dl/downloader/rtmp.py#L29) [places](https://github.com/maleficarium/youtube-dl/blob/master/youtube_dl/downloader/external.py#L67) when filenames are used with subprocessing and might need to be fixed as well.
3. I would like to avoid stuffing renaming code all over the problematic places and make it as generic as possible, ideally RAII-like.

jaimeMF:
This should only be done on windows, since it pronably already works fine other OS.

This may not apply to windows: you can't just use a new filename, because you would write to the current working directory and you may not have write permissions for it (for example you could be running youtube-dl from `/` but use `-o ~/'Downloads/%(title)s-%(id)s.%(ext)s`).

maleficarium:
@jaimeMF Nothing is written in the current directory if it's not the output path. If a different path is specified in -o the files are already stored there and are only renamed there. I don't see how write permissions could be an issue.

@dstftw Is using the video ID only as the filename and only renaming the file at the end of all operations an option? e.g. https://www.youtube.com/watch?v=xCjMABqWdEM would result in "xCjMABqWdEM" as the initial filename instead of "FateGrand Order 7週連続TV-CM 第1弾 セイバー編-xCjMABqWdEM".
This would alter the current file outputs but would solve the issue of unicode characters without the need for renaming or the need to check for the python version & OS.

yan12125:
Another side note: you need to rename files back if postprocessing fails.

jaimeMF:
> Nothing is written in the current directory if it's not the output path. If a different path is specified in -o the files are already stored there and are only renamed there. I don't see how write permissions could be an issue.

You are actually writing to the working directory. For example, here I run youtube-dl from `/` but I use `-o '~/Desktop/foo/%(id)s.%(ext)s'`, and it tries to rename the file to `/<the has>.ext`(which obviously cannot do):
```shell
jaime@jaime /> youtube-dl --ignore-config -v test:youtube --extract-audio --audio-format mp3 -o '~/Desktop/foo/%(id)s.%(ext)s'
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--ignore-config', u'-v', u'test:youtube', u'--extract-audio', u'--audio-format', u'mp3', u'-o', u'~/Desktop/foo/%(id)s.%(ext)s']
[debug] youtube-dl version 2015.04.28
[debug] Git HEAD: 21aa58a
[debug] Python version 2.7.9 - Darwin-13.4.0-x86_64-i386-64bit
[debug] exe versions: avconv 11.3, avprobe 11.3, ffmpeg 2.6.2, ffprobe 2.6.2, rtmpdump 2.4
[debug] Proxy map: {}
[TestURL] Test URL: http://www.youtube.com/watch?v=BaW_jenozKc
[youtube] BaW_jenozKc: Downloading webpage
[youtube] BaW_jenozKc: Extracting video information
[youtube] BaW_jenozKc: Downloading DASH manifest
[debug] Invoking downloader on 'https://r9---sn-h5q7ener.googlevideo.com/videoplayback?id=05a5bf8de9e8cca7&itag=141&source=youtube&requiressl=yes&pl=18&mv=m&ms=au&mm=31&nh=IgpwcjAxLm1hZDA2KgkxMjcuMC4wLjE&ratebypass=yes&mime=audio/mp4&gir=yes&clen=315992&lmt=1387961817988214&dur=9.891&upn=DSiTaOqin0o&fexp=900720,907263,934954,938028,9407109,9408420,9408457,9408706,9408776,9409174,9413010,9413153,948124,952612,952637,952642,957201&sver=3&mt=1430655812&signature=3409D3572B0FAA86F0FD4BE5E3AF1A5F8C75E805.4B422AB70D9E037075E0B93F79C0310BA3366B91&key=dg_yt0&ip=79.156.0.141&ipbits=0&expire=1430677441&sparams=ip,ipbits,expire,id,itag,source,requiressl,pl,mv,ms,mm,nh,ratebypass,mime,gir,clen,lmt,dur'
[download] Destination: ~/Desktop/foo/BaW_jenozKc.m4a
[download] 100% of 308.59KiB in 00:00
[ffmpeg] Correcting container in "~/Desktop/foo/BaW_jenozKc.m4a"
  File "~/local/bin/youtube-dl", line 13, in <module>
    youtube_dl.main()
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/__init__.py", line 401, in main
    _real_main(argv)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/__init__.py", line 391, in _real_main
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/YoutubeDL.py", line 1482, in download
    res = self.extract_info(url)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/YoutubeDL.py", line 661, in extract_info
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/YoutubeDL.py", line 714, in process_ie_result
    extra_info=extra_info)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/YoutubeDL.py", line 661, in extract_info
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/YoutubeDL.py", line 707, in process_ie_result
    return self.process_video_result(ie_result, download=download)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/YoutubeDL.py", line 1153, in process_video_result
    self.process_info(new_info)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/YoutubeDL.py", line 1465, in process_info
    self.post_process(filename, info_dict)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/YoutubeDL.py", line 1527, in post_process
    files_to_delete, info = pp.run(info)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/postprocessor/ffmpeg.py", line 632, in run
    self.run_ffmpeg(filename, temp_filename, options)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/postprocessor/ffmpeg.py", line 160, in run_ffmpeg
    self.run_ffmpeg_multiple_files([path], out_path, opts)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/postprocessor/ffmpeg.py", line 137, in run_ffmpeg_multiple_files
    os.rename(path, hashed_path)
OSError: [Errno 13] Permission denied
```

maleficarium:
@jaimeMF Damn, I messed that up. No matter the permissions, it shouldn't be creating temporary files in the current working directory. Back to the drawing board then.

svnpenn:
Unicode, that is to say UTF-16LE [uses two bytes][1] per character. UTF-8 uses
one or two depending. Consider this command

~~~
$ youtube-dl --get-title gEMAxCAt2eQ | od -tx1c
0000000  54  69  eb  73  74  6f  20  66  65  61  74  75  72  69  6e  67
          T   i 353   s   t   o       f   e   a   t   u   r   i   n   g
~~~

The third character, LATIN LETTER SMALL E WITH DIAERESIS, is represented in
Unicode as

    00 EB

or UTF-8 as

    C3 AB

Notice carefully that youtube-dl is doing neither. So what happens is you get
shit like this when you try to pipe it

    $ youtube-dl --get-title gEMAxCAt2eQ | cat
    Ti▒sto featuring BT - Break My Fall

Please fix this. Regardless of Unicode or UTF-8, this character needs to be two
bytes.

[1]:http://wikipedia.org/wiki/Unicode

svnpenn:
youtube-dl for Windows is using [Windows-1252][1]. In my opinion it should
really be UTF-8. Notice the difference

~~~
$ youtube-dl --get-title gEMAxCAt2eQ | cat
Ti▒sto featuring BT - Break My Fall

$ youtube-dl --get-title gEMAxCAt2eQ | iconv -f cp1252 | cat
Tiësto featuring BT - Break My Fall
~~~

[1]:http://wikipedia.org/wiki/Windows-1252

PR #5601, opened 2015-05-04
[noco] Retrieve video language according to user options

Tassatux:
Get the user audio language preference in order to retrieve video with chosen language instead of 'fr'.


dstftw:
What is the benefit of this improvement?

Tassatux:
Current version of Noco extractor only get a show with French as audio language. Some shows are available in multiple language, for example French and Japanese for some animes.
With this modification, the Noco extractor will get the show audio language according to your account preferences and fallback to French as previously if your preferred audio language isn't available for the show.

dstftw:
Can you add a test video with multiple languages?

dstftw:
Thanks.

PR #5616, opened 2015-05-05
Update AirMozilla test case

blissland:
The existing AirMozzilla test case currently fails for thumbnail and categories


blissland:
Seems to have fixed itself

PR #5628, opened 2015-05-06
[ArchiveOrg] Fix incorrect url condition

blissland:
The condition for assigning to json_url is the wrong way round:

currently for url: aaa.com/xxx

we get:

aaa.com/xxx&output=json

instead of the correct value:

aaa.com/xxx?output=json


jaimeMF:
Thanks!
(The strange part is that they accept both formats).

PR #2817, opened 2015-05-06
faststart

svnpenn:
If I run this command

~~~
youtube-dl -f 22 -x --add-metadata --id 8ATu1BiOPZA
~~~

It does almost everything I need

- downloads the file with the highest quality audio (192k)
- extracts said audio without re-encoding
- adds metadata

However using `-x` invokes FFmpeg. FFmpeg by default will not set [`faststart`](http://multimedia.cx/eggs/improving-qt-faststart)
in fact if a file already has `faststart` it will even be removed. Without
`faststart` these files will not play in my car. This can only be remedied by
running FFmpeg with [the correct flag](http://superuser.com/a/616799)

~~~
ffmpeg -i 8ATu1BiOPZA.m4a -c copy -movflags faststart outfile.m4a
~~~

It would be nice if `youtube-dl` had this as an option, or perhaps even
as the default when using FFmpeg.

### DIFF
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -216,7 +216,7 @@ class FFmpegExtractAudioPP(FFmpegPostProcessor):
                 # Lossless, but in another container
                 acodec = 'copy'
                 extension = 'm4a'
-                more_opts = ['-bsf:a', 'aac_adtstoasc']
+                more_opts = ['-bsf:a', 'aac_adtstoasc', '-movflags', 'faststart']
             elif filecodec in ['aac', 'mp3', 'vorbis', 'opus']:
                 # Lossless if possible
                 acodec = 'copy'
@@ -567,7 +567,7 @@ class FFmpegMetadataPP(FFmpegPostProcessor):
         temp_filename = prepend_extension(filename, 'temp')
 
         if info['ext'] == 'm4a':
-            options = ['-vn', '-acodec', 'copy']
+            options = ['-vn', '-acodec', 'copy', '-movflags', 'faststart']
         else:
             options = ['-c', 'copy']
 
@@ -619,7 +619,7 @@ class FFmpegFixupM4aPP(FFmpegPostProcessor):
         filename = info['filepath']
         temp_filename = prepend_extension(filename, 'temp')
 
-        options = ['-c', 'copy', '-f', 'mp4']
+        options = ['-c', 'copy', '-f', 'mp4', '-movflags', 'faststart']
         self._downloader.to_screen('[ffmpeg] Correcting container in "%s"' % filename)
         self.run_ffmpeg(filename, temp_filename, options)
 


svnpenn:
@jaimeMF please consider this pull request

svnpenn:
@jaimeMF @phihag please consider this pull request

jaimeMF:
Is that the only place where it's needed?, will `--add-metadata` keep it?

(Side note: I don't personally like that issues can be converted to PR, now it looks like it was a PR since the begining and some comments can be out of context. Open a new PR the next time)

svnpenn:
Sorry yes you are right, I have update the commit

svnpenn:
Sorry yes you are right, I have updated the commit

svnpenn:
Yes you are right, it can be needed when

- extracting audio
- adding metadata
- fixing m4a

I have updated the commit

svnpenn:
@jaimeMF @phihag please consider this pull request

svnpenn:
@jaimeMF @phihag please consider this pull request

svnpenn:
@jaimeMF @dstftw can someone comment on this? I have had no comments since I added a commit over a month ago

dstftw:
Have you ensured this has no side effects on avconv 10+ and recent ffmpeg?

svnpenn:
My FFmpeg was built last month, I can test with newer version if necessary

dstftw:
Ideally, it should work with avconv 10+ and ffmpeg of approximately the same release date.

svnpenn:
I have not done extensive testing, but as far as I am aware it does work with what you have listed

svnpenn:
@dstftw @jaimeMF please someone consider this pull request. It has been 6 weeks since the commit was submitted

dstftw:
Shouldn't `-movflags faststart` be applied to mp4 videos as well?
I'm still in doubts making it default is a good idea:
>-movflags faststart
    Run a second pass moving the index (moov atom) to the beginning of the file. **This operation can take a while**, and **will not work in various situations** such as fragmented output, thus it is not enabled by default. 

Probably, making an option would be better.

svnpenn:
@dstftw thanks for the feedback, I will consider writing a patch that way

dstftw:
There is also an upcoming PR #5942 that adds support for passing arbitrary command line to postprocessor with `--postprocessor-args`. In order not to complicate youtube-dl with more options, will `--postprocessor-args` fit your needs?

svnpenn:
@dstftw yeah! I will see if I can try that patch

svnpenn:
#5733

dstftw:
In the next version you'll be able to pass arbitrary command line to ffmpeg, so a command to force `faststart` will look as follows:
```
youtube-dl -f 22 -x --add-metadata --id --postprocessor-args '-movflags faststart' 8ATu1BiOPZA
```
Thanks for your efforts, I'm closing this since a more generic solution has been incorporated.

Kebabpizza:
Note that " (not ') seems required on Windows, or you'll get a ValueError: No closing quotation

PR #5638, opened 2015-05-07
[BildIE] Fix ampersands in xml attributes & update test thumbnails

blissland:
Currently the BildIE test case fails for the following reasons:

1/ The XML downloaded contains raw ampersands in attributes
2/ The expected thumbnail has changed.

This pull request fixes these 2 issues


dstftw:
Thanks.

PR #5641, opened 2015-05-07
[YoutubeDL] Do not force bestvideo+bestaudio when outtmpl is stdout

### LINKED ISSUE #5627: Dumping to `stdout` is broken for version `2015.04.26`
As the subject says, if one tries to dump to `stdout` a video whose ID starts with an underscore (`_`), `youtube-dl` instead tries to save it to disk.

> youtube-dl --output - -- 'https://www.youtube.com/watch?v=_KYvp8PrCFc'

dstftw:
Fixes #5627.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1086,7 +1086,9 @@ class YoutubeDL(object):
         req_format = self.params.get('format')
         if req_format is None:
             req_format_list = []
-            if info_dict['extractor'] in ['youtube', 'ted'] and FFmpegMergerPP(self).available:
+            if (self.params.get('outtmpl', DEFAULT_OUTTMPL) != '-'
+                    and info_dict['extractor'] in ['youtube', 'ted']
+                    and FFmpegMergerPP(self).available):
                 req_format_list.append('bestvideo+bestaudio')
             req_format_list.append('best')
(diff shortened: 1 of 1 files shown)


yan12125:
When outtmpl is ```-```, bestvideo+bestaudio *works* . Maybe a better alternative is adding a warning message to indicate youtube-dl is doing something different from what you think.

jaimeMF:
I think that using `best` with `-o -` is not that bad, since it's conventional to write the result to stdout. With the warning you force people to manually set `-f best`, which can be a bit annoying if you only use it from time to time with `youtube-dl URL -o - | mplayer` (although they can just use a wrapper for that).

dstftw:
@yan12125 as @jaimeMF already mentioned, it does work not as expected. `-` output template is treated by youtube-dl exclusively as streaming to stdout therefore there is a reasonable idiomatic way to use such a tempate - piping to media player with playback starting immediately. Since it's not possible to provide muxing *on-the-fly* (at least while it's not implemented in youtube-dl) I consider it very reasonable to force `-f best` for `-` output template, at least in order not to break idiomatic *pipe-to-media-player* use case scenario.

yan12125:
I see the rationales. Default to ```-f best``` when users intend to write the video to stdout is indeed more intuitive. Sorry I didn't notice that before. Just merge it.

PR #5642, opened 2015-05-07
[BetIE] Updated test data

blissland:
The current tests fail as the test data is out of date. I have updated the data so the tests now pass.

### DIFF
--- a/youtube_dl/extractor/bet.py
+++ b/youtube_dl/extractor/bet.py
@@ -16,11 +16,11 @@ class BetIE(InfoExtractor):
         {
             'url': 'http://www.bet.com/news/politics/2014/12/08/in-bet-exclusive-obama-talks-race-and-racism.html',
             'info_dict': {
-                'id': '740ab250-bb94-4a8a-8787-fe0de7c74471',
+                'id': '500406c1-0d86-4f40-880d-eab2ec1f7587',
                 'display_id': 'in-bet-exclusive-obama-talks-race-and-racism',
                 'ext': 'flv',
-                'title': 'BET News Presents: A Conversation With President Obama',
-                'description': 'md5:5a88d8ae912c1b33e090290af7ec33c6',
+                'title': 'A Conversation With President Obama',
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/bet.py
```diff
@@ -16,11 +16,11 @@ class BetIE(InfoExtractor):
         {
             'url': 'http://www.bet.com/news/politics/2014/12/08/in-bet-exclusive-obama-talks-race-and-racism.html',
             'info_dict': {
-                'id': '740ab250-bb94-4a8a-8787-fe0de7c74471',
+                'id': 'e0c430e7-14e8-414d-bd9d-c14c55cd406f',
```
jaimeMF: I get `2505a456-9506-4078-bf89-9a3b7d797679`

youtube_dl/extractor/bet.py
```diff
@@ -35,7 +35,7 @@ class BetIE(InfoExtractor):
         {
             'url': 'http://www.bet.com/video/news/national/2014/justice-for-ferguson-a-community-reacts.html',
             'info_dict': {
-                'id': 'bcd1b1df-673a-42cf-8d01-b282db608f2d',
+                'id': '0291d0f7-b8da-4423-9c5a-ebfa0368ef53',
```
jaimeMF: And `62e7621a-caac-4ccb-b63d-2ebbb61e64a6 here.

blissland:
Mmm... the id seems to be very dynamic. What's the best way to handle this: match against a regex: (8 chars - 4 chars - 4 chars - 12 chars) or leave out the id from the test info_dict?

jaimeMF:
No, the id must be in the info_dict and should be fixed (we should be able to assume with some confidence that it's unique). I guess that in this case the best option would be to use `news/national/2014/justice-for-ferguson-a-community-reacts` and `news/national/2014/a-conversation-with-president-obama`, since that's what the xml url uses.

blissland:
Ok I updated the pull request to use your suggestion about what to use as the id

jaimeMF:
Thanks!

PR #5657, opened 2015-05-09
[EroProfile] Fix 403 (due to bad URL)

frenchy1983:
Before:

$ python test/test_download.py TestDownload.test_EroProfile
...
ERROR: unable to download video data: HTTP Error 403: Forbidden

After:

$ python test/test_download.py TestDownload.test_EroProfile
...
OK


frenchy1983:
Of course there was a util to do that. Thanks @dstftw!

PR #5658, opened 2015-05-09
[BRIE] Updated two test cases

blissland:
The videos for the first 2 test cases no longer exist. Replaced with new videos.


dstftw:
The second test was an audio test. Can you add such test as well?

blissland:
Added audio test.

dstftw:
Thanks.

PR #5664, opened 2015-05-10
[voicerepublic] Add new extractor

duncankl:
Hello! Just a little extractor for Voice Republic, which hosts spoken word audio content.

### DIFF
--- a/youtube_dl/extractor/voicerepublic.py
+++ b/youtube_dl/extractor/voicerepublic.py
@@ -1,0 +1,55 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+from ..compat import (
+    compat_urllib_request,
+)
+
+
+class VoiceRepublicIE(InfoExtractor):
+    _VALID_URL = r'https?://voicerepublic\.com/talks/(?P<id>[0-9a-z-]+)'
+    _TEST = {
+        'url': 'https://voicerepublic.com/talks/watching-the-watchers-building-a-sousveillance-state',
+        'md5': '0554a24d1657915aa8e8f84e15dc9353',
+        'info_dict': {
+            'id': '2296',
+            'ext': 'm4a',
+            'title': 'Watching the Watchers: Building a Sousveillance State',
+            'thumbnail': 'https://voicerepublic.com/system/flyer/2296.png',
+            'description': 'md5:715ba964958afa2398df615809cfecb1',
+            'creator': 'M. C. McGrath',
+        }
+    }
+
+    def _real_extract(self, url):
+        display_id = self._match_id(url)
+        req = compat_urllib_request.Request(url)
+        # Older versions of Firefox get redirected to an "upgrade browser" page
+        req.add_header('User-Agent', 'youtube-dl')
+        webpage = self._download_webpage(req, display_id)
+        thumbnail = self._og_search_thumbnail(webpage)
+        video_id = self._search_regex(r'/(\d+)\.png', thumbnail, 'id')
+
+        if '<div class=\'vr-player jp-jplayer\'' in webpage:
+            formats = [{
+                'url': 'https://voicerepublic.com/vrmedia/{}-clean.{}'.format(video_id, ext),
+                'ext': ext,
+                'format_id': ext,
+                'vcodec': 'none',
+            } for ext in ['m4a', 'mp3', 'ogg']]
+            self._sort_formats(formats)
+        else:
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/voicerepublic.py
```diff
@@ -0,0 +1,55 @@
...
+    _TEST = {
+        'url': 'https://voicerepublic.com/talks/watching-the-watchers-building-a-sousveillance-state',
+        'md5': '0554a24d1657915aa8e8f84e15dc9353',
+        'info_dict': {
+            'id': '2296',
+            'ext': 'm4a',
+            'title': 'Watching the Watchers: Building a Sousveillance State',
+            'thumbnail': 'https://voicerepublic.com/system/flyer/2296.png',
+            'description': 'md5:715ba964958afa2398df615809cfecb1',
+            'creator': 'M. C. McGrath',
```
dstftw: There is no such field, must be `uploader`.

youtube_dl/extractor/voicerepublic.py
```diff
@@ -0,0 +1,55 @@
...
+        req = compat_urllib_request.Request(url)
+        # Older versions of Firefox get redirected to an "upgrade browser" page
+        req.add_header('User-Agent', 'youtube-dl')
+        webpage = self._download_webpage(req, display_id)
+        thumbnail = self._og_search_thumbnail(webpage)
+        video_id = self._search_regex(r'/(\d+)\.png', thumbnail, 'id')
+
+        if '<div class=\'vr-player jp-jplayer\'' in webpage:
+            formats = [{
+                'url': 'https://voicerepublic.com/vrmedia/{}-clean.{}'.format(video_id, ext),
```
dstftw: Change to `{0}` and `{1}` .

youtube_dl/extractor/voicerepublic.py
```diff
@@ -0,0 +1,55 @@
...
+                'ext': ext,
+                'format_id': ext,
+                'vcodec': 'none',
+            } for ext in ['m4a', 'mp3', 'ogg']]
+            self._sort_formats(formats)
+        else:
+            # Audio is still queued for processing
```
dstftw: You may want to output this meaningful error with `ExtractorError` if this is for sure.

youtube_dl/extractor/voicerepublic.py
```diff
@@ -0,0 +1,55 @@
...
+            formats = []
+
+        return {
+            'id': video_id,
+            'title': self._og_search_title(webpage),
+            'formats': formats,
+            'url': self._og_search_url(webpage),
+            'thumbnail': thumbnail,
+            'description': self._og_search_description(webpage),
+            'creator': self._search_regex(r'<meta content=\'([^\']*?)\' name=\'author\'>', webpage, 'author', fatal=False),
```
dstftw: Use `self._html_search_meta` instead.

duncankl:
Thanks for the feedback, I've added a series of commits to address the comments.

youtube_dl/extractor/voicerepublic.py
```diff
@@ -0,0 +1,55 @@
...
+        video_id = self._search_regex(r'/(\d+)\.png', thumbnail, 'id')
+
+        if '<a>Queued for processing, please stand by...</a>' in webpage:
+            raise ExtractorError('Audio is still queued for processing')
+
+        ext_matches = re.finditer(r'data-\w+=\'/vrmedia/\d+-clean\.(\w+)\'', webpage)
+        exts = [match.group(1) for match in ext_matches]
+
+        formats = [{
+            'url': 'https://voicerepublic.com/vrmedia/{}-clean.{}'.format(video_id, ext),
```
dstftw: You still hardcode the path schema. If it changes to, say, `/vrmedia2/<id>-clean.<ext>` the extraction is broken then. You need to capture the whole path:
```
for ext, path in re.findall(r"data-([^=]+)='([^']+)'", webpage):
    format_url = https://voicerepublic.com + path
    ...
```

duncankl:
Ah yes, I see what you mean. This latest version no longer depends on a hardcoded path schema. I made the regex a bit more strict to prevent it matching other `data-` tags on the page.

dstftw:
Thanks.

duncankl:
Great! Thank you for your help @dstftw.

dstftw:

duncankl:
Actually I'm already in the authors list; I've added one other extractor previously. Thanks for checking though.

PR #5669, opened 2015-05-10
Fallback to `-f best` when merger is outdated

### LINKED ISSUE #5661: Merge issues due to outdated avconv
After switching to `-f bestvideo+bestaudio/best` as default format selection strategy we receive some bugreports caused by outdated avconv (e.g. #5659 #5595) (and probably ffmpeg). It would probably be reasonable to tweak format selection logic a bit, e.g. enable `-f bestvideo+beastaudio/best` only if merger version is new enough (capable of actual merging) and fallback to `-f best` otherwise.

I did some investigations with [original Windows builds](http://builds.libav.org/windows/release-gpl/) and here are the results.

**Test video:** http://www.youtube.com/watch?v=ScMOyURq9os

**Test formats:**

format code | extension | type | codec
--- | --- | --- | ---
278 | webm | video only | vp9
249 | webm | audio only | opus
160 | mp4 | video only | h264
140 | m4a | audio only | aac

**Results:**

video | audio | merge output | avconv 9.18 | avconv 10.6 | avconv 11.3
--- | --- | --- | --- | --- | ---
160 | 140 | mp4 | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:
278 | 249 | webm | :x: | :x: | :heavy_check_mark:
278 | 249 | mkv | :x: | :heavy_check_mark: | :heavy_check_mark:
160 | 249 | mkv | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:
278 | 140 | mkv | :x: | :heavy_check_mark: | :heavy_check_mark:

With libav 0.8.17 it's even [worse](https://github.com/rg3/youtube-dl/issues/5595).

So, we can for example:
* fallback to `-f best` if avconv is older than 10 and if avconv is 10 - force merge output format to mkv if `bestvideo+bestaudio` turns out to be webm+webm.
* fallback to `-f best` if avconv is older than 11.

dstftw:
Fixes #5661.

### DIFF
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -36,7 +36,9 @@ class FFmpegPostProcessor(PostProcessor):
     def check_version(self):
         if not self.available:
             raise FFmpegPostProcessorError('ffmpeg or avconv not found. Please install one.')
+        self.check_outdated()
 
+    def check_outdated(self):
         required_version = '10-0' if self.basename == 'avconv' else '1.0'
         if is_outdated_version(
                 self._versions[self.basename], required_version):
@@ -44,6 +46,8 @@ class FFmpegPostProcessor(PostProcessor):
                 self.basename, self.basename, required_version)
             if self._downloader:
                 self._downloader.report_warning(warning)
+            return True
+        return False
 
     @staticmethod
     def get_versions(downloader=None):
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -36,9 +36,7 @@ class FFmpegPostProcessor(PostProcessor):
     def check_version(self):
         if not self.available:
             raise FFmpegPostProcessorError('ffmpeg or avconv not found. Please install one.')
-        self.check_outdated()
 
-    def check_outdated(self):
         required_version = '10-0' if self.basename == 'avconv' else '1.0'
         if is_outdated_version(
                 self._versions[self.basename], required_version):
@@ -46,8 +44,6 @@ class FFmpegPostProcessor(PostProcessor):
                 self.basename, self.basename, required_version)
             if self._downloader:
                 self._downloader.report_warning(warning)
-            return True
-        return False
 
     @staticmethod
     def get_versions(downloader=None):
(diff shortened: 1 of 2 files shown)


youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -44,6 +46,8 @@ def check_version(self):
                 self.basename, self.basename, required_version)
             if self._downloader:
                 self._downloader.report_warning(warning)
```
jaimeMF: The warning should be printed in `check_version`,  otherwise it's printed when we check it in `YoutubeDL`.

In that case, it may make more sense to rename `check_outdated` to something like `is_updated` or `is_outdated` (you'll have to negate it at least in one place so it's up to you to choose the name).

dstftw: >The warning should be printed in `check_version`,  otherwise it's printed when we check it in `YoutubeDL`.

This is intentional. By this we can tell a user that youtube-dl deviates from the "planned" `bestvideo+bestaudio` strategy and encourage him to update. At least this would be a clue for a user wondering why youtube-dl does not download the best overall quality despite avconv is installed.

There is already `is_outdated_version` so I thought `is_outdated` would be misleading.

jaimeMF: Ok, but we should use a different message, because
> `Your copy of avconv is outdated, update avconv to version 10-0 or newer if you encounter any errors.`

doesn't tell the user that it can't be used for `bestvideo+bestaudio` and that it won't get the best quality.

dstftw:
Added meaningful message, restored `check_version` and added `can_merge` as more convenient place for merge-capable checks.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -591,6 +591,23 @@ def run(self, info):
         os.rename(encodeFilename(temp_filename), encodeFilename(filename))
         return info['__files_to_merge'], info
 
+    def can_merge(self):
+        # TODO: figure out merge-capable ffmpeg version
```
jaimeMF: Probably 2.1 and newer work (I guess that was the version I was using for #2089). Apparently 2.2 [is the one that includes libav 10.6](https://ffmpeg.org/download.html#release_2.2).

dstftw: Well, anyway we have to check some version agains the test set from #5661 to be sure.

PR #5675, opened 2015-05-11
[zingmp3] Set upload_date where available

ping:
Use album "Release Date" as upload_date for for tracks if available.

### DIFF
--- a/youtube_dl/extractor/zingmp3.py
+++ b/youtube_dl/extractor/zingmp3.py
@@ -22,7 +22,7 @@ class ZingMp3BaseInfoExtractor(InfoExtractor):
             'thumbnail': thumbnail,
         }
 
-    def _extract_player_xml(self, player_xml_url, id, playlist_title=None):
+    def _extract_player_xml(self, player_xml_url, id, playlist_title=None, upload_date=None):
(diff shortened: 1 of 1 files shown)


dstftw:
Generally, release date is not related with date media has been uploaded. What is the rationale?

ping:
Actually, I was taking a cue from https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/qqmusic.py#L49 where the release date was used as upload date ("发行时间" in the regex refers to Release date).

If it's not kosher, go ahead and close this. No worries.

yan12125:
If I'm correct, QQ music have cooperation agreements with several major media companies (Warner, Sony, etc.), so songs are uploaded to QQ almost the same time as they are released by the companies. I'm not sure if it is also the case for ZingMp3. If so, it will be great to use the release date as the upload date.

ping:
To be honest, I can't say since I'm not Vietnamese. From what I can tell, the site is very popular in Vietnam but I don't know much beyond that.

yan12125:
I'm closing this PR now. If you find new information for ```upload_date``` or other fields, feel free to open a new pull request.

PR #5676, opened 2015-05-11
Add SouthParkDk extractor

rrooij:
Added an extractor for the Danish South Park site. 

This solves:

https://github.com/rg3/youtube-dl/issues/4605

The imports were also sorted alphabetically, since this wasn't the case previously.


dstftw:
Thanks.

PR #5680, opened 2015-05-12
[qqmusic] Add support for charts / top lists

ping:
Support QQ music charts, for example: http://y.qq.com/#type=toplist&p=top_6 and http://y.qq.com/#type=toplist&p=global_12

### DIFF
--- a/youtube_dl/extractor/qqmusic.py
+++ b/youtube_dl/extractor/qqmusic.py
@@ -188,6 +188,13 @@ class QQMusicToplistIE(QQPlaylistBaseIE):
             'title': 'QQ音乐巅峰榜·欧美',
         },
         'playlist_count': 100,
+    }, {
+        'url': 'http://y.qq.com/#type=toplist&p=global_5',
+        'info_dict': {
+            'id': 'global_5',
+            'title': '韩国mnet排行榜',
+        },
+        'playlist_count': 50,
     }]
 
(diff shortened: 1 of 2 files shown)


yan12125:
The top list http://y.qq.com/#type=toplist&p=global_5 have 50 songs while your extractor gets 40 only. Maybe you need to use the same extraction logic for global_* pages as top_* pages.

ping:
Thanks for the catch. Should be fixed now.

youtube_dl/extractor/qqmusic.py
```diff
@@ -168,3 +169,69 @@ def _real_extract(self, url):
...
+
+        list_page = self._download_webpage(
+            "http://y.qq.com/y/static/toplist/index/%s.html" % list_id, 
+            list_id, 'Download toplist page')
+
+        entries = []
+        jsonp_url = ""
+        if list_type == 'top':
+            jsonp_url = "http://y.qq.com/y/static/toplist/json/top/%s/1.js" % num_id
+        elif list_type == 'global':
```
jaimeMF: since `list_type` can only be `top` or `global` you can change it to `else:` and safely remove `jsonp_url = ""` (which should have been set to `None`).

ping:
Changed.

yan12125:
Thanks. Merged.

PR #5684, opened 2015-05-12
patch for https://github.com/rg3/youtube-dl/issues/5683

GrumpyOldTroll:


jaimeMF:
There a few issues:
1. What happens if all the formats are `m3u8`?, you'll have an empty format list. The user would get confused.
2. You shouldn't be using `FFmpegMergerPP`, instead a new method should be added to the m3u8 downloader.
3. youtube-dl has a builtin m3u8 downloader, so maybe we should use that instead (but it may not work perfectly in all videos)

(Side question: why did you need to modify setup.py and version.py?)

GrumpyOldTroll:
I agree with the points you raised. As I mentioned in rg3#5683, this was more of a quick hack to get me working again than the proper fix you're describing. I considered attempting something better, but chose the easier hack because I didn't want to spend much more time on it. I didn't look too hard, but it wasn't clear to me whether there would be side effects from looking up the downloaders inside select_format.

So I agree the submitted patch is only worth considering if nobody intends to do it properly in the near term, and the only nice thing about it is that it makes at least one site that used to work but stopped working (without ffmpeg) start working again. Also worth noting: I saw several other recent issues with the same error message when searching before logging a new issue, but I wasn't sure if any of those reduced to the same problem or not.

NativeHlsFD is an interesting idea as a fallback plan (maybe even merged into HlsFD in the "not ffpp.available" case). I hadn't noticed it. But it probably has to rewrite the manifest to use relative paths, or many hls videos probably won't work.

Answer to the side question: I'm working on a different package with its own setup.py, and I wanted it to download and install as a dependency a specific version of my fork (one that works for the use case I've worked around here). If I didn't change the name, that installer was getting confused with the youtube_dl package in pypi, and ended up installing a youtube-dl version that didn't work for what I was doing. I imagine there's a better way to do this, but I had trouble discovering it. The internet had some non-examples that claimed it was easy and some examples that didn't work for me. The answer I ended up with was from http://stackoverflow.com/a/3481388, but I also had to rename the egg and the package or the install was failing. There was some trial and error, and I would not be surprised if there's a better way that doesn't require those changes, but my first 8 or so attempts did not work, so I was experimenting.

Thanks for the review.

PR #5686, opened 2015-05-13
Added functionality to retrieve information from YoutubeDL object as

oche-jay:
I have added the functionality to retrieve relevant information from the YoutubeDL object. This information, such as url, filename and formats would normally be written to the console if the options to do so have been set. However, with this patch an application which embeds Youtube-DL could just access an field of YoutubeDL object with that information ( even without the options to print to screen being set). This enhances YoutubeDL as an API (as well as the helpful tool that is is).
e.g:
     print ydl.url, ydl.format, ydl.formats
etc

### DIFF
--- a/README.md
+++ b/README.md
@@ -542,9 +542,14 @@ import youtube_dl
 ydl_opts = {}
 with youtube_dl.YoutubeDL(ydl_opts) as ydl:
     ydl.download(['http://www.youtube.com/watch?v=BaW_jenozKc'])
+	print ydl.url
+	print ydl.format
+	ydl_jsonobject = ydl.info_dict_json
 ```
 
 Most likely, you'll want to use various options. For a list of what can be done, have a look at [youtube_dl/YoutubeDL.py](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L69). For a start, if you want to intercept youtube-dl's output, set a `logger` object.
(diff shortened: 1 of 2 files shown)


yan12125:
Isn't ```extract_info()``` sufficient? For example:
```Python
import youtube_dl
import json

ydl_opts = {'quiet': True}
with youtube_dl.YoutubeDL(ydl_opts) as ydl:
    info_dict = ydl.extract_info('http://www.youtube.com/watch?v=BaW_jenozKc', download=False)
    print(json.dumps(info_dict, indent=4))
```

jaimeMF:
It would make implementing multithread downloading impossible, so it can't be merged in the current state.

As @yan12125 said, `extract_info` should be enough. Maybe we should add a hook that get's called when the extraction is done and gives the info?

oche-jay:
@yan12125 Thanks,I didn't realize extract_info could be used like that, although I did spend a fair bit of time looking for the method that returns the info_dict. Perharps making it a bit more obvious in the README might help.

@jaimeMF the changes are cosmetic so multithreading capabilities shouldn't really be affected.

jaimeMF:
> the changes are cosmetic so multithreading capabilities shouldn't really be affected.

Maybe I wasn't clear, but I wanted to say that using mulithreading with a single YoutubeDL object would be impossible. Because if you were downloading two videos at the same time, the attributes would be from only one of them.

oche-jay:
@jaimeMF I see what you mean, thanks.

PR #5692, opened 2015-05-13
Use thumbnails downloaded by YoutubeDL in EmbedThumbnailPP

yan12125:
Current ```EmbedThumbnailPP``` first downloads the necessary thumbnail, and then merges it into the audio file. Thus in the following invocation:
```
youtube-dl --verbose --embed-thumbnail --no-overwrite --write-all-thumbnails test:xuite
```
Thumbnails are downloaded twice. One in ```YoutubeDL``` and another in ```EmbedThumbnailPP```. I try to fix the scenario in this PR.

There's a deeper advantage from this fix. I believe all resources needed by postprocessors should be downloaded by ```YoutubeDL```, and my implementation conforms to the principle. I think this principle is helpful for solving the problem mentioned by @dstftw in https://github.com/rg3/youtube-dl/commit/1c7e2e64f6328024711d5fa999d4498396f4cb5c#commitcomment-11157068

There are 2 potential issues in my implementation:
1. It's always the highest quality thumbnail being embedded. Users may want to embed a lower quality image, but I think it's rare.
2. If a user invoke youtube-dl consecutively with the 2 following commands:
```
youtube-dl --verbose --write-thumbnail test:xuite
```
```
youtube-dl --verbose --embed-thumbnail test:xuite
```
A thumbnail is downloaded in the first command but deleted in the second command. This issue can be solved by adding a 'already_exists' field for each thumbnail of ```info_dict['thumbnails']```, but I think such checking makes the code too complicated.

cc: @pulpe Most major work in embedthumbnail.py (and previously atomicparsley.py) are done by you, and I'd like to hear some comments from you.

### DIFF
--- a/youtube_dl/postprocessor/embedthumbnail.py
+++ b/youtube_dl/postprocessor/embedthumbnail.py
@@ -7,11 +7,7 @@ import subprocess
 
 from .ffmpeg import FFmpegPostProcessor
 
-from ..compat import (
-    compat_urlretrieve,
-)
 from ..utils import (
-    determine_ext,
     check_executable,
     encodeFilename,
     PostProcessingError,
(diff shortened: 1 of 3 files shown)


youtube_dl/postprocessor/embedthumbnail.py
```diff
@@ -52,7 +52,7 @@ def run(self, info):
             if not check_executable('AtomicParsley', ['-v']):
                 raise EmbedThumbnailPPError('AtomicParsley was not found. Please install.')
 
-            cmd = ['AtomicParsley', filename, '--artwork', temp_thumbnail, '-o', temp_filename]
+            cmd = ['AtomicParsley', filename, '--artwork', thumbnail_filename, '-o', temp_filename]
```
dstftw: Filenames should be `encodeFilename`'ed, otherwise non-ASCII will make subprocess fail on python2 @ Windows.

yan12125: Thanks. Fixed

pulpe:
I like it. Looks much better than before...

dstftw:
LGTM.

yan12125:
Alright merged.

PR #5712, opened 2015-05-15
Sportbox fix

maddoger:
Sportbox.ru extractor was fixed. HD videos support was added.

### DIFF
--- a/youtube_dl/extractor/sportbox.py
+++ b/youtube_dl/extractor/sportbox.py
@@ -11,7 +11,7 @@ from ..utils import (
 
 
 class SportBoxIE(InfoExtractor):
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/sportbox.py
```diff
@@ -11,10 +11,10 @@
 
 
 class SportBoxIE(InfoExtractor):
-    _VALID_URL = r'https?://news\.sportbox\.ru/Vidy_sporta/(?:[^/]+/)+spbvideo_NI\d+_(?P<display_id>.+)'
```
dstftw: Can you add a test for an URL that does not contain `Vidy_sporta/`?

maddoger: Done

dstftw:
Thanks.

PR #5715, opened 2015-05-15
[voxnow] Branch out from rtlnow to fix it (by moving to HLS)

akirk:
Vox Now stopped working some time ago, this fixes it by doing the download via HLS (by re-writing the HDS URLs to HLS).

Maybe this fix also applies to other services served by the rtlnow downloader but I didn't test that.


akirk:
Closed in favor of #5818

PR #5723, opened 2015-05-16
[viki] Add support for shows

ping:
Add support for shows, example: http://www.viki.com/tv/50c-boys-over-flowers

### DIFF
--- a/youtube_dl/extractor/viki.py
+++ b/youtube_dl/extractor/viki.py
@@ -145,3 +145,36 @@ class VikiIE(InfoExtractor):
                 'ext': 'vtt',
             }]
         return res
+
+
+class VikiShowIE(InfoExtractor):
+    IE_NAME = 'viki:show'
+    _VALID_URL = r'^https?://(?:www\.)?viki\.com/tv/(?P<id>[0-9]+c)'
+    _TESTS = [{
+        'url': 'http://www.viki.com/tv/50c-boys-over-flowers',
+        'info_dict': {
+            'id': '50c',
+            'title': 'Boys Over Flowers',
+            'description': 'md5:ecd3cff47967fe193cff37c0bec52790',
+        },
+        'playlist_count': 25,
+    }]
+
+    def _real_extract(self, url):
+        show_id = self._match_id(url)
+        show_page = self._download_webpage(url, show_id, 'Download show page')
+
+        title = self._og_search_title(show_page)
+        description = self._og_search_description(show_page)
+
+        show_json = self._download_json(
+            'http://api.viki.io/v4/containers/%s/episodes.json?app=100000a&per_page=999&sort=number&direction=asc' % show_id,
+            show_id, note='Retrieve show json', errnote='Unable to get show json'
+        )
+        entries = []
+        for video in show_json['response']:
+            video_id = video['id']
+            entries.append(self.url_result(
+                'http://www.viki.com/videos/%s' % video_id, 'Viki', video_id))
+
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/viki.py
```diff
@@ -145,3 +145,35 @@ def _get_subtitles(self, video_id, info_webpage):
...
+
+    def _real_extract(self, url):
+        show_id = self._match_id(url)
+        show_page = self._download_webpage(url, show_id, 'Download show page')
+
+        title = self._og_search_title(show_page)
+        description = self._og_search_description(show_page)
+
+        show_json = self._download_json(
+            'http://api.viki.io/v4/containers/%s/episodes.json?app=100000a&per_page=999&sort=number&direction=asc' % show_id,
```
dstftw: Maximum `per_page` is 25 according [documentation](http://dev.viki.com/v4/overview/#paging). However, I've managed to fetch 50 (for test http://www.viki.com/tv/1354c-poor-nastya-complete). You'll need to implement proper paging approach.

dstftw:
Sometimes there are no episodes, but clips (http://www.viki.com/tv/2500c). May probably reasonable to download them in this case. Or even always download both if present.

ping:
Thanks for the review. I've gone with using 25 per_page, since it's the official number.

dstftw:
Can you make it as variable, e.g. `_PER_PAGE`?

youtube_dl/extractor/viki.py
```diff
@@ -145,3 +145,46 @@ def _get_subtitles(self, video_id, info_webpage):
                 'ext': 'vtt',
             }]
         return res
+
+
+class VikiShowIE(InfoExtractor):
+    IE_NAME = 'viki:show'
```
dstftw: I think this should be renamed to channel or container (for what `c` stands for).

ping:
By _PER_PAGE as a variable, do you mean just as a regular class variable? Or is there some other mechanism that I can hook into?

```
class VikiShowIE(InfoExtractor):
    _PER_PAGE = 25
    # -- snip --

    def _real_extract(self, url):
        # -- snip --
        for video_type in ['episodes', 'clips']:
            json_url = 'http://api.viki.io/v4/containers/%s/%s.json?app=100000a&per_page=%d&sort=number&direction=asc&with_paging=true&page=1' % (show_id, video_type, self._PER_PAGE)
```


 ``show_json['pagination']['next'] `` is a fully formed url, so there's no need to build the url with an incremental ``page`` number

youtube_dl/extractor/viki.py
ping: Will change this to channel

dstftw:
Just a variable would be ok. There are some mechanisms if you still want:
https://github.com/rg3/youtube-dl/blob/master/youtube_dl/utils.py#L1402-L1471
https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/vimeo.py#L648-L665
https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/zdf.py#L124-L153

> ``show_json['pagination']['next'] `` is a fully formed url, so there's no need to build the url with an incremental ``page`` number

Sorry, missed that.

ping:
Changed.

dstftw:
Thanks. Do you want to be credited as one of the authors of youtube-dl? If so, under what name?

ping:
That would be nice. Under "Ping O."

PR #5758, opened 2015-05-20
[dramafever] Add new extractor for dramafever.com

ping:
DramaFever is fairly popular video streaming service, with a large inventory of Asian dramas and movies.


dstftw:
Thanks.

PR #5772, opened 2015-05-21
[TNAFlix] Allow dot (and more) in cat_id and display_id

frenchy1983:
URLs with dots were raising a "UnsupportedError: Unsupported URL" error.

### DIFF
--- a/youtube_dl/extractor/tnaflix.py
+++ b/youtube_dl/extractor/tnaflix.py
@@ -16,20 +16,26 @@ class TNAFlixIE(InfoExtractor):
     _DESCRIPTION_REGEX = r'<h3 itemprop="description">([^<]+)</h3>'
     _CONFIG_REGEX = r'flashvars\.config\s*=\s*escape\("([^"]+)"'
 
-    _TEST = {
-        'url': 'https://www.tnaflix.com/amateur-porn/bunzHD-Ms.Donk/video358632',
-        'md5': '6c431ea56756497e227fb3f01a687869',
-        'info_dict': {
-            'id': '358632',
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/tnaflix.py
```diff
@@ -10,23 +10,23 @@
...
 class TNAFlixIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?tnaflix\.com/(?P<cat_id>[\w-]+)/(?P<display_id>[\w-]+)/video(?P<id>\d+)'
+    _VALID_URL = r'https?://(?:www\.)?tnaflix\.com/(?P<cat_id>[^/]+)/(?P<display_id>[^/]+)/video(?P<id>\d+)'
 
     _TITLE_REGEX = r'<title>(.+?) - TNAFlix Porn Videos</title>'
     _DESCRIPTION_REGEX = r'<h3 itemprop="description">([^<]+)</h3>'
     _CONFIG_REGEX = r'flashvars\.config\s*=\s*escape\("([^"]+)"'
 
     _TEST = {
-        'url': 'http://www.tnaflix.com/porn-stars/Carmella-Decesare-striptease/video553878',
```
dstftw: You should preserve the old test as well. If there are no differences in extraction process apart from dotted URL new test should be set [`only_matching`](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/xhamster.py#L43-L46).
You may probably want to patch EMPFlixIE counterpart the same way.

frenchy1983: Done :)

dstftw:
Thanks.

Issue #1574, opened 2013-10-07
Add support for yle


nemobis:
https://github.com/aajanki/yle-dl by @aajanki may provide what needed

$ ./youtube-dl --verbose http://areena.yle.fi/radio/2048905
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['--verbose', 'http://areena.yle.fi/radio/2048905']
[debug] youtube-dl version 2013.10.07
[debug] Python version 2.6.6 - Linux-3.2.46-grbfs-kapsi-x86_64-with-debian-6.0.7
[debug] Proxy map: {}
WARNING: Falling back on generic information extractor.
[generic] 2048905: Downloading webpage
[generic] 2048905: Extracting information
ERROR: Unsupported URL: http://areena.yle.fi/radio/2048905; please report this issue on https://yt-dl.org/bug . Be sure to call youtube-dl with the --verbose flag and include its complete output. Make sure you are using the latest version; type  youtube-dl -U  to update.
Traceback (most recent call last):
  File "./youtube-dl/youtube_dl/YoutubeDL.py", line 348, in extract_info
    ie_result = ie.extract(url)
  File "./youtube-dl/youtube_dl/extractor/common.py", line 117, in extract
    return self._real_extract(url)
  File "./youtube-dl/youtube_dl/extractor/generic.py", line 152, in _real_extract
    raise ExtractorError(u'Unsupported URL: %s' % url)
ExtractorError: Unsupported URL: http://areena.yle.fi/radio/2048905; please report this issue on https://yt-dl.org/bug . Be sure to call youtube-dl with the --verbose flag and include its complete output. Make sure you are using the latest version; type  youtube-dl -U  to update.

yan12125:
Another form of URL: http://arenan.yle.fi/1-3094091 (from #8896)

rautamiekka:
https://areena.yle.fi/1-4200978 is a link which returns `Unsupported URL`.

Hegezcc:
I am working with this on pull request #20274.

Zuccace:
Great! I think I could test it already. :)

Hegezcc:
It works on my computer, just clone the branch and run it with `python -m youtube_dl https://areena.yle.fi/1-4256816`. 

[...]

PR #5805, opened 2015-05-24
Add support for  YLE (Closes: #1574)

hlintala:
Adds extractors for YLE Areena, YLE Areena Live TV streams, and YLE Elävä Arkisto.

### DIFF
--- a/youtube_dl/aes.py
+++ b/youtube_dl/aes.py
@@ -88,6 +88,32 @@ def aes_cbc_encrypt(data, key, iv):
     return encrypted_data
 
 
+def aes_cfb_decrypt(data, key, iv):
+    """
+    Decrypt with aes in ciphed feedback mode
+
+    @param {int[]} data        cipher
+    @param {int[]} key         16/24/32-Byte cipher key
+    @param {int[]} iv          16-Byte IV
+    @returns {int[]}           decrypted data
+    """
+    expanded_key = key_expansion(key)
+    block_count = int(ceil(float(len(data)) / BLOCK_SIZE_BYTES))
+
+    decrypted_data = []
+    previous_cipher_block = iv
+    for i in range(block_count):
+        block = data[i * BLOCK_SIZE_BYTES: (i + 1) * BLOCK_SIZE_BYTES]
+        block += [0] * (BLOCK_SIZE_BYTES - len(block))
+
+        encrypted_vector = aes_encrypt(previous_cipher_block, expanded_key)
+        decrypted_data += xor(block, encrypted_vector)
+        previous_cipher_block = block
+    decrypted_data = decrypted_data[:len(data)]
+
+    return decrypted_data
+
+
 def key_expansion(data):
     """
     Generate key schedule
--- a/youtube_dl/extractor/yle.py
+++ b/youtube_dl/extractor/yle.py
@@ -15,6 +15,8 @@ from ..utils import (
     intlist_to_bytes,
     parse_duration,
     parse_iso8601,
+    strip_jsonp,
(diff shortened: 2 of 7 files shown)


youtube_dl/aes.py
```diff
@@ -60,6 +60,32 @@ def aes_cbc_decrypt(data, key, iv):
     return decrypted_data
 
 
+def aes_cfb_decrypt(data, key, iv):
```
jaimeMF: Could you add a test into [test/test_aes.py](https://github.com/rg3/youtube-dl/blob/master/test/test_aes.py)? So that we  detect breakage better.

(Also if possible add a way to generate the encrypted data with openssl at [devscripts/generate_aes_testdata.py](https://github.com/rg3/youtube-dl/blob/master/devscripts/generate_aes_testdata.py)

hlintala: Done

youtube_dl/extractor/yle.py
```diff
@@ -0,0 +1,278 @@
...
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        media_id = mobj.group('id')
+
+        lang = 'fi'
+        if mobj.group('host') == 'arenan':
+            lang = 'sv'
+
+        playlist = self._get_playlist(url, media_id)
```
dstftw: Is it really required for a single video URL? All tests you provided end up with empty `data` array and construct `playlist` it from [the original `url`](https://github.com/hlintala/youtube-dl/blob/yle/youtube_dl/extractor/yle.py#L109).

youtube_dl/extractor/yle.py
```diff
@@ -0,0 +1,278 @@
...
+class YLEAreenaLiveTVIE(YLEAreenaIE):
+    _VALID_URL = r'^http://(?P<host>areena|arenan)\.yle\.fi/tv/(?:suorat|direkt)/(?P<id>[\w-]+).*'
+    _PROGRAM_URL = 'http://player.yle.fi/api/v1/services.jsonp?id={pid}&region=fi'
+    _PROGRAM_KEY = 'service'
+    _EVENT_KEY = 'outlet'
+
+    _TEST = {
+        'url': 'http://areena.yle.fi/tv/suorat/yle-tv1',
+        'info_dict': {
+            'title': 'Yle TV1',
```
dstftw: For lives you should build `title` with `self._live_title` and set `is_live` to `True`.

youtube_dl/extractor/yle.py
hlintala: Added test case, which returns list of urls (non-empty `data` array). Also `self._live_title` is now used and `is_live` is set `True` for lives.

hlintala:
Added test case, which returns list of urls (non-empty `data` array). Also `self._live_title` is now used and `is_live` is set `True` for live streams.

yan12125:

rytilahti:
What's the state of this? What else is required besides rebasing to the current head? 

Before finding this PR I was looking a bit into implementing it, and spotted that some information is directly available from OpenGraph properties and `<meta itemprop>` elements. Would it make sense to make use of those? Among properties not currently supported there exists also `alt_title` (for original title) and `thumbnail`.

zenksen:
I'm not a developer, but it seems to me this PR only implements the older (flash player only) HDS/HLS support, but some of the more recent items on Areena have a HTML5 player via Kaltura. Support for that probably needs to be implemented as well.

Zuccace:
Has there been any work towards solvin this issue since last post?

PR #5807, opened 2015-05-24
[pinkbike] new extractor

misterhat:
Solves https://github.com/rg3/youtube-dl/issues/5791.


youtube_dl/extractor/pinkbike.py
```diff
@@ -0,0 +1,78 @@
...
+            r'<source data-quality=\\"([0-9]+)p\\" src=\\"(.*?)\\">',
+            webpage)
+
+        formats = [{'url': fmt[1], 'height': fmt[0]} for fmt in formats]
+
+        return {
+            'id': video_id,
+            'title': title,
+            'description': description,
+            'duration': self._html_search_meta('video:duration', webpage, 'duration'),
```
jaimeMF: this field must be a number.

PR #5828, opened 2015-05-27
[youku] update youku

PeterDing:
[update youku]
Getting video_urls use m3u8 link, this method is from https://github.com/soimort/you-get/blob/develop/src/you_get/extractors/youku.py#L22

### DIFF
--- a/youtube_dl/extractor/youku.py
+++ b/youtube_dl/extractor/youku.py
@@ -1,123 +1,172 @@
 # coding: utf-8
-
 from __future__ import unicode_literals
 
-import math
-import random
 import re
 import time
+import base64
 
 from .common import InfoExtractor
-from ..utils import (
-    ExtractorError,
-)
+from ..utils import ExtractorError
 
+from ..compat import compat_urllib_parse
 
 class YoukuIE(InfoExtractor):
+    IE_NAME = 'youku'
     _VALID_URL = r'''(?x)
         (?:
             http://(?:v|player)\.youku\.com/(?:v_show/id_|player\.php/sid/)|
             youku:)
         (?P<id>[A-Za-z0-9]+)(?:\.html|/v\.swf|)
     '''
+
     _TEST = {
-        'url': 'http://v.youku.com/v_show/id_XNDgyMDQ2NTQw.html',
-        'md5': 'ffe3f2e435663dc2d1eea34faeff5b5b',
-        'params': {
-            'test': False
-        },
-        'info_dict': {
-            'id': 'XNDgyMDQ2NTQw_part00',
-            'ext': 'flv',
-            'title': 'youtube-dl test video "\'/\\ä↭𝕐'
-        }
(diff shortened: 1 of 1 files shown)


yan12125:
you-get is licensed under MIT while youtube-dl is UNLICENSE. We can't (and shouldn't) use their codes directly unless explicit permission. Did you ask the original authors?

youtube_dl/extractor/youku.py
```diff
@@ -1,123 +1,172 @@
...
+                q += 1
+
+            return result
+
+        e_code = trans_e(f_code_1, base64.b64decode(bytes(ep, 'ascii')))
+        sid, token = e_code.split('_')
+        new_ep = trans_e(f_code_2, '%s_%s_%s' % (sid, vid, token))
+        return base64.b64encode(bytes(new_ep, 'latin')), sid, token
+
+    def parse_m3u8(self, cm):
```
yan12125: I think [```_extract_m3u8_formats```](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L847) does the work better. If the existing method does not fit the need, feel free to modify it.

PeterDing:
OK, let me try to remodify it. And I will connect authors of you-get to apply for the permission.
But, I wonder if the authors whose codes is under MIT does not authorize me to use the code, how can I do.

yan12125:
Then you need to write your own codes from scratch.

youtube_dl/extractor/youku.py
```diff
@@ -1,123 +1,172 @@
...
         format = self._downloader.params.get('format', None)
-        supported_format = list(config['data'][0]['streamfileids'].keys())
-
-        # TODO proper format selection
-        if format is None or format == 'best':
-            if 'hd2' in supported_format:
-                format = 'hd2'
+        supported_format = data['streamtypes']
+
+        # DONE proper format selection
```
jaimeMF: Remove this comment, the only purpose of the `TODO` comment was to remind that it had to be implemented.

youtube_dl/extractor/youku.py
```diff
@@ -1,123 +1,172 @@
...
+            "vid": video_id
+        }
+        m3u8_url = 'http://pl.youku.com/playlist/m3u8?' \
+            + compat_urllib_parse.urlencode(m3u8_url_params)
+        cm = self._download_webpage(m3u8_url, video_id, 'M3U8 DOWNLOAD')
+        video_urls = self.parse_m3u8(cm)
+
+        # construct info
+        entries = []
+        for i in range(len(video_urls)):
```
jaimeMF:
```
for i, video_url in enumerate(video_urls):
```

youtube_dl/extractor/youku.py
jaimeMF: Sorry, I should have commented in #5833.

PeterDing:
I rewrite youku.py.
Getting video_url from api.
https://github.com/rg3/youtube-dl/pull/5833/files

PR #5829, opened 2015-05-27
[porn91] Add new extractor for 91porn.com

PeterDing:
An extractor for 91porn.com -- A popular Chinese amateur porn site.

[WARN] 91porn.com has a limit for downloading, one ip could download 10 videos among 24h.


yan12125:
Seems you forgot to commit porn91.py? And don't mix irrelevant commits in a single pull request.

PeterDing:
Thank you, I try again.

PR #5830, opened 2015-05-27
[porn91] Add new extractor for 91porn.com

PeterDing:
An extractor for 91porn.com -- A popular Chinese amateur porn site.

[WARN] 91porn.com has a limit for downloading, one ip could download 10 videos among 24h.

-----

This one will be good.

### DIFF
--- a/youtube_dl/extractor/porn91.py
+++ b/youtube_dl/extractor/porn91.py
@@ -2,7 +2,6 @@
 from __future__ import unicode_literals
 
 import re
-import json
 
 from ..compat import compat_urllib_parse
 from .common import InfoExtractor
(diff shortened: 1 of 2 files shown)


yan12125:
FYI: [```_search_regex```](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L526)/[```_html_search_regex```](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L560) is more robust than ```re.search```. Please consider using these helper functions.

youtube_dl/extractor/porn91.py
```diff
@@ -0,0 +1,63 @@
...
+        title = title.replace('\n', '')
+
+        # get real url
+        n1 = self._search_regex(
+            r'so.addVariable\(\'file\',\'(?P<n1>\d+)\'', webpage, 'n1')
+        n2 = self._search_regex(
+            r'so.addVariable\(\'seccode\',\'(?P<n2>.+?)\'', webpage, 'n2')
+        n3 = self._search_regex(
+            r'so.addVariable\(\'max_vid\',\'(?P<n3>\d+)\'', webpage, 'n3')
+        if not (n1 and n2 and n3):
```
yan12125: ```_search_regex``` raises a ```RegexNotFoundError``` if the given pattern is not found. As a result, this checking does not work.

youtube_dl/extractor/porn91.py
```diff
@@ -0,0 +1,63 @@
...
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('id')
+        url = 'http://91porn.com/view_video.php?viewkey=%s' % video_id
+        self._set_cookie('91porn.com', 'language', 'cn_CN')
+        webpage = self._download_webpage(url, video_id, "get HTML content")
+        title = self._search_regex(
+            r'<div id="viewvideo-title">(?P<title>.+?)</div>',
+            webpage, 'title', flags=re.DOTALL)
+        assert title
```
yan12125: It's not necessary because ```_search_regex``` already raises exceptions in failure.

yan12125:
Thanks for the extractor. Merged with some modifications.

yan12125:
Would you like to be listed in [AUTHORS](https://github.com/rg3/youtube-dl/blob/master/AUTHORS)? If so, under what name?

PeterDing:
@yan12125 It is honor, you can call me Peter Ding.

yan12125:
Thanks. Done in 4053ee9104cd7669f749a267dc2c2a1725ca188b

PR #5833, opened 2015-05-28
[youku] update youku

PeterDing:
Now, the old script gets invalid video_urls, because youku changed its algorithm.  
The commit updates algorithm, generating valid video_urls.  

------

Getting video_url directly from api and algorithm, not from m3u8 file which has incomplete video_url.

### DIFF
--- a/youtube_dl/extractor/youku.py
+++ b/youtube_dl/extractor/youku.py
@@ -1,13 +1,15 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
-import sys
-pyvs = sys.version_info[0]
 import re
 import base64
 
 from .common import InfoExtractor
 from ..utils import ExtractorError
+
+from ..compat import compat_urllib_parse
+
+bytes_is_str = (bytes == str)  # for compatible
 
 class YoukuIE(InfoExtractor):
     IE_NAME = 'youku'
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/youku.py
```diff
@@ -1,123 +1,215 @@
 # coding: utf-8
-
 from __future__ import unicode_literals
 
-import math
-import random
+import sys
+pyvs = sys.version_info[0]
```
yan12125: It's better to put all Python 2/3 compatibility codes to [compat.py](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/compat.py)

youtube_dl/extractor/youku.py
```diff
@@ -1,123 +1,215 @@
...
+            ep = ep.replace('=', '%2D')
+            return ep
+
+        # generate video_urls
+        video_urls_dict = {}
+        for format in data1['streamtypes']:
+            video_urls = []
+            for dt in data1['segs'][format]:
+                n = str(int(dt['no']))
+                video_url = \
```
yan12125: It's better to use ```compat_urllib_parse.urlencode``` in this line. One of the benefits is that there is no need to escape ep in ```generate_ep()``` anymore.

yan12125:
Thanks @PeterDing. Merged with a few modifications.

PR #5849, opened 2015-05-30
[iqiyi] Add new extractor for iqiyi.com

PeterDing:
I use the method that @jackyzy823 found at [here](https://github.com/jackyzy823/you-get/blob/iqiyi-key-again/src/you_get/extractors/iqiyi.py) to generate iqiyi video_url.
Also, I got this [permission](https://github.com/soimort/you-get/pull/523) to use his algorithm to code.
If there is a copyright problem, let me know.
@yan12125

### DIFF
--- a/youtube_dl/extractor/iqiyi.py
+++ b/youtube_dl/extractor/iqiyi.py
@@ -1,0 +1,214 @@
+# coding: utf-8
+
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+from ..compat import (
+    compat_chr,
+    compat_parse_qs,
+    compat_urllib_parse,
+    compat_urllib_request,
+    compat_urlparse,
+    compat_str,
+)
+
+from ..utils import ExtractorError
+
+import re
+import time
+import json
+import uuid
+import math
+import random
+import zlib
+import hashlib
+
+class IqiyiIE(InfoExtractor):
+    IE_NAME = 'iqiyi'
+
+    _VALID_URL = r'http://(?:www\.)iqiyi.com/.+?\.html'
+
+    _TEST = {
+            'url': 'http://www.iqiyi.com/v_19rrojlavg.html',
+            'md5': '260f0f59686e65e886995d0ba791ab83',
+            'info_dict': {
+                'id': '9c1fb1b99d192b21c559e5a1a2cb3c73',
+                'title': '美国德州空中惊现奇异云团 酷似UFO',
+                'ext': 'f4v'
+            }
+    }
+
+    def construct_video_urls(self, data, video_id, _uuid):
+        def do_xor(x, y):
+            a = y % 3
+            if a == 1:
+                return x ^ 121
+            if a == 2:
+                return x ^ 72
+            return x ^ 103
+
+        def get_encode_code(l):
+            a = 0
+            b = l.split('-')
+            c = len(b)
+            s = ''
+            for i in range(c - 1, -1, -1):
+                a = do_xor(int(b[c-i-1], 16), i)
+                s += chr(a)
+            return s[::-1]
+
+        def get_path_key(x):
+            mg = ')(*&^flash@#$%a'
+            tm = self._download_json(
+                'http://data.video.qiyi.com/t?tn=' + str(random.random()), video_id)['t']
+            t = str(int(math.floor(int(tm)/(600.0))))
+            return hashlib.md5(
+                 (t+mg+x).encode('utf8')).hexdigest()
+
+        video_urls_dict = {}
+        for i in data['vp']['tkl'][0]['vs']:
+            if 0 < int(i['bid']) <= 10:
+                format_id = self.get_format(i['bid'])
+
+            video_urls_info = i['fs']
+            if not i['fs'][0]['l'].startswith('/'):
+                t = get_encode_code(i['fs'][0]['l'])
+                if t.endswith('mp4'):
+                    video_urls_info = i['flvs']
+
+            video_urls = []
+            for ii in video_urls_info:
+                vl = ii['l']
+                if not vl.startswith('/'):
+                    vl = get_encode_code(vl)
+                key = get_path_key(
+                    vl.split('/')[-1].split('.')[0])
+                filesize = ii['b']
+                base_url = data['vp']['du'].split('/')
+                base_url.insert(-1, key)
+                base_url = '/'.join(base_url)
+                param = {
+                    'su': _uuid,
+                    'qyid': uuid.uuid4().hex,
+                    'client': '',
+                    'z': '',
+                    'bt': '',
+                    'ct': '',
+                    'tn': str(int(time.time()))
+                }
+                api_video_url = base_url + vl + '?' + \
+                    compat_urllib_parse.urlencode(param)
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/iqiyi.py
```diff
@@ -0,0 +1,242 @@
...
+        cn = zlib.decompress(cn[8:])
+        pt = re.compile(b'MixerRemote\x08(?P<enc_key>.+?)\$&vv')
+        enc_key = self._search_regex(pt, cn, 'enc_key').decode('utf8')
+        return enc_key
+
+    def _real_extract(self, url):
+        webpage = self._download_webpage(
+            url, 'temp_id', note='download video page')
+        tvid = self._search_regex(
+            r'tvId ?= ?(\'|\")(?P<tvid>\d+)', webpage, 'tvid', flags=re.I, group='tvid')
```
yan12125: Is it necessary to use ```re.I``` here? The test case uses "tvid" (all lowercase). Do you have an example that uses "tvId"?

youtube_dl/extractor/iqiyi.py
```diff
@@ -0,0 +1,242 @@
...
+            r'videoId ?= ?(\'|\")(?P<video_id>[a-z\d]+)',
+            webpage, 'video_id', flags=re.I, group='video_id')
+        swf_url = self._search_regex(
+            r'(?P<swf>http://.+?MainPlayer.+?\.swf)', webpage, 'swf')
+        _uuid = uuid.uuid4().hex
+
+        enc_key = self.get_enc_key(swf_url, video_id)
+
+        raw_data = self.get_raw_data(tvid, video_id, enc_key, _uuid)
+        assert raw_data['code'] == 'A000000'
```
yan12125: What's the error indicated by ```raw_data['code'] != 'A000000'```? It's better to give an informational message.

youtube_dl/extractor/iqiyi.py
```diff
@@ -0,0 +1,242 @@
...
+        if not raw_data['data']['vp']['tkl']:
+            raise ExtractorError('No support iQiqy VIP video')
+
+        data = raw_data['data']
+
+        title = data['vi']['vn']
+
+        format = self._downloader.params.get('format', None)
```
yan12125: Extractors should extract information of all possible formats, not the specified format only. Instead, ```YoutubeDL``` class handles format selection.

youtube_dl/extractor/iqiyi.py
PeterDing: http://www.iqiyi.com/v_19rrhnnclk.html
```javascript
33:  Q.PageInfo.playPageInfo = {
34:    albumId: 167419,
35:    tvId: 265686500,
36:    cid: 4,
```

youtube_dl/extractor/iqiyi.py
```diff
@@ -0,0 +1,242 @@
...
+
+            video_urls_info = i['fs']
+            if not i['fs'][0]['l'].startswith('/'):
+                t = get_encode_code(i['fs'][0]['l'])
+                if t.endswith('mp4'):
+                    video_urls_info = i['flvs']
+
+            if int(i['bid']) != int(bid):  # ignore missing match format
+                video_urls.extend(
+                    [('http://example.com/v.flv', ii['b']) for ii in video_urls_info])
```
yan12125: What's the purpose of http://example.com/v.flv here? It always gives a 404 error and I think it's unrelated to iQiyi

youtube_dl/extractor/iqiyi.py
yan12125: There's a line in the this page:
```
data-player-tvid="265686500"
```
As a result, I think ```r'tvid ?= ?(\'|\")(?P<tvid>\d+)'``` without ```re.I``` is enough. (And I would recomment ```r'data-player-tvid\s*=\s*[\'\"](?P<tvid>\d+)'``` for accuracy.) Do you have a page with ```tvId: (tvid)``` only and no ```data-player-tvid="(tvid)"``` line?

youtube_dl/extractor/iqiyi.py
PeterDing: iqiyi sets an expiring timeout (10min) for each generated video link.
And It spends seconds to request information for one video format. If all formats video links be requested, it may be taken minutes.
Usually, people just need one video format, so I let others format video urls to be 'http://example.com/v.flv' and sizes to be original for YoutubeDL.

youtube_dl/extractor/iqiyi.py
PeterDing: No, I do not find. 'data-player-tvid' is fine.

youtube_dl/extractor/iqiyi.py
PeterDing: I has no idea for ```raw_data['code'] != 'A000000'```

youtube_dl/extractor/iqiyi.py
yan12125: I see the rationale. However, this approach makes the following command fails:
```
youtube-dl --all-formats --verbose "http://www.iqiyi.com/v_19rrojlavg.html"
```
It's indeed a problem if the video URL expires too fast. I'll examine the situation.

yan12125: I've tested a generated link and it's still valid after 2.5 hours:
```
$ youtube-dl --get-url test:iqiyi
http://netcncoversea.inter.iqiyi.com/videos/v0/20150528/85/88/721ec4cc33c0ba0a1346c9908859b2fe.f4v?key=099dbd891d07e1f4894d646b70b40cc31&src=iqiyi.com&client=&su=a2ff7a1614f1417c91d9051cc28a03b4&ct=&bt=&qyid=c4bef4174a6446ec841471393a573b3a&z=&tn=1433347310&uuid=8c70f791-556f2524-39
```
(This command is run at Thu 04 Jun 2015 12:01:47 AM CST GMT+8:00)
So I think it's OK to fetch all formats.

It's a good idea to fetch only specified formats, but your approach is not robust enough. Format selection of youtube-dl is a complicated process (see [Format selection](https://github.com/rg3/youtube-dl#format-selection) ). [Sohu extractor](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/sohu.py) suffers from the same problem, too. Before a really robust approach emerges, individual extractors should extract all possible information.

PeterDing: It works.
If there is no expiration, extracting all formats will make sense. But the cost is more time to wait to fetch video_urls.

yan12125:
Thanks for the work. I've merged with several modifications. The most important one is that I remove format selection codes (in 7012620e2b9355d25ddfc855fc5990af938f04d8). If you have a better approach to reduce network traffic, feel free to open a new PR or a new issue for discussion.

dstftw:
I guess we should also ask @jackyzy823 if he/she wants to be credited as the original author.

yan12125:
It's a good idea. Asked in https://github.com/soimort/you-get/pull/523#issuecomment-111564183. Listed in iqiyi.py or AUTHORS?

dstftw:
There is only one place we put credits into - AUTHORS.

PR #5852, opened 2015-05-30
[youtube] Construct a playlist URL in case the page is missing one

ivan:
This fixes jumping from user/channel -> playlist for some users like https://www.youtube.com/user/BitcoinFoundation which lack the HTML that https://github.com/rg3/youtube-dl/commit/386bdfa698a7f06c43df91913677db3732e29900#diff-bd8242a0122c5207531954b67e6e51f0R1416 is looking for.

Playlist URLs are easy to construct, so I don't think we don't have the scan the page for one.

This also removes the superfluous log message "add --no-playlist to just download video VIDEOID" when downloading a user/channel.

cc @dstftw


dstftw:
Thanks.

PR #5859, opened 2015-05-31
add --part-suffix option

ikn:
I want to be able to change '.part' to something else in temporary files so that I can distinguish them.  My file manager uses shared-mime-info to associate MIME types with applications, and shared-mime-info mostly uses filenames.  Lots of applications create '.part' files, but youtube-dl specifically creates '.part' files which are videos, and I only want to associate those ones with my video player.

I wasn't sure if I was supposed to update README.md - I tried, but 'make README.md' seems to think it's already up-to-date.

### DIFF
--- a/test/test_filedownloader.py
+++ b/test/test_filedownloader.py
@@ -1,0 +1,59 @@
+import unittest
+
+from test.helper import FakeYDL
+
+from youtube_dl import FileDownloader
+
+
+class BaseTestFileDownloader(unittest.TestCase):
+    params = {}
+
+    def setUp(self):
+        self.downloader = FileDownloader(FakeYDL(), self.params)
+
+
+class TestPartFileDefaults(BaseTestFileDownloader):
+    params = {}
+
+    def test_temp_name_missing_file(self):
+        # file is missing: should add the default suffix
+        fn = self.downloader.temp_name('some file.ext')
+        self.assertEqual(fn, 'some file.ext.part')
+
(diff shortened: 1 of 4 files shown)


fstirlitz:
You don't need this patch.
```bash
cat > "${XDG_DATA_HOME:-$HOME/.local/share}/mime/packages/partial-video.xml" <<EOF
<?xml version="1.0" encoding="UTF-8"?>
<mime-info xmlns="http://www.freedesktop.org/standards/shared-mime-info">
	<mime-type type="video/x-partial-download+mp4">
		<comment>Partial MP4 download</comment>
		<sub-class-of type="video/mp4"/>
		<glob pattern="*.mp4.part"/>
	</mime-type>
	<mime-type type="video/x-partial-download+x-flv">
		<comment>Partial FLV download</comment>
		<sub-class-of type="video/x-flv"/>
		<glob pattern="*.flv.part"/>
	</mime-type>
</mime-info>
EOF
update-mime-database "${XDG_DATA_HOME:-$HOME/.local/share}/mime"
```

ikn:
It works, but I was trying to avoid listing all the extensions youtube-dl might ever use.  100% chance I'll forget how to add extras a few months down the line when I need to.

(Also, was the intention of the 'sub-class-of' part to have launchers use the superclass type to find a suitable program when there's nothing available for the x-partial-download type?  My file manager doesn't seem to do that, but it's a nice idea.)

PR #5877, opened 2015-06-02
[iprima] Update

slava-sh:


youtube_dl/extractor/iprima.py
```diff
@@ -15,7 +15,7 @@
 
 
 class IPrimaIE(InfoExtractor):
-    _VALID_URL = r'https?://play\.iprima\.cz/[^?#]+/(?P<id>[^?#]+)'
+    _VALID_URL = r'https?://play\.iprima\.cz/([^/]+/)*(?P<id>[^?#]+)'
```
dstftw: Do not capture if you are not going to use it. Change to `(?:[^/]+/)*`.

youtube_dl/extractor/iprima.py
```diff
@@ -102,7 +105,7 @@ def _real_extract(self, url):
 
         return {
             'id': real_id,
-            'title': self._og_search_title(webpage),
+            'title': self._og_search_title(webpage).replace(' | Prima PLAY', ''),
```
dstftw: Use `utils.remove_end` instead.

dstftw:
Thanks.

PR #5882, opened 2015-06-02
Add support for encrypted media in NativeHlsFD

rzhxeo:


yan12125:
Are there advantages to use PyCrypto than the pure python implementation? Current policy is minimizing external library dependency.

rzhxeo:
The python implementation is slow.
It's good for decrypting URLs or subtitles but not for hundreds of megabytes.

yan12125:
Superseded by #8201. Please open an new pull request if you would like to implement a better pure Python AES decryption.

Issue #5884, opened 2015-06-03
Youtube-DL fails to download existing video from youtube.


dopesong:
Command: 

```
youtube-dl --continue --format bestvideo --no-playlist --output "%(autonumber)s-f54852e477709f71b107166f50214f72.%(ext)s" --print-json --ignore-config 'https://www.youtube.com/watch?v=m3YVZ5H166s'
```

Responses:

> WARNING: [youtube] m3YVZ5H166s: Skipping DASH manifest: ExtractorError(u'Cannot decrypt signature without player_url; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.',)
> ERROR: requested format not available

---

> ERROR: unable to download video data: HTTP Error 404: Not Found

Youtube-DL version: 

`youtube-dl is up to date (2015.05.29)`

JSON: 

```
{
  "upload_date":"20150602",
  "extractor":"youtube",
  "height":720,
  "fulltitle":"\u0422\u0440\u0435\u043d\u0438\u043d\u0433_\u041f\u0430\u0432\u043b\u0430_\u0420\u0430\u043a\u043e\u0432\u0430 \u2013 \u0434\u043b\u044f \u043b\u0438\u0434\u0435\u0440\u043e\u0432. \u0410\u0411\u0410 2015-06-02",
  "playlist_index":null,
  "view_count":39,
  "playlist":null,
  "http_headers":{
    "Accept-Charset":"ISO-8859-1,utf-8;q=0.7,*;q=0.7",
    "Accept-Language":"en-us,en;q=0.5",
    "Accept-Encoding":"gzip, deflate",
    "Accept":"text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
    "User-Agent":"Mozilla/5.0 (X11; Linux x86_64; rv:10.0) Gecko/20150101 Firefox/20.0 (Chrome)"
  },
  "_filename":"00001-f54852e477709f71b107166f50214f72.mp4",
  "dislike_count":null,
  "width":1280,
  "subtitles":{

  },
  "age_limit":0,
  "annotations":null,
  "acodec":"none",
  "display_id":"m3YVZ5H166s",
  "automatic_captions":{

  },
  "format":"136 - 1280x720 (DASH video)",
  "requested_subtitles":null,
  "tbr":2200,
  "preference":-40,
[...]

Issue #5754, opened 2015-05-19
unable to download youtube live stream


2011:
This is probably a duplicate of some past bug, but I didn't see anything exactly like it.

This is a live stream (don't know if those eventually get converted to static videos).

$ youtube-dl -F 'https://www.youtube.com/watch?v=pyUCR-Sfenc'
[youtube] pyUCR-Sfenc: Downloading webpage
[youtube] pyUCR-Sfenc: Downloading video info webpage
[youtube] pyUCR-Sfenc: Extracting video information
[youtube] pyUCR-Sfenc: Downloading formats manifest
[youtube] pyUCR-Sfenc: Downloading DASH manifest
[info] Available formats for pyUCR-Sfenc:
format code  extension  resolution note
140          m4a        audio only DASH audio  144k , m4a_dash container, aac  @128k (48000Hz)
141          m4a        audio only DASH audio  272k , m4a_dash container, aac  @256k (48000Hz)
160          mp4        256x144    DASH video  124k , 15fps, video only
133          mp4        426x240    DASH video  258k , 30fps, video only
134          mp4        640x360    DASH video  616k , 30fps, video only
135          mp4        854x480    DASH video 1116k , 30fps, video only
136          mp4        1280x720   DASH video 2216k , 30fps, video only
151          mp4        72p        HLS 
132          mp4        240p       HLS 
92           mp4        240p       HLS 
93           mp4        360p       HLS 
94           mp4        480p       HLS 
95           mp4        720p       HLS  (best)

$ youtube-dl -v 'https://www.youtube.com/watch?v=pyUCR-Sfenc'
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['--restrict-filenames', '-o', '%(title)s-%(id)s-%(uploader)s.%(ext)s', '-w', '-v', 'https://www.youtube.com/watch?v=pyUCR-Sfenc']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.05.15
[debug] Python version 3.4.1 - Linux-3.18.7-x86_64-Intel-R-_Core-TM-_i5_CPU_750_@_2.67GHz
[debug] exe versions: ffmpeg 2.2.14, ffprobe 2.2.14, rtmpdump 2.4
[debug] Proxy map: {}
[youtube] pyUCR-Sfenc: Downloading webpage
[youtube] pyUCR-Sfenc: Downloading video info webpage
[...]

Issue #5703, opened 2015-05-14
Can't download embedded video


Llorx:
This last weekend I recorded some Live events. YouTube has a problem where not saved them like other events, and entering the link wont show the live event but an "event has ended" message: https://www.youtube.com/watch?v=CsmdDsKjzN8

The thing is that the embed link works: https://www.youtube.com/embed/CsmdDsKjzN8 so I thought that is time to download the video to not lose it, just in case YouTUbe definetely deletes it.

The problem comes here. I can't download the vídeo altough it can be watched with the embed code, so I suggest to add a second system to download videos where it uses the same method as the YT embed code instead of the YT main one.

I mean, if the main method fails, use this second method. The first method gives error 400.

I put log files of what I get when downloading:

youtube-dl https://www.youtube.com/embed/CsmdDsKjzN8 -v
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'https://www.youtube.com/embed/CsmdDsKjzN8', u'-v']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.05.10
[debug] Python version 2.6.9 - Linux-3.14.35-28.38.amzn1.x86_64-x86_64-with-glibc2.2.5
[debug] exe versions: ffmpeg git-2015-03-18-f832374, ffprobe git-2015-03-18-f832374, rtmpdump 2.3
[debug] Proxy map: {}
[youtube] CsmdDsKjzN8: Downloading webpage
[youtube] CsmdDsKjzN8: Downloading video info webpage
[youtube] CsmdDsKjzN8: Extracting video information
[youtube] CsmdDsKjzN8: Downloading formats manifest
[youtube] CsmdDsKjzN8: Downloading DASH manifest
[...]

Issue #5702, opened 2015-05-14
YouTube History Feed not downloading


rdvdijk:
I'm trying to download my YouTube History Feed:

```
./youtube-dl-latest ':ythistory' -u **** -p **** -v
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u':ythistory', u'-u', u'PRIVATE', u'-p', u'PRIVATE', u'-v']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.05.10
[debug] Python version 2.7.9 - Darwin-14.3.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg 1.2.4, ffprobe 1.2.4
[debug] Proxy map: {}
[youtube:history] Downloading login page
[youtube:history] Logging in
[youtube:history] history feed: Downloading page 1
[download] Downloading playlist: Youtube Watch History
[youtube:history] playlist Youtube Watch History: Collected 0 video ids (downloading 0 of them)
```

When I use `--write-pages`, I see that the file `history_feed_https_-_www.youtube.com_feed_ajaxaction_load_personal_feed=1_feed_name=history_paging=0.dump` contains:

```
{"load_more_widget_html": "  \n\n\n    \u003cbutton class=\"yt-uix-button yt-uix-button-size-default yt-uix-button-default load-more-button yt-uix-load-more browse-items-load-more-button\" type=\"button\" onclick=\";return false;\" data-uix-load-more-target-id=\"browse-items-primary\" data-uix-load-more-href=\"\/feed_ajax?feed_name=history\u0026amp;action_load_personal_feed=1\"\u003e\u003cspan class=\"yt-uix-button-content\"\u003e  \u003cspan class=\"load-more-loading hid\"\u003e\n      \u003cspan class=\"yt-spinner\"\u003e\n      \u003cspan class=\"yt-spinner-img  yt-sprite\" title=\"Loading icon\"\u003e\u003c\/span\u003e\n\nLoading...\n  \u003c\/span\u003e\n\n  \u003c\/span\u003e\n  \u003cspan class=\"load-more-text\"\u003e\nLoad more\n  \u003c\/span\u003e\n\u003c\/span\u003e\u003c\/button\u003e\n\n", "content_html": "  \n"}%
```

(Looking at that, it seems to be a "Loading..." page.)

Is it a bug, or am I missing something?

jaimeMF:
[...]

PR #5886, opened 2015-06-03
[YoutubeDL] Support DASH manifest downloading

yan12125:
In some videos from youtube, the DASH manifest contains multiple segments URLs for each itag. Existing reports are #5884, #5703, #5702 (the first video is not accessible now, but I remember its DASH manifest contains segments), and I believe #5545, #5754 are affected by the same problem. In fact, downloading any DASH format from a currently live youtube video causes an error.

There are some issues:
1. Is it really needed? For all videos in aforementioned issues, ```-f best``` works fine.
2. My implementation is based on ```NativeHlsFD```, and the two looks similar. Maybe moving common codes to some base class is better.
3. Is it necessary to document ```initialization_url``` and ```segment_urls``` explicitly, in ```common.py``` or so?

### DIFF
--- a/youtube_dl/downloader/dash.py
+++ b/youtube_dl/downloader/dash.py
@@ -1,0 +1,50 @@
+from __future__ import unicode_literals
+from .common import FileDownloader
+from ..compat import compat_urllib_request
+
+import re
+
+
+class DashSegmentsFD(FileDownloader):
+    """
+    Download segments in a DASH manifest
+    """
+    def real_download(self, filename, info_dict):
+        self.report_destination(filename)
+        tmpfilename = self.temp_name(filename)
+        base_url = info_dict['url']
+        segment_urls = info_dict['segment_urls']
+
+        self.byte_counter = 0
+
+        def append_url_to_file(outf, target_url, target_name):
+            self.to_screen('[DashSegments] %s: Downloading %s' % (info_dict['id'], target_name))
+            req = compat_urllib_request.Request(target_url)
+            data = self.ydl.urlopen(req).read()
+            outf.write(data)
+            self.byte_counter += len(data)
+
+        def combine_url(base_url, target_url):
+            if re.match(r'^https?://', target_url):
+                return target_url
+            return '%s/%s' % (base_url, target_url)
+
+        with open(tmpfilename, 'wb') as outf:
+            append_url_to_file(
+                outf, combine_url(base_url, info_dict['initialization_url']),
+                'initialization segment')
+            for i, segment_url in enumerate(segment_urls):
+                append_url_to_file(
+                    outf, combine_url(base_url, segment_url),
+                    'segment %d / %d' % (i + 1, len(segment_urls)))
+
+        self.try_rename(tmpfilename, filename)
+
+        self._hook_progress({
+            'downloaded_bytes': self.byte_counter,
+            'total_bytes': self.byte_counter,
+            'filename': filename,
+            'status': 'finished',
+        })
+
+        return True
--- a/youtube_dl/downloader/dash.py
+++ b/youtube_dl/downloader/dash.py
@@ -16,14 +16,14 @@ class DashSegmentsFD(FileDownloader):
         base_url = info_dict['url']
         segment_urls = info_dict['segment_urls']
 
-        self.byte_counter = 0
+        byte_counter = 0
 
         def append_url_to_file(outf, target_url, target_name):
             self.to_screen('[DashSegments] %s: Downloading %s' % (info_dict['id'], target_name))
             req = compat_urllib_request.Request(target_url)
             data = self.ydl.urlopen(req).read()
             outf.write(data)
-            self.byte_counter += len(data)
+            return len(data)
 
         def combine_url(base_url, target_url):
             if re.match(r'^https?://', target_url):
@@ -35,15 +35,16 @@ class DashSegmentsFD(FileDownloader):
                 outf, combine_url(base_url, info_dict['initialization_url']),
                 'initialization segment')
             for i, segment_url in enumerate(segment_urls):
-                append_url_to_file(
+                segment_len = append_url_to_file(
                     outf, combine_url(base_url, segment_url),
                     'segment %d / %d' % (i + 1, len(segment_urls)))
+                byte_counter += segment_len
 
         self.try_rename(tmpfilename, filename)
 
         self._hook_progress({
-            'downloaded_bytes': self.byte_counter,
-            'total_bytes': self.byte_counter,
+            'downloaded_bytes': byte_counter,
+            'total_bytes': byte_counter,
             'filename': filename,
             'status': 'finished',
         })
--- a/youtube_dl/downloader/dash.py
+++ b/youtube_dl/downloader/dash.py
@@ -16,12 +16,21 @@ class DashSegmentsFD(FileDownloader):
         base_url = info_dict['url']
         segment_urls = info_dict['segment_urls']
 
+        is_test = self.params.get('test', False)
+        remaining_bytes = self._TEST_FILE_SIZE if is_test else None
         byte_counter = 0
 
-        def append_url_to_file(outf, target_url, target_name):
+        def append_url_to_file(outf, target_url, target_name, remaining_bytes=None):
             self.to_screen('[DashSegments] %s: Downloading %s' % (info_dict['id'], target_name))
             req = compat_urllib_request.Request(target_url)
+            if remaining_bytes is not None:
+                req.add_header('Range', 'bytes=0-%d' % (remaining_bytes - 1))
+
             data = self.ydl.urlopen(req).read()
+
+            if remaining_bytes is not None:
+                data = data[:remaining_bytes]
+
             outf.write(data)
             return len(data)
 
@@ -37,8 +46,13 @@ class DashSegmentsFD(FileDownloader):
             for i, segment_url in enumerate(segment_urls):
                 segment_len = append_url_to_file(
                     outf, combine_url(base_url, segment_url),
-                    'segment %d / %d' % (i + 1, len(segment_urls)))
+                    'segment %d / %d' % (i + 1, len(segment_urls)),
+                    remaining_bytes)
                 byte_counter += segment_len
+                if remaining_bytes is not None:
+                    remaining_bytes -= segment_len
+                    if remaining_bytes <= 0:
+                        break
 
         self.try_rename(tmpfilename, filename)
 
--- a/youtube_dl/downloader/dash.py
+++ b/youtube_dl/downloader/dash.py
@@ -1,8 +1,9 @@
 from __future__ import unicode_literals
-from .common import FileDownloader
-from ..compat import compat_urllib_request
 
 import re
+
+from .common import FileDownloader
+from ..compat import compat_urllib_request
 
 
 class DashSegmentsFD(FileDownloader):
--- a/youtube_dl/downloader/http.py
+++ b/youtube_dl/downloader/http.py
@@ -6,7 +6,6 @@ import socket
 import time
 
 from .common import FileDownloader
-from .dash import DashSegmentsFD
 from ..compat import (
     compat_urllib_request,
     compat_urllib_error,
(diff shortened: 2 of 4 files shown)


yan12125:
Issue 4. I think there's no way to determine ```HttpFD``` or ```DashSegmentsFD``` before format selection. Calling another downloader in ```HttpFD``` breaks some abstraction. Maybe there's a better solution?

youtube_dl/downloader/http.py
```diff
@@ -19,6 +20,9 @@
 
 class HttpFD(FileDownloader):
     def real_download(self, filename, info_dict):
+        if info_dict.get('initialization_url') and list(filter(None, info_dict.get('segment_urls', []))):
```
jaimeMF: Why can't this be done in [`get_suitable_downloader`](https://github.com/rg3/youtube-dl/blob/a5ebf77d877ededed377db920faf4f31644573cf/youtube_dl/downloader/__init__.py#L26)?

youtube_dl/downloader/dash.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import unicode_literals
+from .common import FileDownloader
+from ..compat import compat_urllib_request
+
+import re
```
jaimeMF: The import order convention we try to follow is:
1. __future__ imports
2. standard library imports
3. internal imports (aka: all relative imports)

youtube_dl/downloader/dash.py
```diff
@@ -0,0 +1,50 @@
...
+        self.byte_counter = 0
+
+        def append_url_to_file(outf, target_url, target_name):
+            self.to_screen('[DashSegments] %s: Downloading %s' % (info_dict['id'], target_name))
+            req = compat_urllib_request.Request(target_url)
+            data = self.ydl.urlopen(req).read()
+            outf.write(data)
+            self.byte_counter += len(data)
+
+        def combine_url(base_url, target_url):
```
jaimeMF: I haven't tried this code, could `urllib.parse.urljoin` be used instead?

youtube_dl/downloader/http.py
yan12125: I was wrong. Now the detection is in ```get_suitable_downloader```.

youtube_dl/downloader/dash.py
yan12125: Thanks. Fixed.

youtube_dl/downloader/dash.py
yan12125: A sample base_url and target_url is ```https://r2---sn-u5oxu-un5e.googlevideo.com/videoplayback/id/CsmdDsKjzN8.1/itag/135/source/yt_live_broadcast/requiressl/yes/playlist_type/CLEAN/gcr/tw/pmbypass/yes/ratebypass/yes/cmbypass/yes/mime/video%2Fmp4/live/1/gir/yes/upn/8EBB90O9gTY/fexp/9408142,9408288,9408420,9408710,9413503,9415304/signature/5F5F2D0E30C999428C51E2426A741ED87A7FC1B6.668FA2E6A1AA4986A3A77C9AACBB134929E608E9/sver/3/key/cms1/ip/140.112.247.145/ipbits/0/expire/1433448544/sparams/cmbypass,expire,gcr,gir,id,ip,ipbits,itag,live,mime,mm,mn,ms,mv,pl,playlist_type,pmbypass,ratebypass,requiressl,source/mm/32/mn/sn-u5oxu-un5e/ms/lv/mt/1433426887/mv/m/pl/20``` and ```sq/11/clen/683896/lmt/1431241064715340/dur/5.0```. The final URL should be ```https://r2---sn-u5oxu-un5e.googlevideo.com/videoplayback/id/CsmdDsKjzN8.1/itag/135/source/yt_live_broadcast/requiressl/yes/playlist_type/CLEAN/gcr/tw/pmbypass/yes/ratebypass/yes/cmbypass/yes/mime/video%2Fmp4/live/1/gir/yes/upn/8EBB90O9gTY/fexp/9408142,9408288,9408420,9408710,9413503,9415304/signature/5F5F2D0E30C999428C51E2426A741ED87A7FC1B6.668FA2E6A1AA4986A3A77C9AACBB134929E608E9/sver/3/key/cms1/ip/140.112.247.145/ipbits/0/expire/1433448544/sparams/cmbypass,expire,gcr,gir,id,ip,ipbits,itag,live,mime,mm,mn,ms,mv,pl,playlist_type,pmbypass,ratebypass,requiressl,source/mm/32/mn/sn-u5oxu-un5e/ms/lv/mt/1433426887/mv/m/pl/20/sq/11/clen/683896/lmt/1431241064715340/dur/5.0```, and urljoin gives a different result because base_url does not have a trailing slash.

youtube_dl/extractor/youtube.py
```diff
@@ -815,6 +816,12 @@ def decrypt_sig(mobj):
                         'filesize': filesize,
                         'fps': int_or_none(r.attrib.get('frameRate')),
                     }
+                    if segment_list:
```
dstftw: This triggers `FutureWarning` on python 2:
```
youtube_dl\extractor\youtube.py:819: FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
  if segment_list:
```

yan12125: Thanks a lot. Fixed.

dstftw:
I guess this can be merged.

jaimeMF:
I'm testing with https://www.youtube.com/watch?v=sw4hmqVPe0E (if that doesn't work you can look into https://www.youtube.com/live for other live streams). If this change only works for live videos, how are you supposed to get the audio? Also the segments are determined on the video extraction, how does it detects if the live stream has more?

If there's a video know to use the DASH manifest, it would be great to add it as a test.

yan12125:
Currently, no. DASH manifests grows as the live stream goes. As a result, I think bestvideo+bestaudio should be added for non-live streams only. Live streams are served from the start or the current time. It's not reasonable to fetch videos and audios separately.
```Diff
diff --git a/youtube_dl/YoutubeDL.py b/youtube_dl/YoutubeDL.py
index aa6ec9d..c2c0402 100755
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1104,7 +1104,8 @@ class YoutubeDL(object):
         if req_format is None:
             req_format_list = []
             if (self.params.get('outtmpl', DEFAULT_OUTTMPL) != '-' and
-                    info_dict['extractor'] in ['youtube', 'ted']):
+                    info_dict['extractor'] in ['youtube', 'ted'] and
+                    info_dict.get('is_live') is not True):
                 merger = FFmpegMergerPP(self)
                 if merger.available and merger.can_merge():
                     req_format_list.append('bestvideo+bestaudio')
```
(Current ```YoutubeIE``` does not fill ```is_live``` field. That's another problem.)

The only accessible non-live stream that needs this feature seems only https://www.youtube.com/embed/CsmdDsKjzN8 (from #5703). I'll try to add this video to ```_TESTS```.

jaimeMF:
Sounds good (since some people may find useful to use `-f bestaudio` in some live events).

I was checking #5935 and I got a failure:
```
$ youtube-dl --ignore-config -v 'fcTt0LgPMMs'
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['--ignore-config', '-v', 'fcTt0LgPMMs']
[debug] youtube-dl version 2015.05.29
[debug] Git HEAD: 4da31bd
[debug] Python version 3.4.3 - Darwin-13.4.0-x86_64-i386-64bit
[debug] exe versions: avconv 11.4, avprobe 11.4, ffmpeg 2.6.3, ffprobe 2.6.3, rtmpdump 2.4
[debug] Proxy map: {}
[youtube] fcTt0LgPMMs: Downloading webpage
[youtube] fcTt0LgPMMs: Extracting video information
[youtube] fcTt0LgPMMs: Downloading DASH manifest
  File "~/local/bin/youtube-dl", line 13, in <module>
    youtube_dl.main()
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/__init__.py", line 405, in main
    _real_main(argv)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/__init__.py", line 395, in _real_main
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/YoutubeDL.py", line 1502, in download
    res = self.extract_info(url)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/YoutubeDL.py", line 650, in extract_info
    ie_result = ie.extract(url)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/extractor/common.py", line 273, in extract
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/extractor/youtube.py", line 1131, in _real_extract
    video_id, dash_manifest_url, player_url, age_gate)
  File "~/code/proyectos/youtube-dl/youtube-dl/youtube_dl/extractor/youtube.py", line 819, in _parse_dash_manifest
    if len(segment_list):
TypeError: object of type 'NoneType' has no len()
```
I guess doing `if segment_list is not None and len(segment_list):`  is enough but I haven't look further.

yan12125:
Sorry for the mistake, @jaimeMF. ```TypeError``` is fixed in 0c8662d.

```-f bestaudio``` may not be useful. Users may expect youtube-dl downloads to the last minute of the recording, while files extracted from DASH manifest not.

And, I've added a test Youtube_14 for this PR.

**Comment by jaimeMF** [comment]
For livestreams `livestream` and `live_playback` are set to `'1'` in the [ytplayer.config.args](https://github.com/rg3/youtube-dl/blob/02175a7986c4223a0ed27a872c1ca16926913e05/youtube_dl/extractor/youtube.py#L883), so it can be detected. I think we can either don't add dash formats in the extractor or use your patch in YoutubeDL, I dont' have a preference.

**Comment by yan12125** [comment]
There's another example: https://www.youtube.com/watch?v=5VkKartK6qE. In this case, DASH files starts from about 4 hours ago to the current time. I'm not sure whether it's meaningful. Another option can be still listing DASH formats but hiding it (via ```preference```)?

**Comment by jaimeMF** [comment]
I haven't look into all the DASH segements stuff, but maybe they are listing the files since the beginning of the live stream?

>  Another option can be still listing DASH formats but hiding it (via preference)?

I guess that's ok, but you'd still need to make sure `bestvideo+bestaudio` is not selected automatically for live streams.

**Comment by yan12125** [comment]
> maybe they are listing the files since the beginning of the live stream?

I think not, at least for https://www.youtube.com/watch?v=5VkKartK6qE. I download DASH files again and it's started from 21:55:00 (GMT+8) this time, which is exactly 4 hours ago. It's also the leftmost point on the time line of the Youtube player in browsers. I guess DASH manifest contains "recording" of live videos.

Whether it's starting from the beginning of the live stream or from some middle point, it's "history", so I think they are not suitable for live streams.

**Comment by jaimeMF** [comment]
Based on your DASH downloader I have started working on #6012, it's only for livestreams so it's not too useful currently.

Is there something else to do or is it ready to merge?

**Comment by yan12125** [comment]
> Is there something else to do or is it ready to merge?

Nope. Feel free to merge it.

**Comment by jaimeMF** [comment]
I tested yesterday again and normal videos work perfectly, what do we do with livestreams?

**Comment by yan12125** [comment]
I still prefer HLS for youtube live streams as I think most users expect streaming from the current time, while youtube DASH manifests contain 'past' media. This patch is useful for 'just ended' live streams and some [buggy youtube videos](https://github.com/yan12125/youtube-dl/commit/8a1a26ce4c64d7a2c142718fc56f46d9a1c2c4f2).

File: youtube_dl/extractor/youtube.py [hunk 5]
```diff
@@ -516,6 +516,24 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
...
             }
         },
+        # DASH manifest with segment_list
+        {
+            'url': 'https://www.youtube.com/embed/CsmdDsKjzN8',
+            'md5': '8ce563a1d667b599d21064e982ab9e31',
+            'info_dict': {
+                'id': 'CsmdDsKjzN8',
+                'ext': 'mp4',
+                'upload_date': '20150510',
```
jaimeMF: It's quite weird, while in the youtube webpage it says it was uploaded `May 10, 2015`,  the first element where we try to get the date says:
```
<meta itemprop="datePublished" content="2015-05-01">
```

I will change it to `20150501` for now.

jaimeMF:
Merged.

(If I miss any issue fixed by this, feel free to close it yourself).

PR #5894, opened 2015-06-04
Add videowood extractor

luogni:


jaimeMF:
The test video is a copyrighted tv show episode ([Wikipedia article](https://en.wikipedia.org/wiki/Between_%28TV_series%29)) and I doubt they have the rights for distributing that content, and [it seems that a lot of their videos are infringing copyright](https://www.google.com/transparencyreport/removals/copyright/domains/videowood.tv/). [We can't add support for this kind if sites](https://github.com/rg3/youtube-dl/blob/4c8fea92f350d5a3f33d505980ac750b05a9cd34/README.md#can-you-add-support-for-this-anime-video-site-or-site-which-shows-current-movies-for-free).

PR #5896, opened 2015-06-04
[qqmusic] Extract additional formats (mp3-128, mp3-320)

ping:
Add support for higher quality downloads.


yan12125:
Seems mp3-128 have audio bitrate 128kb/s and mp3-320 have 320kb/s. It would be great to add ```abr``` field for these formats.

ping:
Thanks for the ```abr``` suggestion. Added.

yan12125:
Thanks for the work! Merged.

Issue #2910, opened 2014-05-13
[FFmpeg] Add "-movflags faststart" opt when using MP4 container


m132:
This flag moves MP4's moov atom to the beginning of file, so it can be streamed. It works only with ffmpeg, avconv (at least in Ubuntu) doesn't support this yet.

svnpenn:
#5733

svnpenn:
#5942

PR #5942, opened 2015-06-10
Adds support for XviD output with extra parametrization

### LINKED ISSUE #5733: add XviD AVI option to --recode-video
Some TVs, on friend's homes and schools, only support DivX or XviD videos. So, would be nice to have this feature on youtube-dl.

aurium:
As the "LG Time Machine" (a (not so) smart TV) has a limitation for video dimensions (as for codecs), I take to implement an extra parameter `--pp-params` where we can send extra parameterization for the video converter (post-processor).

Example:
```
$ youtube-dl --recode-video=xvid --pp-params='-s 720x480' -c https://www.youtube.com/watch?v=BE7Qoe2ZiXE
```
That works fine on a 4yo LG Time Machine.

Closes #5733

### DIFF
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -171,8 +171,10 @@ def _real_main(argv=None):
     if opts.recodevideo is not None:
         if opts.recodevideo not in ['mp4', 'flv', 'webm', 'ogg', 'mkv', 'xvid']:
             parser.error('invalid video recode format specified')
-    if opts.pp_params is not None:
-        opts.pp_params = opts.pp_params.split()
+    if opts.pp_params is None:
+        opts.pp_params = []
+    else:
+        opts.pp_params = shlex.split(opts.pp_params)
     if opts.convertsubtitles is not None:
         if opts.convertsubtitles not in ['srt', 'vtt', 'ass']:
             parser.error('invalid subtitle format specified')
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -171,10 +171,6 @@ def _real_main(argv=None):
     if opts.recodevideo is not None:
         if opts.recodevideo not in ['mp4', 'flv', 'webm', 'ogg', 'mkv', 'xvid']:
             parser.error('invalid video recode format specified')
-    if opts.pp_params is None:
-        opts.pp_params = []
-    else:
-        opts.pp_params = shlex.split(opts.pp_params)
     if opts.convertsubtitles is not None:
         if opts.convertsubtitles not in ['srt', 'vtt', 'ass']:
             parser.error('invalid subtitle format specified')
@@ -231,7 +227,7 @@ def _real_main(argv=None):
         postprocessors.append({
             'key': 'FFmpegVideoConvertor',
             'preferedformat': opts.recodevideo,
-            'extra_params': opts.pp_params
+            'extra_cmd_args': opts.postprocessor_args,
         })
     if opts.convertsubtitles:
         postprocessors.append({
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -169,8 +169,10 @@ def _real_main(argv=None):
         if not opts.audioquality.isdigit():
             parser.error('invalid audio quality specified')
     if opts.recodevideo is not None:
-        if opts.recodevideo not in ['mp4', 'flv', 'webm', 'ogg', 'mkv']:
+        if opts.recodevideo not in ['mp4', 'flv', 'webm', 'ogg', 'mkv', 'xvid']:
             parser.error('invalid video recode format specified')
+    if opts.pp_params is not None:
+        opts.pp_params = opts.pp_params.split()
     if opts.convertsubtitles is not None:
         if opts.convertsubtitles not in ['srt', 'vtt', 'ass']:
             parser.error('invalid subtitle format specified')
@@ -227,6 +229,7 @@ def _real_main(argv=None):
         postprocessors.append({
             'key': 'FFmpegVideoConvertor',
             'preferedformat': opts.recodevideo,
+            'extra_params': opts.pp_params
         })
     if opts.convertsubtitles:
         postprocessors.append({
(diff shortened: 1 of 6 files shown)


dstftw:
I would prefer this to be implemented similarly to [`--external-downloader-args`](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/options.py#L418-L421). Note that you also miss some places where it should be documented, namely [here](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L261-L263) and possibly [here](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/postprocessor/common.py#L14-L25).

aurium:
Thanks @dstftw. I'll update this MR as you point.

dstftw:
Thanks.

aurium:
These nosetests errors are random?

dstftw:
When suggesting [`--external-downloader-args`](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/options.py#L418-L421)-similar approach I was expecting a generic approach by [exposing the interface for obtaining extra arguments in base class](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/downloader/external.py#L54-L59) that can be used by descendants like [this](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/downloader/external.py#L81) or [this](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/downloader/external.py#L100-L101) if necessary. I would also appreciate similar naming convention as we have for external downloader, i.e. `--postprocessor-args`.

svnpenn:
#2817

svnpenn:
#2910

aurium:
Ok @dstftw, i'll do it. Thanks for helping me to understand the project structure. :-)

svnpenn:
:+1:

aurium:
It is not exactly equal to the externals downloader approach, however the `self._extra_cmd_args` is `PostProcessor` class responsibility.

dstftw:
Why not just use [this approach](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/downloader/external.py#L55) instead of passing `extra_cmd_args` to each processor's ctor?

aurium:
`PostProcessor` class has no `self.params` :-(

I could not add it to `PostProcessor` simply like `FileDownloader` because `FileDownloader` sub-classes has no `__init__`, so it can get the `params` argument without passing super to super. We don't have that look with `PostProcessor`.

As i'm not a python expert you may show me another way. :-)

dstftw:
>`PostProcessor` class has no `self.params` :-(

It's in `self._downloader.params`.

aurium:
Thank you @dstftw! It is much simpler using `downloader.params`, however i wouldn't know it without help.

youtube_dl/__init__.py
```diff
@@ -169,7 +169,7 @@ def _real_main(argv=None):
         if not opts.audioquality.isdigit():
             parser.error('invalid audio quality specified')
     if opts.recodevideo is not None:
-        if opts.recodevideo not in ['mp4', 'flv', 'webm', 'ogg', 'mkv']:
+        if opts.recodevideo not in ['mp4', 'flv', 'webm', 'ogg', 'mkv', 'xvid']:
```
dstftw: I think it's safe to use `avi` here since we use extensions for this option. Moreover, it will simplify [this code](https://github.com/aurium/youtube-dl/blob/master/youtube_dl/postprocessor/ffmpeg.py#L297-L301) a bit.

aurium: I understand your point, but AVI is a so generic container and it may conflict with future needs. Pointing the codec *(that we really need)* may set this feature more secure and help the user to be sure about what they will get.

dstftw: mkv is even more generic but already mentioned.

svnpenn: Frankly why are you even interested in XviD? Doesnt H.264/H.265 beat it in every way?

svnpenn: @aurium why are you even interested in XviD? Doesnt H.264/H.265 beat it in every way?

dstftw: @svnpenn https://github.com/rg3/youtube-dl/pull/5942#issue-86797178.

svnpenn: I am all for the `--postprocessor-args`, but if your Smart TV doesnt support the most ubiquitous codec of the last 5 years, maybe the right thing to do is get a new TV rather than spend weeks writing support for it

aurium: @svnpenn thanks for your proposal, but... i'm on a country well known as "third world"; this model of Smart TV is used by schools; and if i have money to throw away a TV and buy a new one, still that is a really bad action against the planet. So... i understand your good intention, but better software can save money, classrooms and the blue marble ball. :-)

aurium: By the way @dstftw, mkv is also another generic container, this does not reinforce my argument? As XviD is expected to be inside a AVI, is not this a better approach?

Well this is my proposal, but i accept to change the `opts.recodevideo` value to `avi` if this is your final decision. No problem.

svnpenn: @aurium I am confused, what is your argument? That AVI should not be the default container for XviD video? If so I agree with @dstftw, you need to make a case for that, because while it is not the only container that supports XviD, it is certainly the defacto one.

dstftw:
@aurium I've already mentioned [that mkv is even more generic](https://github.com/rg3/youtube-dl/pull/5942#discussion_r33882891) and as you can see `--recode-video` accepts container names not codec names. Therefore sticking with `avi` seems consistent to me. Otherwise, we'll need to rename everything to codec names that will break interface or keep both which will make little sense. Moreover, for an average user `avi` is more likely to say something that `xvid` or `h.264`.

svnpenn:
@dstftw I dont think H.264 even works with AVi

dstftw:
@svnpenn that was just for example that fancy codec name may be totally unclear for an average user.

aurium:
Ok @dstftw, sooner as i can i ll update it.

dstftw:
Thanks, merged.

aurium:
Thank you @dstftw!

PR #5961, opened 2015-06-12
Add --force-generic-extractor

dstftw:
For some extractors that are hard to workout a good `_VALID_URL` we use very vague and unrestrictive ones, e.g. just allowing anything after hostname and capturing part of URL as id.
If some of these extractors happen to have a video embed of some different hoster or platform
and this scenario was not handled in extractor itself we end up with inability to download this embed
until extractor is fixed to support embeds of this kind.
Forcing downloader to use the generic extractor can be a neat temporary solution for this problem.

**Example:**
[5-tv](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/fivetv.py) extractor with [Tvigle](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/generic.py#L1324-L1328) embed (may be georestricted to Russia)
http://www.5-tv.ru/rabota/broadcasts/48/

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -627,15 +627,14 @@ class YoutubeDL(object):
             info_dict.setdefault(key, value)
 
     def extract_info(self, url, download=True, ie_key=None, extra_info={},
-                     process=True):
+                     process=True, force_generic_extractor=False):
         '''
         Returns a list with a dictionary for each video we find.
         If 'download', also downloads the videos.
         extra_info is a dict containing the extra values to add to each result
         '''
 
-        if not ie_key and self._force_generic_extractor_required:
-            self._force_generic_extractor_required = False
+        if not ie_key and force_generic_extractor:
             ie_key = 'Generic'
 
         if ie_key:
@@ -663,7 +662,7 @@ class YoutubeDL(object):
                     }
                 self.add_default_extra_info(ie_result, ie, url)
                 if process:
-                    return self.process_ie_result(ie_result, download, extra_info)
+                    return self.process_ie_result(ie_result, download, extra_info, force_generic_extractor=False)
                 else:
                     return ie_result
             except ExtractorError as de:  # An error we somewhat expected
@@ -688,7 +687,7 @@ class YoutubeDL(object):
             'extractor_key': ie.ie_key(),
         })
 
-    def process_ie_result(self, ie_result, download=True, extra_info={}):
+    def process_ie_result(self, ie_result, download=True, extra_info={}, force_generic_extractor=False):
         """
         Take the result of the ie(may be modified) and resolve all unresolved
         references (URLs, playlist items).
@@ -716,7 +715,8 @@ class YoutubeDL(object):
             return self.extract_info(ie_result['url'],
                                      download,
                                      ie_key=ie_result.get('ie_key'),
-                                     extra_info=extra_info)
+                                     extra_info=extra_info,
+                                     force_generic_extractor=force_generic_extractor)
         elif result_type == 'url_transparent':
             # Use the information from the embedding page
             info = self.extract_info(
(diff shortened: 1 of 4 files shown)


youtube_dl/YoutubeDL.py
```diff
@@ -633,6 +635,10 @@ def extract_info(self, url, download=True, ie_key=None, extra_info={},
         extra_info is a dict containing the extra values to add to each result
         '''
 
+        if not ie_key and self._force_generic_extractor_required:
+            self._force_generic_extractor_required = False
```
jaimeMF: I think that this line would produce unexpected results with multiple urls. The first one would use the generic extractor and the rest would use the normal extractors. (I may have misunderstood it)

dstftw: Good catch. Fixed in 9f4323252abade4f10b0884682f92cedc78b4d4a.

jaimeMF:
I'd prefer to not introduce a global state with `_force_generic_extractor_required`. Implementing multiple concurrent downloads/extraction would be more difficult.

Do you think it can be implemented with a different approach?

dstftw:
Sure it can, we'll just need to pass it as argument all over.

dstftw:
Generally, I agree on concurrency, but we already have some [global state](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L277-L286) that we'll need to get rid of if we eventually will implement parallel downloading.

jaimeMF:
> Sure it can, we'll just need to pass it as argument all over.

I guess it could be a bit ugly, but I think we should compare both solutions

> Generally, I agree on concurrency, but we already have some global state that we'll need to get rid of if we eventually will implement parallel downloading.

Form those, I think that the only two that are problematic are `_download_retcode` and `_num_downloads`, i wouldn't expect the rest to cause too much issues.

dstftw:
Replaced global state with argument passing in 61aa5ba36eea3b7cf8c3570ab33604dd2c13b855.

youtube_dl/YoutubeDL.py
```diff
@@ -683,7 +687,7 @@ def add_default_extra_info(self, ie_result, ie, url):
             'extractor_key': ie.ie_key(),
         })
 
-    def process_ie_result(self, ie_result, download=True, extra_info={}):
+    def process_ie_result(self, ie_result, download=True, extra_info={}, force_generic_extractor=False):
```
jaimeMF: If we are always setting it to `False` I think we can remove the parameter from process_ie_result until we actually need it.

dstftw: Right, missed that. Fixed in 0072afc.

dstftw: However, we may probably need it for url_transparent, playlist and compat_list.

atomicdryad:
Works great on http://www.pbs.org/newshour/bb/terrorism-jan-june10-cia_01-05/

A --generic-fallback option would be great, too: it could be set as a default, and cause GenericIE to run if a site-specific extractor fails (and would likely require blacklisting said extractor, or just running everything after find_video)

dstftw:
>A --generic-fallback option would be great, too: it could be set as a default, and cause GenericIE to run if a site-specific extractor fails

That's a great idea.

dstftw:

jaimeMF:
Only one comment, I would keep the warning so that users don't get surprised if it fails.

dstftw:
Well, generally this warning sounds little bit misleading when `--force-generic-extractor` is used since it does not actually *fallback*. Moreover, users who are going to use this have to clearly understand the intention and be aware of the consequences. However, probably different message will suit it better, like `Forcing on generic information extractor.`?

jaimeMF:
Using a different message sounds good.

dstftw:
Added custom message. Thanks.

PR #5966, opened 2015-06-13
Added support for Geeks and Sundry

ciferkey:
Added support for Geeks and Sundry.

For issue https://github.com/rg3/youtube-dl/issues/5946


dstftw:
I suggest you to read [this](https://github.com/rg3/youtube-dl#developer-instructions) in the first place.

PR #5970, opened 2015-06-13
Brightcove extractor: support customBC.createVideo(...); method (america.aljazeera.com and others)

atomicdryad:
found in http://www.americanbar.org/groups/family_law.html and
http://america.aljazeera.com/watch/shows/america-tonight/2015/6/exclusive-hunting-isil-with-the-pkk.html


dstftw:
Thanks. Merged with some improvements.

PR #5971, opened 2015-06-13
Generic: Return a playlist of all vimeo iframes/embeds

atomicdryad:
This article contains multiple vimeo embeds:

http://www.theguardian.com/environment/2015/jun/10/climate-change-has-left-us-exposed-in-arctic-say-military-experts

This change merges the vimeo iframe and moogaloop embed regexps and returns a playlist.

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -1090,17 +1090,10 @@ class GenericIE(InfoExtractor):
             return _playlist_from_matches(matches, ie='RtlNl')
 
         # Look for embedded (iframe) Vimeo player
-        mobj = re.search(
-            r'<iframe[^>]+?src=(["\'])(?P<url>(?:https?:)?//player\.vimeo\.com/video/.+?)\1', webpage)
-        if mobj:
-            player_url = unescapeHTML(mobj.group('url'))
-            surl = smuggle_url(player_url, {'Referer': url})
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/generic.py
```diff
@@ -1090,17 +1090,10 @@ def _playlist_from_matches(matches, getter=None, ie=None):
             return _playlist_from_matches(matches, ie='RtlNl')
 
         # Look for embedded (iframe) Vimeo player
-        mobj = re.search(
-            r'<iframe[^>]+?src=(["\'])(?P<url>(?:https?:)?//player\.vimeo\.com/video/.+?)\1', webpage)
-        if mobj:
-            player_url = unescapeHTML(mobj.group('url'))
```
dstftw: I guess `unescapeHTML` here was intentional. Are you sure it's not required in new code?

atomicdryad:
From what I could tell no; it extracts matching moogaloop\.swf\?clip_id=\d+

from this bit of VALID_URL (?:(?:play_redirect_hls|moogaloop\.swf)\?clip_id=)?(?:videos?/)?(?P<id>[0-9]+)

So & or &amp; won't matter. (Of course expecting clip_id to be the first param is an issue..)

atomicdryad:
Obsolete due to b407e17, new PR: #6054

PR #5983, opened 2015-06-14
LiveLeak: support more original videos

jomo:
some (old?) videos use `...mp4.h264_270p.mp4...` instead of `...mp4.h264_base.mp4...` in their URLs.

This is an addition to #4768.


dstftw:
Can you add a test?

jomo:
@dstftw done.

dstftw:
Thanks.

PR #5985, opened 2015-06-14
[safari] make url regex more lenient

zx8:
It seems IDs can contain letters as well as numbers - e.g. https://www.safaribooksonline.com/library/view/create-a-nodejs/100000006A0210 - so just match any characters that aren't `/` (although `\w+` would probably be sufficient).


dstftw:
Can you add a test?

PR #5993, opened 2015-06-15
[youtube] Fix missing format details for 60fps DASH formats

anssih:
60fps DASH formats do not appear in the DASH manifest, but the non-DASH
video info page does contain additional parameters for DASH formats that
we can parse.

Use those when they exist.

This fixes "bestvideo" not selecting 60fps formats over similar 30fps
formats just because the file size is unknown.

This can been in e.g.:
youtube-dl -F https://www.youtube.com/watch?v=2KApp699WdE

The formats 298 and 299 have no file size and thus the 30fps formats get preferred over the 60fps formats without this patch.


RSully:
@anssih the tests aren't passing

anssih:
@RSully Fixed, the new failures were on Python 3.x and I had tested just 2.7.

Other failures seem to not be new but happen without the commit as well.

dstftw:
Thanks, cherry-picked in c9afb51ceaeae7cbd74502e0451eb9163a1ab7fb. Sorry for delay.

PR #6009, opened 2015-06-18
[kuwo] New extractor for kuwo.cn

ping:
Kuwo is a legal Chinese music streaming service (similar to qqmusic).


yan12125:
Codes in ```KuwoSingerIE``` and ```KuwoSingerMusicIE``` are much the same. The former can be viewed as a _single page_ version of the latter. Would you like to merge them into a single extractor?

ping:
Merged the two extractors.

yan12125:
Thanks merged.

PR #6010, opened 2015-06-18
[qqmusic] Fix album extraction

ping:
The old webpage based extraction method was not working for new albums. A json endpoint is now available for extracting album info and songs.


yan12125:
Thanks. Merged with some modifications.

PR #6016, opened 2015-06-18
[qqmusic] Fix song extraction when certain formats are unavailable

ping:
Some songs do not have hq (320k) versions available. This fixes it by checking if the url is valid before adding to formats.

Also includes a small change to get the thumbnail url for a song based on the album mid.

### DIFF
--- a/youtube_dl/extractor/qqmusic.py
+++ b/youtube_dl/extractor/qqmusic.py
@@ -9,13 +9,8 @@ from .common import InfoExtractor
 from ..utils import (
     strip_jsonp,
     unescapeHTML,
-    HEADRequest,
-    ExtractorError,
 )
-from ..compat import (
-    compat_urllib_request,
-    compat_HTTPError,
-)
+from ..compat import compat_urllib_request
 
 
 class QQMusicIE(InfoExtractor):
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/qqmusic.py
```diff
@@ -77,14 +104,24 @@ def _real_extract(self, url):
...
-                'url': 'http://cc.stream.qqmusic.qq.com/%s%s.%s?vkey=%s&guid=%s&fromtag=0'
-                       % (details['prefix'], mid, details['ext'], vkey, guid),
-                'format': format_id,
-                'format_id': format_id,
-                'preference': details['preference'],
-                'abr': details.get('abr'),
-            })
+            video_url = 'http://cc.stream.qqmusic.qq.com/%s%s.%s?vkey=%s&guid=%s&fromtag=0' \
+                        % (details['prefix'], mid, details['ext'], vkey, guid)
+            req = HEADRequest(video_url)
```
jaimeMF: Could you use [`_check_formats` or `_is_valid_url`](https://github.com/rg3/youtube-dl/blob/621ed9f5f4d9d82659272ebe01e740e9196fad61/youtube_dl/extractor/common.py#L771-L792)?

ping:
@jaimeMF Thanks, switched to using ``_check_formats`` instead.

yan12125:
Thanks. Merged.

PR #6026, opened 2015-06-19
Support BBC News (bbc.com/news)

atomicdryad:
bbc.com/news uses the same mediaselector format, but a different mediaselector url, and inlined json + xml playlist.sxml methods for obtaining video ids.

_download_media_selector is mostly redundant except for the hardcoded url; perhaps it should be made a third argument?


docs/supportedsites.md
```diff
@@ -50,6 +50,7 @@
  - **Bandcamp**
  - **Bandcamp:album**
  - **bbc.co.uk**: BBC iPlayer
+ - **bbc.com**: BBC news videos
```
yan12125: This file is updated automatically in releases. Please do not modify it in pull requests.

youtube_dl/extractor/bbcnews.py
```diff
@@ -0,0 +1,124 @@
...
+            'title': 'Aerial footage showed the site of the crash in the Alps - courtesy BFM TV',
+            'description': 'Germanwings plane crash site in aerial video - Aerial footage showed the site of the crash in the Alps - courtesy BFM TV',
+            'duration': 47,
+        },
+        'params': {
+            'skip_download': True,
+        }
+    }]
+
+    def _duration_str2int(self, str):
```
yan12125: I haven't tested it thoroughly, but maybe [parse_duration](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/utils.py#L1299) have the same functionality.

docs/supportedsites.md
atomicdryad: Reverted

youtube_dl/extractor/bbcnews.py
atomicdryad: Yeah it does, removed

youtube_dl/extractor/bbcnews.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def _real_extract(self, url):
+        list_id = self._match_id(url)
+        webpage = self._download_webpage(url, list_id)
+
+        list_title = self._html_search_regex(r'<title>(.*?)(?:\s*-\s*BBC News)?</title>', webpage, 'list title')
+
+        pubdate = self._html_search_regex(r'"datePublished":\s*"(\d+-\d+-\d+)', webpage, 'date', default=None)
+        if pubdate:
+           pubdate = pubdate.replace('-','')
```
dstftw: Use [`parse_iso8601`](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/utils.py#L716-L739) to extract timestamp instead.

youtube_dl/extractor/bbcnews.py
```diff
@@ -0,0 +1,109 @@
...
+
+        ret = []
+        # works with bbc.com/news/something-something-123456 articles
+        matches = re.findall(r"data-media-meta='({[^']+})'", webpage)
+        if not matches:
+           # stubbornly generic extractor for {json with "image":{allvideoshavethis},etc}
+           # in http://www.bbc.com/news/video_and_audio/international
+           matches = re.findall(r"({[^{}]+image\":{[^}]+}[^}]+})", webpage)
```
dstftw: Use `_html_search_regex/_search_regex` with list of regex patterns instead.

youtube_dl/extractor/bbcnews.py
```diff
@@ -0,0 +1,109 @@
...
+        if not matches:
+           raise ExtractorError('No video found', expected=True)
+
+        for ent in matches:
+            jent = self._parse_json(ent,list_id)
+
+            programme_id = jent.get('externalId',None)
```
dstftw: You don't need to specify `None` it's default.

youtube_dl/extractor/bbcnews.py
```diff
@@ -0,0 +1,109 @@
...
+
+        ret = []
+        # works with bbc.com/news/something-something-123456 articles
+        matches = re.findall(r"data-media-meta='({[^']+})'", webpage)
+        if not matches:
+           # stubbornly generic extractor for {json with "image":{allvideoshavethis},etc}
+           # in http://www.bbc.com/news/video_and_audio/international
+           matches = re.findall(r"({[^{}]+image\":{[^}]+}[^}]+})", webpage)
+        if not matches:
+           raise ExtractorError('No video found', expected=True)
```
dstftw: After you rewrite aforementioned regexes with `_search_regex` you won't need this anymore.

dstftw:
I would also prefer placing both extractors to a single source file and finally renaming it to `bbc.py`.

atomicdryad:
This says it has merge conflicts but it merges fine?

dstftw:
Thanks, merged.

youtube_dl/extractor/bbc.py
```diff
@@ -377,3 +381,177 @@ def _real_extract(self, url):
...
+                        sval['id'] = skey
+                        jsent.append(sval)
+
+        if len(jsent) == 0:
+            # stubbornly generic extractor for {json with "image":{allvideoshavethis},etc}
+            # in http://www.bbc.com/news/video_and_audio/international
+            # prone to breaking if entries have sourceFiles list
+            jsent = map(
+                lambda m: self._parse_json(m, list_id),
+                re.findall(r"({[^{}]+image\":{[^}]+}[^}]+})", webpage)
```
dstftw: Could you provide more info on that scenario? I was unable to find any example URL that wouldn't be covered by previous patterns.

PR #6033, opened 2015-06-20
[xvx] Add new extractor

devnomnom:


youtube_dl/extractor/xvx.py
```diff
@@ -0,0 +1,40 @@
...
+from .common import InfoExtractor
+from ..compat import (
+    compat_urllib_request,
+    compat_urlparse
+)
+
+class XVXIE(InfoExtractor):
+    _VALID_URL =r'(?:https?://)?(?:www\.)?xvx\.so/view/(?P<id>\d+)'
+    _TEST = {
+        'url': 'http://xvx.so/view/205728768',
```
dstftw: This looks like copyrighted content. youtube-dl does not support sites dedicated to copyright infringement. Do they have rights to distribute this? If not it has to be replaced with one that does not violate any copyrights.

dstftw:
Closing since no example URL that does not infringe copyrights have been provided.

PR #6040, opened 2015-06-21
[neteasemusic] Add new extractor for music.163.com

ping:
http://music.163.com is the music streaming service from [NetEase/网易] (https://en.wikipedia.org/wiki/NetEase).

The site does not geo-restrict AFAIK and has hq downloads available. No login is required.

### DIFF
--- a/youtube_dl/extractor/neteasemusic.py
+++ b/youtube_dl/extractor/neteasemusic.py
@@ -1,0 +1,434 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from hashlib import md5
+from datetime import datetime
+import itertools
+import re
+
+from .common import InfoExtractor
+from ..compat import (
+    compat_urllib_request,
+    compat_urllib_parse,
+)
+
+
+class NetEaseMusicBaseIE(InfoExtractor):
+    _FORMATS = ['bMusic', 'mMusic', 'hMusic']
+    _NETEASE_SALT = '3go8&$8*3*3h0k(2)2'
+    _API_BASE = 'http://music.163.com/api/'
+
+    @classmethod
+    def _encrypt(cls, dfsid):
+        salt_bytes = bytearray(str(cls._NETEASE_SALT))
+        string_bytes = bytearray(str(dfsid))
+        salt_len = len(salt_bytes)
+        for i in xrange(len(string_bytes)):
+            string_bytes[i] = string_bytes[i] ^ salt_bytes[i % salt_len]
+        m = md5()
+        m.update(string_bytes)
+        result = m.digest().encode('base64')[:-1]
+        return result.replace('/', '_').replace('+', '-')
+
+    @classmethod
+    def extract_formats(cls, info):
+        formats = []
+        for song_format in cls._FORMATS:
+            details = info.get(song_format)
+            if not details:
+                continue
+            formats.append({
+                'url': 'http://m1.music.126.net/%s/%s.%s' %
+                       (cls._encrypt(details['dfsId']), details['dfsId'],
+                        details['extension']),
+                'ext': details['extension'],
+                'abr': details['bitrate'] / 1000,
+                'preference': details['bitrate'],
+                'format_id': song_format,
+                'filesize': details['size'],
+                'asr': details['sr']
+            })
+        return formats
+
+    def query_api(self, endpoint, video_id, note):
+        req = compat_urllib_request.Request('%s%s' % (self._API_BASE, endpoint))
+        req.add_header('Referer', self._API_BASE)
+        return self._download_json(req, video_id, note)
+
+
+class NetEaseMusicIE(NetEaseMusicBaseIE):
+    IE_NAME = 'netease:song'
+    _VALID_URL = r'https?://music\.163\.com/(#/)?song\?id=(?P<id>[0-9]+)'
+    _TESTS = [{
+        'url': 'http://music.163.com/#/song?id=32102397',
+        'md5': 'f2e97280e6345c74ba9d5677dd5dcb45',
+        'info_dict': {
+            'id': '32102397',
+            'ext': 'mp3',
+            'title': 'Bad Blood (feat. Kendrick Lamar)',
+            'creator': 'Taylor Swift / Kendrick Lamar',
+            'upload_date': '20150517',
+            'timestamp': 1431878400,
+            'description': 'md5:a10a54589c2860300d02e1de821eb2ef',
+        },
+    }, {
+        'note': 'No lyrics translation.',
+        'url': 'http://music.163.com/#/song?id=29822014',
+        'info_dict': {
+            'id': '29822014',
+            'ext': 'mp3',
+            'title': '听见下雨的声音',
+            'creator': '周杰伦',
+            'upload_date': '20141225',
+            'timestamp': 1419523200,
+            'description': 'md5:a4d8d89f44656af206b7b2555c0bce6c',
+        },
+    }, {
+        'note': 'No lyrics.',
+        'url': 'http://music.163.com/song?id=17241424',
+        'info_dict': {
+            'id': '17241424',
+            'ext': 'mp3',
+            'title': 'Opus 28',
+            'creator': 'Dustin O\'Halloran',
+            'upload_date': '20080211',
+            'timestamp': 1202745600,
+        },
+    }]
+
+    def _process_lyrics(self, lyrics_info):
+        original = lyrics_info.get('lrc', {}).get('lyric')
+        translated = lyrics_info.get('tlyric', {}).get('lyric')
+
+        if not translated:
+            return original
+
+        lyrics_expr = r'(\[[0-9]{2}:[0-9]{2}\.[0-9]{2,}\])([^\n]+)'
+        original_ts_texts = re.findall(lyrics_expr, original)
+        translation_ts_dict = {
+            time_stamp: text for time_stamp, text in re.findall(lyrics_expr, translated)
+        }
+
+        lyrics = '\n'.join([
+            '%s%s / %s' % (time_stamp, text, translation_ts_dict.get(time_stamp, ''))
+            for time_stamp, text in original_ts_texts
+        ])
+        return lyrics
+
+    def _real_extract(self, url):
+        song_id = self._match_id(url)
+
+        params = {
+            'id': song_id,
+            'ids': '[%s]' % song_id
+        }
+        info = self.query_api(
+            'song/detail?' + compat_urllib_parse.urlencode(params),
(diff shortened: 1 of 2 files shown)


yan12125:
Lots of contents on 163.com violates copyright (see the [Google transparency report](https://www.google.com/transparencyreport/removals/copyright/domains/163.com/) ). I'm not sure whether 163.com is the same as music.163.com or not. Can you provide evidences that music.163.com has legal rights over the contents it provides?

yan12125:
According to this [zhihu question](http://www.zhihu.com/question/20981566) (in Chinese), not all songs are legal. Note our policy is that youtube-dl does not support copyright infringing sites, see [this FAQ](https://github.com/rg3/youtube-dl#can-you-add-support-for-this-anime-video-site-or-site-which-shows-current-movies-for-free).

ping:
I'm not entirely familiar with the Chinese music streaming market but I've seen reports of QQMusic and NetEase Cloud Music (music.163.com) suing each other for rights infringement. Both parties have been found guilty in different cases and compelled legally to take down content. http://toutiao.com/a3739968936/

The linked article also refers to Netease acquiring the rights to distribute Taylor Swift's music (with fee).

youtube_dl/extractor/neteasemusic.py
```diff
@@ -0,0 +1,439 @@
...
+                'abr': details['bitrate'] / 1000,
+                'preference': details['bitrate'],
+                'format_id': song_format,
+                'filesize': details['size'],
+                'asr': details['sr']
+            })
+        return formats
+
+    @classmethod
+    def convert_milliseconds(cls, ms):
```
yan12125: Though not thoroughly tested, I think [```int_or_none```](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/utils.py#L1278-1284) has similar functionality.

youtube_dl/extractor/neteasemusic.py
```diff
@@ -0,0 +1,439 @@
...
+        original = lyrics_info.get('lrc', {}).get('lyric')
+        translated = lyrics_info.get('tlyric', {}).get('lyric')
+
+        if not translated:
+            return original
+
+        lyrics_expr = r'(\[[0-9]{2}:[0-9]{2}\.[0-9]{2,}\])([^\n]+)'
+        original_ts_texts = re.findall(lyrics_expr, original)
+        translation_ts_dict = {
+            time_stamp: text for time_stamp, text in re.findall(lyrics_expr, translated)
```
yan12125: This line causes SyntaxError in Python 2.6:
```
$ python2.6 test/test_download.py TestDownload.test_Neteasemusic
  File "test/test_download.py", line 11, in <module>
    from test.helper import (
  File "/home/yen/tmp/youtube-dl-ping-neteasemusic/test/helper.py", line 12, in <module>
    import youtube_dl.extractor
  File "/home/yen/tmp/youtube-dl-ping-neteasemusic/youtube_dl/__init__.py", line 43, in <module>
    from .extractor import gen_extractors, list_extractors
  File "/home/yen/tmp/youtube-dl-ping-neteasemusic/youtube_dl/extractor/__init__.py", line 342, in <module>
    from .neteasemusic import (
  File "/home/yen/tmp/youtube-dl-ping-neteasemusic/youtube_dl/extractor/neteasemusic.py", line 113
    time_stamp: text for time_stamp, text in re.findall(lyrics_expr, translated)
                       ^
SyntaxError: invalid syntax
```

youtube_dl/extractor/neteasemusic.py
```diff
@@ -0,0 +1,439 @@
...
+    _FORMATS = ['bMusic', 'mMusic', 'hMusic']
+    _NETEASE_SALT = '3go8&$8*3*3h0k(2)2'
+    _API_BASE = 'http://music.163.com/api/'
+
+    @classmethod
+    def _encrypt(cls, dfsid):
+        salt_bytes = bytearray(str(cls._NETEASE_SALT))
+        string_bytes = bytearray(str(dfsid))
+        salt_len = len(salt_bytes)
+        for i in xrange(len(string_bytes)):
```
yan12125: ```xrange``` is not defined in Python 3.x

youtube_dl/extractor/neteasemusic.py
```diff
@@ -0,0 +1,439 @@
...
+                'url': 'http://m1.music.126.net/%s/%s.%s' %
+                       (cls._encrypt(details['dfsId']), details['dfsId'],
+                        details['extension']),
+                'ext': details['extension'],
+                'abr': details['bitrate'] / 1000,
+                'preference': details['bitrate'],
+                'format_id': song_format,
+                'filesize': details['size'],
+                'asr': details['sr']
+            })
```
yan12125: Fields other than ```url``` are optional. Use things like ```details.get('XXX')``` to prevent exceptions if any of them is unavailable.

youtube_dl/extractor/neteasemusic.py
```diff
@@ -0,0 +1,439 @@
...
+            details = info.get(song_format)
+            if not details:
+                continue
+            formats.append({
+                'url': 'http://m1.music.126.net/%s/%s.%s' %
+                       (cls._encrypt(details['dfsId']), details['dfsId'],
+                        details['extension']),
+                'ext': details['extension'],
+                'abr': details['bitrate'] / 1000,
+                'preference': details['bitrate'],
```
yan12125: [Total bitrate is already included as a criterion in format sorting](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L756). If there's no special need, leaving this field blank is fine.

youtube_dl/extractor/neteasemusic.py
```diff
@@ -0,0 +1,439 @@
...
+    @classmethod
+    def _encrypt(cls, dfsid):
+        salt_bytes = bytearray(str(cls._NETEASE_SALT))
+        string_bytes = bytearray(str(dfsid))
+        salt_len = len(salt_bytes)
+        for i in xrange(len(string_bytes)):
+            string_bytes[i] = string_bytes[i] ^ salt_bytes[i % salt_len]
+        m = md5()
+        m.update(string_bytes)
+        result = m.digest().encode('base64')[:-1]
```
yan12125: It might be better to use ```base64.b64encode()``` here as that function does not append an extra ```\n```

youtube_dl/extractor/neteasemusic.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+class NetEaseMusicBaseIE(InfoExtractor):
+    _FORMATS = ['bMusic', 'mMusic', 'hMusic']
+    _NETEASE_SALT = '3go8&$8*3*3h0k(2)2'
+    _API_BASE = 'http://music.163.com/api/'
+
+    @classmethod
+    def _encrypt(cls, dfsid):
+        salt_bytes = bytearray(str(cls._NETEASE_SALT))
```
yan12125: In Python 3.x ```bytearray``` accepts only bytes-like objects, and strings are not.

youtube_dl/extractor/neteasemusic.py
```diff
@@ -0,0 +1,439 @@
...
+        self._sort_formats(formats)
+
+        lyrics_info = self.query_api(
+            'song/lyric?id=%s&lv=-1&tv=-1' % song_id, 
+            song_id, 'Downloading lyrics data')
+        lyrics = self._process_lyrics(lyrics_info)
+
+        alt_title = None
+        if info.get('alias'):
+            alt_title = '/'.join(info.get('alias'))
```
yan12125: In all 3 test cases ```info.get('alias')``` is an empty array. Do you have a test case with this field available? Or do you have (official or unofficial) documents describing the meaning of ```alias```? I'm not sure whether it's proper to use ```alias``` as the alternative title.

youtube_dl/extractor/neteasemusic.py
```diff
@@ -0,0 +1,439 @@
...
+    def _real_extract(self, url):
+        mv_id = self._match_id(url)
+
+        info = self.query_api(
+            'mv/detail?id=%s&type=mp4' % mv_id,
+            mv_id, 'Downloading mv info')['data']
+
+        formats = [
+            {'url': mv_url, 'ext': 'mp4', 'format_id': '%sp' % brs, 'preference': int(brs)}
+            for brs, mv_url in info['brs'].items()
```
yan12125: Seems keys in ```info['brs']``` are respective heights of each format. If so, using the field ```height``` is more comprehensive. And there's no need to keep the ```preference``` field anymore.

youtube_dl/extractor/neteasemusic.py
ping: The problem is ``int_or_none`` doesn't implement rounding.
```
milliseconds_example = 1434179341780
print int_or_none(milliseconds_example, scale=1000)			# returns 1434179341
print self.convert_milliseconds(milliseconds_example)		# returns 1434179342
```
This difference can be seen with the timestamps in NetEaseMusicProgramIE.

youtube_dl/extractor/neteasemusic.py
ping: I may have made a mistake. Switched to using ``info.get('transNames')`` for alternative title. Also added a test case. ``transNames`` is the translated title.

youtube_dl/extractor/neteasemusic.py
ping: Changed.

youtube_dl/extractor/neteasemusic.py
ping: Switched to using ``range``.

youtube_dl/extractor/neteasemusic.py
ping: Changed.

youtube_dl/extractor/neteasemusic.py
ping: Removed ``preference``.

youtube_dl/extractor/neteasemusic.py
ping: Changed.

youtube_dl/extractor/neteasemusic.py
ping: Changed.

youtube_dl/extractor/neteasemusic.py
ping: Removed ``preference`` and added ``height`` instead.

ping:
Made the review changes except for the ``int_or_none`` suggestion because of rounding requirements.

yan12125:
Thanks. Merged.

PR #6053, opened 2015-06-23
Save webpage_url to 'copyright' metadata field as well; 'purl' is flv-only

atomicdryad:
youtube-dl --add-metadata fails to save webpage_url in all cases except .flv. Since
'copyright' is unused, yet supported in all formats that support metadata, this makes
use of it. Example of issue from video saved with varying combinations of -f and -x,
(with this patch applied)

for i in *;do ext=${i##*.};mediainfo -f "$i"|egrep '://.+bbc'|sed "s/^/$ext  /";done

flv  copyright                        : http://www.bbc.com/news/world-asia-33179629
flv  purl                             : http://www.bbc.com/news/world-asia-33179629
m4a  Copyright                        : http://www.bbc.com/news/world-asia-33179629
mp3  Copyright                        : http://www.bbc.com/news/world-asia-33179629
mp4  Copyright                        : http://www.bbc.com/news/world-asia-33179629
wav  Copyright                        : http://www.bbc.com/news/world-asia-33179629

More info at http://wiki.multimedia.cx/index.php?title=FFmpeg_Metadata

Another option is to append the url to 'description' however this will be truncated
in some cases due to desc length and media format, and hard to programmatically extract
in others, due to urls in the desc itself.

### DIFF
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -364,6 +364,7 @@ class FFmpegMetadataPP(FFmpegPostProcessor):
             metadata['comment'] = info['description']
         if info.get('webpage_url') is not None:
             metadata['purl'] = info['webpage_url']
+            metadata['copyright'] = info['webpage_url']
         if info.get('album') is not None:
             metadata['album'] = info['album']
 


jaimeMF:
I don't think that using the url in the copyright is right.

atomicdryad:
It's incredibly useful to store the origin url, as opposed to a blank field. Unfortunately there's no 'url' in mp4 or any of the other formats; any idea of where it should go?

jaimeMF:
If there's no field for the url, I don't think we should use another one. Note that you can use `--xattrs ` and the url will be saved in the xattrs (or use `--write-info-json`, but that will use an additional file).

atomicdryad:
xattrs and info json aren't quite the same; I'd like to tag the file itself with it's origin. However, while I don't see what's wrong with making use of the copyright field, I'll admit it may be too kludgy for some.

PR #6064 seems a better alternative for this issue and an overall improvement as far as output customization.

jaimeMF:
My point is that it doesn't seem to be an appropiate field for that, so we shouln't use another one (at least by default).

Can this be closed then?

PR #6064, opened 2015-06-24
--custom-meta option to write specific information to tags

atomicdryad:
This is added after the defaults provided by --add-metadata

Syntax: --custom-meta "TAGNAME=string with optional %(format)s"
Example: --custom-meta "comment=%(webpage_url)s\n%(description)s"

This can be invoked multiple times for different tags.

Example 2:
 youtube-dl --add-metadata -f 18 --meta 'artist=%(uploader_id)s' --meta 'comment=url: %(webpage_url)s\nviews: %(view_count)s / Rating: %(average_rating)s\n\n%(description)s' https://www.youtube.com/watch?v=WzhW20hLp6M
(snip)
[ffmpeg] Adding metadata to 'Breaking Bad Remix (Seasons 3-5)-WzhW20hLp6M.mp4'
 mp4info Breaking\ Bad\ Remix\ (Seasons\ 3-5)-WzhW20hLp6M.mp4
(snip)
 Artist: uploader: placeboing
 Comments: url: https://www.youtube.com/watch?v=WzhW20hLp6M
views: 5601679 / Rating: 4.97684526443

Music and remixing by me.
Seasons 1 and 2 remix: http://www.youtube.com/watch?v=WsqdmqRgrIc
MP3 download: https://www.dropbox.com/s/39luqbbi5cu36du/Chris%20Lohr%20-%20Breaking%20Bad%20Remix%20seasons%203-5.mp3

### DIFF
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -4,7 +4,7 @@ import io
 import os
 import subprocess
 import time
-
+import collections
 
 from .common import AudioConversionError, PostProcessor
 
@@ -368,8 +368,8 @@ class FFmpegMetadataPP(FFmpegPostProcessor):
             metadata['album'] = info['album']
 
         for m in self._downloader.params.get('custommeta'):
-            key,val = m.split('=',1)
-            metadata[key] = val.replace('\\n', '\n') % info
+            key, val = m.split('=', 1)
+            metadata[key] = val.replace('\\n', '\n') % collections.defaultdict(lambda: 'NA', info)
 
         if not metadata:
             self._downloader.to_screen('[ffmpeg] There isn\'t any metadata to add')
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -369,7 +369,7 @@ class FFmpegMetadataPP(FFmpegPostProcessor):
 
         for m in self._downloader.params.get('custommeta'):
             key, val = m.split('=', 1)
-            metadata[key] = val.replace('\\n', '\n') % collections.defaultdict(lambda: 'NA', info)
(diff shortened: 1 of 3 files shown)


youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -367,6 +367,10 @@ def run(self, info):
         if info.get('album') is not None:
             metadata['album'] = info['album']
 
+        for m in self._downloader.params.get('custommeta'):
+            key,val = m.split('=',1)
+            metadata[key] = val.replace('\\n', '\n') % info
```
dstftw: Missing fields should be raplaced with `NA` similarly to [output template](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L568).

dstftw:
The only problem with this is that `--custom-meta` doesn't work without `--add-metadata` that is not clear from the command description. Moreover, I guess it's even more reasonable to separate those two commands so that `--add-metadata` retained it's original semantics as *metadata-by-default*, and `--custom-meta` (I would prefer `--add-custom-metadata` in this case) was treated as a way to add arbitrary metadata without default.

atomicdryad:
Sorry, missed the comment.

As is --(add--)custom-meta effectively activates --add-metadata by adding the postprocessor. The defaults get added, then the custom stuff overrides. I see two potential improvements;

* Explaining 'this activates --add-metadata' in --help

or

* *not* adding default metadata if --add-custom-metadata is used without --add-metadata (and explaining --add-metadata is required if user wants them). This would entail add_metadata going into ydl_opts.

I can see the advantages of the latter but can't think of a use-case, personally.

atomicdryad:
Waitasec, you're right it does squat without --add-metadata; I forgot to cherrypick the commit that did...

atomicdryad:
commit 699a91a can be merged or skipped; if merged --custom-metadata will *not* add defaults without add-metadata

Issue #6066, opened 2015-06-24
[noco.tv] TypeError: Can't convert 'int' object to str implicitly


mitsukarenai:
When a video has only a number as title, it's interpreted as integer instead of string and leads to TypeError.

Example: http://noco.tv/emission/6158/nolife/retro-magic/218-1941
```
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['-v', '-u', 'PRIVATE', '-p', 'PRIVATE', 'http://noco.tv/emission/6158/nolife/retro-magic/218-1941']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.06.15
[debug] Python version 3.4.3 - Linux-4.0.5-1-ARCH-x86_64-with-arch
[debug] exe versions: ffmpeg 2.7.1, ffprobe 2.7.1, rtmpdump 2.4
[debug] Proxy map: {}
[Noco] Logging in as mitsukarenai
[Noco] 6158: Downloading video JSON
[Noco] 6158: Downloading show JSON
[Noco] 6158: Downloading user options JSON
[Noco] 6158: Downloading qualities JSON
[Noco] 6158: Downloading audio-fr_sub-none_LQ video JSON
[Noco] 6158: Downloading audio-fr_sub-none_HQ video JSON
[Noco] 6158: Downloading audio-fr_sub-none_TV video JSON
Traceback (most recent call last):
  File "/usr/lib/python3.4/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.4/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/bin/youtube-dl/__main__.py", line 19, in <module>
  File "/usr/local/bin/youtube-dl/youtube_dl/__init__.py", line 405, in main
  File "/usr/local/bin/youtube-dl/youtube_dl/__init__.py", line 395, in _real_main
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 1502, in download
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 650, in extract_info
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 273, in extract
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/noco.py", line 198, in _real_extract
TypeError: Can't convert 'int' object to str implicitly
[...]

PR #6067, opened 2015-06-24
[noco.tv] Fix issue #6066: title interpreted as integer

mitsukarenai:
Fixes error using `compat_str()` when the title is only a number (interpreted as integer instead of string):
`TypeError: Can't convert 'int' object to str implicitly`
More details in issue


dstftw:
Thanks.

Issue #5946, opened 2015-06-10
Geek and Sundry


Frenzie:
I don't have time to further look into this myself right now, but here's some basic info I gathered for future reference.

Regular page URL:
http://geekandsundry.com/titansgrave-chapter-0/

Relevant HTML:
```html
    <div class="video-wrapper">
        <div class="video-inner-wrapper container">
            
                        <video
                id="video_4279049078001"
                data-account="4036320279001"
                data-player="5d112ed9-283f-485f-a7f9-33f42e8bc042"
                data-embed="default"
                data-video-id="4279049078001"
                class="video-js"
                autoplay="true"
                controls>
            </video>
```

Relevant streaming playlist link used by the Flash player (`videoId` is `data-video-id` and `pubId` is `data-account`): `http://c.brightcove.com/services/mobile/streaming/index/master.m3u8?videoId=4279049078001&pubId=4036320279001`

ciferkey:
Here's my stab at it: https://github.com/rg3/youtube-dl/pull/5965

dstftw:
New Brightcover embeds will be supported in the next version of youtube-dl.

PR #6074, opened 2015-06-24
[geeksandsundry] Add new extractor

ciferkey:
For issue #5946

### DIFF
--- a/youtube_dl/extractor/geeksandsundry.py
+++ b/youtube_dl/extractor/geeksandsundry.py
@@ -1,0 +1,38 @@
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+import re
+
+
+class GeeksAndSundryIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?geekandsundry\.com/(?P<title>.+)'
+    _TEST = {
+        'url': 'http://www.geekandsundry.com/tabletop-bonus-wils-final-thoughts-on-dread/',
+        'md5': '02206df2e7a1805349a75af8df396222',
+        'info_dict': {
+            'id': 'tabletop-bonus-wils-final-thoughts-on-dread/',
+            'ext': 'mp4',
+            'title': 'TableTop Bonus! Wil\u2019s Final Thoughts on Dread | Geek and Sundry',
+        }
+    }
+
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        page = mobj.group('title')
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/geeksandsundry.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        page = mobj.group('title')
+        webpage_url = "https://geekandsundry.com/" + page
+        webpage = self._download_webpage(webpage_url, page)
+
+        self.report_extraction(page)
+
+        video_id = self._html_search_regex(r'data-video-id=\"(\d+)\"', webpage, u'video id')
```
yan12125: There's no need to use ```u``` prefix given ```unicode_literals``` is declared.

youtube_dl/extractor/geeksandsundry.py
```diff
@@ -0,0 +1,38 @@
...
+            'id': 'tabletop-bonus-wils-final-thoughts-on-dread/',
+            'ext': 'mp4',
+            'title': 'TableTop Bonus! Wil\u2019s Final Thoughts on Dread | Geek and Sundry',
+        }
+    }
+
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        page = mobj.group('title')
```
yan12125: Usually variables named ```page``` are for storing downloaded data (webpage, API page, etc.). This variable is usually named as ```video_id```, ```media_id```, etc. Also, it's better to use ```_match_id``` function here.

youtube_dl/extractor/geeksandsundry.py
```diff
@@ -0,0 +1,38 @@
...
+        webpage_url = "https://geekandsundry.com/" + page
+        webpage = self._download_webpage(webpage_url, page)
+
+        self.report_extraction(page)
+
+        video_id = self._html_search_regex(r'data-video-id=\"(\d+)\"', webpage, u'video id')
+        pub_id = self._html_search_regex(r'data-account=\"(\d+)\"', webpage, u'pub id')
+
+        video_url = "http://c.brightcove.com/services/mobile/streaming/index/master.m3u8?videoId=%s&pubId=%s" % (video_id, pub_id)
```
yan12125: It's a meta M3U8. It's better to extract formats in it via ```_extract_m3u8_formats``` function.

dstftw:
This is now covered by generic support for new Brightcove embeds.

Issue #5289, opened 2015-03-26
Travis CI tests exceed 50 minute


yan12125:
References:
[1] https://travis-ci.org/rg3/youtube-dl/builds/55855866
[2] http://docs.travis-ci.com/user/build-configuration/#Job-Timeouts

Travis CI has a hard limit 50 minutes for tests. Lots of recent builds failed because of that. As a result, I can't see the log of failed tests on Travis CI. A possible solution is to run tests in parallel. At least tests for different sites does not depends on each other. Any ideas?

hlintala:
Another possible solution is to split travis build jobs. Also, if log is longer than 10000 lines, produced output log contains only first 10000 lines. This splitting would also solve that problem.

yan12125:
I've identified 3 problematic sites. They cause timeouts from my location.

* http://planetaplay.com/ (Both ping and curl get timed out)
* http://quickvid.org/ (Both ping and curl get timed out)
* http://vube.com/ (Responds to ping while curl gets timed out)

In #6081 I commented out these tests. If someone can confirm any of these sites are not online anymore, please remove the extractor.

@hlintala It's also a good idea, while for now I think commenting out problematic sites is sufficient.

jaimeMF:
If we run multiple tests in parallel we'll need to tweak the `--process-timeout` option for nosetests, I think that the default value is too small for a lot of the tests.

I haven't looked into the travis tests for a while, maybe it would be worth to limit it to the tests in `make offlinetest` and a few extractors like YouTube, Vimeo and Dailymotion, we could even go crazy and limit the tests to the extractors changed in the commit.

yan12125:
Personally I prefer to keep all tests. I think codes that are rarely used should be tested, too.

d4t4king:
Is there any documentation stating this limitation, and/or the reasoning for this limitation?  A "fast" build for my project would be around 90 minutes.  Given the virtualization overhead, etc., I'd expect it to be slightly longer.  Will Travis-CI simply not work for me?

yan12125:
[This page](http://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts) says the limit is 50 minutes. However, seems builds longer than 50 minutes are allowed (for example [this build](https://travis-ci.org/rg3/youtube-dl/builds/79866418) ) since switching to the new infrastructure in f3a65d9636908ee49ff3d50c24efb8067caa32c6, so I'm closing this issue now.

PR #6081, opened 2015-06-25
[planetaplay/quickvid/vube] Skip inaccessible sites

yan12125:
Try to solve #5289. In not a long term solution but practical for now.


yan12125:
The travis CI test goes with much shorter time. I'll merge it if there are no other opinions.

PR #6094, opened 2015-06-26
Added a Playlist Info Extractor for WebOfStories

dufferzafar:
I've tested this by downloading the entire [Donald Knuth](http://www.webofstories.com/playAll/donald.knuth) playlist :smile_cat:

### DIFF
--- a/youtube_dl/extractor/webofstories.py
+++ b/youtube_dl/extractor/webofstories.py
@@ -1,6 +1,8 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
+import re
+
 from .common import InfoExtractor
 from ..utils import int_or_none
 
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/webofstories.py
```diff
@@ -98,3 +100,26 @@ def _real_extract(self, url):
             'description': description,
             'duration': duration,
         }
+
+
+class WebOfStoriesPlaylistIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?webofstories\.com/playAll/(?P<id>[^/]+)'
+    _TESTS = []
```
dstftw: Can you add a test?

dufferzafar: I could definitely try, but can you please tell me how I can get only a single test to run. I don't want to run all other tests while I'm working on this one.

dstftw: Item 6 of [this instruction](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site).

PR #6096, opened 2015-06-26
[snagfilms] Add new extractor

### LINKED ISSUE #6007: Add support for snagfilms.com
Support for snagfilms.com would be very welcome.

remitamine:
it should closes #6007

### DIFF
--- a/youtube_dl/extractor/snagfilms.py
+++ b/youtube_dl/extractor/snagfilms.py
@@ -1,10 +1,10 @@
+from re import match,DOTALL
 from .common import InfoExtractor
 from ..utils import js_to_json
-from re import DOTALL
 
 class SnagFilmsIE(InfoExtractor):
-    _VALID_URL = r'(?:https?://)?(?:www.)?snagfilms\.com/films/title/(?P<display_id>.+?)(?:/|$)'
-    _TEST = {
+    _VALID_URL = r'(?:https?://)?(?:www.|embed.)?snagfilms\.com/(?:films/title/(?P<display_id>.+?)|embed/player\?.*filmId=(?P<id>.+?))(?:&|/|$)'
+    _TESTS = [{
         'url': 'http://www.snagfilms.com/films/title/lost_for_life',
         'info_dict':
         {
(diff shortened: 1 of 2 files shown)


dstftw:
We do not use tabs.

youtube_dl/extractor/snagfilms.py
```diff
@@ -0,0 +1,67 @@
...
+		}
+	}
+
+	def _real_extract(self, url):
+		display_id = self._search_regex(
+			self._VALID_URL,
+			url,
+			'display_id',
+			group='display_id'
+		)
```
dstftw: Uh oh. Just use `self._match_id(url)` and change the symbolic name of regex group to `id`.

youtube_dl/extractor/snagfilms.py
```diff
@@ -0,0 +1,67 @@
...
+
+		json_data = self._parse_json(self._html_search_regex(
+			r'"data":{"film":(?P<data>{.*?}})}',
+			webpage,
+			'data',
+			group='data'
+		), display_id)
+		title = json_data['title']
+		video_id = json_data['id']
+		duration = int(json_data['duration'])
```
dstftw: It should be robust in case of some missing fields.

dstftw:
Generic embeds should be supported as well. Example URL: http://whilewewatch.blogspot.ru/2012/06/whilewewatch-whilewewatch-gripping.html

youtube_dl/extractor/snagfilms.py
```diff
@@ -0,0 +1,67 @@
...
+			'display_id',
+			group='display_id'
+		)
+		webpage = self._download_webpage(url, display_id)
+
+		json_data = self._parse_json(self._html_search_regex(
+			r'"data":{"film":(?P<data>{.*?}})}',
+			webpage,
+			'data',
+			group='data'
```
dstftw: You don't need to specify any symbolic name for a group since you have only one.

youtube_dl/extractor/snagfilms.py
remitamine: i do it like because the acctual id doesn't appear in the url i get it after download the page.
for example the url in the test page:http://www.snagfilms.com/films/title/lost_for_life
the id is not "lost_for_life" it is 0000014c-de2f-d5d6-abcf-ffef58af0017 this what i need to get the embed page that contain the download links

dstftw: It does not matter that it's not an actual id. The code where you extract the actual `video_id` remains the same.

dstftw:
Merged.

dstftw:

remitamine:
my name is Remita Amine.

remitamine:
just asking why not add ```'ie-key' = 'SnagFilmsEmbed'``` to the return of SnagFilmsIE?

PR #6097, opened 2015-06-26
[youtube] Extract formats from multiple DASH manifests (Closes #6093)

### LINKED ISSUE #6093: [youtube] add support for new formats
https://www.youtube.com/watch?v=FIl7x6_3R5Y
youtube-dl misses at least 2 formats on the above video;
810x1440 vp9 30fps DASH video 189.1MB
1080x1920 vp9 30fps DASH video 440.85MB

I've found this using SaveFrom.net helper userscript.
https://sf-addon.com/helper/chrome/helper.user.js

dstftw:
DASH manifest pointed by dashmpd from the video webpage and one pointed by get_video_info may
be different (namely different itag set) - some itags are missing from DASH manifest pointed by
webpage's dashmpd, some - from DASH manifest pointed by get_video_info's dashmpd).
The general idea is to take a union of itags of both DASH manifests (for example video with such
'manifest behavior' see https://github.com/rg3/youtube-dl/issues/6093).

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -518,6 +518,20 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
                 'skip_download': 'requires avconv',
             }
         },
+        # Extraction from multiple DASH manifests (https://github.com/rg3/youtube-dl/pull/6097)
+        {
+            'url': 'https://www.youtube.com/watch?v=FIl7x6_3R5Y',
+            'info_dict': {
+                'id': 'FIl7x6_3R5Y',
+                'ext': 'mp4',
+                'title': 'md5:7b81415841e02ecd4313668cde88737a',
+                'description': 'md5:116377fd2963b81ec4ce64b542173306',
+                'upload_date': '20150625',
+                'uploader_id': 'dorappi2000',
+                'uploader': 'dorappi2000',
+                'formats': 'mincount:33',
+            },
+        }
     ]
 
     def __init__(self, *args, **kwargs):
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -894,27 +894,28 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
                     # Convert to the same format returned by compat_parse_qs
                     video_info = dict((k, [v]) for k, v in args.items())
                     add_dash_mpd(video_info)
-            # We also try looking in get_video_info since it may contain different dashmpd
-            # URL that points to a DASH manifest with possibly different itag set (some itags
-            # are missing from DASH manifest pointed by webpage's dashmpd, some - from DASH
-            # manifest pointed by get_video_info's dashmpd).
-            # The general idea is to take a union of itags of both DASH manifests (for example
-            # video with such 'manifest behavior' see https://github.com/rg3/youtube-dl/issues/6093)
-            self.report_video_info_webpage_download(video_id)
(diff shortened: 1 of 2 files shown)


jaimeMF:
I would prefer to avoid doing an additional network request (potentially 4) by default, it probably makes directly playing the urls (for example with mpv) a bit slower.

dstftw:
In most of the cases there would be only one additional network request. Otherwise, we may loose some formats or even the `bestvideo` if it happens to be served by `get_video_info`'s manifest.

youtube_dl/extractor/youtube.py
```diff
@@ -1118,24 +1131,26 @@ def _map_to_format_list(urlmap):
...
                         'Skipping DASH manifest: %r' % e, video_id)
-                else:
+                if dash_formats:
                     # Remove the formats we found through non-DASH, they
                     # contain less info and it can be wrong, because we use
                     # fixed values (for example the resolution). See
                     # https://github.com/rg3/youtube-dl/issues/5774 for an
                     # example.
-                    dash_keys = set(df['format_id'] for df in dash_formats)
+                    dash_keys = set(df['format_id'] for df in dash_formats.values())
```
jaimeMF: The keys of the `dash_formats` dict are the `format_id` fields themselves, so you could directly `set(dash_formats)` or `set(dash_formats.keys())` to make it clearer.

dstftw: Fixed in d80265c.

yan12125:
Would it be better to skip the additional requests if ```--youtube-skip-dash-manifest``` given? Of course the documentation needs updated if you're going to do that. In case of mpv, the default format is now ```best``` due to mpv-player/mpv#1867. mpv, as well as other players relying youtube-dl, can reduce some network traffic at startup by specifying ```--youtube-skip-dash-manifest``` if they are not interested in video/audio-only formats.

And, I hope there's a test case for this PR.

dstftw:
Since [no DASH manifests downloaded when `--youtube-skip-dash-manifest` is specified](https://github.com/dstftw/youtube-dl/blob/bc93bdb5bbf48b121977a5088ab9607f0fbeb83e/youtube_dl/extractor/youtube.py#L1134), do you mean [not trying to download `get_video_info`](https://github.com/dstftw/youtube-dl/blob/bc93bdb5bbf48b121977a5088ab9607f0fbeb83e/youtube_dl/extractor/youtube.py#L903-L917) when [extraction from the webpage succeeded](https://github.com/dstftw/youtube-dl/blob/bc93bdb5bbf48b121977a5088ab9607f0fbeb83e/youtube_dl/extractor/youtube.py#L893-L896)? Fixing documentation in this case is not enough I guess since no one usually read it and option name (`--youtube-skip-dash-manifest`) clearly states what should it do. Probably some new generic option like `--reduce-traffic` should be introduced which would also be useful for another extractors.

yan12125:
> do you mean not trying to download get_video_info when extraction from the webpage succeeded?

Yes. Sorry that the original statement is not clear enough.

In this PR the additional get_video_info requests are for extracting more DASH urls. When those URLs are not used, I think there's no need to keep the requests. ```--youtube-skip-dash-manifest``` can be interpreted as: skipping requests required for getting valid DASH manifests.

dstftw:
Fixed in 0a3cf9a.

jaimeMF:
Looks good.

dstftw:

yan12125:
I think this PR can be merged.

PR #6098, opened 2015-06-26
[youtube] Pick up codecs info from DASH manifest when not set explicitly

dstftw:


dstftw:

PR #6104, opened 2015-06-27
[moviefap] Add new extractor

gebn:
Extractor for http://moviefap.com/.

Checked with flake8.

### DIFF
--- a/youtube_dl/extractor/moviefap.py
+++ b/youtube_dl/extractor/moviefap.py
@@ -3,7 +3,10 @@ from __future__ import unicode_literals
 import re
 
 from .common import InfoExtractor
-from ..utils import str_to_int
+from ..utils import (
+    xpath_text,
+    str_to_int
+)
 
 
 class MovieFapIE(InfoExtractor):
@@ -82,7 +85,7 @@ class MovieFapIE(InfoExtractor):
             'title': self._html_search_regex(r'<div id="view_title"><h1>(.*?)</h1>', webpage, 'title'),
             'display_id': re.compile(self._VALID_URL).match(url).group('name'),
             'thumbnails': self.__get_thumbnail_data(xml),
-            'thumbnail': xml.find('startThumb').text,
+            'thumbnail': xpath_text(xml, 'startThumb', 'thumbnail'),
             'description': self._html_search_regex(r'name="description" value="(.*?)"', webpage, 'description', fatal=False),
             'uploader_id': self._html_search_regex(r'name="username" value="(.*?)"', webpage, 'uploader_id', fatal=False),
             'view_count': str_to_int(self._html_search_regex(r'<br>Views <strong>([0-9]+)</strong>', webpage, 'view_count, fatal=False')),
@@ -102,7 +105,7 @@ class MovieFapIE(InfoExtractor):
         # work out the video URL(s)
         if xml.find('videoLink') is not None:
             # single format available
-            info['url'] = xml.find('videoLink').text
+            info['url'] = xpath_text(xml, 'videoLink', 'url', True)
         else:
             # multiple formats available
             info['formats'] = []
@@ -110,8 +113,8 @@ class MovieFapIE(InfoExtractor):
             # N.B. formats are already in ascending order of quality
             for item in xml.find('quality').findall('item'):
                 info['formats'].append({
-                    'url': item.find('videoLink').text,
-                    'resolution': item.find('res').text  # 480p etc.
+                    'url': xpath_text(item, 'videoLink', 'url', True),
+                    'resolution': xpath_text(item, 'res', 'resolution', True)  # 480p etc.
                 })
 
         return info
--- a/youtube_dl/extractor/moviefap.py
+++ b/youtube_dl/extractor/moviefap.py
@@ -82,8 +82,36 @@ class MovieFapIE(InfoExtractor):
                 r'flashvars\.config = escape\("(.+?)"', webpage, 'player parameters')
         xml = self._download_xml(info_url, video_id)
 
-        info = {
+        # find the video container
+        if xml.find('videoConfig') is not None:
+            ext = xml.find('videoConfig').find('type').text
+        else:
+            ext = 'flv'  # guess...
+
+        # work out the video URL(s)
+        formats = []
+        if xml.find('videoLink') is not None:
+            # single format available
+            formats.append({
+                'url': xpath_text(xml, 'videoLink', 'url', True),
+                'ext': ext
+            })
+        else:
+            # multiple formats available
+            for item in xml.find('quality').findall('item'):
+                resolution = xpath_text(item, 'res', 'resolution', True)  # 480p etc.
+                formats.append({
+                    'url': xpath_text(item, 'videoLink', 'url', True),
+                    'ext': ext,
+                    'resolution': resolution,
+                    'height': int(re.findall(r'\d+', resolution)[0])
+                })
+
+            self._sort_formats(formats)
+
+        return {
             'id': video_id,
+            'formats': formats,
             'title': self._html_search_regex( \
                     r'<div id="view_title"><h1>(.*?)</h1>', webpage, 'title'),
             'display_id': re.compile(self._VALID_URL).match(url).group('name'),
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/moviefap.py
```diff
@@ -0,0 +1,119 @@
...
+            'id': 'e5da0d3edce5404418f5',
+            'ext': 'flv',
+            'title': 'Jeune Couple Russe',
+            'description': 'Amateur',
+            'thumbnail': 'http://pic.moviefap.com/thumbs/e5/949-18l.jpg',
+            'uploader_id': 'whiskeyjar',
+            'display_id': 'jeune-couple-russe'
+        }
+    }, {
+        'url': 'http://www.moviefap.com/videos/3080837f6712355015c2/busty-british-blonde-takes-backdoor-in-fake-taxi.html',
```
dstftw: This looks like copyrighted content. youtube-dl does not support sites dedicated to [copyright infringement](https://github.com/rg3/youtube-dl#can-you-add-support-for-this-anime-video-site-or-site-which-shows-current-movies-for-free). Do they have rights to distribute this? If not it has to be replaced with one that does not violate any copyrights. Moreover if this test does not test anything special in extraction process it should be removed completely.

gebn: I've swapped the video for another that appears to be copyright-free. Each test checks a different format in which the site provides video URLs - I've now annotated this in the code. Any other comments?

youtube_dl/extractor/moviefap.py
```diff
@@ -0,0 +1,123 @@
...
+        xml = self._download_xml(info_url, video_id)
+
+        # create dictionary of properties we know so far, or can find easily
+        info = {
+            'id': video_id,
+            'title': self._html_search_regex(r'<div id="view_title"><h1>(.*?)</h1>', webpage, 'title'),
+            'display_id': re.compile(self._VALID_URL).match(url).group('name'),
+            'thumbnails': self.__get_thumbnail_data(xml),
+            'thumbnail': xml.find('startThumb').text,
+            'description': self._html_search_regex(r'name="description" value="(.*?)"', webpage, 'description'),
```
dstftw: All the fields except `title` should be passed `default=False` in order not to break extraction if webpage layout for any of these field changes.

youtube_dl/extractor/moviefap.py
```diff
@@ -0,0 +1,123 @@
...
+        # download that XML
+        xml = self._download_xml(info_url, video_id)
+
+        # create dictionary of properties we know so far, or can find easily
+        info = {
+            'id': video_id,
+            'title': self._html_search_regex(r'<div id="view_title"><h1>(.*?)</h1>', webpage, 'title'),
+            'display_id': re.compile(self._VALID_URL).match(url).group('name'),
+            'thumbnails': self.__get_thumbnail_data(xml),
+            'thumbnail': xml.find('startThumb').text,
```
dstftw: Use `utils.xpath_text` instead, again with `fatal=False`.

youtube_dl/extractor/moviefap.py
```diff
@@ -0,0 +1,123 @@
...
+            thumbnails.append({
+                'url': pattern.replace('#', str(i)),
+                'width': width,
+                'height': height
+            })
+        return thumbnails
+
+    def _real_extract(self, url):
+
+        # find the video ID
```
dstftw: I would prefer removing obvious useless comments.

youtube_dl/extractor/moviefap.py
```diff
@@ -0,0 +1,123 @@
...
+        height = str_to_int(timeline.find('imageHeight').text)
+        first = str_to_int(timeline.find('imageFirst').text)
+        last = str_to_int(timeline.find('imageLast').text)
+        pattern = timeline.find('imagePattern').text
+
+        # generate the list of thumbnail information dicts
+        thumbnails = []
+        for i in range(first, last + 1):
+            thumbnails.append({
+                'url': pattern.replace('#', str(i)),
```
dstftw: Use `compat.compat_str` instead.

youtube_dl/extractor/moviefap.py
```diff
@@ -0,0 +1,123 @@
...
+
+        # work out the video URL(s)
+        if xml.find('videoLink') is not None:
+            # single format available
+            info['url'] = xml.find('videoLink').text
+        else:
+            # multiple formats available
+            info['formats'] = []
+
+            # N.B. formats are already in ascending order of quality
```
dstftw: They are in ascending order today. Tomorrow that may change and break the extractor logic. Use generic sorting mechanism: `self._sort_formats`.

youtube_dl/extractor/moviefap.py
gebn: `fatal=False` appears to be the default for that method. Would you still rather it was passed explicitly?

dstftw: No, default is ok. I've just emphasized it should not be fatal in order not to break extraction.

youtube_dl/extractor/moviefap.py
gebn: I think you misunderstood my comment. Each video only has a single format available (e.g. mp4 or flv), however some videos have multiple levels of quality within that format (e.g. 480p and 720p). Is there a similar method to sort quality strings rather than format?

dstftw: format in terms of youtube-dl is a media pointed by URL that has [some metadata that describes it](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L64-L120). `_sort_formats` performs [sorting according this metadata](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L753-L767).

youtube_dl/extractor/moviefap.py
```diff
@@ -0,0 +1,123 @@
...
+            for item in xml.find('quality').findall('item'):
+                info['formats'].append({
+                    'url': item.find('videoLink').text,
+                    'resolution': item.find('res').text  # 480p etc.
```
dstftw: Here you generally have a height that you can put into `height` field and `_sort_formats` will sort according to it.

youtube_dl/extractor/moviefap.py
```diff
@@ -0,0 +1,121 @@
...
+        info_url = self._html_search_regex(r'flashvars\.config = escape\("(.+?)"', webpage, 'player parameters')
+        xml = self._download_xml(info_url, video_id)
+
+        info = {
+            'id': video_id,
+            'title': self._html_search_regex(r'<div id="view_title"><h1>(.*?)</h1>', webpage, 'title'),
+            'display_id': re.compile(self._VALID_URL).match(url).group('name'),
+            'thumbnails': self.__get_thumbnail_data(xml),
+            'thumbnail': xpath_text(xml, 'startThumb', 'thumbnail'),
+            'description': self._html_search_regex(r'name="description" value="(.*?)"', webpage, 'description', fatal=False),
```
dstftw: One more thing, avoid long strings if possible.

youtube_dl/extractor/moviefap.py
gebn: The functionality of `_sort_formats` is very ambiguous. Are you saying I should call `self._sort_formats(info['formats'])` before `return info`? If so, would it be better for youtube-dl to do this automatically rather than delegate responsibility to each extractor?

dstftw: Extractor may need it's own sorting mechanism.

gebn: I don't think `__sort_formats` is helpful due to the extension is always being the same for a given video, and the attributes the site provides. A given video will never be available in flv **and** mp4 format - it's an either/or situation.

What this extractor needs is a generic function that can take a list of formats, e.g. `['480p', '144p', '720p']` and sort it by quality (the result in this case would be `['144p', '480p', '720p']`). I'm happy to write one. It seems to me like `common.py` would be a good place to put it - do you agree?

dstftw: You don't get it. `xpath_text(item, 'res', 'resolution', True)` => `<height>p`. Extract this height, put into format entry along with `url` and `resolution` as `height`. Stuff all formats into `info['formats']`. Run `_sort_formats` on it, it [sorts on height](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L759).

dstftw: Hm, this XML layout reminds me [TNAFlix](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/tnaflix.py) alot.

gebn: Ok I think that's all sorted. Flv can't be simplified as the resolution isn't given at all - the site only provides a URL as there is only one format.

dstftw: Everything apart from `url` is optional.

gebn: Is using `info['formats']` preferred to `info['url']`? Changing to the former doesn't appear to make any difference information-wise.

dstftw: That's just for your info. With this you could build info dict right when return without constructing it by pieces that will result in (imho) more simple and clean code.

gebn: I'm not convinced it makes any difference (I don't like extra variables floating around), but as you think it's more consistent with other extractors that way, I'll change it.

gebn: Committed in 5a9cc19. Flake8 was complaining about my the line breaks, so I fixed them as well in db652ea.

PR #6118, opened 2015-06-28
Update rtlnl.py

corone17:
Better to extract 'http://manifest.us.rtl.nl' from the json, I'd say. And I think it's better to use the default json-url to make it more futureproof.
Succesfully tested with tarball.


dstftw:
Thanks.

PR #6133, opened 2015-06-30
Moved brew install instructions up as this is easiest method imo

robhuzzey:
Trivial change but I thought it may help as I was about to use the curl method until I scrolled & saw brew option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,14 @@ youtube-dl - download videos from youtube.com or other video platforms
 
 # INSTALLATION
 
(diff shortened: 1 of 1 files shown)


dstftw:
I don't think that's a good idea since it may be outdated despite we file a PR every new release to them. I see it very rational to put the original links on the very to as the most relevant and new version.

robhuzzey:
@dstftw ok, now worries... I wasn't aware it may not be the latest on brew... thanks anyway & keep up the good work :)

PR #6144, opened 2015-07-03
HLS WebVTT subtitle downloading support

### LINKED ISSUE #6106: ComCarCoff subtitle support
JSON dumps of CICGC URLs already include a link to sliced English subtitles in WebVTT format that could easily  be downloaded using ffmpeg. It would be nice if the ComCarCoff extractor was able to detect these.

fstirlitz:
This should resolve #6106.

The patches are fairly invasive. Since downloading HLS WebVTT subtitles seems to be relatively heavyweight bandwidth-wise, downloading them right away in the extractor is out of the question; but there was no other way to download these subtitles in an offline-viewable form. Hence the subtitle downloading code was hacked to make use of the downloader infrastructure and a custom HLS WebVTT downloader was written.

The code was tested with Python 3.4.3 and 2.7.10 on a few ComCarCoff videos.

### DIFF
--- a/youtube_dl/downloader/hls.py
+++ b/youtube_dl/downloader/hls.py
@@ -208,3 +208,153 @@ class HlsFD(FragmentFD):
         self._finish_frag_download(ctx)
 
         return True
+
+
+class WebVttHlsFD(FragmentFD):
+    """ A downloader for HLS WebVTT subtitles. """
+    FD_NAME = 'hlswebvtt'
+
+    @staticmethod
+    def _parse_ts(ts):
+        m = re.match('(?:(?:([0-9]+):)?([0-9]+):)?([0-9]+)(?:\.([0-9]+))?', ts)
+        hrs, min, sec, msc = m.groups()
+        return 90 * (
+            int(hrs or 0) * 3600000 +
+            int(min or 0) *   60000 +
+            int(sec or 0) *    1000 +
+            int(msc or 0)
+        )
+
+    @staticmethod
+    def _format_ts(ts):
+        ts  = int(ts / 90)
+        hrs = ts / 3600000
+        ts %=      3600000
+        min = ts /   60000
+        ts %=        60000
+        sec = ts /    1000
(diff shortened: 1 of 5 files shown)


remitamine:
i want you to know that in this commit [avformat/hlsenc: Add WebVtt support in hls](http://git.videolan.org/?p=ffmpeg.git;a=commit;h=379ddc6af44f0c2f3cb517bcc5664610aef9bfc0) has been added to ffmpeg

fstirlitz:
libavcodec still doesn't have any support for `X-TIMESTAMP-MAP` in WebVTT, so it's useless here. Extractors can be systematically modified to use the new method; I retained the old API only to avoid having to modify everything at once and make the patches easier to review. It can be done either way, I don't care.

Anyway, developers: is this any good? Should I adapt this to take advantage of #6392? @dstftw, @jaimeMF?

youtube_dl/downloader/hls.py
```diff
@@ -102,3 +102,138 @@ def real_download(self, filename, info_dict):
...
+        return '%02u:%02u:%02u.%03u' % (hrs, min, sec, ts)
+
+    def real_download(self, filename, info_dict):
+        url = info_dict['url']
+        tmpfilename = self.temp_name(filename)
+
+        self.to_screen(
+            '[hlswebvtt] %s: Downloading manifest' %
+            (info_dict['id']))
+        data = self.ydl.urlopen(url).read()
```
Kagami: This won't work in case of live HLS WebVTT streams because you constantly get new subtitle segments at the same playlist URL.

It's a shame X-TIMESTAMP-MAP support patch hasn't been merged to ffmpeg yet after 3 years, but in my use-case (vlive.tv) it's not required, so dumping HLS WebVTT via ffmpeg works quite good.

jsundram:
This code is awesome -- I had just started writing something just like this -- is there a reason it's not in master?

fstirlitz:
You tell me.

PR #6147, opened 2015-07-04
Update for the wdr extractor

JB26:
I updated the wdr extractor to include .mp4 files.
The tests are also updated.


youtube_dl/extractor/wdr.py
```diff
@@ -16,59 +17,45 @@
 
 
 class WDRIE(InfoExtractor):
-    _PLAYER_REGEX = '-(?:video|audio)player(?:_size-[LMS])?'
-    _VALID_URL = r'(?P<url>https?://www\d?\.(?:wdr\d?|funkhauseuropa)\.de/)(?P<id>.+?)(?P<player>%s)?\.html' % _PLAYER_REGEX
```
dstftw: Now it does not handle URLs like `http://www1.wdr.de/mediathek/video/sendungen/hier_und_heute/videostreetfoodpioniere100-videoplayer_size-L.html`. `(?P<player>%s)?\.html' % _PLAYER_REGEX` was on purpose.

youtube_dl/extractor/wdr.py
```diff
@@ -16,59 +17,45 @@
...
-
+    _PLAYER_REGEX = 'https?://deviceids-medstdp.wdr.de/ondemand/.+?/.+?\.js'
```
dstftw: It would be more reasonable to extract mp4 using the webpage from old approach (e.g. `http://www1.wdr.de/mediathek/video/sendungen/hier_und_heute/videostreetfoodpioniere100-videoplayer_size-L.html`) since it contains different mp4 qualities including one in this JSON. Moreover there is an m3u8 that should be extracted as well,

youtube_dl/extractor/wdr.py
```diff
@@ -95,80 +82,87 @@ class WDRIE(InfoExtractor):
...
-                self.url_result(page_url + href, 'WDR')
-                for href in re.findall(r'<a href="/?(.+?%s\.html)" rel="nofollow"' % self._PLAYER_REGEX, webpage)
-            ]
+        if entries is None:  # Overview page
+            return self._overiew_page_extractor(page_url, page_id, webpage)
 
-            if entries:  # Playlist page
-                return self.playlist_result(entries, page_id)
+        jsonpage = self._download_webpage(entries.group(0), entries.group(0))
+        jsonvars = json.loads(jsonpage[38:-2])
```
dstftw: That's very brittle.

youtube_dl/extractor/wdr.py
```diff
@@ -95,80 +82,87 @@ class WDRIE(InfoExtractor):
...
+                ext = fmt['ext']
+            else:
+                ext = determine_ext(video_url)
 
-        page_id = flashvars['trackerClipId'][0]
-        video_url = flashvars['dslSrc'][0]
-        title = flashvars['trackerClipTitle'][0]
-        thumbnail = flashvars['startPicture'][0] if 'startPicture' in flashvars else None
-        is_live = flashvars.get('isLive', ['0'])[0] == '1'
+            formats.append({'url': video_url, 'ext': ext, 'format_id': _id})
```
dstftw: Extraction from JSON may result in duplicate download links like in your audio test URL.

JB26:
The links in the …-videoplayer_size-L.html don't work for me and I didn't realize that they just need a change in the url to get them working. The new commit uses the old method and also extracts the m3u8 link.

youtube_dl/extractor/wdr.py
```diff
@@ -21,54 +21,41 @@ class WDRIE(InfoExtractor):
 
     _TESTS = [
         {
-            'url': 'http://www1.wdr.de/mediathek/video/sendungen/servicezeit/videoservicezeit560-videoplayer_size-L.html',
+            'url': 'http://www1.wdr.de/mediathek/video/sendungen/hier_und_heute/videostreetfoodpioniere100.html',
```
dstftw: The semantics of tests should be kept. This test should test `*-videoplayer_size-[LMS].html` URL. Same for others: `*-videoplayer.html`, `*-audioplayer.html` (on two different domains).

youtube_dl/extractor/wdr.py
```diff
@@ -163,12 +151,35 @@ def _real_extract(self, url):
...
             ext = determine_ext(video_url)
 
-        description = self._html_search_meta('Description', webpage, 'description')
+        formats.append({'ext': ext, 'url': video_url})
+
+        m3u8_url = re.search(r'<li>\n<a rel="adaptiv" type="application/vnd\.apple\.mpegURL" href="(?P<link>.+?)"', webpage)
+
+        if m3u8_url is not None:
+            m3u8_url = m3u8_url.group('link')
+            formats.append({'ext': 'm3u8', 'url': m3u8_url})
```
dstftw: This should be processed with `self._extract_m3u8_formats` instead.

youtube_dl/extractor/wdr.py
```diff
@@ -163,12 +151,35 @@ def _real_extract(self, url):
...
+
+        webL_quality = -1
+        for video_vars in re.findall(r'<li>\n<a rel="(?P<format_id>web.?)"  href=".+?/(?P<link>fsk.+?)"', webpage):
```
dstftw: Why not just extract the whole `href` and use it as final URL?

youtube_dl/extractor/wdr.py
```diff
@@ -163,12 +151,35 @@ def _real_extract(self, url):
...
+            format_id = video_vars[0]
+            video_url = 'http://ondemand-ww.wdr.de/medstdp/' + video_vars[1]
+            ext = determine_ext(video_url)
+            if format_id == 'webL':
+                quality = webL_quality
+                webL_quality -= 1
+            if format_id == 'webM':
+                quality = -3
+            if format_id == 'webS':
+                quality = -4
```
dstftw: [Use `utils.qualities` instead](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/npo.py#L123).

youtube_dl/extractor/wdr.py
JB26: Doesn't it make more sense to test the url from where `*-videoplayer_size-[LMS].html` gets extracted? That way the link extraction for `*-videoplayer_size-[LMS].html` and the media extraction are getting tested.

dstftw: No. With `*-videoplayer_size-[LMS].html` a single media file URL is tested, otherwise - a playlist.

JB26: I see the problem. The page I test only contains a single video and shouldn't return a playlist. I tried to separate the media extrator and playlist extractor to change that. Still in the same class because there doesn't seem to be any indicator in the url that the current page is a playlist.

youtube_dl/extractor/wdr.py
JB26: Done

youtube_dl/extractor/wdr.py
JB26: Using the whole `href` returns an error page (at least for me) that tells you to install flashplayer.

youtube_dl/extractor/wdr.py
JB26: Done

youtube_dl/extractor/wdr.py
dstftw: Works perfectly for me. What are the URLs from `href` for you?

JB26: I get this one for the url in the first test:

http://mobile-ondemand.wdr.de/CMS2010/mdb/ondemand/weltweit/fsk0/75/750693/750693_8074409.mp4

Tested in firefox (without flash), chrome (with flash), wget curl, and youtube-dl

dstftw: This one doesn't work for me either. What is the original URL?

JB26: http://www1.wdr.de/mediathek/video/sendungen/hier_und_heute/videostreetfoodpioniere100.html

dstftw: Weird. Yesterday it was working:
```
python -m youtube_dl http://mobile-ondemand.wdr.de/CMS2010/mdb/ondemand/weltweit/fsk0/75/752868/752868_8108527.mp4
[download] Destination: 8108527-752868.mp4
[download]   1.4% of 291.34MiB at  8.53MiB/s ETA 00:33
ERROR: Interrupted by user
```
But not now.

JB26: Maybe a caching issue or the server for `mobile-ondemand.wdr.de` is instable?
I found a post on the german [ubuntuusers forum](https://forum.ubuntuusers.de/post/6503252/) from april 2014 that reports problems with `http://mobile-ondemand.wdr.de/CMS2010/mdb/ondemand/…` links.
The url version I use is also used in the html5 player on wdr.

JB26:
Pages with only a single media item are not returned as a playlist anymore.
There doesn't seem to be any indicator in the url if the page is a playlist or not. So I can't use different classes.
It would be possible to use the `-videoplayer.html` part of the url as an indicator. But in my opinion most users will probably input the url without the `-videoplayer.html` part and don't expect a playlist if the page only contained a single video.
At least that's what confused me.

yan12125:
Superseded by #8843.

PR #6179, opened 2015-07-08
Added Biqle.ru extractor.

cewbost:
Added an extractor. That is all.


dstftw:
Instead of reinventing the wheel (i.e. reimplementing vk extractor) you could just extend [it's _VALID_URL](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/vk.py#L24).

PR #6188, opened 2015-07-10
[efukt] Add new extractor

Hazmo:
New extractor for efukt.com

Checked with flake8

### DIFF
--- a/youtube_dl/extractor/efukt.py
+++ b/youtube_dl/extractor/efukt.py
@@ -1,0 +1,51 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import str_to_int
+
+import re
+
+
+class EfuktIE(InfoExtractor):
+    _VALID_URL = r'(?:http://)?(?:www\.)?efukt\.com/(?P<id>[0-9]+)(?:_[^_.]+)+\.html'
+    _TEST = {
+        'url': 'http://efukt.com/21269_Epic_2_Horrible_in_Thirty_Seconds!.html',
+        'md5': 'f07f2f58ae29dc01ce7351e4e825797f',
+        'info_dict': {
+            'id': '21269',
+            'ext': 'mp4',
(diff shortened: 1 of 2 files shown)


dstftw:
Generic extractor works with efukt just fine. What's the rationale apart from additional metadata?

youtube_dl/extractor/efukt.py
```diff
@@ -0,0 +1,51 @@
...
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        if(re.match(r'http://', url) is None):
+            url = 'http://' + url
+
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._html_search_meta('og:title', webpage, 'title')
+        video_url = self._html_search_regex(r'file:\s?"(?P<video_url>(?:http://)?(?:www\.)?assets\.efukt\.com/videos/(?:[^.]+)\.(mp4|flv))"', webpage, 'video_url')
```
dstftw: Here you expect `mp4` or `flv` but [`ext` is always `mp4`](https://github.com/Hazmo/youtube-dl/blob/efukt/youtube_dl/extractor/efukt.py#L48).

youtube_dl/extractor/efukt.py
```diff
@@ -0,0 +1,51 @@
...
+            'ext': 'mp4',
+            'thumbnail': 'http://assets.efukt.com/player-preview/21269.mp4.jpg',
+            'format': 'mp4',
+            'title': 'Epic 2 Horrible in Thirty Seconds!',
+            'description': "She pretty much aces the whole 'prostitute on camera' thing except for one small detail: This newbie pornstar's lady cum has a really unique... consistency. LOLOLOLOLOLOLOLOLOLOL. Full Vid HERE."
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
```
dstftw: You don't need this. This is done behind the scenes if `_VALID_URL` requires URL scheme name.

PR #6196, opened 2015-07-11
[rdsca] New extractor

### LINKED ISSUE #1611: Add support for rds.ca
Hi,

Could you add support for http://www.rds.ca ? 
Thanks.

Example url:

$ youtube-dl --verbose http://www.rds.ca/vid%C3%A9os/un-voyage-positif-3.877934
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['--verbose', 'http://www.rds.ca/vid%C3%A9os/un-voyage-positif-3.877934']
[debug] youtube-dl version 2013.10.17
[debug] Python version 3.3.2 - Linux-3.11.5-1-ARCH-x86_64-with-arch
[debug] Proxy map: {}
[generic] un-voyage-positif-3.877934: Downloading webpage
[generic] un-voyage-positif-3.877934: Extracting information
ERROR: Unsupported URL: http://www.rds.ca/vid%C3%A9os/un-voyage-positif-3.877934; please report this issue on https://yt-dl.org/bug . Be sure to call youtube-dl with the --verbose flag and include its complete output. Make sure you are using the latest version; type  youtube-dl -U  to update.
  File "/usr/lib/python3.3/site-packages/youtube_dl/YoutubeDL.py", line 353, in extract_info
    ie_result = ie.extract(url)
  File "/usr/lib/python3.3/site-packages/youtube_dl/extractor/common.py", line 117, in extract
  File "/usr/lib/python3.3/site-packages/youtube_dl/extractor/generic.py", line 173, in _real_extract
    raise ExtractorError(u'Unsupported URL: %s' % url)
youtube_dl.utils.ExtractorError: Unsupported URL: http://www.rds.ca/vid%C3%A9os/un-voyage-positif-3.877934; please report this issue on https://yt-dl.org/bug . Be sure to call youtube-dl with the --verbose flag and include its complete output. Make sure you are using the latest version; type  youtube-dl -U  to update.

fstirlitz:
This should resolve #1611.

### DIFF
--- a/youtube_dl/extractor/rdsca.py
+++ b/youtube_dl/extractor/rdsca.py
@@ -1,0 +1,50 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import (
+    parse_iso8601,
+    url_basename,
+)
+
+
+class RDScaIE(InfoExtractor):
+    IE_NAME = 'RDS.ca'
+    _VALID_URL = r'http://(?:www\.)?rds\.ca/videos/(?P<id>.*)'
+
+    _TESTS = [{
+        'url': 'http://www.rds.ca/videos/football/nfl/fowler-jr-prend-la-direction-de-jacksonville-3.1132799',
+        'info_dict': {
+            "ext": "mp4",
+            "title": "Fowler Jr. prend la direction de Jacksonville",
+            "description": "Dante Fowler Jr. est le troisième choix du repêchage 2015 de la NFL. ",
+            "timestamp": 1430397346,
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = url_basename(url)
+
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._search_regex(
+            r'<span itemprop="name"[^>]*>([^\n]*)</span>', webpage, 'video title', default=None)
+        video_url = self._search_regex(
+            r'<span itemprop="contentURL" content="([^"]+)"', webpage, 'video URL')
+        upload_date = parse_iso8601(self._search_regex(
+            r'<span itemprop="uploadDate" content="([^"]+)"', webpage, 'upload date', default=None))
+        description = self._search_regex(
+            r'<span itemprop="description"[^>]*>([^\n]*)</span>', webpage, 'description', default=None)
+        thumbnail = self._search_regex(
+            r'<span itemprop="thumbnailUrl" content="([^"]+)"', webpage, 'upload date', default=None)
+
+        return {
+            'id': video_id,
+            'title': title,
+            'description': description,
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/rdsca.py
```diff
@@ -0,0 +1,50 @@
...
+from .common import InfoExtractor
+from ..utils import (
+    parse_iso8601,
+    url_basename,
+)
+
+
+class RDScaIE(InfoExtractor):
+    IE_NAME = 'RDS.ca'
+    _VALID_URL = r'http://(?:www\.)?rds\.ca/videos/(?P<id>.*)'
```
dstftw: Looks like all the URLs satisfy the following format: http://www.rds.ca/videos/.../<display_id>.<id>. `_VALID_URL` should be fixed correspondingly or test with another URL schema should be provided.

youtube_dl/extractor/rdsca.py
```diff
@@ -0,0 +1,50 @@
...
+
+    _TESTS = [{
+        'url': 'http://www.rds.ca/videos/football/nfl/fowler-jr-prend-la-direction-de-jacksonville-3.1132799',
```
dstftw: There should be md5 of the file's first 10k.

youtube_dl/extractor/rdsca.py
```diff
@@ -0,0 +1,50 @@
...
+        'info_dict': {
```
dstftw: `id` and `display_id` should be tested as well.

youtube_dl/extractor/rdsca.py
```diff
@@ -0,0 +1,50 @@
...
+            "ext": "mp4",
```
dstftw: Use quotes consistently - only a single or only a double. Use different kind of quotes only if there is no way to use same kind of quotes.

youtube_dl/extractor/rdsca.py
```diff
@@ -0,0 +1,50 @@
...
+            "title": "Fowler Jr. prend la direction de Jacksonville",
+            "description": "Dante Fowler Jr. est le troisième choix du repêchage 2015 de la NFL. ",
+            "timestamp": 1430397346,
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = url_basename(url)
```
dstftw: As already mentioned `id` and `display_id` should be extracted.

youtube_dl/extractor/rdsca.py
fstirlitz: Well... yes?

youtube_dl/extractor/rdsca.py
fstirlitz: There seems to be no such thing as an "ID", apart from the URL itself.

PR #6199, opened 2015-07-12
[sportschau] Add new extractor

slangangular:
This adds support for videos from sportschau.de, a german sports website. 
Sorry for breaking the convention of using "Add new extractor" in the commit message.


jaimeMF:
Merged, thanks!

Would you like to be credited in the AUTHORS file? If so, under which name?

slangangular:
Thanks! If you don't insist on real names, you could add me with my github username. No problem though, if you don't want to do this.

jaimeMF:
Added in 2711e41bcdaf8e8234dd32e5df6fb76f23179f46

PR #6208, opened 2015-07-13
DiscoveryIE: make playlist extraction tolerate entries with blank 'src' values

atomicdryad:
There are alot here. I assume the videos are disabled.
http://www.discovery.com/tv-shows/other-shows/videos/other-shows-when-we-left-earth-videos/

There is an 'embedurl' for all of the videos, however there is no support for the
urls in question. (todo)

### DIFF
--- a/youtube_dl/extractor/discovery.py
+++ b/youtube_dl/extractor/discovery.py
@@ -41,19 +41,27 @@ class DiscoveryIE(InfoExtractor):
         info = self._download_json(url + '?flat=1', video_id)
 
         video_title = info.get('playlist_title') or info.get('video_title')
-
-        entries = [{
-            'id': compat_str(video_info['id']),
-            'formats': self._extract_m3u8_formats(
-                video_info['src'], video_id, ext='mp4',
-                note='Download m3u8 information for video %d' % (idx + 1)),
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/discovery.py
```diff
@@ -41,19 +41,27 @@ def _real_extract(self, url):
...
-            'duration': parse_duration(video_info.get('video_length')),
-            'webpage_url': video_info.get('href'),
-            'thumbnail': video_info.get('thumbnailURL'),
-            'alt_title': video_info.get('secondary_title'),
-            'timestamp': parse_iso8601(video_info.get('publishedDate')),
-        } for idx, video_info in enumerate(info['playlist'])]
+        entries = []
+        collected = {}
+        for idx, video_info in enumerate(info['playlist']):
+            if collected.get( video_info.get('id') ):
```
jaimeMF: This wasn't used previously, is it really needed?
If the answer is yes, you should use a `set` instead of a dictionary for keeping track of them.

yan12125:
@atomicdryad The mentioned example now works fine with 2016.04.06. Do you have other examples?

dstftw:
Works fine with latest version.

PR #6209, opened 2015-07-13
Add support for reuters.com news videos, and yospace.com backend

atomicdryad:


youtube_dl/extractor/yospace.py
```diff
@@ -0,0 +1,202 @@
...
+                    ret.append({
+                        'id': vid.get('id', video_id),
+                        'title': vid.get('title', video_id),
+                        'description': vid.get('description'),
+                        'webpage_url': url,
+                        'formats': formats,
+                    })
+        if not ret:
+            raise ExtractorError('No video found', expected=True)
+        if len(ret) > 1:
```
jaimeMF: I would always return a playlist.

Please add a test with more than one video, like [in the ted extractor](https://github.com/rg3/youtube-dl/blob/6621ca39a3f03e44648b54bd8ff57814362f5dbc/youtube_dl/extractor/ted.py#L64-L69).

youtube_dl/extractor/yospace.py
```diff
@@ -0,0 +1,202 @@
...
+        ret = []
+        webpage = self._download_webpage(url, video_id)
+        vids = self._scrape_javascript(webpage)
+        for vid in vids:
+            vurl = vid.get('flv', vid.get('mpeg'))
+            if vurl:
+                formats = []
+                formats.append({
+                    'url': vurl,
+                    'format_id': 'embed-flv-' + str(vid.get('vidnum')),
```
jaimeMF: Why do you add the `vidnum` to the `format_id`?

youtube_dl/extractor/yospace.py
```diff
@@ -0,0 +1,202 @@
...
+
+        return {
+            'id': video_id.replace('/', '_'),
+            'display_id': display_id,
+            'title': video_id.replace('/', '_'),
+            'formats': formats,
+        }
+
+
+class ReutersIE(YospaceIE):
```
jaimeMF: This can be placed in its own file.

youtube_dl/extractor/yospace.py
```diff
@@ -0,0 +1,202 @@
...
+
+        if hls_url is not None:
+            formats.extend(self._extract_m3u8(hls_url))
+
+        self._sort_formats(formats)
+
+        return {
+            'id': video_id.replace('/', '_'),
+            'display_id': display_id,
+            'title': video_id.replace('/', '_'),
```
jaimeMF: Since this extractor by itself can't provide too much info, maybe it would be better to remove the _real_extract and _VALID_URL, don't import it in `__init__` and do something similar to `MTVServicesInfoExtractor`, which is the base class for the extractors that need it.

PR #6211, opened 2015-07-13
Add support for democracynow.org

atomicdryad:
Supports downloading clips or entire shows. Subtitle support

### DIFF
--- a/youtube_dl/extractor/democracynow.py
+++ b/youtube_dl/extractor/democracynow.py
@@ -1,0 +1,100 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import json
+import time
+import hmac
+import hashlib
+import itertools
+import re
+from ..utils import (
+    ExtractorError,
+    int_or_none,
+    parse_age_limit,
+    parse_iso8601,
+)
+from ..compat import compat_urllib_request
+from .common import InfoExtractor
+
+
+class DemocracynowIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?democracynow.org/?(?P<id>[^\?]*)'
+    IE_NAME = 'democracynow'
+    _TESTS = [{
+        'url': 'http://www.democracynow.org/shows/2015/7/3',
+        'info_dict': {
+            'id': '2015-0703-001',
+            'ext': 'mp4',
+            'title': 'July 03, 2015 - Democracy Now!',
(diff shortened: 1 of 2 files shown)


atomicdryad:
broken merge, resubmitting

atomicdryad:
Closed the wrong PR

youtube_dl/extractor/democracynow.py
```diff
@@ -0,0 +1,85 @@
...
+        formats = []
+        subtitles = {}
+        for key in ('caption_file', '.......'):
+            # ....... = pending vtt support that doesn't clobber srt 'chapter_file':
+            url = js.get(key, '')
+            if url == '' or url is None:
+                continue
+            if not re.match(r'^https?://', url):
+                url = base_host + url
+            ext = re.search(r'\.([^\.]+)$', url).group(1)
```
yan12125: Seems this line just extracts the file extension from the url. There's no need to explicitly provide ```ext``` if file extensions can be determined from URLs.

youtube_dl/extractor/democracynow.py
```diff
@@ -0,0 +1,85 @@
...
+        re_desc = re.search(r'<meta property=.og:description. content=(["\'])(.+?)\1', webpage, re.DOTALL)
+        description = re_desc.group(2) if re_desc else ''
+
+        jstr = self._search_regex(r'({.+?"related_video_xml".+?})', webpage, 'json', default=None)
+        js = self._parse_json(jstr, display_id)
+        video_id = None
+        formats = []
+        subtitles = {}
+        for key in ('caption_file', '.......'):
+            # ....... = pending vtt support that doesn't clobber srt 'chapter_file':
```
yan12125: Not quite understand. Why not add vtt files to ```subtitles```, too?

youtube_dl/extractor/democracynow.py
```diff
@@ -0,0 +1,85 @@
...
+                url = url + 'start=' + str(js.get('start'))
+            formats.append({
+                'format_id': purl.group('dir'),
+                'ext': purl.group('ext'),
+                'url': url,
+            })
+        self._sort_formats(formats)
+        ret = {
+            'id': video_id,
+            'title': js.get('title'),
```
yan12125: If 'title' is not in ```js```, there should be a fallback title.

yan12125:
Thanks for your work @atomicdryad. I've made some changes, especially removing ```start=X``` and ```end=Y``` parameters from the final URLs.

And, would you like to be listed in [AUTHORS](https://github.com/rg3/youtube-dl/blob/master/AUTHORS)? If so, under what name?

PR #6212, opened 2015-07-13
Load additional extractor module .py files placed in ~/.config/youtube-dl/modules, $XDG_CACHE_HOME/youtube-dl/modules, or a directory specified with --module-dir

atomicdryad:
Modules can simply be placed in the directory without editing an __init__.py
Modules are written as normal extractor modules and require no changes,
(ie from .common import InfoExtractor, etc) as they're loaded into
the youtube_dl.extractor. namespace

### DIFF
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -40,7 +40,7 @@ from .update import update_self
 from .downloader import (
     FileDownloader,
 )
-from .extractor import gen_extractors, list_extractors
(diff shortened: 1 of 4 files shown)


dstftw:
Here is the rationale why external extractors is not a good idea: https://github.com/rg3/youtube-dl/issues/1185#issuecomment-67412997, https://github.com/rg3/youtube-dl/pull/1198#issuecomment-67413240.

PR #6214, opened 2015-07-13
Load additional extractor module .py files placed in ~/.config/youtube-dl/modules (or equivalent), or path set by --module-dir option 

atomicdryad:
Modules can simply be placed in the directory without editing an __init__.py

Modules are written as normal extractor modules and require no changes,
(ie from .common import InfoExtractor, etc) as they're loaded into
the youtube_dl.extractor. namespace

youtube_dl/auto/*.py is autoloaded as well

### DIFF
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -40,7 +40,7 @@ from .update import update_self
 from .downloader import (
     FileDownloader,
 )
-from .extractor import gen_extractors, list_extractors
+from .extractor import gen_extractors, list_extractors, add_extractors
 from .YoutubeDL import YoutubeDL
 
 
@@ -55,6 +55,10 @@ def _real_main(argv=None):
     setproctitle('youtube-dl')
 
     parser, opts, args = parseOpts(argv)
+
+
+    from .autoload import load_dynamic_extractors
+    add_extractors( load_dynamic_extractors(opts.module_dir) )
 
     # Set user agent
     if opts.user_agent is not None:
--- a/youtube_dl/autoload.py
+++ b/youtube_dl/autoload.py
@@ -1,0 +1,42 @@
+import os
+import glob
+import imp
+from .compat import (
+    compat_expanduser,
+    compat_getenv,
+)
+
+
+mdirs = [ os.path.join(os.path.dirname(__file__),'/auto') ]
+
+def confdirs():
+    cfg_home = compat_getenv('XDG_CONFIG_HOME') or compat_getenv('appdata') or os.path.join(compat_expanduser('~'), '.config')
+
+    if cfg_home:
+       cfg_dir = os.path.join(cfg_home, 'youtube-dl','modules')
+       if os.path.isdir(cfg_dir):
+           return [cfg_dir]
+    return []
+
+def load_dynamic_extractors(module_dir=None):
+    mdirs.extend(confdirs())
+    if module_dir != None:
+        if not os.path.isdir(module_dir):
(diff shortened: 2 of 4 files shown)


jaimeMF:
I'm not completely against adding support for this but as explained in https://github.com/rg3/youtube-dl/pull/6212#issuecomment-120941892 some devs would prefer to avoid it unless it's actually needed.

remitamine:
for me it help to remove modules that i never use from the local instalation witch will make the execution speed better(less regex to try)

jaimeMF:
@remitamine youtube-dl will start faster but most of the execution time is spent extracting the information not in finding the right extractor. This can be tested with the following script:
```python
from __future__ import unicode_literals, division
import timeit

setup = '''
from youtube_dl import YoutubeDL

ydl = YoutubeDL({'quiet': True, 'no_warnings': True})
def match(url, extract=False):
    if extract:
        return ydl.extract_info(url, download=False)
    for ie in ydl._ies:
        if ie.suitable(url):
            return True
    return False
'''

nr = 100

def repeat(stmt):
    print(stmt)
    ts = timeit.repeat(stmt=stmt, setup=setup, number=nr)
    print([t/nr for t in ts])

repeat('match("http://podcastfeeds.nbcnews.com/audio/podcast/MSNBC-MADDOW-NETCAST-M4V.xml")')
repeat('match("http://podcastfeeds.nbcnews.com/audio/podcast/MSNBC-MADDOW-NETCAST-M4V.xml", True)')
repeat('match("http://www.youtube.com/watch?v=BaW_jenozKcj")')
repeat('match("http://www.youtube.com/watch?v=BaW_jenozKcj", True)')
```
Which outputs:
```
match("http://podcastfeeds.nbcnews.com/audio/podcast/MSNBC-MADDOW-NETCAST-M4V.xml")
[0.0013781707599991933, 0.0013171957299346103, 0.0013914224097970873]
match("http://podcastfeeds.nbcnews.com/audio/podcast/MSNBC-MADDOW-NETCAST-M4V.xml", True)
[0.12700892633001787, 0.11449998143012635, 0.11422629146953113]
match("http://www.youtube.com/watch?v=BaW_jenozKcj")
[0.00033350404992233963, 0.00040478369046468286, 0.0003109663899522275]
match("http://www.youtube.com/watch?v=BaW_jenozKcj", True)
[0.903358261960093, 0.8863492684200174, 0.9325759381300304]
```
As you can see iterating over all of the extractors is two order of magnitude faster than extracting the info, unless you are running a huge amount of extractions you won't benefit that much.

remitamine:
yes, this is right, because info extraction depend on network latency.
there another thing every time i update youtube-dl, i have to change __init__.py(and other files) to include the extractor that i've made and that are not accepted yet and there are extractors that i don't send because they depend on other pull requests(usa network and fox)

jaimeMF:
I'm closing the PR, read the links in https://github.com/rg3/youtube-dl/pull/6212#issuecomment-120941892 for the reasons.

@remitamine I use a wrapper around youtube-dl to switch between python versions:
```python
#!/usr/bin/env python3
import youtube_dl

youtube_dl.main()
```

You can modify it to load or delete extractors, for example:
```python
#!/usr/bin/env python3
import youtube_dl
import youtube_dl.extractor

for ie in load_external_extractors(...):
        youtube_dl.extractor._ALL_CLASSES.insert(0, ie)

youtube_dl.main()
```
Of course it can stop working at any time, but it should do the work while we don't add this functionality to youtube-dl.

PR #6217, opened 2015-07-13
Copy change: "skipping" to "all done" with -x

riking:
A copy change.

This happens when you specify -x for a YT video and an acceptable format (.m4a)
could be downloaded directly.


dstftw:
This is intentional. When playlist is downloaded phrase "all done" in the middle of the playlist sounds misleading especially along with the downloading continuation.

jaimeMF:
I thinks it's ok now, additionally other postprocessors could be run after it.

PR #6221, opened 2015-07-14
Fix video duration reporting for Dailymotion extractor

cazulu:
Quick patch to fix the duration reporting for the dailymotion metadata extractor.


jaimeMF:
Thanks! Cherry picked and updated the test in 50aa2bb6b94805d901f00d4da7a1b4ff1fcd5169.

PR #6242, opened 2015-07-15
[f4m] Improvements

dstftw:
1. `_extract_f4m_formats` now handles malformed f4m manifests (97f4aec).
2. URLs starting with `https?://` in 2.0 maniests are now handled properly (cc357c4).
3. Child f4m manifests are now extracted recursively (70f0f5a) in order to prevent format by requested bitrate resolve problem when parent and child manifest's bitrates differ.
4. prosiebensat1 switched to generic f4m extraction (f01f731).

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -28,6 +28,7 @@ from ..utils import (
     clean_html,
     compiled_regex_type,
     ExtractorError,
+    fix_xml_ampersands,
     float_or_none,
     int_or_none,
     RegexNotFoundError,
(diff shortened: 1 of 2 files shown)


jaimeMF:
It looks good, do you have a url that is not georestricted to test?

dstftw:
>a url that is not georestricted to test

[Yes](https://github.com/dstftw/youtube-dl/blob/f01f731107010e0c10fc94782daa7a3ba543e92a/youtube_dl/extractor/prosiebensat1.py#L25).

jaimeMF:
Thanks, it works. Feel free to merge it.

PR #6253, opened 2015-07-16
Fixed loading of some blip.tv videos.

cherez:
Some blip videos lack a runtime, so the current extractor crashes on grabbing the duration.


cherez:
For an example: http://blip.tv/rss/flash/6700880
In particular, note the line:
```      <blip:runtime/>```

dstftw:
Can you add a test?

dstftw:
Thanks. Fixed with xpath in 9441f77faaf078aca2947b7613f61f46b6d0aea2.

PR #6259, opened 2015-07-17
[howstuffwoks] fix _VALID_URL regex

remitamine:
make the extractor works with url that does not contain a number in the url like that:
http://shows.howstuffworks.com/stuff-to-blow-your-mind/optical-illusions-video.htm

### DIFF
--- a/youtube_dl/extractor/howstuffworks.py
+++ b/youtube_dl/extractor/howstuffworks.py
@@ -46,6 +46,17 @@ class HowStuffWorksIE(InfoExtractor):
                 'thumbnail': 're:^https?://.*\.jpg$',
             },
         },
+        {
+            'url': 'http://shows.howstuffworks.com/stuff-to-blow-your-mind/optical-illusions-video.htm',
+            'info_dict': {
+                'id': '855410',
+                'ext': 'mp4',
+                'title': 'Stuff to Blow Your Mind',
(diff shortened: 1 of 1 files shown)


dstftw:
Can you add a test?

youtube_dl/extractor/howstuffworks.py
```diff
@@ -10,7 +10,7 @@
 
 
 class HowStuffWorksIE(InfoExtractor):
-    _VALID_URL = r'https?://[\da-z-]+\.howstuffworks\.com/(?:[^/]+/)*\d+-(?P<id>.+?)-video\.htm'
+    _VALID_URL = r'https?://[\da-z-]+\.howstuffworks\.com/(?:[^/]+/)*(\d+-)*(?P<id>.+?)-video\.htm'
```
dstftw: That's wrong by the way. It matches zero on more `\d+-` parts. Also do not capture group if you dont not use it.

remitamine:
i correct the mistakes you point at is this enough.

youtube_dl/extractor/howstuffworks.py
```diff
@@ -46,6 +46,17 @@ class HowStuffWorksIE(InfoExtractor):
                 'thumbnail': 're:^https?://.*\.jpg$',
             },
         },
+        {
+            'url': 'http://shows.howstuffworks.com/stuff-to-blow-your-mind/optical-illusions-video.htm',
+            'info_dict': {
```
dstftw: `'only_matching': True` is enough.

PR #6273, opened 2015-07-18
[tudou]Fix extraction

linhua55:
for http://www.tudou.com/albumplay/cJAHGih4yYg.html


yan12125:
Thanks for fixing it. Would you like to be listed in [AUTHORS](https://github.com/rg3/youtube-dl/blob/master/AUTHORS)? If so, under what name?

PR #6277, opened 2015-07-18
[shahid] Add new extractor

remitamine:


youtube_dl/extractor/shahid.py
```diff
@@ -0,0 +1,56 @@
...
+        description = json_data['description']
+        player_json_data = self._download_json(
+            'https://shahid.mbc.net/arContent/getPlayerContent-param-.id-'+video_id+'.type-player.html',
+            video_id
+        )['data']
+        if 'url' in player_json_data:
+            m3u8_url = player_json_data['url']
+        else:
+            for error in json_data['error'].values():
+                self.report_warning(error)
```
dstftw: Raise `ExtractorError` instead.

youtube_dl/extractor/shahid.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        json_data = self._parse_json(
+            get_element_by_id('jsonld', webpage),
+            video_id
+        )
+        title = json_data['name']
+        thumbnail = json_data['image']
```
dstftw: Everything except title should be optional.

remitamine: I thought that it was supposed to extract as much information as possible.
Do I have to remove the extraction of optional data.

dstftw: It does. It does not supposed to fail completely if some of the optional fields are missing.

youtube_dl/extractor/shahid.py
```diff
@@ -0,0 +1,81 @@
...
+    ExtractorError,
+    int_or_none
+)
+
+
+class ShahidIE(InfoExtractor):
+    _VALID_URL = r'https?://shahid\.mbc\.net/ar/episode/(?P<id>\d+)/?'
+    _TESTS = [
+        {
+            'url': 'https://shahid.mbc.net/ar/episode/108084/%D8%AE%D9%88%D8%A7%D8%B7%D8%B1-%D8%A7%D9%84%D9%85%D9%88%D8%B3%D9%85-11-%D8%A7%D9%84%D8%AD%D9%84%D9%82%D8%A9-1.html',
```
dstftw: This URL is now subscriber only. I've managed to find free one:
```
https://shahid.mbc.net/ar/episode/90574/%D8%A7%D9%84%D9%85%D9%84%D9%83-%D8%B9%D8%A8%D8%AF%D8%A7%D9%84%D9%84%D9%87-%D8%A7%D9%84%D8%A5%D9%86%D8%B3%D8%A7%D9%86-%D8%A7%D9%84%D9%85%D9%88%D8%B3%D9%85-1-%D9%83%D9%84%D9%8A%D8%A8-3.html
```

youtube_dl/extractor/shahid.py
```diff
@@ -0,0 +1,81 @@
...
+                'duration': 1166,
+            },
+            'params': {
+                # m3u8 download
+                'skip_download': True,
+            }
+        },
+        {
+            # shahid plus subscriber only
+            'url': 'https://shahid.mbc.net/ar/series/90497/%D9%85%D8%B1%D8%A7%D9%8A%D8%A7-2011.html',
```
dstftw: This is not matched by `_VALID_URL`.

youtube_dl/extractor/shahid.py
```diff
@@ -0,0 +1,81 @@
...
+        )['data']
+        if 'url' in player_json_data:
+            m3u8_url = player_json_data['url']
+        else:
+            for error in player_json_data['error'].values():
+                raise ExtractorError(error)
+        formats = self._extract_m3u8_formats(m3u8_url, video_id)
+
+        video_info = self._download_json(
+            player_info['url'] + '/' + player_type + '/' + video_id + '?apiKey=sh%40hid0nlin3&hash=b2wMCTHpSmyxGqQjJFOycRmLSex%2BBpTK%2Fooxy6vHaqs%3D',
```
dstftw: Have you seen any example when `player_type` is something different from `episode`? If not I suppose it's reasonable to just hardcode the API URL instead of extracting it and `playerType` from `flashvars` since it looks very brittle and have higher probability to change.

youtube_dl/extractor/shahid.py
remitamine: it's a mistake.
i put a series url instead of an episode url.
i will fix it now.

youtube_dl/extractor/shahid.py
remitamine: it's just what they use internaly in their player.

youtube_dl/extractor/shahid.py
remitamine: this is strange for me it's available for free.
anyway, i will change it as the url you give me is working for both of us.

remitamine:
i changed the tests.

youtube_dl/extractor/shahid.py
dstftw: So can you name any example URL that requests `http://api.shahid.net/api/v1_1/<not_episode_string_here>/...`?

remitamine: no, i don't have an example of url with a different type i just that in their code they use the player type variable in the api url.

dstftw: I'm aware of that. What I'm trying to tell is that it would be more robust to just hardcode `player_info['type']` and `player_info['playerType']` or create default fallbacks if extraction of these fails. [This part](https://github.com/remitamine/youtube-dl/blob/shahid/youtube_dl/extractor/shahid.py#L41-L44) is very brittle and will likely break the extraction if layout inside `flashvars` changes.

remitamine: if i hardcode the variables in the flashvars and keep the part that could break the extraction as comment. is this acceptable?

dstftw: What's the point keeping code as comment?

remitamine: if it appears that some of the urls don't work because of the difference in the player type or other hardcoded values it will be easy to reuse it.

dstftw: Then just keep this code and
>create default fallbacks if extraction of these fails

remitamine: ok, i will work on it.

remitamine: how about this?
```python
    _API_VARS = {
        'type': 'player',
        'url': 'http://api.shahid.net/api/v1_1',
        'playerType': 'episode',
    }

    def _real_extract(self, url):

        player_info = ''
        for line in self._search_regex('var flashvars = ({[^}]+})', webpage, 'flashvars').splitlines():
            if '+' not in line and '(' not in line and ')' not in line:
                player_info += line
        player_info = self._parse_json(player_info, video_id, js_to_json, False)
        if player_info is not None:
            for key in self._API_VARS:
                if key in player_info:
                    self._API_VARS[key] = player_info[key]

        player_json_data = self._download_json(
            'https://shahid.mbc.net/arContent/getPlayerContent-param-.id-' + video_id + '.type-' + self._API_VARS['type'] + '.html',
            video_id
        )['data']
        if 'url' in player_json_data:
            m3u8_url = player_json_data['url']
        else:
            for error in player_json_data['error'].values():
                raise ExtractorError(error)
        formats = self._extract_m3u8_formats(m3u8_url, video_id)

        video_info = self._download_json(
            self._API_VARS['url'] + '/' + self._API_VARS['playerType'] + '/' + video_id + '?apiKey=sh%40hid0nlin3&hash=b2wMCTHpSmyxGqQjJFOycRmLSex%2BBpTK%2Fooxy6vHaqs%3D',
            video_id
        )['data']
```

dstftw: Upper case is idiomatic for constants.

remitamine: what do you mean?

dstftw: UPPER_CASE is idiomatically used for constants in python. Therefore `_API_VARS` should not change it's state since initially defined.
Here:
```python
self._API_VARS[key] = player_info[key]
```
you modify it.

remitamine: so, i will use lowercase for this variable, what else?

remitamine:
i added the default fallback for api vars

dstftw:
Thanks.

PR #6292, opened 2015-07-19
pbs: fix vague 'Full Episode' titles; prepend name of show

atomicdryad:
Many videos are given the title 'Full Episode' or 'Episode x', etc;
in the info json 'title' is the episode title (for oneshots this is
'Full Episode') while the program name is 'program':{'title'. This
changes the title from '(episode title)' to
 '(program name) - (episode title)'

The following urls demonstrate useless titles

    http://video.pbs.org/video/2365367186/
    http://video.pbs.org/video/2365519307/
    http://video.pbs.org/video/2365527039/
    http://video.pbs.org/video/2365530605/

Before change

    Full Episode
    Episode 5 | Preview
    Season 3 | Episode 4 Preview
    July 17, 2015

After change

    To Catch A Comet - Full Episode
    The Crimson Field - Episode 5 | Preview
    Last Tango in Halifax - Season 3 | Episode 4 Preview
    Charlie Rose The Week - July 17, 2015

### DIFF
--- a/youtube_dl/extractor/pbs.py
+++ b/youtube_dl/extractor/pbs.py
@@ -249,8 +249,8 @@ class PBSIE(InfoExtractor):
 
         # video.pbs.org video.pbs.org/videoInfo/... frequently provides an obscure 'title' value, like
         # 'Full Episode', 'Episode 5', etc. prepend program->title
-        alt_title = info.get('program', {}).get('title', '')
-        if alt_title != '':
+        alt_title = info.get('program', {}).get('title')
+        if alt_title:
             info['title'] = alt_title + ' - ' + re.sub(r'^' + alt_title + '[\s\-\:]+', '', info['title'])
 
         return {
--- a/youtube_dl/extractor/pbs.py
+++ b/youtube_dl/extractor/pbs.py
@@ -32,7 +32,7 @@ class PBSIE(InfoExtractor):
             'info_dict': {
                 'id': '2365006249',
                 'ext': 'mp4',
-                'title': 'A More Perfect Union',
+                'title': 'Constitution USA with Peter Sagal - A More Perfect Union',
                 'description': 'md5:ba0c207295339c8d6eced00b7c363c6a',
                 'duration': 3190,
             },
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/pbs.py
```diff
@@ -232,6 +247,12 @@ def _real_extract(self, url):
                 'url': closed_captions_url,
             }]
 
+        # video.pbs.org video.pbs.org/videoInfo/... frequently provides an obscure 'title' value, like
+        # 'Full Episode', 'Episode 5', etc. prepend program->title
+        alt_title = info.get('program', {}).get('title', '')
+        if alt_title != '':
```
dstftw: Use more python idiomatic `.get('title')` and `if alt_title:` instead.

atomicdryad:
Would that catch u'' and None?

dstftw:
Yes.

youtube_dl/extractor/pbs.py
```diff
@@ -232,6 +247,12 @@ def _real_extract(self, url):
                 'url': closed_captions_url,
             }]
 
+        # video.pbs.org video.pbs.org/videoInfo/... frequently provides an obscure 'title' value, like
+        # 'Full Episode', 'Episode 5', etc. prepend program->title
+        alt_title = info.get('program', {}).get('title')
+        if alt_title:
+            info['title'] = alt_title + ' - ' + re.sub(r'^' + alt_title + '[\s\-\:]+', '', info['title'])
```
dstftw: Wouldn't it be easier to just use `info['title']` when it contains `alt_title` instead of cutting it from `info['title']` and then prepending again?

atomicdryad: http://video.pbs.org/video/2365530886/
info['title']:  PBS NewsHour full episode July 17, 2015
alt_title: PBS NewsHour

'PBS NewsHour - full episode July 17, 2015' is consistent with the '(show name) - (episode title)' naming scheme

dstftw:
Thanks, merged.

Issue #5118, opened 2015-03-03
Skipping DASH manifest


amkrankruleuen:
user@sub2:~/conversion# youtube-dl -e "https://www.youtube.com/watch?v=7f8nSWtJLQo"

WARNING: [youtube] 7f8nSWtJLQo: Skipping DASH manifest: ExtractorError(u'Could not download DASH manifest: HTTP Error 403: Forbidden (caused by HTTPError()); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.',)
Becky G - Magik 2.0 (Lyric Video)  ft. Austin Mahone

dstftw:
>Be sure to call youtube-dl with the --verbose flag and include its complete output

amkrankruleuen:
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['-e', '--verbose', 'https://www.youtube.com/watch?v=7f8nSWtJLQo']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.03.03.1
[debug] Python version 2.7.3 - Linux-3.14.32-xxxx-grs-ipv6-64-x86_64-with-Debian-7
[debug] exe versions: avconv 0.8.16-6, avprobe 0.8.16-6, ffmpeg 2.3, ffprobe 2.3
[debug] Proxy map: {}
WARNING: [youtube] 7f8nSWtJLQo: Skipping DASH manifest: ExtractorError(u'Could not download DASH manifest: HTTP Error 403: Forbidden (caused by HTTPError()); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.',)
Becky G - Magik 2.0 (Lyric Video)  ft. Austin Mahone

ivan:
The same error is printed for https://www.youtube.com/watch?v=RQuY8kERaU0 (though the video downloads anyway.)

baruns:
What is the solution then?? It's happening for latest version also.. Video gets downloaded without problem... Strange!!!

[test@my ~]# youtube-dl -s -4 http://www.youtube.com/watch?v=5OAW3TKwfyk --verbose
[debug] System config: []
[debug] User config: []
[...]

PR #6304, opened 2015-07-20
[youtube] Don't use the DASH manifest from 'get_video_info' if 'use_cipher_signature' is True (#5118)

jaimeMF:
Currently they give a 403 Forbidden error.

I don't know if it's worth trying to find a way to download those manifests, but at least we can skip one network request and won't get more issue reports.


dstftw:
LGTM.

jaimeMF:
Cherry picked in 87dc451108b278eb5e5600a0159bdbd337272392.

PR #6306, opened 2015-07-20
[youtube] Extract start_time

jaimeMF:
From the 't=*' in the url.
Currently youtube-dl doesn't use the value, but it was requested for the mpv plugin (https://github.com/mpv-player/mpv/issues/1519).

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -900,6 +900,8 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
             query = compat_parse_qs(component)
             if start_time is None and 't' in query:
                 start_time = parse_duration(query['t'][0])
+            if start_time is None and 'start' in query:
+                start_time = parse_duration(query['start'][0])
             if end_time is None and 'end' in query:
                 end_time = parse_duration(query['end'][0])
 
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/youtube.py
```diff
@@ -889,6 +892,14 @@ def _real_extract(self, url):
             'http' if self._downloader.params.get('prefer_insecure', False)
             else 'https')
 
+        start_time = None
+        parsed_url = compat_urllib_parse_urlparse(url)
+        for component in [parsed_url.fragment, parsed_url.query]:
+            query = compat_parse_qs(component)
+            if 't' in query:
```
dstftw: It also recognizes it as `start`. Moreover it supports `end` counterpart (e.g. http://www.youtube.com/v/4OuHSDsuhXc?start=10&end=25) that would be reasonable to add as well hoping youtube-dl will support downloading fragments some day.

jaimeMF: Done in https://github.com/jaimeMF/youtube-dl/commit/297a564beeb20ca8b00d94f5707532110631f409 and https://github.com/jaimeMF/youtube-dl/commit/2929fa0e79dfd3a1366e7e23eb4344bc93dd3a10

PR #6312, opened 2015-07-21
[nbc] remove unavailable tests

remitamine:
this pages are not available anymore:
http://www.nbc.com/the-tonight-show/episodes/176:
```
Thank you, 404 Error, for being our website's way of saying, "We know you're looking for something, but we couldn't find anything, so here's nothing."
```
http://www.nbcnews.com/feature/dateline-full-episodes/full-episode-family-business-n285156
```
This page is unavailable.
```

### DIFF
--- a/youtube_dl/extractor/nbc.py
+++ b/youtube_dl/extractor/nbc.py
@@ -30,16 +30,6 @@ class NBCIE(InfoExtractor):
             },
         },
         {
-            'url': 'http://www.nbc.com/the-tonight-show/episodes/176',
-            'info_dict': {
-                'id': 'XwU9KZkp98TH',
-                'ext': 'flv',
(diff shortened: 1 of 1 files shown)


yan12125:
Usually a ```'skip': '(reason)',``` line is added instead of removing the whole test case. We try to retain the test case orders. (test_NBC_1, test_NBC_2, etc.)

remitamine:
i just want to understand:
- what is the benefit of keeping broken urls in the tests?
- why trying to retain the test case orders if its generated automaticaly?

https://github.com/rg3/youtube-dl/blob/master/test/test_download.py#L221

yan12125:
Sometimes test names are referred in issues/pull requests or commit messages.

remitamine:
i will close this pull request and an other pull request to skip this test or just if you can add it directly

PR #6322, opened 2015-07-21
Extend twitter support (status urls, titles, vmaps)

atomicdryad:


youtube_dl/extractor/twitter.py
```diff
@@ -44,6 +57,20 @@ def _real_extract(self, url):
                 unescapeHTML(self._search_regex(
                     r'data-player-config="([^"]+)"', webpage, 'data player config')),
                 video_id)
+            if 'playlist' not in config:
+                if 'vmapUrl' in config:
+                    webpage = self._download_webpage(config['vmapUrl'], video_id + ' (xml)')
```
yan12125: Using ```self._download_xml``` is better here. And by the way, it's uncommon to use prefixes or suffixes in the second field.

youtube_dl/extractor/twitter.py
```diff
@@ -44,6 +57,20 @@ def _real_extract(self, url):
...
+                    video_url = self._search_regex(
+                        r'<MediaFile>\s*<!\[CDATA\[(https?://.+?)\]\]>', webpage, 'data player config (xml)')
+                    f = {
+                        'url': video_url,
+                    }
+                    ext = re.search(r'\.([a-z0-9]{2,4})(\?.+)?$', video_url)
```
yan12125: [```determine_ext```](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/utils.py#L901) is more robust. However, usually you don't need to specify ```ext``` in formats dictionary.

youtube_dl/extractor/twitter.py
```diff
@@ -70,3 +97,47 @@ def _real_extract(self, url):
             'duration': duration,
             'formats': formats,
         }
+
+
+class TwitterIE(TwitterCardIE):
```
yan12125: Inherit from ```InfoExtractor``` is enough.

youtube_dl/extractor/twitter.py
```diff
@@ -70,3 +97,47 @@ def _real_extract(self, url):
...
+        },
+    }]
+
+    def _real_extract(self, url):
+        id = self._match_id(url)
+        username, twid = re.match(r'([^/]+)/status/(\d+)', id).groups()
+        name = username
+        url = re.sub(r'https?://(m|mobile)\.', 'https://', url)
+        webpage = self._download_webpage(url, 'tweet: ' + url)
+        description = unescapeHTML(self._search_regex('<title>\s*(.+?)\s*</title>', webpage, 'title'))
```
yan12125: You may want ```self._html_search_regex``` here.

yan12125:
Thanks. Merged.

PR #6340, opened 2015-07-22
new param "extra_http_headers", dictionary of extra headers to send w…

nlevitt:
…ith every http request

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -256,6 +256,7 @@ class YoutubeDL(object):
                        If it returns None, the video is downloaded.
                        match_filter_func in utils.py is one example for this.
     no_color:          Do not emit color codes in output.
+    extra_http_headers: Extra http headers to include with every request.
 
     The following options determine which downloader is picked:
     external_downloader: Executable of the external downloader to call.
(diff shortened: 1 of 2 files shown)


dstftw:
What's the point in sending it with every request? Why [this is not enough](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L116-L117)?

jaimeMF:
I would prefer to modify [YoutubeDLHandler](https://github.com/rg3/youtube-dl/blob/9c29bc69f7d6365835f495dff10f3c5f49671a55/youtube_dl/utils.py#L610) so that it can use a header [dictionary other than `std_headers`](https://github.com/rg3/youtube-dl/blob/9c29bc69f7d6365835f495dff10f3c5f49671a55/youtube_dl/utils.py#L653), in my opinion it would be cleaner.

nlevitt:
@jaimeMF I updated the pull request according to your suggestion. Hadn't noticed that part of the code.

@dstftw My use case is kind of specific. It's for use in a new web crawler (https://github.com/nlevitt/brozzler) coupled with an archiving http proxy (https://github.com/internetarchive/warcprox). Every http request from the crawler needs to include a custom header to pass context information to the proxy.

I could understand if this is not general enough of a use case to warrant inclusion in mainline youtube-dl. But I will need to maintain my own fork if you guys make that decision.

Another approach that would work for my use case would be to allow users to add custom handlers to the OpenerDirector chain. I don't see a good way to do that in the existing code, though I might be missing something. If you prefer that, I could prepare a different pull request.

PR #6353, opened 2015-07-24
Add --socks-proxy parameter to use socks proxy

### LINKED ISSUE #402: SOCKS proxy support
The necessity arose in #171 and even I needed it to debug a geo-blocked video.
There was an attempt on #305 that faced license issues.

I implemented it some time ago with SocksiPy-branch but we don't like external dependencies, do we?

Don42:
resolves #402
This allows the user to specify a socks proxy to
tunnel the connection through. 
This does add a soft dependency on socksipy. Normal operation will work without it,
but socks support obviously depends on it.

I did some positive tests and a videos were downloaded from youtube via the socks proxy.
I did not test other hosts, but assume it should work the same.
I am not certain what happens if --socks-proxy and --proxy are specified simultaneously. Did not have a http proxy to test this.
The unittests have not finished running yet, but since none of my code (except adding the option)  is actually run if the --socks-proxy option is not set it seems highly unlikely for it break existing functionality.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -59,6 +59,7 @@ from .utils import (
     locked_file,
     make_HTTPS_handler,
     MaxDownloadsReached,
+    OptionalDependencyNotFound,
     PagedList,
     parse_filesize,
     PerRequestProxyHandler,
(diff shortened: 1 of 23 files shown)


yan12125:
This approach forces all network traffic going through the SOCKS proxy. The [urllib2 handler](https://github.com/Anorov/PySocks#urllib2-handler) approach might be better.

Don42:
Why is it a bad thing if all traffic goes through the proxy? Which traffic do you want bypassing the proxy?

yan12125:
For some Chinese sites ([letv](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/letv.py#L91-93), [sohu](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/sohu.py#L101-103) and [youku](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/youku.py#L181-183)), it's possible to use different proxies for different requests. In such cases your solution introduce traffic waste.

Don42:
I think I nearly got it. Will update this PR tomorrow.

dstftw:
As already mentioned in IRC conversation you should ensure there are no problems bundling SocksiPy with py2exe in the first place.

Don42:
Sorry about all the commits. I'm going to close this one and create a new clean PR.

PR #6360, opened 2015-07-25
[youtube] Add support for multifeed videos

dstftw:
Adds support for downloading multifeed videos (from multiple cameras) as playlist (e.g. https://www.youtube.com/watch?v=jqWvoWXjCVs)

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -1089,7 +1089,9 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
             else:
                 video_description = ''
 
-        if not smuggled_data.get('force_singlefeed', False) and 'multifeed_metadata_list' in video_info:
+        if (not self._downloader.params.get('noplaylist') and
+                not smuggled_data.get('force_singlefeed', False) and
+                'multifeed_metadata_list' in video_info):
             entries = []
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/youtube.py
```diff
@@ -1005,6 +1062,50 @@ def add_dash_mpd(video_info):
...
+            ''', r'\1', video_description)
+            video_description = clean_html(video_description)
+        else:
+            fd_mobj = re.search(r'<meta name="description" content="([^"]+)"', video_webpage)
+            if fd_mobj:
+                video_description = unescapeHTML(fd_mobj.group(1))
+            else:
+                video_description = ''
+
+        if (not self._downloader.params.get('noplaylist') and
```
jaimeMF: I would add a message like [the one for playlists](https://github.com/rg3/youtube-dl/blob/1af330f29f0ff8d8eba245591c7aa17ca2b7c3ed/youtube_dl/extractor/youtube.py#L1465) about how to avoid this behaviour.

dstftw: Thanks, fixed in 5e1eddb939a4fbc2a5ef10111c8141c842cf01d6.

jaimeMF:
Merged in 34866b4836eab8dd2fcaae88dc1ed5c79a742c92.

PR #6370, opened 2015-07-26
[spiegel] Accept iframe urls

tippfeler:
Some video URLs on spiegel.de contain "-iframe" instead of "-embed" (or no suffix). This commit enables the extraction for these URLs as well. The actual extraction process remains unchanged.


jaimeMF:
Thanks! Cherry picked in aeb7b41d44313b6bb007b2f1cd0fc8cda84e59d5, I have set the test to only match the url because we already download the video in the previous test.

PR #6372, opened 2015-07-26
Remove redundant (and wrong) class parameters

raphaelm:
I just stumpled upon those two lines which are absolutely redundant, as the parameters are set in the constructor for every class instance. At the same time, class parameters and instance parameters are very different things in python so this might even cause harm :)


dstftw:
Thanks.

PR #6389, opened 2015-07-28
[youtube] Added argument --write-tags, that ads the list of tags to t…

purdeaandrei:
…he end of the description file. Implies --write-description.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -150,6 +150,7 @@ class YoutubeDL(object):
     logger:            Log messages to a logging.Logger instance.
     logtostderr:       Log messages to stderr instead of stdout.
     writedescription:  Write the video description to a .description file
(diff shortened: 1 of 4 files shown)


dstftw:
Why not just use `--write-info-json`?

purdeaandrei:
well, indeed, only the changes in youtube.py are necessary in that case.

I like having the option of including the keyword in the description too, but of course whoever merges it will decide if they want to merge it. I will make another pull-request with just the youtube.py changes that adds the tags to the json.

dstftw:
I see little sense adding `--write-*` option for an arbitrary field until there is a clear rationale and purpose (e.g. some well defined format consumed by a 3rd party software).

purdeaandrei:
Well, one could argue that it better matches what you see when you click "Show More" on youtube. But really to argue that, one would also have to add the other fields that are showed there, not just Keywords, so I see your point :)

PR #6390, opened 2015-07-28
[youtube] save keywords in info jason when --write-info-json is used

purdeaandrei:
Simpler version of this pull request: https://github.com/rg3/youtube-dl/pull/6389

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -1072,9 +1072,8 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
         else:
             video_categories = None
 
-        m = re.findall(r'''<meta(?:\s+[a-zA-Z0-9:._-]+(?:=[a-zA-Z0-9:._-]+|="[^"]+"|='[^']+'))*?\s+property=['"]?og:video:tag['"]?(?:\s+[a-zA-Z0-9:._-]+(?:=[a-zA-Z0-9:._-]+|="[^"]+"|='[^']+'))*?\s+content=['"]?([^>'"]+?)['"]?\s*>'''
+        video_tags = re.findall(r'''<meta(?:\s+[a-zA-Z0-9:._-]+(?:=[a-zA-Z0-9:._-]+|="[^"]+"|='[^']+'))*?\s+property=['"]?og:video:tag['"]?(?:\s+[a-zA-Z0-9:._-]+(?:=[a-zA-Z0-9:._-]+|="[^"]+"|='[^']+'))*?\s+content=['"]?([^>'"]+?)['"]?\s*>'''
         , video_webpage, re.DOTALL | re.IGNORECASE);
-        video_tags = ", ".join(m)
         # description
         video_description = get_element_by_id("eow-description", video_webpage)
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/youtube.py
```diff
@@ -1072,6 +1072,9 @@ def add_dash_mpd(video_info):
         else:
             video_categories = None
 
+        m = re.findall(r'''<meta(?:\s+[a-zA-Z0-9:._-]+(?:=[a-zA-Z0-9:._-]+|="[^"]+"|='[^']+'))*?\s+property=['"]?og:video:tag['"]?(?:\s+[a-zA-Z0-9:._-]+(?:=[a-zA-Z0-9:._-]+|="[^"]+"|='[^']+'))*?\s+content=['"]?([^>'"]+?)['"]?\s*>'''
```
dstftw: Use `self._html_search_meta()` instead.

youtube_dl/extractor/youtube.py
```diff
@@ -1072,6 +1072,9 @@ def add_dash_mpd(video_info):
...
+        , video_webpage, re.DOTALL | re.IGNORECASE);
+        video_tags = ", ".join(m)
```
dstftw: I would prefer `tags` to be similar to `categories` i.e. a list instead of string.

dstftw:
It should be [documented](https://github.com/rg3/youtube-dl/blob/289bbb350e11ae1460db515560345088d7c58663/youtube_dl/extractor/common.py#L190).

youtube_dl/extractor/youtube.py
purdeaandrei: I can't do that, because youtube returns multiple meta tags, one per keyword. _html_search_meta() uses re.search under the hood, so it only finds the first instance.

dstftw: Can you name an example URL and add a test?

purdeaandrei: >$ curl https://www.youtube.com/watch?v=OPf0YbXqDm0 | grep og:video:tag

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   534    0   534    0     0   3581      0 --:--:-- --:--:-- --:--:--  3560        <meta property="og:video:tag" content="Uptown Funk">
        <meta property="og:video:tag" content="Mark">
        <meta property="og:video:tag" content="Ronson">
        <meta property="og:video:tag" content="feat.">
        <meta property="og:video:tag" content="Bruno">
        <meta property="og:video:tag" content="Mars">
        <meta property="og:video:tag" content="Pop">
        <meta property="og:video:tag" content="Columbia">
100  146k    0  146k    0     0   412k      0 --:--:-- --:--:-- --:--:--  412k


What kind of test do you want?

youtube_dl/extractor/youtube.py
purdeaandrei: done

purdeaandrei:
> It should be documented.
done

dstftw:
Thanks.

Issue #6069, opened 2015-06-24
f4m and hls using the external downloader


remitamine:
i'm asking if it is possible to make f4m and hls downloading the segments using the external downloader if the --externl-downloader option is used.
as i see in the code the F4mFD is using HttpFD and the NativeHlsFD use the youtube wrapper(compat_urllib_request).
it even possible to share the downloading part of the HlsFD and NativeHlsFD and the only difference is how to combine the segments together.
i made a simple script to download m3u8 segments and create a m3u8 file working with the local parts.
it depends on m3u8 module and the aria2 but it's possible to make it work with other downloaders.
[M3U8 downloader](https://gist.github.com/remitamine/2e0cbef7d50e8e7ef2d2)

Gaploid:
Any hope that it will be implemented? It will be so awesome to use full internet speed to download hls videos.

dneise:
Can noobs help?

sergeevabc:
There is a [feature request][1] to make HLS possible in Aria2.

[1]: https://github.com/aria2/aria2/issues/1271

Issue #6057, opened 2015-06-23
Viewster: uncomplete video file


remitamine:
when i download videos from the viewster the last part of the video is missing.
in the video player it show the right duration but when i reach to about the last 10-20% of the video the player fail and get closed.
i'm not sure but i think the problem is reletad to the f4m downloader may be it doesn't get the right number of segments to download.

yan12125:
Which URL are you trying to download?

remitamine:
this http://www.viewster.com/movie/1244-16928-001
and this http://www.viewster.com/movie/1244-16928-002

yan12125:
Both videos give "Oops, this content is not available in your country :(". Do you have a non-georestricted example?

Or, what's your location? I may try to find a proxy.

dstftw:
@yan12125 for me it works via tor through US exit node.

remitamine:
yes i download the videos with US IP.

yan12125:
Well, for me the downloaded file can be played in VLC to the last second. What player are you using? Did you get any error messages either in downloading or from the player?

My file is 571426074 bytes with MD5 b8321050ac8307bc58e15800d579e125. If you have an identical file, it could be the problem from your player rather than youtube-dl.

remitamine:
i'm using mpv.
i forget to say i stop the download and continue multiple times.
as i remember youtube-dl finished without an error but in the video player i didn't see what the terminal output in the last section.
i start thinking that the problem happen from youtube-dl beacause it happen 2 times.
i already removed the files but i will redownload it again and post the exact file size and md5 checksum.

remitamine:
i download the two files again.
for the first episode it stops playing after 2 seconds i think because of an error at the beginning of the download:
[download]   0.0% of ~630.72MiB at  932.60B/s ETA --:--:--ERROR: content too short (expected 1503094 bytes and served 135808)
but when i complete the download it continue like there is no error before.
for the second episode no error happened in the download process i stop the download multiple times(31.7% , 60.7% , 71.6%) and after it complete i play the video and it stops at 71%.
i put the output where the error happen in multiple players.
the output of mpv:
AV: 00:31:37 / 00:44:08 (71%) A-V:  0.000


Exiting... (End of file)

the output of ffplay after seeking to the last section:
Seek to 93% ( 0:41:04) of total duration ( 0:44:08)       B f=0/0   
[...]

PR #6392, opened 2015-07-28
Generalized fragmented media file downloader

dstftw:
* Adds a base file downloader class to be used by fragmented media file downloaders (e.g. f4m/m3u8 manifests).
* Rewrites `F4mFD` and `NativeHlsFD` in terms of `FragmentFD`.
* Adds generic progress output and resume for `NativeHlsFD`.

### DIFF
--- a/youtube_dl/downloader/f4m.py
+++ b/youtube_dl/downloader/f4m.py
@@ -7,8 +7,7 @@ import os
 import time
 import xml.etree.ElementTree as etree
 
-from .common import FileDownloader
-from .http import HttpFD
+from .fragment import FragmentFD
 from ..compat import (
     compat_urlparse,
     compat_urllib_error,
@@ -16,8 +15,6 @@ from ..compat import (
 from ..utils import (
     struct_pack,
     struct_unpack,
-    encodeFilename,
-    sanitize_open,
     xpath_text,
 )
 
@@ -226,15 +223,12 @@ def _add_ns(prop):
     return '{http://ns.adobe.com/f4m/1.0}%s' % prop
 
 
-class HttpQuietDownloader(HttpFD):
-    def to_screen(self, *args, **kargs):
-        pass
-
-
-class F4mFD(FileDownloader):
+class F4mFD(FragmentFD):
     """
     A downloader for f4m manifests or AdobeHDS.
     """
+
+    FD_NAME = 'f4m'
 
     def _get_unencrypted_media(self, doc):
         media = doc.findall(_add_ns('media'))
@@ -288,7 +282,7 @@ class F4mFD(FileDownloader):
     def real_download(self, filename, info_dict):
         man_url = info_dict['url']
         requested_bitrate = info_dict.get('tbr')
-        self.to_screen('[download] Downloading f4m manifest')
+        self.to_screen('[%s] Downloading f4m manifest' % self.FD_NAME)
         manifest = self.ydl.urlopen(man_url).read()
 
         doc = etree.fromstring(manifest)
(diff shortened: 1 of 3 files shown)


jaimeMF:
Looks good, nice work.

remitamine:
i test this branch because i like the idea and it is good.
there is actually two things i would like to consider they are a problems already present in the current master branch.
- when downloading a file there is a duplication, the size of temporary file is the same size of the segments except the last segment currently downloaded,i think it's better to combine the segments only at after downloading all the segments or combine them when the file is being downloaded but don't keep the segment files.
- the second is a problem i already report about it https://github.com/rg3/youtube-dl/issues/6057 it happen because the segments download didn't support resume(range header) for this case it should not keep the last partially downloaded segment or keep the segment in the memory and only write it when the segment download is finished.

there is one more thing:
- i request it in this issue https://github.com/rg3/youtube-dl/issues/6069 about using external downloader with hls native and f4m downloader.
it is simple to make(i already try it with aria2 and it is better in the speed) but the progess is not possible to be the same as the progress of f4m downloader(because it's hard to get the progress from external downloader), it can be like the old progress in the hls native downloader.

i will close this pull request https://github.com/rg3/youtube-dl/pull/6284 because it's now present in this pull request.

jaimeMF:
> when downloading a file there is a duplication, the size of temporary file is the same size of the segments except the last segment currently downloaded,i think it's better to combine the segments only at after downloading all the segments or combine them when the file is being downloaded but don't keep the segment files

Segments should be appended to the file as they are being downloaded, because it allows to directly play a livestream.
About keeping the segments: that's how we currently support resuming, it could be better to store the list of downloaded segments but it would make the code a bit more complex.

> the second is a problem i already report about it #6057 it happen because the segments download didn't support resume(range header) for this case it should not keep the last partially downloaded segment or keep the segment in the memory and only write it when the segment download is finished.

I haven't look into the problem in detail, but we're using `HttpFD` which supports resuming partial downloads.

>  i request it in this issue #6069 about using external downloader with hls native and f4m downloader. it is simple to make(i already try it with aria2 and it is better in the speed) but the progess is not possible to be the same as the progress of f4m downloader(because it's hard to get the progress from external downloader), it can be like the old progress in the hls native downloader.

It could be done. Since they are usually small files, is the improvement that great?

remitamine:
> I haven't look into the problem in detail, but we're using HttpFD which supports resuming partial downloads.

when i download a file using the f4m downloader when i resume the download i had to remove the last file manually before continue the download to get at the end a working file if i didn't do it the result file can't be played until the end.
i'm talking about the range because when i change the f4m downloader to work with aria2 it complains about getting a different range from the server.
> It could be done. Since they are usually small files, is the improvement that great?

yes it is, because for example i can pass all the links to aria2 and make it download them in parallel.
the download of playlist and videos with multiple parts(like chinese video sites) also can benefit from this change by letting aria2c handle the parallelism(using the --max-concurrent-downloads=N option) instead of adding more code to youtube-dl,this already requested here https://github.com/rg3/youtube-dl/issues/3746.

remitamine:
> I haven't look into the problem in detail, but we're using HttpFD which supports resuming partial downloads.

when i download a file using the f4m downloader when i resume the download i had to remove the last file manually before continue the download to get at the end a working file if i didn't do it the result file can't be played until the end.
i'm talking about the range because when i change the f4m downloader to work with aria2 it complains about getting a different range from the server.
> It could be done. Since they are usually small files, is the improvement that great?

yes it is, because for example i can pass all the links to aria2 and make it download them in parallel.
the download of playlist and videos with multiple parts(like chinese video sites) also can benefit from this change by letting aria2c handle the parallelism(using the --max-concurrent-downloads=N option) instead of adding more code to youtube-dl,this already requested here https://github.com/rg3/youtube-dl/issues/3746.

remitamine:
> Segments should be appended to the file as they are being downloaded, because it allows to directly play a livestream.
About keeping the segments: that's how we currently support resuming, it could be better to store the list of downloaded segments but it would make the code a bit more complex.

for m3u8 there is way to keep only the segments and you can still resume the download and play it as it's downloading but for f4m i don't know it's differnte.
for m3u9 by keeping a m3u8 file with relative file path to the local segments and the key if it's exist i try it before and it's working.
i already post a wrapper around youtube-dl with this feature in this issue: https://github.com/rg3/youtube-dl/issues/6069
about viewster https://github.com/jaimeMF said https://github.com/rg3/youtube-dl/issues/6057#issuecomment-126023517 that disabling the continuedl work, i'm using it now as temporary solution, but i'm thinking it's better to fix the problem, i post here https://github.com/rg3/youtube-dl/issues/6057#issuecomment-126129799 what i think is the problem in HttpFD but i don't know a lot about if some work on it can say if it implement the case(checking for return of content-range) or not.

jaimeMF:
@remitamine The best solution would be to make HttpFD recognize those situations, I don't know how hard it is.

@dstftw The PR looks good to me, feel free to merge it.

PR #6402, opened 2015-07-29
Added BASH script that organizes downloading videos

psprint:
The script has been developed, used and tested for 17 months. It is a very thoroughly integrated set of features that allow to manage youtube-dl downloads and to utilize cloud services. A training in writing compact, rock solid Bash code


dstftw:
That's too specific and out of scope of youtube-dl. If you want to contribute some concrete functionality you'll have to implement it in python.

psprint:
Bash is good to start, restart etc. processes. The script could fit into some "contrib" or similar directory in youtube-dl tree.

PR #6411, opened 2015-07-30
added ruv.is

iliasku:


dstftw:
What did you mean by this? By the way, `supportedsites.md` is autogenerated.

iliasku:
my apologies, didn't know it was autogenerated.
i just tried it on ruv.is and it worked, that's why i wanted to add it

jaimeMF:
We only add sites that have its own extractor, but the generic extractor works in different sites.

PR #6367, opened 2015-07-26
[iqiyi] fix 1080p parse

sceext2:
If you set `um` to `0`, *iqiyi* will only return `720p` video information. 
But if change `um` to `1`, *iqiyi* will return `1080p` video information. 

+ **`um=0`**
  
  ```
$ python -m youtube_dl --get-format --all-formats "http://www.iqiyi.com/v_19rrole0lg.html"
h6 - unknown
h5 - unknown
720p
h6 - unknown
h5 - unknown
720p
h6 - unknown
h5 - unknown
720p
h6 - unknown
h5 - unknown
720p
h6 - unknown
h5 - unknown
720p
h6 - unknown
h5 - unknown
720p
h6 - unknown
h5 - unknown
720p
h6 - unknown
h5 - unknown
720p
$ 
  ```

+ **`um=1`**
  
  ```
$ python -m youtube_dl --get-format --all-formats "http://www.iqiyi.com/v_19rrole0lg.html"
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
$ 
  ```

PR #6315, opened 2015-07-21
[iqiyi] add some format descriptions to format_id h1, h2, h3

sceext2:
Add some human-readable descriptions (`format`) to `format_id` `h1`, `h2`, `h3`. 

+ Before add
  ```
$ python -m youtube_dl --get-format --all-formats "http://www.iqiyi.com/v_19rrolq8ak.html"
h6 - unknown
h5 - unknown
h3 - unknown
h2 - unknown
h6 - unknown
h5 - unknown
h3 - unknown
h2 - unknown
h6 - unknown
h5 - unknown
h3 - unknown
h2 - unknown
h6 - unknown
h5 - unknown
h3 - unknown
h2 - unknown
h6 - unknown
h5 - unknown
h3 - unknown
h2 - unknown
h6 - unknown
h5 - unknown
h3 - unknown
h2 - unknown
h6 - unknown
h5 - unknown
h3 - unknown
h2 - unknown
h6 - unknown
h5 - unknown
h3 - unknown
h2 - unknown
$ 
  ```

+ After add
  ```
$ python -m youtube_dl --get-format --all-formats "http://www.iqiyi.com/v_19rrolq8ak.html"
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
h6 - unknown
h5 - unknown
720p
1080p
$ 
  ```

PR #6415, opened 2015-07-31
[iqiyi] Fix iqiyi (2015-07-30): update the md5 salt (enc key) and other improvements

sceext2:
iqiyi updated its flash player at 2015-07-30 and changed the md5 salt (enc key). 

According to iqiyi's habit, now the old salt is still usable. 
But soon (maybe tomorrow or several hours later) iqiyi will refuse the old salt. 
Then must use the new salt. 

Other improvements:
+ fix 1080p parse (#6367)
+ add some formats description (#6315)

### DIFF
--- a/youtube_dl/extractor/iqiyi.py
+++ b/youtube_dl/extractor/iqiyi.py
@@ -214,7 +214,7 @@ class IqiyiIE(InfoExtractor):
         return raw_data
 
     def get_enc_key(self, swf_url, video_id):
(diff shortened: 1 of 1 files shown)


sceext2:
This is out of date.

yan12125:
Could you split the enc key part into another PR?

sceext2:
@yan12125 
OK, it is #6533 .

PR #6437, opened 2015-08-02
[pornhub] Fix video url regular expression.

gebn:
PornHub seems to have subtly changed their JavaScript. Before, video URL strings were embedded directly in the video's `flashvars_*` object, but they are now assigned to variables of the form `player_quality_*`, which are then added to this object later under the relevant quality key.

The fix is to simply update the regex to match where these URL variables are initialised.


dstftw:
Cherry-picked in 754e70cf3e74218ae5d840985fbf07bbe274332a. Thanks.

PR #6439, opened 2015-08-02
[facebook] extract uploader

remitamine:
related to this issue https://github.com/rg3/youtube-dl/issues/4591.
for the does not provide the resolution for both formats(sd,hd)


dstftw:
Can you add a test?

remitamine:
i added the uploader value to the previous tests and they pass.

dstftw:
Thanks.

PR #6441, opened 2015-08-02
[twitch:vod] Parse start_time from 't'

haasn:
Eg. for VOD links like http://www.twitch.tv/gamesdonequick/v/9136645?t=14h29m15s

### DIFF
--- a/youtube_dl/extractor/twitch.py
+++ b/youtube_dl/extractor/twitch.py
@@ -8,12 +8,14 @@ import random
 from .common import InfoExtractor
 from ..compat import (
     compat_str,
+    compat_parse_qs,
     compat_urllib_parse,
     compat_urllib_request,
 )
 from ..utils import (
     ExtractorError,
     parse_iso8601,
+    parse_duration,
 )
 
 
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/twitch.py
```diff
@@ -216,6 +219,14 @@ def _real_extract(self, url):
             item_id, 'mp4')
         self._prefer_source(formats)
         info['formats'] = formats
+
+        info['start_time'] = None
+        parsed_url = compat_urllib_parse.urlparse(url)
+        for component in [parsed_url.fragment, parsed_url.query]:
```
jaimeMF: Using the time in the fragmet (like `...#t=5m10s`) doesn't work for me, does it work for you?

haasn: No idea, I've never used the official twitch client. Just copied the code from the youtube.py script and figured it would probably the same-ish.

jaimeMF:
Cherry picked in e704f87f869b98bbed56d7dd0fe27710306c8272, thanks!

Issue #5921, opened 2015-06-07
Proposal: A more robust solution for truncating file names


yan12125:
Currently in [this line](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L1239), ```info_dict['title']``` is truncated to 200 characters. This is first done by @phihag in 0599ef8c0827068184f78745a299d043927db4d9. This implementation can't handle wide characters correctly. An examples are #5740. I think YoutubeDL should switch to a utf-8 friendly implementation.

There's an implementation at [a stackoverflow answer](http://stackoverflow.com/a/13665637/3786245), although I haven't tested it. Any opinions?

dstftw:
As an addition: it still may not be enough since max filename limitation in general can be arbitrary (see #5908).

yan12125:
Another side note: on Windows file system encoding is not always utf-8, which make thinks more complicated. For example, on Traditional Chinese Windows file system uses big5 (cp950), where each Chinese character occupies 2 bytes rather than 3 bytes in utf-8.

Invertex:
This is still an issue, at least in Windows, and can be further exacerbated depending on the character length of the directory path that youtube-dl is outputting files in. `--restrict-filenames` doesn't trim it and there's no option to truncate the filename either still. A simple `-truncate ##` parameter would suffice just fine, or perhaps even an `auto` value that checks total file path length.

shillshocked:
Also an issue in Linux, and moderator keeps closing new issues on this without referencing or addressing the original issue.

wizzup:
I've got the same error message, add here for reference and seaching

https://github.com/ytdl-org/youtube-dl/issues/23702

Issue #5740, opened 2015-05-18
instagram: filenames are too long


pabs3:
```
$ youtube-dl --verbose http://instagram.com/schekobon/
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['--verbose', 'http://instagram.com/schekobon/']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.02.28
[debug] Python version 2.7.10rc1 - Linux-4.0.0-1-amd64-x86_64-with-debian-stretch-sid
[debug] exe versions: avconv 11.3-6, avprobe 11.3-6, rtmpdump 2.4
[debug] Proxy map: {}
[instagram:user] schekobon: Downloading page 1 
[instagram:user] schekobon: Downloading page 2 
[instagram:user] schekobon: Downloading page 3 
[instagram:user] schekobon: Downloading page 4 
[instagram:user] schekobon: Downloading page 5 
[instagram:user] schekobon: Downloading page 6 
[instagram:user] schekobon: Downloading page 7 
[instagram:user] schekobon: Downloading page 8 
[instagram:user] schekobon: Downloading page 9 
[instagram:user] schekobon: Downloading page 10 
[instagram:user] schekobon: Downloading page 11 
[instagram:user] schekobon: Downloading page 12 
[download] Downloading playlist: schekobon
[instagram:user] playlist schekobon: Collected 14 video ids (downloading 14 of them)
[download] Downloading video 1 of 14
[debug] Invoking downloader on u'https://scontent-lax.cdninstagram.com/hphotos-xfa1/t50.2886-16/11244209_893039364086936_414992731_n.mp4'
ERROR: unable to open for writing: [Errno 36] File name too long: 'Animation basics - timing and spasing , following arcs, tv paint 11 trial. Practice with a clock)))#art #animationbasics #animation #2danimation #study #clock #tvpaint #\xd1\x83\xd0\xbf\xd1\x80\xd0\xb0\xd0\xb6\xd0\xbd\xd0\xb5\xd0\xbd\xd0\xb8\xd1\x8f #\xd0\xb0\xd0\xbd\xd0\xb8\xd0\xbc\xd0\xb0\xd1\x86\xd0\xb8\xd1\x8f #\xd1\x87\xd0\xb0\xd1\x81\xd1\x8b ...-986928499285239423_1520751736.mp4.part'
Traceback (most recent call last):
[...]

PR #6442, opened 2015-08-03
[YoutubeDL] truncate long filenames taking in count the NAME_MAX and filesystem encoding

remitamine:
related to this issue https://github.com/rg3/youtube-dl/issues/5908.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -60,6 +60,7 @@ from .utils import (
     format_bytes,
     formatSeconds,
     GeoRestrictedError,
+    get_filesystem_encoding,
     int_or_none,
     ISO3166Utils,
     locked_file,
@@ -1681,8 +1682,6 @@ class YoutubeDL(object):
                 raise MaxDownloadsReached()
 
         info_dict['fulltitle'] = info_dict['title']
-        if len(info_dict['title']) > 200:
-            info_dict['title'] = info_dict['title'][:197] + '...'
 
         if 'format' not in info_dict:
             info_dict['format'] = info_dict['ext']
@@ -1740,6 +1739,23 @@ class YoutubeDL(object):
 
         if not ensure_dir_exists(sanitize_path(encodeFilename(filename))):
             return
+
+        filesystem_enc = get_filesystem_encoding()
+        max_filename_len = 200
+        if hasattr(os, 'pathconf') and hasattr(os, 'pathconf_names'):
+            try:
+                dn = os.path.dirname(sanitize_path(encodeFilename(filename)))
+                max_filename_len = int(os.pathconf(os.path.abspath(dn), os.pathconf_names['PC_NAME_MAX']))
+            except OSError as err:
+                self.report_error('the system does not support PC_NAME_MAX ' + compat_str(err))
+        if len((os.path.basename(filename) + '.part').encode(filesystem_enc)) > max_filename_len:
+            trun_len = max_filename_len - len(('....part.' + info_dict['ext']).encode(filesystem_enc))
+            while True:
+                try:
+                    info_dict['title'] = info_dict['title'].encode(filesystem_enc)[:trun_len].decode() + '...'
+                    break
+                except:
+                    trun_len -= 1
 
         if self.params.get('writedescription', False):
             descfn = replace_extension(filename, 'description', info_dict.get('ext'))


yan12125:
As I've mentioned in #5921, encodings should be taken into consideration, too.

remitamine:
i only follow the instuction related to issue where @dstftw say that the problem should be with NAME_MAX https://github.com/rg3/youtube-dl/issues/5908#issuecomment-109600112, for encoding i will try to find a solution, but i don't have a windows machie and a url that contain a long name to test with(max name for me is 255).
for the calculation i should consider the `.part'(5) and the ext like `.webm`(5).

yan12125:
The encoding problem exists on all platforms. For example, in Python ```len('中文')``` is 2, while on UTF-8 filesystems the corresponding filename length is 6. (Chinese characters are usually 3-bytes long in UTF-8). An problematic example can be found in #5740.
I didn't require you to handle encoding problems in this PR, too. Just a reminder that this PR does not solve the problem completely.

remitamine:
with this code i think it will calculate the lenght of the name according to file system encoding(please tell me if this is right), now i should figure how much needed for the file name length using the file system encoding.
```python
filesystem_enc = sys.getfilesystemencoding()
if filesystem_enc is None:
    filesystem_enc = 'utf-8'
max_filename_len = 200
if sys.platform.startswith('linux'):
    max_filename_len = int(os.pathconf(os.path.abspath(dn), 'PC_NAME_MAX'))
if len(info_dict['title'].encode(filesystem_enc)) > max_filename_len:
```

yan12125:
This approach cause problems in some cases. For example assume ```max_filename_len``` is 256, ```info_dict['title']``` is ```'a' * 252 + '中文'```, and ```filesystem_enc``` is utf-8. ```info_dict['title'].encode(filesystem_enc)[:max_filename_len - 3]``` would be ```b'a' * 252 + b'\xe4'```, which is not a valid UTF-8 string.

remitamine:
i put this code to handle this case it will try to decode the string if it can't it will raise an exception than i remove 1 byte and try again.
```
decoded = False
    while not decoded:
        try:
            info_dict['title'] = info_dict['title'].encode(filesystem_enc)[:trun_len].decode() + '...'
            decoded = True
        except:
            trun_len = trun_len - 1
```

yan12125:
A good idea. Just do not use bare ```except:```.

remitamine:
i know only about UnicodeEncodeError if there no other exception related to decoding i will change it to:
```python
except UnicodeEncodeError:
```

youtube_dl/YoutubeDL.py
```diff
@@ -1291,6 +1289,22 @@ def process_info(self, info_dict):
             self.report_error('unable to create directory ' + compat_str(err))
             return
 
+        filesystem_enc = sys.getfilesystemencoding()
+        if filesystem_enc is None:
+            filesystem_enc = 'utf-8'
+        max_filename_len = 200
+        if sys.platform.startswith('linux'):
```
dstftw: I would prefer testing with `hasattr` instead to avoid [possible breakages](https://github.com/rg3/youtube-dl/commit/22603348aa0b3e02c520589dea092507a04ab06a) I've learned about the hard way.

remitamine: what extactly the thing i need to check with hasattr.
both sys.getfilesystemencoding and sys.platform are available in python 2 and 3:
https://docs.python.org/2/library/sys.html#sys.getfilesystemencoding
https://docs.python.org/3/library/sys.html#sys.getfilesystemencoding
https://docs.python.org/2/library/sys.html#sys.platform
https://docs.python.org/3/library/sys.html#sys.platform

dstftw: `os.pathconf` and `os.pathconf_names`.

youtube_dl/YoutubeDL.py
```diff
@@ -1291,6 +1289,22 @@ def process_info(self, info_dict):
             self.report_error('unable to create directory ' + compat_str(err))
             return
 
+        filesystem_enc = sys.getfilesystemencoding()
```
dstftw: Use `utils.get_filesystem_encoding` instead.

youtube_dl/YoutubeDL.py
remitamine: os.pathconf and os.pathconf_names both available on unix and i use them only inside the condition to check for linux and because i can't test it on mac os i didn't add it.
and about test if os.pathconf_names contains PC_NAME_MAX i will add it now.
and i want to ask when i decode the bytes should i let python choose the encoding or use the same filesystem encoding that encode the string with.
```python
info_dict['title'] = info_dict['title'].encode(filesystem_enc)[:trun_len].decode() + '...'
```
or
```python
info_dict['title'] = info_dict['title'].encode(filesystem_enc)[:trun_len].decode(filesystem) + '...'
```

dstftw: >os.pathconf and os.pathconf_names both available on unix

Not necessarily, [they can be disabled](https://github.com/python/cpython/blob/9586a26986ab6fe8baac15d6db29b5e19c09ba65/Modules/posixmodule.c#L10483-L10489):
```
>>> import os
>>> dir(os.pathconf)
  File "<stdin>", line 1, in <module>
AttributeError: module object has no attribute 'pathconf'
```

remitamine: ok i wil add a test for them

remitamine:
added make the requested changes is there anything else

youtube_dl/YoutubeDL.py
```diff
@@ -1291,6 +1290,26 @@ def process_info(self, info_dict):
             self.report_error('unable to create directory ' + compat_str(err))
             return
 
+        filesystem_enc = get_filesystem_encoding()
+        if filesystem_enc is None:
+            filesystem_enc = 'utf-8'
+        max_filename_len = 200
+        if sys.platform.startswith('linux'):
```
dstftw: This check is not necessary now as you test pathconf anyway.

remitamine: this is right, so it can work on other systems if it is available.

dstftw: >it can work on other systems if it is available

That's the main point.

youtube_dl/YoutubeDL.py
```diff
@@ -1291,6 +1290,25 @@ def process_info(self, info_dict):
...
+        max_filename_len = 200
+        if hasattr(os, 'pathconf') and hasattr(os, 'pathconf_names'):
+            try:
+                max_filename_len = int(os.pathconf(os.path.abspath(dn), os.pathconf_names['PC_NAME_MAX']))
+            except OSError as err:
+                self.report_error('the system does not support PC_NAME_MAX ' + compat_str(err))
+        if len((info_dict['title'] + '.part.' + info_dict['ext']).encode(filesystem_enc)) > max_filename_len:
+            trun_len = max_filename_len - len(('....part.' + info_dict['ext']).encode(filesystem_enc))
+            decoded = False
+            while not decoded:
```
jaimeMF: why don't you do:
```python
while True:
    try:
       info_dict['title'] = ...
   except ...:
     trun_len -= 1
   break
```
?

jaimeMF:
Instead of checking the title, shouldn't it check the filename?, which will usually be longer than the title.

youtube_dl/YoutubeDL.py
jaimeMF: Sorry:
```python
while True:
    try:
       info_dict['title'] = ...
       break
   except ...:
     trun_len -= 1
```

remitamine: this is better

remitamine:
like this?
```python
filesystem_enc = get_filesystem_encoding()
if filesystem_enc is None:
    filesystem_enc = 'utf-8'
max_filename_len = 200
if hasattr(os, 'pathconf') and hasattr(os, 'pathconf_names'):
    try:
        max_filename_len = int(os.pathconf(os.path.abspath(dn), os.pathconf_names['PC_NAME_MAX']))
    except OSError as err:
        self.report_error('the system does not support PC_NAME_MAX ' + compat_str(err))
if len(os.path.basename(filename) + '.part').encode(filesystem_enc)) > max_filename_len:
    trun_len = max_filename_len - len(('....part.' + info_dict['ext']).encode(filesystem_enc))
    decoded = False
    while True:
        try:
            info_dict['title'] = info_dict['title'].encode(filesystem_enc)[:trun_len].decode() + '...'
            break
        except:
            trun_len -= 1
```

youtube_dl/YoutubeDL.py
```diff
@@ -1291,6 +1290,25 @@ def process_info(self, info_dict):
             self.report_error('unable to create directory ' + compat_str(err))
             return
 
+        filesystem_enc = get_filesystem_encoding()
+        if filesystem_enc is None:
+            filesystem_enc = 'utf-8'
```
dstftw: This is already done by `utils.get_filesystem_encoding`.

jaimeMF:
> like this?
> ```python
> if len((os.path.basename(filename) + '.part').encode(filesystem_enc)) > max_filename_len:
> ```
> i actually checking before for filename indirectly by concatenating the title and the ext

I think so.

remitamine:
it didn't work yet i think because this should be done after the name is truncated.
```python
info_dict['_filename'] = filename = self.prepare_filename(info_dict)
```

remitamine:
i moved the code here https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L575 with some modification, it work for the builtin downloader but it didn't work for aria2c even if i count the '.aria2', i don't know why even the length of the filename is the same as the NAME_MAX it fail.

remitamine:
this is the code after i move it:
```python
filesystem_enc = get_filesystem_encoding()
max_filename_len = 200
if hasattr(os, 'pathconf') and hasattr(os, 'pathconf_names'):
    try:
        max_filename_len = int(os.pathconf(os.path.dirname(os.path.abspath(filename)), os.pathconf_names['PC_NAME_MAX']))
    except OSError as err:
        self.report_error('the system does not support PC_NAME_MAX ' + compat_str(err))
if len((os.path.basename(filename) + '.part').encode(filesystem_enc)) > max_filename_len:
    trun_len = max_filename_len - len(('....part.aria2.' + info_dict['ext']).encode(filesystem_enc))
    while True:
        try:
            filename = filename.encode(filesystem_enc)[:trun_len].decode() + '....' + template_dict['ext']
            break
        except:
            trun_len -= 1
```

martinetd:
This is a bit of PR necromancy, but I'm still interested in that fix as I hit the bug today.

I'm not quite sure what's missing at this point, is there anything more than a rebase required to take this in?
Happy to help as possible.

remitamine:
moving the implementation to `prepare_filename` is not enough, because the filename can be modified after calling the function(like in the case of download dash formats(video+audio)), i don't won't to continue working on this for now, closing the issue might be better.
this is the last diff that i was working with to move the code to `prepare_filename`(a bit messy):
```diff
diff --git a/youtube_dl/YoutubeDL.py b/youtube_dl/YoutubeDL.py
index 807cff59a..7f33348b1 100755
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -703,7 +703,32 @@ class YoutubeDL(object):
             # to workaround encoding issues with subprocess on python2 @ Windows
             if sys.version_info < (3, 0) and sys.platform == 'win32':
                 filename = encodeFilename(filename, True).decode(preferredencoding())
-            return sanitize_path(filename)
+            filename = sanitize_path(filename)
+
+            filesystem_enc = get_filesystem_encoding()
+            max_filename_len = 200
+            if hasattr(os, 'pathconf') and hasattr(os, 'pathconf_names'):
+                try:
+                    max_filename_len = int(os.pathconf(os.getcwd(), os.pathconf_names['PC_NAME_MAX'])) - 1
+                except OSError as err:
+                    self.report_error('the system does not support PC_NAME_MAX ' + compat_str(err))
+
+            suff_without_part = '....' + info_dict['ext']
+            len_suff_with_part = len(suff_without_part + '.part')
+            base_filename = os.path.basename(filename)
+            part_filename = (base_filename + '.part').encode(filesystem_enc)
+            part_filename_len = len(part_filename)
+            if part_filename_len > max_filename_len:
+                b_trun_len = max_filename_len - len_suff_with_part
+                while True:
+                    try:
+                        trun_len = len(part_filename[:b_trun_len].decode())
+                        break
+                    except UnicodeDecodeError:
+                        b_trun_len -= 1
+                filename = filename[:-(len(base_filename) - trun_len)] + suff_without_part
+
+            return filename
         except ValueError as err:
             self.report_error('Error in output template: ' + str(err) + ' (encoding: ' + repr(preferredencoding()) + ')')
             return None
@@ -1740,23 +1765,6 @@ class YoutubeDL(object):
         if not ensure_dir_exists(sanitize_path(encodeFilename(filename))):
             return
 
-        filesystem_enc = get_filesystem_encoding()
-        max_filename_len = 200
-        if hasattr(os, 'pathconf') and hasattr(os, 'pathconf_names'):
-            try:
-                dn = os.path.dirname(sanitize_path(encodeFilename(filename)))
-                max_filename_len = int(os.pathconf(os.path.abspath(dn), os.pathconf_names['PC_NAME_MAX']))
-            except OSError as err:
-                self.report_error('the system does not support PC_NAME_MAX ' + compat_str(err))
-        if len((os.path.basename(filename) + '.part').encode(filesystem_enc)) > max_filename_len:
-            trun_len = max_filename_len - len(('....part.' + info_dict['ext']).encode(filesystem_enc))
-            while True:
-                try:
-                    info_dict['title'] = info_dict['title'].encode(filesystem_enc)[:trun_len].decode() + '...'
-                    break
-                except:
-                    trun_len -= 1
-
         if self.params.get('writedescription', False):
             descfn = replace_extension(filename, 'description', info_dict.get('ext'))
             if self.params.get('nooverwrites', False) and os.path.exists(encodeFilename(descfn)):
```

yan12125:
Thanks for all the work. Maybe using a hash filename in downloaders and postprocessors can fix everything.

remitamine:
> Maybe using a hash filename in downloaders and postprocessors can fix everything.

yes, it should fix the remaining problem.

mbnoimi:
This solution doesn't fix the problem!
I'm still getting an error of **File name too long** of the following link:

```
youtube-dl --newline -R 999 -i --all-subs -o "/home/Ahmet/%(title)s.%(ext)s" -f 18 -v --ignore-config --hls-prefer-native "https://www.youtube.com/attribution_link?a=yHtgUtfX2yhQ9ODk&u=/watch%3Fv%3DUJs-RFmPjoY%26feature%3Dem-uploademail"
```
Error message:
```
[Sat Feb 24 12:21:31 2018] [debug] System config: []
[Sat Feb 24 12:21:31 2018] [debug] User config: []
[Sat Feb 24 12:21:31 2018] [debug] Custom config: []
[Sat Feb 24 12:21:31 2018] [debug] Command-line args: [u'--newline', u'-R', u'999', u'-i', u'--all-subs', u'-o', u'/home/Ahmet/%(title)s.%(ext)s', u'-f', u'18', u'-v', u'--ignore-config', u'--hls-prefer-native', u'https://www.youtube.com/attribution_link?a=yHtgUtfX2yhQ9ODk&u=/watch%3Fv%3DUJs-RFmPjoY%26feature%3Dem-uploademail']
[Sat Feb 24 12:21:31 2018] [debug] Encodings: locale UTF-8, fs UTF-8, out None, pref UTF-8
[Sat Feb 24 12:21:31 2018] [debug] youtube-dl version 2018.02.22
[Sat Feb 24 12:21:31 2018] [debug] Python version 2.7.12 (CPython) - Linux-4.10.0-38-generic-x86_64-with-LinuxMint-18.3-sylvia
[Sat Feb 24 12:21:31 2018] [debug] exe versions: ffmpeg 2.8.11-0ubuntu0.16.04.1, ffprobe 2.8.11-0ubuntu0.16.04.1
[Sat Feb 24 12:21:31 2018] [debug] Proxy map: {}
[Sat Feb 24 12:21:31 2018] WARNING: video doesn't have subtitles
[Sat Feb 24 12:21:31 2018] ERROR: unable to open for writing: [Errno 36] File name too long: '/home/Ahmet/\xd8\xa7\xd9\x86\xd8\xa7\xd8\xb4\xd9\x8a\xd8\xaf \xd8\xa7\xd9\x84\xd8\xb1\xd9\x88\xd8\xb6\xd8\xa9 - \xd8\xaa\xd8\xb9\xd9\x84\xd9\x8a\xd9\x85 \xd8\xa7\xd9\x84\xd8\xa7\xd8\xb7\xd9\x81\xd8\xa7\xd9\x84 - \xd8\xb9\xd8\xa7\xd9\x84\xd9\x85 \xd8\xa7\xd9\x84\xd8\xad\xd9\x8a\xd9\x88\xd8\xa7\xd9\x86\xd8\xa7\xd8\xaa (1)- \xd8\xa7\xd9\x84\xd8\xa7\xd8\xb3\xd8\xaf Learn land and sea animals For Kids -Lion.mp4.part'
[Sat Feb 24 12:21:31 2018] Traceback (most recent call last):
[Sat Feb 24 12:21:31 2018]   File "/home/Ahmet/.config/youtube-dlg/youtube-dl/youtube_dl/downloader/http.py", line 249, in download
[Sat Feb 24 12:21:31 2018]     ctx.tmpfilename, ctx.open_mode)
[Sat Feb 24 12:21:31 2018]   File "/home/Ahmet/.config/youtube-dlg/youtube-dl/youtube_dl/utils.py", line 457, in sanitize_open
[Sat Feb 24 12:21:31 2018]     stream = open(encodeFilename(filename), open_mode)
[Sat Feb 24 12:21:31 2018] IOError: [Errno 36] File name too long: '/home/Ahmet/\xd8\xa7\xd9\x86\xd8\xa7\xd8\xb4\xd9\x8a\xd8\xaf \xd8\xa7\xd9\x84\xd8\xb1\xd9\x88\xd8\xb6\xd8\xa9 - \xd8\xaa\xd8\xb9\xd9\x84\xd9\x8a\xd9\x85 \xd8\xa7\xd9\x84\xd8\xa7\xd8\xb7\xd9\x81\xd8\xa7\xd9\x84 - \xd8\xb9\xd8\xa7\xd9\x84\xd9\x85 \xd8\xa7\xd9\x84\xd8\xad\xd9\x8a\xd9\x88\xd8\xa7\xd9\x86\xd8\xa7\xd8\xaa (1)- \xd8\xa7\xd9\x84\xd8\xa7\xd8\xb3\xd8\xaf Learn land and sea animals For Kids -Lion.mp4.part'
[Sat Feb 24 12:21:31 2018] 
```

PR #6501, opened 2015-08-08
add option to make a lite version without porn site extractors

remitamine:
related to this issue https://github.com/rg3/youtube-dl/issues/6497.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,6 +92,9 @@ youtube-dl.fish: youtube_dl/*.py youtube_dl/*/*.py devscripts/fish-completion.in
 
 fish-completion: youtube-dl.fish
 
+remove-porn:
+	python devscripts/remove-porn.py
+
 youtube-dl.tar.gz: youtube-dl README.md README.txt youtube-dl.1 youtube-dl.bash-completion youtube-dl.zsh youtube-dl.fish
 	@tar -czf youtube-dl.tar.gz --transform "s|^|youtube-dl/|" --owner 0 --group 0 \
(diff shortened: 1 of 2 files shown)


dstftw:
That's very very brittle. Any change to the layout may break youtube-dl completely (just like gentoo no-offensive breaks with every other release). You also forget that removed extractors may be referenced from another place apart from generix extractor with `self.url_result` or even worse with arbitrary code.

remitamine:
this just a try to make an automated solution for what i request:
> simple way to delete them.

i will list the ways i think about how to do this:
- script to remove them(as the one in this pull request).
- move them to a separate folder(so it can just deleted) and also the imports of this type of extractors(`special/__init__.py`) and than just place an import like this:
```
try:
    from .special import *
```
- move them to another repo and in at the time to release a new version make two version one from the core repo only and one merged with the separate repo.
if nothing is accptable this pull request will be closed and i will left with only one last option making a personal fork and from time to time merge the latest changes from the upsteam(just to provide this option and i won't bother you anymore).
i will continue to support the project(when i come back after 20 days).

oskar456:
Sorry for probably dumb question, but what is the reason to delete some functionality in the first place? Even without deleting the porn site extractors you don't get any porn, unless you ask for it. If you ask for it, chances are you get them utilizing the generic extractor. Does that mean you want to delete generic extractor as well?

remitamine:
i didn't remove the generic extractor i remove the parts of generic extractor that depend on porn site extractors other the python interpreter will raise an error.

PR #6509, opened 2015-08-09
[playtvak] Initial support for videos from Mafra group servers

oskar456:
Support for videos in articles from idnes.cz, lidovky.cz, metro.cz, as
well as video and live streams from playtvak.cz

Signed-off-by: Ondřej Caletka <ondrej@caletka.cz>

### DIFF
--- a/youtube_dl/extractor/playtvak.py
+++ b/youtube_dl/extractor/playtvak.py
@@ -1,0 +1,127 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from ..compat import (
+    compat_urlparse,
+    compat_urllib_parse,
+)
+from ..utils import ExtractorError
+
+
+def _extract_json(code):
+    return re.sub(
+        r'(?s)^VideoPlayer.data\("", ({.*})\);?\s*?(?://[^\n]*)*$', r'\1', code)
+
+
+class PlaytvakIE(InfoExtractor):
+    _VALID_URL = r'https?://.*?(playtvak|idnes|lidovky|metro)\.cz/.*\?c=(?P<id>[A-Z][0-9]{6}_[0-9]{6}_.*)'
+    _TESTS = [{
+        'url': 'http://www.playtvak.cz/vyzente-vosy-a-srsne-ze-zahrady-dn5-/hodinovy-manzel.aspx?c=A150730_150323_hodinovy-manzel_kuko',
+        'md5': '4525ae312c324b4be2f4603cc78ceb4a',
+        'info_dict': {
+            'id': 'A150730_150323_hodinovy-manzel_kuko',
+            'ext': 'mp4',
+            'title': 'Vyžeňte vosy a sršně ze zahrady',
+            'thumbnail': 'http://oidnes.cz/15/074/mobil/KUK5cea00_010hodmanel58154.jpg',
+            'description': 'Málo co kazí atmosféru venkovního posezení tak jako neustálé bzučení kolem hlavy.  Vyzkoušejte náš lapač a odpuzovač vos a sršňů.',
+        }
+    }, {  # live video test
+        'url': 'http://slowtv.playtvak.cz/planespotting-0pr-/planespotting.aspx?c=A150624_164934_planespotting_cat',
+        'info_dict': {
+            'id': 'A150624_164934_planespotting_cat',
+            'ext': 'flv',
+            'title': 're:^Přímý přenos iDNES.cz [0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}$',
+            'thumbnail': 'http://data.idnes.cz/soubory/servisni-play-porady/89A150630_ACEK_026_VIDEOPLAYER-STREA.PNG',
+            'description': 'Sledujte provoz na ranveji Letiště Václava Havla v Praze',
+            'is_live': True,
+        },
+        'params': {
+            'skip_download': True,  # requires rtmpdump
+        },
+    }, {  # idnes.cz
+        'url': 'http://zpravy.idnes.cz/pes-zavreny-v-aute-rozbijeni-okynek-v-aute-fj5-/domaci.aspx?c=A150809_104116_domaci_pku',
+        'md5': '819832ba33cd7016e58a6658577fe289',
+        'info_dict': {
+            'id': 'A150809_104116_domaci_pku',
+            'ext': 'mp4',
+            'title': 'Zavřeli jsme mraženou pizzu do auta. Upekla se',
+            'thumbnail': 'http://i.idnes.cz/15/081/vidw/SHA5d1786_pizzaauto.jpg',
+            'description': 'Na sociálních sítích se objevila výzva, aby lidé, kteří v horkých letních dnech uvidí v zaparkovaném autě zavřeného psa, neváhali rozbít okénko. Zastánci tohoto postoje argumentují zdravím zvířete, které v dusnu může zkolabovat. Policie doporučuje nejprve volat tísňovou linku.',
+        }
+    }, {  # lidovky.cz
+        'url': 'http://www.lidovky.cz/dalsi-demonstrace-v-praze-o-migraci-duq-/video.aspx?c=A150808_214044_ln-video_ELE',
+        'md5': 'c7209ac4ba9d234d4ad5bab7485bcee8',
+        'info_dict': {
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/playtvak.py
```diff
@@ -0,0 +1,86 @@
...
+    }]
+
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        xmlinfourl = self._html_search_regex(r'Misc.videoFLV\({ data: "([^"]+)"', webpage, 'xmlinfourl')
+        xmlinfo = self._download_xml(xmlinfourl, video_id)
+
+        item = xmlinfo.find('./items/item[title]')
```
dstftw: This is not supported in python 2.6. Use `utils.find_xpath_attr` instead.

youtube_dl/extractor/playtvak.py
```diff
@@ -0,0 +1,86 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..compat import compat_urlparse
+
+
+class PlaytvakIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:.*\.)?(playtvak|idnes|lidovky|metro)\.cz/.*\?c=(?P<id>.*)'
```
dstftw: Do not capture group if you don't use it. `id` must match at least one character.

youtube_dl/extractor/playtvak.py
dstftw: This is not supported in python 2.6. Use `utils.find_xpath_attr` instead.

oskar456:
Thanks @dstftw, hopefully fixed.

youtube_dl/extractor/playtvak.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        xmlinfourl = self._html_search_regex(r'Misc.videoFLV\({ data: "([^"]+)"', webpage, 'xmlinfourl')
+        xmlinfo = self._download_xml(xmlinfourl, video_id)
+
+        item = None
+        for i in xmlinfo.findall('./items/item'):
+            if i.find('title') is not None:
```
dstftw: It's probably better to test for `type` tag to be equal to `video`.

oskar456: Well, it can be also a livevideo (see second test). From my observation, every video has a title.

youtube_dl/extractor/playtvak.py
```diff
@@ -0,0 +1,95 @@
...
+                item = i
+                break
+        if item is None:
+            raise ExtractorError('No suitable stream found')
+        title = xpath_text(item, './title')
```
dstftw: For lives `self._live_title` should be use.

youtube_dl/extractor/playtvak.py
```diff
@@ -0,0 +1,95 @@
...
+            'thumbnail': 'http://i.idnes.cz/15/081/vidw/PID5d1d52_vandas3.jpg',
+            'description': 'Desítky lidí se sešly v Praze na protest proti imigrantům. Současně probíhala i demonstrace na jejich podporu. Na Staroměstském náměstí vystoupil i předseda dělnické strany Tomáš Vandas a kontroverzní slovenský politik Marian Kotleba. Dalšího slovenského nacionalistu Mariána Magáta odvedla policie.',
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        xmlinfourl = self._html_search_regex(r'Misc.videoFLV\({ data: "([^"]+)"', webpage, 'xmlinfourl')
+        xmlinfo = self._download_xml(xmlinfourl, video_id)
```
dstftw: This is also available as JSON if `&type=js` is appended to the query. With JSON extra webm format is served.

oskar456: I'll look at that. What browser/OS do you use? I get XML playlist both on FIrefox and Chromium on Linux.

oskar456:
Added `self._live_title` for live streams, changed the check to item type `video` or `livevideo`.

I'll look onto parsing the JS-type playlist some time later, it will probably require to drop everything and start over :)

oskar456:
I have rewritten the extractor to use the JS-based playlist. There are new medium format quality streams in `webm` format and there is also HLS support for live streams, as an alternative to RTMP.

However, at least the planespotting stream (second test) uses non-standard MP3 audio, which get corrupted by hardcoded `ffmpeg` option `aac_adtstoaac`. That's why I lowered HLS stream preference.

oskar456:
Fixed flake8 errors.

dstftw:
Broken on python 2:
```
py27yt http://www.playtvak.cz/vyzente-vosy-a-srsne-ze-zahrady-dn5-/hodinovy-manzel.aspx?c=A150730_150323_hodinovy-manzel_kuko --wri
te-pages
[Playtvak] A150730_150323_hodinovy-manzel_kuko: Downloading webpage
[Playtvak] Saving request to A150730_150323_hodinovy-manzel_kuko_http_-_www.playtvak.cz_vyzente-vosy-a-srsne-ze-zahrady-dn5-_hodinovy-manzel.aspxc=A150730_15032
3_hodinovy-manzel_kuko.dump
  File "C:\python\python2710\lib\runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\python\python2710\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\Dev\git\youtube-dl\master\youtube_dl\__main__.py", line 19, in <module>
    youtube_dl.main()
  File "youtube_dl\__init__.py", line 410, in main
    _real_main(argv)
  File "youtube_dl\__init__.py", line 400, in _real_main
  File "youtube_dl\YoutubeDL.py", line 1653, in download
  File "youtube_dl\YoutubeDL.py", line 655, in extract_info
    ie_result = ie.extract(url)
  File "youtube_dl\extractor\common.py", line 287, in extract
  File "youtube_dl\extractor\playtvak.py", line 74, in _real_extract
    newquery = compat_urlparse.urlencode(qs, True)
AttributeError: 'module' object has no attribute 'urlencode'
```

oskar456:
Oh s**t ;)
I fixed it using undocumented python2 urllib.unparse. Hope this works even with 2.6 (don't have such old python by hand).

dstftw:
I suggest you not to amend the commits but keep them separate since it's hard to track changes.

oskar456:
Sorry about that. Only change was on line 75, utilizing `compat_urllib_parse`. Tests now pass on both python flavours.

youtube_dl/extractor/playtvak.py
```diff
@@ -0,0 +1,127 @@
...
+from ..compat import (
+    compat_urlparse,
+    compat_urllib_parse,
+)
+from ..utils import ExtractorError
+
+
+def _extract_json(code):
+    return re.sub(
+        r'(?s)^VideoPlayer.data\("", ({.*})\);?\s*?(?://[^\n]*)*$', r'\1', code)
```
dstftw: What's the purpose of regex part after `;`? Can you name any example?

oskar456: Copied from [`strip_jsonp` function](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/utils.py#L1622). It should catch and drop anything after the interesting json.

youtube_dl/extractor/playtvak.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class PlaytvakIE(InfoExtractor):
+    _VALID_URL = r'https?://.*?(playtvak|idnes|lidovky|metro)\.cz/.*\?c=(?P<id>[A-Z][0-9]{6}_[0-9]{6}_.*)'
```
dstftw: Is there any example URL for metro.cz?

dstftw:
Thanks, merged.

youtube_dl/extractor/playtvak.py
oskar456: They mostly embed youtube videos, but few of them share the same publication pattern, [like this.](http://www.metro.cz/video-pod-billboardem-se-na-vltavske-roztocil-kolotoc-deti-vozil-jen-par-hodin-1hx-/metro-extra.aspx?c=A141111_173251_metro-extra_row)

PR #6518, opened 2015-08-10
[youtube] Use the first v= argument in the URL

puckipedia:
It seems that youtube-dl prefers the second (yes, I tested it) v= argument in an URL, while python `parse_qs` and the youtube site itself seem to prefer the first one, which could cause discrepancies in the shown video. This could cause unwanted rickrolling, and who wants that?

This seems to be caused due to the `(?:.*?&)` part of the regex, that does the pre-`v=` arguments handling, is greedy and causes it to match the first `v=`, while the second match is then used to match the plain `v=` afterwards.


dstftw:
Thanks.

PR #6523, opened 2015-08-11
Use `{title}` style for --output templates rather than `%(title)s`

cousteaulecommandant:
Currently youtube-dl uses Python's native string interpolation for automated output file naming.  For example, `%(title)s` will be replaced with the video's title.  However, this syntax is ugly and counter-intuitive, specially for people who are not familiar with Python but still want to use this program.

This patch replaces the `%` interpolation with str.format() formatting, so that instead of `%(title)s` the user can just write `{title}`, which in my opinion looks quite better.  Plus it seems to be the preferred formatting method nowadays and it has more options.

Backwards compatibility is provided by replacing all `%(whatever)s` in the template string with `{whatever}` using a regex, and `%%` with `%`; however this is a temporary (and ugly) workaround that can probably be removed in a future when users have gotten used to the new format.

This change requires Python 2.6 or greater to work, since `str.format()` is new in version 2.6 (if that version can be considered "new" at all nowadays).

PS: I moved the call to `sanitize_path()` AFTER formatting the file name; it makes little sense to sanitize a format string rather than an actual file name/path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,17 +114,26 @@ which means you can modify it, redistribute it or use it however you like.
 ## Filesystem Options:
     -a, --batch-file FILE            File containing URLs to download ('-' for stdin)
     --id                             Use only video ID in file name
-    -o, --output TEMPLATE            Output filename template. Use %(title)s to get the title, %(uploader)s for the uploader name, %(uploader_id)s for the uploader
-                                     nickname if different, %(autonumber)s to get an automatically incremented number, %(ext)s for the filename extension, %(format)s for
-                                     the format description (like "22 - 1280x720" or "HD"), %(format_id)s for the unique id of the format (like YouTube's itags: "137"),
-                                     %(upload_date)s for the upload date (YYYYMMDD), %(extractor)s for the provider (youtube, metacafe, etc), %(id)s for the video id,
-                                     %(playlist_title)s, %(playlist_id)s, or %(playlist)s (=title if present, ID otherwise) for the playlist the video is in,
-                                     %(playlist_index)s for the position in the playlist. %(height)s and %(width)s for the width and height of the video format.
-                                     %(resolution)s for a textual description of the resolution of the video format. %% for a literal percent. Use - to output to stdout.
-                                     Can also be used to download to a different directory, for example with -o '/my/downloads/%(uploader)s/%(title)s-%(id)s.%(ext)s' .
-    --autonumber-size NUMBER         Specify the number of digits in %(autonumber)s when it is present in output filename template or --auto-number option is given
+    -o, --output TEMPLATE            Output filename template. Use 
+                                     {title} to get the title, 
+                                     {uploader} for the uploader name, {uploader_id} for the uploader nickname if different, 
+                                     {autonumber} to get an automatically incremented number, 
+                                     {ext} for the filename extension, 
+                                     {format} for the format description (like "22 - 1280x720" or "HD"), 
+                                     {format_id} for the unique id of the format (like YouTube's itags: "137"), 
+                                     {upload_date} for the upload date (YYYYMMDD), 
+                                     {extractor} for the provider (youtube, metacafe, etc), 
+                                     {id} for the video id, 
+                                     {playlist_title}, {playlist_id}, or {playlist} (=title if present, ID otherwise) for the playlist the video is in, 
+                                     {playlist_index} for the position in the playlist, 
+                                     {height} and {width} for the width and height of the video format, 
+                                     {resolution} for a textual description of the resolution of the video format, 
+                                     {{ }} %% for literal braces and percent. 
+                                     Use - to output to stdout. 
+                                     Can also be used to download to a different directory, for example with -o '/my/downloads/{uploader}/{title}-{id}.{ext}' .
+    --autonumber-size NUMBER         Specify the number of digits in {autonumber} when it is present in output filename template or --auto-number option is given
     --restrict-filenames             Restrict filenames to only ASCII characters, and avoid "&" and spaces in filenames
-    -A, --auto-number                [deprecated; use  -o "%(autonumber)s-%(title)s.%(ext)s" ] Number downloaded files starting from 00000
+    -A, --auto-number                [deprecated; use  -o '{autonumber}-{title}.{ext}' ] Number downloaded files starting from 00000
     -t, --title                      [deprecated] Use title in file name (default)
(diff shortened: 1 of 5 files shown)


cousteaulecommandant:
I made this as 3 separate commits, one per file, because I'm not quite an expert with git so I edited and committed the files directly from GitHub; but if they can be squashed into a single commit that'd be nice.

dstftw:
`test_YoutubeDL.py` now fails.

cousteaulecommandant:
My bad. :(  Turns out it's not a good idea to use "".format(**d) with a defaultdict.  Using string.Formatter().vformat() now, but now the solution doesn't look as elegant as before, I'm starting to stop liking it.

jaimeMF:
I prefer the `{ }` syntax, I think we should print a warning if the old syntax is used.
There are other places with templates to update: https://github.com/rg3/youtube-dl/blob/369e195a44530b43bfd0e95087c9e1a238d91184/youtube_dl/__init__.py#L192-L199.

cousteaulecommandant:
In order to print a warning, I think the easiest way would be checking whether the template string has changed after the backwards compatibility hack, i.e.

    tmpl2 = re.sub(r'(?<!%)((?:%)*)\1%\(([^)]*)\)s', r'\1{\2}', tmpl).replace('%%','%')
    if tmpl != tmpl2:
        print "%(key)s formatting style is deprecated; please use {key} style instead"
        tmpl = tmpl2

However this should be done only once, maybe right after parsing the options.  If it's done where I'm doing it right now, the user would get one warning per downloaded file.  (Plus it does the replacement multiple times which doesn't feel OK)

cousteaulecommandant:
Another idea I have in mind, if this PR gets accepted, is to modify the behavior to take advantage of format specifiers.  For example, `{autonumber:03}` is more convenient than doing `--autonumber-size=3` (although the label without `:03` should default to the current behavior, which could be done with a custom int-like class).
Another example: if upload dates are changed to a `datetime.date` object, users could specify arbitrary date formats such as ` {upload_date:%Y}` to write the year, or ` {upload_date:%Y/%m}` if the user wants to create subdirectories for each year and month.
(This would require the ` sanitize` method be applied only to string members and not other object types; otherwise they'd lose the ability to be formatted.)

jaimeMF:
> Another idea I have in mind, if this PR gets accepted, is to modify the behavior to take advantage of format specifiers. For example, {autonumber:03} is more convenient than doing --autonumber-size=3 (although the label without :03 should default to the current behavior, which could be done with a custom int-like class).

I would like that, it could also be used for playlist_index. To reduce complexity I would even remove the default behaviour of using 5 digits.

> Another example: if upload dates are changed to a datetime.date object, users could specify arbitrary date formats such as {upload_date:%Y} to write the year, or {upload_date:%Y/%m} if the user wants to create subdirectories for each year and month.

Using folder per date is a features some people may find useful, but we would need to make sure that it defaults to the current format.

cousteaulecommandant:
> Using folder per date is a features some people may find useful, but we would need to make sure that it defaults to the current format.

Python's datetime.date defaults to YYYY-mm-dd (standard ISO format) rather than YYYYmmdd, would that be a problem?
(Personally I like the one with dashes more; the one without dashes could still be achieved with `{upload_date:%Y%m%d}`)

cousteaulecommandant:
If there's still any interest in this functionality (although the PR has long gone "rotten" and was unmergeable anyway) I could give it a try to re-create it from scratch on the newest master.

dirkf:
I hadn't seen this PR which pre-dates the start of my earliest first PR review.

The PR was closed by some unfathomable weirdness in GH, not intentionally.

However, it would probably be a better use of time to make yt-dl's template formatting match a larger subset of what _yt-dlp_ has implemented, extending the original syntax but not changing to `{}`.

cousteaulecommandant:
Oh, that's weird.

Yes; in all fairness this was an OLD pr; it even predates the time when this repo was closed and reopened (as well as all its forks; maybe that has to do with why this got closed). 

Back in the day this was an interesting PR because it replaced Python's old (and rather ugly) `%(...)s` syntax with its newer (and more readable and versatile) `{...}`, which also allows additional formatting options (such as the `{upload_date:%Y%m%d}` example).  But nowadays I don't know how the features in yt-dlp compare with what my idea would be able to do; I had a look at it and saw that it introduces quite a few new syntax features.  I might have another look at it.

cousteaulecommandant:
After having a more in-depth look at the --output format in yt-dlp, I see an issue: it is basically implementing a custom formatter loosely inspired in Python's %-formatting rather than using Python's native support.

The reason I pushed for a `{field}` syntax in my PR was that it was already natively supported by Python (as was the % syntax), so I wouldn't need to reinvent the wheel writing a custom formatter (in the PR, I'm not parsing the format string looking for `{` and `}` delimiters at all; I'm just relying on Python's native string formatter).  My thought was that the simpler the PR, the more likely to be merged (and the easier it would be to maintain afterwards).

So, having a look at yt-dlp's formatting options (which are quite messy), I can see that the following would be supported out of the box (showing [yt-dlp]→[this PR]):
* ✔️ Integer formatting: `%(NAME)05d` → `{NAME:05}`
* ✔️ Object traversal: `%(subtitles.en.0.ext)s` → `{subtitles[en][0][ext]}`.
    - Also supports object attributes in addition to dict/list indexes (e.g., `{foo[bar][0].baz}` would give you `infodict["foo"]["bar"][0].baz`).
    - But negative indexes such as `[-1]` or slices such as `[3:7:-1]` are not supported.
* ✔️ Datetime formatting: `%(upload_date>%Y-%m-%d)s` → `{upload_date:%Y-%m-%d}`

The following are not, and would require quite a bit of hacking to be implemented (resulting in harder to maintain code):
* ❌ Build dictionaries with specific keys (`%(foo:{bar,baz})j`).
* ❌ Alternatives and replacement, such as `%(title,id|Unknown)s`.
* ❌ Addition, such as `%(n_entries+1-playlist_index)d`.

The following could be supported with some minor hacking (mostly using Python's native formatting features with some minor additions):
* ❓ Conversions, for example, JSON output: `%(NAME)j` could be implemented as `{NAME!j}` by implementing a custom `!j` conversion.  This may apply to other conversions (such as `!S` to sanitize file name, `!D` to add suffixes, etc).

So, bottom line, yt-dlp --output is too complicated to implement and would require a custom parser, whereas my goal in the PR was to leverage the existing Python features.

PR #6533, opened 2015-08-12
 [iqiyi] update md5 salt (2015-08-10)

sceext2:
Update the md5 salt for iqiyi's player `MainPlayer_5_2_26_c3_3_7_1.swf` (updated at 2015-08-10).


yan12125:
Thanks!

PR #6554, opened 2015-08-14
Updated line numbers in the fragment portion of README links.

reddraggone9:
These two links no longer point to the the part of the file they originally referenced. This commit updates them.


yan12125:
I guess it might be better to move these docs to [docs/](https://github.com/rg3/youtube-dl/tree/master/docs)

PR #6564, opened 2015-08-14
[msnbc] new extractor

fstirlitz:


yan12125:
I'm working on MSNBC, too. I'll commit my work ASAP.

yan12125:
I've adapted some of your codes in 26e1c3514f4af1ed60cd1114a653fe49e1fa8d11. Would you like to be listed in [AUTHORS](https://github.com/rg3/youtube-dl/blob/master/AUTHORS)? If so, under what name?

PR #6621, opened 2015-08-20
[vlive] New extractor for vlive.tv

ping:
"V" http://www.vlive.tv is a live broadcasting mobile platform targeted at South Korean celebrities

Similar to Periscope, videos are available for viewing after the live broadcast, but only on the app for now.

Example: http://m.vlive.tv/video/1326

### DIFF
--- a/youtube_dl/extractor/vlive.py
+++ b/youtube_dl/extractor/vlive.py
@@ -25,7 +25,6 @@ class VLiveIE(InfoExtractor):
             'ext': 'mp4',
             'title': '[V] Girl\'s Day\'s Broadcast',
             'creator': 'Girl\'s Day',
-            'upload_date': '20150817',
         },
     }
     _SECRET = 'rFkwZet6pqk1vQt6SxxUkAHX7YL3lmqzUMrU4IDusTo4jEBdtOhNfT4BYYAdArwH'
@@ -41,21 +40,14 @@ class VLiveIE(InfoExtractor):
         thumbnail = self._og_search_thumbnail(webpage)
         creator = self._html_search_regex(
             r'<span class="name">([^<>]+)</span>', webpage, 'creator')
-        upload_date = self._html_search_regex(
-            r'<span class="time">(\d{4}\.\d{2}\.\d{2})</span>', webpage,
-            'upload date', default=None, fatal=False)
-        if upload_date:
-            upload_date = upload_date.replace('.', '')
-
+        
         url = 'http://global.apis.naver.com/globalV/globalV/vod/%s/playinfo?' % video_id
-        msgpad = {'msgpad': '%.0f' % (time() * 1000)}
-        md = {
-            'md': b64encode(
-                hmac.new(self._SECRET.encode('ascii'),
-                         (url[:255] + msgpad['msgpad']).encode('ascii'), sha1).digest())
-        }
-        url += '&' + compat_urllib_parse.urlencode(msgpad) + '&' + compat_urllib_parse.urlencode(md)
-
+        msgpad = '%.0f' % (time() * 1000)
+        md = b64encode(
+            hmac.new(self._SECRET.encode('ascii'),
+                     (url[:255] + msgpad).encode('ascii'), sha1).digest()
+        )
+        url += '&' + compat_urllib_parse.urlencode({'msgpad': msgpad, 'md': md})
         playinfo = self._download_json(url, video_id, 'Downloading video json')
 
         if playinfo.get('message', '') != 'success':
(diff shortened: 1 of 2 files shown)


yan12125:
Why did you remove ```upload_date```?

youtube_dl/extractor/vlive.py
```diff
@@ -0,0 +1,85 @@
...
+        msgpad = '%.0f' % (time() * 1000)
+        md = b64encode(
+            hmac.new(self._SECRET.encode('ascii'),
+                     (url[:255] + msgpad).encode('ascii'), sha1).digest()
+        )
+        url += '&' + compat_urllib_parse.urlencode({'msgpad': msgpad, 'md': md})
+        playinfo = self._download_json(url, video_id, 'Downloading video json')
+
+        if playinfo.get('message', '') != 'success':
+            raise ExtractorError(playinfo['message'])
```
yan12125: Here you should consider the case that ```playinfo``` does not have the ```'message'``` field, too.

youtube_dl/extractor/vlive.py
```diff
@@ -0,0 +1,85 @@
...
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(
+            'http://m.vlive.tv/video/%s' % video_id,
+            video_id, note='Download video page')
+
+        title = self._og_search_title(webpage)
+        thumbnail = self._og_search_thumbnail(webpage)
+        creator = self._html_search_regex(
+            r'<span class="name">([^<>]+)</span>', webpage, 'creator')
```
yan12125: ```<span[^>]+class="name">...``` is better.

youtube_dl/extractor/vlive.py
```diff
@@ -0,0 +1,85 @@
...
+        'url': 'http://m.vlive.tv/video/1326',
+        'md5': 'cc7314812855ce56de70a06a27314983',
+        'info_dict': {
+            'id': '1326',
+            'ext': 'mp4',
+            'title': '[V] Girl\'s Day\'s Broadcast',
+            'creator': 'Girl\'s Day',
+        },
+    }
+    _SECRET = 'rFkwZet6pqk1vQt6SxxUkAHX7YL3lmqzUMrU4IDusTo4jEBdtOhNfT4BYYAdArwH'
```
yan12125: Is it possible to get the secret from the website rather using an hardcoded one? The key may be changed in the future and the extractor is broken before the new key is committed.

ping:
@yan12125 ``upload_date`` was removed because they switched to displaying relative dates, like "10 minutes ago", "yesterday", "12 days ago", "1 month ago" instead of the yyyy.mm.dd format.

youtube_dl/extractor/vlive.py
ping: Changed.

youtube_dl/extractor/vlive.py
ping: Changed.

youtube_dl/extractor/vlive.py
ping: Not to my knowledge. At present, videos are only playable through the official ios/android apps and not the website.

youtube_dl/extractor/vlive.py
```diff
@@ -0,0 +1,85 @@
...
+from ..utils import (
+    ExtractorError,
+    determine_ext
+)
+from ..compat import compat_urllib_parse
+
+
+class VLiveIE(InfoExtractor):
+    IE_NAME = 'vlive'
+    _VALID_URL = r'https?://(?:(www|m)\.)?vlive\.tv/video/(?P<id>[0-9]+)'
```
yan12125: I've tried http://www.vlive.tv/video/1326 and got a 404 error. Do you have a www.vlive.tv example URL?

ping: They used to redirect http://www.vlive.tv/video/1326 to http://m.live.tv/video/1326 for all requests, but it seems that they only do the redirect for mobile devices now (you can try with your phone, or toggle device mode in Chrome's Developer Tools). Not sure if this change in behaviour is a temporary bug.

When you share a link from the V app to Twitter, the link generated is still http://vlive.tv/video/1326 which redirects to http://www.vlive.tv/video/1326.

I'm inclined to leave www in the url regex for this.

yan12125: Thanks for the information. Could you add some comments?

ping:
@yan12125 Forgot to leave a comment that I've added a note about the mobile redirect in code. Not sure  if there's more you want documented. Let me know.

yan12125:
Thanks for the contribution!

PR #6649, opened 2015-08-23
Update subtitles for NetEaseMusic and QQMusic

fqj1994:
1. Copy description to subtitles.zh-CN.data
2. Remove the 3rd digit after dot in NetEase's timestamp to fit
LRC format (to be compatible with mpv)

### DIFF
--- a/test/helper.py
+++ b/test/helper.py
@@ -89,7 +89,7 @@ def gettestcases(include_onlymatching=False):
 md5 = lambda s: hashlib.md5(s.encode('utf-8')).hexdigest()
 
 
-def expect_info_dict(self, got_dict, expected_dict):
+def expect_dict(self, got_dict, expected_dict):
     for info_field, expected in expected_dict.items():
         if isinstance(expected, compat_str) and expected.startswith('re:'):
             got = got_dict.get(info_field)
@@ -127,6 +127,22 @@ def expect_info_dict(self, got_dict, expected_dict):
             got = got_dict.get(info_field)
             self.assertTrue(isinstance(got, expected),
                             'Expected type %r for field %s, but got value %r of type %r' % (expected, info_field, got, type(got)))
+        elif isinstance(expected, dict) and isinstance(got_dict.get(info_field, None), dict):
+            expect_dict(self, got_dict.get(info_field), expected)
+        elif isinstance(expected, list) and isinstance(got_dict.get(info_field, None), list):
+            got = got_dict.get(info_field, None)
+            self.assertEqual(len(expected), len(got),
+                             'Expect a list of length %d, but got a list of length %d' % (
+                             len(expected), len(got)))
+            _id = 0
+            for i, j in zip(got, expected):
+                _type_i = type(i)
+                _type_j = type(j)
+                self.assertEqual(_type_j, _type_i,
+                                 'Type doesn\'t match at element %d of the list in field %s, expect %s, got %s' % (
+                                 _id, info_field, _type_j, _type_i))
+                expect_dict(self, {'_': i}, {'_': j})
+                _id += 1
         else:
             if isinstance(expected, compat_str) and expected.startswith('md5:'):
                 got = 'md5:' + md5(got_dict.get(info_field))
@@ -149,6 +165,9 @@ def expect_info_dict(self, got_dict, expected_dict):
             self.assertEqual(expected, got,
                              'invalid value for field %s, expected %r, got %r' % (info_field, expected, got))
 
+
+def expect_info_dict(self, got_dict, expected_dict):
+    expect_dict(self, got_dict, expected_dict)
     # Check for the presence of mandatory fields
     if got_dict.get('_type') not in ('playlist', 'multi_video'):
         for key in ('id', 'url', 'title', 'ext'):
--- a/youtube_dl/extractor/neteasemusic.py
+++ b/youtube_dl/extractor/neteasemusic.py
@@ -66,6 +66,7 @@ class NetEaseMusicIE(NetEaseMusicBaseIE):
     IE_DESC = '网易云音乐'
     _VALID_URL = r'https?://music\.163\.com/(#/)?song\?id=(?P<id>[0-9]+)'
     _TESTS = [{
+        'note': 'origin + translated lyrics, with time tag need to be fixed',
         'url': 'http://music.163.com/#/song?id=32102397',
         'md5': 'f2e97280e6345c74ba9d5677dd5dcb45',
         'info_dict': {
@@ -75,7 +76,10 @@ class NetEaseMusicIE(NetEaseMusicBaseIE):
             'creator': 'Taylor Swift / Kendrick Lamar',
             'upload_date': '20150517',
             'timestamp': 1431878400,
-            'description': 'md5:a10a54589c2860300d02e1de821eb2ef',
+            'subtitles': {
+                'origin': [{'ext': 'lrc', 'data': 'md5:eb9ae90502b435de7d9e99fc7602adb4'}],
+                'translated': [{'ext': 'lrc', 'data': 'md5:cadca69fdfb7b679d273cc01a518f7dd'}],
+            }
         },
     }, {
         'note': 'No lyrics translation.',
@@ -87,7 +91,9 @@ class NetEaseMusicIE(NetEaseMusicBaseIE):
             'creator': '周杰伦',
             'upload_date': '20141225',
             'timestamp': 1419523200,
-            'description': 'md5:a4d8d89f44656af206b7b2555c0bce6c',
+            'subtitles': {
+                'origin': [{'ext': 'lrc', 'data': 'md5:a1766edaa6dbc85357f0ae9feabc867b'}],
+            }
         },
     }, {
         'note': 'No lyrics.',
@@ -99,6 +105,7 @@ class NetEaseMusicIE(NetEaseMusicBaseIE):
             'creator': 'Dustin O\'Halloran',
             'upload_date': '20080211',
             'timestamp': 1202745600,
+            'subtitles': {}
         },
     }, {
         'note': 'Has translated name.',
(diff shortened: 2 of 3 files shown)


youtube_dl/extractor/qqmusic.py
```diff
@@ -119,6 +119,12 @@ def _real_extract(self, url):
             'upload_date': publish_time,
             'creator': singer,
             'description': lrc_content,
+            'subtitles': {
+                'zh-CN': [{
```
dstftw: This is not always the case. Here is the `en` lyrics http://y.qq.com/#type=song&mid=001JyApY11tIp6.

youtube_dl/extractor/qqmusic.py
```diff
@@ -119,6 +119,12 @@ def _real_extract(self, url):
...
+                    'ext': 'lrc',
```
dstftw: It's not an [`.lrc` format](https://en.wikipedia.org/wiki/LRC_%28file_format%29) but just a plain text.

youtube_dl/extractor/qqmusic.py
fqj1994: @dstftw Yes, there're lyrics in other languages. I don't know what to write actually. zh-CN might be the best option if it has to be a valid language code. Please advise.

dstftw: The code should match the content. If it's not possible to figure out the lang code there is no point placing it in subtitles. Moreover it's not time bound so can't even barely be treated as subtitles.

fqj1994: Actually mpv supports .lrc as subtitles (except some issue for the last item).

dstftw: It's not an .lrc since it does not follow the format and not bound to time tags.

fqj1994: new code is added to filter out non-lrc plain text lyrics.
And, doing language detect would require a language model, which requires a lot of code. But introducing new dependencies doesn't sounds good. any advice?

youtube_dl/extractor/qqmusic.py
```diff
@@ -112,15 +117,35 @@ def _real_extract(self, url):
...
+        else:
+            return {
+                    'id': mid,
+                    'formats': formats,
+                    'title': song_name,
+                    'upload_date': publish_time,
+                    'creator': singer,
+                    'description': lrc_content,
+                    'thumbnail': thumbnail_url,
+                    }
```
dstftw: Just update the dictionary with `subtitles` key when `.lrc` detected. There is no need in code duplication. Also replace `zh-CN` with something more generic like `origin`.

fqj1994: updated and rebased.

fqj1994:
there're still some issue on neteasemusic's _process_lyrics

youtube_dl/extractor/qqmusic.py
```diff
@@ -58,6 +58,11 @@ def m_r_get_ruin():
         curMs = int(time.time() * 1000) % 1000
         return int(round(random.random() * 2147483647) * curMs % 1E10)
 
+    def _filter_lrc(self, data):
+        lyrics_expr = r'(\[[0-9]{2}:[0-9]{2}\.[0-9]{2,}\])([^\n]+)'
+        texts = re.findall(lyrics_expr, data)
+        return ''.join("%s%s\n" % (time_stamp, text) for time_stamp, text in texts)
```
dstftw: .lrc may contain another tags apart from time tags (ID tags) that are completely filtered out now.

fqj1994: fixed and rebased

fqj1994:
updated @dstftw

fqj1994:
The last line of .lrc will be ignored(and not shown) by mpv as it doesn't have a duration. and mpv cannot handle .lrc with other tags.
So adding a conversion to .srt for neteasemusic.

fqj1994:
ping @dstftw 
Is there any update for this PR or anything should I modify?
(conversion to .srt is discarded as mpv fixed a bug handling .lrc in their recent commits)

youtube_dl/extractor/neteasemusic.py
```diff
@@ -108,27 +106,44 @@ class NetEaseMusicIE(NetEaseMusicBaseIE):
...
+        lyrics_expr_multitimetag = r'(\[[0-9]{2}:[0-9]{2}\.[0-9]{2,}\])+([^\n]*)'
         original_ts_texts = re.findall(lyrics_expr, original)
         translation_ts_dict = dict(
-            (time_stamp, text) for time_stamp, text in re.findall(lyrics_expr, translated)
+            (time_stamp, text) for time_stamp, text in re.findall(lyrics_expr_multitimetag, translated)
         )
-        lyrics = '\n'.join([
-            '%s%s / %s' % (time_stamp, text, translation_ts_dict.get(time_stamp, ''))
+        lyrics = ''.join([
+            '%s%s / %s\n' % (self._fix_timestamp(time_stamp), text, translation_ts_dict.get(time_stamp)) if translation_ts_dict.get(time_stamp, None)
```
dstftw: What's the point combining original and translated lyrics to a single file? There should be 2 separated entries in `subtitles` instead.

dstftw:
Push changes as new commits since otherwise it's hard to track changes.

youtube_dl/extractor/neteasemusic.py
fqj1994: The original code to parse lyrics combined the original and translated lyrics into single .lrc data in description.
This is just to keep the original code.
Should I return 3 lyrics in pseudo lang-code: origin, translation, both?

dstftw: Originally, it was a description. I don't see much point keeping duplicate data. So `original` and `zh-CN` should be enough.

fqj1994: The latest commit just pushed is origin + translation + both.
If it is origin + zh-CN, the zh-CN should be the combined version, am I right? As seeing only translated lyrics while listening music sounds no point.

fqj1994: updated the latest to return only two lyrics.
@dstftw

dstftw: Combining is a task of a player or any other consumer, youtube-dl should not do that. Otherwise we would need to generate all `n! / (n - 2)!` partial permutations of unique subtitles for consistency.

dstftw: Combining could be moved to a postprocessor and exposed as a new generic option.

fqj1994: But in this case,
actually we don't know origin is `zh-CN` or not. (Chinese songs only have original lyrics, but not translated lyrics). For Chinese songs, this may make `zh-CN` not-existed as we cannot detect `origin` is `zh-CN` or not unless we blindly duplicate `origin` to `zh-CN` if no translation existed. But simply duplicating that also has a issue because for some songs lyrics are not translated, making `zh-CN` a non-Chinese lyrics.
Above all, maybe keeping 3 versions is the most reasonable way.
Any advice?

dstftw: Then mark them as `origin` and `translated` (like in source JSON). When some of them are missing they will miss from `subtitles` as well.

fqj1994: done in latest commit.

fqj1994: wait, an issue in latest commit.
_fix_timestamp is missed.

fqj1994: done. @dstftw

youtube_dl/extractor/neteasemusic.py
```diff
@@ -108,30 +106,55 @@ class NetEaseMusicIE(NetEaseMusicBaseIE):
...
+            return "[" + match_result.group(1) + "]"
+        else:
+            # return a valid timestamp to avoid other applications' error
+            return "[00:00.00]"
+
+    def _fix_lyric_timestamp(self, line):
+        text = line
+        corrected_times = ""
+        lyrics_expr = r'(\[[0-9]{2}:[0-9]{2}\.[0-9]{2,}\])([^\n]*)'
+        # Handle time tag with one or more time in one line.
```
dstftw: Can you name any example URL of such song?

fqj1994: Existed test cases(like the 1st case of NetEaseMusicIE) already includes .lrc with wrong time tag need to be fixed.
but md5: is not supported in lyrics.xx.data.
putting a complete corrected lyrics in codes doesn't seem good.
any advice?

fqj1994: I did some modification to test/helper.py to allow use md5: for string within dict or list.
I don't know it's proper or not.
It shouldn't break other tests but I don't have a good network to run all tests to check.

dstftw: I expected an example URL for
>more time in one line

fqj1994: URL for multiple time in one line is `http://music.163.com/#/song?id=4926366`
added as a test in latest commit

fqj1994: and as for multiple time tag in one line (whose time need to be fixed) is `http://music.163.com/#/song?id=22826396`
both tests are added.

fqj1994:
ping @dstftw

fqj1994:
Hello @dstftw  or others, is there any update or further changes needed for this PR?

test/helper.py
```diff
@@ -127,6 +131,15 @@ def expect_info_dict(self, got_dict, expected_dict):
             got = got_dict.get(info_field)
             self.assertTrue(isinstance(got, expected),
                             'Expected type %r for field %s, but got value %r of type %r' % (expected, info_field, got, type(got)))
+        elif isinstance(expected, dict) and isinstance(got_dict.get(info_field, None), dict):
+            expect_dict(self, got_dict.get(info_field), expected, True)
```
dstftw: This should be relaxed since in most cases there is no need in exact match of dicts or dicts' keys. Moreover some internal, compatibility or autocalculated fields may be placed in original dict. So, `same_keys` feature should be removed completely.

test/helper.py
```diff
@@ -127,6 +131,15 @@ def expect_info_dict(self, got_dict, expected_dict):
...
+        elif isinstance(expected, list) and isinstance(got_dict.get(info_field, None), list):
+            got = got_dict.get(info_field, None)
+            self.assertEqual(len(expected), len(got),
+                             "Expect a list of length %d, but got a list of length %d" % (
+                             len(expected), len(got)))
+            for i, j in zip(got, expected):
+                expect_dict(self, {'_': i}, {'_': j})
```
dstftw: This dict wrapping trick is only requred when both `i` and `j` are not dicts. This approach will fail with misleading assertion (`invalid value for field _`) when corresponding list items are of different type.

test/helper.py
```diff
@@ -127,6 +131,15 @@ def expect_info_dict(self, got_dict, expected_dict):
             got = got_dict.get(info_field)
             self.assertTrue(isinstance(got, expected),
                             'Expected type %r for field %s, but got value %r of type %r' % (expected, info_field, got, type(got)))
+        elif isinstance(expected, dict) and isinstance(got_dict.get(info_field, None), dict):
+            expect_dict(self, got_dict.get(info_field), expected, True)
+        elif isinstance(expected, list) and isinstance(got_dict.get(info_field, None), list):
+            got = got_dict.get(info_field, None)
+            self.assertEqual(len(expected), len(got),
+                             "Expect a list of length %d, but got a list of length %d" % (
```
dstftw: Prefer consistent using of single quotes when possible.

dstftw:
Could you also split this PR into 3 separated:
* Recursion support for test data
* NetEaseMusic changes
* QQMusic changes

QQMusic part could be merged long time ago but I can neither merge nor cherry-pick since it's overlaped with another code.

fqj1994:
hello @dstftw 
test helper has been modified as your requirement.
And PR has been spited into 3 commits.

dstftw:
Actually I've expected 3 separated **pull requests** rather than 3 commits.

fqj1994:
Please wait for up to 10 minutes.

fqj1994:
Submitted 3 PRs at #6984, #6985, #6986 
#6986 depends on #6985 but doesn't include the commit existed at #6985 . So tests of #6986 will fail unless #6985 is merged.

Issue #6577, opened 2015-08-16
wat.tv some videos can't be downloaded


dabiboo:
Hello,

I'm having troubles downloading some wat.tv videos.
Ex : 
youtube-dl -v http://www.wat.tv/video/anna-bergendahl-for-you-2015-7dvjn_76lkz_.html

--> 
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://www.wat.tv/video/anna-bergendahl-for-you-2015-7dvjn_76lkz_.html']
[debug] Encodings: locale cp1252, fs mbcs, out None, pref cp1252
[debug] youtube-dl version 2015.08.16
[debug] Python version 2.7.8 - Windows-7-6.1.7601-SP1
[debug] exe versions: ffmpeg N-43060-ga85b4a5, ffprobe N-43060-ga85b4a5
[debug] Proxy map: {}
[wat.tv] anna-bergendahl-for-you-2015: Downloading webpage
[wat.tv] 12404723: Downloading JSON metadata
ERROR: wat.tv returned error: <a href="http://www.wat.tv/video/anna-bergendahl-for-you-2015-7dvjn_76lkz_.html">Ce contenu n'est pas disponible directement dans ce player<br /><br />Pour le consulter cliquer ici</a>
Traceback (most recent call last):
  File "youtube_dl\YoutubeDL.pyo", line 655, in extract_info
  File "youtube_dl\extractor\common.pyo", line 287, in extract
  File "youtube_dl\extractor\wat.pyo", line 67, in _real_extract
ExtractorError: wat.tv returned error: <a href="http://www.wat.tv/video/anna-bergendahl-for-you-2015-7dvjn_76lkz_.html">Ce contenu n'est pas disponible directement dans ce player<br /><br />Pour le consulter cliquer ici</a>

Are theses videos protected or something ?

dstftw:
Can you watch it in browser?

dabiboo:
Yes, in my browser I can watch it.

I've just figured out the video is not reachable from the android app.

dstftw:
I can't watch it in browser as well as download. However I can both view in browser and download, for example, this video http://www.wat.tv/video/20-heures-aout-2015-7jgjt_2i0u7_.html.

dabiboo:
Ok so I guess you're not in France and this content is geoblocked. It cames from "Universal Music France"  so It makes sense they block it.

dstftw:
[...]

PR #6657, opened 2015-08-24
[universalmusicfrance] Add new extractor

dabiboo:
hi,

cf #6577 : adding support for universalmusicfrance.
Next step is to make the wat extractor redirect to this one when the video is hosted on universalmusicfrance like http://www.wat.tv/video/anna-bergendahl-for-you-2015-7dvjn_76lkz_.html
how can we do this ?


dabiboo:
Hi,

tests are failing because this extractor require a french ip. What do I do ? Do I add a skip on theses tests ?

dabiboo:
Hi,

It's been 15 days... Is there a problem ? could I get a feedback ?

yan12125:
Please don't include irrelevant commits in this pull request. There's no need to merge master in pull requests. Open a new one with desired changes only.

PR #6669, opened 2015-08-25
[bdsmstreak] Add new extractor

mackosajt:
This is my first extractor, please be gentle :)


dstftw:

mackosajt:
@dstftw Unfortunately I'm not sure how to decide which content violates any copyrights and which does not.

PR #6673, opened 2015-08-25
[MTV] add support for German site mtv.de

bastik:
This patch adds support for the German MTV site.


dstftw:
Instead of patching the bunch of classes why not inherit from `MTVIE` and override anything you need?

bastik:
Yes, this is better (done).

PR #6681, opened 2015-08-26
[vivo] Add new extractor

mall0c:


dstftw:
That's already implemented in [shared.sx extractor](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/shared.py) `_VALID_URL` of which I've just expanded to cover vivo.sx. Thanks anyway.

PR #6697, opened 2015-08-28
md5 on-the-fly for dash and http

xvagainstall:
calculating md5 and put in file with same filename as video + ".md5"

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1524,7 +1524,7 @@ class YoutubeDL(object):
                         fd.add_progress_hook(ph)
                     if self.params.get('verbose'):
                         self.to_stdout('[debug] Invoking downloader on %r' % info.get('url'))
-                    return fd.download(name, info)
+                    return fd.download(name, info, self.params.get("md5"))
 
                 if info_dict.get('requested_formats') is not None:
                     downloaded = []
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -162,9 +162,9 @@ def parseOpts(overrideArguments=None):
         '--ignore-config',
         action='store_true',
         help='Do not read configuration files. '
-        'When given in the global configuration file /etc/youtube-dl.conf: '
-        'Do not read the user configuration in ~/.config/youtube-dl/config '
-        '(%APPDATA%/youtube-dl/config.txt on Windows)')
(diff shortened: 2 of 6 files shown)


dstftw:
It should be calculated just once after all postprocessing is done [here](https://github.com/rg3/youtube-dl/blob/054d43bb11590da234e7e8f46a595a34b19efb8c/youtube_dl/YoutubeDL.py#L1639). Calculating it for some particular downloaders chunk by chunk makes no sense.

youtube_dl/YoutubeDL.py
```diff
@@ -1524,7 +1524,7 @@ def dl(name, info):
                         fd.add_progress_hook(ph)
                     if self.params.get('verbose'):
                         self.to_stdout('[debug] Invoking downloader on %r' % info.get('url'))
-                    return fd.download(name, info)
+                    return fd.download(name, info, self.params.get("md5"))
```
dstftw: It `md5` flag should not go here as an arguments. Instead it should be extracted in particular downloader from `self.params`.

youtube_dl/options.py
```diff
@@ -175,6 +175,11 @@ def _hide_login_info(opts):
         action='store_true', dest='no_color',
         default=False,
         help='Do not emit color codes in output')
+    general.add_option(
+        '--md5',
```
dstftw: This should be renamed to `--write-md5`.

youtube_dl/YoutubeDL.py
xvagainstall: calculate md5 for chunks we got lowering memory usage. its valuable for our project

xvagainstall: but ok, i'll change dat

dstftw: Calculate it in chunks after the final file is on disk. As said it makes no sense to calculate it immediately since the final file may be modified by postprocessor.

xvagainstall: We know about post process changes. But my teammates who directly use ydl have also harddrive issusa caus use it in 20-30 threads. It is ok if i write pre and postprocess md5 calculation?

dstftw: I guess no, duplicated options would be misleading for end-user.

jaimeMF:
Why don't you just use `--exec 'md5 {} > {}.md5'`?

PR #6703, opened 2015-08-29
Handle Screenwave Media embeds in the generic extractor

### LINKED ISSUE #6695: Request: TheCinemaSnob.com extractor for ScreenwaveMedia embeds
Given that there's an extractor for Cinemassacre (precedent) and that TheCinemaSnob.com appears to have switched to Screenwave Media embeds in response to YouTube ["holding the Elm Street 2: Freddy's Revenge episode hostage"](https://www.youtube.com/watch?v=Ui84pB-w154&t=1395), it'd probably be a good idea to add a TheCinemaSnob.com extractor to remove the need for users to poke around in the page source to find the specific ScreenwaveMedia URL that youtube-dl knows how to handle.

fstirlitz:
This should resolve #6695 (at least partially). Only the first commit is necessary to resolve the issue; the other two are tangential tweaks.

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -1718,6 +1718,15 @@ class GenericIE(InfoExtractor):
         if snagfilms_url:
             return self.url_result(snagfilms_url)
 
+        # Look for ScreenwaveMedia embeds
+        mobj = re.search(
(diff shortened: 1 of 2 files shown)


dstftw:
Thanks, cherry-picked,

PR #6704, opened 2015-08-29
[kaltura] add html5 player urls

sourcerect:
Kaltura has an HTML5 player with different, newer scheme of URLs. Luckily, the same API calls still work, but the numbers gotten from the URL are in different order, so the regex looks worse. I did my best...


dstftw:
Thanks, cherry-picked in ef49b5905345255bc6913e279b9d76a0bb83e1ce.

PR #6717, opened 2015-08-31
Support video-password on Youku.

psjay:


youtube_dl/extractor/youku.py
```diff
@@ -185,10 +196,17 @@ def retrieve_data(req_url, note):
...
 
+        video_password = self._downloader.params.get('videopassword', None)
+
         # request basic data
-        data1 = retrieve_data(
-            'http://v.youku.com/player/getPlayList/VideoIDS/%s' % video_id,
-            'Downloading JSON metadata 1')
+        if video_password:
+            data1 = retrieve_data(
+                'http://v.youku.com/player/getPlayList/VideoIDS/%s?password=%s' % (video_id, video_password),
```
dstftw: Just append `?password=%s` to the URL when `videopassword` is present instead of duplicating calls to `retrieve_data`.

psjay: good suggestion.

dstftw:
Thanks, cherry-picked.

PR #6725, opened 2015-09-01
Typo

Pitel:


yan12125:
It's not a typo. The right bracket is there:
```
[deprecated; use  -o "%(autonumber)s-%(title)s.%(ext)s" ] Number downloaded files starting from 00000
                                                        ^
```

PR #6734, opened 2015-09-01
[limelight] Add new extractor

remitamine:


dstftw:
Cherry-picked in ef5acfe32de4c995625f9800cfe0776237961436, thanks.

PR #6748, opened 2015-09-02
[dcn] add live streams and season extraction and support for other types of urls

remitamine:


youtube_dl/extractor/dcn.py
```diff
@@ -82,3 +104,122 @@ def _real_extract(self, url):
...
+            compat_urllib_parse.urlencode(data),
+            {
+                'Origin': 'http://www.dcndigital.ae',
+                'Content-Type': 'application/x-www-form-urlencoded'
+            })
+
+        show = self._download_json(request, show_id)
+        season_id = season_id or show['default_season']
+        season = {}
+        for _ in show['seasons']:
```
dstftw: `_` is idiomatic way to denote unused variables.

remitamine:
i changed the variable name to season.

PR #6752, opened 2015-09-03
Generic extractor enhancements

fstirlitz:
Two rather unrelated patches to the generic extractor. The `<audio>` tag support is a no-brainer, but I'm less sure about the other patch; it may introduce some false positives. On the other hand, if the generic extractor reaches this point, there's probably nothing better to do.

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -1797,7 +1797,7 @@ class GenericIE(InfoExtractor):
                 found = filter_video(re.findall(r'<meta.*?property="og:video".*?content="(.*?)"', webpage))
         if not found:
             # HTML5 video
-            found = re.findall(r'(?s)<video[^<]*(?:>.*?<source[^>]*)?\s+src=["\'](.*?)["\']', webpage)
+            found = re.findall(r'(?s)<(?:video|audio)[^<]*(?:>.*?<source[^>]*)?\s+src=["\'](.*?)["\']', webpage)
         if not found:
             REDIRECT_REGEX = r'[0-9]{,2};\s*(?:URL|url)=\'?([^\'"]+)'
             found = re.search(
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/generic.py
```diff
@@ -1817,6 +1817,11 @@ def filter_video(urls):
                     'url': new_url,
                 }
         if not found:
+            # Last resort: just look for anything that looks like a media URL
+            found = set(re.findall(
+                r'\b((?:https?:)?//[A-Za-z0-9.:[\]]+/[A-Za-z0-9\-\._~:/()*+,;=%]+\.(?:mp3|mp4|flv|m3u8|smil)(?:\?[A-Za-z0-9\-\._~:/[\]()*+,;=%]*)?)\b',
```
dstftw: This is probably unwanted since it may match just any plain text (e.g. user comment) if URL happens to be there.

fstirlitz: Perhaps. My other idea was to scan the contents of `<script>` tags, `data-*` attributes and flashvars (passed in whichever manner) for URLs of this form. That still isn't ideal; were it applied to ComCarrCoff, it would match lots of spurious URLs (given that the embedded JSON contains the data for all the videos, not just the current one). On the other hand, if the generic extractor reaches this point, there's little else you can do anyway.

dstftw: I've yet cherry-picked your first commit in 12439dd5ec46ccac1ef13db7280473ecef44a096, thanks. As of second lets wait to hear other opinions.

jaimeMF: As you have pointed, it could match unwanted urls so I would avoid adding it. If you are only interested in ComCarrCoff, you could try to add a custom extractor because the method may not be used in too many sites..

dstftw: As a compromise we could add an `--aggressive-extraction` option that will try aggressive strategies like this one.

jaimeMF: I guess it's ok, I wouldn't use it but I don't mind.

fstirlitz: To be honest, I don't care that much about this change. I'll be fine if this PR is closed without merging.

PR #6755, opened 2015-09-03
[downloader/external] Respect --no-check-certificate for curl and aria2c and --proxy for curl

remitamine:
if i didn't set `--no-check-certificate` the value of `param` is `False` then it pass this condition:
```python
if isinstance(param, bool):
```
that makes it apply regardless of setting it or not.
so i added this condition to fix it:
```python
or param is False
```

### DIFF
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -53,6 +53,14 @@ class ExternalFD(FileDownloader):
             return [command_option]
         return [command_option, param]
 
+    def _bool_option(self, command_option, param, true_value='true', false_value='false', separator=None):
+        param = self.params.get(param)
+        if not isinstance(param, bool):
+            return []
+        if separator:
+            return [command_option + separator + (true_value if param else false_value)]
+        return [command_option, true_value if param else false_value]
+
     def _configuration_args(self, default=[]):
         ex_args = self.params.get('external_downloader_args')
         if ex_args is None:
@@ -123,7 +131,7 @@ class Aria2cFD(ExternalFD):
             cmd += ['--header', '%s: %s' % (key, val)]
         cmd += self._option('--interface', 'source_address')
         cmd += self._option('--all-proxy', 'proxy')
-        cmd += self._option('--check-certificate=false', 'nocheckcertificate')
+        cmd += self._bool_option('--check-certificate', 'nocheckcertificate', 'false', 'true', '=')
         cmd += ['--', info_dict['url']]
         return cmd
 
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -47,7 +47,7 @@ class ExternalFD(FileDownloader):
 
     def _option(self, command_option, param):
         param = self.params.get(param)
-        if param is None:
+        if param is None or param is False:
             return []
         if isinstance(param, bool):
             return [command_option]
@@ -80,6 +80,8 @@ class CurlFD(ExternalFD):
         for key, val in info_dict['http_headers'].items():
             cmd += ['--header', '%s: %s' % (key, val)]
         cmd += self._option('--interface', 'source_address')
+        cmd += self._option('--proxy', 'proxy')
+        cmd += self._option('--insecure', 'nocheckcertificate')
         cmd += self._configuration_args()
         cmd += ['--', info_dict['url']]
         return cmd
@@ -121,6 +123,7 @@ class Aria2cFD(ExternalFD):
             cmd += ['--header', '%s: %s' % (key, val)]
         cmd += self._option('--interface', 'source_address')
         cmd += self._option('--all-proxy', 'proxy')
+        cmd += self._option('--check-certificate=false', 'nocheckcertificate')
         cmd += ['--', info_dict['url']]
         return cmd
 
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -47,7 +47,7 @@ class ExternalFD(FileDownloader):
 
     def _option(self, command_option, param):
         param = self.params.get(param)
-        if param is None or param is False:
+        if param is None:
             return []
         return [command_option, param]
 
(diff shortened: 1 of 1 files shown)


dstftw:
This breaks the original semantics for [bool options without value](https://github.com/remitamine/youtube-dl/blob/external_downloader_options/youtube_dl/downloader/external.py#L52-L53). These have to be always passed as bare option switch without value whether they `true` or not. If you really need to pass value to bool option then it's probably better to add a new method something similar to this:
```python
def _bool_option(self, command_option, param, true_value='true', false_value='false'):
    if not isinstance(param, bool):
        return []
    return [command_option, true_value if param else false_value]
```

remitamine:
but it won't work with aria2c because this command is working:
```
aria2c --check-certificate=false url
```
and this don't work:
```
aria2c --check-certificate false url
Exception caught
Exception: [download_helper.cc:417] errorCode=1 Unrecognized URI or unsupported protocol: false
```
because aria2c will execute the command like it was
```
aria2c --check-certificate=true false url
```

dstftw:
Add a `separator=None` parameter.

remitamine:
i added the function and used it.

youtube_dl/downloader/external.py
```diff
@@ -47,12 +47,20 @@ def supports(cls, info_dict):
 
     def _option(self, command_option, param):
         param = self.params.get(param)
-        if param is None:
+        if param is None or param is False:
```
dstftw: Second check should be removed.

remitamine: if i remove it the same problem will happen for curl and wget:
always the --insecure will be passed to curl and and --no-check-certificate to wget.
for example(i didn't pass the --no-check-certificate to youtube-dl but the insecure is passed to curl because the value of nocheckcertificate is False witch make the if `isinstance(param, bool`): pass):
```
youtube-dl -v --ignore-config --external-downloader curl https://www.youtube.com/watch?v=H8E3hLAXHRs
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['-v', '--ignore-config', '--external-downloader', 'curl', 'https://www.youtube.com/watch?v=H8E3hLAXHRs']
[debug] youtube-dl version 2015.08.28
[debug] Python version 3.4.3 - Linux-4.1.6-1-ARCH-x86_64-with-arch-Arch-Linux
[debug] exe versions: ffmpeg 2.7.2, ffprobe 2.7.2, rtmpdump 2.4
[debug] Proxy map: {}
[youtube] H8E3hLAXHRs: Downloading webpage
[youtube] H8E3hLAXHRs: Downloading video info webpage
[youtube] H8E3hLAXHRs: Extracting video information
[youtube] H8E3hLAXHRs: Downloading DASH manifest
[youtube] H8E3hLAXHRs: Downloading DASH manifest
[debug] Invoking downloader on 'https://r6---sn-nx5cvox-2hvl.googlevideo.com/videoplayback?id=1fc13784b0171d1b&itag=133&source=youtube&requiressl=yes&pl=18&mm=31&mn=sn-nx5cvox-2hvl&ms=au&mv=m&ratebypass=yes&mime=video/mp4&gir=yes&clen=1479105&lmt=1441024129364364&dur=48.444&signature=5E5B9FDAD629A621E614855BBE23A61DAE2AB7A5.3559D2A13BCD9209576CC781830180120CB0BA53&mt=1441394976&sver=3&key=dg_yt0&fexp=9407992,9408710,9409069,9413122,9415365,9415485,9416023,9416126,9416729,9417280,9417707,9418153,9418448,9418465,9419445,9420022,9420152,9420348,9420434&upn=INxCwen4Ys4&ip=197.205.100.112&ipbits=0&expire=1441416621&sparams=ip,ipbits,expire,id,itag,source,requiressl,pl,mm,mn,ms,mv,ratebypass,mime,gir,clen,lmt,dur'
[download] Destination: خلط بين نشرة الأخبار ونشرة الأحوال الجوية على المباشر؟!-H8E3hLAXHRs.f133.mp4
False
[debug] curl command line: curl --location -o 'خلط بين نشرة الأخبار ونشرة الأحوال الجوية على المباشر؟!-H8E3hLAXHRs.f133.mp4.part' --header 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:10.0) Gecko/20150101 Firefox/20.0 (Chrome)' --header 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8' --header 'Accept-Charset: ISO-8859-1,utf-8;q=0.7,*;q=0.7' --header 'Accept-Language: en-us,en;q=0.5' --header 'Accept-Encoding: gzip, deflate' --insecure -- 'https://r6---sn-nx5cvox-2hvl.googlevideo.com/videoplayback?id=1fc13784b0171d1b&itag=133&source=youtube&requiressl=yes&pl=18&mm=31&mn=sn-nx5cvox-2hvl&ms=au&mv=m&ratebypass=yes&mime=video/mp4&gir=yes&clen=1479105&lmt=1441024129364364&dur=48.444&signature=5E5B9FDAD629A621E614855BBE23A61DAE2AB7A5.3559D2A13BCD9209576CC781830180120CB0BA53&mt=1441394976&sver=3&key=dg_yt0&fexp=9407992,9408710,9409069,9413122,9415365,9415485,9416023,9416126,9416729,9417280,9417707,9418153,9418448,9418465,9419445,9420022,9420152,9420348,9420434&upn=INxCwen4Ys4&ip=197.205.100.112&ipbits=0&expire=1441416621&sparams=ip,ipbits,expire,id,itag,source,requiressl,pl,mm,mn,ms,mv,ratebypass,mime,gir,clen,lmt,dur'
```

dstftw: Yes, argumentless options handling should be moved in separate method as well:
```python
def _argless_option(self, command_option, param, expected_value=True):
    ...
    return [command_option] if param == expected_value else []
```

remitamine: when i add this is these part of code will still useful:
```python
if isinstance(param, bool):
    return [command_option]
```

dstftw: It may be useful if default option's value is `None` (that I overlooked for `nocheckcertificate`), but I guess it's better to remove it and require usage of argless option method in such cases.

dstftw: Also `_valueless_option` is probably a better name for it.

remitamine:
i added the `_argless_option` function and used it for `nocheckcertificate`

remitamine:
i changed the function name to _valueless_option.

dstftw:
Thanks.

PR #6765, opened 2015-09-05
minor refactoring to code

suhussai:
Refactored some function definitions in youtube_dl/compat.py

### DIFF
--- a/youtube_dl/compat.py
+++ b/youtube_dl/compat.py
@@ -228,10 +228,7 @@ except ImportError:  # Python < 3.3
 
 
 def compat_ord(c):
-    if type(c) is int:
-        return c
-    else:
-        return ord(c)
+    return c if type(c) is int else ord(c)
 
 
 if sys.version_info >= (3, 0):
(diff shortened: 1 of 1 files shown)


youtube_dl/compat.py
```diff
@@ -207,15 +207,10 @@ def _parse_qsl(qs, keep_blank_values=False, strict_parsing=False,
...
         pairs = _parse_qsl(qs, keep_blank_values, strict_parsing,
                            encoding=encoding, errors=errors)
-        for name, value in pairs:
-            if name in parsed_result:
-                parsed_result[name].append(value)
-            else:
-                parsed_result[name] = [value]
-        return parsed_result
+        return {name:(parsed_result[name] + [value] if name in parsed_result 
+                      else [value]) for name, value in pairs}
```
yan12125: This syntax is not supported on Python 2.6.

jaimeMF:
The new `compat_parse_qs` doesn't work because `parsed_result`  is not defined, additionally I don't think it can be implemented with your approach because this query string is valid: `'d=2&d=4&d=5'`.

About the change in `compat_ord`: I think it was fine before, the expression is not too complex but for me `if *: else:` is usually more clear.

suhussai:
Ok that makes sense.

PR #6781, opened 2015-09-07
[test/test_write_annotations] fix expected test filename

remitamine:
because `ext` is replaced here:https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L1464


jaimeMF:
Thanks.

PR #6783, opened 2015-09-07
[yahoo] change test md5 checksum

remitamine:
and about the test with this url:https://ca.sports.yahoo.com/video/program-makes-hockey-more-affordable-013127711.html
it's not available anymore.


yan12125:
Thanks!

PR #6790, opened 2015-09-07
[canal13cl] fix info extraction

remitamine:


youtube_dl/extractor/tele13.py
```diff
@@ -0,0 +1,77 @@
...
+class Tele13IE(InfoExtractor):
+    _VALID_URL = r'^http://(?:www\.)?t13\.cl/videos(?:/[^/]+)+/(?P<id>[\w-]+)'
+    _TESTS = [
+        {
+            'url': 'http://www.t13.cl/videos/actualidad/el-circulo-de-hierro-de-michelle-bachelet-en-su-regreso-a-la-moneda',
+            'md5': '4cb1fa38adcad8fea88487a078831755',
+            'info_dict': {
+                'id': 'el-circulo-de-hierro-de-michelle-bachelet-en-su-regreso-a-la-moneda',
+                'ext': 'mp4',
+                'title': 'El c\u00edrculo de hierro de Michelle Bachelet en su regreso a La Moneda',
```
jaimeMF: I don't see any reason to use unicode escape sequences here, you can directly use `'í'`.

remitamine: i copied it from the --write-info-json output.

jaimeMF: python's `json` use escape sequences, but both are supported. I prefer to directly use the character (imaging having to use escape sequences for titles in chines, arabic ...).

remitamine: ok i will change it.

remitamine:
i changed the unicode escape sequences and skip the test.

PR #6806, opened 2015-09-09
[vier] fix VierVideos extraction

remitamine:


dstftw:
Extraction could be fixed just by fixing several lines of code of the old approach. Apart from better mimicry of the website behavior and ~20kb per page gain what are the benefits? Now extraction relies completely on JS layout of the page and any change will result in broken extractor.

remitamine:
i folowed what the site is downing whenever i click on `Meer` button i see that it makes the request i've added.
the site already changed and it's not possible to avoid these changes.
i think now it will be broken only if they choose to use anothor CMS.

dstftw:
I'm aware that you are trying to mimick the website. I don't see much benefit in this approach so sticking with the old one.

remitamine:
it was possible to add the old code as a fallback.

PR #6810, opened 2015-09-09
[fktv] fix info extraction

remitamine:


yan12125:
An extractor should raise an error if something goes wrong (regex not found, etc.) Returns nothing from ```_real_extract``` causes youtube-dl silently fails.

yan12125:
Thanks. Merged.

PR #6812, opened 2015-09-09
Extract all vid.me formats

lalinsky:
The current vid.me extractor tries to download "application/dash+xml" documents as media for many videos. For example, for https://vid.me/Gc6M it downloads the contents of https://api.vid.me/video/4176611/stream?format=dash.

I switched the extractor to use vid.me's JSON API and now it extracts all formats.

### DIFF
--- a/youtube_dl/extractor/vidme.py
+++ b/youtube_dl/extractor/vidme.py
@@ -1,10 +1,10 @@
 from __future__ import unicode_literals
 
-from .common import InfoExtractor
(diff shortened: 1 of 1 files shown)


dstftw:
Cherry-picked in d9c19db340c568db1afcd6d85ab5b9c4681dd7df, thanks.

lalinsky:
Thank you for merging it so quickly!

PR #6851, opened 2015-09-13
Fix bad formatting of man page (Closes #6510)

### LINKED ISSUE #6510: [minor bug] bad formatting of man page
Some sections in the man page are badly formatted, notably in the OPTIONS section and the FAQ section. The problem is most likely caused by hard-coded spaces/tabs and line breaks.

Consider python(1):
python_1
and youtube-dl(1):
youtube-dl_1

Also the Debian bug report:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=793054

When I run `make youtube-dl.1`, it shows:

    python devscripts/prepare_manpage.py >youtube-dl.1.temp.md
    pandoc -s -f markdown -t man youtube-dl.1.temp.md -o youtube-dl.1
    rm -f youtube-dl.1.temp.md

Should I try augmenting `devscripts/prepare_manpage.py`? Thanks!

-- System Information:
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v']
[debug] youtube-dl version 2015.06.04.1
[debug] Python version 2.7.10 - Linux-4.1.0-1-amd64-x86_64-with-debian-stretch-sid
[debug] exe versions: avconv 11.4-6, avprobe 11.4-6, rtmpdump 2.4
[debug] Proxy map: {}

alexvong1995:
Previously, the `Makefile` uses `pandoc` to convert `README.md` to `youtube-dl.1` but the formatting is not satisfactory. Therefore, I change `devscripts/prepare_manpage.py` to convert` README.md` to **Perl Pod** format, which is then piped to `pod2man` to produce the `youtube-dl.1` man page.

Note that a system with **Python** installed usually implies it has **Perl** installed. Since `pod2man` comes with a standard installation of **Perl**, we may assume the existence of `pod2man` on most developer's system.

A picture showing the result is attached below:
youtube-dl_1_fixed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@ BINDIR ?= $(PREFIX)/bin
 MANDIR ?= $(PREFIX)/man
 SHAREDIR ?= $(PREFIX)/share
 PYTHON ?= /usr/bin/env python
+VERSION ?= $(shell echo 'print(__version__)' | cat youtube_dl/version.py - | python)
+DATE ?= $(shell echo '$(VERSION)' | sed -e 's/\./-/g')
 
 # set SYSCONFDIR to /etc if PREFIX=/usr or PREFIX=/usr/local
 ifeq ($(PREFIX),/usr)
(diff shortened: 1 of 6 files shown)


yan12125:
I have a (possibly) simpler implementation for #6510. I'll create a PR as long as I've tested my codes on all major platforms.

alexvong1995:
Thanks, my implementation is a bit slow. Build time increased from 1.5s to 2s on my computer.  Your simpler implementation (if possible) should be preferred. You can email me if you encounter any problems. The original script is written in Scheme: <https://gitlab.com/alexvong1995/youtube-dl_man-page/blob/master/prepare_manpage.scm>.

yan12125:
@alexvong1995 I've posted my codes at #6852. Feel free to leave comments for it.

alexvong1995:
Add more photos to illustrate how different Markdown's markups are handled.

**OPTIONS** section:
youtube-dl_1_fixed_more_1

List of items:
youtube-dl_1_fixed_more_2

Verbatim paragraph:
youtube-dl_1_fixed_more_3

**FAQ** section:
youtube-dl_1_fixed_more_4

**pod2man** does a better job in spacing than **pandoc** IMHO.

jaimeMF:
I think that now the only difference with your images is in some of the underlined text. Pandoc only adds underlines when using `*FOO*` but for all of the options, output template fields and others we use
 ```
`FOO`
```
which I think is the standard method in markdown.

I personally think its fine as it is, so I'm closing the issue.

PR #6853, opened 2015-09-13
Add pip instructions to upgrade error message

hickford:


dstftw:
I don't see much point putting upgrade instructions here just for a particular pip scenario to update youtube-dl. Here it would be more reasonable to just provide a link to [update FAQ](https://yt-dl.org/update) and cover all the cases in FAQ entry.

PR #6864, opened 2015-09-15
[fivethirtyeight] Add new extractor

bikeseat:
For http://fivethirtyeight.com

### DIFF
--- a/youtube_dl/extractor/fivethirtyeight.py
+++ b/youtube_dl/extractor/fivethirtyeight.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+from .common import InfoExtractor
+from youtube_dl.compat import compat_urllib_request
+from youtube_dl import utils
+
+
+class FivethirtyeightIE(InfoExtractor):
+    _VALID_URL = r'http://fivethirtyeight\.com/.+'
+    _TEST = {
+        'url': 'http://fivethirtyeight.com/features/rage-against-the-machines/',
+        'md5': 'c825a057981316c4d4444fefea35a108',
+        'info_dict': {
+            'id': '11694550',
+            'ext': 'mp4',
+            'title': 'Rage Against The Machines',
+            'description': 'This is an excerpted chapter from “The Signal and the Noise: Why So Many Predictions Fail — but Some Don’t” by Nate Silver, editor in chief of FiveThirtyEight. …',
+            'duration': 1037,
+        }
+    }
+
+    def _real_extract(self, url):
+        webpage = self._download_webpage(url, 'video_id')
+        video_id = self._html_search_regex(r'.*data-video-id=\'(.*)\' data-cms.*', webpage, 'video_id')
+        title = self._html_search_regex(r'<title>(.*)\s*\|', webpage, 'title')
+
+        data = self._download_json(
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/fivethirtyeight.py
```diff
@@ -0,0 +1,52 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+from .common import InfoExtractor
+from youtube_dl.compat import compat_urllib_request
+from youtube_dl import utils
+
+
+class FivethirtyeightIE(InfoExtractor):
+    _VALID_URL = r'http://fivethirtyeight\.com/.+'
```
dstftw: You should capture a part of URL that represents a video in unique way...

youtube_dl/extractor/fivethirtyeight.py
```diff
@@ -0,0 +1,52 @@
...
+            'id': '11694550',
+            'ext': 'mp4',
+            'title': 'Rage Against The Machines',
+            'description': 'This is an excerpted chapter from “The Signal and the Noise: Why So Many Predictions Fail — but Some Don’t” by Nate Silver, editor in chief of FiveThirtyEight. …',
+            'duration': 1037,
+        }
+    }
+
+    def _real_extract(self, url):
+        webpage = self._download_webpage(url, 'video_id')
```
dstftw: ...and output it here instead of const string `video_id` here that makes no sense.

youtube_dl/extractor/fivethirtyeight.py
```diff
@@ -0,0 +1,52 @@
...
+        video_id = self._html_search_regex(r'.*data-video-id=\'(.*)\' data-cms.*', webpage, 'video_id')
```
dstftw: `.*` on both ends of regex make no sense. Also when capturing at least one character is required - there is no sense capturing empty video id.

youtube_dl/extractor/fivethirtyeight.py
```diff
@@ -0,0 +1,52 @@
...
+        title = self._html_search_regex(r'<title>(.*)\s*\|', webpage, 'title')
```
dstftw: Same here.

youtube_dl/extractor/fivethirtyeight.py
```diff
@@ -0,0 +1,52 @@
...
+
+        data = self._download_json(
+            'http://espn.go.com/videohub/video/util/getMinifiedClipJsonById?id=%s&cms=espn&device=mobile&omniReportSuite=wdgespvideo,wdgespfivethirtyeight,wdgespge&xhr=1' % video_id, video_id)
+
+        url = data["videos"][0]["links"]["mobile"]["href"]
+
+        request = compat_urllib_request.Request(url)
+        request.add_header('User-Agent', 'ipad')
+
+        formats = self._extract_m3u8_formats(request, 'display_id', 'mp4')
```
dstftw: Do not use static strings instead of actual video id.

bikeseat:
Thanks for the feedback. Let me know how the changes look.

B

yan12125:

yan12125:
Ping @remitamine 

This can be closed, right? Apparently there's a bug in GitHub's commit message parser.

remitamine:
i guess it thinks that `(closes` is part of the url(i should've remove the `http://` part).

PR #6867, opened 2015-09-15
Use `insert` for all sys.path manipulations

yan12125:
For solving the problem mentioned in https://github.com/rg3/youtube-dl/pull/6852#issuecomment-140127015.
With this patch, ```./youtube_dl/__main__.py``` no longer imports the wrong package.


jaimeMF:
Thanks, cherry picked in 95240b8093c6fe9007e33f0991ed8d54b42eaf3b.

PR #6872, opened 2015-09-15
Added A Script that Adds Easier Downloading

atorres1929:
Process of how script works:
-Looks through Download.txt
-Looks element up using Google Ajax Tool
-Returns link to youtube-dl to be downloaded
-Downloads all elements to links folder


dstftw:
That's out of scope of youtube-dl. Moreover OS-specific.

atorres1929:
Okay understood, thank you for the comment!

PR #6880, opened 2015-09-17
[togglesg] New extractor for toggle.sg

ping:
Toggle.sg is an online service from a broadcast station in Singapore.

The extractor works for clips and programmes available for streaming on the site, example http://video.toggle.sg/en/series/lion-moms-tif/trailers/lion-moms-premier/343115

Some content are geo-restricted or drm-protected.

### DIFF
--- a/youtube_dl/extractor/togglesg.py
+++ b/youtube_dl/extractor/togglesg.py
@@ -1,0 +1,159 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import json
+import re
+import itertools
+
+from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    int_or_none,
+    determine_ext,
+    parse_iso8601,
+    remove_end
+)
+from ..compat import compat_urllib_request
+
+
+class ToggleSgIE(InfoExtractor):
+    IE_NAME = 'togglesg'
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/togglesg.py
```diff
@@ -0,0 +1,159 @@
...
+    _API_USER = 'tvpapi_147'
+    _API_PASS = '11111'
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, video_id, note='Downloading video page')
+
+        api_user = self._search_regex(
+            r'apiUser:\s*"([^"]+)"', webpage, 'apiUser', default=self._API_USER, fatal=False)
```
dstftw: There is no need in `fatal` when `default` is provided.

youtube_dl/extractor/togglesg.py
```diff
@@ -0,0 +1,159 @@
...
+    determine_ext,
+    parse_iso8601,
+    remove_end
+)
+from ..compat import compat_urllib_request
+
+
+class ToggleSgIE(InfoExtractor):
+    IE_NAME = 'togglesg'
+    _VALID_URL = r'https?://video\.toggle\.sg/(?:(en|zh))/(?:(series|clips|movies))/.+?/(?P<id>[0-9]+)'
```
dstftw: Don't capture groups if you are not going to use them.

ping:
Changes made.

dstftw:
Cherry-picked, thanks. Sorry for long wait.

PR #6883, opened 2015-09-17
[postprocessor/ffmpeg] Always use the 'file:' protocol for filenames (fixes #6874)

### LINKED ISSUE #6874: Using ':' in -o breaks ffmpeg
I was trying to download a full playlist from youtube, and I wanted the videos numbered so I could watch in order, so I used:
youtube-dl --playlist-reverse -f bestvideo+171/bestvideo+140 -o "%(autonumber)s: %(title)s-%(id)s.%(ext)s" https://www.youtube.com/playlist?list=PL26B36130624FCAF5

The unfortunate side effect is that ffmpeg believes that the filename is specifying a protocol.  I could just use a different character, but I was able to fix it by modifying FFmpegPostProcessor.run_ffmpeg_multiple_files and simply prepending 'file:' to each input_path and the out_path.

jaimeMF:
**WARNING**: I haven't tested it in windows.

@yan12125 I think that we don't call `run_ffmpeg_multiple_files` with an url in any place, but `_ffmpeg_filename_argument` can be adapted to support urls.

### DIFF
--- a/youtube_dl/downloader/hls.py
+++ b/youtube_dl/downloader/hls.py
@@ -31,7 +31,7 @@ class HlsFD(FileDownloader):
         args = [
             encodeArgument(opt)
             for opt in (ffpp.executable, '-y', '-i', url, '-f', 'mp4', '-c', 'copy', '-bsf:a', 'aac_adtstoasc')]
-        args.append(encodeFilename(tmpfilename, True))
+        args.append(encodeFilename(ffpp._ffmpeg_filename_argument(tmpfilename), True))
 
         self._debug_cmd(args)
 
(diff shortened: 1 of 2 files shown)


dstftw:
Works fine on Windows. Note that #6874 does not affect Windows since [path is sanitized from forbidden characters](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/utils.py#L358).

yan12125:
> I think that we don't call ```run_ffmpeg_multiple_files``` with an url in any place

You're right. Just merge it.

jaimeMF:
Cherry picked in 8a7bbd160641dbbc13f68aaa26dfa49b71483a8e

PR #6897, opened 2015-09-19
[traileraddict] fix info extraction

remitamine:


youtube_dl/extractor/traileraddict.py
```diff
@@ -1,64 +1,70 @@
...
-            fvar = "fvarhd"
-        else:
-            fvar = "fvar"
+        video_info = self._get_video_info(video_id)
 
-        info_url = "http://www.traileraddict.com/%s.php?tid=%s" % (fvar, str(video_id))
-        info_webpage = self._download_webpage(info_url, video_id, "Downloading the info webpage")
+        formats = [{
+            'url': video_info['fileurl'][0].strip(),
+            'width': int_or_none(video_info.get('vidwidth')[0]),
```
dstftw: This will fail if `vidwidth` is missing.

youtube_dl/extractor/traileraddict.py
```diff
@@ -1,64 +1,70 @@
...
-                                       info_webpage, 'Download url').replace('%3F', '?')
-        thumbnail_url = self._search_regex(r'&image=(.+?)&',
-                                           info_webpage, 'thumbnail url')
+        if video_info.get('hdurl')[0].startswith('http://'):
+            formats.append({
+                'url': video_info['hdurl'][0].strip(),
+                'width': int_or_none(video_info.get('hd_vidwidth')[0]),
+                'height': int_or_none(video_info.get('hd_vidheight')[0]),
+                'format_id': 'hd',
+            })
```
dstftw: This code looks similar to `sd` format and can be extracted to a function.

remitamine: how about this:
```python
def _get_formats(self, video_info, format_spec):
        return {
            'url': video_info[format_spec['url']][0].strip(),
            'width': int_or_none(video_info.get(format_spec['width'])[0]) if format_spec['width'] in video_info else None,
            'height': int_or_none(video_info.get(format_spec['height'])[0]) if format_spec['height'] in video_info else None,
            'format_id': format_spec['format_id'],
        }
```

PR #6916, opened 2015-09-21
[ninegag] fix _VALID_URL regex and handle the use of other external providers

remitamine:


dstftw:
Why not just extract `post_view['sourceUrl']`?

remitamine:
this is right i will change it to get the url from sourceUrl and keep _EXTERNAL_VIDEO_PROVIDER to get the ie-key.

dstftw:
No, just pass it without `ie_key` in order to delegate resolve to youtube-dl.

remitamine:
if ie_key is known in advance. youtube-dl can access the right extractor directly.

dstftw:
If this external video provider mapping changes on 9gag the whole extractions will be broken. In case of passing `sourceUrl` without `ie_key` it will work as usual.

remitamine:
thanks for explanation,
i will remove _EXTERNAL_VIDEO_PROVIDER and use only `sourceUrl`.

dstftw:
You can preserve `_EXTERNAL_VIDEO_PROVIDER` scenario as a fallback when `sourceUrl` is missing.

remitamine:
i apply the requested changes.

youtube_dl/extractor/ninegag.py
```diff
@@ -42,19 +37,44 @@ class NineGagIE(InfoExtractor):
...
 
         webpage = self._download_webpage(url, display_id)
 
         post_view = json.loads(self._html_search_regex(
             r'var postView = new app\.PostView\({\s*post:\s*({.+?}),\s*posts:\s*prefetchedCurrentPost', webpage, 'post view'))
 
-        youtube_id = post_view['videoExternalId']
+        ie_key = None
+        source_url = post_view.get('sourceUrl')
+        if not source_url or source_url == '':
```
dstftw: `not source_url` implies `source_url == ''`.

remitamine: i didn't know this. i will change it.

remitamine:
i removed the unnecessary condition

dstftw:
Thanks, cherry-picked.

PR #6921, opened 2015-09-22
[ninegag] VALID_URL fix

stannynuytkens:
Domain changed from 9gag.tv to 9gag.com(/tv/...)
Someone still needs to update the testcases for this extractor!


yan12125:
Thanks for fixing it, while it's a duplicate of #6916.

stannynuytkens:
Yes, I noticed it too late..

PR #6934, opened 2015-09-23
[condenast] fix extraction and add support for other sites

remitamine:


youtube_dl/extractor/condenast.py
```diff
@@ -86,8 +97,8 @@ def _extract_video(self, webpage, url_type):
         info_url = base_info_url + data
         info_page = self._download_webpage(info_url, video_id,
                                            'Downloading video info')
-        video_info = self._search_regex(r'var video = ({.+?});', info_page, 'video info')
-        video_info = json.loads(video_info)
+        video_info = self._search_regex(r'var\s*video\s*=\s*({.+?});', info_page, 'video info')
```
dstftw: Should be `r'var\s+video...`.

remitamine:
> Should be `r'var\s+video...`.

i changed it.

dstftw:
Thanks, cherry-picked.

PR #6956, opened 2015-09-24
[keek] fix info extraction

remitamine:


dstftw:
Uploader still can be extracted.

remitamine:
i added code to extract description, uploader and uploader id.

dstftw:
```
  File "youtube_dl/__main__.py", line 16, in <module>
    import youtube_dl
  File "C:\Dev\git\youtube-dl\master\youtube_dl\__init__.py", line 43, in <module>
    from .extractor import gen_extractors, list_extractors
  File "C:\Dev\git\youtube-dl\master\youtube_dl\extractor\__init__.py", line 272, in <module>
    from .keek import KeekIE
  File "C:\Dev\git\youtube-dl\master\youtube_dl\extractor\keek.py", line 17
SyntaxError: Non-ASCII character '\xc3' in file C:\Dev\git\youtube-dl\master\youtube_dl\extractor\keek.py on line 17, but no encoding declared; see http://www.python.org/peps/pep-0263.html for details
```

remitamine:
it works now in both python 2 and python 3

youtube_dl/extractor/keek.py
```diff
@@ -1,46 +1,42 @@
...
-        'md5': '09c5c109067536c1cec8bac8c21fea05',
+        'url': 'https://www.keek.com/keek/NODfbab',
+        'md5': '9b0636f8c0f7614afa4ea5e4c6e57e83',
         'info_dict': {
             'id': 'NODfbab',
             'ext': 'mp4',
-            'uploader': 'youtube-dl project',
-            'uploader_id': 'ytdl',
-            'title': 'test chars: "\'/\\\u00e4<>This is a test video for youtube-dl.For more information, contact phihag@phihag.de .',
+            'title': 'test chars: "\'/\\\xe4<>This is a test video for youtube-dl.For more information, contact phihag@phihag.de . - Video - Videos on Keek',
```
dstftw: You should add a coding cookie on top of the file instead.

remitamine: > You should add a coding cookie on top of the file instead.

you mean?
```python
# coding: utf-8
```

dstftw: Yes.

remitamine:
i added the utf-8 coding cookie

youtube_dl/extractor/keek.py
```diff
@@ -1,46 +1,43 @@
...
-        if raw_desc:
-            uploader = self._html_search_regex(
-                r'Watch (.*?)\s+\(', raw_desc, 'uploader', fatal=False)
-            uploader_id = self._html_search_regex(
-                r'Watch .*?\(@(.+?)\)', raw_desc, 'uploader_id', fatal=False)
-        else:
-            uploader = None
-            uploader_id = None
+        uploader = uploader_id = None
+        matches = re.search(r'data-username="(?P<uploader>[^"]+)"[^>]*data-user-id="(?P<uploader_id>[^"]+)"', webpage)
```
dstftw: Why not just use `self._search.regex` twice?

remitamine: because they are extracted from the same tag.

dstftw: This does not prevent from using `self._search_regex` in any way.

remitamine: i will change to use self._search_regex.

remitamine:
i changed it to use _search_regex.

dstftw:
Cherry-picked, thanks.

PR #6966, opened 2015-09-25
[kuwo] fix title extraction and update test

remitamine:


youtube_dl/extractor/kuwo.py
```diff
@@ -78,7 +79,7 @@ def _real_extract(self, url):
             errnote='Unable to get song detail info')
 
         song_name = self._html_search_regex(
-            r'<h1[^>]+title="([^"]+)">', webpage, 'song name')
+            r'(?s)class="(?:[^" ]+ +)*title(?: +[^" ]+)*".*?<h1[^>]+title="([^"]+)"', webpage, 'song name')
```
yan12125: According to the [W3C HTML5 syntax spec](http://www.w3.org/TR/html5/syntax.html), using ```'\s'``` to match whitespaces is better here.

yan12125:
Could you add a check for the 'offline' error, too?

remitamine:
i added the chack for offline error page and change match for whitespaces with `\s`

youtube_dl/extractor/kuwo.py
```diff
@@ -76,9 +77,11 @@ def _real_extract(self, url):
         webpage = self._download_webpage(
             url, song_id, note='Download song detail info',
             errnote='Unable to get song detail info')
+        if '对不起，该歌曲由于版权问题已被下线，将返回网站首页' in webpage:
+            raise ExtractorError('this song has been offline because of copyright issues')
```
yan12125: Please add ```expected=True``` here as it's not a youtube-dl's fault.

remitamine:
i changed the offline error into an expected ExtractorError

yan12125:
Thanks. Merged.

PR #6978, opened 2015-09-26
[fktv] get format_id from video file ext

remitamine:
completion of this commit 4866b72eb2bf5747cde4654e75b1c1be0d456456


jaimeMF:
Thanks for catching this.

PR #6985, opened 2015-09-26
[test] recursively check dict and list in expect_info_dict

fqj1994:
This allows to use md5:, re:, etc within the str inside a list
or dict.

Separate to 3 PRs per discussion at issue #6649 .

### DIFF
--- a/test/helper.py
+++ b/test/helper.py
@@ -89,81 +89,78 @@ def gettestcases(include_onlymatching=False):
 md5 = lambda s: hashlib.md5(s.encode('utf-8')).hexdigest()
 
 
+def expect_value(self, got, expected, field):
+    if isinstance(expected, compat_str) and expected.startswith('re:'):
+        match_str = expected[len('re:'):]
+        match_rex = re.compile(match_str)
+
+        self.assertTrue(
+            isinstance(got, compat_str),
+            'Expected a %s object, but got %s for field %s' % (
+                compat_str.__name__, type(got).__name__, field))
+        self.assertTrue(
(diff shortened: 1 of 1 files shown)


jaimeMF:
In my opinion The part for checking lists is a bit messy, maybe it would be better to split the function in two:
* One for checking if two fields are equal
```
def expect_value(got, expected, field):
     # all the if isintance(...)
    ...
```
* And other for checking dictionaries:
```
def expect_dict(self, got_dict, expected_dict):
    for info_field, expected in expected_dict.items():
        got = got_dict.get(info_field)
        expect_value(got, expected, info_field)
```

test/helper.py
```diff
@@ -89,66 +89,82 @@ def gettestcases(include_onlymatching=False):
...
+        self.assertTrue(isinstance(got, expected),
+                        'Expected type %r for field %s, but got value %r of type %r' % (expected, field, got, type(got)))
+    elif isinstance(expected, dict) and isinstance(got, dict):
+        expect_dict(self, got, expected)
+    elif isinstance(expected, list) and isinstance(got, list):
+        self.assertEqual(len(expected), len(got),
+                         'Expect a list of length %d, but got a list of length %d' % (
+                         len(expected), len(got)))
+        _id = 0
+        for i, j in zip(got, expected):
```
dstftw: Use `for index, (item_got, item_expected) in enumerate(zip(got, expected)):` instead. Either use descriptive names for type variables or don't extract them as variables at all.

dstftw:
Cherry-picked, thanks.

PR #6992, opened 2015-09-27
YouTube: Get titles from playlist (fix #6699)

### LINKED ISSUE #6699: YouTube: -j --flat-playlist no longer gets titles
```
$ youtube-dl --verbose -j --flat-playlist https://www.youtube.com/user/rhettandlink2
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--verbose', u'-j', u'--flat-playlist', u'https://www.youtube.com/user/rhettandlink2']
[debug] Encodings: locale UTF-8, fs UTF-8, out None, pref UTF-8
[debug] youtube-dl version 2015.08.28
[debug] Python version 2.7.6 - Linux-3.16.0-46-generic-x86_64-with-Ubuntu-14.04-trusty
[debug] exe versions: avconv 11-6, avprobe 11-6, rtmpdump 2.4
[debug] Proxy map: {}
{"url": "GE6ipRyomnA", "_type": "url", "ie_key": "Youtube", "id": "GE6ipRyomnA"}
{"url": "fx3jLYHP8oE", "_type": "url", "ie_key": "Youtube", "id": "fx3jLYHP8oE"}
...
```
Earlier this year I reported a similar issue, #4971.  I guess YouTube has changed something on their end again.  Hopefully it's still possible to get the video titles without fetching each video's page.  It seems like it should be possible, because if I leave off `-j`, I can see it fetching the playlist pages, and those include the video titles (at least, the HTML pages do), but I guess `youtube-dl` just isn't parsing them out.

If I leave off `--flat-playlist`, I can get JSON with the `fulltitle` element, but the JSON also includes *everything*, while all I want is the ID and title.

Thanks for your help.

alphapapa:
Parse video titles from playlist pages' HTML/JSON.  This broke a
while back when YouTube changed something.  Now running
"-j --flat-playlist" will return both IDs and titles of videos again.

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -1586,11 +1586,19 @@ class YoutubePlaylistIE(YoutubeBaseInfoExtractor):
                 matches = self._VIDEO_RE.finditer(content_html)
 
                 # Get videos from current page. Using OrderedDict to
-                # avoid duplicates would make this much simpler.
+                # avoid duplicates would make this much
+                # simpler. Lacking that, we store the order of the
+                # videos as video_num so we can sort the dict, keeping
+                # the order of the playlist. We have to avoid
+                # duplicates because it seems that every video in the
+                # playlist shows up in the HTML/JSON twice: once
+                # without a title, and once with a title. Maybe using
+                # something like bs4 instead of regexps would also be a
+                # good idea.
                 new_videos = {}
+                num = 0
                 for m in matches:
-                    video_index = m.group('index')
-                    if video_index == '0':
+                    if m.group('index') == '0':
                         # Ignore link with index 0
                         continue
 
@@ -1602,21 +1610,18 @@ class YoutubePlaylistIE(YoutubeBaseInfoExtractor):
                         video_title = None
 
                     if video_id in new_videos:
-                        # Duplicate video
-
+                        # Video is already in dict
                         if video_title and not new_videos[video_id]['title']:
                             # Set missing title
                             new_videos[video_id]['title'] = video_title
-
-                        new_videos[video_id]['index'] = video_index
-
                     else:
-                        # New video
-                        new_videos[video_id] = {'index': int(video_index),
-                                                'title': video_title}
+                        # Video not in dict
+                        new_videos[video_id] = {'num': num, 'title': video_title}
 
-                # Sort videos by index
-                new_videos = sorted(new_videos.iteritems(), key=lambda v: v[1]['index'])
+                    num += 1
+
+                # Sort videos by playlist order
+                new_videos = sorted(new_videos.iteritems(), key=lambda v: v[1]['num'])
 
                 # Yield current list of videos
                 for video in new_videos:
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -1440,23 +1440,7 @@ class YoutubePlaylistIE(YoutubeBaseInfoExtractor):
                         ((?:PL|LL|EC|UU|FL|RD|UL)[0-9A-Za-z-_]{10,})
                      )"""
     _TEMPLATE_URL = 'https://www.youtube.com/playlist?list=%s'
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/youtube.py
```diff
@@ -1553,25 +1557,37 @@ def _extract_playlist(self, playlist_id):
...
-                new_ids = orderedSet(m.group('id') for m in matches if m.group('index') != '0')
-                for vid_id in new_ids:
-                    yield self.url_result(vid_id, 'Youtube', video_id=vid_id)
 
+                # Make list of videos
+                new_videos = [{'id': m.group('id'),
+                               'title': m.group('title').strip()}
+                              for m in matches
+                              if m.group('index') != '0'  # Ignore link with index 0
+                              if m.group('title').strip()]  # Ignore links without titles, which also prevents duplicates
```
dstftw: No, entries without title should not be ignored. On this stage we only require video id/URL.

youtube_dl/extractor/youtube.py
```diff
@@ -1440,7 +1440,7 @@ class YoutubePlaylistIE(YoutubeBaseInfoExtractor):
                         ((?:PL|LL|EC|UU|FL|RD|UL)[0-9A-Za-z-_]{10,})
                      )"""
     _TEMPLATE_URL = 'https://www.youtube.com/playlist?list=%s'
-    _VIDEO_RE = r'href="\s*/watch\?v=(?P<id>[0-9A-Za-z_-]{11})&amp;[^"]*?index=(?P<index>\d+)'
+    _VIDEO_RE = r'href="\s*/watch\?v=(?P<id>[0-9A-Za-z_-]{11})&amp;[^"]*?index=(?P<index>\d+)[^>]+>(?P<title>[^<]+)'
```
dstftw: Title part should be optional.

youtube_dl/extractor/youtube.py
```diff
@@ -1553,25 +1557,37 @@ def _extract_playlist(self, playlist_id):
                     'Invalid parameters. Maybe URL is incorrect.',
                     expected=True)
             elif re.match(r'[^<]*Choose your language[^<]*', match):
+                # Looks good; continue
```
dstftw: Avoid senseless comments.

youtube_dl/extractor/youtube.py
alphapapa: Are there videos without titles, though?  Does YouTube even permit that?

jaimeMF: Even if all videos have a title the regex may stop to detect the title someday and I prefer getting all the videos in the playlist instead of missing some or all of them.

Also, I haven't tested but titles could be empty or have only spaces (I don't know if YouTube checks that before allowing you to submit the video).

dstftw: It does not matter whether it's permitted or not. Relying on mandatory title where it's technically not required is more likely to result in broken extraction.

alphapapa: Good points.  I fixed it to handle videos without titles.  It would be nice to use an OrderedDict to make the code for avoiding duplicate videos cleaner (because, when accepting videos without titles, it appears that every video that has a title ends up matching twice, once without and once with title; something weird about the HTML/JSON YouTube returns).

OrderedDicts are in Python 2.7.  Will we be able to use them in the future?

youtube_dl/extractor/youtube.py
```diff
@@ -1440,7 +1440,23 @@ class YoutubePlaylistIE(YoutubeBaseInfoExtractor):
...
+
+        # Index
+        index=(?P<index>\d+)[^>]+
+
+        # End of <a> tag
+        >
+
+        # Video title (optional)
+        (?P<title>[^<]+)?
+        """, re.VERBOSE)
```
dstftw: That's unreadable. Revert to single line without any duplicating comment.

alphapapa:
Is there anything else I need to do to make this acceptable?  Thanks.

Issue #6819, opened 2015-09-10
Vidzi is broken


davek-cygwin:
C:\Projects\rg3\youtube-dl>python -m youtube_dl "http://vidzi.tv/jnk3om3hftz4.html" --verbose 2>&1 | tee fail-vidzi.log

[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'http://vidzi.tv/jnk3om3hftz4.html', u'--verbose']
[debug] Encodings: locale cp1252, fs mbcs, out None, pref cp1252
[debug] youtube-dl version 2015.09.09
[debug] Git HEAD: 1721fef
[debug] Python version 2.7.8 - Windows-7-6.1.7601-SP1
[debug] exe versions: ffmpeg 0.8.1, rtmpdump 2.4
[debug] Proxy map: {}
[Vidzi] jnk3om3hftz4: Downloading webpage
ERROR: Unable to extract video url; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "youtube_dl\YoutubeDL.py", line 660, in extract_info
    ie_result = ie.extract(url)
  File "youtube_dl\extractor\common.py", line 287, in extract
    return self._real_extract(url)
  File "youtube_dl\extractor\vidzi.py", line 24, in _real_extract
    r'{\s*file\s*:\s*"([^"]+)"\s*}', webpage, 'video url')
  File "youtube_dl\extractor\common.py", line 584, in _html_search_regex
    res = self._search_regex(pattern, string, name, default, fatal, flags, group)
  File "youtube_dl\extractor\common.py", line 575, in _search_regex
    raise RegexNotFoundError('Unable to extract %s' % _name)
RegexNotFoundError: Unable to extract video url; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.

davek-cygwin:
Body of webpage: https://gist.github.com/davek-cygwin/9ba2d360d963dbc0495f

Relevant part is:

<script type="text/javascript" src="http://static.vidzi.tv/nplayer/jwplayer.js"></script>
[...]

PR #6998, opened 2015-09-27
[vidzi] Fix extraction as described in #6819

kaltsimon:
Fixed as described in #6819, but without using gmpy2.

### DIFF
--- a/youtube_dl/extractor/vidzi.py
+++ b/youtube_dl/extractor/vidzi.py
@@ -2,6 +2,8 @@
 from __future__ import unicode_literals
 
 from .common import InfoExtractor
+import re
+import string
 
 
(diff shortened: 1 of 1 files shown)


yan12125:
The ```int2base``` function is almost the same as [
Alex Martelli's answer on StackOverflow](http://stackoverflow.com/a/2267446). To be used in youtube-dl, these codes should be released into public domain by the original author. Have you asked him?

yan12125:
Thanks for this work. I've pushed similar changes to master. See commits between 8f4a2124a914207912bf9fc37e593210e8dd423b and efbd6fb8bb86c07e6f924a7ec2c4bd486face3a4. I use my own ```base_n``` function to avoid licensing issues.

Issue #7036, opened 2015-10-01
Tumblr: No Format selection / Gets low-res video only


Hrxn:
This applies to all videos on Tumblr which have that nice 'HD' button in the lower right corner.

Tumblr now serves two different videos (two different files), or maybe even more, don't know about their API or backend.

2015-10-04:
~~I found an example: http://larastonesbitch.tumblr.com/post/130035771559/honestlyiconic~~
Great. Just a few days later, and that blog already moved..
Example is now here, it seems: http://larastoneswitch.tumblr.com/post/130035771559/honestlyiconic

    youtube-dl --get-url http://larastoneswitch.tumblr.com/post/130035771559

gives us this link:
https://www.tumblr.com/video_file/130279545660/tumblr_nvda1gkdqg1uqlov4/480
*(To get the file: Right-click, save link as)*

You will get a .mp4 file.

If you use this direct link:
https://www.tumblr.com/video_file/130279545660/tumblr_nvda1gkdqg1uqlov4

You get a different .mp4 file. Still the same video, but higher resolution (better quality)

*(You can also click the links directly, the video gets now played in your browser (usually), but you can clearly see the different videos/files/resolutions)*

Please note: The quality in general isn't that great in my example here, but this was just the first video I found that had the HD button.

**Edit:**

This video is SFW. Looks like a short clip from an older Victoria's Secret Fashion Show, I think.
Finding a NSFW video would have been faster, though ;-)

Hrxn:
For the record:

The old example link is not longer working anymore.

The post with the video has moved here: http://larastoneswitch.tumblr.com/post/130035771559/honestlyiconic

OTOH, thank you, good job!

dstftw:
HD videos will be supported in the next version.

PR #7040, opened 2015-10-01
[tumblr] Added support for HD video where available (#7036)

oliciv:
Added support + tests for HD videos on Tumblr where available (reported in #7036)

### DIFF
--- a/youtube_dl/extractor/tumblr.py
+++ b/youtube_dl/extractor/tumblr.py
@@ -7,6 +7,7 @@ from .common import InfoExtractor
 
 from ..utils import RegexNotFoundError
 
+
 class TumblrIE(InfoExtractor):
     _VALID_URL = r'http://(?P<blog_name>.*?)\.tumblr\.com/(?:post|video)/(?P<id>[0-9]+)(?:$|[/?#])'
     _TESTS = [{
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/tumblr.py
```diff
@@ -68,8 +99,35 @@ def _real_extract(self, url):
...
+            format_id = hd_video_url.split("/")[-1] + 'p'
+            if len(format_id) != 4:
+                format_id = "720p"
+            video_urls.append({
+                'ext': 'mp4',
+                'format_id': 'hd',
+                'url': hd_video_url,
+                'resolution': format_id,
+            })
+        except RegexNotFoundError:
```
yan12125: Passing ```default=None``` to ```_search_regex``` is cleaner.

youtube_dl/extractor/tumblr.py
```diff
@@ -68,8 +99,35 @@ def _real_extract(self, url):
...
+            'url': sd_video_url,
+            'resolution': format_id,
+        })
+
+        try:
+            hd_video_url = self._search_regex(r'hdUrl":"([^"]+)"',
+                                              iframe, 'hd video url')
+            hd_video_url = hd_video_url.replace("\\", "")
+            format_id = hd_video_url.split("/")[-1] + 'p'
+            if len(format_id) != 4:
```
yan12125: This may not be a good check. ```format_id``` like 'abcp' passes the check, too. On the other hand, we usually leave the resolution field ```None``` if it's unknown.

oliciv:
@yan12125 Thanks, appreciate the feedback, will make some changes

oliciv:
Have cleaned up HD URL regex error handling and set resolution to None if we can't be sure. From all the examples I found, looks like we can trust the <video url>.split("/")[-1] to be the resolution if it's 3 characters, but no longer make assumptions if it's not there.

youtube_dl/extractor/tumblr.py
```diff
@@ -29,6 +29,34 @@ class TumblrIE(InfoExtractor):
             'thumbnail': 're:http://.*\.jpg',
         }
     }, {
+        'url': 'http://hdvideotest.tumblr.com/post/130323439814/test-description-for-my-hd-video',
+        'md5': '99a84522f60972bf064a0b80f87bcbb5',
+        'resolution': 'hd',
```
jaimeMF: Unless I'm missing something, this field is not being checked, it should go inside `'info_dict`'  instead.

dstftw:
Cherry-picked in 2a27e66234ec0030d2ca6f1ddf229db7f7eb8ded, thanks.

PR #7041, opened 2015-10-01
[sfdebris] Added new extractor

### LINKED ISSUE #7039: Feature request: Screenwave support
SF Debris has switched from Blip to Screenwave, so I cannot use youtube-dl on it anymore. Fortunately the Javascript is reasonably easy to read; the mechanism for retrieving video URLs is as follows.

1. Retrieve the source of the page (try `http://sfdebris.com/videos/animation/transformerss1e01.php` as a test case).
2. Match the regex `http://player[0-9]+.screenwavemedia.com/player.php?id=[^']+`.
3. Retrieve the source of the URL matched in step 2.
4. Make a note of the value of `vidid` in the `attributes` map.
5. Match the regex `http://player[0-9]+.screenwavemedia.com/player.js`.
6. Retrieve the source of the URL matched in step 4.
7. The `sources` list inside `jwoptions.playlist` contains the video URLs at decreasing levels of fidelity. `thisObj.options.videoserver` is a reference to an object declared earlier in the file; `playerVidId` is the value of `vidid` which was recorded in step 4.

Some of this errs on the side of trying not to take advantage of what may be coincidences and trying to minimise the possible necessity of future maintenance. For instance,

* The value of `vidid` recorded in step 4 is the same as the `id` parameter of the script whose URL is matched in step 2.
* There appears to be only one `player` server, `player2.screenwavemedia.com`, at the moment.

ngld:
This fixes #7039.
The extractor is based on ```TeamFourIE``` in ```youtube_dl/extractor/screenwavemedia.py```.

### DIFF
--- a/youtube_dl/extractor/__init__.py
+++ b/youtube_dl/extractor/__init__.py
@@ -534,6 +534,7 @@ from .senateisvp import SenateISVPIE
 from .servingsys import ServingSysIE
 from .sexu import SexuIE
 from .sexykarma import SexyKarmaIE
+from .sfdebris import SfDebrisIE
 from .shahid import ShahidIE
 from .shared import SharedIE
 from .sharesix import ShareSixIE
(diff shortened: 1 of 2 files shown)


dstftw:
Screenwave is already supported by youtube-dl and http://sfdebris.com/videos/animation/transformerss1e01.php is processed by generic extractor just fine.

PR #7043, opened 2015-10-02
Voissa Add new extractor

tiburce:
Add Voissa support

### DIFF
--- a/youtube_dl/extractor/ruutu.py
+++ b/youtube_dl/extractor/ruutu.py
@@ -74,7 +74,7 @@ class RuutuIE(InfoExtractor):
                         preference = -1 if proto == 'rtmp' else 1
                         label = child.get('label')
                         tbr = int_or_none(child.get('bitrate'))
-                        width, height = [int_or_none(x) for x in child.get('resolution', 'x').split('x')]
+                        width, height = [int_or_none(x) for x in child.get('resolution', 'x').split('x')[:2]]
                         formats.append({
                             'format_id': '%s-%s' % (proto, label if label else tbr),
                             'url': video_url,


tiburce:
This site lag sometimes ... the script work well.

docs/supportedsites.md
```diff
@@ -636,6 +636,7 @@
  - **vlive**
  - **Vodlocker**
  - **VoiceRepublic**
+ - **Voissa**
```
dstftw: This is an autogenerated file.

tiburce:
Sorry ,  I know now ...

youtube_dl/extractor/voissa.py
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('id')
+        display_id = mobj.group('display_id')
+        webpage = self._download_webpage(url, video_id)
+
+        url_info = self._download_webpage('http://communaute.voissa.com/playlist.php?id=' + str(video_id), 'Url inf')
```
dstftw: Use `self._download_xml` instead and process as XML.

youtube_dl/extractor/voissa.py
```diff
@@ -0,0 +1,48 @@
...
+        video_url = self._search_regex(r'<media:content url="([^"]+)"', url_info, 'video URL')
+        thumbnail = self._search_regex(r'<media:thumbnail url="([^"]+)"', url_info, 'thumbnail URL', fatal=False)
+
+        title = self._search_regex(r'<title>Vidéo ([^.]+).+Voissa</title>', webpage, 'title')
+
+        mobj = re.search(r'<span itemprop="name">(?P<uploader>[^<]+)</span>.+vu (?P<viewcount>\d+)', webpage)
+        uploader = mobj.group('uploader')
+        view_count = mobj.group('viewcount')
```
dstftw: `view_count` must be int.

Issue #5362, opened 2015-04-05
NBA falls back to generic information extractor


RingoTheDog:
URIL:  http://www.nba.com/kings/sim-bhullar-media-availability

C:\>youtube-dl -v http://www.nba.com/kings/sim-bhullar-media-availability
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://www.nba.com/kings/sim-bhullar-media-
availability']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2015.04.03
[debug] Python version 2.7.8 - Windows-7-6.1.7601-SP1
[debug] exe versions: ffmpeg N-40824-g31dfe20
[debug] Proxy map: {}
[generic] sim-bhullar-media-availability: Requesting header
WARNING: Falling back on generic information extractor.
[generic] sim-bhullar-media-availability: Downloading webpage
[generic] sim-bhullar-media-availability: Extracting information
ERROR: Unsupported URL: http://www.nba.com/kings/sim-bhullar-media-availability
Traceback (most recent call last):
  File "youtube_dl\extractor\generic.pyo", line 846, in _real_extract
  File "youtube_dl\utils.pyo", line 1528, in parse_xml
  File "xml\etree\ElementTree.pyo", line 1300, in XML
  File "xml\etree\ElementTree.pyo", line 1642, in feed
  File "xml\etree\ElementTree.pyo", line 1506, in _raiseerror
ParseError: duplicate attribute: line 3, column 2
Traceback (most recent call last):
  File "youtube_dl\YoutubeDL.pyo", line 651, in extract_info
  File "youtube_dl\extractor\common.pyo", line 275, in extract
  File "youtube_dl\extractor\generic.pyo", line 1345, in _real_extract
UnsupportedError: Unsupported URL: http://www.nba.com/kings/sim-bhullar-media-av
ailability

Same video....different URL...similar result:
URL:  http://www.nba.com/kings/video/teams/kings/2015/04/03/SimShootIntsmmov-3563067

C:\>youtube-dl -v http://www.nba.com/kings/video/teams/kings/2015/04/03/SimShoot
Intsmmov-3563067
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://www.nba.com/kings/video/teams/kings/
2015/04/03/SimShootIntsmmov-3563067']
[...]

Issue #4167, opened 2014-11-12
NBA URL


RingoTheDog:
URL: http://www.nba.com/raptors/video/2014/11/11/NBAWEB141111BOUCHARDRAPTORmp4-3433066

youtube-dl -v http://www.nba.com/raptors/video/2014/11/11/NBAWEB141111BOUCHARDRAPTORmp4-3433066
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['-v', 'http://www.nba.com/raptors/video/2014/11/11/N
BAWEB141111BOUCHARDRAPTORmp4-3433066']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2014.11.12.1
[debug] Python version 2.7.8 - Windows-7-6.1.7601-SP1
[debug] exe versions: ffmpeg N-40824-
[debug] Proxy map: {}
[generic] NBAWEB141111BOUCHARDRAPTORmp4-3433066: Requesting header
WARNING: Falling back on generic information extractor.
[generic] NBAWEB141111BOUCHARDRAPTORmp4-3433066: Downloading webpage
[generic] NBAWEB141111BOUCHARDRAPTORmp4-3433066: Extracting information
ERROR: Unsupported URL: http://www.nba.com/raptors/video/2014/11/11/NBAWEB141111
BOUCHARDRAPTORmp4-3433066; please report this issue on https://yt-dl.org/bug . B
e sure to call youtube-dl with the --verbose flag and include its complete outpu
t. Make sure you are using the latest version; type  youtube-dl -U  to update.
Traceback (most recent call last):
  File "youtube_dl\extractor\generic.pyo", line 599, in _real_extract
  File "youtube_dl\utils.pyo", line 1310, in parse_xml
  File "xml\etree\ElementTree.pyo", line 1300, in XML
  File "xml\etree\ElementTree.pyo", line 1642, in feed
  File "xml\etree\ElementTree.pyo", line 1506, in _raiseerror
ParseError: duplicate attribute: line 3, column 2
Traceback (most recent call last):
  File "youtube_dl\YoutubeDL.pyo", line 536, in extract_info
  File "youtube_dl\extractor\common.pyo", line 199, in extract
  File "youtube_dl\extractor\generic.pyo", line 980, in _real_extract
ExtractorError: Unsupported URL: http://www.nba.com/raptors/video/2014/11/11/NBA
[...]

PR #7044, opened 2015-10-02
[nba] extract all video formats and extract more info

remitamine:


remitamine:
i find a better way to extract the info and formats.
it's better than hardcoded formats.

remitamine:
i improve the extraction by extracting the video info from xml feed

remitamine:
i added support for team subsites video like in this issues(#4167, #5362, #7068)
not all of them(i didn't add support for urls that does not contain `/video/` in the url like this http://www.nba.com/timberwolves/wiggins-shootaround and this http://www.nba.com/kings/sim-bhullar-media-availability
in this case the id should be extracted from the nbaVideoCore.firstVideo variable in the webpage).

PR #7055, opened 2015-10-03
[izlesene] extract info using api request

remitamine:


youtube_dl/extractor/izlesene.py
```diff
@@ -28,13 +24,8 @@ class IzleseneIE(InfoExtractor):
...
                 'ext': 'mp4',
                 'title': 'Sevinçten Çıldırtan Doğum Günü Hediyesi',
-                'description': 'md5:253753e2655dde93f59f74b572454f6d',
                 'thumbnail': 're:^http://.*\.jpg',
-                'uploader_id': 'pelikzzle',
-                'timestamp': int,
-                'upload_date': '20140702',
-                'duration': 95.395,
-                'age_limit': 0,
+                'duration': 95395,
```
jaimeMF: Duration **must** be in seconds, it's in milliseconds now.

jaimeMF:
Ideally we would extract all the information from a lightweight source (the json api in this case), but I don't think it's reasonable to stop getting the same information as before (even if we have to download the normal webpage).

PR #7067, opened 2015-10-04
Added support for Qianmo.com

taichatha:
This should work! I tested on a few videos, and it reads the video correctly and downloads them.

### DIFF
--- a/youtube_dl/extractor/qianmo.py
+++ b/youtube_dl/extractor/qianmo.py
@@ -29,15 +29,10 @@ class QianmoIE(InfoExtractor):
         video_id = self._match_id(url)
 
         webpage = self._download_webpage(url, video_id)
-        print type(webpage)
-        # TODO more code goes here, for example ...
         title = self._html_search_regex(r'<title>(.*?)</title>', webpage, 'title')
-        print title
-        # postInfo = self._html_search_regex(r'postInfo =(.*?);', webpage, u'postInfo')
-        # regex = 
+        
         video_url = self._html_search_regex(r'"urls":\["(.*?)"\],', webpage, 'url')
-        print video_url
-        print "Hanh"
+        
 
         return {
             'id': video_id,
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/qianmo.py
```diff
@@ -0,0 +1,43 @@
+# http://qianmo.com/bazingayue/324
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from lxml import html
+import requests
```
dstftw: These dependencies are unacceptable. Moreover you don'y use them.

youtube_dl/extractor/qianmo.py
```diff
@@ -0,0 +1,43 @@
...
+
+class QianmoIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?qianmo\.com/(?P<user>[A-Za-z]+)/(?P<id>[0-9]+)'
+    _TEST = {
+        'url': 'http://qianmo.com/kawaei/281',
+        'info_dict': {
+            'id': '281',
+            'ext': 'mp4',
+            'title': '[中字] Google新logo宣传片 「Google, 进化」 - 阡陌视频社区',
+            # TODO more properties, either as:
```
dstftw: Either **do** or remove.

youtube_dl/extractor/qianmo.py
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, video_id)
+        title = self._html_search_regex(r'<title>(.*?)</title>', webpage, 'title')
+        
+        video_url = self._html_search_regex(r'"urls":\["(.*?)"\],', webpage, 'url')
```
dstftw: `postInfo` contains far more metadata that should also be extracted. Consider [this](https://github.com/taichatha/youtube-dl/blob/master/youtube_dl/extractor/common.py#L63-L198) for a set of relevant fields.

youtube_dl/extractor/qianmo.py
taichatha: Sorry, I must have left those in there by accident.

yan12125:

PR #7089, opened 2015-10-07
Voissa new extractor

tiburce:
Add support  for this site with needed corrections
I learn git ... sorry for my fist commit and pull request erased by error.

Voissa site lag sometimes , but the script work well


tiburce:
I have testing again the script and there is no problem, i don't understand why travis test fail

TomGijselinck:
@tiburce To view the cause of the travis build failure, view the details of the Travis CI build. There you can check whether or not your test caused the build failure. As a side note, all builds are failing because some tests do not pass. So it's possible that your tests pass, but the build fails.

tiburce:
I've see that , but my test run and the travis test for this extractor fail.

yan12125:

PR #7092, opened 2015-10-07
Make the native hls downloader use disk space more efficiently

e00E:
It now saves the url of the most recently completed segment to figure out from where it has to resume.
Before, it kept ALL segments which resulted in using double the disk space neccessary.

### DIFF
--- a/youtube_dl/downloader/fragment.py
+++ b/youtube_dl/downloader/fragment.py
@@ -49,7 +49,10 @@ class FragmentFD(FileDownloader):
 
     def _start_frag_download(self, ctx):
         total_frags = ctx['total_frags']
-        downloaded_bytes = os.path.getsize(ctx['tmpfilename']) if ctx.get('continue_dl') else 0
+        try:
+            downloaded_bytes = os.path.getsize(ctx['tmpfilename']) if ctx.get('continue_dl') else 0
+        except os.error as e:
+            downloaded_bytes = 0
         frag_index = ctx['continue_fragment'] + 1 if ctx.get('continue_fragment') else 0
         # This dict stores the download progress, it's updated by the progress
         # hook
--- a/youtube_dl/downloader/fragment.py
+++ b/youtube_dl/downloader/fragment.py
@@ -21,11 +21,11 @@ class FragmentFD(FileDownloader):
     A base file downloader class for fragmented media (e.g. f4m/m3u8 manifests).
     """
 
-    def _prepare_and_start_frag_download(self, ctx, continue_dl=False, continue_fragment=None):
-        self._prepare_frag_download(ctx, continue_dl)
-        self._start_frag_download(ctx, continue_fragment)
+    def _prepare_and_start_frag_download(self, ctx):
+        self._prepare_frag_download(ctx)
+        self._start_frag_download(ctx)
 
-    def _prepare_frag_download(self, ctx, continue_dl=False):
+    def _prepare_frag_download(self, ctx):
         self.to_screen('[%s] Total fragments: %d' % (self.FD_NAME, ctx['total_frags']))
         self.report_destination(ctx['filename'])
         dl = HttpQuietDownloader(
@@ -40,20 +40,17 @@ class FragmentFD(FileDownloader):
             }
         )
         tmpfilename = self.temp_name(ctx['filename'])
-        dest_stream, tmpfilename = sanitize_open(tmpfilename, 'ab' if continue_dl else 'wb')
+        dest_stream, tmpfilename = sanitize_open(tmpfilename, 'ab' if ctx.get('continue_dl') else 'wb')
         ctx.update({
             'dl': dl,
             'dest_stream': dest_stream,
             'tmpfilename': tmpfilename,
         })
 
-    def _start_frag_download(self, ctx, continue_fragment=None):
-        # continue_fragment is the last fragment that was already downloaded
-        # when continuing an old download or None when not continuing
-
+    def _start_frag_download(self, ctx):
         total_frags = ctx['total_frags']
-        downloaded_bytes = 0 if continue_fragment is None else os.path.getsize(ctx['tmpfilename'])
-        frag_index = 0 if continue_fragment is None else continue_fragment + 1
+        downloaded_bytes = os.path.getsize(ctx['tmpfilename']) if ctx.get('continue_dl') else 0
+        frag_index = ctx['continue_fragment'] + 1 if ctx.get('continue_fragment') else 0
         # This dict stores the download progress, it's updated by the progress
         # hook
         state = {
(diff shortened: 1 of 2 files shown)


e00E:
This is for #7087 again.

youtube_dl/downloader/hls.py
```diff
@@ -81,7 +89,10 @@ def real_download(self, filename, info_dict):
                     line
                     if re.match(r'^https?://', line)
                     else compat_urlparse.urljoin(man_url, line))
-                fragment_urls.append(segment_url)
+                if arrived_at_last_downloaded_segment:
+                    fragment_urls.append(segment_url)
+                elif segment_url == last_downloaded_segment:
```
dstftw: Relying on segment's URL is bad idea since it may contain a dynamic part.

youtube_dl/downloader/hls.py
```diff
@@ -81,7 +89,10 @@ def real_download(self, filename, info_dict):
                     line
                     if re.match(r'^https?://', line)
                     else compat_urlparse.urljoin(man_url, line))
-                fragment_urls.append(segment_url)
+                if arrived_at_last_downloaded_segment:
+                    fragment_urls.append(segment_url)
```
dstftw: Since now you skip already downloaded segments the total fragment message displays wrong value after restarting: `[hlsnative] Total fragments: ...`.

youtube_dl/downloader/hls.py
```diff
@@ -102,11 +112,15 @@ def real_download(self, filename, info_dict):
             down, frag_sanitized = sanitize_open(frag_filename, 'rb')
             ctx['dest_stream'].write(down.read())
             down.close()
-            frags_filenames.append(frag_sanitized)
+            os.remove(encodeFilename(frag_sanitized))
+            segments_file = open(last_downloaded_segment_filename, 'w')
```
dstftw: Use `sanitize_open` or at least `encodeFilename`

youtube_dl/downloader/hls.py
```diff
@@ -102,11 +112,15 @@ def real_download(self, filename, info_dict):
...
+            segments_file.write(frag_url + "\n")
+            segments_file.close()
+            
 
         self._finish_frag_download(ctx)
 
-        for frag_file in frags_filenames:
-            os.remove(encodeFilename(frag_file))
+        if last_downloaded_segment is not None:
+            os.remove(last_downloaded_segment_filename)
```
dstftw: Use `encodeFilename`.

youtube_dl/downloader/hls.py
```diff
@@ -102,11 +112,15 @@ def real_download(self, filename, info_dict):
...
+            segments_file = open(last_downloaded_segment_filename, 'w')
+            segments_file.write(frag_url + "\n")
+            segments_file.close()
+            
 
         self._finish_frag_download(ctx)
 
-        for frag_file in frags_filenames:
-            os.remove(encodeFilename(frag_file))
+        if last_downloaded_segment is not None:
```
dstftw: Just `if last_downloaded_segment:` is enough.

youtube_dl/downloader/hls.py
```diff
@@ -102,11 +112,15 @@ def real_download(self, filename, info_dict):
             down, frag_sanitized = sanitize_open(frag_filename, 'rb')
             ctx['dest_stream'].write(down.read())
             down.close()
-            frags_filenames.append(frag_sanitized)
+            os.remove(encodeFilename(frag_sanitized))
+            segments_file = open(last_downloaded_segment_filename, 'w')
+            segments_file.write(frag_url + "\n")
```
dstftw: Consistently use single quotes.

dstftw:
Another problems:
* It does not work at all. After restarting it overwrite the old file since `FragementFD` does not open it in append mode.
* Fragment on which download was cancelled will be completely redownloaded upon resume. It should resume downloading segment from the place where is was interrupted whenever possible.

e00E:
Thank you for reviewing dstftw.

I fixed  most of the stuff but im not sure about your last two points in your last post.

The behavior for individual fragments that were interrupted should be unchanged from before. My patch does not touch this. I think it all relies on fragment.py setting the dl method in ctx correctly. To me it seems this should be a resumable download anyway (see fragment.py + 34).

I dont know how it would leave fragments undeleted on disk. Each downloaded fragments gets deleted right after copying its contents to the final file.

youtube_dl/downloader/fragment.py
```diff
@@ -21,11 +21,11 @@ class FragmentFD(FileDownloader):
     A base file downloader class for fragmented media (e.g. f4m/m3u8 manifests).
     """
 
-    def _prepare_and_start_frag_download(self, ctx):
-        self._prepare_frag_download(ctx)
-        self._start_frag_download(ctx)
+    def _prepare_and_start_frag_download(self, ctx, continue_dl=False, continue_fragment=None):
```
dstftw: Instead of adding new parameters put them into `ctx`.

e00E:
I tried to make it so other modules using FragmentFD that are not aware of the additional ctx parameters keep working as before.

e00E:
Any update?

youtube_dl/downloader/fragment.py
```diff
@@ -49,12 +49,14 @@ def _prepare_frag_download(self, ctx):
 
     def _start_frag_download(self, ctx):
         total_frags = ctx['total_frags']
+        downloaded_bytes = os.path.getsize(ctx['tmpfilename']) if ctx.get('continue_dl') else 0
```
dstftw: This breaks streaming to stdout.

e00E:
This is only used for for the progress display which you cannot see anyway when writing to stdout right?
In that case I could just check if the filename is `-` and simply set downloaded_bytes to 0 since the progress is irrelevant.
Do you agree?

The other solution I have in mind is storing the downloaded bytes in the `last_segment` file which should work but it would be unnecessary in the above case.

e00E:
I just tried writing to stdout and then redirecting that into a file (`> out.mp4`) and you do see the progress so I will implement the second solution.

dstftw:
Generally, not using `getsize` with `-` should be enough.

e00E:
Alright, so the first case after all. I used the exception as I thought it was more error proof than just checking tmpfile for `-` because maybe reading the size can file for other reasons too.

dstftw:
If it fails with correct filepath I don't think it should continue further since it will most likely fail on another file operations as well.

e00E:
I was thinking maybe there are cases where you cannot read but can write to a file or the filesystem doesnt allow you to get the size or something else esoteric.
If it fails on writing later then it will still just stop and I dont think having the file size call fail as an implicit check for write fails later makes sense.

PR #7103, opened 2015-10-08
Fixed 4tube.com extractor to pull metadata from associated Javascript…

magicalvidyakitty:
… and not the HTML of the desired page.


magicalvidyakitty:
cool thanks

PR #7105, opened 2015-10-08
[pbs] detect errors from http error code

remitamine:


youtube_dl/extractor/pbs.py
```diff
@@ -231,13 +233,18 @@ def _real_extract(self, url):
...
-                        'right restrictions')
+                http_code = redirect_info['http_code']
+                if http_code == 403:
+                    message = 'We\'re sorry, but this video is not available in your region due to right restrictions.'
+                elif http_code == 101:
+                    message = 'We\'re sorry, but this video is not yet available.'
+                elif http_code == 404:
+                    message = 'We are experiencing technical difficulties that are preventing us from playing the video at this time. Please check back again soon.'
+                elif http_code == 410:
+                    message = 'This video has expired and is no longer available for online streaming.'
```
dstftw: This could be better placed into a dict.

remitamine:
i place the errors in a dict.

dstftw:
Thanks.

PR #7114, opened 2015-10-09
[srgssr] Add generic extractor for SRGSSR Group sites

remitamine:
add support for srf.ch, rts.ch, rsi.ch, rtr.ch and swissinfo.ch play sites with a single extractor.


youtube_dl/extractor/srgssr.py
```diff
@@ -0,0 +1,127 @@
...
+        'md5': '0a274ce38fda48c53c01890651985bc6',
+        'info_dict': {
+            'id': '677f5829-e473-4823-ac83-a1087fe97faa',
+            'ext': 'flv',
+            'upload_date': '20130710',
+            'title': 'Jaguar XK120, Shadow und Tornado-Dampflokomotive',
+            'description': 'md5:88604432b60d5a38787f152dec89cd56',
+            'timestamp': 1373493600,
+        },
+    }]
```
jaimeMF: Add at least one test with a domain different to srf.ch

jaimeMF:
`urn:` looks too generic and unrelated to SRGSSR, you should use other shortcut.

remitamine:
i added another test for rtr audio.
and change the shortcut to start with srgssr.

PR #7116, opened 2015-10-09
[expotv] extract m3u8 formats

remitamine:


remitamine:
the m3u8 manifest is a Master Playlist even if it contain only one Media Playlist.

PR #7121, opened 2015-10-10
[adultswim] detect when video needs authentication

remitamine:


dstftw:
That's wrong. The video http://www.adultswim.com/videos/rick-and-morty/interdimensional-cable-2-tempting-fate/ requires authentication but can be downloaded as clips without it.

remitamine:
> That's wrong. The video http://www.adultswim.com/videos/rick-and-morty/interdimensional-cable-2-tempting-fate/ requires authentication but can be downloaded as clips without it.

it's not totally wrong, but as it can bypassed for some videos i changed the code to raise the error only if it can't find the clips.

dstftw:
Cherry-picked, thanks.

PR #7124, opened 2015-10-10
[5min] extract more video info and formats

remitamine:


youtube_dl/extractor/fivemin.py
```diff
@@ -59,26 +106,39 @@ def _real_extract(self, url):
...
+        if qs:
+            qs = '?' + qs
+        for rendition in info['Renditions']:
+            if rendition['RenditionType'] == 'm3u8':
+                formats.extend(self._extract_m3u8_formats(rendition['Url'], video_id, m3u8_id='hls'))
+            elif rendition['RenditionType'] == 'aac':
+                continue
+            else:
+                rendition_url = video_url[:10] + video_url[10:].replace('//', '/%s/' % rendition['ID'])
+                rendition_url = rendition_url[:rendition_url.rindex('.')] + '.' + rendition['RenditionType'] + qs
```
jaimeMF: I think this two lines are better using  compat_urllib_parse_urlparse and the `_replace` method:
* It's not immediately clear clear why you use `video_url[:10]`, it's better if you explicitly change the path
* `rendition_url.rindex('.')` won't works as expected if the query contains a dot, and normal `index` will match the dot from the domain.

remitamine: the code is a translation of the js code.
i will try to use the `compat_urllib_parse_urlparse` and `_replace`.

remitamine:
i changed the code to use `compat_urllib_parse_urlparse`,`compat_urlparse` and `replace_extension` to create rendition url.

remitamine:
i changed the code to use `compat_urllib_parse_urlparse`,`compat_urlparse` and `replace_extension` to create rendition url.

jaimeMF:
Thanks, cherry-picked in 26669ea3cf596f2ea4bce9e21ce73c1d8fc3ff72 with some minor changes (I recommend you to run `flake8 .` to check if there's any style issue).

PR #7143, opened 2015-10-12
[fc-zenit] Added new extractor

Kaspi:


youtube_dl/extractor/fczenit.py
```diff
@@ -0,0 +1,52 @@
...
+
+import os.path
+import re
+import json
+
+from ..compat import compat_urllib_parse_unquote
+from ..utils import url_basename
+from .common import InfoExtractor
+
+class fczenitIE(InfoExtractor):
```
jaimeMF: By convention class names start with a capital letter.

youtube_dl/extractor/fczenit.py
```diff
@@ -0,0 +1,52 @@
...
+    _VALID_URL = r'(?:https?://(?:www\.)?fc-zenit\.ru/video/gl(?P<id>[0-9]+))'
```
jaimeMF: There's no need to use  a group for the whole url.

youtube_dl/extractor/fczenit.py
```diff
@@ -0,0 +1,52 @@
...
+                "url" : sources[0],
+                "tbr": sources[1]
+        } for sources in bitrates]
+
+        self._sort_formats(formats)
+
+        return {
+            'id' : video_id,
+            'title' : video_title,
+            'url' : webpage_url,
```
jaimeMF: `'url'` is for the actual video url, since you already use `'formats'` it's not needed.

youtube_dl/extractor/fczenit.py
```diff
@@ -0,0 +1,52 @@
...
+            'ext' : u'mp4',
```
jaimeMF: `'ext'` it's automatically inferred from the url if you don't give it, in this case it can be omitted.

youtube_dl/extractor/fczenit.py
```diff
@@ -0,0 +1,52 @@
...
+import json
+
+from ..compat import compat_urllib_parse_unquote
+from ..utils import url_basename
+from .common import InfoExtractor
+
+class fczenitIE(InfoExtractor):
+    _VALID_URL = r'(?:https?://(?:www\.)?fc-zenit\.ru/video/gl(?P<id>[0-9]+))'
+    _TEST = {
+    u'url': u'http://fc-zenit.ru/video/gl6785/',
```
jaimeMF: Since you used `from __future__ import unicode_literals`, you don't need the `u` prefix.

youtube_dl/extractor/fczenit.py
```diff
@@ -0,0 +1,52 @@
...
+import re
+import json
+
+from ..compat import compat_urllib_parse_unquote
+from ..utils import url_basename
+from .common import InfoExtractor
+
+class fczenitIE(InfoExtractor):
+    _VALID_URL = r'(?:https?://(?:www\.)?fc-zenit\.ru/video/gl(?P<id>[0-9]+))'
+    _TEST = {
```
jaimeMF: We indent the dictionary entries one level.

youtube_dl/extractor/fczenit.py
```diff
@@ -0,0 +1,52 @@
...
+    u'url': u'http://fc-zenit.ru/video/gl6785/',
+    u'md5' : '458bacc24549173fe5a5aa29174a5606',
+    u'info_dict': {
+        u"id": u"6785",
```
jaimeMF: We always use single quotes.

youtube_dl/extractor/fczenit.py
```diff
@@ -0,0 +1,52 @@
...
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('id')
+        webpage_url = 'http://fc-zenit.ru/video/gl' + video_id
+        webpage = self._download_webpage(webpage_url, video_id)
+
+        video_title = self._html_search_regex(r'<div class=\"photoalbum__title\">([^<]+)', webpage, u"title")
+
+        # Log that we are starting to parse the page
+        self.report_extraction(video_id)
```
jaimeMF: We don't this method in new extractors, we already log in `_download_webpage` which is usually the slowest part.

jaimeMF:
Thanks, cherry picked in 3dc582e5ea69af4ad7f51d30c1d87cf93aa6b72b with the comments I made fixed.

jaimeMF:
Do you want to be credited in the AUTHORS file? if so under which name?

PR #7145, opened 2015-10-12
[Canvas] Add new extractor 

### LINKED ISSUE #7079: Unsupported url: canvas.be
```
$ youtube-dl http://www.canvas.be/video/radio-gaga/najaar-2015/internaat-turnhout --verbose
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['http://www.canvas.be/video/radio-gaga/najaar-2015/internaat-turnhout', '--verbose']
[debug] youtube-dl version 2014.08.05
[debug] Python version 2.7.9 - Linux-3.16.0-4-amd64-x86_64-with-debian-8.2
[debug] Proxy map: {}
[generic] internaat-turnhout: Requesting header
[generic] internaat-turnhout: Downloading webpage
[generic] internaat-turnhout: Extracting information
ERROR: Unsupported URL: http://www.canvas.be/video/radio-gaga/najaar-2015/internaat-turnhout; please report this issue on https://yt-dl.org/bug . Be sure to call youtube-dl with the --verbose flag and include its complete output. Make sure you are using the latest version; type  youtube-dl -U  to update.
  File "/usr/lib/python2.7/dist-packages/youtube_dl/extractor/generic.py", line 457, in _real_extract
    doc = parse_xml(webpage)
  File "/usr/lib/python2.7/dist-packages/youtube_dl/utils.py", line 1417, in parse_xml
    return xml.etree.ElementTree.XML(s.encode('utf-8'), **kwargs)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1300, in XML
    parser.feed(text)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1642, in feed
    self._raiseerror(v)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1506, in _raiseerror
    raise err
ParseError: not well-formed (invalid token): line 8, column 373
  File "/usr/lib/python2.7/dist-packages/youtube_dl/YoutubeDL.py", line 516, in extract_info
    ie_result = ie.extract(url)
  File "/usr/lib/python2.7/dist-packages/youtube_dl/extractor/common.py", line 170, in extract
  File "/usr/lib/python2.7/dist-packages/youtube_dl/extractor/generic.py", line 752, in _real_extract
    raise ExtractorError('Unsupported URL: %s' % url)
ExtractorError: Unsupported URL: http://www.canvas.be/video/radio-gaga/najaar-2015/internaat-turnhout; please report this issue on https://yt-dl.org/bug . Be sure to call youtube-dl with the --verbose flag and include its complete output. Make sure you are using the latest version; type  youtube-dl -U  to update.
```

TomGijselinck:
Resolves #7079. Code is checked with flake8.

### DIFF
--- a/youtube_dl/extractor/canvas.py
+++ b/youtube_dl/extractor/canvas.py
@@ -1,0 +1,77 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from youtube_dl import utils
+from .common import InfoExtractor
+
+
+class CanvasIE(InfoExtractor):
+    _VALID_URL = r'(?:https?://)?(?:www\.)?canvas\.be/video/(?P<id>.+)'
+    _TEST = {
+        'url': 'http://www.canvas.be/video/de-afspraak/najaar-2015/de-afspraak-veilt-voor-de-warmste-week',
+        'md5': 'ea838375a547ac787d4064d8c7860a6c',
+        'info_dict': {
+            'id': 'de-afspraak/najaar-2015/de-afspraak-veilt-voor-de-warmste-week',
+            'title': 'De afspraak veilt voor de Warmste Week',
+            'ext': 'mp4',
+            'duration': 49,
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._search_regex(
+            r'<h1 class="video__body__header__title">(.+?)</h1>', webpage,
+            'title')
+        data_video = self._html_search_regex(
+            r'data-video=(["\'])(?P<id>.+?)\1', webpage, 'data-video', group='id')
+        json_url = 'https://mediazone.vrt.be/api/v1/canvas/assets/' + data_video
+        data = self._download_json(json_url, video_id)
+
+        formats = []
+        for target in data['targetUrls']:
+            if 'type' and 'url' in target:
+                extension = utils.determine_ext(target['url'])
+                if target['type'] == 'PROGRESSIVE_DOWNLOAD':
+                    formats.append({
+                        'format_id': extension,
+                        'url': target['url'],
+                        'protocol': 'http',
+                    })
+                elif target['type'] == 'HLS':
+                    formats.extend(self._extract_m3u8_formats(
+                        target['url'], video_id, entry_protocol='m3u8_native',
+                        ext='mp4',
+                        preference=0,
+                        fatal=False,
+                        m3u8_id='hls'))
+                elif target['type'] == 'HDS':
+                    formats.append({
+                        'format_id': extension,
+                        'url': target['url'],
+                        'protocol': 'HDS',
+                    })
+                elif target['type'] == 'RTMP':
+                    formats.append({
+                        'format_id': extension,
+                        'url': target['url'],
+                        'protocol': 'rtmp',
+                    })
+                elif target['type'] == 'RTSP':
+                    formats.append({
+                        'format_id': extension,
+                        'url': target['url'],
+                        'protocol': 'rtsp',
+                    })
+
+        self._sort_formats(formats)
+        duration = utils.int_or_none(data.get('duration')) / 1000
+        return {
+            'id': video_id,
+            'title': title,
+            'formats': formats,
+            'duration': duration,
+        }
--- a/youtube_dl/extractor/__init__.py
+++ b/youtube_dl/extractor/__init__.py
@@ -70,6 +70,7 @@ from .camdemy import (
 from .canal13cl import Canal13clIE
 from .canalplus import CanalplusIE
 from .canalc2 import Canalc2IE
+from .canvas import CanvasIE
 from .cbs import CBSIE
 from .cbsnews import CBSNewsIE
 from .cbssports import CBSSportsIE


youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,69 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+import requests
```
dstftw: Dependency on `requests` is unacceptable.

TomGijselinck:
Dependency on `requests` removed.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,68 @@
...
+        'info_dict': {
+            'id': 'radio-gaga/najaar-2015/internaat-turnhout',
+            'title': 'Internaat Turnhout',
+            'ext': 'mp4',
+            'duration': 2467160,
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = re.match(self._VALID_URL, url).group('id')
```
yan12125: Use ```self._match_id``` is better.

TomGijselinck: Thanks for pointing that out!

TomGijselinck:
Build failure is **not** caused by `test_Canvas`, which passes without any problems.

TomGijselinck:
@yan12125 Is there something I have to fix? Because the PR won't get accepted and I do not know what's wrong.

yan12125:
I have little time on writing codes these days. @dstftw just merges it if you think it's OK.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,66 @@
...
+                })
+
+        self._sort_formats(formats)
+        ext = 'mp4'
+        duration = data['duration']
+        return {
+            'id': video_id,
+            'title': title,
+            'formats': formats,
+            'ext': ext,
```
dstftw: `ext` here makes no sense.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,66 @@
...
+                    'url': target['url'],
+                    'protocol': 'http',
+                })
+            elif target['type'] == 'HLS':
+                formats.extend(self._extract_m3u8_formats(
+                    target['url'], video_id, entry_protocol='m3u8_native', ext='mp4',
+                    preference=0))
+            elif target['type'] == 'RTMP':
+                formats.append({
+                    'format-id': 'mp4',
```
dstftw: Correct field name is `format_id`.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._html_search_regex(
+            r'<h1 class="video__body__header__title">(.*?)</h1>', webpage, 'title')
+        data_video = self._html_search_regex(
```
dstftw: `self._search_regex` is enough here.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,66 @@
...
+            r'data-video="(.*?)"', webpage, 'data-video')
```
dstftw: Matched data-video should not be empty.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,66 @@
...
+        json_url = 'https://mediazone.vrt.be/api/v1/canvas/assets/' + data_video
+        data = self._download_json(json_url, video_id)
+
+        formats = []
+        for target in data['targetUrls']:
+            if target['type'] == 'PROGRESSIVE_DOWNLOAD':
```
dstftw: Missing `type` and `url` in `target` should be tolerated without extractor breakage.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,66 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class CanvasIE(InfoExtractor):
+    _VALID_URL = r'(?:https?://)?(?:www\.)?canvas\.be/video/(?P<id>((([[a-z0-9]+-?)+/?)+))'
```
dstftw: No groups apart from `id` should be captured.

midas02:
As per my comment on https://github.com/rg3/youtube-dl/issues/7079#issuecomment-168800544, there is an existing extractor for Flemish broadcaster VRT called vrt.py. Please include this website into that one, instead of adding another extractor which will basically do the same.

dstftw:
@midas02 they use totally different embedding.

TomGijselinck:
@dstftw Thanks for taking your time for reviewing. I applied your remarks and I hope it's now ready to merge.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,68 @@
...
+                        'url': target['url'],
+                        'protocol': 'rtmp',
+                    })
+                elif target['type'] == 'RTSP':
+                    formats.append({
+                        'format_id': 'mp4',
+                        'url': target['url'],
+                        'protocol': 'rtsp',
+                    })
+        except KeyError:
```
dstftw: If first target happens to fail no formats will be extracted at all.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,68 @@
...
+class CanvasIE(InfoExtractor):
+    _VALID_URL = r'(?:https?://)?(?:www\.)?canvas\.be/video/(?P<id>.+)'
+    _TEST = {
+        'url': 'http://www.canvas.be/video/slijk/2015/isaac-kimeli',
+        'md5': 'c8a68bce36a4587c1dfb9a7e96a796fe',
+        'info_dict': {
+            'id': 'slijk/2015/isaac-kimeli',
+            'title': 'Veldlopen: Isaac Kimeli',
+            'ext': 'mp4',
+            'duration': 1731000,
```
dstftw: Duration must be in seconds.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,68 @@
+# coding: utf-8
+from __future__ import unicode_literals
+from .common import InfoExtractor
+
+
+class CanvasIE(InfoExtractor):
+    _VALID_URL = r'(?:https?://)?(?:www\.)?canvas\.be/video/(?P<id>.+)'
+    _TEST = {
+        'url': 'http://www.canvas.be/video/slijk/2015/isaac-kimeli',
```
dstftw: This video is georestricted.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,68 @@
...
+                    formats.append({
+                        'format_id': 'mp4',
+                        'url': target['url'],
+                        'protocol': 'rtsp',
+                    })
+        except KeyError:
+            pass
+
+        self._sort_formats(formats)
+        duration = data['duration']
```
dstftw: This also should not break extraction if it's missing.

youtube_dl/extractor/canvas.py
TomGijselinck: @dstftw I tried multiple other videos with a proxy server and none would show up. Is it possible that all videos on this website have georestriction? And in this case, should I remove the test? I implemented your other remarks and everything works fine.

dstftw: This one is not restricted http://www.canvas.be/video/de-afspraak/najaar-2015/de-afspraak-veilt-voor-de-warmste-week.

TomGijselinck: @dstftw Thank you! This video neither had georestriction nor a date limit. Perfect.

TomGijselinck:
@dstftw I fixed the issues you mentioned in previous diff. I hope everything is alright now.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,67 @@
...
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._search_regex(
+            r'<h1 class="video__body__header__title">(.*?)</h1>', webpage,
+            'title')
+        data_video = self._html_search_regex(
+            r'data-video="(.+)"', webpage, 'data-video')
```
dstftw: Change to `r'data-video=(["\'])(?P<id>.+?)\1, webpage, 'data-video', group='id')`.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._search_regex(
+            r'<h1 class="video__body__header__title">(.*?)</h1>', webpage,
```
dstftw: Change to `(.+?)`.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,67 @@
...
+                elif target['type'] == 'RTSP':
+                    formats.append({
+                        'format_id': 'mp4',
+                        'url': target['url'],
+                        'protocol': 'rtsp',
+                    })
+
+        self._sort_formats(formats)
+        if 'duration' in data:
+            duration = int(data['duration']) / 1000
```
dstftw: Use `utils.int_or_none` instead. Use `data.get('duration')` instead of testing keys manually.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,67 @@
...
+                    formats.append({
+                        'format_id': 'mp4',
+                        'url': target['url'],
+                        'protocol': 'http',
+                    })
+                elif target['type'] == 'HLS':
+                    formats.extend(self._extract_m3u8_formats(
+                        target['url'], video_id, entry_protocol='m3u8_native',
+                        ext='mp4',
+                        preference=0))
```
dstftw: Pass `fatal=False`.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,67 @@
...
+            r'data-video="(.+)"', webpage, 'data-video')
+        json_url = 'https://mediazone.vrt.be/api/v1/canvas/assets/' + data_video
+        data = self._download_json(json_url, video_id)
+
+        formats = []
+        for target in data['targetUrls']:
+            if 'type' and 'url' in target:
+                if target['type'] == 'PROGRESSIVE_DOWNLOAD':
+                    formats.append({
+                        'format_id': 'mp4',
```
dstftw: `format_id` should be meaningful and not the same for every format.

youtube_dl/extractor/canvas.py
TomGijselinck: @dstftw If I change it like that, running the `test/test_download.py` gives me a ` DownloadError` because the JSON metadata is not found.

dstftw: Missed `'` after \1.

youtube_dl/extractor/canvas.py
```diff
@@ -0,0 +1,67 @@
...
+                        'url': target['url'],
+                        'protocol': 'http',
+                    })
+                elif target['type'] == 'HLS':
+                    formats.extend(self._extract_m3u8_formats(
+                        target['url'], video_id, entry_protocol='m3u8_native',
```
dstftw: Also pass `m3u8_id='hls'`.

youtube_dl/extractor/canvas.py
TomGijselinck: @dstftw This gives me another `DownloadError: Unable to extract data-video;`. I actually don't understand why it should be this (`r'data-video=(["\'])(?P<id>.+?)\1\''`) regex instead of `r'data-video="(.+)"'` because the latter will find the complete `data-video` value which I can not see how the former can do that.

dstftw: `r'data-video="(.+)"'` is greedy and therefore will match everything from the first to the last quote. If webpage layout will change to something like `data-video="mz-ast-5e5f90b6-2d72-4c40-82c2-e134f884e93e" data-preroll="true"` your regex will match `mz-ast-5e5f90b6-2d72-4c40-82c2-e134f884e93e" data-preroll="true`.

dstftw: Just copy paste the whole line I've posted.

dstftw:
When finished squash commits to a single one.

youtube_dl/extractor/canvas.py
TomGijselinck: @dstftw Ah, yes, I understand. Thank you, it works now.

TomGijselinck:
@dstftw I fixed all issues you mentioned.

TomGijselinck:
@dstftw I fixed all issues you mentioned.

dstftw:
Cherry-picked in be2d40a58a4f89682859bb23789b75330ee756a6, thansk.

PR #7147, opened 2015-10-12
[nowtv] added playlist support

AndroKev:
Now you can download a season or the full serie!
example:
http://www.nowtv.at/rtl/das-supertalent/list/free-staffel-8
http://www.nowtv.at/rtl/das-supertalent

both works now and download the season 8 (url1) or the whole serie (url2)


dstftw:
It should be implemented as separated extractor.

AndroKev:
I tried it but I had an problem with the valid url:

http://www.nowtv.de/rtl2/berlin-tag-nacht/berlin-tag-nacht-folge-934/player
_VALID_URL = r'https?://(?:www\.)?nowtv\.(?:de|at|ch)/(?:rtl|rtl2|rtlnitro|superrtl|ntv|vox)/(?P<id>.+?)/(?:player|preview)' <--works

but now i need a valid url for this:
http://www.nowtv.at/rtl/stern-tv/list/aktuell
and this:
http://www.nowtv.at/rtl/das-supertalent/list/free-staffel-8

AndroKev:
any ideas how i get this to work?

ps. https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/prosiebensat1.py
uses also only one extractor

Issue #7032, opened 2015-10-01
vidto.me Unsupported URL


follex:
Hi :)
if I use the command
<code> youtube-dl -F http://vidto.me/j38j9zd9g6ge.html </code>

It returns these errors
<code>
[generic] j38j9zd9g6ge: Requesting header
WARNING: Falling back on generic information extractor.
[generic] j38j9zd9g6ge: Downloading webpage
[generic] j38j9zd9g6ge: Extracting information
ERROR: Unsupported URL: http://vidto.me/j38j9zd9g6ge.html
</code>

Yet Vidto.me is in the list

yan12125:
I guess you're looking at vid.me entry on [supported sites list](https://github.com/rg3/youtube-dl/blob/master/docs/supportedsites.md)? They're different sites.

Another point is that this video looks like [a copyrighted artwork](https://en.wikipedia.org/wiki/Wag_the_Dog). [youtube-dl does not support sites with illegal media](https://github.com/rg3/youtube-dl#can-you-add-support-for-this-anime-video-site-or-site-which-shows-current-movies-for-free). Do you have evidence that vidto.me has the rights for playing this video? For example, did they buy/rent the copyright of this video from the original copyright owners?

follex:
Oh sorry, I confused  Vid.me with Vidto.me

yan12125:
According to [Google transparency report for vidme.to](https://www.google.com/transparencyreport/removals/copyright/domains/vidto.me/), this website is full of illegal contents. Sorry youtube-dl will NOT support it.

PR #7167, opened 2015-10-14
[vidto] Add extractor

ferama:


docs/supportedsites.md
```diff
@@ -619,6 +620,7 @@
  - **VideoTt**: video.tt - Your True Tube
  - **videoweed**: VideoWeed
  - **Vidme**
+ - **vidto**: VidTo.me
```
yan12125: This file is updated automatically in each stable release. Please do not change its content in normal commits.

youtube_dl/extractor/vidto.py
```diff
@@ -0,0 +1,82 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+import sys
+from .common import InfoExtractor
```
yan12125: Built-in modules should be imported first.

youtube_dl/extractor/vidto.py
```diff
@@ -0,0 +1,82 @@
...
+        'url': 'http://vidto.me/ku5glz52nqe1.html',
+        'info_dict': {
+            'id': 'ku5glz52nqe1',
+            'ext': 'mp4',
+            'title': 'test.mp4'
+        }
+    }
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
```
yan12125: ```self._match_id``` is better here.

yan12125:
Well, vidto.me specializes in copyright infringement (see #7032). This website will not be supported by youtube-dl until they've removed illegal contents. Sorry for annoyance and being late.

dstftw:
Content on vidto.me is uploaded by users. The general rule is that if example URL that does not violate any copyright (e.g. one provided in this PR) can be found then the website can be supported by youtube-dl.

youtube_dl/extractor/vidto.py
```diff
@@ -0,0 +1,82 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+import sys
+from .common import InfoExtractor
+import time
+
+from ..utils import (
+    encode_dict,
```
yan12125: Parentheses are not necessary when there's only one symbol.

youtube_dl/extractor/vidto.py
```diff
@@ -0,0 +1,82 @@
...
+            'usr_login': '',
+            'id': id_value,
+            'fname': title,
+            'referer': '',
+            'hash': hash_value,
+        }
+        post_data = compat_urllib_parse.urlencode(encode_dict(form_str)).encode('ascii')
+        req = compat_urllib_request.Request(url, post_data)
+        req.add_header('Content-type', 'application/x-www-form-urlencoded')
+        for key, morsel in cookies.iteritems():
```
yan12125: ```dict.iteritems``` exists only on python 2.

youtube_dl/extractor/vidto.py
```diff
@@ -0,0 +1,82 @@
...
+            req.add_header('Cookie', '%s=%s' % (morsel.key, morsel.value))
```
yan12125: It should be ```(key, morsel)``` instead of ```(morsel.key, morsel.value)``` here. Also, cookies should be merged into a single header line as multiple ```Cookie:``` headers violate the [HTTP standard](http://tools.ietf.org/html/rfc6265#section-5.4).

youtube_dl/extractor/vidto.py
```diff
@@ -0,0 +1,82 @@
...
+
+        print("Waiting for countdown...")
+        time.sleep(7)
```
yan12125: Is is necessary for correct extraction?
And don't use ```print()``` directly. Use ```self.to_screen()``` instead.

youtube_dl/extractor/vidto.py
```diff
@@ -0,0 +1,82 @@
...
+        post_result = self._download_webpage(
+            req, None,
+            note='Proceed to video...', errnote='unable to proceed', fatal=True)
```
yan12125: For ```self._download_webpage```:
1. The second parameter (video_id) should not be ```None```.
2. The default value for ```fatal``` is already ```True```.

youtube_dl/extractor/vidto.py
```diff
@@ -0,0 +1,82 @@
...
+
+        file_link_regex = r'file_link ?= ?\'(https?:\/\/[0-9a-zA-z.\/\-_]+)'
```
yan12125:
```
r'file_link\s*=\s*\'(https?:\/\/[0-9a-zA-z.\/\-_]+)'
```
is better.

youtube_dl/extractor/vidto.py
ferama: Yes is necessary

ferama:
This is no more an extractor pull request, I'm applying my fast fixes on this branch :) For my needs I think it's a better solution. I leave the request open, just in case.

yan12125:
In usual cases, it's not a good idea to mix multiple sites in a single pull request. I guess Nowvideo is not related to Vidto? If so, please open a separate pull request.

ferama:
You are slow (for what I can see) at merge and I need an "always working for my needs branch"

Nowvideo is not related to vidto. Close the pull request if you want :)

Inviato da iPad

yan12125:
Cherry-picked at a625e5654365a8a246fa5fc5f5d347adc67c5617 and 42fc93c70928d34d46ceda644eb870e86daa8393. By the way, NowVideo is fixed in rg3/master. Please check it and propse your own patch if there are still problems.

yan12125:
@ferama Would you like to be listed in [AUTHORS](https://github.com/rg3/youtube-dl/blob/master/AUTHORS)? If so, under what name?

ferama:
As you wish, I'm Marco Ferragina. I can use rg3/master again, thanks :)

yan12125:
Well ```vidto.py``` is merged into ```gorillavid.py```, and the latter is renamed to ```xfileshare.py```. Anyway, I still list your name in AUTHORS for your other contributions.

Issue #7134, opened 2015-10-11
clyp.it support


0x27:
I can probably add this one myself, but its going to take a day or three to acquaint myself with the codebase properly, but if someone else gets there first or whatever thats cool :)

TL;DR, [here is my reference implementation for downloading individual files from clyp.it](https://github.com/0x27/clypdown). Its fairly straightforward.

Example URL: https://clyp.it/hrvttfzp  
Basically, you rewrite it to `https://api.clyp.it/hrvttfzp`, extract `Title` and `Mp3Url` from the JSON response, and download `Mp3Url`.

dstftw:
Post [an example URL](https://github.com/rg3/youtube-dl/blob/master/README.md#is-the-description-of-the-issue-itself-sufficient).

0x27:
Updated with example url and description of how their API is used in this instance. Still reading over some of the extractor examples, to figure out how youtube-dl does it :)

CianLR:
Just made a pull request adding clyp.it

dstftw:
Clyp will be supporte in the next version.

PR #7194, opened 2015-10-16
Added clyp.it support (Issue: #7134)

CianLR:


youtube_dl/extractor/clyp.py
```diff
@@ -0,0 +1,53 @@
...
+            'title': 'Krisson80 - bits wip wip',
+            'description': '#Krisson80BitsWipWip #chiptune\n#wip',
+        },
+    }, {
+        'url': 'https://clyp.it/ojz2wfah',
+        'only_matching': True,
+    }]
+
+    def _real_extract(self, url):
+        video_id = url.replace("https://clyp.it/", "")
```
dstftw: The idiomatic way to extract `id` is to use a group in `_VALID_URL`.

youtube_dl/extractor/clyp.py
```diff
@@ -0,0 +1,53 @@
...
+        api_url = 'https://api.clyp.it/' + video_id
+        metadata = self._download_json(api_url, video_id)
+
+        title = metadata['Title']
+        description = metadata['Description']
```
dstftw: Every field apart from title should be treated as optional and extraction should not fail when it's missing.

dstftw:
Cherry-picked in 4e16c1f80b009001aaea6f8baca5dfbfa9b58c11. Thanks.

Issue #5766, opened 2015-05-20
[site-support-request] music.naver.com


githubtea:
http://music.naver.com/artist/videoPlayer.nhn?videoId=50290

Support this site, please~
Thanks in advance. :)

Issue #2943, opened 2014-05-18
[site-support-request] movie.naver.com


ubless607:
<pre>C:\rtmp>youtube-dl --verbose http://movie.naver.com/movie/bi/mi/mediaView.nhn?code=57278&mid=22567

[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['--verbose', 'http://movie.naver.com/movie/bi/mi/mediaView.nhn?code=57278']
[debug] Encodings: locale cp949, fs mbcs, out cp949, pref cp949
[debug] youtube-dl version 2014.08.29
[debug] Python version 2.7.8 - Windows-7-6.1.7601-SP1
[debug] Proxy map: {}
[generic] mediaView: Requesting header
WARNING: Falling back on generic information extractor.
[generic] mediaView: Downloading webpage
[generic] mediaView: Extracting information
ERROR: Unsupported URL: http://movie.naver.com/movie/bi/mi/mediaView.nhn?code=57278; please report this issue on https://yt-dl.org/bug . Be sure to call youtube-dl with the --verbose flag and include its complete output. Make sure you are using the latest version; type  youtube-dl -U  to update.
Traceback (most recent call last):
  File "youtube_dl\extractor\generic.pyo", line 530, in _real_extract
  File "youtube_dl\utils.pyo", line 1469, in parse_xml
  File "xml\etree\ElementTree.pyo", line 1300, in XML
  File "xml\etree\ElementTree.pyo", line 1642, in feed
  File "xml\etree\ElementTree.pyo", line 1506, in _raiseerror
ParseError: mismatched tag: line 84, column 2
Traceback (most recent call last):
  File "youtube_dl\YoutubeDL.pyo", line 521, in extract_info
  File "youtube_dl\extractor\common.pyo", line 178, in extract
  File "youtube_dl\extractor\generic.pyo", line 869, in _real_extract
ExtractorError: Unsupported URL: http://movie.naver.com/movie/bi/mi/mediaView.nhn?code=57278; please report this issue on https://yt-dl.org/bug . Be sure to call youtube-dl with the --verbose flag and include its complete output. Make sure you are using the latest version; type  youtube-dl -U  to update.

'mid'은(는) 내부 또는 외부 명령, 실행할 수 있는 프로그램, 또는
배치 파일이 아닙니다.
</pre>

mpenkov:
Still unsupported in the current master:

```
[...]

Issue #2873, opened 2014-05-09
[site-support-request] sports.naver.com


ubless607:
<blockquote>http://sports.news.naver.com/sports/index.nhn?category=baseball&ctg=news&mod=read&office_id=295&article_id=0000001186</blockquote>

Issue #2872, opened 2014-05-09
[site-support-request] news.naver.com


ubless607:
<blockquote>http://news.naver.com/main/read.nhn?mode=LPOD&mid=tvh&oid=437&aid=0000020470<br> http://news.naver.com/main/read.nhn?mode=LPOD&mid=tvh&oid=214&aid=0000340102 <br>http://news.naver.com/main/read.nhn?mode=LPOD&mid=tvh&oid=056&aid=0010013683</blockquote>

PR #7203, opened 2015-10-17
[naver] add support for other naver sites

remitamine:
it works for most of the urls in this issues: #2943, #2873, #2872, #5766.


jgschwartz:
Why did these changes never get pulled?

Issue #7144, opened 2015-10-12
letv.py can not download true video


newpoyang:
letv.py can not download true video file, it download letv's ad video file

yan12125:
FYI: #6376 works for most cases, while it has some flaws.

yan12125:
Letv (now renamed to LeEco) is working after #7208 merged. Feel free to open new issues if there are still problems.

PR #6376, opened 2015-07-27
Letv fix

yan12125:
This fixes #6375.

Sorry that I've done the work several days ago, but I hesitated to push it to master as the implementation looks dirty. I can't figure out a better solution.

The final Letv m3u8 manifest is 'encrypted', so I need to decrypt it in ```HlsFD```/```NativeHlsFD```. My approach is to register a callback function in the extractor and call it in the downloader. A function pool is necessary because all values in ```info_dict``` should be serializable to JSON.

PR #7208, opened 2015-10-17
[Letv] Fix extraction

### LINKED ISSUE #6192: Letv Incomplete download
Recent usage of youtube-dl with letv.com causes problems where the video is incompletely fetched. For example, using it on the following video:
<pre>
youtube-dl -f 350 http://www.letv.com/ptv/vplay/1839416.html
</pre>
will download a 24 MB file. But originally, the lowest resolution (at 350) should be 79 MB. The same goes for other resolutions. The video notably stops at around the 10 minute mark, even though the entire video should be much longer.

It may be that the videos by Letv are now split up like Youku and Tudou. The flvcd site, though, suggests this is not the case.

yan12125:
This is an enhancement of #6376. This patch also fixes #6192 and #7144.

```compat_urllib_request_DataHandler``` is ported from [cpython hg rev 1733b3bd46db](https://hg.python.org/cpython/file/1733b3bd46db/Lib/urllib/request.py#l1599).

Cc @jaimeMF

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -37,6 +37,7 @@ from .compat import (
     compat_tokenize_tokenize,
     compat_urllib_error,
     compat_urllib_request,
+    compat_urllib_request_DataHandler,
 )
 from .utils import (
     ContentTooShortError,
@@ -1967,8 +1968,9 @@ class YoutubeDL(object):
         debuglevel = 1 if self.params.get('debug_printtraffic') else 0
         https_handler = make_HTTPS_handler(self.params, debuglevel=debuglevel)
         ydlh = YoutubeDLHandler(self.params, debuglevel=debuglevel)
+        data_handler = compat_urllib_request_DataHandler()
         opener = compat_urllib_request.build_opener(
-            proxy_handler, https_handler, cookie_processor, ydlh)
+            proxy_handler, https_handler, cookie_processor, ydlh, data_handler)
 
         # Delete the default user-agent header, which would otherwise apply in
         # cases where our custom HTTP handler doesn't come into play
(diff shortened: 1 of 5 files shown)


jaimeMF:
Everything seem to work, just two comments:
*  I would clean up the commit message from: https://github.com/yan12125/youtube-dl/commit/0a740fdbc07f763ef727c0cfad8801b7ee1b77e4, it references a few things that are not used anymore (function_pool and "Support preprocessing of m3u8 manifests").
* It seems that mime types are usually in lower-case, is there some reason to use `application/x-mpegURL`?

yan12125:
I took the MIME type from Apple's doc [Deploying HTTP Live Streaming](https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/StreamingMediaGuide/DeployingHTTPLiveStreaming/DeployingHTTPLiveStreaming.html), and according to section 5.1 "
Syntax of the Content-Type Header Field" of [RFC2045](https://www.ietf.org/rfc/rfc2045.txt), upper-case english letters are allowed in the ```Content-Type``` field.

yan12125:
Commit messages modified via ```rebase```. Thanks for pointing that out.

jaimeMF:
IANA uses `application/vnd.apple.mpegurl`: https://www.iana.org/assignments/media-types/application/vnd.apple.mpegurl, https://www.iana.org/assignments/media-types/media-types.xhtml#application (this is what python's mimetypes module follows), at least vimeo also uses that mimetype for m3u8 manifest. If that's the most common we could use that instead, but it's not really important.

yan12125:
Thanks. Updated as I think following Python's implementation is better.

jaimeMF:
Unless someone disagrees with using `data:*`urls, I think you can merge it. Nice work.

dstftw:
LGTM.

yan12125:
Merged.

PR #7213, opened 2015-10-17
[soundcloud] Add Soundcloud search extractor

reiv:
It would be nice to be able to search for tracks on Soundcloud directly from youtube-dl.

I've added a `SoundcloudSearchIE` extractor that is invoked with `scsearch:"query"`. It uses the Soundcloud API v2.

Thanks.

### DIFF
--- a/youtube_dl/extractor/soundcloud.py
+++ b/youtube_dl/extractor/soundcloud.py
@@ -532,11 +532,7 @@ class SoundcloudSearchIE(SearchInfoExtractor, SoundcloudIE):
             collection_id='Query "{0}"'.format(query),
             q=query.encode('utf-8'))
 
-        for _ in range(n):
-            try:
-                track = next(tracks)
-            except StopIteration:
-                break
+        for track in itertools.islice(tracks, n):
             uri = track['uri']
             title = track['title']
             results.append(self.url_result(url=uri))
--- a/youtube_dl/extractor/soundcloud.py
+++ b/youtube_dl/extractor/soundcloud.py
@@ -535,7 +535,7 @@ class SoundcloudSearchIE(SearchInfoExtractor, SoundcloudIE):
 
         if not results:
             raise ExtractorError(
-                '[soundcloud] No track results', expected=True)
+                'Soundcloud said: No track results', expected=True)
         
-        return self.playlist_result(results[:n], playlist_title=query)
+        return self.playlist_result(results, playlist_title=query)
 
--- a/youtube_dl/extractor/soundcloud.py
+++ b/youtube_dl/extractor/soundcloud.py
@@ -526,16 +526,12 @@ class SoundcloudSearchIE(SearchInfoExtractor, SoundcloudIE):
             next_url = response.get('next_href', None)
 
     def _get_n_results(self, query, n):
-        results = []
-
         tracks = self._get_collection('/search/tracks',
             collection_id='Query "{0}"'.format(query),
             q=query.encode('utf-8'))
 
-        for track in itertools.islice(tracks, n):
-            uri = track['uri']
-            title = track['title']
-            results.append(self.url_result(url=uri))
+        results = [self.url_result(url=track['uri'])
+            for track in itertools.islice(tracks, n)]
 
         if not results:
             raise ExtractorError(
--- a/youtube_dl/extractor/soundcloud.py
+++ b/youtube_dl/extractor/soundcloud.py
@@ -488,15 +488,12 @@ class SoundcloudSearchIE(SearchInfoExtractor, SoundcloudIE):
 
     _SEARCH_KEY = 'scsearch'
     _RESULTS_PER_PAGE = 50
+    _API_V2_BASE = 'https://api-v2.soundcloud.com'
 
     def _get_collection(self, endpoint, collection_id, **query):
-        import itertools
-
         query['limit'] = self._RESULTS_PER_PAGE
         query['client_id'] = self._CLIENT_ID
         query['linked_partitioning'] = '1'
-
-        api_base_url = '{0}//api-v2.soundcloud.com'.format(self.http_scheme())
 
         total_results = self._MAX_RESULTS
         collected_results = 0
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/soundcloud.py
```diff
@@ -469,3 +472,86 @@ def _real_extract(self, url):
...
+                break
+
+            next_url = response.get(u'next_href', None)
+
+    def _get_n_results(self, query, n):
+
+        results = []
+
+        tracks = self._get_collection('/search/tracks',
+            collection_id='Query "{}"'.format(query),
```
dstftw: `{}` doesn't work in python 2.6.

youtube_dl/extractor/soundcloud.py
```diff
@@ -469,3 +472,86 @@ def _real_extract(self, url):
...
+            q=query.encode('utf-8'))
+
+        for _ in range(n):
+            try:
+                track = next(tracks)
+            except StopIteration:
+                break
+            uri = track[u'uri']
```
dstftw: Don't use `u`.

youtube_dl/extractor/soundcloud.py
```diff
@@ -469,3 +472,86 @@ def _real_extract(self, url):
...
+            title = track[u'title']
+            username = track[u'user'][u'username']
```
dstftw: It shouldn't fail if `user` or `username` is missing.

youtube_dl/extractor/soundcloud.py
```diff
@@ -469,3 +472,86 @@ def _real_extract(self, url):
...
+    _RESULTS_PER_PAGE = 50
+
+    def _get_collection(self, endpoint, collection_id, **query):
+        import itertools
+
+        query['limit'] = self._RESULTS_PER_PAGE
+        query['client_id'] = self._CLIENT_ID
+        query['linked_partitioning'] = '1'
+
+        api_base_url = '{0}//api-v2.soundcloud.com'.format(self.http_scheme())
```
dstftw: This makes no sense since it redirected to https anyway.

youtube_dl/extractor/soundcloud.py
```diff
@@ -469,3 +472,86 @@ def _real_extract(self, url):
...
+            'title': 'post-avant jazzcore',
+        },
+        'playlist_count': 15,
+    }]
+
+    _SEARCH_KEY = 'scsearch'
+    _RESULTS_PER_PAGE = 50
+
+    def _get_collection(self, endpoint, collection_id, **query):
+        import itertools
```
jaimeMF: This is already imported and (in general) you should only use `import`s at the top level.

youtube_dl/extractor/soundcloud.py
```diff
@@ -469,3 +472,86 @@ def _real_extract(self, url):
...
+
+    def _get_n_results(self, query, n):
+
+        results = []
+
+        tracks = self._get_collection('/search/tracks',
+            collection_id='Query "{}"'.format(query),
+            q=query.encode('utf-8'))
+
+        for _ in range(n):
```
jaimeMF: Is there some reason why you can't just use `for track in tracks:` here?

reiv: `_get_collection()` is a generator that loads lazily from a paginated source. Writing `for track in tracks:` would load all results, but we want to stop at `n`. This way we avoid making unnecessary requests.

youtube_dl/extractor/soundcloud.py
reiv: Fixed.

youtube_dl/extractor/soundcloud.py
reiv: Turns out it wasn't necessary anyway.

youtube_dl/extractor/soundcloud.py
reiv: @dstftw: Okay, removed.

youtube_dl/extractor/soundcloud.py
reiv: Fixed.

youtube_dl/extractor/soundcloud.py
jaimeMF: I missed that, then you should either use [itertools.islice](https://docs.python.org/2/library/itertools.html#itertools.islice) or
```python
for i, track in enumerate(tracks):
    if i >= n:
        break
```

reiv: Thanks, I rewrote it with `itertools.islice`.

youtube_dl/extractor/soundcloud.py
```diff
@@ -469,3 +472,70 @@ def _real_extract(self, url):
...
+        tracks = self._get_collection('/search/tracks',
+            collection_id='Query "{0}"'.format(query),
+            q=query.encode('utf-8'))
+
+        results = [self.url_result(url=track['uri'])
+            for track in itertools.islice(tracks, n)]
+
+        if not results:
+            raise ExtractorError(
+                '[soundcloud] No track results', expected=True)
```
yan12125: Usually we use syntaxes like 'soundcloud said: ' but not brackets in error messages.

reiv: Changed.

youtube_dl/extractor/soundcloud.py
```diff
@@ -469,3 +472,70 @@ def _real_extract(self, url):
             'description': data.get('description'),
             'entries': entries,
         }
+
+
+class SoundcloudSearchIE(SearchInfoExtractor, SoundcloudIE):
+    IE_NAME = 'soundcloud:search'
+    IE_DESC = 'Soundcloud search'
+    _MAX_RESULTS = 200
```
dstftw: Any particular reason for limiting this at all?

reiv: @dstftw: it's in the API docs. 200 is the maximum number of items returned in a collection. See https://developers.soundcloud.com/docs#pagination

dstftw: >The number of items in the collection returned is limited to 50 by default with a maximum value of 200

Collection is a single page thus 200 is the limit for a single page. So, `_RESULTS_PER_PAGE` should be set to 200 and `_MAX_RESULTS` to `float('inf')`.

reiv: I took `limit` to mean the limit on the number of results altogether, but your interpretation makes more sense.

dstftw:
Cherry-picked. Thanks.

dstftw:

reiv:
Cheers! You can use my github username.

PR #7218, opened 2015-10-18
[NPR] New extractor for NPR.org mediaplayer

Kaspi:


youtube_dl/extractor/npr.py
```diff
@@ -0,0 +1,71 @@
...
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('id')
+        webpage_url = 'http://www.npr.org/player/v2/mediaPlayer.html?id=' + video_id
+        webpage = self._download_webpage(webpage_url, video_id)
+        key = 'MDAzMzQ2MjAyMDEyMzk4MTU1MDg3ZmM3MQ010'
+        xml_url = 'http://api.npr.org/query?id=%s&apiKey=%s' % (video_id, key)
+
+        config = self._download_xml(xml_url,video_id, note='Downloading XML')
+
+        audio = config.findall('./list/story/audio[@type="standard"]')
```
dstftw: This syntax is not supported in python 2.6. You should use `xpath_*` from `utils`.

youtube_dl/extractor/npr.py
```diff
@@ -0,0 +1,71 @@
...
+        'id': '445367719',
+        'ext': 'mp4',
+        'title': 'VEGA INTL. Night School'
+    }
+}
+
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('id')
```
dstftw: Use `self._matching_id`.

youtube_dl/extractor/npr.py
```diff
@@ -0,0 +1,71 @@
...
+
+            for r in regex:
+                t = song.find(r)
+                if t is not None:
+                    format.append({'format': t.get('type', t.tag),
+                               'url' : t.text})
+
+            entries.append({ "title":song_title,
+                             "id":song_id,
+                             "duration": str(int(song_duration) / 60) +":"+ str(int(song_duration) % 60) ,
```
dstftw: duration must be `int`.

youtube_dl/extractor/npr.py
```diff
@@ -0,0 +1,71 @@
...
+                             "formats":format})
```
dstftw: Avoid shadowing built-in names.

youtube_dl/extractor/npr.py
```diff
@@ -0,0 +1,71 @@
...
+}
+
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('id')
+        webpage_url = 'http://www.npr.org/player/v2/mediaPlayer.html?id=' + video_id
+        webpage = self._download_webpage(webpage_url, video_id)
+        key = 'MDAzMzQ2MjAyMDEyMzk4MTU1MDg3ZmM3MQ010'
+        xml_url = 'http://api.npr.org/query?id=%s&apiKey=%s' % (video_id, key)
```
remitamine: you can get json output by appending `&format=json` to the api request url.

PR #7225, opened 2015-10-18
adds thumbnail support for ZDF Mediathek extractor

kennell:
This commit adds thumbnail support to the ZDF Mediathek extractor

### DIFF
--- a/youtube_dl/extractor/zdf.py
+++ b/youtube_dl/extractor/zdf.py
@@ -75,9 +75,10 @@ def extract_from_xml_url(ie, video_id, xml_url):
         for node in fnode:
             thumbnail = {'url': node.text}
             if 'key' in node.attrib:
-                if re.match("^[0-9]+x[0-9]+$", node.attrib['key']):
-                    thumbnail['width'] = int_or_none(node.attrib['key'].split('x')[0])
-                    thumbnail['height'] = int_or_none(node.attrib['key'].split('x')[1])
+                m = re.match('^([0-9]+)x([0-9]+)$', node.attrib['key'])
+                if m:
+                    thumbnail['width'] = int(m.group(1))
+                    thumbnail['height'] = int(m.group(2))
             thumbnails.append(thumbnail)
         return thumbnails
 
--- a/youtube_dl/extractor/zdf.py
+++ b/youtube_dl/extractor/zdf.py
@@ -70,6 +70,23 @@ def extract_from_xml_url(ie, video_id, xml_url):
             '_available': is_available,
         }
 
+    def xml_to_thumbnails(fnode):
+        thumbnails = list()
+        for node in fnode:
+            width_x_height = node.attrib['key']
+            thumbnail = {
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/zdf.py
```diff
@@ -81,6 +98,8 @@ def xml_to_format(fnode):
         'title': title,
         'description': description,
         'duration': duration,
+        'thumbnail': thumbnail,
```
dstftw: You may not specify this since you provide `thumbnails`. It will be autocalculated.

youtube_dl/extractor/zdf.py
```diff
@@ -70,6 +70,23 @@ def xml_to_format(fnode):
             '_available': is_available,
         }
 
+    def xml_to_thumbnails(fnode):
+        thumbnails = list()
+        for node in fnode:
+            width_x_height = node.attrib['key']
```
dstftw: Extraction should not fail due to changes in XML layout. If `key` is removed the whole extraction will be broken.

youtube_dl/extractor/zdf.py
```diff
@@ -70,6 +70,23 @@ def xml_to_format(fnode):
...
+            thumbnail = {
+                'url': node.text,
+                'width': int(width_x_height.split('x')[0]),
```
dstftw: Same here. And use `utils.int_or_none` instead.

kennell:
HI @dstftw - i addressed these issues in latest commit.

youtube_dl/extractor/zdf.py
```diff
@@ -70,6 +70,21 @@ def xml_to_format(fnode):
...
         }
 
+    def xml_to_thumbnails(fnode):
+        thumbnails = list()
+        for node in fnode:
+            thumbnail = {'url': node.text}
+            if 'key' in node.attrib:
+                width_x_height = node.attrib['key']
+                thumbnail['width'] = int_or_none(width_x_height.split('x')[0])
+                thumbnail['height'] = int_or_none(width_x_height.split('x')[1])
```
dstftw: This will fail if `width_x_height` will not contain `x`.

kennell:
Adressed @dstftw

youtube_dl/extractor/zdf.py
```diff
@@ -70,6 +70,21 @@ def xml_to_format(fnode):
             '_available': is_available,
         }
 
+    def xml_to_thumbnails(fnode):
+        thumbnails = list()
+        for node in fnode:
+            thumbnail = {'url': node.text}
+            if 'key' in node.attrib:
+                if re.match("^[0-9]+x[0-9]+$", node.attrib['key']):
```
dstftw: Since you use regex anyway you can just capture both numbers and use plain `int`. Also consistently use single quotes.

kennell:
@dstftw - again, fixed  :)

dstftw:
Thanks.

Issue #7162, opened 2015-10-13
Site Request Stitcher


canamcy:
http://www.stitcher.com/podcast/the-talking-machines/e/40789481?autoplay=true

canamcy:
[generic] 40789481?autoplay=true: Requesting header
[generic] 40789481?autoplay=true: Downloading webpage
[generic] 40789481?autoplay=true: Extracting information

[generic] 40789481?autoplay=true: Requesting header
[generic] 40789481?autoplay=true: Downloading webpage
[generic] 40789481?autoplay=true: Extracting information
[Error] ytdl for => [debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-f', u'worst', u'-i', u'--verbose', u'--restrict-f
ilenames', u'--youtube-skip-dash-manifest', u'-o', u'..\\..\\..\\yo0\\M\\yodl\\y
outube-dl\\%(extractor)s\\%(title)s-%(id)s.%(ext)s', u'--', u'http://www.stitche
r.com/podcast/the-talking-machines/e/40789481?autoplay=true']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2015.10.13
[debug] Python version 2.7.8 - Windows-7-6.1.7601-SP1
[debug] exe versions: ffmpeg N-56800-gad8fbdd, ffprobe N-56800-gad8fbdd
[debug] Proxy map: {}
WARNING: Falling back on generic information extractor.
ERROR: Unsupported URL: http://www.stitcher.com/podcast/the-talking-machines/e/4
0789481?autoplay=true
Traceback (most recent call last):
  File "youtube_dl\extractor\generic.pyo", line 1240, in _real_extract
  File "youtube_dl\utils.pyo", line 1656, in parse_xml
  File "xml\etree\ElementTree.pyo", line 1300, in XML
  File "xml\etree\ElementTree.pyo", line 1642, in feed
  File "xml\etree\ElementTree.pyo", line 1506, in _raiseerror
ParseError: not well-formed (invalid token): line 4, column 407
Traceback (most recent call last):
  File "youtube_dl\YoutubeDL.pyo", line 660, in extract_info
  File "youtube_dl\extractor\common.pyo", line 290, in extract
  File "youtube_dl\extractor\generic.pyo", line 1838, in _real_extract
UnsupportedError: Unsupported URL: http://www.stitcher.com/podcast/the-talking-m
[...]

PR #7228, opened 2015-10-19
Stitcher Add new extractor (Issue: #7162)

mjdubell:
New extractor

### DIFF
--- a/youtube_dl/extractor/stitcher.py
+++ b/youtube_dl/extractor/stitcher.py
@@ -1,0 +1,37 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class StitcherIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?stitcher\.com/podcast/[\/a-z\-]+\d+\?.+'
+    _TEST = {
+        'url': 'http://www.stitcher.com/podcast/the-talking-machines/e/40789481?autoplay=true',
+        'md5': '391dd4e021e6edeb7b8e68fbf2e9e940',
+        'info_dict': {
+            'id': '40789481',
+            'ext': 'mp3',
+            'title': 'Machine Learning Mastery and Cancer Clusters from Talking Machines',
+        }
+    }
+
+    def _real_extract(self, url):
+        audio_id = self._search_regex(r'[a-z\/\-\:\/\/.]+?(\d+?)\?.+', url, "audio_id")
+
+        webpage = self._download_webpage(url, audio_id)
+
+        title = self._og_search_title(webpage)
+        url = self._search_regex(r'[\s\S]*episodeURL: "(.+?)"[\s\S]*', webpage, 'url')
+        episode_image = self._search_regex(r'[\s\S]*episodeImage: "(.+?)"[\s\S]*', webpage, 'thumbnail')
+        duration = int(self._search_regex(r'[\s\S]*duration: (\d+?),[\s\S]*', webpage, 'duration')) / 60
+
+        return {
+            'id': audio_id,
+            'url': url,
+            'title': title,
+            'duration': duration,
+            'thumbnail': episode_image,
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/stitcher.py
```diff
@@ -0,0 +1,37 @@
...
+        'md5': '391dd4e021e6edeb7b8e68fbf2e9e940',
+        'info_dict': {
+            'id': '40789481',
+            'ext': 'mp3',
+            'title': 'Machine Learning Mastery and Cancer Clusters from Talking Machines',
+        }
+    }
+
+    def _real_extract(self, url):
+        audio_id = self._search_regex(r'[a-z\/\-\:\/\/.]+?(\d+?)\?.+', url, "audio_id")
```
jaimeMF: The method we follow here is to use the `_VALID_URL`, which you would define as:
```python
_VALID_URL = r'https?://(?:www\.)?stitcher\.com/podcast/[\/a-z\-]+(?P<id>\d+)\?.+'
```
And then you only need to do:
```python
audio_id = self._match_id(url)
```

youtube_dl/extractor/stitcher.py
```diff
@@ -0,0 +1,37 @@
...
+
+        webpage = self._download_webpage(url, audio_id)
+
+        title = self._og_search_title(webpage)
+        url = self._search_regex(r'[\s\S]*episodeURL: "(.+?)"[\s\S]*', webpage, 'url')
```
dstftw: `[\s\S]*` on both sides are noop due to `*`.

youtube_dl/extractor/stitcher.py
```diff
@@ -0,0 +1,37 @@
...
+        episode_image = self._search_regex(r'[\s\S]*episodeImage: "(.+?)"[\s\S]*', webpage, 'thumbnail')
```
dstftw: Should be `fatal=False`.

youtube_dl/extractor/stitcher.py
```diff
@@ -0,0 +1,37 @@
...
+        duration = int(self._search_regex(r'[\s\S]*duration: (\d+?),[\s\S]*', webpage, 'duration')) / 60
```
dstftw: Should be `fatal=False`. Use `utils.int_or_none`. It must be in seconds.

mjdubell:
Updated the code

youtube_dl/extractor/stitcher.py
```diff
@@ -0,0 +1,36 @@
+# coding: utf-8
+from __future__ import unicode_literals
+from .common import InfoExtractor
+
+
+class StitcherIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?stitcher\.com/podcast/[\/a-z\-]+(?P<id>\d+)|\?[a-z=]+'
```
dstftw: What are you trying to achieve with `|\?[a-z=]+`?

youtube_dl/extractor/stitcher.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def _real_extract(self, url):
+        audio_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, audio_id)
+
+        title = self._og_search_title(webpage)
+        url = self._search_regex(r'episodeURL: "(.+?)"', webpage, 'url')
+        episode_image = self._search_regex(r'episodeImage: "(.+?)"', webpage, 'episode_image', fatal=False)
+        duration = self._search_regex(r'simpleDuration: "(\d+?) minutes"', webpage, 'duration', fatal=False)
```
dstftw: As said duration must be int.

youtube_dl/extractor/stitcher.py
mjdubell: I was trying to match:
```
?autoplay=true
```
But I see now that isn't necessary.

mjdubell:
Updated code

PR #7233, opened 2015-10-19
[Oddshot] Add new extractor

aaearon:
Proposed extractor for http://oddshot.tv/


dstftw:
What's the point? It's downloaded with generic extractor just fine.

PR #7237, opened 2015-10-19
[vidme] Stream URL fallback, better error message for suspended videos

lalinsky:
It seem that vidme does not return "formats" in the API response in some cases, but we can still use the "complete_url" field from there. See e.g. https://vid.me/e5g


dstftw:
Thanks.

PR #7240, opened 2015-10-20
[ImoocVideoIE] Add new extractor

bumaociyuan:
[ImoocVideoIE] Add new extractor

### DIFF
--- a/youtube_dl/extractor/imooc.py
+++ b/youtube_dl/extractor/imooc.py
@@ -1,0 +1,49 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from ..compat import (
+    compat_urllib_parse_unquote,
+    compat_urllib_request,
+)
+
+class ImoocVideoIE(InfoExtractor):
+    _VALID_URL = r'http://www.imooc.com/video/(?P<id>[0-9]+)'
+    _TEST = {
+        'url': 'http://www.imooc.com/video/5454',
+        'md5': '03a0f36327721551fce08776fe8f70f1',
+        'info_dict': {
+            'id': '5454',
+            'ext': 'mp4',
+            'title': '3-1 网络环境查看命令',
+        }
+    }
+
+    # _ANDROID_USER_AGENT = 'Mozilla/5.0 (iPhone; U; CPU iPhone OS 4_3_3 like Mac OS X; en-us) AppleWebKit/533.17.9 (KHTML, like Gecko) Version/5.0.2 Mobile/8J2 Safari/6533.18.5'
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/imooc.py
```diff
@@ -0,0 +1,48 @@
...
+    }
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        json_url = 'http://www.imooc.com/course/ajaxmediainfo/?mid=%s&mode=flash' % video_id
+        data = self._download_json(json_url, video_id, 'Downloading video formats')
+
+        url = data['data']['result']['mpath'][0]
```
remitamine: it better to extract all the urls in the `mpath` array.

youtube_dl/extractor/imooc.py
```diff
@@ -0,0 +1,48 @@
...
+            'title': '网络环境查看命令',
+        }
+    }
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        json_url = 'http://www.imooc.com/course/ajaxmediainfo/?mid=%s&mode=flash' % video_id
+        data = self._download_json(json_url, video_id, 'Downloading video formats')
```
remitamine: it's better to check if `data['result'] == 0` if not than raise an ExtractorError with `data['msg']`.

docs/supportedsites.md
```diff
@@ -220,6 +220,7 @@
  - **imdb:list**: Internet Movie Database lists
  - **Imgur**
  - **ImgurAlbum**
+ - **Imooc**: 慕课网
```

youtube_dl/extractor/imooc.py
```diff
@@ -0,0 +1,67 @@
...
+
+        if data['result'] == 0:
+            urls = data['data']['result']['mpath']
+            title = data['data']['result']['name']
+
+            for i, url in enumerate(urls):
+                req = HEADRequest(url)
+                res = self._request_webpage(
+                    req, video_id, note='Testing video URL %d' % i, errnote=False)
```
remitamine: i didn't mean to check the urls, i mean to extract the `mpath` urls by putting them in a `formats` array and return them with id and title.

youtube_dl/extractor/imooc.py
```diff
@@ -0,0 +1,67 @@
...
+                if res:
+                    break
+            else:
+                raise ExtractorError('No working video URLs found')
+
+        else:
+            print data['msg']
+            raise ValueError(data['msg'])
```
yan12125: The previous ```print``` looks like a debug statement. Please remove it. And, ```ExtractorError``` (with ```expected=True```) is better than ```ValueError``` here.

yan12125:

PR #7242, opened 2015-10-20
[vimeo] Fix error parsing

lalinsky:
Example video that fails - https://vimeo.com/109815029


dstftw:
Can you add a test?

lalinsky:
Is there some way to add a test that checks for the expected error message? I wanted to add a test from the beginning, but I couldn't find anything. I could add a test with `only_matching`, which would be only mostly useful for manual testing, but not actually test the behavior.

Or do I add something like `expected_errors` (similar to `expected_warnings`) to `test_download.py`?

dstftw:
only_matching is enough for now. It's just for reference.

lalinsky:
Done.

dstftw:
Thanks.

PR #7255, opened 2015-10-21
fix regex in utils to avoid breaking json structure

hynek-jemelik:
Stripping out html tags not done correctly. It looks for longest regex, which can span multiple tags for even multiple json fields, which in some cases can break the whole json structure and make it unparsable.

For example in https://github.com/rg3/youtube-dl/issues/7254


jaimeMF:
I haven't yet looked into your original issue. But your change is problematic, because currently:
```python
>>> clean_html('<div class="foo">Hello</div>')
'Hello'
```
but with your change:
```python
>>>  clean_html('<div class="foo">Hello</div>')
'<div class="foo">Hello'
```
I'm not 100% sure, but we probably rely on the current behaviour in some extractor.

Issue #7193, opened 2015-10-16
Tightening up CI Integration 


devoxel:
I think the current test architecture needs a re-examination.

I've identified a couple of issues:
# Verbose Logs

Logs are verbose. 

``` sh
$ curl https://s3.amazonaws.com/archive.travis-ci.org/jobs/85403103/log.txt | wc -l
15385
```

Ouch. Take a step back: that's literally unreadable. 
# Tests are slow

Take a look at the [travis-ci build history](https://travis-ci.org/rg3/youtube-dl/builds)

Every build runs for about 5hrs and runs in a single command. Compare that to the [ ruby on rails](https://travis-ci.org/rails/rails/builds/85624063) setup. Builds are separated based on what's being tested and things are allowed to fail (namely some tests with new versions of ruby)

---

I suggest fixing these by making these changes:
- Separating tests into canonical pieces
- Allowing some tests to fail, be it because of network errors, 404s or just non-core functionality

I'm going to start on this stuff on a fork, but I'd like community input too so do shout at me.

PR #7267, opened 2015-10-22
Parallel nosetests

yan12125:
With this patch, there will be no more congestion in Travis CI tests. Tries to solve #5289 and #7193.

Implementation considerations:
1. 4 processors: just my intuition. Larger values are fine.
2. 540-second timeout: Travis CI sets a 10-minute limit for output, I use 9 minutes.
3. `_multiprocess_shared_`: If I'm correct, `setUp` function of `TestDownload` is not reentrant. Is it?
4. Print test names in case of network errors. Just for easier debugging of such errors.

See https://travis-ci.org/yan12125/youtube-dl/builds/86834580 for the result.

### DIFF
--- a/devscripts/run_tests.sh
+++ b/devscripts/run_tests.sh
@@ -16,4 +16,4 @@ case "$YTDL_TEST_SET" in
     ;;
 esac
 
-nosetests test --verbose $test_set
+nosetests test --verbose $test_set --processes=4 --process-timeout=540
--- a/devscripts/run_tests.sh
+++ b/devscripts/run_tests.sh
@@ -3,6 +3,7 @@
 DOWNLOAD_TESTS="age_restriction|download|subtitles|write_annotations|iqiyi_sdk_interpreter"
 
 test_set=""
+multiprocess_args=""
 
 case "$YTDL_TEST_SET" in
     core)
@@ -10,10 +11,11 @@ case "$YTDL_TEST_SET" in
     ;;
     download)
         test_set="-I test_(?!$DOWNLOAD_TESTS).+\.py"
+        multiprocess_args="--processes=4 --process-timeout=540"
     ;;
     *)
         break
     ;;
 esac
 
-nosetests test --verbose $test_set --processes=4 --process-timeout=540
+nosetests test --verbose $test_set $multiprocess_args
--- a/test/test_download.py
+++ b/test/test_download.py
@@ -75,7 +75,7 @@ class TestDownload(unittest.TestCase):
 # Dynamically generate tests
 
 
-def generator(test_case):
+def generator(test_case, tname):
 
     def test_template(self):
         ie = youtube_dl.extractor.get_info_extractor(test_case['name'])
@@ -148,7 +148,7 @@ def generator(test_case):
                         raise
 
                     if try_num == RETRIES:
-                        report_warning('Failed due to network errors, skipping...')
+                        report_warning('%s failed due to network errors, skipping...' % tname)
                         return
 
                     print('Retrying: {0} failed tries\n\n##########\n\n'.format(try_num))
@@ -223,12 +223,12 @@ def generator(test_case):
 
 # And add them to TestDownload
 for n, test_case in enumerate(defs):
-    test_method = generator(test_case)
     tname = 'test_' + str(test_case['name'])
     i = 1
     while hasattr(TestDownload, tname):
         tname = 'test_%s_%d' % (test_case['name'], i)
         i += 1
+    test_method = generator(test_case, tname)
     test_method.__name__ = str(tname)
     setattr(TestDownload, test_method.__name__, test_method)
     del test_method
--- a/test/test_download.py
+++ b/test/test_download.py
@@ -65,6 +65,8 @@ defs = gettestcases()
 
 
 class TestDownload(unittest.TestCase):
+    _multiprocess_shared_ = True
+
     maxDiff = None
 
     def setUp(self):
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -7,7 +7,7 @@ python:
   - "3.4"
   - "3.5"
 sudo: false
-script: nosetests test --verbose
+script: nosetests test --verbose --processes=4 --process-timeout=540
```
jaimeMF: You can pass `--processes=-1` ([docs](https://nose.readthedocs.org/en/latest/man.html#cmdoption--processes)) and it will automatically select  the number based on the number of cores. Maybe that's better?

jaimeMF:
> 3. _multiprocess_shared_: If I'm correct, setUp function of TestDownload is not reentrant. Is it?

Actually it seems that we can remove it, `self.defs`  isn't used anymore.

test/test_download.py
```diff
@@ -64,6 +64,8 @@ def _file_md5(fn):
 
 
 class TestDownload(unittest.TestCase):
+    _multiprocess_shared_ = True
```
jaimeMF: I think there should be a link to https://nose.readthedocs.org/en/latest/doc_tests/test_multiprocess/multiprocess.html, to make it clear that this is not from `unittest.TestCase` (that was where I first looked for its documentation).

jaimeMF:
On [test_compat_expanduser](https://github.com/rg3/youtube-dl/blob/3513d4143657c1bdc7da72f048a63d4b1ee794d3/test/test_compat.py#L31) we temporary modify the `HOME` environment variable, could it interfere with some test?

.travis.yml
jaimeMF: I have seen you have try how many cores would use in Travis (https://travis-ci.org/yan12125/travis-ci-test/builds/87175866#L87), 16 is probably too much. It seems fine to use 4.

jaimeMF:
> On test_compat_expanduser we temporary modify the HOME environment variable, could it interfere with some test?

I have run a simple test with multiprocessing` and the different processes don't share `os.environ`. So it shouldn't be a problem.



Since the other two remarks are made are minor, I think we can merge it in the current state (unless you want to add/modify something).

yan12125:
Sorry I don't have much time on this work these days. Here's another concern: info JSON files may have the same names. For example:
```
class FooIE(InfoExtractor):
    _TEST = {
        'url': 'http://foo.com/video/1234'
        'info_dict': {
            'id': '1234',
            'ext': 'mp4',
            'title': 'youtube-dl test video',
        }
    }
class BarIE(InfoExtractor):
    _TEST = {
        'url': 'http://bar.com/video/1234'
        'info_dict': {
            'id': '1234',
            'ext': 'mp4',
            'title': 'youtube-dl test video',
        }
    }
```
Then both tests write ```youtube-dl test video-1234.info.json```. A possible solution is prepending IE names in info JSON filenames. However, I guess lots of applications using youtube-dl rely on info JSON files. Doing this will break them. Or we can use a different [info JSON filename](https://github.com/rg3/youtube-dl/blob/8a06999ba0f9c948f8d2a1ef89c73eedbfb09cfb/youtube_dl/YoutubeDL.py#L1519) when testing?

jaimeMF:
We can just prepend `tname` (which also covers the case when multiple test from a IE reuse the video, I don't know if we actually have any of this):
```diff
diff --git a/test/test_download.py b/test/test_download.py
index f7f7423..e8ccdeb 100644
--- a/test/test_download.py
+++ b/test/test_download.py
@@ -103,6 +103,7 @@ def generator(test_case, tname):
                 return
 
         params = get_params(test_case.get('params', {}))
+        params['outtmpl'] = tname + '_' + params['outtmpl']
         if is_playlist and 'playlist' not in test_case:
             params.setdefault('extract_flat', True)
             params.setdefault('skip_download', True)
```

yan12125:
Thanks @jaimeMF. It's a good idea.

yan12125:
Now all tests run more than 50 minutes and interrupted by Travis. I believe this change is ready; I'll squash and merge it in a few days if no objection.

A recent build: https://travis-ci.org/yan12125/youtube-dl/builds/200858625

dstftw:
I would prefer to separate builds for core and download tests first, something similar to https://github.com/dstftw/youtube-dl/commits/travis-separate-core-and-extractors.

yan12125:
That's also the way to go. Is there something missing in your branch before a pull request?

dstftw:
Well, I'm not sure, but it seems to work. Do you want me to open a PR or will you just borrow these codes into this PR?

yan12125:
The two changes come from different ideas, so a different PR is better.

dstftw:
Done in #12100.

yan12125:
Rebased after #12100 merged

devscripts/run_tests.sh
```diff
@@ -16,4 +16,4 @@ case "$YTDL_TEST_SET" in
     ;;
 esac
 
-nosetests test --verbose $test_set
+nosetests test --verbose $test_set --processes=4 --process-timeout=540
```
dstftw: Basically, there is a little sense running core test parallely.

dstftw:
LGTM.

yan12125:
Seems it's running fine. Good to go now?

dstftw:
Feel free to merge.

yan12125:
Done in 8936f68a0ba3284c88ec619fb4cc22eb0499e7f3. Thanks for previous work on spliting tests and reviewing on this!

PR #7274, opened 2015-10-23
[abc] add support for audio extraction

remitamine:


youtube_dl/extractor/abc.py
```diff
@@ -12,7 +12,7 @@
 
 class ABCIE(InfoExtractor):
     IE_NAME = 'abc.net.au'
-    _VALID_URL = r'http://www\.abc\.net\.au/news/[^/]+/[^/]+/(?P<id>\d+)'
+    _VALID_URL = r'http://www\.abc\.net\.au/news/(?:[^/]+/){1,2}(?P<id>\d+)'
```
jaimeMF: It's not clear in the commit why you are changing this, if you need it you should add an example url in the tests.

youtube_dl/extractor/abc.py
```diff
@@ -58,13 +67,22 @@ def _real_extract(self, url):
...
                 self.url_result(url_info['url']) for url_info in urls_info])
 
-        formats = [{
-            'url': url_info['url'],
-            'width': int_or_none(url_info.get('width')),
-            'height': int_or_none(url_info.get('height')),
-            'tbr': int_or_none(url_info.get('bitrate')),
-            'filesize': int_or_none(url_info.get('filesize')),
-        } for url_info in urls_info]
+        if mobj.group('type') == 'Video':
```
jaimeMF: You could just do:
```python
        is_video = mobj.group('type') == 'Video'
        formats = [{
            'url': url_info['url'],
            'vcodec': url_info.get('codec') if is_video else 'none',
            'width': int_or_none(url_info.get('width')),
            'height': int_or_none(url_info.get('height')),
            'tbr': int_or_none(url_info.get('bitrate')),
            'filesize': int_or_none(url_info.get('filesize')),
        } for url_info in urls_info]
```

remitamine:
i added the test for the case that require the change in the _VALID_REGEX and simplify the extraction.

jaimeMF:
Cherry picked in 7687b354c59efea076fae762206c00de273fbe04 and d97da29da26c920ae31fde94bba5e3b4e1f5a36a, thanks.
For future PR, use separate commits for independent changes (adding audio support and improving the _VALID_URL in this case), keep in in mind that joining commits is quite easy with `git rebase --interactive` while splitting in different commits requires more work.

PR #7282, opened 2015-10-24
[spiegeltv] Fix Accept-Encoding issue (server chokes on gzip)

cdoepmann:
The content server's responses are incomplete when the Accept-Encoding header includes `gzip`, so I removed it (Accept-Encoding in std_headers is normally set to `gzip,deflate`). As an alternative, one could set it to `*`. Seems to work fine, too.

The whole issue seems to be a bug on the server side, which we need to cope with.

Example video that fails: http://www.spiegel.tv/filme/zdfe-ice-pilots-nachwuchssorgen/


dstftw:
Cherry-picked in 0198807ef95338bec69cfdcd67249c007e4d4141. Thanks.

Issue #7178, opened 2015-10-14
"TypeError: must be unicode, not str" when using --write-description


sebma:
Hi, I have the following error :

"TypeError: must be unicode, not str"

The command I typed is the following :

youtube-dl --verbose --ignore-config --write-description http://www.veoh.com/watch/v46093745wbEGkakh
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--verbose', u'--ignore-config', u'--write-description', u'http://www.veoh.com/watch/v46093745wbEGkakh']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.10.13
[debug] Python version 2.7.8 - Linux-3.16.0-44-generic-x86_64-with-Ubuntu-14.10-utopic
[debug] exe versions: avconv 11.2-6, avprobe 11.2-6, ffmpeg 2.6.2, ffprobe 2.6.2, rtmpdump 2.4
[debug] Proxy map: {}
[Veoh] v46093745wbEGkakh: Downloading video XML
[info] Writing video description to: Pirates Of Silicon Valley-46093745.description
Traceback (most recent call last):
File "/usr/local/bin/youtube-dl", line 11, in
sys.exit(main())
File "/usr/local/lib/python2.7/dist-packages/youtube_dl/init.py", line 410, in main
real_main(argv)
File "/usr/local/lib/python2.7/dist-packages/youtube_dl/__init_.py", line 400, in _real_main
retcode = ydl.download(all_urls)
File "/usr/local/lib/python2.7/dist-packages/youtube_dl/YoutubeDL.py", line 1665, in download
url, force_generic_extractor=self.params.get('force_generic_extractor', False))
File "/usr/local/lib/python2.7/dist-packages/youtube_dl/YoutubeDL.py", line 671, in extract_info
return self.process_ie_result(ie_result, download, extra_info)
File "/usr/local/lib/python2.7/dist-packages/youtube_dl/YoutubeDL.py", line 717, in process_ie_result
return self.process_video_result(ie_result, download=download)
File "/usr/local/lib/python2.7/dist-packages/youtube_dl/YoutubeDL.py", line 1335, in process_video_result
self.process_info(new_info)
File "/usr/local/lib/python2.7/dist-packages/youtube_dl/YoutubeDL.py", line 1465, in process_info
descfile.write(info_dict['description'])
TypeError: must be unicode, not str

Any idea ?

Seb.

[...]

PR #7296, opened 2015-10-25
Use a wrapper around xml.etree.ElementTree.fromstring in python 2.x (…

jaimeMF:
…#7178)

Attributes aren't unicode objects, so they couldn't be directly used in info_dict fields (for example '--write-description' doesn't work with bytes).

### DIFF
--- a/youtube_dl/compat.py
+++ b/youtube_dl/compat.py
@@ -216,8 +216,7 @@ except ImportError:  # Python 2.6
 if sys.version_info[0] >= 3:
     compat_etree_fromstring = xml.etree.ElementTree.fromstring
 else:
-    # on python 2.x the the attributes of a node are str objects instead of
-    # unicode
+    # on python 2.x the the attributes of a node aren't always unicode objects
     etree = xml.etree.ElementTree
 
     # on 2.6 XML doesn't have a parser argument, function copied from CPython
@@ -231,7 +230,8 @@ else:
     def _element_factory(*args, **kwargs):
         el = etree.Element(*args, **kwargs)
         for k, v in el.items():
-            el.set(k, v.decode('utf-8'))
+            if isinstance(v, bytes):
+                el.set(k, v.decode('utf-8'))
         return el
 
     def compat_etree_fromstring(text):
--- a/youtube_dl/compat.py
+++ b/youtube_dl/compat.py
@@ -216,8 +216,18 @@ except ImportError:  # Python 2.6
 if sys.version_info[0] >= 3:
     compat_etree_fromstring = xml.etree.ElementTree.fromstring
 else:
-    # on python 2.x the the attributes of a node aren't always unicode objects
+    # on python 2.x the attributes and text of a node aren't always unicode
+    # objects
     etree = xml.etree.ElementTree
+
+    try:
+        _etree_iter = etree.Element.iter
+    except AttributeError:  # Python <=2.6
+        def _etree_iter(root):
+            for el in root.findall('*'):
+                yield el
+                for sub in _etree_iter(el):
+                    yield sub
 
     # on 2.6 XML doesn't have a parser argument, function copied from CPython
     # 2.7 source
@@ -235,7 +245,11 @@ else:
         return el
 
     def compat_etree_fromstring(text):
-        return _XML(text, parser=etree.XMLParser(target=etree.TreeBuilder(element_factory=_element_factory)))
+        doc = _XML(text, parser=etree.XMLParser(target=etree.TreeBuilder(element_factory=_element_factory)))
+        for el in _etree_iter(doc):
+            if el.text is not None and isinstance(el.text, bytes):
+                el.text = el.text.decode('utf-8')
+        return doc
 
 try:
     from urllib.parse import parse_qs as compat_parse_qs
--- a/youtube_dl/compat.py
+++ b/youtube_dl/compat.py
@@ -14,6 +14,7 @@ import socket
 import subprocess
 import sys
 import itertools
+import xml.etree.ElementTree
 
 
 try:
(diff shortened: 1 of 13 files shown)


yan12125:
Fails for Chinese contents:
```Diff
diff --git a/test/test_compat.py b/test/test_compat.py
index 2b08604..870005c 100644
--- a/test/test_compat.py
+++ b/test/test_compat.py
@@ -78,5 +78,9 @@ class TestCompat(unittest.TestCase):
         doc = compat_etree_fromstring(xml.encode('utf-8'))
         self.assertTrue(isinstance(doc.attrib['foo'], compat_str))
 
+        xml = '<el foo="中文"></el>'
+        doc = compat_etree_fromstring(xml.encode('utf-8'))
+        self.assertTrue(isinstance(doc.attrib['foo'], compat_str))
+
 if __name__ == '__main__':
     unittest.main()
```
And the result:
```
$ python2 test/test_compat.py
.E.....
ERROR: test_compat_etree_fromstring (__main__.TestCompat)
  File "test/test_compat.py", line 82, in test_compat_etree_fromstring
    doc = compat_etree_fromstring(xml.encode('utf-8'))
  File "/home/yen/Executables/Multimedia/youtube-dl/youtube_dl/compat.py", line 238, in compat_etree_fromstring
    return _XML(text, parser=etree.XMLParser(target=etree.TreeBuilder(element_factory=_element_factory)))
  File "/home/yen/Executables/Multimedia/youtube-dl/youtube_dl/compat.py", line 228, in _XML
    parser.feed(text)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1640, in feed
    self._parser.Parse(data, 0)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1543, in _start_list
    return self.target.start(tag, attrib)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1419, in start
    self._last = elem = self._factory(tag, attrs)
  File "/home/yen/Executables/Multimedia/youtube-dl/youtube_dl/compat.py", line 234, in _element_factory
    el.set(k, v.decode('utf-8'))
  File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)

Ran 7 tests in 0.002s

FAILED (errors=1)
```

jaimeMF:
Thanks for catching that, fixed in 387db16a789fea25795433538d80513c18d0f699 (I should have obviously tried with non ascii characters).

dstftw:
This does not solve the problem completely, `Element.text` is also `str`.

jaimeMF:
We also have [parse_xml in utils](https://github.com/rg3/youtube-dl/blob/d01949dc89feb2441f251e42e8a6bfa4711b9715/youtube_dl/utils.py#L1660-L1674). Since it's only required in python 2.x I'd merge it with compat_etree_fromstring`, but I don't know if defining the `doctype` methods is required because it should be optional (that's what I understand in the docs).

jaimeMF:
Now `Element.text` is also an unicode object (I've removed `utils.parse_xml`).

dstftw:
It's also good to fix [`unified_strdate`](https://github.com/rg3/youtube-dl/blob/0a192fbea798c843ad6fef37106901d431f39b6e/youtube_dl/utils.py#L899), [`extractor` field](https://github.com/rg3/youtube-dl/blob/0a192fbea798c843ad6fef37106901d431f39b6e/youtube_dl/YoutubeDL.py#L691) and [`extractor_key` field](https://github.com/rg3/youtube-dl/blob/0a192fbea798c843ad6fef37106901d431f39b6e/youtube_dl/YoutubeDL.py#L694).
And it's probably nice to have something similar to the following code snippet to track byte fields further at least in download tests:
```python
def check_dict(d):
    for key, value in d.items():
        check_item(key, value)

def check_item(key, item):
    if isinstance(item, bytes):
        self.report_warning(
            'Extractor %s returned a byte field "%s" in info_dict that should be replaced with unicode string.'
            % (info_dict.get('extractor'), key))
    elif isinstance(item, (list, tuple,)):
        for sub_item in item:
            check_item(key, sub_item)
    elif isinstance(item, dict):
        check_dict(item)

check_dict(info_dict)
```

jaimeMF:
I think it would be better to do it in separate PRs (I'd prefer if this PR doesn't get too big).

youtube_dl/compat.py
```diff
@@ -212,6 +213,43 @@ def data_open(self, req):
 except ImportError:  # Python 2.6
     from xml.parsers.expat import ExpatError as compat_xml_parse_error
 
+if sys.version_info[0] >= 3:
+    compat_etree_fromstring = xml.etree.ElementTree.fromstring
+else:
+    # on python 2.x the attributes and text of a node aren't always unicode
+    # objects
```
yan12125: "not always" sounds strange. Maybe it's better to explicitly state that in Python 2.x [```ElementTree._fixtext()```](https://hg.python.org/cpython/file/2.7/Lib/xml/etree/ElementTree.py#l1508) decode unicode strings to ascii if possible?

jaimeMF: Modified in ae37338e681319a28d98dc551253d9fa1830969a. Thanks for pointing this.

dstftw:
This can now be merged I suppose?

jaimeMF:
I think so, I'll try to look into making other fields Unicode as soon as possible.

dstftw:
>I'll try to look into making other fields Unicode as soon as possible.

@jaimeMF I've just manager this in ae12bc3ebb4cb377c2b4337ec255e652b36f5143 and dc519b5421366a8cac681455a817ae25f7f4aa83.

dstftw:
Thanks.

PR #7298, opened 2015-10-26
Add subtitle support to francetvinfo.fr extractor

Frenzie:


dstftw:
Cherry-picked in 5dadae079bd053c822353b081e94d9daff333208. Thanks.

Frenzie:
You're welcome. And thanks for the refactoring; it makes more sense that way around. Besides, I wasn't aware of some of the internal youtube-dl functionality in the first place. :)

PR #7305, opened 2015-10-27
[francetv] Accept mobile URLs

fenollp:


jaimeMF:
Which URLs have you tried with?, http://pluzz.francetv.fr/videos/un_ora_,130129525.html works but the mobile version http://m.pluzz.francetv.fr/videos/un_ora_,130129525.html fails (instead of looking for `data-diffusion` it seems that we need to look for `<meta name="id_video" content="`).

fenollp:
Thanks for fixing it! I missed adding the `s%://m.%://%` part in there…

PR #7322, opened 2015-10-30
[vgtv] extract videos from FTV, Aftenposten, Aftonbladet using VGTVIE

remitamine:


youtube_dl/extractor/aftenposten.py
```diff
@@ -1,23 +0,0 @@
...
-            'ext': 'mov',
-            'title': 'TRAILER: "Sweatshop" - I can´t take any more',
-            'description': 'md5:21891f2b0dd7ec2f78d84a50e54f8238',
-            'timestamp': 1416927969,
-            'upload_date': '20141125',
-        }
-    }
-
-    def _real_extract(self, url):
-        return self.url_result('xstream:ap:%s' % self._match_id(url), 'Xstream')
```
dstftw: Looks old videos with 5 digit length video id are still available via xstream in much better quality than vgtv.

youtube_dl/extractor/vgtv.py
```diff
@@ -11,16 +11,19 @@
 
 
 class VGTVIE(InfoExtractor):
-    IE_DESC = 'VGTV and BTTV'
+    IE_DESC = 'VGTV, BTTV, FTV, Aftenposten and Aftonbladet'
     _VALID_URL = r'''(?x)
-                    (?:
-                        vgtv:|
```
dstftw: Shortcuts should be preserved.

youtube_dl/extractor/aftenposten.py
remitamine: newer videos doesn't work with the xstream:
youtube-dl 'http://www.aftenposten.no/webtv/#!/video/102250/star-wars-tilbake-til-70-tallet'
[Xstream] 102250: Downloading XML
  File "/usr/bin/youtube-dl", line 9, in <module>
    load_entry_point('youtube-dl==2015.10.24', 'console_scripts', 'youtube-dl')()
  File "/usr/lib/python3.5/site-packages/youtube_dl/__init__.py", line 410, in main
    _real_main(argv)
  File "/usr/lib/python3.5/site-packages/youtube_dl/__init__.py", line 400, in _real_main
  File "/usr/lib/python3.5/site-packages/youtube_dl/YoutubeDL.py", line 1666, in download
  File "/usr/lib/python3.5/site-packages/youtube_dl/YoutubeDL.py", line 672, in extract_info
  File "/usr/lib/python3.5/site-packages/youtube_dl/YoutubeDL.py", line 725, in process_ie_result
    extra_info=extra_info)
  File "/usr/lib/python3.5/site-packages/youtube_dl/YoutubeDL.py", line 661, in extract_info
    ie_result = ie.extract(url)
  File "/usr/lib/python3.5/site-packages/youtube_dl/extractor/common.py", line 291, in extract
  File "/usr/lib/python3.5/site-packages/youtube_dl/extractor/xstream.py", line 53, in _real_extract
    video_id)
  File "/usr/lib/python3.5/site-packages/youtube_dl/extractor/common.py", line 465, in _download_xml
    return xml.etree.ElementTree.fromstring(xml_string.encode('utf-8'))
  File "/usr/lib/python3.5/xml/etree/ElementTree.py", line 1322, in XML
    return parser.close()
xml.etree.ElementTree.ParseError: no element found: line 1, column 0

dstftw: I'm aware of that.

youtube_dl/extractor/vgtv.py
remitamine: the newer regex is not compatible with the old use of the shortcut and it's used only by BTArticleIE.

dstftw: Then make it compatible? It's not about using in another extractor but about using as shortcuts by user in the first place. And it obviously should be preserved for backward compatibility

remitamine: i can make it compatible but then it can match urls that are not actualy available in the site.

youtube_dl/extractor/aftenposten.py
remitamine: i think getting videos in lower quality(720p instead 1080p) is better than getting an error in newer videos.

dstftw: I've already suggested how to cover both scenarios without getting any error.
>Looks like old videos with **5 digit length video id** are still available via xstream in much better quality than vgtv.

New video ids are 6 digit length.

remitamine: i added some code to extract 5 digit length video id using both xstream and vgtv extractors.

PR #7324, opened 2015-10-30
[baidu] improve info extraction

remitamine:


yan12125:
What's the purpose of this PR? The current approach works just fine, and it matches what browsers do.

remitamine:
- extract info directly from the api is better than downloading the webpage and using regex for extraction.
- extract direct links instead of making head requests for every url in the playlist.

yan12125:
> extract info directly from the api is better than downloading the webpage and using regex for extraction.

It's not the case in youtube-dl. youtube-dl tries to mimic browsers as much as possible so that videos available on browsers are always downloadable. If these API calls are seen in actual browsing (from different IPs, embedded pages, or mobile devices etc.), then it's acceptable. Where did you find ```xqinfo``` API calls?

> extract direct links instead of making head requests for every url in the playlist.

It's indeed a problem. However, there's no need to handle redirections in individual info extractors. [GenericIE](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/generic.py#L1820) works fine. It's my bad that I didn't notice that at the first time.

remitamine:
> Where did you find xqinfo API calls?
in mobile devices.

yan12125:
Thanks merged with some modifications.

PR #7334, opened 2015-11-01
[flickr] extract more info and formats

remitamine:


youtube_dl/extractor/flickr.py
```diff
@@ -1,67 +1,77 @@
...
+        'md5': '164fe3fa6c22e18d448d4d5af2330f31',
         'info_dict': {
             'id': '5645318632',
-            'ext': 'mp4',
-            "description": "Waterfalls in the Springtime at Dark Hollow Waterfalls. These are located just off of Skyline Drive in Virginia. They are only about 6/10 of a mile hike but it is a pretty steep hill and a good climb back up.",
-            "uploader_id": "forestwander-nature-pictures",
-            "title": "Dark Hollow Waterfalls"
+            'ext': 'mpg',
+            'description': 'Waterfalls in the Springtime at Dark Hollow Waterfalls. These are located just off of Skyline Drive in Virginia. They are only about 6/10 of a mile hike but it is a pretty steep hill and a good climb back up.',
+            'uploader_id': 'forestwander-nature-pictures',
```
jaimeMF: This field appears twice.

remitamine:
i removed the duplication in test case.
and fixed the error of expired api key.

youtube_dl/extractor/flickr.py
```diff
@@ -1,67 +1,81 @@
...
 
-        first_url = 'https://secure.flickr.com/apps/video/video_mtl_xml.gne?v=x&photo_id=' + video_id + '&secret=' + secret + '&bitrate=700&target=_self'
-        first_xml = self._download_xml(first_url, video_id, 'Downloading first data webpage')
+        api_key = self._download_json('https://www.flickr.com/hermes_error_beacon.gne', video_id, 'Downloading api key',)['site_key']
 
-        node_id = find_xpath_attr(
-            first_xml, './/{http://video.yahoo.com/YEP/1.0/}Item', 'id',
-            'id').text
+        video_info = self._call_api('photos.getInfo', video_id, api_key, 'Downloading video info')['photo']
+        if video_info['media'] == 'video':
```
jaimeMF: I think it should raise an error if it's not a video (like https://www.flickr.com/photos/forestwander-nature-pictures/5788311733/in/photostream/), otherwise it's not clear what is happening.

remitamine: i added `raise ExtractorError` if it's not a video.

PR #7336, opened 2015-11-01
[giga] improve extraction and add support for posts with multiple videos

remitamine:


youtube_dl/extractor/giga.py
```diff
@@ -45,58 +122,37 @@ class GigaIE(InfoExtractor):
...
-            'title': title,
-            'description': description,
-            'thumbnail': thumbnail,
-            'duration': duration,
-            'timestamp': timestamp,
-            'uploader': uploader,
-            'view_count': view_count,
-            'formats': formats,
-        }
+            if len(entries) == 1:
```
jaimeMF: I would always return a `multi_video` result.

youtube_dl/extractor/giga.py
```diff
@@ -7,13 +7,47 @@
...
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        default = self._call_api('default', video_id)
+        playlist = self._call_api('playlist', video_id)[0]
+
+        quality = qualities(['normal', 'hd720'])
+
+        formats = []
+        for format_id in itertools.count(0):
```
jaimeMF: Isn't `0` already the default?

jaimeMF: I don't have a strong preference, but you can achieve the same with:
```python
        for format_id, fmt in playlist.items():
            if format_id.isdigit():
                formats.append({
                    'url': fmt['src'],
                    'format_id': '%s-%s' % (fmt['quality'], fmt['type'].split('/')[-1]),
                    'quality': quality(fmt['quality']),
                })
```

remitamine: > Isn't 0 already the default?

i didn't change the code in this part.
it's already in the master branch https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/giga.py#L61

PR #7357, opened 2015-11-04
fixed vidzi and sharesix

misterhat:
Minor website changes broke these extractors.


yan12125:
Seems the two sites are irrelevant. Could you open 2 separate pull requests for each site?

PR #7387, opened 2015-11-06
[sharesix] fixed. finds new code and navigates to video page 

misterhat:
Sharesix changed their method of navigating to the video page.


dstftw:
Can you rebase?

PR #7390, opened 2015-11-06
[videolecture] add support for sites using viidea service and add support for multi part videos (fixes #7384)

### LINKED ISSUE #7384: Not able to download several videos/parts from the same videolectures.net url
Hi,

First of all thank you for this wonderful tool. I was trying to download the videos from a videolectures.net url (youtube-dl http://videolectures.net/mlss09uk_bishop_ibi/) but I have not been able to download all the videos that are listed in the web page (part1 and part2), it always downloads the first part. Is it possible to select which part to download? (I have used the playlist options but they seem not to work in this case).  I have used youtube-dl version 2015.11.02.

remitamine:


dstftw:
```
F
  File "test/test_all_urls.py", line 97, in test_no_duplicates
AssertionError: ViideaIE should not match URL u'http://videolectures.net/promogram_igor_mekjavic_eng/video/1/smil.xml' . That URL belongs to Generic.

Ran 1 test in 5.469s

FAILED (failures=1)
```

remitamine:
i will add `(?!/\d+.xml)` to the end of regex.

dstftw:
No, add the same suffix as videolectures extractor has now in master.

remitamine:
ok i will add it.

remitamine:
i added the suffix from master branch and fixed the multipart extraction.

youtube_dl/extractor/viidea.py
```diff
@@ -0,0 +1,144 @@
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from ..compat import (
+    compat_HTTPError,
```
dstftw: This is unused.

youtube_dl/extractor/viidea.py
```diff
@@ -0,0 +1,144 @@
...
+    compat_urlparse,
+)
+from ..utils import (
+    ExtractorError,
```
dstftw: And this as well.

youtube_dl/extractor/viidea.py
```diff
@@ -0,0 +1,144 @@
...
+        }],
+        'playlist_count': 2,
+    }]
+
+    def _real_extract(self, url):
+        lecture_slug, part = re.match(self._VALID_URL, url).groups()
+
+        webpage = self._download_webpage(url, lecture_slug)
+
+        cfg = self._parse_json(self._search_regex(r'cfg\s*:\s*({[^}]+})', webpage, 'cfg'), lecture_slug, js_to_json)
```
dstftw: This will fail if `cfg` turns out to contain inner dictionary.

youtube_dl/extractor/viidea.py
```diff
@@ -0,0 +1,144 @@
...
+
+        lecture_id = str(cfg['obj_id'])
```
dstftw: Use `compat_str`.

youtube_dl/extractor/viidea.py
remitamine: and this `cfg\s*:\s*({.+?}),xhr`?

dstftw: I also saw `cfg:{.+?},ga:function(`, so let's just add both:
```python
[r'cfg\s*:\s*({.+?}),[\da-zA-Z_]:\(?function', r'cfg\s*:\s*({[^}]+})']
```

youtube_dl/extractor/viidea.py
```diff
@@ -0,0 +1,144 @@
...
+                info['id'] = '%s_part%s' % (lecture_id, part)
+                switch = smil.find('.//switch')
+                if switch is not None:
+                    info['duration'] = parse_duration(switch.attrib.get('dur'))
+                return info
+            else:
+                for part in parts:
+                    entries.append(self.url_result('%s/%s/video/%s' % (base_url, lecture_slug, part), 'Viidea'))
+                lecture_info['_type'] = 'multi_video'
+        else:
```
dstftw: This branch should probably take place always when `cfg['type']` equals `"evt"` regardless of the `cfg['videos']` content - potentially `videos` of `evt` item may contain some promo videos that will "shadow" the actual videos of this event with current approach.

remitamine: may be there are other types that contain playlists?

dstftw: Maybe, who knows. At least we can try playlist if videos is empty and if type is evt (or not vpr).

remitamine: ok i will add the condition.

remitamine:
i made the requested changes.

youtube_dl/extractor/viidea.py
```diff
@@ -0,0 +1,143 @@
...
+            'display_id': lecture_slug,
+            'title': lecture_data['title'],
+            'timestamp': parse_iso8601(lecture_data.get('time')),
+            'description': lecture_data.get('description_wiki'),
+            'thumbnail': lecture_data.get('thumb'),
+        }
+
+        entries = []
+        parts = cfg.get('videos')
+        if lecture_data.get('type') != 'evt' and parts:
```
dstftw: Actually I mean to process `videos` always when `videos` present and also try playlist when either no `videos` present or `type` is `evt`.

remitamine:
is this what you mean(e7a90483509634939824f408e25ccc7da6cef350)

dstftw:
Yes.

youtube_dl/extractor/viidea.py
```diff
@@ -0,0 +1,143 @@
...
+    }]
+
+    def _real_extract(self, url):
+        lecture_slug, part = re.match(self._VALID_URL, url).groups()
+
+        webpage = self._download_webpage(url, lecture_slug)
+
+        cfg = self._parse_json(self._search_regex([r'cfg\s*:\s*({.+?}),[\da-zA-Z_]:\(?function', r'cfg\s*:\s*({[^}]+})'], webpage, 'cfg'), lecture_slug, js_to_json)
+
+        lecture_id = str(cfg['obj_id'])
```
dstftw: One more to `compat_str`.

remitamine:
i changed it to compat_str(87e59de943eef05b8bedb8878699b50bd8556560)

dstftw:
Cherry-picked. Thanks.

PR #7396, opened 2015-11-07
[crunchyroll] the video_title wasn't working.

tijder:
I changed the self._html_search_regex.


dstftw:
Post an example URL where it was not working.

tijder:
Example url: http://www.crunchyroll.com/fairy-tail/episode-82-tartaros-arc-wings-of-despair-687655
Gives "Crunchyroll" as title instead of "Fairy Tail Series 2 Episode 82 – Tartaros Arc: Wings of Despair"

dstftw:
It's better to check for `<span itemprop="title">` since first `<h1>` contains `itemscope` as well but in different order that may change in future.

PR #7436, opened 2015-11-10
Adding proxy to update procedure

davidbz:


youtube_dl/YoutubeDL.py
```diff
@@ -1994,6 +1994,9 @@ def get_encoding(self):
             encoding = preferredencoding()
         return encoding
 
+    def get_opener(self):
```
jaimeMF: Directly access the opener, after all we are only using it internally.

Kasher:
+1

youtube_dl/YoutubeDL.py
davidbz: Done

youtube_dl/update.py
```diff
@@ -13,7 +13,7 @@
     compat_str,
     compat_urllib_request,
```
dstftw: It's unused now.

davidbz: Done

dstftw:
Thanks.

PR #7455, opened 2015-11-11
Change deprecated function

sieben:
https://docs.python.org/3/library/os.path.html#os.path.splitunc

### DIFF
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -361,7 +361,7 @@ def sanitize_path(s):
         return s
     drive_or_unc, _ = os.path.splitdrive(s)
     if sys.version_info < (2, 7) and not drive_or_unc:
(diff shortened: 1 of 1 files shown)


yan12125:
This line is a fallback of the ```splitdrive``` call two lines above. It's for Python 2.6.x only, in which ```splitdrive``` does not provide sufficient functionality. FYI, here's the ```splitdrive``` function in Python 2.6.9:

```Python
# Split a path in a drive specification (a drive letter followed by a
# colon) and the path specification.
# It is always true that drivespec + pathspec == p
def splitdrive(p):
    """Split a pathname into drive and path specifiers. Returns a 2-tuple
"(drive,path)";  either part may be empty"""
    if p[1:2] == ':':
        return p[0:2], p[2:]
    return '', p
```

PR #7467, opened 2015-11-12
Fix for downloading multiple formats with --download-archive argument

dundua:
Fix for issue described in issue https://github.com/rg3/youtube-dl/issues/7022


dstftw:
`record_download_archive` must be called **only** after media was successfully downloaded and postprocessed. This change does not respect this at all. With this change video will be always written to archive even if it was skipped due to filters or one of simulate or `--skip-download` options, some error occured while downloading, some postprocessor failed, etc.

PR #7480, opened 2015-11-12
Updated fix for downloading multiple formats with --download-archive argument

dundua:
A updated fix for issue described in https://github.com/rg3/youtube-dl/issues/7022. This fixes the issues brought up in https://github.com/rg3/youtube-dl/pull/7467 where a record would be saved in a download archive even when a error occurs.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1330,11 +1330,15 @@ class YoutubeDL(object):
         if download:
             if len(formats_to_download) > 1:
                 self.to_screen('[info] %s: downloading video in %s formats' % (info_dict['id'], len(formats_to_download)))
+            download_success = 0
             for format in formats_to_download:
                 new_info = dict(info_dict)
                 new_info.update(format)
-                self.process_info(new_info)
-            self.record_download_archive(info_dict)
+                success = self.process_info(new_info)
+                if success:
+                    download_success += 1
+            if download_success == len(formats_to_download):
+                self.record_download_archive(info_dict)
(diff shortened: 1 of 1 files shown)


dstftw:
Now you have another problem due to the fact you write archive only when all formats succeeded when it could be a partial success. In this case nothing will be written to the archive despite of some succeeded formats. So when the same command will be initiated again youtube-dl will again try to redownload already succeeded formats that is incorrect. Or another example: if one decides after fully successful archiving to download some more formats of same videos he will get an erroneous response that it's already archived.
All these bogus scenarios are caused by archive format not taking `format_id` into account. Since `format_id` in general is not guaranteed to be stable enough as video `id` is I'm doubtful whether this improvement worth the trouble.

dundua:
In the first example, youtube-dl should not redownload succeeded formats since it will report the file already exists and continue downloading from the last file. In the latter example, this issue already exists in the current version.

Since these issues are mainly caused by the archive file not including format_id, would it be possible to update the archive file format to include it. I am hesitant to do so, since it will create backwards compatibility issues.

dstftw:
You don't get it. In the first case it should not even try to redownload. If you'll delete or move archived files from the original location they were downloaded to but not from archive file youtube-dl still **must not** download already archived files. Currently it will redownload them with this patch. The latter problem does not exist in current version since currently download archive is not expected to be used for multiple formats.
I've already mentioned relying on `format_id` is questionable due to it's possible unstability. In any case if `format_id` is going to be added to download archive format it still must support the old syntax for backward compatibility.

PR #7505, opened 2015-11-15
[tnaflix] Added handling of embedded videos

devnomnom:
Looks for embedded tnaflix player
Added a test for it

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -55,7 +55,6 @@ from .snagfilms import SnagFilmsEmbedIE
 from .screenwavemedia import ScreenwaveMediaIE
 from .mtv import MTVServicesEmbeddedIE
 
-
 class GenericIE(InfoExtractor):
     IE_DESC = 'Generic downloader that works on some sites'
     _VALID_URL = r'.*'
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/generic.py
```diff
@@ -1778,6 +1790,13 @@ def _playlist_from_matches(matches, getter=None, ie=None):
                 self._proto_relative_url(unescapeHTML(mobj.group(1))),
                 'AdobeTVVideo')
 
+        # Look for embedded tnaflix player
+        tnaflix_urls = [url for url in re.findall(
+            r'(?P<url>https?://(?:www\.)?tnaflix\.com/[^/]+/[^/]+/video\d+)',
```
dstftw: This is too broad. It must not capture plain text URLs.

youtube_dl/extractor/generic.py
```diff
@@ -55,7 +55,6 @@
 from .screenwavemedia import ScreenwaveMediaIE
 from .mtv import MTVServicesEmbeddedIE
 
-
```
dstftw: Do not break PEP 8 guidelines.

PR #7514, opened 2015-11-16
[neteasemusic] Fixes #7301

### LINKED ISSUE #7301: neteasemusic no longer works

ping:
Switching to http://m5.music.126.net (original http://m1.music.126.net) fixes the download problem.

### DIFF
--- a/youtube_dl/extractor/neteasemusic.py
+++ b/youtube_dl/extractor/neteasemusic.py
@@ -40,7 +40,7 @@ class NetEaseMusicBaseIE(InfoExtractor):
             if not details:
                 continue
             formats.append({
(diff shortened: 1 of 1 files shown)


yan12125:
Currently ```m5.music.126.net``` is hardcoded. Can it be extracted from some webpages? Netease may change the URL again in the future.

ping:
Not that I know of. The netease music website uses ``m2.music.126.net`` which is geoblocked afaik and doesn't resolve outside of mainland China. I cannot use the website at all unless I edit ``/etc/hosts``.

On further investigation it looks like any of ``m5.music.126.net / p1.music.126.net / p2.music.126.net / p3.music.126.net / p4.music.126.net`` will work.

Not sure how to best incorporate that into code though.

Issue #7517, opened 2015-11-16
How to know when all playlist videos as been downloaded?


hedii:
When downloading playlist files, the line 'Downloading video x of x' is useful.
But it would be even more useful to have a line like that at the end of the output:
'All playlist videos downloaded'

I am using youtube-dl in a php app. I am parsing result lines to get progress, ETA, bandwidth, filename, etc..
But there is no information at all that i can use to know when the playlist download as ended.

Is there something to know when all playlist videos as been downloaded?

jaimehrubiks:
I don't see why you need that. You just check one by one and in the last one x/x you can see when it ends

hedii:
no, because when the last x/x is displayed, the video has not yet been downloaded, or converted.

jaimehrubiks:
I see your point, and it is a good idea. Anyway I checked what I did and it is to wait for the process to end (it is in java anyway), you can also read the percentage after last x/x or even look for the word "error" for some cases.

hedii:
@jaimehrubiks thanks for your answer

- waiting for the process to end is not relevant if youtube-dl is fired via an api: we just launch the download command and don't wait for a response. the response is the parsing of the command log (stored somewhere) each x sec. this is it that can indicate progress, download finish, conversion, etc...

- read the percentage after last x/x is not relevant because there could be a converting process after the last download. if we want to do something on all the playlist files (in my case, zip the whole playlist), we have to wait for the end of the last converting process. To know the last download/convert as ended, we have to use very ugly, long, performance-killing regex that parse not the just the last line, but the whole file (last line and up to 'Downloading video x of x').

- when youtube-dl is embeded in an external app (i've done it several times), it is better for performance to only parse the last line of the cmd output. We can use --exec echo 'blablabla' to output something that indicates the end of the download/conversion process when asking for a single video, but --exec is fired for each videos in a playlist, so this workaround is ineffective for playlists.

What about creating a new post processing option like that:

--exec-playlist CMD : Execute a command after all playlist's videos has been downloaded.

Or even easier, output a short message when the last video has been downloaded/converted: 'Process finished'.
It could be added to everything, not just playlists...

I don't know the python language, so i cannot propose a pull request :(

jaimehrubiks:
@hedii I see now. I wish someone adds this.

Anyway, again, I will try to help you :P
Just this: 
youtube-dl https://www.youtube.com/playlist?list=PLAYLISTID end
and the last line is...
[...]

PR #7525, opened 2015-11-17
Add an info string when playlist download is finished

hedii:
see #7517 
When the playlist is fully downloaded/converted, output an info message.
Without an indicator, it is very difficult to know when all playlist videos processing have finished  when youtube-dl is embedded in an external app. Just adding this line solves a lot of code/performance when building an app based on youtube-dl fro downloading playlists.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -833,7 +833,7 @@ class YoutubeDL(object):
                                                       extra_info=extra)
                 playlist_results.append(entry_result)
             ie_result['entries'] = playlist_results
-            self.to_screen('[download] Playlist %s : download process finished' % playlist)
+            self.to_screen('[download] Finished downloading playlist: %s' % playlist)
             return ie_result
         elif result_type == 'compat_list':
             self.report_warning(
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -833,7 +833,7 @@ class YoutubeDL(object):
                                                       extra_info=extra)
                 playlist_results.append(entry_result)
             ie_result['entries'] = playlist_results
-            self.to_screen('[download] Playlist download process finished')
+            self.to_screen('[download] Playlist %s : download process finished' % playlist)
             return ie_result
         elif result_type == 'compat_list':
             self.report_warning(
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -833,7 +833,7 @@ class YoutubeDL(object):
                                                       extra_info=extra)
                 playlist_results.append(entry_result)
             ie_result['entries'] = playlist_results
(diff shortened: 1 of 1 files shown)


youtube_dl/YoutubeDL.py
```diff
@@ -833,6 +833,8 @@ def iter_playlistitems(format):
                                                       extra_info=extra)
                 playlist_results.append(entry_result)
             ie_result['entries'] = playlist_results
+            if result_type == 'playlist':
```
dstftw: Why is this condition?

youtube_dl/YoutubeDL.py
```diff
@@ -833,6 +833,8 @@ def iter_playlistitems(format):
...
+                self.to_screen('[info] Playlist download process finished')
```
dstftw: This does not stylistically matches [opening message](https://github.com/hedii/youtube-dl/blob/output-message-when-playlist-downloaded/youtube_dl/YoutubeDL.py#L747), should be same prefix. Also it does not mention the playlist name.

youtube_dl/YoutubeDL.py
hedii: because I don't know what is 'multi_video', and this would apply to 'multi_video' without this condition. Do I have to remove it?

dstftw: Opening message is already applied for `multi_video`, I don't see any reason for closing message not to be applied for it as well.

youtube_dl/YoutubeDL.py
hedii: Is [download] relevant if the whole process finishes after an [ffmpeg] or [avconv] conversion?

dstftw: It does since there may be no postprocessing at all.

youtube_dl/YoutubeDL.py
```diff
@@ -833,6 +833,7 @@ def iter_playlistitems(format):
                                                       extra_info=extra)
                 playlist_results.append(entry_result)
             ie_result['entries'] = playlist_results
+            self.to_screen('[download] Playlist %s : download process finished' % playlist)
```
dstftw: Why not just `[download] Finished downloading playlist: %s`?

yan12125:
@dstftw Maybe the opening message needs some changes, too. Here's an example: http://v.baidu.com/tv/23137.htm. It's "a playlist of multi-videos".

dstftw:
In fact it had to be changed when `multi_video` was introduced. I don't know whether someone already relies on this already or not but mentioning playlist name or id is well enough to guess pair of opening/closing messages.

youtube_dl/YoutubeDL.py
hedii: yes, i've change the message

dstftw:
Cherry-picked in 371c3b796cb4168dbb2cb1ad48e3cf12745a8601.

PR #7531, opened 2015-11-17
added openings.moe support

Francesco149:
example usages
```bash

youtube_dl http://openings.moe # random video
youtube_dl http://openings.moe/?video=InstallLinux.webm # specific video

```


jaimeMF:
This doesn't seem to do anything that the generic extractor is not doing, why do you think an specific extractor is needed?

> youtube_dl http://openings.moe/?video=InstallLinux.webm # specific video

That link redirects me to http://openings.moe/

Francesco149:
I mistakenly thought the default extractor wasn't enough because when I first tested it I used https, which was not supported by openings.moe.
My bad, haha!

PR #7569, opened 2015-11-19
Force protocol-less URLs to HTTP

dstftw:
The general intention is to mitigate the number of failures due to non-`_proto_relative_url`'d protocol-less URLs (e.g. https://github.com/rg3/youtube-dl/issues/7532) that are very frequent nowadays.
The idea is to require the usage of `compat_urllib_request_Request` instead of `compat_urllib_request.Request` that treats all protocol-less URLs as HTTP ones (I've tried another approaches with patching requests on-fly and before `urlopen` but in python 3 protocol is validated inside `Request`'s ctor so this seems to be the most reasonable workaround). I guess HTTP is most probable scenario and even if it's not in particular case we will be most likely redirected to HTTPS when trying HTTP.
In fact this routine is not a compatibility one so it's probably better to move it to utils. It could also be useful for any other `Request` preprocessing that may be added in future, e.g. we can wrap in `compat_urllib_request_Request` [all this unwieldy stuff](https://github.com/rg3/youtube-dl/blob/6d8f1076397a9193ceb174e4e404364511e4d781/youtube_dl/extractor/vimeo.py#L43) needed to properly encode `data` on python 2 that tend to be constantly forgotten and allow passing `data` dictionaries directly to `compat_urllib_request_Request`).
If we agree on this proposal I'll change all remaining usages of `compat_urllib_request.Request` to `compat_urllib_request_Request` throughout the codebase.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -28,6 +28,7 @@ if os.name == 'nt':
     import ctypes
 
 from .compat import (
+    compat_basestring,
     compat_cookiejar,
     compat_expanduser,
     compat_get_terminal_size,
@@ -38,6 +39,7 @@ from .compat import (
     compat_urllib_error,
     compat_urllib_request,
     compat_urllib_request_DataHandler,
+    compat_urllib_request_Request,
 )
 from .utils import (
     ContentTooShortError,
@@ -1871,6 +1873,8 @@ class YoutubeDL(object):
 
     def urlopen(self, req):
         """ Start an HTTP download """
+        if isinstance(req, compat_basestring):
+            req = compat_urllib_request_Request(req)
         return self._opener.open(req, timeout=self._socket_timeout)
 
     def print_debug_header(self):
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -39,7 +39,6 @@ from .compat import (
     compat_urllib_error,
     compat_urllib_request,
     compat_urllib_request_DataHandler,
-    compat_urllib_request_Request,
 )
 from .utils import (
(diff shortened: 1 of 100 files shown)


jaimeMF:
> In fact this routine is not a compatibility one so it's probably better to move it to utils

I think so, if you are running the last python version you should be able to only keep the renames in compat.py.

dstftw:
Did not quite get it about renames.
Renamed to `sanitized_Request` and moved to `utils`. Alternative naming suggestions are welcome.

jaimeMF:
I'm OK with the PR.

> Did not quite get it about renames.

What I wanted to say is that on python3 a working `compat.py` would just contain simple statements:
```python
import urllib.request as compat_urllib_request
import urllib.error as compat_urllib_error
...
```
So there wouldn't be any real relevant code, after all it's mean to just bring missing features to old python versions (that's the reason you pointed in the PR message) not adding extra functionality.
(I think that the only part that is not a simple statement is the definition of `compat_print`.)

dstftw:
Switched codebase to use `sanitized_Request` instead of `compat_urllib_request.Request`.

jaimeMF:
I would include all the switchs to `sanitized_Request`, it would make the git log cleaner.

dstftw:
Include where?

jaimeMF:
> Include where?

I would include all the switchs to sanitized_Request *in a single commit* (sorry, I should have read the message a second time).

dstftw:
Squashed in 3d460ebdce6ef085e615e804edb8128890a1cd74.

dstftw:
I'd merge this if there are no more suggestions yet there are no that much conflicts.

jaimeMF:
I'm OK with the changes.

PR #7589, opened 2015-11-21
[pluralsight] do not extract videos if user doesn't have permission

remitamine:


youtube_dl/extractor/pluralsight.py
```diff
@@ -215,7 +219,7 @@ def _real_extract(self, url):
         for module in course_data:
             for clip in module.get('clips', []):
                 player_parameters = clip.get('playerParameters')
-                if not player_parameters:
+                if not player_parameters or clip.get('userMayViewClip') is False:
```
dstftw: Relying on `userMayViewClip` is probably a [bad idea](https://github.com/rg3/youtube-dl/commit/2b6bda1ed86e1b64242b33c032286dc315d541ae#diff-0b85b33765d2939b773726fda5a55b06).

remitamine: i don't have an acount to test with.so if these can cause problems for registred users then just ignore it.
if raising an error if `authorized` is false in clip info is also doesn't work than i will close this PL and let an other one that have an account to fix this.

dstftw: Neither do I. But afair when I had one, `userMayViewClip` caused some problems by not always correlating with videos that can't actually be downloaded.

PR #7599, opened 2015-11-22
[youtube] More explicit player config JSON extraction (fixes #7468)

### LINKED ISSUE #7468: ValueError when downloading a video from YouTube
Downloading [this video](https://www.youtube.com/watch?v=Ms7iBXnlUO8) seems to fail:

```
$ PYTHONPATH=`pwd`  ./bin/youtube-dl --verbose --list-formats 'https://www.youtube.com/watch?v=Ms7iBXnlUO8'
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--verbose', u'--list-formats', u'https://www.youtube.com/watch?v=Ms7iBXnlUO8']
[debug] youtube-dl version 2015.11.10
[debug] Git HEAD: fcd817a
[debug] Python version 2.7.9 - Linux-3.19.0-33-generic-x86_64-with-Ubuntu-15.04-vivid
[debug] exe versions: ffmpeg 2.5.8-0ubuntu0.15.04.1, ffprobe 2.5.8-0ubuntu0.15.04.1, rtmpdump 2.4
[debug] Proxy map: {}
[youtube] Ms7iBXnlUO8: Downloading webpage
  File "./bin/youtube-dl", line 6, in <module>
    youtube_dl.main()
  File "/home/lukas/work/youtube-dl/youtube_dl/__init__.py", line 410, in main
    _real_main(argv)
  File "/home/lukas/work/youtube-dl/youtube_dl/__init__.py", line 400, in _real_main
  File "/home/lukas/work/youtube-dl/youtube_dl/YoutubeDL.py", line 1666, in download
  File "/home/lukas/work/youtube-dl/youtube_dl/YoutubeDL.py", line 661, in extract_info
    ie_result = ie.extract(url)
  File "/home/lukas/work/youtube-dl/youtube_dl/extractor/common.py", line 290, in extract
  File "/home/lukas/work/youtube-dl/youtube_dl/extractor/youtube.py", line 1080, in _real_extract
    ytplayer_config = json.loads(json_code)
  File "/usr/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python2.7/json/decoder.py", line 382, in raw_decode
    obj, end = self.scan_once(s, idx)
ValueError: Unterminated string starting at: line 1 column 6498 (char 6497)
```

lalinsky:
Fixed downloads for these videos:

 - https://www.youtube.com/watch?v=Ms7iBXnlUO8
 - https://www.youtube.com/watch?v=lsguqyKfVQg

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -674,7 +674,23 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
         {
             'url': 'http://vid.plus/FlRa-iH7PGw',
             'only_matching': True,
-        }
+        },
+        {
+            # Title with JS-like syntax "};"
+            'url': 'https://www.youtube.com/watch?v=lsguqyKfVQg',
+            'info_dict': {
+                'id': 'lsguqyKfVQg',
+                'ext': 'mp4',
+                'title': '{dark walk}; Loki/AC/Dishonored; collab w/Elflover21',
+                'description': 'md5:8085699c11dc3f597ce0410b0dcbb34a',
+                'upload_date': '20151119',
+                'uploader_id': 'IronSoulElf',
+                'uploader': 'IronSoulElf',
+            },
+            'params': {
+                'skip_download': True,
+            },
+        },
     ]
 
     def __init__(self, *args, **kwargs):
@@ -858,16 +874,25 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
             return {}
         return sub_lang_list
 
+    def _get_ytplayer_config(self, webpage):
+        patterns = [
+            r';ytplayer\.config\s*=\s*({.*?});ytplayer',
+            r';ytplayer\.config\s*=\s*({.*?});',
+        ]
+        for pattern in patterns:
+            config = self._search_regex(pattern, webpage, 'ytconfig.player', default=None)
+            if config is not None:
+                return json.loads(uppercase_escape(config))
+
     def _get_automatic_captions(self, video_id, webpage):
         """We need the webpage for getting the captions url, pass it as an
            argument to speed up the process."""
         self.to_screen('%s: Looking for automatic captions' % video_id)
-        mobj = re.search(r';ytplayer.config = ({.*?});', webpage)
+        player_config = self._get_ytplayer_config(webpage)
         err_msg = 'Couldn\'t find automatic captions for %s' % video_id
-        if mobj is None:
+        if player_config is None:
             self._downloader.report_warning(err_msg)
             return {}
-        player_config = json.loads(mobj.group(1))
         try:
             args = player_config['args']
             caption_url = args['ttsurl']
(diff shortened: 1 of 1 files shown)


dstftw:
This does not fix the actual problem. If this is intentional YouTube measure against downloading (I can't find another reasonable explanation for that) they'll just changed it to `info: {}});ytplayer}, n);}};(function(w, startTick` and it will be broken again. Instead we should correctly process quoted strings obviously not with regexes since it will cause lots of backtracking.

dstftw:
If you suggest this as temporary "solution" then preserve both regexes keeping `;ytplayer`-one first (e.g. with `_search_regex`).

lalinsky:
This is all user content, it's not intentional from YouTube. The last video I posted to the issue has `"{dark walk}; Loki/AC/Dishonored; collab w/Elflover21"` as the title and the `};` part confuses the regex. Obviously it would be great to have a proper JS parser, but this seems like a simple enough fix. I'm sure this kind of problem exists on many places in youtube-dl, which tries to extract JSON from HTML using regexes and it was always fixed by adding a more specific regex.

lalinsky:
Ok, I'll switch the code to `_search_regex` and add a fallback to the original one.

lalinsky:
Done. There was one more instance of this, so I made a new function for the extraction. Please have a look.

dstftw:
`_search_regex` already supports list of patterns.

lalinsky:
Oh, I didn't know that. Will change.

dstftw:
Thanks.

RomanStone:
Found link to this page in "src" code while trying to do `ytplayer.config` extract in php

Please let me suggest another method, 
it's written in php because I don't know how to "code" in python, but I like youtube-dl,
and want to help.

```
<?php
$VideoID = 'aj9coPGuFkM';
$file = file_get_contents('https://www.youtube.com/watch?v=' . $VideoID);


$pattern1 = '/ytplayer\.config\s*=\s*\{(.+?)\}.*/';
preg_match($pattern1, $file, $matches1);

if(!empty($matches1[0]) && strpos($matches1[0], '}') !== false) {

  //don't need it now
  $matches1 = preg_replace("/ytplayer\.config\s*=\s*/", '', $matches1[0]);
  if(!empty($matches1)) {

    // explode by ending bracket
    $js_array = explode('}', $matches1);
    unset($matches1);

    $b = '';
    foreach($js_array as $str) {
      // and reasemble sequentially
      $b .= $str . '}';
      // try to decode any string
      $out = json_decode($b, true);
      if(!empty($out)) {
        // ok found something, turn it back to json;
        $ytplayer_config = json_encode(array('ytplayer.config' => $out));
        unset($b); unset($js_array); unset($out);
        break;
      }
    }
  }
}


// Do something
if(!empty($ytplayer_config)) {
  print_r($ytplayer_config);
} else {
  echo "Failed to extract ytplayer_config";
}
?>

```

Not really nice way to extract, but it's work


Php-sandbox
http://sandbox.onlinephpfunctions.com/code/fec6f56d8d4d7d3938941ca767acfb1c428751e5

yan12125:
@RomanStone Seems your code is what youtube-dl used in the past, which has caused problems for some videos.

remitamine:
i think, `https://www.youtube.com/watch?v=video_id&spf=navigate` can be used as another fallback to extract ytplayer config.

PR #7618, opened 2015-11-23
Update README.md

alanyee:
Fat meat is greasy -- shortening word length


dstftw:
I don't see any benefit changing it, sorry.

PR #7619, opened 2015-11-23
Add MTV81 extractor

desseim:
Support for videos from http://www.mtv81.com/ .

Composition of `MTVServicesInfoExtractor`'s feed URL is also moved `_get_feed_url` to allow its overriding.

### DIFF
--- a/youtube_dl/extractor/mtv.py
+++ b/youtube_dl/extractor/mtv.py
@@ -4,10 +4,13 @@ import re
 
 from .common import InfoExtractor
 from ..compat import (
-    compat_urllib_parse,
     compat_str,
 )
 from ..utils import (
+    compat_urlparse,
+    compat_urllib_parse,
+    compat_urllib_parse_urlparse,
+    compat_parse_qs,
     ExtractorError,
     find_xpath_attr,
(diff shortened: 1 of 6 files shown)


youtube_dl/extractor/mtv.py
```diff
@@ -170,12 +178,7 @@ def _get_video_info(self, itemdoc):
     def _get_videos_info(self, uri):
         video_id = self._id_from_uri(uri)
         feed_url = self._get_feed_url(uri)
-        data = compat_urllib_parse.urlencode({'uri': uri})
-        info_url = feed_url + '?'
-        if self._LANG:
-            info_url += 'lang=%s&' % self._LANG
-        info_url += data
-        return self._get_videos_info_from_url(info_url, video_id)
+        return self._get_videos_info_from_url(feed_url, video_id)
```
jaimeMF: This change breaks MTVServicesEmbeddedIE

desseim: Thanks, I've refactored it a bit and `MTVServicesEmbeddedIE` test should now pass again.

yan12125:

PR #7673, opened 2015-11-27
Tweak francetv.py to handle france3-regions subdomain.

wapiflapi:
Small changes to regexes to handle the france3-regions subdomain for francetvinfo.fr.
Thanks to @thc202 on IRC for the patch.


dstftw:
Add a test.

PR #7678, opened 2015-11-28
Basic playlist generation

cherez:
Added a `--write-playlists` option to generate .m3u playlists when downloading playlists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,7 @@ which means you can modify it, redistribute it or use it however you like.
     --write-info-json                Write video metadata to a .info.json file
     --write-annotations              Write video annotations to a
                                      .annotations.xml file
+    --write-playlists                Write playlists to .m3u files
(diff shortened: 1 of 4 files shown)


dstftw:
1. Playlist's `info_dict` should not contain `ext` field, moreover `--write-playlists` should not affect existing `info_dict`'s in any way.
2. It should be `m3u8` and UTF-8 encoded.
3. It should not fail on non-ASCII video filenames.
4. It should not fail when output template is specified (e.g. `-o %(playlist)s/%(title)s.%(ext)s`).
4. It should write metadata according to m3u8 specification (otherwise I don't see much point in this feature since same may be achieved with https://github.com/rg3/youtube-dl/issues/6664#issuecomment-134570722).
5. There should be a `to_screen` message telling user about m3u8 playlist being written.
6. I don't see any reason for not allowing this feature for non playlists, i.e. single videos.

PR #7057, opened 2015-10-03
[cspan] correct the clip info extraction (fixes #7335)

remitamine:

PR #7681, opened 2015-11-28
[skynewsarabia] Add new extractor

remitamine:


youtube_dl/extractor/skynewsarabia.py
```diff
@@ -0,0 +1,115 @@
...
+            '_type': 'url_transparent',
+            'url': 'limelight:media:%s' % self._get_limelight_media_id(video_data['videoUrl'][0]['url']),
+            'id': video_id,
+            'title': video_data['headline'],
+            'description': video_data.get('summary'),
+            'thumbnail': self._get_image_url(video_data['mediaAsset']['imageUrl']),
+            'timestamp': parse_iso8601(video_data.get('date')),
+            'duration': parse_duration(video_data.get('runTime')),
+            'tags': video_data.get('tags'),
+            'categories': [video_data.get('topicTitle')],
```
dstftw: When there is no `topicTitle`, `categories` will be `[None]` that does not make any sense.

dstftw:
Just a reminder: you can commit right in master.

youtube_dl/extractor/skynewsarabia.py
remitamine: i changed the code to return empty array if there is no topicTitle.

remitamine:
> Just a reminder: you can commit right in master.
i didn't know this before.
anyway, Still need assistance to find the errors in the code that need to be fixed.

dstftw:
Haven't you received mail from me dated November 9 I've sent to your email found in git log?

remitamine:
it happen by mistake that i make the github to watch for the youtube-dl repository and when i connect next week i find a lot of notification so i marked them all as read.
and i didn't check the email for a long time.
i will search now for the message, i didn't read it yet.

remitamine:
I can not access the old notification in github(may be because i participated in the new interface) i will search in the email.

dstftw:
I've just resent it again.

remitamine:
if i can.i will add the changes in #7057 as they fix an error in the implementation(downloading program when the clip is requested) and next week when i have time i will test the other pull requests that i've send before.

PR #7690, opened 2015-11-29
[lrt] fix the broken extractor

GiedriusS:
The website has moved to using m3u8 so changes are needed as a result. The "real" format is still mp4. Also, makes the test pass again with no changes to it

### DIFF
--- a/youtube_dl/extractor/lrt.py
+++ b/youtube_dl/extractor/lrt.py
@@ -25,7 +25,7 @@ class LRTIE(InfoExtractor):
             'duration': 1783,
         },
         'params': {
-            'skip_download': True,  # HLS download
+            'skip_download': True,  # m3u8 download
         },
     }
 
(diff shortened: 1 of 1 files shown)


GiedriusS:
@naglis ping

GiedriusS:
Ping

GiedriusS:
Ping

youtube_dl/extractor/lrt.py
```diff
@@ -37,26 +33,11 @@ def _real_extract(self, url):
...
-                    'url': data['streamer'],
-                    'play_path': 'mp4:%s' % data['file'],
-                    'preference': -1,
-                    'rtmp_real_time': True,
-                })
-            else:
-                formats.extend(
-                    self._extract_m3u8_formats(data['file'], video_id, 'mp4'))
+        link = self._search_regex(r'file: "(.*)" \+ location\.hash\.substring\(1\)', webpage, 'link to m3u8')
+        link += video_id
```
jaimeMF: The test video (http://www.lrt.lt/mediateka/irasas/54391/) seems to work fine without this line, and it doesn't seem to be used by their webpage (`location.hash` is empty), so I'm going to remove it.

jaimeMF:
Merged, thanks! (And sorry for the delay)

PR #7691, opened 2015-11-29
Always use PYTHON env var in Makefile

ryandesign:
If "python" is python 2.5 (or presumably older), building youtube-dl fails, even if the PYTHON environment variable has been set to a newer version of python, such as 2.7. See [this MacPorts bug report](https://trac.macports.org/ticket/49840). The solution is to fix the Makefile so that it always uses the PYTHON environment variable rather than sometimes calling "python".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,34 +60,34 @@ youtube-dl: youtube_dl/*.py youtube_dl/*/*.py
 	chmod a+x youtube-dl
 
 README.md: youtube_dl/*.py youtube_dl/*/*.py
-	COLUMNS=80 python youtube_dl/__main__.py --help | python devscripts/make_readme.py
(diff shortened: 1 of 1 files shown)


yan12125:
Not tested, but isn't setting ```$PATH``` enough?

ryandesign:
You already provide a PYTHON environment variable, which people can set when they run the Makefile to indicate which python they want to use. You should either support the environment variable completely, with this patch, or remove it entirely.

yan12125:
Thanks. Merged.

dstftw:
@ryandesign note that youtube-dl does not support any python version prior to 2.6.

ryandesign:
@dstftw Exactly. But I still want to build youtube-dl on operating systems that provide python 2.5 (and I provide elsewhere a copy of python 2.7). That is why I am specifying the path to a copy of python 2.7 in the PYTHON env var, and was surprised that the build system was not respecting that directive. Now it is.

Issue #5697, opened 2015-05-14
Drop python 2.6 support


jaimeMF:
[REHL 7 ships with python 2.7 since June 2014](https://access.redhat.com/articles/676453) and the [same for CentOS 7](https://wiki.centos.org/About/Product?highlight=%28python%29), so I think that all the current releases for main Linux distributions have stopped shipping python 2.6 only, OS X includes 2.7 for a few releases and on *BSDs 2.7 can be easily installed from the ports. Therefore all users with updated software could continue to use it.

Some of the problems with python 2.6:
* It won't get any security updates, so it can be dangerous.
* We have to workaround around some features missing in the standard library. For example some `assert*` methods are not available in `unittest` and the `xml` module doesn't support xpath attributes queries or namespaces. One of the most annoying thing that is missing is the need to explicitly set the index for the `.format` method. That's also true to some degree for 2.7 vs 3.x, but 2.7 is currently widely deploy so dropping it would mean that a lot of users would stop using youtube-dl.

If someone is using outdated software it would unfortunately stop to work. This can annoy some users, for example those using a PPC Mac where youtube-dl can be the only option for viewing videos from the Internet; although they can probably manually build python 2.7 (I think I managed to compile python 3.{something} in OS X 10.4) or use something like [Tigerbrew](https://github.com/mistydemeo/tigerbrew).

Some interesting read about the problem with supporting python 2.6 is [this post by Nick Coghlan](http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html) and [this one from Alex Gaynor](https://alexgaynor.net/2015/mar/30/red-hat-open-source-community/).

yan12125:
Are we going to remove python 2.6-specific workarounds? Or just change the policy to "Python 2.6 series is not officially supported" and declare it in documents? The latter option brings less annoyance to users.

jaimeMF:
I would first say that we only support 2.7+ and then start removing the workarounds, additionally for any new code that is added we wouldn't check if it works on 2.6 and we would assume that all the features in 2.7 are available.
I don't think that keeping the workarounds is useful, I would say a big part of dropping support for 2.6 is getting rid of them, and eventually users will get annoyed when some change we made is incompatible.

dstftw:
[...]

PR #7706, opened 2015-11-30
Added argparse and multiple match/reject title options

hitbox:
Added support for argparse for Python >= 2.7.0 and allowance for multiple `--reject-title` and `--match-title` options. Added test `test_optParse.py`.

I was having a lot of frustration trying to reject multiple titles until I figured out that the program will only honor the last one given.

### DIFF
--- a/test/test_optParse.py
+++ b/test/test_optParse.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+from __future__ import unicode_literals
+
+# Allow direct execution
+import os
+import sys
+import unittest
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+try:
+    from StringIO import StringIO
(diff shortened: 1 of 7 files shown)


jaimeMF:
We are not going to maintain two separate versions of the option handling code. [I think the `append` action](https://docs.python.org/2/library/optparse.html#other-actions) can be used in optparse to handle your case.

hitbox:
Might we switch to argparse? I saw [some interest in it over a year ago](https://github.com/rg3/youtube-dl/issues/583). I can eliminate the old optParse class.

jaimeMF:
> Might we switch to argparse? I saw some interest in it over a year ago. I can eliminate the old optParse class.

Not until we stop supporting python2.6 (see #5697), as said in https://github.com/rg3/youtube-dl/issues/583#issuecomment-43994673.

PR #7727, opened 2015-12-01
Multiple title options

hitbox:
Added support for multiple `--reject-title` and `--match-title` options. Added test `test_options.py`.

### DIFF
--- a/test/test_options.py
+++ b/test/test_options.py
@@ -1,0 +1,28 @@
+#!/usr/bin/env python
+from __future__ import unicode_literals
+
+# Allow direct execution
+import os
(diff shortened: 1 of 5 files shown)


dstftw:
What's the point actually? Both are regexes, you can just pass arbitrary regex in single command.

hitbox:
I think if it's going to accept multiples of these options it should handle them in some way, rather than just throw away all but the last. Even if it's to fail and say it only accepts one and advise the user.
While I was in there I accidently noticed it accepted `--reject` without thowing up an "unrecognized option" error and exiting.

PR #7728, opened 2015-12-02
[trollvids] Add new extractor

nexAkari:
Extractor for trollvids.com. Very basic, supports only the most common player type for now (nuevo).


dstftw:
Config XML is similar to [anitube](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/anitube.py) and [trutube](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/trutube.py) and should be generalized in reusable method.

nexAkari:
@dstftw No, that would throw a ValueError for float literals. Although I see that some extractors simply put a float float in the "duration" key and that seems to cause no problems. The InfoExtractor documentation says it should be an integer, perhaps it should be updated?

nexAkari:
@dstftw Alright, I could do that. Where would be the best place to put that method then? Simply in the InfoExtractor base class? I'm not very familiar with the codebase yet.

dstftw:
Create a base class for this player similar to [spruto](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/vimple.py#L7-L26) and inherit all 3 extractors from it.

nexAkari:
@dstftw ok, done.

youtube_dl/extractor/nuevo.py
```diff
@@ -0,0 +1,33 @@
...
+
+class NuevoBaseIE(InfoExtractor):
+    def _extract_nuevo(self, config_url, video_id, info=None, ignore_hd=False):
+        if info is None:
+            info = {}
+
+        sdformats, hdformats = [], []
+        tree = self._download_xml(config_url, video_id, transform_source=lambda s: s.strip())
+
+        for child in tree:
```
dstftw: Instead of this loop use [anitube's approach](https://github.com/nexAkari/youtube-dl/blob/6722ebd43720e836af8217fa078fa1a604b98229/youtube_dl/extractor/anitube.py#L35-L51). For text fields use `utils.xpath_text`.

youtube_dl/extractor/nuevo.py
```diff
@@ -0,0 +1,33 @@
+# encoding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+from ..utils import float_or_none
+
+class NuevoBaseIE(InfoExtractor):
+    def _extract_nuevo(self, config_url, video_id, info=None, ignore_hd=False):
```
dstftw: hd should be always tried to be extracted whether it's present or not. There is no need in this flag.

dstftw: Also there is no need in `info` here - you can return info dict from here and add more fields at the call place once it's returned.

youtube_dl/extractor/trollvids.py
```diff
@@ -0,0 +1,46 @@
...
+
+from ..compat import (
+    compat_urllib_parse_unquote
+)
+
+import re
+
+
+class TrollvidsIE(NuevoBaseIE):
+    _VALID_URL = r"http://(?:www\.)?trollvids\.com/+video/+(?P<id>[0-9]+)/+(?P<title>[^?&]+)"
```
dstftw: Use single quotes consistently.

youtube_dl/extractor/nuevo.py
nexAkari: @dstftw: I added the ignore_hd parameter because trutube's filehd always 404s (and that was mentioned in the extractor's code). The original extractor also ignored it for this reason. I removed the parameter for now and made the trutube extractor remove the bogus hd format from the list instead, but that's a little ugly. The test fails without this.

dstftw: `InfoExtractor._check_formats` should be used for validating download URLs.

youtube_dl/extractor/trollvids.py
```diff
@@ -0,0 +1,49 @@
...
+
+        video_id = match.group('id')
+        raw_video_title = match.group('title')
+        url = 'http://trollvids.com/video/%s/%s' % (video_id, raw_video_title)
+        config_url = 'http://trollvids.com/nuevo/player/config.php?v=%s' % video_id
+
+        info = self._extract_nuevo(config_url, video_id)
+
+        info.update({
+            'webpage_url': url,
```
dstftw: This field is added automatically no need to add it by hand.

youtube_dl/extractor/trollvids.py
```diff
@@ -0,0 +1,49 @@
...
+            'age_limit': 18
+        })
+
+        if 'title' not in info:
```
dstftw: Can you name any example URL that ends up in missing title in XML?

youtube_dl/extractor/nuevo.py
nexAkari: @dstftw [Well, this obviously won't work.](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L837)

```python
    ...
    
    def _check_formats(self, formats, video_id):
        if formats:
            formats[:] = filter(
                lambda f: self._is_valid_url(
                    f['url'], video_id,
                    item='%s video format' % f.get('format_id') if f.get('format_id') else 'video'),
                formats)

    def _is_valid_url(self, url, video_id, item='video'):
        url = self._proto_relative_url(url, scheme='http:')
        # For now assume non HTTP(S) URLs always valid
        if not (url.startswith('http://') or url.startswith('https://')):
            return True
    
    ...
```

dstftw:
Cherry-picked in d570746e45cff3c0f89654bf748e44a5da75a924, thanks.

PR #7731, opened 2015-12-02
Download archive update

AndroKev:


dstftw:
Download archive is not supposed to work with `--skip-download`.

AndroKev:
but maybe it should:
scenario:
you have a playlist, but only want to download the new items. so you download the playlist with --skip-download and then when you load it again only the new item are loaded( of course then without --skip-download)

dstftw:
If you plan to download new items in future you have to use download archive from the beginning. Even if you don't have a download archive but a playlist and still want to download only new items, just run the same command used for playlist downloading appending `--download-archive` and it will generate a download archive.

PR #7753, opened 2015-12-04
fix warning

programadorhedonista:
```
$ unzip -t youtube-dl
Archive:  youtube-dl
warning [youtube-dl]:  22 extra bytes at beginning or within zipfile
  (attempting to process anyway)
    testing: youtube_dl/YoutubeDL.py   OK
    testing: youtube_dl/__init__.py   OK
    testing: youtube_dl/__main__.py   OK
...
```


programadorhedonista:
Ermm... Better adding `--quiet`

dstftw:
It's not supposed to be unzipped by hand.

programadorhedonista:
It's not supposed to be unzipped by hand, yes.
But the zip offsets are incorrect, after all.

PR #7767, opened 2015-12-05
[funimation] Add new extractor

Slyneth:


youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,110 @@
+# coding: utf-8
+from __future__ import unicode_literals
+from .common import InfoExtractor
+from ..utils import (sanitized_Request, ExtractorError, urlencode_postdata)
+
+import re
+
+class FunimationBaseIE(InfoExtractor):
```
yan12125: There's no need to create a base class if there's only one derived class.

dstftw: What's the purpose of the base class if it's only inherited once?

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,110 @@
...
+        'skip': 'requires actual password',
+    }
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+
+        request = sanitized_Request(url)
+        request.add_header('User-Agent', self._USER_AGENT)
+
+        video_address = mobj.group('address')
```
dstftw: This makes no sense - you already have it in `url`.

yan12125: Seems it's exactly ```url```. Is it necessary to create a new variable?

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,110 @@
...
+        webpage = self._download_webpage(request, video_address)
```
yan12125: Usually we don't use the URL as ```video_id```. If ```video_id``` is unknown yet, it's OK to use part of the URL, for example "a-housecat-a-stray-god-and-a-tail".

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,110 @@
...
+            'vbr': 3500,
+            'resolution': '720p',
+            'url': video_base + '-3500K.mp4' + video_auth,
+        }, {
+            'format_id': 'highest',
+            'quality': 6,
+            'vbr': 4000,
+            'resolution': '1080p',
+            'url': video_base + '-4000K.mp4' + video_auth,
+        }]
```
yan12125: It's better to generate these values in a loop. Another point: are all qualities available for all videos?

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,110 @@
...
+
+        video_base = self._search_regex(r'"sdUrl":"(.+?)(?:-1500K.mp4)"', webpage, 'video_base')
+        video_id = self._search_regex(r'"FUNImationID":"(.+?)"', webpage, 'video_id')
+        video_auth = self._search_regex(r'"authToken":"(.+?)"', webpage, 'video_auth')
+
+        video_base = video_base.replace('\\', '')
+
+        formats = [{
+            'format_id': 'lowest',
+            'quality': 1,
```
yan12125: In this case ```quality``` is determined from other attributes, like the bitrate or the resolution. There's no need to specify ```quality``` values explicitly.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,110 @@
...
+        })
+        login_request = sanitized_Request(login_url, data)
+        login_request.add_header('Content-Type', 'application/x-www-form-urlencoded')
+        self._download_webpage(login_request, login_url)
+
+    def _real_initialize(self):
+        self._login()
+
+class FunimationIE(FunimationBaseIE):
+    _VALID_URL = r'(?P<address>https?://(?:www\.)?funimation\.com/shows/.*/videos/official/.+)'
```
yan12125: There should be an ```id``` group.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,110 @@
...
+            'vbr': 4000,
+            'resolution': '1080p',
+            'url': video_base + '-4000K.mp4' + video_auth,
+        }]
+        self._sort_formats(formats)
+
+        return {
+            'id': video_id,
+            'ext': 'mp4',
+            'title': self._og_search_title(webpage),
```
yan12125: There should be a fallback title if ```_og_search_title()``` returns ```None```.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,110 @@
...
+            'quality': 6,
+            'vbr': 4000,
+            'resolution': '1080p',
+            'url': video_base + '-4000K.mp4' + video_auth,
+        }]
+        self._sort_formats(formats)
+
+        return {
+            'id': video_id,
+            'ext': 'mp4',
```
yan12125: In this case video extensions can be correctly determined from URLs. There's no need to specify here.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,110 @@
+# coding: utf-8
+from __future__ import unicode_literals
+from .common import InfoExtractor
+from ..utils import (sanitized_Request, ExtractorError, urlencode_postdata)
```
yan12125: Usually we place each imported name in individual lines. For example:
```
from ..utils import (
    sanitized_Request,
    ExtractorError,
    urlencode_postdata,
)
```

youtube_dl/extractor/funimation.py
Slyneth: I tried to use taht first, but couldn't handle the regex. The url sometimes end with title-of-ep, and sometimes with title-of-ep?watch=sub and title-of-ep?watch=dub

I wasn't able to make the last part of the regex optional.

youtube_dl/extractor/funimation.py
Slyneth: Older archive is all on Hulu, and all the video of the recent seasons I checked had them. But I don't understand what do you mean by a loop.

youtube_dl/extractor/funimation.py
yan12125: Maybe ```(?P<id>[^?]+)``` works?

Slyneth: Worked, thanks.

But now video gets downloaded as Show - Title of Ep-title-of-ep.mp4.

youtube_dl/extractor/funimation.py
yan12125: For example: (I didn't test it)
```
formats_map = [
    ('lowest', 480, 1000),
    ('lower', 480, 1500),
    ...
]
formats = [{
    'format_id': format_id,
    'vbr': video_bitrate,
    ...
} for format_id, height, video_bitrate in formats_map]
```

For another problem: then could you add some checking (something like ```_search_regex``` or so) for Hulu videos and raise an ```ExtractorError```? Or we'll get misleading error messages.

youtube_dl/extractor/funimation.py
yan12125: Just assign the correct value to ```video_id``` later.

youtube_dl/extractor/funimation.py
Slyneth: How can I do that?

yan12125: An example is the last few lines of [facebook.py](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/facebook.py).

youtube_dl/extractor/funimation.py
Slyneth: Actually, it seems I was wrong about Hulu, when you login, even 10 year old videos becomes Funimation hosted. I checked 50 or so from different seasons going back 10 years. 

Only the video without login are Hulu hosted, it seems. I already have an error for not providing login information.

About the formats, all formats have different urls. I tried what you wrote but I have problems with url. I did it like this:

        formats_map = [
            ('lowest', 480, 1000),
            ('lower', 480, 1500),
            ('low', 480, 2000),
            ('medium', 720, 2500),
            ('high', 720, 3500),
            ('highest', 1080, 4000),
        ]
        formats = [{
            'format_id': format_id,
            'vbr': video_bitrate,
            'resolution': height + 'p',
            'url': video_base + '-' + video_bitrate + '.mp4',
        } for format_id, height, video_bitrate in formats_map]
```
What I am getting wrong? It says *TypeError: unsupported operand type(s) for +: 'int' and 'unicode'*.

**Comment by Slyneth** [comment]
I tried to fix everything but the formats, I am couldn't get it to work.

Also, I was wrong about Hulu, as long as you are subbed, you stream all the files directly from Funimation, and Funimation doesn't work without subbing.

File: youtube_dl/extractor/funimation.py (hunk 4 above)
Comment (yan12125) [inline youtube_dl/extractor/funimation.py]: It's not allowed in Python to add strings and integers directly. The following syntax should work:
```
'%s-%d.mp4' % (video_base, video_bitrate)
```

**Comment by Slyneth** [comment]
I have added all the required errors.

Hulu only happens when you are not logged in, I have made logging in mandatory. Geographic restrictions and wrong login information have their own errors too.

File: youtube_dl/extractor/funimation.py [hunk 10]
```diff
@@ -0,0 +1,98 @@
...
+        })
+        login_request = sanitized_Request(login_url, data)
+        login_request.add_header('Content-Type', 'application/x-www-form-urlencoded')
+
+        try:
+            login = self._download_webpage(login_request, login_url)
+        except ExtractorError as e:
+            if isinstance(e.cause, compat_HTTPError) and e.cause.code == 403:
+                raise ExtractorError('%s is now available in your region.' % self.IE_NAME, expected=True)
+                return
```
yan12125: There's no need to ```return``` after ```raise```.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,98 @@
...
+            ('low', 480, 2000),
+            ('medium', 720, 2500),
+            ('high', 720, 3500),
+            ('highest', 1080, 4000),
+        ]
+        formats = [{
+            'format_id': format_id,
+            'vbr': video_bitrate,
+            'resolution': '{}p'.format(height),
+            'url': '{}-{}K.mp4{}'.format(video_base, video_bitrate, video_auth),
```
yan12125: Such a format is not available in Python 2.6. Use ```'{0}-{1}K.mp4{2}'``` instead.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,98 @@
...
+        'skip': 'requires actual password',
+    }
+
+    def _real_initialize(self):
+        self._login()
+
+    def _login(self):
+        (username, password) = self._get_login_info()
+        if username is None:
+                raise ExtractorError('No login info available, needed for using %s.' % self.IE_NAME, expected=True)
```
dstftw: Making logging in with credentials mandatory prevents ability to authenticate with cookies.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,98 @@
...
+            'password_field': password,
+        })
+        login_request = sanitized_Request(login_url, data)
+        login_request.add_header('Content-Type', 'application/x-www-form-urlencoded')
+
+        try:
+            login = self._download_webpage(login_request, login_url)
+        except ExtractorError as e:
+            if isinstance(e.cause, compat_HTTPError) and e.cause.code == 403:
+                raise ExtractorError('%s is now available in your region.' % self.IE_NAME, expected=True)
```
yan12125: I guess it a typo? now -> not

youtube_dl/extractor/funimation.py
Slyneth: So, should I remove this and put a different error (not having learned how to do them) for trying to download not logged in?

This might actually be better as I could also give an error for subscription requirement.

Slyneth:
I found a free wo watch stream, doesn't even need login and not on Hulu. Changed the test to it. Now only errors are geographic restrictions, and Hulu videos.

Slyneth:
One more thing, I want to change preferred format to lower and print a message about that. It is the only one that all shows support, free, dub or sub subscriptions or not logged-in.

How can I do that?

Slyneth:
It is complete now, I think.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,92 @@
...
+        'params': {
+            'format': 'lower',
+        }
+    }
+
+    def _real_initialize(self):
+        self._login()
+
+    def _login(self):
+        (username, password) = self._get_login_info()
```
dstftw: You try logging in even when no username/password is specified.

dstftw:
Commits should be squashed to a single commit.

youtube_dl/extractor/funimation.py
Slyneth: How will I fix that?

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,92 @@
...
+        })
+        login_request = sanitized_Request(login_url, data)
+        login_request.add_header('Content-Type', 'application/x-www-form-urlencoded')
+
+        try:
+            login = self._download_webpage(login_request, login_url)
+        except ExtractorError as e:
+            if isinstance(e.cause, compat_HTTPError) and e.cause.code == 403:
+                raise ExtractorError('%s is not available in your region.' % self.IE_NAME, expected=True)
+
```
dstftw: You don't check whether login succeeded or not.

youtube_dl/extractor/funimation.py
dstftw: Don't try logging in when `username` is `None` obviously.

youtube_dl/extractor/funimation.py
Slyneth: Alright, fixed both. Now, how could I do this squashing? I am doing everything from web expect testing locally.

dstftw: You should consult some git manual.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,97 @@
...
+        if re.search(r'<meta property="og:url" content="http://www.funimation.com/login"/>', login) is not None:
+                raise ExtractorError('Unable to login, wrong username or password', expected=True)
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('id')
+        request = sanitized_Request(url)
+        request.add_header('User-Agent', self._USER_AGENT)
+        try:
+            webpage = self._download_webpage(request, url)
```
dstftw: Second argument should be `video_id`.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,97 @@
...
+        except ExtractorError as e:
+            if isinstance(e.cause, compat_HTTPError) and e.cause.code == 403:
+                raise ExtractorError('%s is not available in your region.' % self.IE_NAME, expected=True)
+        if re.search(r'"sdUrl":"http', webpage) is None:
+                raise ExtractorError('You are not logged-in or the stream requires subscription.', expected=True)
+        self.report_extraction(url)
```
dstftw: This line is unnecessary.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,97 @@
...
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('id')
+        request = sanitized_Request(url)
+        request.add_header('User-Agent', self._USER_AGENT)
+        try:
+            webpage = self._download_webpage(request, url)
+        except ExtractorError as e:
+            if isinstance(e.cause, compat_HTTPError) and e.cause.code == 403:
+                raise ExtractorError('%s is not available in your region.' % self.IE_NAME, expected=True)
```
dstftw: You should not silence all another exceptions but re-raise.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    def _real_initialize(self):
+        self._login()
+
+    def _login(self):
+        (username, password) = self._get_login_info()
+        if username is None:
+            return
+        self.report_login()
```
dstftw: This line is unnecessary.

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,97 @@
...
+        login_url = 'http://www.funimation.com/login'
+        data = urlencode_postdata({
+            'formname': 'loginForm2',
+            'email_field': username,
+            'password_field': password,
```
dstftw: All fields that may potentially contain non-ASCII must be [utf-8 encoded](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/vimeo.py#L43).

youtube_dl/extractor/funimation.py
Slyneth: I fixed all the others but I do not understand what do you mean by this.

Thanks for all the help by the way, I knew next to nothing 2 days ago =)

youtube_dl/extractor/funimation.py
```diff
@@ -0,0 +1,97 @@
...
+        video_base = video_base.replace('\\', '')
+
+        # lower might be better as the default format, it is the most supported.
+        formats_map = [
+            ('lowest', 480, 1000),
+            ('lower', 480, 1500),
+            ('low', 480, 2000),
+            ('medium', 720, 2500),
+            ('high', 720, 3500),
+            ('highest', 1080, 4000),
```
dstftw: `medium`, `high` and `highest` result in a webpage downloaded instead of media:
```
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'http://www.funimation.com/shows/air/videos/official/breeze', u'-f', u'highest', u'-v', u'--proxy', u'159.203.253.115:8080']
[debug] youtube-dl version 2015.12.05
[debug] Git HEAD: 7b99b5f
[debug] Python version 2.6.6 - Windows-2003Server-5.2.3790-SP2
[debug] exe versions: ffmpeg N-75573-g1d0487f, ffprobe N-75573-g1d0487f, rtmpdump 2.4
[debug] Proxy map: {u'http': u'159.203.253.115:8080', u'https': u'159.203.253.115:8080'}
[Funimation] http://www.funimation.com/shows/air/videos/official/breeze: Downloading webpage
[Funimation] http://www.funimation.com/shows/air/videos/official/breeze: Extracting information
[debug] Invoking downloader on u'http://wpc.8c48.edgecastcdn.net/008C48/SV/480/AIRENG0001/AIRENG0001-480-4000K.mp4?RGJ2KdXHZ87YZHjym4dN6OYXCAJDiGuI4QWB4gyktVzSPnQmAbKYudcnkn0mya80YoaX5BYPLd2LZeLymJGe9zsjAuZy7bZWL3o-Wx-fj1lLOTipN9wMF6aA4YaNWiIC-DXAGmKINCMYJ-3rc9x6_jEvBw'
[download] Destination: Air - Breeze-AIRENG0001.mp4
[download] 100% of 10.29KiB in 00:00
```

Slyneth: Not all formats are available for all episodes. Trying higher formats says this for me:

    [debug] System config: []
    [debug] User config: []
    [debug] Command-line args: [u'http://www.funimation.com/shows/air/videos/official/breeze', u'--format', u'high', u'--verbose']
    [debug] Encodings: locale US-ASCII, fs utf-8, out US-ASCII, pref US-ASCII
    [debug] youtube-dl version 2015.11.27.1
    [debug] Git HEAD: 4617ebb
    [debug] Python version 2.7.10 - Darwin-15.0.0-x86_64-i386-64bit
    [debug] exe versions: ffmpeg 2.8.3, ffprobe 2.8.3
    [debug] Proxy map: {}
    [Funimation] breeze: Downloading webpage
    [debug] Invoking downloader on u'http://wpc.8c48.edgecastcdn.net/008C48/SV/480/AIRENG0001/AIRENG0001-480-3500K.mp4?Z0QmTAlHnPRKQx5Lj0gl8LsPqtkUGwKoQHCdXsS0mTACzVG8qKJxMyO_Jk3LvQs_XxH1xiqPzKb6J85mlEYB_rKzJgd5-jbPhSOodUFm4IvEfhdHf_Cgb3suGBo3cXzWz7dl20zGy6k2X-6o71vlDXo'
    ERROR: unable to download video data: HTTP Error 404: Not Found

I tried to find a way to report this, but couldn't handle it.

youtube_dl/extractor/funimation.py
dstftw: Put `raise` after `if` at the same indent.

youtube_dl/extractor/funimation.py
dstftw: This can be easily detected from m3u8 that's served in `sdUrl` for default youtube-dl UA.

youtube_dl/extractor/funimation.py
Slyneth: I did this:

            try:
                webpage = self._download_webpage(request, video_id)
            except ExtractorError as e:
                if isinstance(e.cause, compat_HTTPError) and e.cause.code == 403:
                raise ExtractorError('%s is not available in your region.' % self.IE_NAME, expected=True)
            if re.search(r'"sdUrl":"http', webpage) is None:
            raise ExtractorError('You are not logged-in or the stream requires subscription.', expected=True)

And it gives this error:

        raise ExtractorError('%s is not available in your region.' % self.IE_NAME, expected=True)
            ^
    IndentationError: expected an indented block

youtube_dl/extractor/funimation.py
Slyneth: I couldn't workout m3u8 when I first started but I think I learned a very little but since then. I will try again.

youtube_dl/extractor/funimation.py
dstftw:
```python
        try:
            webpage = self._download_webpage(request, url)
        except ExtractorError as e:
            if isinstance(e.cause, compat_HTTPError) and e.cause.code == 403:
                raise ExtractorError('%s is not available in your region.' % self.IE_NAME, expected=True)
            raise
```

PR #7771, opened 2015-12-06
[novamov] simplify info extraction

remitamine:


dstftw:
1. Media downloaded with this approach is of worse quality than one of the current approach.
2. This approach does not detect some errors (e.g. "The video is being transfered"), a webpage is downloaded instead of throwing an error.

PR #7819, opened 2015-12-09
[putlocker] Add new extractor

dgoguerra:
Added extractor for [putlocker.is](http://putlocker.is/).


dstftw:

dstftw:
By checking several pages I can state that putlocker.is is clearly dedicated to copyright infringement therefore can't be supported by youtube-dl.

dgoguerra:
It does serve copyrighted content. I didn't even think about it, even though you state it very clearly in the project description. My fault, sorry!

PR #7821, opened 2015-12-09
[FFmpegPostProcessor] Default of prefer ffmpeg

joksnet:
When no `downloader` is passed to `FFmpegPostProcessor`
an exception was raised trying to get the prefer ffmpeg param.

    AttributeError: 'NoneType' object has no attribute 'params'

This fixes and defaults to `False`.


dstftw:
Thanks.

PR #7827, opened 2015-12-10
Updating README.md

habi:
As part of http://24pullrequests.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-**youtube-dl** - download videos from youtube.com or other video platforms
+youtube-dl - download videos from youtube.com or other video platforms
 
 - [INSTALLATION](#installation)
 - [DESCRIPTION](#description)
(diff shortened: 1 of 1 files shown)


dstftw:
I don't see any point emphasizing **youtube-dl** all over the places. Only important parts of text should be emphasized for catching attention.

habi:
The first time it’s mentioned unemphasised, the next two times it’s emphasised, then again not.
I thought it would be good to harmonise that.

phihag:
No, it should basically be never emphasized, except maybe once or twice at the very beginning, or if following other conventions. 67 times is at least sixty times too many.

habi:
OK, the emphasis is back to what it was before.
Only Markdown changes and added links remain.

phihag:
Thank you very much for your contribution! Merged.

habi:
Very welcome!

PR #7887, opened 2015-12-17
Update funimation.py

Slyneth:
I think not having FUNimationID in the format code would be better for automation purposes, scripts and such.


dstftw:
How is this better for automation purposes?
With this patch it's impossible to distinguish between ENG/JPN formats:
```
[info] Available formats for 31128:
http-480p-0  mp4        480p       1500k
http-480p-1  mp4        480p       1500k  (best)
```
With master:
```
[info] Available formats for 31128:
format code           extension  resolution note
HSNJPN0001-http-480p  mp4        480p       1500k
HSNENG0001-http-480p  mp4        480p       1500k  (best)
```

Slyneth:
I haven't thought of that.

With every episode having a different format name, it was hard to create an automated command to download and watch 720p (for example) without asking for formats first. I wrote a little script on my end to solve this.

By the way, that language codes aren't always correct, don't depend on them to select the right language.

PR #7894, opened 2015-12-18
Update iqiyi.py

gam2046:
Fix part of the address can not be resolved.
eg:
http://www.iqiyi.com/w_19rt6o8t9p.html
http://www.iqiyi.com/a_19rrhbc6kt.html


yan12125:
Please add test cases with ```'only_matching': True``` to ```_TESTS```.

yan12125:
Also it's better to use:
```
r'http://(?:www\.)iqiyi\.com/(?:v|w|a)_[^.]+\.html'
```
Sorry I didn't correct the regex in [the earlier commit](https://github.com/rg3/youtube-dl/commit/865ab62f43eb94a9f4f757a464df147e983cb439).

gam2046:
`r'http://(?:www\.)iqiyi\.com/(?:v|w|a)_[^.]+\.html'` is correct, but I'm not sure if there are other forms of URL.And now iqiyi.com was changed.

yan12125:
Did you mean #7896? I've updated ```iqiyi.py``` with the latest key.

gam2046:
Yes, on the one hand is the key has failed, on the other hand, I think [the earlier commit][link1] will be better. Like `http://(?:www\.)iqiyi.com/.+?\.html`

Even the URL may be like this `http://yule.iqiyi.com/pcb.html`, so if you only need to check the URL of the host name?

Regular expressions may be like this `http://[^.]+?\.iqiyi.com/.+?\.html`

[link1]:https://github.com/rg3/youtube-dl/commit/865ab62f43eb94a9f4f757a464df147e983cb439

yan12125:
Good example, so it should be
```
http://(?:[^.]+\.)iqiyi\.com/.+\.html
```
The following URLs are matched:

http://www.iqiyi.com/v_19rrojlavg.html
http://www.iqiyi.com/a_19rrhbc6kt.html
http://iqiyi.com/a_19rrhbc6kt.html
http://yule.iqiyi.com/pcb.html

But not matching:
http://www.notiqiyi.com/123.html
http://iqiyi-com/123.html

gam2046:
`http://(?:[^.]+\.)iqiyi\.com/.+\.html` is not matched `http://iqiyi.com/a_19rrhbc6kt.html`, may be like this `http:/(?:.*?)(/|\.)iqiyi\.com/.+\.html`. 

I don't understand what you mean. Why do you want to match those addresses?(like `http://www.notiqiyi.com/123.html`) Those are not legal addresses.

gam2046:
I'm not python, I don't know about Python support for regular expressions, may not support `. *?`, can be replaced by `.{0,999}`

yan12125:
Sorry there's a typo:
```
http://(?:[^.]+\.)?iqiyi\.com/.+\.html
```

yan12125:
Thanks for the contribution, cherry-picked at 82e02ea5fce6bbc595dbfefc8eafce21e2bb53cc and bee83e84f64248a47f2f0b79e8c2698bc89abf55 and add some tests. Github does not send notifications when there are new commits in a pull request, so please leave a comment the next time. And by the way, it's usually not recommended to include merge commits in pull requests.

Would you like to be listed in [AUTHORS](https://github.com/rg3/youtube-dl/blob/master/AUTHORS)? If so, under what name?

gam2046:
No thanks. I believe that if it is someone else, it will do so. The spirit of open source should be like this, I am very grateful to you to write such a useful program.

Thanks again.

PR #7921, opened 2015-12-20
[pandoratv] Add new extractor (closes #6884)

### LINKED ISSUE #6884: add support for pandora.tv
http://channel.pandora.tv/channel/video.ptv?ch_userid=jump7580&prgid=36253542&categid=33146188&page=1&ref=ch&lot=cthum2_1_2

bit:


youtube_dl/extractor/__init__.py
```diff
@@ -473,6 +473,7 @@
     ORFFM4IE,
     ORFIPTVIE,
 )
+from .pandora_tv import PandoraTVIE
```
dstftw: We don't use underscore in extractor file names.

youtube_dl/extractor/pandora_tv.py
```diff
@@ -0,0 +1,40 @@
...
+    qualities,
+)
+
+
+class PandoraTVIE(InfoExtractor):
+    _VALID_URL = r'http://channel.pandora.tv/channel/video.ptv\?ch_userid=(?P<userid>.*?)&prgid=(?P<prgid>.*?)&.*'
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('userid') + '_' + mobj.group('prgid')
```
dstftw: User id is not related to video id hence shouldn't be part of `video_id`.
You shouldn't rely on query arguments order, `http://channel.pandora.tv/channel/video.ptv?prgid=...&ch_userid=...` should be matched as well.

youtube_dl/extractor/pandoratv.py
```diff
@@ -0,0 +1,40 @@
...
+    compat_parse_qs,
+    compat_urllib_parse_urlparse,
+)
+from ..utils import (
+    find_xpath_attr,
+    unified_strdate,
+    get_element_by_attribute,
+    int_or_none,
+    qualities,
+)
```
dstftw: All these import are unused, check your code with flake8.

youtube_dl/extractor/pandoratv.py
```diff
@@ -0,0 +1,40 @@
+# encoding: utf-8
```
dstftw: There must be [`from __future__ import unicode_literals` on top](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site).

youtube_dl/extractor/pandoratv.py
```diff
@@ -0,0 +1,40 @@
...
+    unified_strdate,
+    get_element_by_attribute,
+    int_or_none,
+    qualities,
+)
+
+
+class PandoraTVIE(InfoExtractor):
+    _VALID_URL = r'http://channel.pandora.tv/channel/video.ptv\?ch_userid=(?P<userid>.*?)&prgid=(?P<prgid>.*?)&.*'
+
```
dstftw: Tests are missing.

youtube_dl/extractor/pandoratv.py
```diff
@@ -0,0 +1,40 @@
...
+    find_xpath_attr,
+    unified_strdate,
+    get_element_by_attribute,
+    int_or_none,
+    qualities,
+)
+
+
+class PandoraTVIE(InfoExtractor):
+    _VALID_URL = r'http://channel.pandora.tv/channel/video.ptv\?ch_userid=(?P<userid>.*?)&prgid=(?P<prgid>.*?)&.*'
```
dstftw: Subdomains are not matched: http://jp.channel.pandora.tv/channel/video.ptv?c1=&prgid=53294230&ch_userid=mikakim&ref=main&lot=cate_01_2.

youtube_dl/extractor/pandoratv.py
```diff
@@ -0,0 +1,40 @@
...
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('userid') + '_' + mobj.group('prgid')
+        webpage = self._download_webpage(url, video_id)
+        video_set = self._html_search_regex(
+                r'var videoSet = ({.*?});',
+                webpage, 'video set')
+        video_set = json.loads(video_set)
+        info_url ="http://channel.pandora.tv/channel/video.ptv?m=getSmartData&ch_userid={userid}&prgid={prgid}".format(**video_set)
+        info = self._download_webpage(info_url, None)
+        info = json.loads(info[1:-1])
```
dstftw: Use `self._download_json` instead.

bit: how would _download_json now that its a jsonp request that needs to be unwrapped(info[1:-1])?

dstftw: `transform_source`.

bit: never mind, found transform_source...

youtube_dl/extractor/pandoratv.py
```diff
@@ -0,0 +1,37 @@
...
+    }]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('prgid')
+        webpage = self._download_webpage(url, video_id)
+        video_set = self._html_search_regex(
+                r'var videoSet = ({.*?});',
+                webpage, 'video set')
+        video_set = json.loads(video_set)
```
dstftw: Use `self._parse_json` instead.

youtube_dl/extractor/pandoratv.py
```diff
@@ -0,0 +1,37 @@
+# encoding: utf-8
+from __future__ import unicode_literals
+
+import re
+import json
+
+from .common import InfoExtractor
+
+class PandoraTVIE(InfoExtractor):
+    _VALID_URL = r'http://(.*?\.)?channel.pandora.tv/channel/video.ptv\?.*?prgid=(?P<prgid>.*?)&.*'
```
dstftw: `ch_userid` is mandatory, without it 404 is returned, e.g. http://channel.pandora.tv/channel/video.ptv?prgid=53294230&ref=main&lot=cate_01_2 shouldn't be matched.
`http://(.*?\.)?` should be `http://(?:.+?\.)?`.
`(?P<prgid>.*?)` should be `(?P<prgid>.+?)`.
`.` representing dots should be `\.`.

dstftw: [Correct way](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/pluralsight.py#L95-L100) to implement order independent argument matching.

youtube_dl/extractor/pandoratv.py
```diff
@@ -0,0 +1,47 @@
...
+        user_id = qs.get('ch_userid', [None])[0]
+        if any(not f for f in (video_id, user_id,)):
+            raise ExtractorError('Invalid URL', expected=True)
+
+        webpage = self._download_webpage(url, video_id)
+        video_set = self._html_search_regex(
+                r'var videoSet = ({.*?});',
+                webpage, 'video set')
+        video_set = self._parse_json(video_set, None)
+        info_url = "http://channel.pandora.tv/channel/video.ptv?m=getSmartData&ch_userid={userid}&prgid={prgid}".format(**video_set)
```
dstftw: You've already extracted `prgid` and `ch_userid` from URL, you should not additionally rely on these fields in `video_set`.

dstftw: Also always use single quotes whenever possible.

youtube_dl/extractor/pandoratv.py
```diff
@@ -0,0 +1,47 @@
...
+from ..compat import (
+    compat_urlparse,
+)
+from ..utils import (
+    ExtractorError,
+)
+
+
+class PandoraTVIE(InfoExtractor):
+    _VALID_URL = r'http://(?:.+?\.)?channel.pandora.tv/channel/video.ptv\?.*?prgid=(?P<prgid>.+?)&.*'
```
dstftw: Part after `\?` should be removed since it's not used anymore.

youtube_dl/extractor/pandoratv.py
```diff
@@ -0,0 +1,47 @@
...
+        if any(not f for f in (video_id, user_id,)):
+            raise ExtractorError('Invalid URL', expected=True)
+
+        webpage = self._download_webpage(url, video_id)
+        video_set = self._html_search_regex(
+                r'var videoSet = ({.*?});',
+                webpage, 'video set')
+        video_set = self._parse_json(video_set, video_id)
+        info_url = 'http://channel.pandora.tv/channel/video.ptv?m=getSmartData&ch_userid={userid}&prgid={prgid}'.format(userid=user_id, prgid=video_id)
+        info = self._download_json(info_url, None, transform_source=lambda info: info[1:-1])
```
dstftw: Second argument should be `video_id`.

dstftw:
According to webplayer there are more video quality formats avaiable, all should be extracted. More metadata can be also extracted from the `webpage` like upload date, view count and so on.

bit:
no idea how to access other resolutions. right now using the html5 lookup which only has one.

remitamine:
the flash player use this url `http://channel.pandora.tv/video/php/php.player.query.php?ch_userid=?&prgid=?` to extract video info but it require to get the keys manualy.
the mobile site use this url `http://m.pandora.tv/?c=view&m=viewJsonApi&ch_userid=?&prgid=?` to extract info it contain all the information(no need for encryption).

dstftw:
Cherry-picked in 9accfed4e724d3048e894c2323b3a478951d670d, thanks.

PR #7968, opened 2015-12-23
niconico: Support authentication with cookie

jlhg:
https://github.com/rg3/youtube-dl/issues/6003#issuecomment-113364979
I also got an incorrect thumbPlayKey in some machines, and don't know why.
As a workaround, the cookie file can be applied for authentication alternative.


yan12125:
@jlhg How niconico.py handles authentication is quite different now, so I rewrite your codes. Thanks for the idea and the first implementation!

PR #7991, opened 2015-12-23
[gogoanime] Add new extractor

l29ah:


remitamine:
youtube-dl does not include support for services that specialize in infringing copyright.
https://github.com/rg3/youtube-dl#can-you-add-support-for-this-anime-video-site-or-site-which-shows-current-movies-for-free

PR #7992, opened 2015-12-23
[ceskatelevize] Work with old embed URLs

che0:
Extended CeskaTelevize extractor to work with embed URLs (for some old shows these are the only available).


che0:
Nope, those URLs are not permanent, so the test only passes for a short time :(

PR #7994, opened 2015-12-24
[Weiqitv] Add new extractor

FounderSG:


youtube_dl/extractor/weiqitv.py
```diff
@@ -0,0 +1,89 @@
...
+        info_json_str = self._html_search_regex(
+            'var video = (.*});',
+            page, 'info_json_str', default=None)
+        info_json = self._parse_json(info_json_str, media_id)
+
+        play_json_param = self._html_search_regex(
+            r'letvurl.*?bcloud.html\?([^"]+)',
+            page, 'play_json_param', default=None)
+        play_json_req = sanitized_Request(
+            'http://api.letvcloud.com/gpc.php?cf=html5&sign=signxxxxx&ver=2.2&format=json&' +
```
yan12125: This site is using the service provided by Letv. There should be standalone ```LetvCloudIE```. Example URLs for ```LetvCloudIE``` can be http://yuntv.letv.com/bcloud.html?uu=p7jnfw5hw9&vu=467623dedf.

FounderSG: do you mean after find out letvurl then delegate to LetvCloudIE for the rest of work?
the LetvCloudIE should be in letv.py.

yan12125: Yep.

FounderSG: can you show me example how to delegate to another extractor? 
i use media_info = LetvCloudIE()._real_extract(letvcloud_url)
it is working, but unit test test_download failed.

yan12125: You need [```url_transparent```](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L232-L237). For example:
```
return {
    '_type': 'url_transparent',
    'ie_key': 'LetvCloud',
    'description': description,
    ...
}
```

FounderSG: i have done the changes and updated pull request. please review the new code.

youtube_dl/extractor/letv.py
```diff
@@ -239,3 +240,75 @@ class LetvPlaylistIE(LetvTvIE):
         },
         'playlist_mincount': 7
     }]
+
+
+class LetvCloudIE(InfoExtractor):
+    IE_DESC = 'LetvCloud'
+    _VALID_URL = r'http://yuntv\.letv\.com/bcloud.html\?(?P<id>.*)$'
```
yan12125: The group ```id``` is not for IDs but URL parameters. This can be misleading, and I suggest a different name.

youtube_dl/extractor/letv.py
```diff
@@ -239,3 +240,75 @@ class LetvPlaylistIE(LetvTvIE):
...
+            'timestamp': 1445946428,
+            'upload_date': '20151027',
+        },
+    }]
+
+    def _real_extract(self, url):
+        media_params = self._match_id(url)
+        uu = re.search("uu=([\w]+)", media_params).group(1)
+        vu = re.search("vu=([\w]+)", media_params).group(1)
+        media_id = uu + vu
```
yan12125: I guess ```p7jnfw5hw9_187060b6fd``` or ```p7jnfw5hw9-187060b6fd``` is better than ```p7jnfw5hw9187060b6fd``` for readability.

youtube_dl/extractor/letv.py
```diff
@@ -239,3 +240,75 @@ class LetvPlaylistIE(LetvTvIE):
...
+
+        play_json_req = sanitized_Request(
+            'http://api.letvcloud.com/gpc.php?cf=html5&sign=signxxxxx&ver=2.2&format=json&' +
+            "uu=" + uu + "&vu=" + vu)
+        play_json = self._download_json(play_json_req, media_id, 'Downloading playJson data')
+
+        urls = []
+        for v in play_json['data']['video_info']['media'].values():
+            url_info_dict = {
+                'url': base64.b64decode(v['play_url']['main_url']).decode("utf-8"),
```
yan12125: This will fail in Python 3.2 as old ```base64.b64decode``` requires bytestrings. You need:
```
base64.b64decode(v['play_url']['main_url'].encode('utf-8')).decode('utf-8')
```

youtube_dl/extractor/letv.py
```diff
@@ -239,3 +240,75 @@ class LetvPlaylistIE(LetvTvIE):
...
+                'ext': 'mp4',
```
yan12125: Are videos always mp4?

youtube_dl/extractor/letv.py
```diff
@@ -239,3 +240,75 @@ class LetvPlaylistIE(LetvTvIE):
...
+                'format_id': int(v['play_url']['vtype']),
+                'format_note': v['play_url']['definition'],
+                'width': int(v['play_url']['vwidth']),
+                'height': int(v['play_url']['vheight']),
```
yan12125: Some fields (for example width, height, format_note, etc.) are optional. Invalid or missing values should be gated. For example:
```
int_or_none(v.get('play_url', {}).get('vheight'))
```

youtube_dl/extractor/letv.py
```diff
@@ -239,3 +240,75 @@ class LetvPlaylistIE(LetvTvIE):
...
+            }
+            urls.append(url_info_dict)
+
+        publish_time = parse_iso8601(
+            play_json['data']['video_info']['add_time'],
+            delimiter=' ', timezone=datetime.timedelta(hours=8))
```
yan12125: Likely, ```timestamp``` is optional, too. See https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L64-L201 for more information.

youtube_dl/extractor/letv.py
```diff
@@ -239,3 +240,75 @@ class LetvPlaylistIE(LetvTvIE):
...
+        vu = re.search("vu=([\w]+)", media_params).group(1)
+        media_id = uu + vu
+
+        play_json_req = sanitized_Request(
+            'http://api.letvcloud.com/gpc.php?cf=html5&sign=signxxxxx&ver=2.2&format=json&' +
+            "uu=" + uu + "&vu=" + vu)
+        play_json = self._download_json(play_json_req, media_id, 'Downloading playJson data')
+
+        urls = []
+        for v in play_json['data']['video_info']['media'].values():
```
yan12125: Here using list comprehension simplifies the code:
```
urls = [{
    'url': ...
    ...
} for v in play_json['data']['video_info']['media'].values()]
```

And by the way, there should be a meaningful name instead of ```v```.

youtube_dl/extractor/letv.py
```diff
@@ -239,3 +240,75 @@ class LetvPlaylistIE(LetvTvIE):
...
+            'ext': 'mp4',
+            'title': 'p7jnfw5hw9187060b6fd',
+            'timestamp': 1445946428,
+            'upload_date': '20151027',
+        },
+    }]
+
+    def _real_extract(self, url):
+        media_params = self._match_id(url)
+        uu = re.search("uu=([\w]+)", media_params).group(1)
```
yan12125: We use single quotes as much as possible.

youtube_dl/extractor/letv.py
```diff
@@ -239,3 +240,75 @@ class LetvPlaylistIE(LetvTvIE):
         },
         'playlist_mincount': 7
     }]
+
+
+class LetvCloudIE(InfoExtractor):
+    IE_DESC = 'LetvCloud'
```
yan12125: Information extractor description should be a more meaningful name. You can even use non-English names, for example ```乐视云```.

youtube_dl/extractor/letv.py
```diff
@@ -239,3 +240,75 @@ class LetvPlaylistIE(LetvTvIE):
...
+            urls.append(url_info_dict)
+
+        publish_time = parse_iso8601(
+            play_json['data']['video_info']['add_time'],
+            delimiter=' ', timezone=datetime.timedelta(hours=8))
+
+        return {
+            'id': media_id,
+            'title': media_id,
+            'formats': urls,
```
yan12125: Don't forget to call ```self._sort_formats()```, or the formats are not sorted:
```
$ youtube-dl -F test:letvcloud 
[TestURL] Test URL: http://yuntv.letv.com/bcloud.html?uu=p7jnfw5hw9&vu=467623dedf
[LetvCloud] p7jnfw5hw9467623dedf: Downloading playJson data
[info] Available formats for p7jnfw5hw9467623dedf:
52           mp4        1920x1072  1080p 
13           mp4        960x544    高清 
51           mp4        1280x720   720P 
28           mp4        1280x720   原画 
22           mp4        1280x720   超清 
21           mp4        640x352    标清  (best)
```

youtube_dl/extractor/letv.py
```diff
@@ -239,3 +240,75 @@ class LetvPlaylistIE(LetvTvIE):
...
+        'md5': '26450599afd64c513bc77030ad15db44',
+        'info_dict': {
+            'id': 'p7jnfw5hw9467623dedf',
+            'ext': 'mp4',
+            'title': 'p7jnfw5hw9467623dedf',
+            'timestamp': 1445922586,
+            'upload_date': '20151027',
+        },
+        'params': {
+            'format': '[width=1920]',
```
yan12125: I guess it's not necessary after ```_sort_formats()```?

youtube_dl/extractor/weiqitv.py
```diff
@@ -0,0 +1,63 @@
...
+            'upload_date': '20151027',
+        },
+    }]
+
+    def _real_extract(self, url):
+        media_id = self._match_id(url)
+        page = self._download_webpage(url, media_id)
+
+        info_json_str = self._html_search_regex(
+            'var video = (.*});',
```
yan12125: Not tested, but ```var\s+video\s*=\s*([^;]+);``` may be better.

youtube_dl/extractor/weiqitv.py
```diff
@@ -0,0 +1,63 @@
...
+            'timestamp': 1445946428,
+            'upload_date': '20151027',
+        },
+    }]
+
+    def _real_extract(self, url):
+        media_id = self._match_id(url)
+        page = self._download_webpage(url, media_id)
+
+        info_json_str = self._html_search_regex(
```
yan12125: It's better to use ```_search_regex``` for JavaScript codes instead of ```_html_search_regex``` as the latter is designed for HTML markups.

youtube_dl/extractor/weiqitv.py
```diff
@@ -0,0 +1,63 @@
...
+            'var video = (.*});',
+            page, 'info_json_str', default=None)
```
yan12125: In ```_html_search_regex``` (as well as ```_search_regex```), the default value for default is ```None```, so there's no need to specify it explicitly.

youtube_dl/extractor/letv.py
FounderSG: how about name it as urlparams?

youtube_dl/extractor/letv.py
FounderSG: sounds valid, i choose p7jnfw5hw9_187060b6fd

youtube_dl/extractor/letv.py
FounderSG: this is something i don't know, i will fix it as you advised in next PR.

youtube_dl/extractor/letv.py
FounderSG: there are two api, one for mp4 and one for flv, i use the one for mp4, all videos are mp4.

youtube_dl/extractor/letv.py
FounderSG: I will fix it as you advised in next PR.

youtube_dl/extractor/letv.py
FounderSG: for simplicity, i will remove publish_time. i can add it again if somebody ask for it.

youtube_dl/extractor/letv.py
FounderSG: how about name it as media?

youtube_dl/extractor/letv.py
FounderSG: I will fix it as you advised in next PR.

youtube_dl/extractor/letv.py
FounderSG: I will fix it as you advised in next PR.

youtube_dl/extractor/letv.py
FounderSG: the format code returned by letvcloud looks wild, e.g., 13 is actually better than 21, is it necessary to have a remap?

youtube_dl/extractor/letv.py
FounderSG: 1920x1080 have the format id 52, which is the largest, test pass even without format params and _sort_formats.

youtube_dl/extractor/weiqitv.py
FounderSG: (.*}); match string end with }; 
([^;]+); match string end with ;
if there are ; inside json variable, then ([^;]+); can not catch the whole variable.

youtube_dl/extractor/weiqitv.py
FounderSG: this is something i don't know, i will fix it as you advised in next PR.

youtube_dl/extractor/weiqitv.py
FounderSG: I will fix it as you advised in next PR.

youtube_dl/extractor/letv.py
yan12125: It's good.

youtube_dl/extractor/letv.py
yan12125: If there are two APIs, it's better to include both APIs and let users to choose from flv and mp4.

youtube_dl/extractor/letv.py
yan12125: If such information is available, it's better to include it in the extractor.

youtube_dl/extractor/letv.py
yan12125: Good idea.

youtube_dl/extractor/letv.py
yan12125: I guess format codes are some denotions not related to the quality. ```_sort_formats()``` will sort the formats by resolutions, bitrates, etc., so it would not be a problem. If formats are not sorted, users may download videos with worse qualities. For example in the case I've mentioned previously, 640x352 is downloaded instead the 1920x1072 one.

youtube_dl/extractor/letv.py
FounderSG: _VALID_URL = r'http://yuntv\.letv\.com/bcloud.html\?.*$'
is it ok to set _VALID_URL without id group?

youtube_dl/extractor/letv.py
yan12125: FYI: ```_sort_formats``` uses multiple criteria for sorting, and ```format_id``` is the last one.
https://github.com/rg3/youtube-dl/blob/bd3f9ecabeb5e189f16580c61c56d93148079363/youtube_dl/extractor/common.py#L810-L826

youtube_dl/extractor/weiqitv.py
yan12125: Then maybe ```var\s+video\s*=\s*(.+});```?

youtube_dl/extractor/letv.py
yan12125: It's OK to have no ```id``` group for complicated cases, such as query parameters. [Here's another example](https://github.com/rg3/youtube-dl/blob/bd3f9ecabeb5e189f16580c61c56d93148079363/youtube_dl/extractor/youtube.py#L1940). However, usually it's better to have some named groups.

youtube_dl/extractor/letv.py
FounderSG: _sort_formats has been added, please review updated PR.

youtube_dl/extractor/letv.py
FounderSG: i don't have official flv api document, somebody reverse engineering letvcloud flash player and post to internet, http://tieba.baidu.com/p/3636578571, but the information maybe outdated, it doesn't work on my test.

youtube_dl/extractor/letv.py
FounderSG: i misunderstand format_id, params removed in updated PR.

youtube_dl/extractor/weiqitv.py
FounderSG: var\s+video\s*=\s*(.+}); works fine.

FounderSG:
what should i do to get my pull request accept?

yan12125:
Sorry, but Github does not send notifications if there are only code updates. The next time please leave a comment after codes done.
Would you like to be listed in [AUTHORS](https://github.com/rg3/youtube-dl/blob/master/AUTHORS) with the name Founder Fang?

FounderSG:
yes, Founder Fang is my name, it's my pleasure to be listed in AUTHORS.

yan12125:
Done in 56aa074538afc4941b437e2d91a3e298d5f20eb9.

PR #7995, opened 2015-12-24
[zdf] expand valid-url pattern for channels

Boris-de:
The webpage also creates URLs which include additional text that defines
the sorting order on the page like "aktuellste" (most current) and
"meist-gesehen" (most seen), e.g.:

http://www.zdf.de/ZDFmediathek/kanaluebersicht/aktuellste/332
http://www.zdf.de/ZDFmediathek/kanaluebersicht/meist-gesehen/332


dstftw:
Cherry-picked in e41604227ce14ded0b87385a17490bfa65d35fde, thanks.

PR #8001, opened 2015-12-24
Fix format selection issue.

nepella:
I came across [a YouTube video](http://youtu.be/D09YfuAkNag) for which a lower-resolution video format (134, 480×360 vs. 135, 640×480) is chosen by `bestvideo+bestaudio`. That’s not what we want: in addition to having a lower resolution, the video has a lower bitrate (by ffmpeg) and is visually inferior. The issue is caused by YouTube incorrectly reporting a higher total bitrate for the lower-resolution format. Putting height and width before bitrate in `_sort_formats` fixes this.

[Here’s](https://github.com/rg3/youtube-dl/files/71848/format-listing.txt) the format listing for the video (without my patch).


dstftw:
This may break sorting for some another extractor that returns invalid height/width but correct bitrate. I would prefer not to change the generic sorting behaviour but rather tweak it on per extractor basis.

nepella:
Like so? It seems kludgy to duplicate _sort_formats, but I can’t think of a better way to do it.

remitamine:
> Like so? It seems kludgy to duplicate _sort_formats, but I can’t think of a better way to do it.

you can use something like this https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/vimeo.py#L444.

nepella:
Oh! thank you. How's this?

dstftw:
There is no such fields: `ext_preference`, `audio_ext_preference`. This breaks `--prefer-free-formats`.

nepella:
I am no longer able to repeat this issue, even against c24044635b6e7f67274899e7c5e7a5efcce947af (the base for this pull request): I see that the available formats ([reported by `youtube-dl -F`](<https://github.com/rg3/youtube-dl/files/2231965/format-listing-2.txt>) has changed and is in a different order.

PR #8018, opened 2015-12-27
Update README.md

Azulan:
Add explanation and workaround FAQ for YouTube SSL certificate error on Windows OS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -531,6 +531,10 @@ By default, youtube-dl intends to have the best options (incidentally, if you ha
 
 Most people asking this question are not aware that youtube-dl now defaults to downloading the highest available quality as reported by YouTube, which will be 1080p or 720p in some cases, so you no longer need the `-b` option. For some specific videos, maybe YouTube does not report them to be available in a specific high quality format you're interested in. In that case, simply request it with the `-f` option and youtube-dl will try to download it.
 
+### I get a [SSL: CERTIFICATE_VERIFY_FAILED] error when trying to download a YouTube page on my Windows computer. How do I fix this?
+
+Your computer likely has an expired Google security certificate. Go to https://pki.google.com and download the "Google Internet Authority G2" certificate file. Double click the downloaded file to install the certificate. You may need to restart your computer. The issue should be resolved.
+
 ### I get HTTP error 402 when trying to download a video. What's this?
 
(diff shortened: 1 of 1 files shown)


yan12125:
"Google Internet Authority G2" (serial number 02:3A:83) is not a root CA, but "GeoTrust Global CA" (serial number 02:34:56) is. Users should check GeoTrust Global CA instead of downloading Google's certificates.

Azulan:
@yan12125 But it is Google's cert that was expired.

yan12125:
That's strange. When Python tries to connect to ```www.youtube.com:443```, youtube.com returns a certificate chain with 3 items: *.google.com (with *.youtube.com in subject alternative names), Google Internet Authority G2 and GeoTrust Global CA. Normal devices have only valid GeoTrust Global CA, so the whole certificate chain is trusted. Have you installed Google Internet Authority G2 certificate before?

yan12125:
By https://github.com/rg3/youtube-dl/issues/7309#issuecomment-167414048, I think it's better to suggest users to remove expired Google Internet Authority G2 certificates.

Azulan:
@yan12125 Perhaps, but do you really want non-technical users trying to poke around in the cert store and tell them to delete something? Mistakes could be disastrous. The offered solution is far simpler and less risky, I think.

PotcFdk:
@Azulan Poking around in the cert store without knowing what you're doing is always a bad idea. If you add the wrong cert (note: it's not easy to tell what's a "good" and what's a "bad" certificate - even if you're technically competent), it could lead to all sorts of disasters.  
The one expired Google Internet Authority G2 certificate that I mentioned [here](https://github.com/rg3/youtube-dl/issues/7309#issuecomment-167414048) expired in April 2015, but I don't remember having experienced these issues that long ago. Thus, I'm not sure whether something added these useless certificates to the system recently or if previous versions of youtube-dl/python/whatever just did things in a different way.

yan12125:
@Azulan your concern is possible. It would be better to have detailed instructions and warnings in boldface rather than a simple sentence "remove the expired Google Internet Authority G2 in certmgr.msc". You can add the steps to a file in [docs/](https://github.com/rg3/youtube-dl/tree/master/docs) as it might be too long to put all the things in FAQ. Of course users have rights to trust the downloaded certificate based on others' opinions, but it's inproper to suggest a wrong solution in ```README.md```.

PR #8023, opened 2015-12-27
[common] simplify the use of _extract_m3u8_formats and _extract_f4m_formats

remitamine:
instead of:
```
m3u8_formats = self._extract_m3u8_formats(m3u8_url, video_id, 'mp4', m3u8_id='hls', fatal=False)
if m3u8_formats:
    formats.extend(m3u8_formats)
f4m_formats = self._extract_f4m_formats(f4m_url, video_id, f4m_id='hds', fatal=False)
if f4m_formats:
    formats.extend(f4m_formats)
```
just use:
```
formats.extend(self._extract_m3u8_formats(m3u8_url, video_id, 'mp4', m3u8_id='hls', fatal=False))
formats.extend(self._extract_f4m_formats(f4m_url, video_id, f4m_id='hds', fatal=False))
```
and may be even making mp4 as the default ext and hls as the default id for m3u8_formats
and hds as the default id for f4m_formats
if this acceptable i will change all the parts that use this two methods in the other extractors.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -885,7 +885,7 @@ class InfoExtractor(object):
             fatal=fatal)
 
         if manifest is False:
-            return manifest
+            return []
 
         formats = []
         manifest_version = '1.0'
@@ -956,7 +956,7 @@ class InfoExtractor(object):
             errnote=errnote or 'Failed to download m3u8 information',
             fatal=fatal)
         if res is False:
-            return res
+            return []
         m3u8_doc, urlh = res
         m3u8_url = urlh.geturl()
         last_info = None


jaimeMF:
It think it's a reasonable change.

About setting `m3u8_id` and `f4m_id`: if a website only provides m3u8 or f4m formats the prefix is not needed so I don't know if we should set it (although it's not really important since it's better to use format filters instead of directly using the ids).

dstftw:
LGTM.

remitamine:
> if this acceptable i will change all the parts that use this two methods in the other extractors.

i will just merge it because if i modify every use if this methods it will take a long time(there is more than 400 file that use this method) and just keep this change when there is a modification of an extractor or for the new extrators.

dstftw:
It's fairly easy to deal with most common cases with this rough snippet:
```python
import glob
import re

for filename in glob.glob('./youtube_dl/extractor/*.py'):
    with open(filename, 'r+') as f:
        src = f.read().decode('utf-8')
        src_tmp = re.sub(
            r'(?s)([a-zA-Z_][\da-zA-Z_]*) = (self\._extract_(?:f4m|m3u8|smil)_formats\(.+?\))\s*if \1:\s*([a-zA-Z_][\da-zA-Z_]*)\.extend\(\1\)',
            r'\3.extend(\2)', src)
        if src == src_tmp:
            continue
        f.seek(0)
        f.truncate()
        f.write(src_tmp.encode('utf-8'))
```

remitamine:
if it is safe to apply this modification than it will a lot better than changing every file manualy.

dstftw:
Obviously you should look through changed files before committing. Visually I've found no issues applying it. I can commit it myself if you want.

remitamine:
> I can commit it myself if you want.

yes and thanks for the help.

Issue #7850, opened 2015-12-12
Site Support Request: pornorama.com


UnveilTech:
Hi All,
Possible to pump Pornorama ?

./youtube-dl http://www.pornorama.com/video1416287/0/amy_my_neighbor_fucked
[generic] amy_my_neighbor_fucked: Requesting header
[redirect] Following redirect to http://www.pornorama.com/video1416287/amy_my_neighbor_fucked
[generic] amy_my_neighbor_fucked: Requesting header
WARNING: Falling back on generic information extractor.
[generic] amy_my_neighbor_fucked: Downloading webpage
[generic] amy_my_neighbor_fucked: Extracting information
ERROR: Unsupported URL: http://www.pornorama.com/video1416287/amy_my_neighbor_fucked

PR #8041, opened 2015-12-29
[Pornorama extractor]  support site: http://www.pornorama.com/

Celthi:
#7850

### DIFF
--- a/youtube_dl/extractor/pornorama.py
+++ b/youtube_dl/extractor/pornorama.py
@@ -27,10 +27,6 @@ class PornoramaIE(InfoExtractor):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
 
-        mobj = re.search(r'<h1 class="inlineError">(.+?)</h1>', webpage)
-        if mobj:
-            raise ExtractorError('%s said: %s' % (self.IE_NAME, clean_html(mobj.group(1))), expected=True)
-
         video_url = compat_urllib_parse_unquote(
             self._search_regex(r'flv_url=(.+?)&', webpage, 'video URL'))
         video_title = self._html_search_regex(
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/pornorama.py
```diff
@@ -0,0 +1,50 @@
...
+from ..utils import (
+    clean_html,
+    ExtractorError
+)
+
+
+class PornoramaIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?pornorama\.com/video(?P<id>[0-9]+)(?:.*)'
+    _TEST = {
+        'url': 'http://www.pornorama.com/video1416287/amy_my_neighbor_fucked',
```
dstftw: This looks like copyrighted content. youtube-dl does not support sites dedicated to [copyright infringement](https://github.com/rg3/youtube-dl#can-you-add-support-for-this-anime-video-site-or-site-which-shows-current-movies-for-free). Do they have rights to distribute this? If not provide [an example URL](https://github.com/rg3/youtube-dl/blob/master/README.md#is-the-description-of-the-issue-itself-sufficient) that does not violate any copyrights.

Hrxn:
Why do you believe that?

This is just a front-end for a video on xvideos?

Celthi:
@Hrxn yeah, you are right. At first I thought this is the reason that the issue #7850 is remaining open,  piece of case to implement it. I do this for coding the simple one since I am a rookie, only for learning purpose.

youtube_dl/extractor/pornorama.py
Celthi: @dstftw  I found  the open site support request #7850 and it is as @Hrxn  says. So I thought I may give it a try. The example URL is in the #7850. If you feel it looks like copyrighted content, please reject this pull request.

Hrxn:
It's always a good idea to learn this stuff, I can completely understand and support your decision! :+1:

Celthi:
@Hrxn Thank you.

PR #8051, opened 2015-12-29
[common] fail only if the extractor didn't return formats

remitamine:
format extraction will fail if one of _extract_*_formats methods didn't return any formats this is not desirable in extractors that has more than one type of formats.
this change make the error happen only if the extractor didn't return formats.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1208,8 +1208,6 @@ class YoutubeDL(object):
             raise ExtractorError('Missing "id" field in extractor result')
         if 'title' not in info_dict:
             raise ExtractorError('Missing "title" field in extractor result')
-        if not info_dict.get('formats'):
-            raise ExtractorError('No video formats found')
 
         if 'playlist' not in info_dict:
             # It isn't part of a playlist
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1208,6 +1208,8 @@ class YoutubeDL(object):
             raise ExtractorError('Missing "id" field in extractor result')
         if 'title' not in info_dict:
             raise ExtractorError('Missing "title" field in extractor result')
+        if not info_dict.get('formats'):
+            raise ExtractorError('No video formats found')
 
         if 'playlist' not in info_dict:
             # It isn't part of a playlist
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -765,7 +765,7 @@ class InfoExtractor(object):
 
     def _sort_formats(self, formats, field_preference=None):
         if not formats:
-            raise ExtractorError('No video formats found')
+            return
 
         def _formats_key(f):
             # TODO remove the following workaround


dstftw:
This will prevent the ability to quit early during the extraction and half the rest of the extraction in cases when `_sort_formats` is called by extractors.

remitamine:
> This will prevent the ability to quit early during the extraction and halt the rest of the extraction in cases when _sort_formats is called by extractors.

can you explain why this wil happen and if this won't work is there a better way to did it.

dstftw:
The most reasonable way to write an extractor is to extract formats in the first place and stop extraction early if no formats available:
```
def _real_extract(self, url):
    ...
    # First extract formats
    formats = make_formats(...)
    self._sort_formats(formats)  # If no formats here extraction should be stopped
                                 # since there is no sense to continue
    ...
    # Extract title, if no title - stop extraction
    ...
    # The rest of the extraction goes here. There can be some
    # heavy network I/O fetching some additional metadata and
    # so on that would be just time waste when no formats or title
    # available.
    ...
    return info_dict
```
Currently, `_sort_formats` is used as this stop mechanism, so if this check is removed the remaining part of extraction code will be always executed even when no formats have been extracted in the first step.

dstftw:
The least tedious way would be to add `allow_empty_formats=False` flag to `_sort_formats` and set it on when called from `_extract_*_formats` methods.

remitamine:
i follwed the flow of execution and i find this close to the end of the extraction:
_real_extract -> extract -> https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L665 -> https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L676

remitamine:
also i think not every extractor call the sort_formats method.

youtube_dl/YoutubeDL.py
```diff
@@ -1208,6 +1208,8 @@ def process_video_result(self, info_dict, download=True):
             raise ExtractorError('Missing "id" field in extractor result')
         if 'title' not in info_dict:
             raise ExtractorError('Missing "title" field in extractor result')
+        if not info_dict.get('formats'):
```
jaimeMF: Note that the `formats` field is optional, the `url` can be set in the main dict. For example the appleconnect test URL (https://itunes.apple.com/us/post/idsa.4ab17a39-2720-11e5-96c5-a5b38f6c42d3) fails.

remitamine:
when i read the rest of the method i find that there is already a check for formats existence https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L1271.

dstftw:
A real world example of possible empty formats list is #8573: f4m formats are DRM protected, it's reasonable to filter them out in `_extract_f4m_formats` but in this case we end up with an empty list passed to `_sort_formats`.
Since this problem becomes more actual now, I'd like to bump this PR and drop some ideas on how to resolve this better:
1. Stick with current approach of this PR and inspect all the places in code where we will need to explicitly add checks for early quit (see above).
2. Remove `_sort_formats` from `_extract_*_formats` methods completely and inspect all the places in code where we will need to insert calls to `_sort_formats`.
3. Add a boolean flag `sort` with default value `True` to all the `_extract_*_formats` routines and set it to false when empty formats list is allowed.
I find #2 to be most rational since in 90-95% of extraction cases (according to our code base) multiple (f4m/m3u8/smil/...) formats are extracted and it makes no sense to do sorting in between. It also more transparent since it decouples the responsibility: `_extract_*_formats` should extract formats not sort them.
On the other hand #3 will require less changes and won't require any code inspection but it will bring a clutter in future code when we will need to pass `sort=False` in extract routines.
Would like to hear another opinions.

remitamine:
> I find #2 to be most rational since in 90-95% of extraction cases (according to our code base) multiple (f4m/m3u8/smil/...) formats are extracted and it makes no sense to do sorting in between. It also more transparent since it decouples the responsibility: _extract_*_formats should extract formats not sort them.

i agree,#2 is the best solution even if it mean more work.

```
    # First extract formats
    formats = make_formats(...)
    self._sort_formats(formats)  # If no formats here extraction should be stopped
                                 # since there is no sense to continue
    ...
    # Extract title, if no title - stop extraction
```
isn't better to extract the title before the formats(most of the time extracting the title didn't require more requests than the webpage or the main json/xml info).

remitamine:
applying `#2` will require at least adding _sort_formats to this extractors
```
lrt.py
dfb.py
dplay.py
discovery.py
laola1tv.py
abc7news.py
telecinco.py
rtve.py
videomore.py
chaturbate.py
generic.py
cwtv.py
vier.py
ynet.py
animeondemand.py
sportbox.py
funnyordie.py
mitele.py
shahid.py
restudy.py
bet.py
matchtv.py
rtvnh.py
tubitv.py
```

dstftw:
So any other opinions? I guess we should adopt #2. We may also add a flag `sort=False` to all `_extract_*_formats` for convenience.

dstftw:
>isn't better to extract the title before the formats(most of the time extracting the title didn't require more requests than the webpage or the main json/xml info)

Yes, it makes sense when title does not require additional network I/O but it's not always the case.

Issue #7858, opened 2015-12-13
Site request: espnfc


RingoTheDog:
ESPN Football site...I think should use the same extractor as ESPN, just need the URL to be recognized.

URL: http://www.espnfc.us/video/mls-highlights/150/video/2743663/must-see-moments-best-of-the-mls-season
OUTPUT: 
youtube-dl -v "http://www.espnfc.us/video/mls-highlights/150/video/2743663/m
ust-see-moments-best-of-the-mls-season"
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://www.espnfc.us/video/mls-highlights/1
50/video/2743663/must-see-moments-best-of-the-mls-season']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2015.12.06
[debug] Python version 2.7.10 - Windows-7-6.1.7601-SP1
[debug] exe versions: ffmpeg N-71346-gdf4fca2
[debug] Proxy map: {}
[generic] must-see-moments-best-of-the-mls-season: Requesting header
WARNING: Falling back on generic information extractor.
[generic] must-see-moments-best-of-the-mls-season: Downloading webpage
[generic] must-see-moments-best-of-the-mls-season: Extracting information
[debug] Invoking downloader on u'https://sw88.go.com/b/ss/wdgespvideo/5.4/REDIR/
031615570333296051449504376310?D=..&url=https://once.unicornmedia.com/now/od/aut
o/07355f66-a73e-46c2-885b-e769dac7c817/20879ce0-ad85-4acf-b0b4-f4a91a6c1517/ac99
a7e5-7c12-4327-919f-a1a18597cd52/content.once?UMADPARAMreferer=http://espnfc.com
/video/espnfc/video?id=2743663&pe=media&mediaName=SOCCER::FCUS_MLS::+Must+See+Mo
ments:+Best+of+the+MLS+season_2743663&mediaLength=00:03:31&mediaPlayer=embed09&m
ediaSession=0-0&c3=videoid=SOCCER::FCUS_MLS::+Must+See+Moments:+Best+of+the+MLS+
season_2743663&v2=videoid=SOCCER::FCUS_MLS::+Must+See+Moments:+Best+of+the+MLS+s
eason_2743663&events=event1&v52=inlinePlayer&c52=inlinePlayer&v53=inlinePlayer:t
witter&c53=inlinePlayer:twitter'

[...]

PR #8053, opened 2015-12-30
[ESPNFC] Add new extractor

davidjameshowell:
This is my first commit and pull request, go easy on me!

Per #7858, created a new extractor for ESPNFC. I decided to keep this separate from the ESPN extractor due how the links are formatted.  I decided to use the Once auto link since it will return an MP4/stitched for browsers that do not support native HLS.

I added some additional comments to #7858 after realizing that we can just swap the ID of the video in with "intl:<videoID>" and use the existing extractor.

Let me know what the best path would be (as well as suggestions for the next extractor with this commit).

### DIFF
--- a/youtube_dl/extractor/__init__.py
+++ b/youtube_dl/extractor/__init__.py
@@ -184,6 +184,7 @@ from .eporner import EpornerIE
 from .eroprofile import EroProfileIE
 from .escapist import EscapistIE
 from .espn import ESPNIE
+from .espnfc import ESPNFCIE
 from .esri import EsriVideoIE
 from .europa import EuropaIE
 from .everyonesmixtape import EveryonesMixtapeIE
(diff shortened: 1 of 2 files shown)


jaimeMF:
Actually you don't need the `intl:` prefix, `http://espn.go.com/video/clip?id=<id>` seems to work fine on a browser.

I think it's better to just redirect to that url, but the espn.go.com extractor is unfinished (7e760fc18897663db7c0717434e28a8cca9f3810) and it doesn't work for me.

davidjameshowell:
Alright. I think I'll work on the base on itself. Again, sorry for amateurish, example/advice on validating two different urls schemes or would this be a base to redirect to the ESPN?

dstftw:
espn extractor now works fine due to changes to ooyala made some time inbetween.

davidjameshowell:
Advice on how to bake it into the ESPN extractor?

yan12125:

yan12125:
Domain names like www.espnfc.com should be included as well.

Issue #7954, opened 2015-12-23
Feature request: Add extractor for cultureunplugged.com


sjjh:
Looking at the supported sites http://www.cultureunplugged.com/ is not supported yet. I would love if an extractor could be added for it.

````
simon@laptop:~$ youtube-dl -v "http://www.cultureunplugged.com/documentary/watch-online/play/9096/Laula"
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://www.cultureunplugged.com/documentary/watch-online/play/9096/Laula']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.06.04.1
[debug] Python version 2.7.10 - Linux-4.2.0-22-generic-x86_64-with-Ubuntu-15.10-wily
[debug] exe versions: ffmpeg 2.7.3-0ubuntu0.15.10.1, ffprobe 2.7.3-0ubuntu0.15.10.1, rtmpdump 2.4
[debug] Proxy map: {}
[generic] Laula: Requesting header
WARNING: Falling back on generic information extractor.
[generic] Laula: Downloading webpage
[generic] Laula: Extracting information
ERROR: Unsupported URL: http://www.cultureunplugged.com/documentary/watch-online/play/9096/Laula
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/youtube_dl/extractor/generic.py", line 997, in _real_extract
    doc = parse_xml(webpage)
  File "/usr/lib/python2.7/dist-packages/youtube_dl/utils.py", line 1558, in parse_xml
    tree = xml.etree.ElementTree.XML(s.encode('utf-8'), **kwargs)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1300, in XML
    parser.feed(text)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1642, in feed
    self._raiseerror(v)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1506, in _raiseerror
    raise err
ParseError: undefined entity: line 5, column 1
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/youtube_dl/YoutubeDL.py", line 650, in extract_info
    ie_result = ie.extract(url)
  File "/usr/lib/python2.7/dist-packages/youtube_dl/extractor/common.py", line 273, in extract
    return self._real_extract(url)
[...]

PR #8060, opened 2015-12-30
[CultureUnplugged] Add new extractor

davidjameshowell:
For #7954, adding extractor for CultureUnplugged. May add additional metadata in the future, but testing looks good.

### DIFF
--- a/youtube_dl/extractor/cultureunplugged.py
+++ b/youtube_dl/extractor/cultureunplugged.py
@@ -1,0 +1,48 @@
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+
+
+class CultureUnpluggedIE(InfoExtractor):
+    _VALID_URL = r'http?://(?:www\.)?cultureunplugged\.com/documentary\/watch-online\/play\/(?P<id>\d+)\/(?P<display_id>[^/]+)'
+    _TEST = {
+        'url': 'http://www.cultureunplugged.com/documentary/watch-online/play/53662/The-Next--Best-West',
+        'md5': 'ac6c093b089f7d05e79934dcb3d228fc',
+        'info_dict': {
+            'id': '53662',
+            'display_id': 'The-Next--Best-West',
+            'description': 'The Next, Best West explores our changing relationship with the land that sustains us. It tells the story of how the conventional American concept of progress has steered our exploitation of the Western landscape, and takes you to three places – Colorado’s San Luis Valley, the high plains of eastern Montana and the Elwha River on Washington’s Olympic Peninsula – where a vibrant new understanding of progress presages a better future. The Next, Best West shows how our interpretation of progress has shaped the singular landscape of the American West, and how a new understanding of progress may be our best hope for a bright and healthy future. The West is a place of pure beauty that has provided us so much, yet we have cared for it too little. But that is beginning to change.',
+            'ext': 'mp4',
+            'title': 'The Next, Best West',
+            'thumbnail': 'http://cdn.cultureunplugged.com/thumbnails_16_9/lg/53662.jpg',
+        }
+    }
+
+    def _real_extract(self, url):
+
+        video_id = self._match_id(url)
+        display_id = re.match(self._VALID_URL, url).group('display_id')
+        json_url = 'http://www.cultureunplugged.com/movie-data/cu-%s.json' % (video_id)
+        json_output = self._download_json(json_url, video_id)
+
+        title = json_output['title']
+        description = json_output['synopsis']
+        creator = json_output['producer']
+        thumbnail = json_output['large_thumb']
+
+        formats = [{
+            'url': json_output['url'],
+            'format': 'mp4'
+            }]
+
+        return {
+            'id': video_id,
+            'title': title,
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/cultureunplugged.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def _real_extract(self, url):
+
+        video_id = self._match_id(url)
+        display_id = re.match(self._VALID_URL, url).group('display_id')
+        json_url = 'http://www.cultureunplugged.com/movie-data/cu-%s.json' % (video_id)
+        json_output = self._download_json(json_url, video_id)
+
+        title = json_output['title']
+        description = json_output['synopsis']
```
dstftw: All fields apart from `title` and `formats` should be treated optional and extraction must not fail when any of these fields are missing.

davidjameshowell: Forgive my ignorance - what's the best approach in this instance (for the future)? Is there a method in commons for this type of check?

dstftw: `json_output.get('synopsis')`. You can find best practices in any recently modified extractor.

youtube_dl/extractor/cultureunplugged.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+
+
+class CultureUnpluggedIE(InfoExtractor):
+    _VALID_URL = r'http?://(?:www\.)?cultureunplugged\.com/documentary\/watch-online\/play\/(?P<id>\d+)\/(?P<display_id>[^/]+)'
```
dstftw: No need to escape forward slash. Should be `https?`. Part after `(?P<id>\d+)` should be optional since `id` is only used for extraction.

davidjameshowell: Valid point for the slashes. Still refining regex. It doesn't appear the server returns HTTPS, but no harm in adding it. 

I added the display ID as another optional ID, which mimics the title. Should this be extracted outside of the valid URL?

dstftw: I mean it should not be `http?`. `display_id` should be extracted from the `url` if present or may be extracted from further downloaded pages when does not present in the `url`.

davidjameshowell: I've gone through a handful of the websites and it appears that the component after the ID correlates to the display name slugged out. I believe it is safe to say this is a constant and is included with each playable video; their video slug structure.

davidjameshowell:
Made some changes. I used Charles to rewrite and intercept the JSON response to check for null values and all passed successfully (just entered null as value in the JSON file). How are we looking?

youtube_dl/extractor/cultureunplugged.py
```diff
@@ -0,0 +1,48 @@
...
+    _TEST = {
+        'url': 'http://www.cultureunplugged.com/documentary/watch-online/play/53662/The-Next--Best-West',
+        'md5': 'ac6c093b089f7d05e79934dcb3d228fc',
+        'info_dict': {
+            'id': '53662',
+            'display_id': 'The-Next--Best-West',
+            'description': 'The Next, Best West explores our changing relationship with the land that sustains us. It tells the story of how the conventional American concept of progress has steered our exploitation of the Western landscape, and takes you to three places – Colorado’s San Luis Valley, the high plains of eastern Montana and the Elwha River on Washington’s Olympic Peninsula – where a vibrant new understanding of progress presages a better future. The Next, Best West shows how our interpretation of progress has shaped the singular landscape of the American West, and how a new understanding of progress may be our best hope for a bright and healthy future. The West is a place of pure beauty that has provided us so much, yet we have cared for it too little. But that is beginning to change.',
```
dstftw: Can you replace it with `'md5:hash'`?

youtube_dl/extractor/cultureunplugged.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+
+
+class CultureUnpluggedIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?cultureunplugged\.com/documentary/watch-online/play/(?P<id>\d+)/(?P<display_id>[^/]+)'
```
dstftw: You didn't make `display_id` part option. http://www.cultureunplugged.com/documentary/watch-online/play/53662 URLs works just fine in browser.

youtube_dl/extractor/cultureunplugged.py
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    def _real_extract(self, url):
+
+        video_id = self._match_id(url)
+        display_id = re.match(self._VALID_URL, url).group('display_id')
+        json_url = 'http://www.cultureunplugged.com/movie-data/cu-%s.json' % (video_id)
+        json_output = self._download_json(json_url, video_id)
+
+        title = json_output.get('title')
```
dstftw: `title` must be mandatory I've already told about this.

Issue #4793, opened 2015-01-28
download udemy videos own section folder 


muhasturk:
Hello, I would like to download udemy videos inside own section folders. Youtube-dl downloads all videos inside same folder.

https://yadi.sk/i/p8KHU35GeHSCt

greatcityness:
You may download udemy videos with Allavsoft.

It also download videos from MySpace, Google Video, Yahoo Video, Break.com and other more that 750+ video websites.

yan12125:
@muhasturk https://yadi.sk/i/p8KHU35GeHSCt is gone. ("Nothing found") If you have another example, please propose it here and I'll reopen the issue.

dstftw:
The original request was to be able to put downloads in separate folders by course chapters, e.g. https://www.udemy.com/java-tutorial/#/lecture/172757
```
~/Downloads/Chapter 1 - Introduction/
    Lecture 1 - Introduction and Installation
    Lecture 2 - What Java Is and How It Works
    Lecture 3 - Getting a Job and What to Study After Completing Basic Java
    ....
~/Chapter 2 - Programming Core Java
    Lecture 6 - A Hello World Program
    ...
```

greatcityness:
It works very good.

dstftw:
This feature will be supported in the next version via output template.
Example: `youtube-dl -o '%(chapter_number)s - %(chapter)s/%(title)s.%(ext)s' -u user -p pass https://www.udemy.com/java-tutorial/`.

PR #8061, opened 2015-12-30
Introduce chapter and series fields

dstftw:
This PR proposes some new optional fields to video info dictionary in order to cover the gap for videos organized to logical chapters and video episodes belonging to season of a series or a programme.

Field | Usage | Description
------------ | ------------- | -------------
`chapter` | chapter-only | Name or title of the chapter the video belongs to.
`chapter_number` | chapter-only | Number of the chapter the video belongs to, as an integer.
`chapter_id` | chapter-only | Id of the chapter the video belongs to, as a unicode string.
`series` | series-only | Title of the series or programme the video episode belongs to.
`season` | series-only | Title of the season the video episode belongs to.
`season_number` | series-only | Number of the season the video episode belongs to, as an integer.
`season_id` | series-only | Id of the season the video episode belongs to, as a unicode string.
`episode` | series-only | Title of the video episode. Unlike mandatory video `title` field, this field should denote the exact title of the video episode without any kind of decoration.
`episode_number` | series-only | Number of the video episode within a season, as an integer.
`episode_id` | series-only | Id of the video episode, as a unicode string.

Notes:
* `*_id` fields are not `*_index` (i.e. numeric) in order to tolerate some potential exotic naming schemas apart from autoincremented numbering.
* `episode` is supposed to be potentially different from `title` and should denote the exact name of the video episode while `title` is allowed to be decorated and imposes less strictness (e.g. can contain the title displayed on a website that is not necessarily should match episode name).

Pros:
* Collecting and providing this potentially useful information.
* An elegant way to solve issues like #4793 with help of output templates.

Cons:
* Duplicated fields for videos.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -200,6 +200,26 @@ class InfoExtractor(object):
     end_time:       Time in seconds where the reproduction should end, as
                     specified in the URL.
 
+    The following fields should only be used when the video belongs to some logical
+    chapter or section:
+
+    chapter:        Name or title of the chapter the video belongs to.
+    chapter_id:     Number or id of the chapter the video belongs to, as an integer
+                    or unicode string.
+
+    The following fields should only be used when the video is an episode of some
+    series or programme:
+
+    series:         Title of the series or programme the video episode belongs to.
+    season:         Title of the season the video episode belongs to.
+    season_id:      Number or id of the season the video episode belongs to, as an
+                    integer or unicode string.
+    episode:        Title of the video episode. Unlike mandatory video title field,
+                    this field should denote the exact title of the video episode
+                    without any kind of decoration.
+    episode_id:     Number or id of the video episode within a season, as an integer
+                    or unicode string.
+
     Unless mentioned otherwise, the fields should be Unicode strings.
 
     Unless mentioned otherwise, None is equivalent to absence of information.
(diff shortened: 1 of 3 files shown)


dstftw:
I'm going to merge this if no objections or field naming suggesstions.

jaimeMF:
Why do `*_id` fields accept integers? Wouldn't it be more consistent to just use unicode strings? This could be problematic if we allow in the future to specify the number of digits for numeric fields in the output template, it would cause `TypeError`s  if it's a string.

dstftw:
Accepting only unicode strings would involve boilerplate `compat_str` wrapping in extractors. As for output template, one obviously should not use numeric formatters for fields that can potentially be non-numeric, at the same time `%s` forks just fine for numeric fields.

dstftw:
Alternatively `*_index` fields may also be introduced with numeric semantics.

jaimeMF:
> Accepting only unicode strings would involve boilerplate compat_str wrapping in extractors since most of the time these fields are originally integers.

It's not that common but it also happens with the `id` field: sometimes we have to convert from an integer (usually coming from a JSON file). Since we can't guarantee that every site uses numeric episode ids I think it's more reasonable to just use a single type.

> As for output template, one obviously should not use numeric formatters for fields that can potentially be non-numeric, at the same time %s forks just fine for numeric fields.

You are right, but some people like to use 0 for padding numbers in filenames and they will eventually ask it. So your suggestion of using `*_index` sound good.

By the way: are you aware of any supported site that use non-numeric ids?

dstftw:
I remember some used string identifiers for seasons but it's more like unique identifiers rather that an indexes.

dstftw:
Added `*_number` fields.

dstftw:
I'm merging this. Further suggestions and improvements to be committed right in master.

PR #8063, opened 2015-12-30
[extractor/rte.py] Add support for RTE radio player

bpfoley:
While here, correct RteIE filename extension to .flv as this is the
container format F4mFD outputs.

### DIFF
--- a/youtube_dl/extractor/rte.py
+++ b/youtube_dl/extractor/rte.py
@@ -5,16 +5,19 @@ from .common import InfoExtractor
 
 from ..utils import (
     float_or_none,
+    unescapeHTML,
 )
 
 
 class RteIE(InfoExtractor):
+    IE_NAME = 'rte'
+    IE_DESC = 'Raidió Teilifís Éireann TV'
     _VALID_URL = r'https?://(?:www\.)?rte\.ie/player/[^/]{2,3}/show/[^/]+/(?P<id>[0-9]+)'
     _TEST = {
         'url': 'http://www.rte.ie/player/ie/show/iwitness-862/10478715/',
         'info_dict': {
             'id': '10478715',
-            'ext': 'mp4',
+            'ext': 'flv',
             'title': 'Watch iWitness  online',
             'thumbnail': 're:^https?://.*\.jpg$',
             'description': 'iWitness : The spirit of Ireland, one voice and one minute at a time.',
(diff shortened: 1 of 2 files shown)


docs/supportedsites.md
```diff
@@ -463,7 +463,7 @@
  - **RottenTomatoes**
  - **Roxwel**
  - **RTBF**
- - **Rte**
+ - **RTE** Raidió Teilifís Éireann TV and radio
```

bpfoley:
Extractor descriptions moved class constants.

youtube_dl/extractor/rte.py
```diff
@@ -60,3 +62,56 @@ def _real_extract(self, url):
...
+
+class RteRadioIE(InfoExtractor):
+    IE_NAME = 'rte.ie'
+    IE_DESC = 'Raidió Teilifís Éireann Radio'
+    _VALID_URL = (
+        r'https?://(?:www\.)?rte\.ie/radio/utils/radioplayer/rteradioweb\.html'
+        r'#!rii=(?P<channel_id>[0-9]+)(?:%3A|:)'
+        r'(?P<id>[0-9]+)(?:%3A|:)'
+        r'(?P<pia>[0-9]+)(?:%3A|:)'
+        r'(?P<date>[0-9]{2}(%2D|-)[0-9]{2}(%2D|-)[0-9]{4})(?:%3A|:)'
```
dstftw: Groups apart from `id` are not used, no need to capture them.

youtube_dl/extractor/rte.py
```diff
@@ -60,3 +62,56 @@ def _real_extract(self, url):
...
+        }
+    }
+
+    def _real_extract(self, url):
+        item_id = self._match_id(url)
+        feeds_url = 'http://www.rte.ie/rteavgen/getplaylist/?type=web&format=json&id=' + item_id
+        json_string = self._download_json(feeds_url, item_id)
+
+        show = json_string['shows'][0]
+        title, description, thumbnail = show['title'], show['description'], show['thumbnail']
```
dstftw: Everything apart from `title` should be optional.

youtube_dl/extractor/rte.py
```diff
@@ -60,3 +62,56 @@ def _real_extract(self, url):
...
+        duration = float_or_none(show['duration'], 1000)
+
+        f4m_url = show['media:group'][0]['hds_server'] + show['media:group'][0]['hds_url']
```
dstftw: hls and rtmp are available as well.

youtube_dl/extractor/rte.py
```diff
@@ -60,3 +62,56 @@ def _real_extract(self, url):
...
+        f4m_formats = self._extract_f4m_formats(f4m_url, item_id)
+        f4m_formats = [{
+            'format_id': f['format_id'],
+            'url': f['url'],
+            'ext': 'flv',
+        } for f in f4m_formats]
```
dstftw: What's the point of lines 104-108, ext is already flv.

youtube_dl/extractor/rte.py
bpfoley: Done.

youtube_dl/extractor/rte.py
bpfoley: Done.

youtube_dl/extractor/rte.py
bpfoley: It doesn't appear to work that way. When I leave out the 'ext' item the extension of the file written to disk is .f4m.

dstftw: Can't be. `_extract_f4m_formats` always [returns flv](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L941).

bpfoley: Ah. I just understood what I did wrong. Nevermind! The perils of copy-and-paste coding.

youtube_dl/extractor/rte.py
```diff
@@ -60,3 +56,71 @@ def _real_extract(self, url):
...
+
+class RteRadioIE(InfoExtractor):
+    IE_NAME = 'rte.ie'
+    IE_DESC = 'Raidió Teilifís Éireann Radio'
+    _VALID_URL = r'''(?x)
+        https?://(?:www\.)?rte\.ie/radio/utils/radioplayer/rteradioweb\.html
+        [#]!rii=(?:[0-9]+)(?:%3A|:)  # Channel ID
+        (?P<id>[0-9]+)(?:%3A|:)      # Item ID
+        (?:[0-9]+)(?:%3A|:)          # Playable item (ie program) ID
+        (?:[0-9]{2}(%2D|-)[0-9]{2}(%2D|-)[0-9]{4})(?:%3A|:) # Start date
```
dstftw: Everything after `id` may be skipped and not matched since http://www.rte.ie/radio/utils/radioplayer/rteradioweb.html#!rii=16:10507902 works just fine.

youtube_dl/extractor/rte.py
```diff
@@ -60,3 +56,71 @@ def _real_extract(self, url):
...
         }
+
+
+
+class RteRadioIE(InfoExtractor):
+    IE_NAME = 'rte.ie'
+    IE_DESC = 'Raidió Teilifís Éireann Radio'
+    _VALID_URL = r'''(?x)
+        https?://(?:www\.)?rte\.ie/radio/utils/radioplayer/rteradioweb\.html
+        [#]!rii=(?:[0-9]+)(?:%3A|:)  # Channel ID
```
dstftw: There is no need in `[]` around `#`.

bpfoley: ACK. I'll escape it instead.

dstftw: No need to be escaped either.

bpfoley: It's a verbose regexp, so `#`s have to be escaped, otherwise they look like the start of a comment.

dstftw: I guess it's better to put it on single line since it's much shorter now.

bpfoley: OK. Generally, for documentation purposes, I prefer to split non-trivial regexps into fragments that can be individually commented, but I'm not going to bikeshed over it.

By the way, do you know if there's any way to delete the now-redundant travis-ci jobs #8130 and #8132 before they run?

dstftw: This regex is fairly straightforward and `# Channel ID` comment practically does not bring much benefit since it's not used anywhere. [Here](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/youtube.py#L248-L277) is the non-trivial one.
There is a cancel build X button on the right hand side https://travis-ci.org/rg3/youtube-dl/builds/100002208 but not sure if it's available for you. Why bother at all?

youtube_dl/extractor/rte.py
```diff
@@ -60,3 +56,68 @@ def _real_extract(self, url):
             'thumbnail': thumbnail,
             'duration': duration,
         }
+
+
+
+class RteRadioIE(InfoExtractor):
+    IE_NAME = 'rte.ie'
```
dstftw: Extractors should have unique names.

bpfoley: Done.

youtube_dl/extractor/rte.py
bpfoley: Yeah, looks like I just don't have permission for this project. I was just hoping to avoid a bunch of pointless tests building up in the build queue. They take quite a while to run.

dstftw:
Thanks.

bpfoley:
And thanks for all the feedback.

PR #8090, opened 2015-12-31
[bbccouk] Skip tests for unavailable content

bpfoley:


dstftw:
These test should be replaced with working ones instead of just skipping especially when there are no problems finding.

PR #8091, opened 2015-12-31
[youtube] Update some tests to match Youtube output

bpfoley:
test_YoutubeShow: Show URLs now have the suffix /playlists

test_YoutubeUser: User playlists have the title 'Uploads from Foo'
                  and produce a playlist ID.


dstftw:
Show id should not have /playlists suffix.

PR #6811, opened 2015-09-09
[extractor/generic]  add full support for HTML5 media tags(extract single and multiple formats, subtitles, poster...)

remitamine:
before:
```
youtube-dl -F http://www.dota2.com/reborn/part1/
[generic] part1: Requesting header
WARNING: Falling back on generic information extractor.
[generic] part1: Downloading webpage
[generic] part1: Extracting information
[download] Downloading playlist: None
[generic] playlist None: Collected 4 video ids (downloading 4 of them)
[download] Downloading video 1 of 4
[info] Available formats for AllNewDashboard:
format code  extension  resolution note
0            webm       unknown    
[download] Downloading video 2 of 4
[info] Available formats for RedesignedHeroBrowser:
format code  extension  resolution note
0            webm       unknown    
[download] Downloading video 3 of 4
[info] Available formats for NewWaysToWatch:
format code  extension  resolution note
0            webm       unknown    
[download] Downloading video 4 of 4
[info] Available formats for RevampedLearningExperience:
format code  extension  resolution note
0            webm       unknown
```
after:
```
youtube-dl -F http://www.dota2.com/reborn/part1/
[generic] part1: Requesting header
WARNING: Falling back on generic information extractor.
[generic] part1: Downloading webpage
[generic] part1: Extracting information
[download] Downloading playlist: None
[generic] playlist None: Collected 4 video ids (downloading 4 of them)
[download] Downloading video 1 of 4
[info] Available formats for :
format code  extension  resolution note
0            webm       unknown    
1            mp4        unknown    (best)
[download] Downloading video 2 of 4
[info] Available formats for :
format code  extension  resolution note
0            webm       unknown    
1            mp4        unknown    (best)
[download] Downloading video 3 of 4
[info] Available formats for :
format code  extension  resolution note
0            webm       unknown    
1            mp4        unknown    (best)
[download] Downloading video 4 of 4
[info] Available formats for :
format code  extension  resolution note
0            webm       unknown    
1            mp4        unknown    (best)
```

PR #8092, opened 2016-01-01
[twitter] Fix thumbnail matching for Twitter.

bpfoley:
Optional poster group and greedy matching caused poster="..."
to be dropped. This makes test_Twitter_1 pass.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -28,6 +28,7 @@ from youtube_dl.utils import (
     encodeFilename,
     escape_rfc3986,
     escape_url,
+    extract_attributes,
     ExtractorError,
     find_xpath_attr,
     fix_xml_ampersands,
@@ -75,6 +76,7 @@ from youtube_dl.utils import (
     cli_bool_option,
 )
 from youtube_dl.compat import (
+    compat_chr,
     compat_etree_fromstring,
 )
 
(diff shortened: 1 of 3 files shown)


dstftw:
Poster part must be optional.

bpfoley:
Hmm. The problem seems to be in the r"[^>]+"  subexpressions between each element. Since they're greedy matches and the capture groups are optional, the greedy matches end consuming more of the input than intended.

For the test case, Twitter now produces output like 
`<video class="animated-gif" data-media-id="123" data-height="4" data-width="5" poster="http://poster.png" loop>`

If we add capture groups for the greedy matches in the regex, we can see the problem
 ```'''(?x)
  <video(?P<GAP1>[^>]+)class="animated-gif"(?P<GAP2>[^>]+)
      (?:data-height="(?P<height>\d+)")?(?P<GAP3>[^>]+)
      (?:data-width="(?P<width>\d+)")?(?P<GAP4>[^>]+)
      (?:poster="(?P<poster>[^"]+)")?(?P<GAP5>[^>]*)>\s*'''
```
The output of `mobj.dictgroups()` looks like
```
{
	'GAP1': ' ',
	'GAP2': ' data-media-id="123" data-height="4" data-width="5" poster="http://poster.png" l',
	'height': None,
	'GAP3': 'o',
	'width': None,
	'GAP4': 'o',
	'poster': None,
	'GAP5': 'p'
}
```
That is, the GAP2 group is consuming most of the string because data-width, data-height and poster are all optional. I'll look at this further and see if I can come up with some more robust solution.

**Comment by dstftw** [comment]
Making gaps non-greedy produces the same effect, what's the point of this overcomplication?

**Comment by bpfoley** [comment]
Technically, yes, making the gaps non-greedy works, but it's still fragile. For example, the video element on the test twitter page now contains the markup `<video .. data-height="" data-width=""  ... >` (note the empty values) and because the width and height patterns in the regexp expect one or more digits, this breaks and produces
```
{
 'GAP1': ' ',
 'GAP2': ' ',
 'height': None,
 'GAP3': 'd',
 'width': None
 'GAP4': 'a',
 'poster': None,
 'GAP5': 'ta-height="" data-width="" poster="http://poster.png" loop',
}
```

Also the regexp will break if the order of the attributes changes for some reason.

I can just make the regexp work for the current twitter markup if you think that's the best thing to do, but I thought `_html_attrs_as_dict()` might be a more useful and easier to read solution that could be used elsewhere.

**Comment by remitamine** [comment]
> but I thought _html_attrs_as_dict() might be a more useful and easier to read solution that could be used elsewhere.

i already proposed something like this in https://github.com/remitamine/youtube-dl/blob/9550ca506fccf9c9d795816cc0a7817ff262ef45/youtube_dl/utils.py#L251

**Comment by dstftw** [comment]
If you want to provide generic attribute extraction routine it should be placed in utils and well covered with tests. Note that it's much more complex than your current implementation: http://www.w3.org/TR/html5/syntax.html#attributes-0.

**Comment by bpfoley** [comment]
Oh awesome. I see your `extract_attributes` regexp handles whitespace around the = which mine doesn't, but yours will do the wrong thing if there's a quote in the value eg `val="doesn't work"` or `val='A "quoted" value'`

**Comment by remitamine** [comment]
i created it the first time to work with brighcove in page embeds and it was working without a problem.

**Comment by bpfoley** [comment]
dstftw: I agree, doing full HTML/HTML5 parsing is non-trivial, and I deliberately avoided handling unicode, entities, unquoted values and all the rest. If we're going to go down this route, I wonder would it be worth just offloading it to Python's [HTMLParser](https://docs.python.org/2/library/htmlparser.html)?
```python
try:
    from html.parser import HTMLParser
except ImportError:  # Python 2
    from HTMLParser import HTMLParser

class HTMLAttributeParser(HTMLParser):
    def __init__(self):
        HTMLParser.__init__(self)
    def handle_starttag(self, tag, attrs):
        self.attrs = dict(attrs)

def extract_attributes(html_element):
    parser = HTMLAttributeParser()
    parser.feed(html_element)
    return parser.attrs
```

dstftw:
@bpfoley that's actually preferable but I didn't test how it's behave on semi-valid HTML and whether it supports quoteless attributes and so on.

bpfoley:
It handles hex entities, XML entities (ie `&quot;` and friends), and HTML 3.2 & 4.0 named entities (eg `&pound;` and `&lambda;`), as well as quoteless attributes. Looking at [test_htmlparser.py](https://github.com/python/cpython/blob/2.7/Lib/test/test_htmlparser.py#L411), it does appear to try to handle a lot of malformed cases too, so it sounds like this might be our solution.

bpfoley:
@dstfw OK, after a bit of trial and error `test_twitter_1` and the new `test_extract_attributes` pass for all 6 builds. Does this look good to you?

dstftw:
The curious thing is that `HTMLParser` has been used before but was abandoned in favour of regexes (see 3828505646a58d5e19f0ffe77821650e25d72b3a). It's unclear what was the rationale. Pinging @phihag.

dstftw:
Ping @phihag.

phihag:
@dstftw Pings do nothing for me, sorry - I am just reachable via email. HTMLParser doesn't do DOM/Xpath/CSS query selectors, so it's not useful in the typical case of having to match something like `<h2><div>(.*)</`. To extract attributes, it's perfectly fine.

phihag:
Oh, I see that we previously used HTMLParser precisely to parse attributes. I'm sorry to say I don't remember why we switched. Before accepting this PR, please add a couple of testcases for exotic Unicode characters though (at least one inside and outside the BMP).

bpfoley:
Done. There is a gotcha though. The default build of Python 2.x, and 3.y for y<3 (including the system python on MacOS X) doesn't support unicode codepoints outside BMP. Literals like u'\u0012345' are split into surrogate pairs; unichr(0x12345) raises a ValueError; and HTMLParser leaves character refs undecoded. See https://docs.python.org/2/c-api/unicode.html and https://docs.python.org/3/c-api/unicode.html

bpfoley:
Ping. Any further thoughts on this one?

yan12125:
Sorry for not noticing this pull request, but I've pushed 7efc1c2b493aa1d68e5f1b248980144bbacf943a in which a naive approach applied to metadata extraction. Maybe this PR is no longer necessary?

remitamine:
even if the problem with twitter is fixed. but the part related to element attribute parsing is important to continue the work in #6811.

yan12125:
Then could you pull relevant parts from this PR with @bpfoley's permission? If #6811 is finally merged, Twitter can benefit from it, too.

remitamine:
> Then could you pull relevant parts from this PR with @bpfoley's permission?

if it's possible to merge part of PR(this commit bpfoley@0f867a684683184e65b307461bc3cc7d73d9a4ea) directly to master branch, and i will rebase on master and continue the work.

bpfoley:
OK, I've pushed a rebased version of just the extract_attributes code for now. I can look at updating the twitter code to use this function rather than hand-parsing HTML later if you're interested.

remitamine:
> If #6811 is finally merged, Twitter can benefit from it, too.

the changes in #6811 can take some time. and using extract_attributes for twitter now won't be difficult.

Issue #6018, opened 2015-06-18
--prefer-free-formats does not prefer webm to mp4


kparal:
```
$ youtube-dl 'https://www.youtube.com/watch?v=r9rGX91rq5I' -v -F
[debug] System config: [u'--prefer-free-formats']
[debug] User config: []
[debug] Command-line args: [u'https://www.youtube.com/watch?v=r9rGX91rq5I', u'-v', u'-F']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.06.04.1
[debug] Python version 2.7.10 - Linux-4.0.5-300.1.kparal.fc22.x86_64-x86_64-with-fedora-22-Twenty_Two
[debug] exe versions: ffmpeg 2.6.3, ffprobe 2.6.3
[debug] Proxy map: {}
[youtube] r9rGX91rq5I: Downloading webpage
[youtube] r9rGX91rq5I: Extracting video information
[youtube] r9rGX91rq5I: Downloading DASH manifest
[info] Available formats for r9rGX91rq5I:
format code  extension  resolution note
171          webm       audio only DASH audio  108k , audio@128k (44100Hz), 3.84MiB
140          m4a        audio only DASH audio  127k , m4a_dash container, aac  @128k (44100Hz), 4.76MiB
278          webm       256x144    DASH video   62k , webm container, VP9, 15fps, video only, 1.23MiB
242          webm       426x240    DASH video  112k , 30fps, video only, 1.88MiB
160          mp4        256x144    DASH video  122k , 15fps, video only, 3.37MiB
243          webm       640x360    DASH video  205k , 30fps, video only, 3.59MiB
244          webm       854x480    DASH video  281k , 30fps, video only, 6.15MiB
134          mp4        640x360    DASH video  297k , 30fps, video only, 3.44MiB
133          mp4        426x240    DASH video  307k , 30fps, video only, 7.54MiB
135          mp4        854x480    DASH video  626k , 30fps, video only, 7.05MiB
247          webm       1280x720   DASH video  645k , 30fps, video only, 14.85MiB
248          webm       1920x1080  DASH video 1204k , 30fps, video only, 27.42MiB
136          mp4        1280x720   DASH video 1286k , 30fps, video only, 14.79MiB
137          mp4        1920x1080  DASH video 2521k , 30fps, video only, 29.69MiB
[...]

PR #8125, opened 2016-01-02
Customizable sort formats

dstftw:
This PR proposes experimental `_sort_formats` behavior extension by introducing the notion of criterion.

The main problem with sorting now is that it's impossible to tweak the default sorting behavior without copy pasting sorting key routine or providing `field_preference` that is also flawed since it does not support calculated conditions. Thus the aim of this improvement is to be able to easily customize the default sorting behavior with minimal overhead and maximum convenience.

For this to happen the notion of _criterion_ is introduced. _Criterion_ is a either a field from formats dictionary or a synthetic expression calculated based on some data from formats dictionary. _Criterion_ is referenced as a plain string denoting it's name, _criteria_ is either a list or a tuple of criterions.

Currently following criteria are supported for sorting customization:

**Field criteria**:
- `format_id`
- `width`
- `height`
- `tbr`
- `abr`
- `vbr`
- `fps`
- `filesize`
- `filesize_approx`
- `language_preference`
- `quality`
- `source_preference`

**Synthetic criteria:**

| Synthetic criterion | Explanation |
| --- | --- |
| `preference` | Calculated based on `preference` field with some corrections |
| `proto_preference` | Calculated based on `protocol` field and prioritizes direct HTTP(s) URLs |
| `ext_preference` | Video extension preference, calculated based on `ext` field and value of `prefer_free_formats` setting |
| `audio_ext_preference` | Audio extension preference, calculated based on `ext` field and value of `prefer_free_formats` setting |

With this approach sorting can be customized in several ways.

**Way 1: Hardcode criteria preference**
Use only specified criterions when sorting formats removing any other default criteria from consideration. This is achieved by passing criteria to `criteria_preference` in order of preference.

**Example:** Override default criteria with set of only 5 criteria in exactly this preference:

``` python
# Bitrates are completely broken. Single m3u8 may contain entries in kbps and bps
# at the same time without actual units specified. This lead to wrong sorting.
self._sort_formats(formats, criteria_preference=('preference', 'height', 'width', 'fps', 'format_id'))
```

**Way 2: Tweak relative criteria preference**
Tweak criteria relative preference between several criteria keeping all other default criteria inplace. This is achieved by passing relative critea list to `criteria_relative_preference`.

**Example:** Prefer `width` over `height`, all other default criteria stay as before

``` python
self._sort_formats(formats, criteria_relative_preference=('width', 'height'))
```

**Way 3: Exclude criteria from consideration**
Remove criteria from consideration completely. This is achieved by passing criteria list to exclude to `criteria_exclusion`.

**Example:** Remove `tbr` from consideration, all other default criteria stay as before:

``` python
self._sort_formats(formats, criteria_exclusion=('tbr',))
```

Way 2 and Way 3 may be combined.

This solves #6018, #8001.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -904,7 +904,7 @@ class InfoExtractor(object):
                         default_criteria_order.remove(field)
 
             default_key_dict = dict(default_key)
-            return [default_key_dict[field] for field in criteria_order]
+            return [default_key_dict[criterion] for criterion in criteria_order]
 
         formats.sort(key=_formats_key)
 
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -764,7 +764,57 @@ class InfoExtractor(object):
             html, '%s form' % form_id, group='form')
         return self._hidden_inputs(form)
 
-    def _sort_formats(self, formats, field_preference=None):
+    def _sort_formats(self, formats, criteria_preference=None,
+                      criteria_relative_preference=None, criteria_exclusion=None):
+        """
+        Sort formats.
+
+        Sorting behavior can be customized by specifying absolute or relative criteria
+        order and criteria exclusion. Criteria is either a field from formats dictionary
+        or a synthetic expression calculated based on some data from formats dictionary.
+        Currently following criteria are supported for sorting customization:
+
+        Field criteria (see description of formats dict for each field explanation):
+            format_id
+            width
+            height
+            tbr
+            abr
+            vbr
+            fps
+            filesize
+            filesize_approx
+            language_preference
+            quality
+            source_preference
+
+        Synthetic criteria:
+            preference              Calculated based on preference field with some corrections
+            proto_preference        Calculated based on protocol field and prioritizes
+                                    direct HTTP(s) URLs
+            ext_preference          Video extension preference, calculated based on ext field
+                                    and value of prefer_free_formats setting
+            audio_ext_preference    Audio extension preference, calculated based on ext field
+                                    and value of prefer_free_formats setting
+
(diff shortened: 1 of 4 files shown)


youtube_dl/extractor/common.py
```diff
@@ -806,24 +853,59 @@ def _formats_key(f):
...
+                synthetic_criterion('audio_ext_preference', audio_ext_preference),
+                field_criterion('fps'),
+                field_criterion('filesize_approx'),
+                field_criterion('source_preference'),
+                field_criterion('format_id', ''),
             )
+
+            if (not criteria_preference and not criteria_relative_preference and
+                    not criteria_exclusion):
+                return default_key
```
jaimeMF: You are returning a tuple of `(filed, value)` item instead of a list of values as you do later.

youtube_dl/extractor/common.py
```diff
@@ -806,24 +853,59 @@ def _formats_key(f):
...
+
+            default_criteria_order = list(list(zip(*default_key))[0])
```
jaimeMF: I think it's more explicit to use a list comprehension:
```python
default_criteria_order = [name for name, _ in default_key]
```

youtube_dl/extractor/common.py
dstftw: Fixed in 69c5f149fa5478d5436c845b58a89f35383f32c7. Thanks.

youtube_dl/extractor/common.py
dstftw: Fixed in 8afe782893c99bfce2ce7c482f142fa6bdea8f6b.

eirnym:
@dstftw Could this branch be revised and eventually merged?

PR #8130, opened 2016-01-03
[youtube] added vcodec/acodec/abr for multiple itags

dyn888:
Should make downloading with filters more precise and easier, ie. bestvideo[vcodec=h264]. By default a lot of codecs are specified as avc1.xxxxxx and unique for each format, which makes them unusable for bestvideo selection.


yan12125:
How did you get bitrates? Any reference data?

dyn888:
Most of it by downloading and testing, then also from this page https://en.wikipedia.org/wiki/YouTube#Quality_and_formats for formats which are not in use anymore.

yan12125:
Good point. Bitrates parts can be merged. For acodec/vcodec parts, I have a different idea: #8218. Codecs names like ```avc1.640028``` provides more information than ```h264```, so I guess they should be preserved.

dyn888:
Yes [vcodec^=avc1] would also be a great functionality. However, with YouTube each avc.xxxxxx is unique for each format, so basically it is the same as specifying format directly. For instance, '-f 133' will always be 'avc1.4d4015', then '-f 134' will always be 'avc1.4d401e'. Then, with VP9 all of them are 'vp9', so it would make sense to simply merge avc1.* to 'h264', just like YouTube already uses 'vp9' for VP9 codecs. Is there really any reason to keep using avc1.xxxxxx besides format selection? It is much simpler to use x264, like with aac/vorbis/opus/vp9/...

dyn888:
Something to keep in mind:
- some formats already have 'h264' specified, so those would need to be removed in order to use avc1.* for all
- this update also adds codec info for h263, mp4v/mp4a, mp3, ... it is not limited to avc1, for instance '-f 5' only returns 'small' now, with this update it outputs 'small , h263, mp3  @ 64k' -- it adds more selection, so those additions should be preserved

youtube_dl/extractor/youtube.py
```diff
@@ -277,55 +277,55 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
...
-        # 3d videos
-        '82': {'ext': 'mp4', 'height': 360, 'format_note': '3D', 'preference': -20},
-        '83': {'ext': 'mp4', 'height': 480, 'format_note': '3D', 'preference': -20},
-        '84': {'ext': 'mp4', 'height': 720, 'format_note': '3D', 'preference': -20},
-        '85': {'ext': 'mp4', 'height': 1080, 'format_note': '3D', 'preference': -20},
-        '100': {'ext': 'webm', 'height': 360, 'format_note': '3D', 'preference': -20},
-        '101': {'ext': 'webm', 'height': 480, 'format_note': '3D', 'preference': -20},
-        '102': {'ext': 'webm', 'height': 720, 'format_note': '3D', 'preference': -20},
+        '5': {'ext': 'flv', 'width': 400, 'height': 240, 'acodec': 'mp3', 'abr': 64, 'vcodec': 'h263'},
+        '6': {'ext': 'flv', 'width': 450, 'height': 270, 'acodec': 'mp3', 'abr': 64, 'vcodec': 'h263'},
```
yan12125: For http://www.youtube.com/watch?v=BaW_jenozKcj, ffprobe says flv1+mp3 for format 5:
```
$ ffprobe "youtube-dl test video ''_ä↭𝕐-BaW_jenozKc.flv"   
ffprobe version 2.8.4 Copyright (c) 2007-2015 the FFmpeg developers
  built with gcc 5.3.0 (GCC)
  configuration: --prefix=/usr --disable-debug --disable-static --disable-stripping --enable-avisynth --enable-avresample --enable-fontconfig --enable-gnutls --enable-gpl --enable-ladspa --enable-libass --enable-libbluray --enable-libdcadec --enable-libfreetype --enable-libfribidi --enable-libgsm --enable-libmodplug --enable-libmp3lame --enable-libopencore_amrnb --enable-libopencore_amrwb --enable-libopenjpeg --enable-libopus --enable-libpulse --enable-libschroedinger --enable-libsoxr --enable-libspeex --enable-libssh --enable-libtheora --enable-libv4l2 --enable-libvidstab --enable-libvorbis --enable-libvpx --enable-libwebp --enable-libx264 --enable-libx265 --enable-libxvid --enable-shared --enable-version3 --enable-x11grab
  libavutil      54. 31.100 / 54. 31.100
  libavcodec     56. 60.100 / 56. 60.100
  libavformat    56. 40.101 / 56. 40.101
  libavdevice    56.  4.100 / 56.  4.100
  libavfilter     5. 40.101 /  5. 40.101
  libavresample   2.  1.  0 /  2.  1.  0
  libswscale      3.  1.101 /  3.  1.101
  libswresample   1.  2.101 /  1.  2.101
  libpostproc    53.  3.100 / 53.  3.100
Input #0, flv, from 'youtube-dl test video ''_ä↭𝕐-BaW_jenozKc.flv':
  Metadata:
    starttime       : 0
    totalduration   : 10
    totaldatarate   : 258
    bytelength      : 319093
    canseekontime   : true
    sourcedata      : BCAA9AF48
    purl            : 
    pmsg            : 
    httphostheader  : r5---sn-5njj-u2xe.googlevideo.com
  Duration: 00:00:09.80, start: 0.000000, bitrate: 260 kb/s
    Stream #0:0: Video: flv1, yuv420p, 426x240, 191 kb/s, 30 fps, 30 tbr, 1k tbn, 1k tbc
    Stream #0:1: Audio: mp3, 22050 Hz, stereo, s16p, 66 kb/s
```
Is ```h263``` the same (or almost the same) as ```flv1```?

dyn888:
Yes, it's Sorenson's implementation of h263 aka Sorenson Spark. FLV can be different codecs, that's why h263 is more precise.
https://en.wikipedia.org/wiki/Sorenson_Media#Sorenson_Spark

dyn888:
So, what would be best -- to use 'h264' or use '^=avc1' (when #8218 is ready)? In any case your patch is great, but also you can see YT already uses united 'vp9' for VP9 videos, 'vorbis'  and 'opus' for webm audio, so 'h264' is also logical. Some itags in existing code also have 'h264' already specified, my update expanded this to all of them.

dyn888:
So, do we merge this or not?

dyn888:
Anyone alive? :)

dstftw:
I would prefer changing the semantics of defaults to act as fallbacks when no corresponding metadata available rather that to override.

yan12125:
Thanks @dstftw for supporting my idea. I have some work for this idea but I didn't publish it before. Now it's at #8293.

dyn888:
But the question remains -- why? Is there any advantage of doing this at all, since YouTube won't change their formats later, instead, they will use new itags. In any case, this patch could be merged into master already, even if you later make it not to override returned values. In case of vp9, vorbis and opus this is not an issue anyway, only with avc1.xxxxxx it is.

About avc1<>h264, as stated before, there are some itags defined in extractor which already have 'h264' already specified there, my patch only expanded this to all itags, so in case 'avc1' is preferred those should be removed, regardless if this patch is merged or not.

dyn888:
@dstftw Not sure what data is returned at any given time, but in general it is better to override returned data when it provides no usable info, for instance -f 5 by default says 'small', with this patch it says 'small , h263, mp3 @ 64k', which is much better. Something to keep in mind, instead of giving priority to returned data. As said previously, itags are added as new formats emerge, old itags seem to be always the same.

yan12125:
Sorry for being late. Merged.

dyn888:
Thank you!

PR #8142, opened 2016-01-05
[update] fix (unexploitable) BB'06 vulnerability in rsa_verify

FiloSottile:
The rsa_verify code was vulnerable to a BB'06 attack, allowing to forge
signatures for arbitrary messages if and only if the public key exponent is
3.  Since the updates key is hardcoded to 65537, there is no risk for
youtube-dl, but I don't want vulnerable code in the wild.

The new function adopts a way safer approach of encoding-and-comparing to
replace the dangerous parsing code.

### DIFF
--- a/test/test_update.py
+++ b/test/test_update.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python
+
+from __future__ import unicode_literals
+
+# Allow direct execution
+import os
+import sys
+import unittest
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+
+import json
+from youtube_dl.update import rsa_verify
+
+
+class TestUpdate(unittest.TestCase):
+    def test_rsa_verify(self):
+        UPDATES_RSA_KEY = (0x9d60ee4d8f805312fdb15a62f87b95bd66177b91df176765d13514a0f1754bcd2057295c5b6f1d35daa6742c3ffc9a82d3e118861c207995a8031e151d863c9927e304576bc80692bc8e094896fcf11b66f3e29e04e3a71e9a11558558acea1840aec37fc396fb6b65dc81a1c4144e03bd1c011de62e3f1357b327d08426fe93, 65537)
+        with open(os.path.join(os.path.dirname(os.path.abspath(__file__)), 'versions.json'), 'rb') as f:
+            versions_info = f.read().decode()
+        versions_info = json.loads(versions_info)
+        signature = versions_info['signature']
+        del versions_info['signature']
+        self.assertTrue(rsa_verify(
+            json.dumps(versions_info, sort_keys=True).encode('utf-8'),
+            signature, UPDATES_RSA_KEY))
+
+
+if __name__ == '__main__':
+    unittest.main()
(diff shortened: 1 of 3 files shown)


FiloSottile:
More details: https://blog.filippo.io/bleichenbacher-06-signature-forgery-in-python-rsa/

FiloSottile:
```
ERROR: test_rsa_verify (test.test_update.TestUpdate)
  File "/home/travis/build/rg3/youtube-dl/test/test_update.py", line 27, in test_rsa_verify
    signature, UPDATES_RSA_KEY))
  File "/home/travis/build/rg3/youtube-dl/youtube_dl/update.py", line 21, in rsa_verify
    signature = b'%x' % pow(int(signature, 16), key[1], key[0])
TypeError: unsupported operand type(s) for %: 'bytes' and 'int'
```

yan12125:
Well this format is unavailable in Python < 3.5. [1]

[1] https://www.python.org/dev/peps/pep-0460/

jaimeMF:
> `TypeError: unsupported operand type(s) for %: 'bytes' and 'int'`

With this alternate form the test passes in all python versions:
```python
signature = ('%x' % pow(int(signature, 16), key[1], key[0])).encode()
```

Please run `flake8` and fix the style issues.

FiloSottile:
```
test_rsa_verify (test.test_update.TestUpdate) ... ok
test_rsa_verify (test.test_update.TestUpdate) ... ok
test_rsa_verify (test.test_update.TestUpdate) ... ok
test_rsa_verify (test.test_update.TestUpdate) ... ok
test_rsa_verify (test.test_update.TestUpdate) ... ok
test_rsa_verify (test.test_update.TestUpdate) ... ok
```

jaimeMF:
Is there some reason to use the last `versions.json`? It uses a lot of space (for example the tar.gz file produced with `python setup.py sdist` goes from 784K to 868K), we could use a smaller file (like https://github.com/rg3/youtube-dl/blob/77b66b65bd7ad85d84a66b7cfc1944d7e03328cb/update/versions.json).

Additionally: due to my lack of knowledge I can't comment on the actual change, since you know more feel free to merge it after some time if nobody disagrees.

FiloSottile:
Very good point.

The crypto itself was reviewed by a couple people that know what they are doing, so should be alright. Or at least better than the one before.

I just want to be sure I'm not bricking the update process.

jaimeMF:
> I just want to be sure I'm not bricking the update process.

Well, I have tried to update and it works. Feel free to merge when you think it's ready.

yan12125:
I've just learned [```hmac.compare_digest```](https://docs.python.org/2/library/hmac.html#hmac.compare_digest). Maybe this function can be used for comparison? It's available only in Python 2.7.7+ or 3.3+. Nevertheless, try to use it:
```Python
import hmac

# First try to use the safe comparison function to prevent timing analysis attack
if hasattr(hmac, 'compare_digest'):
    return hmac.compare_digest(expected, signature)
else:  # Otherwise, fallback to unsafe direct comparison
    return expected == signature
```
This funciton is written in C for providing timing-analysis proof. (https://hg.python.org/cpython/file/e0df94327586/Modules/_operator.c#l165) I guess it's impossible to port it to pure Python for older Python versions.

PR #6454, opened 2015-08-04
Support arbitrary stream merges

fstirlitz:
This supersedes #5581 and #6435.

With this change, the merge operator may join any number of media streams, video or audio. The streams are downloaded in the order specified.

Use case:

```
youtube-dl http://www.gdcvault.com/play/1014631/Classic-Game-Postmortem-PAC -f speaker+slides+en
```

shall download the `slides` (video only), `speaker` (video+audio) and `en` (audio only) streams and merge them into a single file.

Remarks:
- If more than one video or audio streams are to be merged, the default output format is set to `mkv`. This isn't always necessary: `webm`, `ogg` and `mp4` will do fine with multiple audio and/or video streams; `flv`, on the other hand, will not (and possibly others).
- The metadata for a stream of a particular type are copied into the `info_dict` of the merged stream only if there is a single stream of that type (i.e. only one audio stream and only one video stream). This seems sensible (after all, questions like "what is the audio codec of `bestaudio[acodec=opus]+bestaudio[acodec=vorbis]`" are somewhat ill-formed), but other behaviours might be preferable here (e.g. copying the metadata of the "best-quality" stream of a given type, or copying them if they are the same for every stream and leaving them otherwise, or creating a list of all the values for all the streams).

PR #8157, opened 2016-01-07
[YoutubeDL] Gracefully handle audio/video format reversal (Fixes #5154)

### LINKED ISSUE #5154: Format Merging Auto Correction
When downloading different formats and merging them, but entering `-f someaudio+somevideo`, the programme returns: `ERROR: The first format must contain the video, try using "-f somevideo+someaudio"`. If it already "understands" the formats and can offer the alternate form, why can the programme not be run with the alternate form, preserving normal stream order in the merge? :confused:

nateschiffer:
Improved error handling by automatically attempting the suggested fix and swapping formats, failing with error only in the case where two audio-only formats are given. Simplified audio format handling logic and comments, for readability.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1105,15 +1105,17 @@ class YoutubeDL(object):
             elif selector.type == MERGE:
                 def _merge(formats_info):
                     format_1, format_2 = [f['format_id'] for f in formats_info]
-                    # The first format must contain the video and the
-                    # second the audio
+                    # The first format must contain the video
+                    # If the formats are reversed, swap them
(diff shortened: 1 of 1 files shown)


dstftw:
Formats should be downloaded in order they appear in format selection expression. There is already a more generic proposal for downloading and merging arbitrary formats: #6454.

nateschiffer:
>Formats should be downloaded in order they appear in format selection expression.

Ok, but what does that break that's *worse* than the current behavior?

Because the currently behavior has the formats not downloaded at all, failing by suggesting a fix that could easily have been checked for automatically (since youtube-dl already knows which formats contain audio/video by that point).

PR #8174, opened 2016-01-08
[v.qq.com]qq video extractor

52Jolynn:
extractor of v.qq.com


youtube_dl/extractor/__init__.py
```diff
@@ -18,11 +18,11 @@
 from .aftonbladet import AftonbladetIE
 from .airmozilla import AirMozillaIE
 from .aljazeera import AlJazeeraIE
+from .allocine import AllocineIE
```
dstftw: These changes (alphabetic reordering) are not related to qq extractor thus should not be present in this PR.

PR #8175, opened 2016-01-08
[v.ifeng.com]Ifeng extractor

52Jolynn:
extractor of v.ifeng.com.
凤凰网


dstftw:
This PR should not contain changes unrelated to ifeng extractor.

PR #8176, opened 2016-01-08
tvp.pl extractor update

d42:
Basically, the public tv overlords have changed the url routing, and some of the api.
As a bonus it handles shows in their own subdomains now :3

### DIFF
--- a/youtube_dl/extractor/tvp.py
+++ b/youtube_dl/extractor/tvp.py
@@ -75,9 +75,10 @@ class TvpIE(InfoExtractor):
         'playlist_count': 86,
     }]
 
-    def _get_json(self, url, entry_id):
+    def _get_json(self, url, entry_id, fatal=True):
         formatted_url = url.format(id=int(entry_id))
-        return self._download_json(formatted_url, entry_id)
+        json = self._download_json(formatted_url, entry_id, fatal=fatal)
+        return {} if json is None else json
 
     def _format_formats(self, formats, video_id):
 
@@ -164,7 +165,7 @@ class TvpIE(InfoExtractor):
         entry_id = self._match_id(url)
         ctx = self._get_json(self._META_URL, entry_id)['context']
         if ctx['format_id'] == 0:
-            file_info = self._get_json(self._FILE_INFO_URL, entry_id)
+            file_info = self._get_json(self._FILE_INFO_URL, entry_id, fatal=False)
             original_id = file_info.get('copy_of_object_id')
             if original_id:
                 ctx = self._get_json(self._META_URL, original_id)['context']
--- a/youtube_dl/extractor/tvp.py
+++ b/youtube_dl/extractor/tvp.py
@@ -1,4 +1,6 @@
 # -*- encoding: utf-8 -*-
+from __future__ import unicode_literals
+
 from .common import InfoExtractor, ExtractorError
 
 
--- a/youtube_dl/extractor/tvp.py
+++ b/youtube_dl/extractor/tvp.py
@@ -150,10 +150,9 @@ class TvpIE(InfoExtractor):
                 if 'directory_video' in item['types']:
                     ids.append(item['_id'])
                 if 'video' in item['types'] and item['is_released']:
-                    yield {
-                        '_type': 'url',
-                        'title': self._guess_title(item),
-                        'url': item['url']}
+                    yield self.url_result(item['url'],
+                                          video_id=item['_id'],
+                                          video_title=self._guess_title(item))
 
     def _get_playlist(self, context):
         pls_id = str(context['material_id'])
--- a/youtube_dl/extractor/tvp.py
+++ b/youtube_dl/extractor/tvp.py
@@ -48,7 +48,7 @@ class TvpApi:
 
 class TvpIE(InfoExtractor):
     IE_NAME = 'tvp.pl'
-    _VALID_URL = r'https?://(?:vod|www)\.(\w+\.)?tvp\.pl/(?P<id>\d+)/.*'
+    _VALID_URL = r'https?://(\w+\.)+tvp\.pl/(?P<id>\d+)/.*'
 
     _TESTS = [{
         'url': 'http://vod.tvp.pl/4278035/odc-2',
--- a/youtube_dl/extractor/tvp.py
+++ b/youtube_dl/extractor/tvp.py
@@ -1,12 +1,181 @@
-# -*- coding: utf-8 -*-
-from __future__ import unicode_literals
+from .common import InfoExtractor, ExtractorError
 
-import re
+VIDEO_LISTING_URL = ('http://www.api.v3.tvp.pl/shared/listing.php'
+                     '?dump=json&direct=true&count=-1&parent_id={id}')
+META_URL = 'http://www.tvp.pl/shared/video_data.php?dump=json&video_id={id}'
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/tvp.py
```diff
@@ -1,139 +1,247 @@
...
+VIDEO_LISTING_URL = ('http://www.api.v3.tvp.pl/shared/listing.php'
+                     '?dump=json&direct=true&count=-1&parent_id={id}')
+META_URL = 'http://www.tvp.pl/shared/video_data.php?dump=json&video_id={id}'
+TOKENIZER_URL = 'http://www.tvp.pl/shared/cdn/tokenizer_v2.php?object_id={id}'
+FILE_INFO_URL = 'http://www.tvp.pl/pub/stat/videofileinfo?video_id={id}'
+IGNORED_MIMETYPES = 'application/vnd.ms-ss', 'application/x-mpegurl'
 
-from .common import InfoExtractor
+
+class TvpApi:
```
dstftw: I don't see much point extracting it in separate class. Base class inherited from `InfoExtractor` is preferred.

youtube_dl/extractor/tvp.py
```diff
@@ -1,139 +1,247 @@
...
-from __future__ import unicode_literals
+# -*- encoding: utf-8 -*-
+from .common import InfoExtractor, ExtractorError
 
-import re
+VIDEO_LISTING_URL = ('http://www.api.v3.tvp.pl/shared/listing.php'
+                     '?dump=json&direct=true&count=-1&parent_id={id}')
+META_URL = 'http://www.tvp.pl/shared/video_data.php?dump=json&video_id={id}'
+TOKENIZER_URL = 'http://www.tvp.pl/shared/cdn/tokenizer_v2.php?object_id={id}'
+FILE_INFO_URL = 'http://www.tvp.pl/pub/stat/videofileinfo?video_id={id}'
```
dstftw: I don't see much point in there constants since each is only used once.

youtube_dl/extractor/tvp.py
```diff
@@ -1,139 +1,247 @@
...
+        json = self._get_json(VIDEO_LISTING_URL, id)
+        return json
+
+    def meta(self, id):
+        json = self._get_json(META_URL, id)
+        return json
+
+    def info(self, id):
+        json = self._get_json(FILE_INFO_URL, id)
+        return json
```
dstftw: Code duplication in these three methods. Again they are only used once, so there is really no need in them.

youtube_dl/extractor/tvp.py
```diff
@@ -1,139 +1,247 @@
...
+        json = self._get_json(TOKENIZER_URL, id)
+        status = json['status']
+        if status == 'NOT_PLAYABLE':
+            raise ExtractorError("video is not playable")
+        if status != 'OK':
+            raise ExtractorError("unknown status: %s", status)
+        return json['formats']
+
+    def _get_json(self, url, id):
+        id = int(id)
```
dstftw: Avoid shadowing built-in names.

youtube_dl/extractor/tvp.py
```diff
@@ -1,139 +1,247 @@
...
-            m3u8_url = re.sub('([^/]*)\.ism/manifest', r'\1.ism/\1.m3u8', video_url)
-            formats = self._extract_m3u8_formats(m3u8_url, video_id, 'mp4')
+            title = ', '.join([context['website_title'], context['title']])
+        url = context['url']
+        description = context['description_root']
+
+        try:
+            formats = self._format_formats(self.api.formats(id), id)
+        except ExtractorError as e:
+            self.to_screen("%s: %s" % (title, e))
```
dstftw: What's the point of this?

youtube_dl/extractor/tvp.py
```diff
@@ -1,139 +1,247 @@
...
         else:
-            m3u8_url = re.sub('([^/]*)\.ism/manifest', r'\1.ism/\1.m3u8', video_url)
-            formats = self._extract_m3u8_formats(m3u8_url, video_id, 'mp4')
+            title = ', '.join([context['website_title'], context['title']])
+        url = context['url']
+        description = context['description_root']
+
+        try:
+            formats = self._format_formats(self.api.formats(id), id)
+        except ExtractorError as e:
```
dstftw: If any of the formats fails the whole extraction is broken.

youtube_dl/extractor/tvp.py
```diff
@@ -1,139 +1,247 @@
...
+            item_id = ids.pop()
+            listing = self.api.listing(item_id)
+            for item in listing['items']:
+                if 'directory_video' in item['types']:
+                    ids.append(item['_id'])
+                if 'video' in item['types'] and item['is_released']:
+                    meta = self.api.context(item['_id'])
+                    try:
+                        yield self._get_video(meta)
+                    except ExtractorError:
```
dstftw: This defeats the whole point of `ExtractorError`. It should be propagated and handled outside the extractor not muted in extractor.

youtube_dl/extractor/tvp.py
```diff
@@ -1,139 +1,247 @@
...
+
+    def context(self, id):
+        meta = self.meta(id)
+        return meta['context']
+
+    def formats(self, id):
+        json = self._get_json(TOKENIZER_URL, id)
+        status = json['status']
+        if status == 'NOT_PLAYABLE':
+            raise ExtractorError("video is not playable")
```
dstftw: Use single quotes consistently whenever possible. If this is expected error pass `expected=True`.

youtube_dl/extractor/tvp.py
```diff
@@ -1,139 +1,247 @@
...
+        if context['title_root']:
+            title = context['title_root']
+        elif not context['website_title']:
+            title = context['title']
         else:
-            m3u8_url = re.sub('([^/]*)\.ism/manifest', r'\1.ism/\1.m3u8', video_url)
-            formats = self._extract_m3u8_formats(m3u8_url, video_id, 'mp4')
+            title = ', '.join([context['website_title'], context['title']])
+        url = context['url']
+        description = context['description_root']
```
dstftw: Extration should be tolerate to any missing optional data. Anything apart from title and data required for formats extraction considered optional. So if `description_root` is missing if should not break the extraction.

dstftw:
I recommend you looking through the code of any recent extractor for best practices and idiomatic coding style.

youtube_dl/extractor/tvp.py
d42: Sometimes playlists contain videos that don't provide any sources, so I ended up raising exception and ignoring it if it was raised inside a playlist,
Also i wanted to know that something like that happened, so i reraised it here to print out which video was at fault.

Is there a mechanism for handling something like that?

youtube_dl/extractor/tvp.py
d42: API always provides that key, even if value is just an empty string, should I handle it nonetheless?

youtube_dl/extractor/tvp.py
dstftw: Video that does not provide any sources sounds like unexpected outcome therefore it should not be muted. If user wants to ignore this she have to pass `-i`.

youtube_dl/extractor/tvp.py
dstftw: Today it provides, tomorrow it doesn't. But extractor is broken across all the packages some of which are not tend to update fast enough.

PR #8177, opened 2016-01-08
[extractor] v.ifeng.com(凤凰网)

52Jolynn:
extractor of v.ifeng.com

### DIFF
--- a/youtube_dl/extractor/ifeng.py
+++ b/youtube_dl/extractor/ifeng.py
@@ -1,0 +1,64 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/ifeng.py
```diff
@@ -0,0 +1,78 @@
...
+                'url': element.get('VideoPlayUrl'),
+                'ext': 'mp4',
+                'quality' : quality
+            })
+
+        return {
+            'id': video_id,
+            'title': title,
+            '_type': 'multi_video',
+            'entries': [entry],
```
dstftw: What's the point using `multi_video` if there is always only one entry in `entries`?

PR #8194, opened 2016-01-09
[Bigflix] Add new extractor for bigflix.com

vickyg3:
Add an IE to support bigflix.com. It uses some sort of silverlight
plugin whose video url is being populated using base64 encoded
flashvars. So it is quite straightforward to extract.

### DIFF
--- a/youtube_dl/extractor/bigflix.py
+++ b/youtube_dl/extractor/bigflix.py
@@ -1,0 +1,42 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from base64 import b64decode
+
+from .common import InfoExtractor
+from ..compat import compat_urllib_parse_unquote
+
+
+class BigflixIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?bigflix\.com/.*/(?P<id>[0-9]+)'
+    _TEST = {
+        'url': 'http://www.bigflix.com/Hindi-movies/Action-movies/Singham-Returns/16537',
+        'md5': 'ec76aa9b1129e2e5b301a474e54fab74',
+        'info_dict': {
+            'id': '16537',
+            'ext': 'mp4',
+            'title': 'Singham Returns',
+            'description': 'md5:3d2ba5815f14911d5cc6a501ae0cf65d',
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._html_search_regex(
+            r'<div[^>]+class=["\']pagetitle["\'][^>]*>(.+?)</div>',
+            webpage, 'title')
+
+        video_url = b64decode(compat_urllib_parse_unquote(self._search_regex(
+            r'file=([^&]+)', webpage, 'video url')).encode('ascii')).decode('utf-8')
+
+        description = self._html_search_meta('description', webpage)
+
+        return {
+            'id': video_id,
+            'title': title,
+            'url': video_url,
+            'description': description,
+        }
--- a/youtube_dl/extractor/__init__.py
+++ b/youtube_dl/extractor/__init__.py
@@ -61,6 +61,7 @@ from .beeg import BeegIE
 from .behindkink import BehindKinkIE
 from .beatportpro import BeatportProIE
 from .bet import BetIE
+from .bigflix import BigflixIE
 from .bild import BildIE
 from .bilibili import BiliBiliIE
 from .bleacherreport import (


docs/supportedsites.md
```diff
@@ -65,6 +65,7 @@
  - **Beeg**
  - **BehindKink**
  - **Bet**
+ - **Bigflix**
```

youtube_dl/extractor/bigflix.py
```diff
@@ -0,0 +1,56 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from base64 import b64decode
+from urllib import unquote
```
dstftw: Use `compat_urllib_parse_unquote` instead.

youtube_dl/extractor/bigflix.py
```diff
@@ -0,0 +1,56 @@
...
+                'title': 'Madarasapatinam',
+                'description': 'md5:63b9b8ed79189c6f0418c26d9a3452ca',
+            }
+        },
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        request = sanitized_Request(url)
```
dstftw: No need to use `sanitized_Request` request here, pass url directly to download method.

youtube_dl/extractor/bigflix.py
```diff
@@ -0,0 +1,56 @@
...
+        webpage = self._download_webpage(request, video_id)
+
+        title = self._html_search_regex(
+            r'<div[^>]+class=["\']pagetitle["\'][^>]*>(.+?)</div>',
+            webpage, 'title')
+
+        video_url = b64decode(unquote(self._search_regex(
```
dstftw: 1. `video_url` must be unicode string.
2. On python 3.2 `b64decode` accepts only bytes.

docs/supportedsites.md
vickyg3: Removed this from the PR. Would it be okay if i added this to .gitignore then?

youtube_dl/extractor/bigflix.py
vickyg3: Done.

youtube_dl/extractor/bigflix.py
vickyg3: Done.

youtube_dl/extractor/bigflix.py
vickyg3: Fixed 2 by encoding the input to b64decode as ASCII.
Fixed 1 by encoding the output of b64decode as utf-8.

vickyg3:
Thanks for the quick review. Have addressed your comments. Please take another look. :)

docs/supportedsites.md
dstftw: No, it's committed on every new release.

vickyg3: alright.

youtube_dl/extractor/bigflix.py
```diff
@@ -0,0 +1,54 @@
...
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._html_search_regex(
+            r'<div[^>]+class=["\']pagetitle["\'][^>]*>(.+?)</div>',
+            webpage, 'title')
+
+        video_url = b64decode(compat_urllib_parse_unquote(self._search_regex(
+            r'file=([^&]+)', webpage, 'video url')).encode('ASCII')).encode('utf-8')
```
dstftw: It should be `decode('utf-8')` not encode. Also use lowercase for encoding name for consistency.

youtube_dl/extractor/bigflix.py
```diff
@@ -0,0 +1,54 @@
...
+            'md5': 'ec76aa9b1129e2e5b301a474e54fab74',
+            'info_dict': {
+                'id': '16537',
+                'ext': 'mp4',
+                'title': 'Singham Returns',
+                'description': 'md5:3d2ba5815f14911d5cc6a501ae0cf65d',
+            }
+        },
+        {
+            'url': 'http://www.bigflix.com/Tamil-movies/Drama-movies/Madarasapatinam/16070',
```
dstftw: If both tests test the same functionality leave only one. Otherwise put short comment for additional test.

dstftw:
Could you also squash commits after you fix everything?

youtube_dl/extractor/bigflix.py
vickyg3: done.

youtube_dl/extractor/bigflix.py
vickyg3: done.

vickyg3:
Have squashed everything into a single commit.

dstftw:
Thanks.

PR #8212, opened 2016-01-12
Composer package

Rudloff:
Hello,

Would it be possible to add this composer.json file to the repo?
It would allow PHP-based projects to use youtube-dl as a [Composer](https://getcomposer.org/) dependency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ test/testdata
 youtube-dl.zsh
 .idea
 .idea/*
+vendor/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -32,4 +32,5 @@ test/testdata
 .tox
 youtube-dl.zsh
 .idea
-.idea/*
\ No newline at end of file
+.idea/*
+vendor/
```
Rudloff: Composer creates a *vendor/* directory, it should not be added to the repo.

yan12125:
youtube-dl is a python package but not PHP. If you want to download youtube-dl automatically, you should handle it in PHP.

PR #8215, opened 2016-01-12
[pluralsight] Fix new player update

zendyani:
Pluralsight updated their player, so this is a little fix to make Pluralsight extractor work again.
Thanks


youtube_dl/extractor/pluralsight.py
```diff
@@ -232,7 +232,7 @@ def guess_allowed_qualities():
         # { a = author, cn = clip_id, lc = end, m = name }
 
         return {
-            'id': clip['clipName'],
+            'id': clip['name'],
```
dstftw: This breaks the old player.

PR #8218, opened 2016-01-13
[WIP] Introduce CSS3 like string operators

yan12125:
As pointed out by #8130, codec names like ```avc1.640028``` is difficult for format selection. I borrow some ideas from [CSS3 selectors](http://www.w3.org/TR/css3-selectors/#selectors). Here's an usage example:

```
$ youtube-dl -j -f 'worstvideo[vcodec^=avc1]' test:youtube | jq .format
"160 - 256x144 (DASH video)"
```

Currently there are no tests. I'll fill them up if this idea is acceptable.


jaimeMF:
Looks fine, it should also be documented in the README.

Issue #8227, opened 2016-01-13
HLS vulnerability in both FFMpeg and NativeHlsFD


yan12125:
A recent post points out that FFMpeg can cause contents of arbitrary files (for example ```/etc/passwd```) being accessible on the Internet via a malicious input file. [1][2] The reaction of Arch Linux developers is disabling affected components before they are fixed. [3] In this commit, ```concat:``` protocol and HLS support are disabled. The former one is not used in the mainline codebase, just in some pull requests (#2844). The latter one is more serious. I've just updated my copy to the latest official Arch binary. Downloading an YouTube live stream gives:

```
$ youtube-dl -v "https://www.youtube.com/watch?v=clO8XxFkrj4"
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['-v', 'https://www.youtube.com/watch?v=clO8XxFkrj4']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.01.09
[debug] Git HEAD: 40cf7fc
[debug] Python version 3.5.1 - Linux-4.3.3-2-ARCH-x86_64-with-arch-Arch-Linux
[debug] exe versions: ffmpeg 2.8.4, ffprobe 2.8.4, rtmpdump 2.4
[debug] Proxy map: {}
[youtube] clO8XxFkrj4: Downloading webpage
[youtube] clO8XxFkrj4: Downloading video info webpage
[youtube] clO8XxFkrj4: Extracting video information
[youtube] clO8XxFkrj4: Downloading formats manifest
[youtube] clO8XxFkrj4: Downloading DASH manifest
[...]

PR #8228, opened 2016-01-13
[YoutubeDL] urlopen: disable the 'file:' protocol (#8227)

jaimeMF:
If someone is running youtube-dl on a server to deliver files, the user could input 'file:///some/important/file' and youtube-dl would save that file as a video giving access to sensitive information to the user.
'file:' urls can be filtered, but the user can use an URL to a crafted m3u8 manifest like:

    #EXTM3U
    #EXT-X-MEDIA-SEQUENCE:0
    #EXTINF:10.0
    file:///etc/passwd
    #EXT-X-ENDLIST

With this patch 'file:' URLs raise URLError like for unknown protocols.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1994,7 +1994,7 @@ class YoutubeDL(object):
         file_handler = compat_urllib_request.FileHandler()
 
         def file_open(*args, **kwargs):
-            raise compat_urllib_error.URLError('file:/// protocol is explicitly disabled in youtube-dl for security reasons')
+            raise compat_urllib_error.URLError('file:// scheme is explicitly disabled in youtube-dl for security reasons')
         file_handler.file_open = file_open
 
         opener = compat_urllib_request.build_opener(
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1986,8 +1986,14 @@ class YoutubeDL(object):
         https_handler = make_HTTPS_handler(self.params, debuglevel=debuglevel)
         ydlh = YoutubeDLHandler(self.params, debuglevel=debuglevel)
         data_handler = compat_urllib_request_DataHandler()
-        opener = compat_urllib_request.build_opener(
-            proxy_handler, https_handler, cookie_processor, ydlh, data_handler)
+        unknown_handler = compat_urllib_request.UnknownHandler()
+        handlers = (proxy_handler, https_handler, cookie_processor, ydlh, data_handler, unknown_handler)
+        # we don't use build_opener because it automatically adds FileHandler,
+        # which can be used for malicious purposes (see
+        # https://github.com/rg3/youtube-dl/issues/8227)
+        opener = compat_urllib_request.OpenerDirector()
+        for handler in handlers:
+            opener.add_handler(handler)
 
         # Delete the default user-agent header, which would otherwise apply in
         # cases where our custom HTTP handler doesn't come into play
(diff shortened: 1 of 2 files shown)


yan12125:
LGTM. Thanks for such a clean solution!

jaimeMF:
I have added a commit for using build_opener again (I would squash them before merging), because otherwise we manually need to add some handlers for HTTP redirects.

yan12125:
Personally I prefer whitelists than blacklists, but either way is OK in this case as I don't think ```default_classes``` in ```build_opener``` would change in near future.

If you are going to use the blacklist way, a longer error message may be better:
```
file:/// protocol is explicitly disabled in youtube-dl for security reasons
```

jaimeMF:
> Personally I prefer whitelists than blacklists, but either way is OK in this case as I don't think default_classes in build_opener would change in near future.

I also prefer whitelists, I haven't looked into it but you probably have to add a few handlers which makes the code a bit ugly. Additionally if in the future they split or merge some handlers we don't have to update the code.

> If you are going to use the blacklist way, a longer error message may be better:

Good idea, changed in https://github.com/rg3/youtube-dl/commit/4240d504963bb6d1c7bd7c288a7874f9d8dc042b

youtube_dl/YoutubeDL.py
```diff
@@ -1986,8 +1986,19 @@ def _setup_opener(self):
...
         data_handler = compat_urllib_request_DataHandler()
+
+        # When passing our own FileHandler instance, build_opener won't add the
+        # default FileHandler and allows us to disable the file protocol, which
+        # can be used for malicious purposes (see
+        # https://github.com/rg3/youtube-dl/issues/8227)
+        file_handler = compat_urllib_request.FileHandler()
+
+        def file_open(*args, **kwargs):
+            raise compat_urllib_error.URLError('file:/// protocol is explicitly disabled in youtube-dl for security reasons')
```
dstftw: As of [RFC1378](https://www.ietf.org/rfc/rfc1738.txt) file URL scheme *does not specify an Internet protocol or access method for such files*. So I suppose it'd be technically more correct to refer to it as `file:// URL scheme` or simply `file:// scheme` (this is how it's referred in urllib). Note two slashes since more generic form is `file://<host>/<path>`.

jaimeMF: Changed in https://github.com/jaimeMF/youtube-dl/commit/30e2f2d76f6dd52803effce14fa14f3a8051c84a, thanks.

dstftw:
LGTM.

clytras:
There are filesystem permissions to handle access. If a user has access to `/etc/passwd`, then that user does not need `youtube-dl` to download/get that file, they can simply `cat /etc/passwd` or even `curl file:///etc/passwd` to access it. Only `root` should have access to that kind of files.

Does `curl` or `wget` restrict `file://` downloads? No. Why should a CLI tool such as `youtube-dl` restrict the file protocol?

Implementing such "security" methods have no point at all. Operating systems and especially NIX systems have already a very strong security filesystem mechanism.

Please check ['Allow "file://" URLs'](https://github.com/ytdl-org/youtube-dl/issues/22408) (https://github.com/ytdl-org/youtube-dl/issues/22408) so at least if you still want this king of (_to me unnecessary_) security feature, then let the users to disable it by using configuration files.

mp3butcher:
Not developping ydl for so long but here's my opinion (i can be wrong):
This so call security issue isn't harmful: one could retrieve its own files through YDL ... it's kind a normal behavior for user with correct file permission... The only way I can see it could harm is if one put code in ydl to SEND these file elsewhere. 

Further, YDL extractors requires to embed a payload to reconstruct standard fileformat (hls,dash..) (payloads are site specifics)
and after that YDL leave the choice between native and ffmpeg processing of these standard formats.
So YDL require a local storage accessible to ffmpeg in order to process these reconstructed files.
For me, the simplier to allow that is to allow file:///... 
An other way to go seams to  create a new ydl://localstore:// protocol for  temporary ydl files and a translation to file:/// protocol to communicate with ffmpeg
Any feedback welcome..

PR #8236, opened 2016-01-14
Including feature to increase or decrease media (volume) audio level

phnahes:
This changes adds a new feature to increase or decrease media audio level  (volume). Sometimes after download, the levels are too low, so amplifies the levels.

I add a option to select how many dB use. And default value on 4 dB.


README.md
```diff
@@ -361,6 +361,9 @@ which means you can modify it, redistribute it or use it however you like.
                                      a value between 0 (better) and 9 (worse)
                                      for VBR or a specific bitrate like 128K
                                      (default 5)
+    --audio-volume VOLUME            Specify ffmpeg/avconv audio volume, insert 
+				     a value in dB to increase audio volume.
+                                     (default 4)
```
dstftw: This file is autogenerated.

youtube_dl/__init__.py
```diff
@@ -377,7 +383,7 @@ def _real_main(argv=None):
     with YoutubeDL(ydl_opts) as ydl:
         # Update version
         if opts.update_self:
-            update_self(ydl.to_screen, opts.verbose, ydl._opener)
```
dstftw: This change is unrelated and breaks updating.

youtube_dl/options.py
```diff
@@ -688,6 +688,12 @@ def _hide_login_info(opts):
         '--audio-quality', metavar='QUALITY',
         dest='audioquality', default='5',
         help='Specify ffmpeg/avconv audio quality, insert a value between 0 (better) and 9 (worse) for VBR or a specific bitrate like 128K (default %default)')
+
+    postproc.add_option(
+        '--audio-volume', metavar='QUALITY',
+        dest='audiovolume', default=None,
+        help='Specify ffmpeg/avconv audio volume, insert a value in dB to increase audio volume (default %default)')
```
dstftw: Default part should be removed since there is no default.

youtube_dl/options.py
```diff
@@ -688,6 +688,12 @@ def _hide_login_info(opts):
         '--audio-quality', metavar='QUALITY',
         dest='audioquality', default='5',
         help='Specify ffmpeg/avconv audio quality, insert a value between 0 (better) and 9 (worse) for VBR or a specific bitrate like 128K (default %default)')
+
+    postproc.add_option(
+        '--audio-volume', metavar='QUALITY',
```
dstftw: audio volume is a tautology. `metavar` should be `VOLUME`.

fstirlitz:
How about a ReplayGain postprocessor instead?

PR #8246, opened 2016-01-15
Initial JSON-LD metadata extraction support

dstftw:


remitamine:
there is also the [VideoObject schema](http://schema.org/VideoObject).
this is out of the scope of this PR but some times i see the use of TVEpisode and VideoObject with the microdata format(like in dailymotion video page) it will be good to support it also.
the VideoObject schema can be helpful in the generic extractor because it can contain the embed url that can lead to a supported extractor.

dstftw:
There are much much more metadata available. I don't set a goal to cover everything right now but rather extend it as needed.

yan12125:
ld+json format is also found in ```bbc.py```. I guess it's be better to use ```_search_json_ld``` in bbc.py, too.

PR #8247, opened 2016-01-15
[vidzi] Updated extractor

Tropfchen:


dstftw:
It works fine with current master for me. Moreover there is already a more generic solution [proposed](https://github.com/rg3/youtube-dl/pull/6998).

PR #8249, opened 2016-01-16
[neteasemusic] Fixes #8239

### LINKED ISSUE #8239: Netease broken again
Tests in master for extractor will fail.
Looks like they switched download server again.

ping:
203.130.59.9 is a reverse proxy for the netease cdn [via https://github.com/Unblocker/Unblock-Youku/issues/488#issuecomment-162314538]

### DIFF
--- a/youtube_dl/extractor/neteasemusic.py
+++ b/youtube_dl/extractor/neteasemusic.py
@@ -40,7 +40,7 @@ class NetEaseMusicBaseIE(InfoExtractor):
             if not details:
                 continue
             formats.append({
-                'url': 'http://m5.music.126.net/%s/%s.%s' %
+                'url': 'http://203.130.59.9/m1.music.126.net/%s/%s.%s' %
                        (cls._encrypt(details['dfsId']), details['dfsId'],
                         details['extension']),
                 'ext': details.get('extension'),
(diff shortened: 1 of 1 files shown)


yan12125:
If I understand it correct, users in Mainland China can download netease music without problem. We shouldn't enforce proxies on all users.

ping:
@yan12125 If that's the case, will you accept a PR that is something like this:
```python
song_url = 'http://m5.music.126.net/%s/%s.%s' % (
    self._encrypt(details['dfsId']), details['dfsId'], details['extension'])
if not self._is_valid_url(song_url, info['id']):
    song_url = 'http://203.130.59.9/m1.music.126.net/%s/%s.%s' % (
        self._encrypt(details['dfsId']), details['dfsId'], details['extension'])
```

yan12125:
Sorry but it's not a good idea to use hardcoded proxies in the codebase. You can add a new option and use this option in ```neteasemusic.py```.

ping:
@yan12125 Even if the reverse proxy apparently belongs to netease? (as claimed by the poster in the in the linked comment). 

I'm curious what you mean by "add a new option", do you mean an extractor-specific option? Do you have example of an extractor doing this?

yan12125:
Well I thought 黄易 is different from 网易. As a record, 203.130.59.9 belongs to the same company as netease.com. [1][2]

This fix is OK. Just eliminate duplicated codes and add some comments.

[1] http://whois.domaintools.com/203.130.59.255
[2] http://whois.domaintools.com/netease.com

yan12125:
Thanks merged.

hoschi111:
I guess this proxy-feature is broken again.
Test:
http://198.47.104.134/m1.music.126.net/9Oq39CoHx4H4W2i438mBjQ==/2888417047551516.mp3

=> Access Denied

PR #8265, opened 2016-01-17
[SVTPlay] Add subtitle support

maix:
basic subtitle support for svtplay.se

### DIFF
--- a/youtube_dl/extractor/svt.py
+++ b/youtube_dl/extractor/svt.py
@@ -37,6 +37,12 @@ class SVTBaseIE(InfoExtractor):
                 })
         self._sort_formats(formats)
 
+        # SVT does not tell us the language, so we assume swedish.
+        subtitles = {}
+        for sr in video_info['subtitleReferences']:
+            if 'url' in sr:
+                subtitles.setdefault('sv', []).append({'url': sr['url']})
+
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/svt.py
```diff
@@ -37,6 +37,12 @@ def _extract_video(self, url, video_id):
                 })
         self._sort_formats(formats)
 
+        # SVT does not tell us the language, so we assume swedish.
+        subtitles = {}
+        for sr in video_info['subtitleReferences']:
+            if 'url' in sr:
+                subtitles.setdefault('sv', []).append({'url': sr['url']})
```
dstftw: Valid ISO 3166-1 code for Sweden is `se`.

dstftw:
Can you add a test?

youtube_dl/extractor/svt.py
maix: yes but the language code (ISO-639) for swedish is sv :) https://en.wikipedia.org/wiki/Swedish_language

maix:
> Can you add a test?
It seems that all the videos on the site are only accessible within a specific timeframe, mostly only 7 or 30 days, a few for a year—that would probably mean that the test will fail after some time. So I'd say it's better not to include one at all, what do you think?

PR #8273, opened 2016-01-18
[OraTV] update extractor

CeruleanSky:
"current" is now "video"
"hls_stream" is now hls_stream without quotes
video_id is now id
duration for current video is not present(for other videos it is)

### DIFF
--- a/youtube_dl/extractor/ora.py
+++ b/youtube_dl/extractor/ora.py
@@ -21,7 +21,6 @@ class OraTVIE(InfoExtractor):
             'ext': 'mp4',
             'title': 'Vine & YouTube Stars Zach King & King Bach On Their Viral Videos!',
             'description': 'md5:ebbc5b1424dd5dba7be7538148287ac1',
-            'duration': 1477,
         }
     }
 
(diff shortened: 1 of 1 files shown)


CeruleanSky:
This fixes the broken Ora TV extractor due to changes in the variable names in the webpage.  It removes the duration test , as that information is no longer present for the current video that is being requested to be downloaded.  For some reason duration is still listed for the suggested videos though, so it may return, but in the meantime it is better not to test for something that isn't there.

youtube_dl/extractor/ora.py
```diff
@@ -30,9 +29,9 @@ def _real_extract(self, url):
         webpage = self._download_webpage(url, display_id)
 
         video_data = self._search_regex(
-            r'"current"\s*:\s*({[^}]+?})', webpage, 'current video')
+            r'"video"\s*:\s*({[^}]+?})', webpage, 'current video')
         m3u8_url = self._search_regex(
-            r'"hls_stream"\s*:\s*"([^"]+)', video_data, 'm3u8 url', None)
+            r'hls_stream\s*:\s*"([^"]+)', video_data, 'm3u8 url', None)
```
dstftw: Both quoted and quoteless should be tried. Looking at their JSON they can switch back any moment.

CeruleanSky:
Modified regex to take into account that hls_stream may be put in quotes again in the future.

dstftw:
Cherry-picked in 1f78ed189aab9b7ea17594ac1f33662d37433455, thanks.

PR #8360, opened 2016-01-29
Remove extra spaces & trim title

dyn888:
This removes extra spaces in title, also removes any leading and trailing whitespace, before title size check and after title is copied to 'fulltitle'. Repeated spaces can waste space in a filename, leading spaces can mess up sorting, and make file handling more difficult.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1423,6 +1423,7 @@ class YoutubeDL(object):
                 raise MaxDownloadsReached()
 
         info_dict['fulltitle'] = info_dict['title']
+        info_dict['title'] = " ".join(info_dict['title'].split())
         if len(info_dict['title']) > 200:
             info_dict['title'] = info_dict['title'][:197] + '...'
 


dstftw:
Extra whitespace may be intentional, even the trailing one. Particular extractor should handle this individually.

dyn888:
Title whitespace intentional? Any example of this? How about a commandline option to offer users a choice about this?

dstftw:
>Title whitespace intentional? Any example of this?

Say, I'm the author of some YouTube video and I want it to be entitled "A&nbsp;&nbsp;&nbsp;B", and don't want it to be "A B" under no cirsumstances. Beforehand you don't know my intentions whether it's intentional or not.

dyn888:
Then it could be perhaps a command line option, like: `--trim-title` ... That way users could use it when needed.

dstftw:
I don't see much point for such specific options. Also idiomatic trim does not trim inner whitespace so suggested name does not reflect the semantics properly. Moreover there are lots of other textual options apart from title adding option for each is overkill.

dyn888:
Then it can be renamed to something else. I see no reason not to allow one option to clean white space, not sure why are you opposed to it either. It is useful option to have, required one simple line of code, and it can be Off by default, so there is nothing change if it is not explicitly used.

dstftw:
I've already pointer out why - it's too specific.
Why only trim all whitespace, somebody may want to trim only inner, add another one option for that? One may want to trim only left or right, or only tabs - another bunch of options? Why only title, there are description/uploader/series/etc, add options for them also?

dyn888:
Because title goes to filename. Some videos have leading and trailing spaces and it creates files with those, which is generally not desired. This could also be done at the end with the final filename, not limited to the title -- filename is more important. I am in no rush to add this, but it could be an useful option, limited to filenames, similar to `--restrict-filenames`, just for whitespace removing.

dyn888:
Hehe, did you know underscores "_" are already trimmed and collapsed? Try with IKZhQnc5cc8 -- so what now if someone wants a title like __super__cool__title__ .... it doesn't work. And it is not extractor-specific, it is inside utils.py. :)

dstftw:
It's just buggy legacy filename sanitize implementation.

dyn888:
It's because sanitization can result in multiple consequent `_` and those are later collapsed. Not sure what's buggy about that, it works OK for what it does.

dstftw:
It should not collapse the original underscores but only those that represent invalid characters. Moreover I don't even see much point collapsing underscores that were previously invalid characters.

dyn888:
It should do the same thing for spaces, nobody needs extra spaces in filenames.

PR #8362, opened 2016-01-29
Trim and collapse whitespace from filename

dyn888:
Removes leading, trailing and repeated spaces/whitespace.


dyn888:
Note: This could be added as a command-line option, if that's preferable.
Something similar to ` --restrict-filenames` (which already collapses spaces converted to underscores).

PR #8376, opened 2016-01-30
[schooltv] Add extractor for SchoolTV playlists

### LINKED ISSUE #8163: Site request: schooltv.nl
It belongs to the npo.nl network of sites, so I suspect you can use the same code.
[Example url](http://www.schooltv.nl/video/ademhaling-de-hele-dag-haal-je-adem-maar-wat-gebeurt-er-dan-eigenlijk-in-je-lichaam/)
The ID of the video can be found in the `data-mid` attribute of a div with class `VideoPlayer`.

rrooij:
This closes #8163

### DIFF
--- a/youtube_dl/extractor/npo.py
+++ b/youtube_dl/extractor/npo.py
@@ -406,6 +406,37 @@ class NPORadioFragmentIE(InfoExtractor):
         }
 
 
+class SchoolTVIE(InfoExtractor):
+    IE_NAME = 'schooltv'
+    _VALID_URL = r'https?://(?:www\.)?schooltv\.nl/video/(?P<id>[^/?#&]+)'
+
+    _TEST = {
+        'url': 'http://www.schooltv.nl/video/ademhaling-de-hele-dag-haal-je-adem-maar-wat-gebeurt-er-dan-eigenlijk-in-je-lichaam/',
+        'info_dict': {
+            'id': 'WO_NTR_429477',
+            'display_id': 'ademhaling-de-hele-dag-haal-je-adem-maar-wat-gebeurt-er-dan-eigenlijk-in-je-lichaam',
+            'title': 'Ademhaling: De hele dag haal je adem. Maar wat gebeurt er dan eigenlijk in je lichaam?',
+            'ext': 'mp4',
+            'description': 'md5:abfa0ff690adb73fd0297fd033aaa631'
+        },
+        'params': {
+            # Skip because of m3u8 download
+            'skip_download': True
+        }
+    }
+
+    def _real_extract(self, url):
+        display_id = self._match_id(url)
+        webpage = self._download_webpage(url, display_id)
+        video_id = self._search_regex(r'data-mid="([^"]+)"', webpage, 'video_id')
+        return {
+            '_type': 'url_transparent',
+            'ie_key': 'NPO',
+            'url': 'npo:%s' % video_id,
+            'display_id': display_id
+        }
+
+
 class VPROIE(NPOIE):
     IE_NAME = 'vpro'
     _VALID_URL = r'https?://(?:www\.)?(?:tegenlicht\.)?vpro\.nl/(?:[^/]+/){2,}(?P<id>[^/]+)\.html'
--- a/youtube_dl/extractor/__init__.py
+++ b/youtube_dl/extractor/__init__.py
@@ -483,6 +483,7 @@ from .npo import (
     NPOLiveIE,
     NPORadioIE,
     NPORadioFragmentIE,
+    SchoolTVIE,
     VPROIE,
     WNLIE
 )


youtube_dl/extractor/npo.py
```diff
@@ -406,6 +406,42 @@ def _real_extract(self, url):
         }
 
 
+class SchoolTVIE(NPOIE):
```
dstftw: No need to inherit from `NPOIE`.

youtube_dl/extractor/npo.py
```diff
@@ -406,6 +406,42 @@ def _real_extract(self, url):
...
+    IE_NAME = 'schooltv'
+    _VALID_URL = r'https?://(?:www\.)?schooltv\.nl/video/(?P<id>.*[^/])?/'
```
dstftw: `id` should not be optional. No need in trailing `/`.

youtube_dl/extractor/npo.py
```diff
@@ -406,6 +406,42 @@ def _real_extract(self, url):
...
+        }
+    ]
+
+    def _real_extract(self, url):
+        playlist_id = self._match_id(url)
+        webpage = self._download_webpage(url, playlist_id)
+
+        entries = [
+            self.url_result('npo:%s' % video_id)
+            for video_id in re.findall(r'data-mid="([^"]+)"', webpage)
```
dstftw: This is too broad and detects the same video twice.

youtube_dl/extractor/npo.py
```diff
@@ -406,6 +406,42 @@ def _real_extract(self, url):
...
+            }
+        },
+        {
+            'url': 'http://www.schooltv.nl/video/waarom-zijn-beroemde-kunstenaars-vaak-mannen/',
+            'playlist_count': 2,
+            'info_dict': {
+                'id': 'waarom-zijn-beroemde-kunstenaars-vaak-mannen',
+                'title': 'Schooltv: ARTclips - Waarom zijn beroemde kunstenaars vaak mannen?',
+            }
+        }
```
dstftw: If both tests test the same scenario leave only one.

rrooij:
Wow, thanks for the feedback! Should I fix the issues and squash it in the commit?

dstftw:
Yes.

rrooij:
Done.

One question: is it normal behavior that the video file in the test returns a m3u8 playlist? The VPRO IE returns a .m4v and they are both from NPO, so I'm curious whether this is correct.

youtube_dl/extractor/npo.py
```diff
@@ -406,6 +406,32 @@ def _real_extract(self, url):
         }
 
 
+class SchoolTVIE(InfoExtractor):
+    IE_NAME = 'schooltv'
+    _VALID_URL = r'https?://(?:www\.)?schooltv\.nl/video/(?P<id>.*[^/])/?'
```
dstftw: Change to `(?P<id>[^/]+)`.

youtube_dl/extractor/npo.py
```diff
@@ -406,6 +406,32 @@ def _real_extract(self, url):
...
+    def _real_extract(self, url):
+        playlist_id = self._match_id(url)
+        webpage = self._download_webpage(url, playlist_id)
+
+        entries = [
+            self.url_result('npo:%s' % video_id)
+            for video_id in re.findall(r'VideoPlayer.*data-mid="([^"]+)"', webpage)
+        ]
+        playlist_title = self._og_search_title(webpage)
+        return self.playlist_result(entries, playlist_id, playlist_title)
```
dstftw: From what I've seen there is always only one video on the page thus no need in playlist.

dstftw:
>test returns a m3u8

Fixed in a7685b3a6b02be62ebd53e2978ab43b1cde3ec4b.

rrooij:
I think it is fixed by now.

* It's much shorter now that it doesn't try to find a playlist (it is pointless anyway since no page seems to contain a playlist
* The VALID_URL regex is replaced by `(?P<id>[^/?#&]+)` as you said
* The test now assumes a mp4 to be returned instead of a m3u8.

youtube_dl/extractor/npo.py
```diff
@@ -406,6 +406,27 @@ def _real_extract(self, url):
...
+            'title': 'Ademhaling: De hele dag haal je adem. Maar wat gebeurt er dan eigenlijk in je lichaam?',
+            'ext': 'mp4',
+            'description': 'md5:abfa0ff690adb73fd0297fd033aaa631'
+        }
+    }
+
+    def _real_extract(self, url):
+        playlist_id = self._match_id(url)
+        webpage = self._download_webpage(url, playlist_id)
+        video_id = re.findall(r'VideoPlayer.*data-mid="([^"]+)"', webpage)[0]
```
dstftw: Use `self._search_regex` instead.

rrooij:
Done. I also changed the variable names since the id doesn't resemble a playlist id anymore.

youtube_dl/extractor/npo.py
```diff
@@ -406,6 +406,27 @@ def _real_extract(self, url):
...
+            'title': 'Ademhaling: De hele dag haal je adem. Maar wat gebeurt er dan eigenlijk in je lichaam?',
+            'ext': 'mp4',
+            'description': 'md5:abfa0ff690adb73fd0297fd033aaa631'
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        npo_video_id = self._search_regex(r'data-mid="([^"]+)"', webpage, video_id)
```
dstftw: 3rd argument should be the name of field you search for `'video id'`.

youtube_dl/extractor/npo.py
```diff
@@ -406,6 +406,27 @@ def _real_extract(self, url):
...
+        'info_dict': {
+            'id': 'WO_NTR_429477',
+            'title': 'Ademhaling: De hele dag haal je adem. Maar wat gebeurt er dan eigenlijk in je lichaam?',
+            'ext': 'mp4',
+            'description': 'md5:abfa0ff690adb73fd0297fd033aaa631'
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
```
dstftw: This is usually called `display_id` and included in info dict as well.

rrooij:
I changed the variable to `display_id`. However, I didn't include it in the info_dict because the test fails. It expects the video_id instead (`WO_NTR_...`).

dstftw:
I mean it should be included as `display_id` field.

rrooij:
Did I do it correctly?

As far as I can see, the difference between the `_type` being `url_transparant` and `url` is that the former lets you override the info_dict of the other extractor? Is this correct? Since it does the display_id correctly now, instead of when I manually added the display_id to the dictionary result from `url_result`.

Thanks for your help by the way!

youtube_dl/extractor/npo.py
```diff
@@ -406,6 +406,33 @@ def _real_extract(self, url):
...
+
+    _TEST = {
+        'url': 'http://www.schooltv.nl/video/ademhaling-de-hele-dag-haal-je-adem-maar-wat-gebeurt-er-dan-eigenlijk-in-je-lichaam/',
+        'info_dict': {
+            'id': 'WO_NTR_429477',
+            'display_id': 'ademhaling-de-hele-dag-haal-je-adem-maar-wat-gebeurt-er-dan-eigenlijk-in-je-lichaam',
+            'title': 'Ademhaling: De hele dag haal je adem. Maar wat gebeurt er dan eigenlijk in je lichaam?',
+            'ext': 'mp4',
+            'description': 'md5:abfa0ff690adb73fd0297fd033aaa631'
+        }
```
dstftw: m3u8 downloads with ffmpeg should be [skipped](https://github.com/rg3/youtube-dl/commit/60ad3eb9706861d4182ab44ee19d64350ca2e36e).

rrooij:
Done. And I understand, downloading the video is redundant in test cases if the json contains all the necessary information.

dstftw:
No. ffmpeg or rtmpdump may stall making the whole test build stall.

dstftw:
Thanks.

PR #8378, opened 2016-01-30
[Gamekings] Fix failed extraction and update URL

rrooij:
Some old videos that aren't on Vimeo are being uploaded to YouTube under the
'Gamekings Vault' channel. They use YouTube now for some videos as video
hosting instead of Vimeo or their own hosting. The first test failed to
succeed under the existing code, but works now by using the YouTube
extractor.

The Regex is changed to find the new gogoVideo JavaScript line with the
YouTube embed. Checking if there is a YouTube embed is done by a String
find, which is probably not the best method of checking this.

The URLs are also updated. Gamekings uses the `.nl` TLD now instead of `.tv`.

### DIFF
--- a/youtube_dl/extractor/gamekings.py
+++ b/youtube_dl/extractor/gamekings.py
@@ -9,9 +9,9 @@ from ..utils import (
 
 
 class GamekingsIE(InfoExtractor):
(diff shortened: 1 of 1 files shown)


yan12125:
Thanks merged.

PR #8394, opened 2016-02-01
force id to be string for download_archive

dennishofmann:
As far as I know, netease uses numbers, which are no strings and need to get converted.


phihag:
In python 2, `str is bytes`, so this doesn't do what you expect. Instead, why not fix the extractor and document that `id` must be a (character) string?

PR #8439, opened 2016-02-05
Add --skip-protocols to exclude specific protocols from format extraction

remitamine:
this is different from using `protocol!=` because it won't request manifest urls at all for the specified protocols.

### DIFF
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -340,14 +340,6 @@ def parseOpts(overrideArguments=None):
         action='store_true', dest='listformats',
         help='List all available formats of requested videos')
     video_format.add_option(
-        '--youtube-include-dash-manifest',
-        action='store_true', dest='youtube_include_dash_manifest', default=True,
-        help=optparse.SUPPRESS_HELP)
-    video_format.add_option(
-        '--youtube-skip-dash-manifest',
-        action='store_false', dest='youtube_include_dash_manifest',
-        help='Do not download the DASH manifests and related data on YouTube videos')
-    video_format.add_option(
         '--merge-output-format',
         action='store', dest='merge_output_format', metavar='FORMAT', default=None,
         help=(
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -354,6 +354,11 @@ def parseOpts(overrideArguments=None):
             'If a merge is required (e.g. bestvideo+bestaudio), '
             'output to given container format. One of mkv, mp4, ogg, webm, flv. '
             'Ignored if no merge is required'))
+    video_format.add_option(
+        '--skip-protocols',
+        action='callback', dest='skip_protocols', type='str',
+        default=[], callback=_comma_separated_values_options_callback,
+        help='Protocols to exclude from the format extraction separated by commas (Possible values: m3u8, f4m, dash and smil)')
 
     subtitles = optparse.OptionGroup(parser, 'Subtitle Options')
     subtitles.add_option(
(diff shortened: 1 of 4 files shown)


youtube_dl/options.py
```diff
@@ -340,20 +340,17 @@ def _hide_login_info(opts):
         action='store_true', dest='listformats',
         help='List all available formats of requested videos')
     video_format.add_option(
-        '--youtube-include-dash-manifest',
-        action='store_true', dest='youtube_include_dash_manifest', default=True,
-        help=optparse.SUPPRESS_HELP)
-    video_format.add_option(
-        '--youtube-skip-dash-manifest',
-        action='store_false', dest='youtube_include_dash_manifest',
-        help='Do not download the DASH manifests and related data on YouTube videos')
```
dstftw: These options should be kept and probably marked deprecated in order to avoid breakage of user configurations and code that uses them. `youtube_include_dash_manifest` logic should be expressed via `skip_protocols`.

youtube_dl/options.py
```diff
@@ -340,20 +340,17 @@ def _hide_login_info(opts):
...
-    video_format.add_option(
         '--merge-output-format',
         action='store', dest='merge_output_format', metavar='FORMAT', default=None,
         help=(
             'If a merge is required (e.g. bestvideo+bestaudio), '
             'output to given container format. One of mkv, mp4, ogg, webm, flv. '
             'Ignored if no merge is required'))
+    video_format.add_option(
+        '--skip-protocols',
```
dstftw: Technically these are not protocols. I'm aware we are using some of them in `protocol` metafield in info dict but still I'd prefer diffrerent wording. Moreover I'd prefer something even more generic like `--skip-extraction` with comma separated value list of possible values: `m3u8`, `f4m`, `mpd`, `smil`, `metadata` etc. (possibly even extractor-specific ones). This would allow to fine-tune the extraction not to download unnecessary data and minimize network I/O.

youtube_dl/options.py
```diff
@@ -340,20 +340,17 @@ def _hide_login_info(opts):
...
+        action='callback', dest='skip_protocols', type='str',
+        default=[], callback=_comma_separated_values_options_callback,
+        help='Protocols to exclude from the format extraction separated by commas (Possible values: m3u8, f4m, dash and smil)')
```
dstftw: This should be clarified that it's not guaranteed to actually exclude processing of mentioned things since extractor may use own extraction routines.

youtube_dl/options.py
remitamine: if we try to add other types of steps i think we should update extractor to use common methods so it will be simpler to add this contraits like.
_real_extract_media_formats, _real_extract_subtitles, _real_extract_metadata... that extractors will override and wrap them inside methods without _real where we can check for the --skip-extraction-steps and may add other behavior(messages about starting the specific step)

youtube_dl/options.py
remitamine: isn't possible to notify users when they update that there are some changes that they need to apply.

dstftw: This will require them to remove these options from configurations/code anyway.

remitamine: if we don't ask user to remove them than this options will stay in the code even if there is a better and generic way to do it.
by asking them to use the new options we can remove deprecated code quickly.

remitamine:
i will work on the idea of skiping steps as you request when i have more time.

PR #8445, opened 2016-02-06
[rte:radio] Add support for RTMP downloads, alternate URL style

bpfoley:
This is useful as
a) RTMP downloads are a good deal faster to download
b) Older items are available only as RTMP streams


dstftw:
Thanks.

PR #8446, opened 2016-02-06
[vlive] Updated to new V App/VLive api.

EraYaN:
More robust with getting keys and ids from website.

The only thing to make it more robust is to not hard code this part:
```
http://global.apis.naver.com/rmcnmv/rmcnmv/vod_play_videoInfo.json
```

Sadly I couldn't really see where it was actually defined in the webpage.

It has some variables at the top but those seem to be for different things.

```javascript
	var nsc = "v.all";
	
	var lcs_SerName = "global.apis.naver.com";
	var ccsrv = "global.apis.naver.com";
	var svr = "fvweb06.vlive";
```

### DIFF
--- a/youtube_dl/extractor/vlive.py
+++ b/youtube_dl/extractor/vlive.py
@@ -9,17 +9,18 @@ from time import time
 from .common import InfoExtractor
 from ..utils import (
     ExtractorError,
-    determine_ext
+    determine_ext,
+    int_or_none
 )
 from ..compat import compat_urllib_parse
 
 
 class VLiveIE(InfoExtractor):
     IE_NAME = 'vlive'
-    # www.vlive.tv/video/ links redirect to m.vlive.tv/video/ for mobile devices
+    # vlive.tv/video/ links redirect to www.vlive.tv/video/ 
     _VALID_URL = r'https?://(?:(www|m)\.)?vlive\.tv/video/(?P<id>[0-9]+)'
     _TEST = {
-        'url': 'http://m.vlive.tv/video/1326',
+        'url': 'http://www.vlive.tv/video/1326',
         'md5': 'cc7314812855ce56de70a06a27314983',
         'info_dict': {
             'id': '1326',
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/vlive.py
```diff
@@ -28,50 +28,64 @@ class VLiveIE(InfoExtractor):
...
 
         webpage = self._download_webpage(
-            'http://m.vlive.tv/video/%s' % video_id,
+            'http://www.vlive.tv/video/%s' % video_id,
             video_id, note='Download video page')
 
+        webpage_output = webpage.encode('utf8', 'replace')
+        text_file = open("webpage.html", "w", )
+        text_file.write(webpage_output)
+        text_file.close()
```
dstftw: Debug code must be removed.

youtube_dl/extractor/vlive.py
```diff
@@ -28,50 +28,64 @@ class VLiveIE(InfoExtractor):
...
     }
-    _SECRET = 'rFkwZet6pqk1vQt6SxxUkAHX7YL3lmqzUMrU4IDusTo4jEBdtOhNfT4BYYAdArwH'
+    
+    def clean_shitty_vlive_json(self, jsonstring):
+        start = jsonstring.find("{");
+        end = jsonstring.rfind("}");
+        json_output = jsonstring.encode('utf8', 'replace')
+        text_file = open("vlive.json", "w", )
+        text_file.write(json_output)
+        text_file.close()
```
dstftw: Debug code must be removed.

youtube_dl/extractor/vlive.py
```diff
@@ -28,50 +28,64 @@ class VLiveIE(InfoExtractor):
             'creator': 'Girl\'s Day',
         },
     }
-    _SECRET = 'rFkwZet6pqk1vQt6SxxUkAHX7YL3lmqzUMrU4IDusTo4jEBdtOhNfT4BYYAdArwH'
+    
+    def clean_shitty_vlive_json(self, jsonstring):
```
dstftw: This routine is not used.

EraYaN:
Good point, Removed and squashed.

youtube_dl/extractor/vlive.py
```diff
@@ -28,50 +28,49 @@ class VLiveIE(InfoExtractor):
...
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
 
         webpage = self._download_webpage(
-            'http://m.vlive.tv/video/%s' % video_id,
+            'http://www.vlive.tv/video/%s' % video_id,
             video_id, note='Download video page')
 
+        #TODO One regex with two captures? (re.search() https://docs.python.org/2/library/re.html )
```
dstftw: Either implement or remove.

youtube_dl/extractor/vlive.py
```diff
@@ -28,50 +28,49 @@ class VLiveIE(InfoExtractor):
...
+        long_video_id = self._html_search_regex(
```
dstftw: `_search_regex` is enough here.

youtube_dl/extractor/vlive.py
```diff
@@ -28,50 +28,49 @@ class VLiveIE(InfoExtractor):
...
+            r'vlive\.tv\.video\.ajax\.request\.handler\.init\("[0-9]+",\s?"[^"]*",\s?"([^"]+)",\s?"[^"]+",\s?"[^"]*",\s?"[^"]*"\)', webpage, 'long_video_id')
+
+        key = self._html_search_regex(
```
dstftw: `_search_regex` is enough here.

youtube_dl/extractor/vlive.py
```diff
@@ -28,50 +28,49 @@ class VLiveIE(InfoExtractor):
...
+            r'vlive\.tv\.video\.ajax\.request\.handler\.init\("[0-9]+",\s?"[^"]*",\s?"[^"]+",\s?"([^"]+)",\s?"[^"]*",\s?"[^"]*"\)', webpage, 'long_video_id')
```
dstftw: Field name is supposed to be `key` not `long_video_id`.

youtube_dl/extractor/vlive.py
```diff
@@ -28,50 +28,49 @@ class VLiveIE(InfoExtractor):
...
-        url += '&' + compat_urllib_parse.urlencode({'msgpad': msgpad, 'md': md})
-        playinfo = self._download_json(url, video_id, 'Downloading video json')
+            r'<div class="info_area">\s*<strong[^>]+class="name">([^<>]+)</strong>', webpage, 'creator')
 
-        if playinfo.get('message', '') != 'success':
-            raise ExtractorError(playinfo.get('message', 'JSON request unsuccessful'))
+        # doct = document type (xml or json), cpt = caption type (vtt or ttml)
+        url = "http://global.apis.naver.com/rmcnmv/rmcnmv/vod_play_videoInfo.json?videoId=%s&key=%s&ptc=http&doct=json&cpt=vtt" % (long_video_id, key)
+        
+        playinfo = self._download_json(url, video_id, 'Downloading video json',encoding='utf-8',fatal=False)
```
dstftw: No need to specify encoding. `fatal` should be `True` (default) since you are not providing alternative extraction scenario.

youtube_dl/extractor/vlive.py
```diff
@@ -28,50 +28,49 @@ class VLiveIE(InfoExtractor):
...
 
-        if not playinfo.get('result'):
+        if not playinfo.get('videos'):
             raise ExtractorError('No videos found.')
```
dstftw: No need for this check, this is already checked in `_sort_formats`.

youtube_dl/extractor/vlive.py
```diff
@@ -28,50 +28,49 @@ class VLiveIE(InfoExtractor):
...
+        for vid in playinfo.get('videos', {}).get('list', []):
             formats.append({
                 'url': vid['source'],
                 'ext': 'mp4',
                 'abr': vid.get('bitrate', {}).get('audio'),
                 'vbr': vid.get('bitrate', {}).get('video'),
                 'format_id': vid['encodingOption']['name'],
-                'height': vid.get('height'),
-                'width': vid.get('width'),
+                'height': vid.get('encodingOption').get('height'),
```
dstftw: This will fail if `encodingOption` is missing. Also wrap in `int_or_none` from `utils`.

EraYaN: Any idea on how you check for this in python?

in C(++/#) i would have done something like
```
vid.get('encodingOption') ? int_or_none(vid.get('encodingOption').get('height')) : None
```

dstftw: You've already did this with `videos` and `captions`: `int_or_none(vid.get('encodingOption', {}).get('height'))`

youtube_dl/extractor/vlive.py
```diff
@@ -28,50 +29,45 @@ class VLiveIE(InfoExtractor):
...
+            r'vlive\.tv\.video\.ajax\.request\.handler\.init\("[0-9]+",\s?"[^"]*",\s?"([^"]+)",\s?"[^"]+",\s?"[^"]*",\s?"[^"]*"\)', webpage, 'long_video_id')
+
+        key = self._search_regex(
+            r'vlive\.tv\.video\.ajax\.request\.handler\.init\("[0-9]+",\s?"[^"]*",\s?"[^"]+",\s?"([^"]+)",\s?"[^"]*",\s?"[^"]*"\)', webpage, 'key')
+
         title = self._og_search_title(webpage)
         thumbnail = self._og_search_thumbnail(webpage)
         creator = self._html_search_regex(
-            r'<span[^>]+class="name">([^<>]+)</span>', webpage, 'creator')
+            r'<div class="info_area">\s*<strong[^>]+class="name">([^<>]+)</strong>', webpage, 'creator')
```
dstftw: `fatal` should be `False` for this field.

dstftw:
Cherry-picked in 52f5889f77e04420634d70145d689660ea6bfe24, thanks.

PR #8454, opened 2016-02-07
New extractor Animefullx

felixdollack:
Added extractor for animefullx.
Includes 3 _TESTS which are running fine and shows no problems with flake8.


dstftw:

dstftw:
As of https://www.google.com/transparencyreport/removals/copyright/domains/animefullx.com/ and brief skim through pages of this site is obvious that it's dedicated to copyright infringement and does not contain any user generated content thus can't be supported by youtube-dl.

felixdollack:
yep. didn't see that.

PR #8473, opened 2016-02-09
[arte:magazine] Add extractor

Schnouki:


youtube_dl/extractor/arte.py
```diff
@@ -276,6 +276,29 @@ class ArteTVCinemaIE(ArteTVPlus7IE):
...
+            'title': 'Trepalium - Extrait Ep.01',
+        },
+    }, {
+        'url': 'http://www.arte.tv/magazine/metropolis/de/frank-woeste-german-paris-metropolis',
+        'md5': '53185a2b4876383552c037c21aeca8f7',
+        'info_dict': {
+            'id': '066149-000-A',
+            'ext': 'mp4',
+            'title': 'Bonus: Frank Woeste - ein deutscher Jazzmusiker in Paris',
+        },
```
dstftw: If this both test the same functionality leave only one or make the other `'only_matching': True`.

Issue #8301, opened 2016-01-22
Broken: laola1.tv (site layout changed)


egolus:
as the site changed its layout, youtube-dl can't find the stream any more.

example link: `http://www.laola1.tv/watch/video/460627`
the livestreams don't work either.

~~~
 .\youtube-dl.exe -v http://www.laola1.tv/watch/video/460627
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://www.laola1.tv/watch/video/460627']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2016.01.15
[debug] Python version 2.7.10 - Windows-8-6.2.9200
[debug] exe versions: ffmpeg 2.8.4, ffprobe 2.8.4
[debug] Proxy map: {}
[generic] 460627: Requesting header
[redirect] Following redirect to http://www.laola1.tv/de-de/video/460627
[generic] 460627: Requesting header
WARNING: Falling back on generic information extractor.
[generic] 460627: Downloading webpage
[generic] 460627: Extracting information
ERROR: Unsupported URL: http://www.laola1.tv/de-de/video/460627
Traceback (most recent call last):
  File "youtube_dl\extractor\generic.pyo", line 1289, in _real_extract
  File "youtube_dl\compat.pyo", line 248, in compat_etree_fromstring
  File "youtube_dl\compat.pyo", line 237, in _XML
  File "xml\etree\ElementTree.pyo", line 1642, in feed
  File "xml\etree\ElementTree.pyo", line 1506, in _raiseerror
ParseError: not well-formed (invalid token): line 57, column 32
Traceback (most recent call last):
  File "youtube_dl\YoutubeDL.pyo", line 665, in extract_info
  File "youtube_dl\extractor\common.pyo", line 312, in extract
  File "youtube_dl\extractor\generic.pyo", line 1908, in _real_extract
UnsupportedError: Unsupported URL: http://www.laola1.tv/de-de/video/460627
~~~

(run on windows 10)

blackwinter:
This is fixed in 2016.02.22.

egolus:
it works for videos now, but livestreams still error out

example livestream right now: http://www.laola1.tv/de-de/livestream/2016-02-23-cska-moscow-hc-slovan-bratislava-german
~~~
[...]

PR #8478, opened 2016-02-09
[laola1tv] Fixes for changed site layout.

blackwinter:
* Fixed valid URLs (w/ tests).
* Fixed iframe URL extraction.
* Fixed token URL extraction.
* Fixed variable extraction.
* Fixed uploader spelling.
* Added upload_date to result dictionary.

Addresses #8301.

### DIFF
--- a/youtube_dl/extractor/laola1tv.py
+++ b/youtube_dl/extractor/laola1tv.py
@@ -1,86 +1,113 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-import random
 import re
 
 from .common import InfoExtractor
 from ..utils import (
+    compat_urlparse,
     ExtractorError,
+    sanitized_Request,
+    xpath_element,
     xpath_text,
+    unified_strdate,
+    urlencode_postdata,
 )
 
 
 class Laola1TvIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?laola1\.tv/(?P<lang>[a-z]+)-(?P<portal>[a-z]+)/.*?/(?P<id>[0-9]+)\.html'
-    _TEST = {
+    _VALID_URL = r'https?://(?:www\.)?laola1\.tv/(?P<lang>[a-z]+)-(?P<portal>[a-z]+)/.*?/(?P<slug>[\w-]+)'
+    _TESTS = [{
         'url': 'http://www.laola1.tv/de-de/video/straubing-tigers-koelner-haie/227883.html',
         'info_dict': {
-            'id': '227883',
-            'ext': 'mp4',
-            'title': 'Straubing Tigers - Kölner Haie',
             'categories': ['Eishockey'],
+            'ext': 'flv',
+            'id': '227883',
             'is_live': False,
+            'title': 'Straubing Tigers - Kölner Haie',
+            'upload_date': '20140912',
         },
         'params': {
             'skip_download': True,
         }
-    }
+    }, {
+        'url': 'http://www.laola1.tv/de-de/video/straubing-tigers-koelner-haie',
+        'info_dict': {
+            'categories': ['Eishockey'],
+            'ext': 'flv',
+            'id': '464602',
+            'is_live': False,
+            'title': 'Straubing Tigers - Kölner Haie',
+            'upload_date': '20160129',
+        },
+        'params': {
+            'skip_download': True,
+        }
+    }]
 
     def _real_extract(self, url):
         mobj = re.match(self._VALID_URL, url)
-        video_id = mobj.group('id')
         lang = mobj.group('lang')
         portal = mobj.group('portal')
 
-        webpage = self._download_webpage(url, video_id)
+        webpage = self._download_webpage(url, mobj.group('slug'))
         iframe_url = self._search_regex(
-            r'<iframe[^>]*?class="main_tv_player"[^>]*?src="([^"]+)"',
+            r'<iframe[^>]*?id="videoplayer"[^>]*?src="([^"]+)"',
             webpage, 'iframe URL')
 
-        iframe = self._download_webpage(
-            iframe_url, video_id, note='Downloading iframe')
-        flashvars_m = re.findall(
-            r'flashvars\.([_a-zA-Z0-9]+)\s*=\s*"([^"]*)";', iframe)
-        flashvars = dict((m[0], m[1]) for m in flashvars_m)
+        video_id = self._search_regex(
+            r'videoid=(\d+)', iframe_url, 'video ID')
+
+        iframe = self._download_webpage(compat_urlparse.urljoin(
+            url, iframe_url), video_id, note='Downloading iframe')
 
         partner_id = self._search_regex(
-            r'partnerid\s*:\s*"([^"]+)"', iframe, 'partner id')
+            r'partnerid\s*:\s*"([^"]+)"', iframe, 'partner ID')
 
         xml_url = ('http://www.laola1.tv/server/hd_video.php?' +
                    'play=%s&partner=%s&portal=%s&v5ident=&lang=%s' % (
                        video_id, partner_id, portal, lang))
         hd_doc = self._download_xml(xml_url, video_id)
 
-        title = xpath_text(hd_doc, './/video/title', fatal=True)
-        flash_url = xpath_text(hd_doc, './/video/url', fatal=True)
-        uploader = xpath_text(hd_doc, './/video/meta_organistation')
-        is_live = xpath_text(hd_doc, './/video/islive') == 'true'
+        _v = lambda x, **k: xpath_text(hd_doc, './/video/' + x, **k)
+        title = _v('title', fatal=True)
 
-        categories = xpath_text(hd_doc, './/video/meta_sports')
+        categories = _v('meta_sports')
         if categories:
             categories = categories.split(',')
 
-        ident = random.randint(10000000, 99999999)
-        token_url = '%s&ident=%s&klub=0&unikey=0&timestamp=%s&auth=%s' % (
-            flash_url, ident, flashvars['timestamp'], flashvars['auth'])
+        time_date = _v('time_date')
+        time_start = _v('time_start')
+        upload_date = None
+        if time_date and time_start:
+            upload_date = unified_strdate(time_date + ' ' + time_start)
 
+        json_url = ('https://club.laola1.tv/sp/laola1/api/v3/user/session' +
+                    '/premium/player/stream-access?videoId=%s&target=2' +
+                    '&label=laola1tv&area=%s') % (video_id, _v('area'))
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/laola1tv.py
```diff
@@ -25,27 +27,60 @@ class Laola1TvIE(InfoExtractor):
...
 
+        _x = lambda x: re.search(x + r'=(\w+)', iframe_url).group(1)
+        if (video_id is None):
+            video_id = _x('videoid')
+        if (lang is None):
+            lang = _x('lang')
+        if (portal is None):
+            portal = _x('portal')
+
+        iframe_url = urlparse.urljoin(url, iframe_url)
```
dstftw: Use `compat_urlparse.urljoin` instead.

youtube_dl/extractor/laola1tv.py
```diff
@@ -54,16 +89,17 @@ def _real_extract(self, url):
...
+        uploader = xpath_text(hd_doc, './/video/meta_organisation')
         is_live = xpath_text(hd_doc, './/video/islive') == 'true'
 
         categories = xpath_text(hd_doc, './/video/meta_sports')
         if categories:
             categories = categories.split(',')
 
-        ident = random.randint(10000000, 99999999)
+        timestamp = time.strftime('%Y%m%d%H%M%S', time.gmtime())
+        ident = '%d%s' % (random.randint(10000000, 99999999), timestamp)
```
dstftw: Correct player timestamp can be extracted from https://club.laola1.tv/sp/laola1/api/v3/user/session/premium/player/stream-access?videoId=227883&target=2&label=laola1tv&area=ice_hockey_del.

youtube_dl/extractor/laola1tv.py
blackwinter: Done.

youtube_dl/extractor/laola1tv.py
blackwinter: Can you be more specific? I get "Page not found!" for this link.

dstftw: See firebug/developer tools.

blackwinter: I appreciate your assistance, but I still don't understand how I can get the timestamp from the URL you gave.

dstftw: When requested with proper headers/cookies it returns JSON:
```json
{"status":"error","message":"Please log in to get access to this content!","data":{"stream-access":["https
:\/\/streamaccess.unas.tv\/hdflash2\/vod\/22\/471029.xml?streamid=471029&partnerid=22&label=laola1tv
&area=ice_hockey_del&ident=2796309820160209163815&klub=0&unikey=0&timestamp=20160209163815&auth=f7a094f27a0946900ad049524a9e45f3"
,"https:\/\/streamaccess.unasmedia.tv\/hdflash2\/vod\/22\/471029.xml?streamid=471029&partnerid=22&label
=laola1tv&area=ice_hockey_del&ident=2796309820160209163815&klub=0&unikey=0&timestamp=20160209163815&auth
=f7a094f27a0946900ad049524a9e45f3"]},"code":5}
```

blackwinter: Thanks, I think I got it. This also seems to solve the recurring token error.

blackwinter:
I cleaned the patch up some more and consider it done now. There are certainly things that can be improved (I'm not too familiar with Python), but I'm using it already and the tests pass locally (Python 2.7.10).

CI is another matter. Python 2.6 fails completely (`AttributeError: 'NoneType' object has no attribute 'attrib'`), but that was already the case before these changes. On all other versions the first two tests pass, the third fails due to some encoding issue (`AssertionError: Invalid value for field title, expected u'Straubing Tigers - K\xf6lner Haie', got u'Straubing Tigers - Kolner Haie'`).

blackwinter:
I discovered two more CI errors:

0. `AssertionError: Laola1TvIE should match URL u'http://www.laola1.tv/watch/video/464602'` in `test_no_duplicates`. This is because `/watch/...` is a redirect; I simplified `_VALID_URL` to only match the target URL (same with `/video/<slug>/<id>.html`).
0. `AssertionError: Invalid value for field categories, expected 'Eishockey', got 'Ice Hockey'` in `test_Laola1Tv_2`. This seems to be caused by `lang=en` which the redirect might infer from geolocation; at least, that would explain why the test is passing for me locally. This would also explain the different title mentioned in my previous comment.

Should I try to do anything about these errors?

dstftw:
1. Shouls be fixed either by fixing `_VALID_URL`, changing the test case or removing it. If you are sure this is always a redirect replace with `http://www.laola1.tv/en-int/video/464602`. Also switch it to `'only_matching': True`.
2. Replace with `list`.
3. To make extraction not to fail on python 2.6 use `xpath_element` for xpath element extraction.

youtube_dl/extractor/laola1tv.py
```diff
@@ -72,15 +101,15 @@ def _real_extract(self, url):
...
-            'title': title,
-            'url': video_url,
-            'uploader': uploader,
             'categories': categories,
             'ext': 'mp4',
+            'id': video_id,
+            'is_live': _v('islive') == 'true',
+            'title': title,
+            'upload_date': unified_strdate(
+                _v('time_date') + ' ' + _v('time_start')),
```
dstftw: This will fail if any of time fields is missing.

blackwinter: Done.

blackwinter:
> Shouls be fixed either by fixing `_VALID_URL`, changing the test case or removing it. If you are sure this is always a redirect replace with `http://www.laola1.tv/en-int/video/464602`. Also switch it to `'only_matching': True`.

I wanted to express that `/watch/...` will work, too. But as it's not really needed, I just removed the test.

> Replace with `list`.

What exactly should I replace?

> To make extraction not to fail on python 2.6 use `xpath_element` for xpath element extraction.

I've used `xpath_attr` instead.

youtube_dl/extractor/laola1tv.py
```diff
@@ -1,86 +1,109 @@
...
+        json_url = ('https://club.laola1.tv/sp/laola1/api/v3/user/session' +
+                    '/premium/player/stream-access?videoId=%s&target=2' +
+                    '&label=laola1tv&area=%s') % (video_id, _v('area'))
+        req = sanitized_Request(json_url, urlencode_postdata(
+            dict((i, v) for i, v in enumerate(_v('req_liga_abos').split(',')))))
+
+        token_url = self._download_json(req, video_id)['data']['stream-access'][0]
         token_doc = self._download_xml(
             token_url, video_id, note='Downloading token')
-        token_attrib = token_doc.find('.//token').attrib
```
dstftw: `xpath_element(token_doc, './/token').attrib`.

youtube_dl/extractor/laola1tv.py
```diff
@@ -1,86 +1,109 @@
...
-        uploader = xpath_text(hd_doc, './/video/meta_organistation')
-        is_live = xpath_text(hd_doc, './/video/islive') == 'true'
+        _v = lambda x, **k: xpath_text(hd_doc, './/video/' + x, **k)
+        title = _v('title', fatal=True)
 
-        categories = xpath_text(hd_doc, './/video/meta_sports')
-        if categories:
-            categories = categories.split(',')
+        categories = _v('meta_sports')
+        if categories: categories = categories.split(',')
```
dstftw: This is not pythonic to place it on a single line. Check code with flake8.

youtube_dl/extractor/laola1tv.py
```diff
@@ -1,86 +1,109 @@
...
+            video_url = '%s?hdnea=%s&hdcore=3.2.0' % (token_url, token_auth)
 
         return {
+            'categories': categories,
+            'ext': 'mp4',
             'id': video_id,
-            'is_live': is_live,
+            'is_live': _v('islive') == 'true',
             'title': title,
+            'upload_date': upload_date,
```
dstftw: This will result in reference to unassigned variable when time fields are missing.

youtube_dl/extractor/laola1tv.py
blackwinter: Done. Now passes flake8.

youtube_dl/extractor/laola1tv.py
blackwinter: Done, I think.

dstftw: Same with `video_url`.

youtube_dl/extractor/laola1tv.py
```diff
@@ -1,86 +1,114 @@
...
-        video_url = '%s?hdnea=%s&hdcore=3.2.0' % (
-            token_attrib['url'], token_attrib['auth'])
+        token_attrib = xpath_element(token_doc, './/token').attrib
+        token_auth = token_attrib['auth']
+        video_url = token_attrib['url']
+
+        if token_auth in ('blocked', 'restricted'):
+            raise ExtractorError(
+                'Token error: %s' % token_attrib['comment'], expected=True)
+        elif token_auth and video_url:
```
dstftw: This check makes no sense.

youtube_dl/extractor/laola1tv.py
```diff
@@ -1,86 +1,114 @@
...
         return {
+            'categories': categories,
+            'ext': 'mp4',
             'id': video_id,
-            'is_live': is_live,
+            'is_live': _v('islive') == 'true',
             'title': title,
+            'upload_date': upload_date,
+            'uploader': _v('meta_organisation'),
             'url': video_url,
```
dstftw: Instead of `video_url` there should be `formats` extracted with `self._extract_f4m_formats`.

blackwinter: Done.

youtube_dl/extractor/laola1tv.py
blackwinter: This would fail if `auth` or `url` fields are missing. But I changed it.

youtube_dl/extractor/laola1tv.py
blackwinter: Done.

youtube_dl/extractor/laola1tv.py
dstftw: If will fail anyway if any of these fields is missing and it's better to fail early explicitly in extractor than in some generic code.

dstftw:
Cherry-picked in 5e19323ed91a2ac3014c8f803fa65d2774a6a4f4, thanks.

blackwinter:
Thank you, Sergey!

beta2k:
is there a release somewhere where laola1tv works? please provide me with a link...

blackwinter:
As of today there is. Just run `youtube-dl -U`.

PR #8479, opened 2016-02-09
[downloader/dash] Implement dashsegments fd in terms of fragment fd

remitamine:


dstftw:
LGTM.

Noctem:
Awesome, this seems to have fixed the main issues I had with some YouTube videos.

Some background: it seems that YouTube doesn't actually generate the DASH segments until they're being viewed for the first time (someone correct me if I'm wrong), so on videos with 0 views my download would almost always reach a segment before YouTube has it and I'd get a 404. So I have a script that would just keep repeating until it succeeds, but that involved repeatedly downloading the same chunks until it reaches new segments, 404s again, and has to start over. This meant that downloading a single video would often take 20 or more attempts.

So anyways, this change has made that situation a lot better in that:
- it doesn't appear to be 404ing
  - I'm guessing that it actually is, but it's just transparently retrying those segments?
- it's able to properly resume downloads if there is a failure
- it isn't spamming my terminal with _"downloading segment **x** of **y**"_ hundreds of times.
- is providing output size and remaining time estimates

The only drawback I've noticed is that it seems to be keeping all of the segments both individually and appended together, so the storage required during the download is doubled. Apparently they're [looped over and deleted as soon as the download is complete](https://github.com/rg3/youtube-dl/pull/8479/files#diff-361ab33b924d01350c4c4bca2bc9a6b2R56), but ideally the temporary storage required for the download wouldn't be twice the size of the video, especially since I usually download to a RAMDisk before processing/remuxing. Is there any reason (resumability perhaps?) that each segment couldn't be deleted immediately after it's appended to the main file?

Anyways, thanks. This commit will save me a lot of time and bandwidth.

PR #8487, opened 2016-02-09
Added support for hotstar url

ashutosh2k12:
Added support for hotstar short url(url by id)

### DIFF
--- a/youtube_dl/extractor/hotstar.py
+++ b/youtube_dl/extractor/hotstar.py
@@ -10,8 +10,8 @@ from ..utils import (
 
 
 class HotStarIE(InfoExtractor):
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/hotstar.py
```diff
@@ -10,8 +10,8 @@
 
 
 class HotStarIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?hotstar\.com/.*?[/-](?P<id>\d{10})'
-    _TEST = {
+    _VALID_URL = r'https?://(?:www\.)?hotstar\.com/.*?(?P<id>\d{10})|[/-](?P<iq>\d{10})'
```
dstftw: `iq` will never match.

PR #8505, opened 2016-02-10
[screenjunkies] Add new extractor

TingPing:


youtube_dl/extractor/screenjunkies.py
```diff
@@ -0,0 +1,71 @@
...
+import re
+import json
+
+from .common import (
+    InfoExtractor,
+    ExtractorError
+)
+
+class ScreenJunkiesIE(InfoExtractor):
+    _VALID_URL = r'http://www.screenjunkies.com/video/(?P<id>.*)'
```
dstftw: `id` must match at least one character.

youtube_dl/extractor/screenjunkies.py
```diff
@@ -0,0 +1,71 @@
...
+            'title': "Honest Trailers - 'The Dark Knight'",
+        },
+    }, {
+        'url': 'http://www.screenjunkies.com/video/knocking-dead-ep-1-the-show-so-far-3003285',
+        'only_matching': True,
+    }]
+
+    def _real_extract(self, url):
+        video_link = self._match_id(url)
+        video_id = self._search_regex(r'.*-(\d+)', url, 'id', default=None, fatal=False)
```
dstftw: `id` should be extracted from `_VALID_URL` with `self._match_id`.

youtube_dl/extractor/screenjunkies.py
```diff
@@ -0,0 +1,71 @@
...
+
+        if not video_id: # Older urls didn't have the id in them, but we can grab it manually
+            webpage = self._download_webpage(url, video_link)
+            video_id = self._search_regex(r'src="/embed/(\d+)"', webpage, 'video id')
+
+        webpage = self._download_webpage('http://www.screenjunkies.com/embed/%s' %video_id, video_id)
+        video_vars_str = self._html_search_regex(r'embedVars = (\{.+\})\s*</script>', webpage, 'video variables', flags=re.S)
+        video_vars = json.loads(video_vars_str)
```
dstftw: Use `_parse_json` instead.

youtube_dl/extractor/screenjunkies.py
```diff
@@ -0,0 +1,71 @@
...
+
+        # TODO: Figure out required cookies
+        if video_vars['subscriptionLevel'] > 0:
+            raise ExtractorError('This video requires ScreenJunkiesPlus')
```
dstftw: `expected=True`.

youtube_dl/extractor/screenjunkies.py
TingPing: Not all URLs have the id included though.

youtube_dl/extractor/screenjunkies.py
```diff
@@ -0,0 +1,71 @@
...
+                'width': f['width'],
+                'height': f['height'],
+                'format': 'mp4',
+            })
+        formats.sort(key=lambda d: d['tbr'])
+
+        return {
+            'id': video_id,
+            'title': video_vars['contentName'],
+            'thumbnail': video_vars['thumbUri'],
```
dstftw: Extraction should be tolerate to missing fields. Everything apart from title and formats considered optional and should not break the extraction if missing.

youtube_dl/extractor/screenjunkies.py
```diff
@@ -0,0 +1,71 @@
...
+            'duration': video_vars['videoLengthInSeconds'],
```
dstftw: `int_or_none`.

youtube_dl/extractor/screenjunkies.py
```diff
@@ -0,0 +1,71 @@
...
+            raise ExtractorError('This video requires ScreenJunkiesPlus')
+
+        formats = []
+        for f in video_vars['media']:
+            if not f['mediaPurpose'] == 'play':
+                continue
+
+            formats.append({
+                'url': f['uri'],
+                'tbr': f['bitRate'],
```
dstftw: `int_or_none`.

youtube_dl/extractor/screenjunkies.py
```diff
@@ -0,0 +1,71 @@
...
+                'width': f['width'],
```
dstftw: `int_or_none`.

youtube_dl/extractor/screenjunkies.py
```diff
@@ -0,0 +1,71 @@
...
+                'height': f['height'],
```
dstftw: `int_or_none`.

youtube_dl/extractor/screenjunkies.py
```diff
@@ -0,0 +1,71 @@
...
+            })
+        formats.sort(key=lambda d: d['tbr'])
+
+        return {
+            'id': video_id,
+            'title': video_vars['contentName'],
+            'thumbnail': video_vars['thumbUri'],
+            'duration': video_vars['videoLengthInSeconds'],
+            'formats': formats,
+            'tags': video_vars['tags'],
```
dstftw: Must be list, not a string.

youtube_dl/extractor/screenjunkies.py
dstftw: >when present.

dstftw:
Cherry-picked in b9e8d7140aa8584f8516c3575c20b21172278fef, thanks.

TingPing:
> Do you want to be credited as one of the authors of youtube-dl? If so under what name?

sure: Patrick Griffis

TingPing:
@dstftw Also thanks a ton for getting the Plus only videos working!

dstftw:
Done in fe80df30804736db5e644ad15ac09149c57ddf33.

Issue #8325, opened 2016-01-26
bad working with a particular URL


dufoin:
Hi,

the following link entered in the terminal window returns :
ERROR: Unsupported URL: http://replay.publicsenat.fr/vod/documentaire/a-la-recherche-de-bernard-maris,-l-anti-economiste/192668
What is your advice ? thanks

db

dufoin:
OK, very good indeed !
thanks

PR #8521, opened 2016-02-11
[PublicSenat] add new extractor

crazyiop:
solve #8325


dstftw:
This should be handled by generic extractor.

PR #8537, opened 2016-02-13
[Canvas] Add subtitles

RobinHoutevelts:
This PR adds the subtitle functionality of shows at canvas.be

### DIFF
--- a/youtube_dl/extractor/canvas.py
+++ b/youtube_dl/extractor/canvas.py
@@ -53,6 +53,15 @@ class CanvasIE(InfoExtractor):
                     'url': format_url,
                 })
         self._sort_formats(formats)
+        
+        subtitles = {}
+        for target in data['subtitleUrls']:
+            format_url, format_type = target.get('url'), target.get('type')
+            if format_type == 'CLOSED':
+                subtitles['nl'] = [{
+                    'ext': 'vtt',
+                    'url': format_url,
+                }]
 
         return {
             'id': video_id,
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/canvas.py
```diff
@@ -53,6 +53,15 @@ def _real_extract(self, url):
                     'url': format_url,
                 })
         self._sort_formats(formats)
+        
+        subtitles = {}
+        for target in data['subtitleUrls']:
```
dstftw: Extraction must tolerate missing `subtitleUrls`.

youtube_dl/extractor/canvas.py
```diff
@@ -53,6 +53,15 @@ def _real_extract(self, url):
...
+            format_url, format_type = target.get('url'), target.get('type')
```
dstftw: If there is no `url` you'll end up with invalid entry in `subtitles`.

youtube_dl/extractor/canvas.py
RobinHoutevelts: Even when it's _always_ there?
for example: https://mediazone.vrt.be/api/v1/canvas/assets/mz-ast-a30df0cf-590c-4c4f-b1fe-7c28fc3b7b8c

dstftw: Today it's there - tomorrow it's not. Extractors must be as robust as possible and tolerate to layout changes. Subtitles is not a critical metadata to extract thus when `subtitleUrls` disappears (if ever) if must not break the main extraction routine.

RobinHoutevelts: Seems unlikely but reasonable, I'll update and squash the commits tomorrow ( it's 5am, I'm off to bed )
Thanks reviewing the code

PR #8550, opened 2016-02-13
fix netzkino extractor for subcategory urls

dingensundso:
Some movies on netzkino are in subcategories, but the api url is only for main category. E.g. http://www.netzkino.de/#!/science-fiction-kinoab18/downstream will not load with current extractor. This PR fixes this by splitting the category and taking only the last part.

### DIFF
--- a/youtube_dl/extractor/netzkino.py
+++ b/youtube_dl/extractor/netzkino.py
@@ -36,7 +36,7 @@ class NetzkinoIE(InfoExtractor):
 
     def _real_extract(self, url):
         mobj = re.match(self._VALID_URL, url)
-        category_id = mobj.group('category').split('-')[-1]
+        category_id = mobj.group('category').split('-')[0]
         video_id = mobj.group('id')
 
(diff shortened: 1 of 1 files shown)


dstftw:
1. There is another URL format: http://www.netzkino.de/#!/kinoab18-all/downstream.
2. Tests should be added.

dingensundso:
Oh. Just noticed the URL I provided doesn't work with my fix. Before I was using a movie from the "Prickelkino" category.
So there are different formats

1. http://www.netzkino.de/#!/kinoab18-all/downstream
   where the main-category is in front of the dash

2. http://www.netzkino.de/#!/prickelkino-kinoab18/night-eyes-iii-ov
  where it's the other way around.

Interestingly the first URL works only with the first part `kinoab18` as category, but the second works with either `prickelkino` or `kinoab18` but not with `prickelkino-kinoab18`. So I changed my PR.

dstftw:
http://www.netzkino.de/#!/science-fiction-kinoab18/downstream does not work now.

dingensundso:
Dann. When I have some time, I'll scrape their entire website go find out wjoxh different formats they use.

PR #8574, opened 2016-02-16
[uStudio] Add new extractor

mutantmonkey:


dstftw:
There are more qualities available.

youtube_dl/extractor/ustudio.py
```diff
@@ -0,0 +1,56 @@
...
+                video_id),
+            video_id,
+            note='Downloading video info',
+            errnote='Failed to download video info')
+
+        formats = [
+            {
+                'url': quality.attrib['url'],
+                'width': int(quality.attrib['width']),
+                'height': int(quality.attrib['height']),
```
dstftw: extraction must be tolerate to missing fields.

youtube_dl/extractor/ustudio.py
```diff
@@ -0,0 +1,56 @@
...
+        return {
+            'id': video_id,
+            'title': self._og_search_title(webpage),
+            'thumbnail': self._og_search_thumbnail(webpage),
+            'formats': formats,
+            'description': self._og_search_description(webpage),
+            'uploader': self._html_search_regex(
+                r'<a href=".*/user/.*">(.*)</a> on',
+                webpage,
+                'uploader'),
```
dstftw: This should be `fatal=False`.

youtube_dl/extractor/ustudio.py
```diff
@@ -0,0 +1,56 @@
...
+        self._sort_formats(formats)
+
+        return {
+            'id': video_id,
+            'title': self._og_search_title(webpage),
+            'thumbnail': self._og_search_thumbnail(webpage),
+            'formats': formats,
+            'description': self._og_search_description(webpage),
+            'uploader': self._html_search_regex(
+                r'<a href=".*/user/.*">(.*)</a> on',
```
dstftw: It should match at least one character.

dstftw:
Cherry-picked, thanks. Do you want to be credited as one of the authors of youtube-dl? If so under what name?

PR #8575, opened 2016-02-16
[KUSI] Add new extractor

mutantmonkey:


youtube_dl/extractor/kusi.py
```diff
@@ -0,0 +1,69 @@
...
+
+        quality_options = doc.find('{http://search.yahoo.com/mrss/}group').findall('{http://search.yahoo.com/mrss/}content')
+
+        QUALITIES = ['180', '360']
+        formats = []
+        for quality in quality_options:
+            if 'height' in quality.attrib:
+                format_id = quality.attrib['height']
+                formats.append({
+                    'url': compat_urllib_parse.unquote_plus(quality.attrib['url']),
```
yan12125: Use ```compat_urllib_parse_unquote_plus``` instead.

youtube_dl/extractor/kusi.py
```diff
@@ -0,0 +1,69 @@
...
+                    'format_id': format_id,
+                    'preference': QUALITIES.index(format_id) if format_id in QUALITIES else -1,
```
yan12125: Use ```height``` attribute is better here. ```_sort_formats``` can sort formats by heights.

youtube_dl/extractor/kusi.py
```diff
@@ -0,0 +1,69 @@
...
+
+        xml_url = 'http://www.kusi.com/build.asp?buildtype=buildfeaturexml'\
+                  'request&featureType=Clip&featureid={}&affiliateno=956&'\
+                  'clientgroupid=1&rnd=562461'.format(video_id)
+        doc = self._download_xml(xml_url, video_id,
+                                 note='Downloading video info',
+                                 errnote='Failed to download video info')
+
+        video_title = doc.find('HEADLINE').text
+        duration = int(doc.find('DURATION').text)
```
yan12125: ```duration``` is optional. This line shouldn't fail if doc does not contain a ```DURATION``` element.

youtube_dl/extractor/kusi.py
```diff
@@ -0,0 +1,69 @@
...
+
+        if mobj.group('clipId') is not None:
+            video_id = mobj.group('clipId')
+        else:
+            webpage = self._download_webpage(url, mobj.group('path'))
+            video_id = self._html_search_regex(r'"clipId", "(\d+)"', webpage,
+                                               'clipId')
+
+        xml_url = 'http://www.kusi.com/build.asp?buildtype=buildfeaturexml'\
+                  'request&featureType=Clip&featureid={}&affiliateno=956&'\
```
yan12125: Unnumbered placeholders are not supported in Python 2.6.

youtube_dl/extractor/kusi.py
```diff
@@ -0,0 +1,69 @@
...
+        'info_dict': {
+            'id': '12203019',
+            'ext': 'mp4',
+            'title': 'Turko Files: Case Closed! & Put On Hold!',
+            'duration': 231000,
+        }
+    }
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
```
yan12125: Use ```_sort_formats```.

yan12125:
Thanks merged. Would you like to be listed in [AUTHORS](https://github.com/rg3/youtube-dl/blob/master/AUTHORS)? If so under what name?

mutantmonkey:
Sure, please list me as "mutantmonkey"

PR #8577, opened 2016-02-16
Fixed dailymotion view_count parsing

cazulu:
Fixed dailymotion view_count parsing when the decimal marker is a whitespace, e.g., '101 101'

### DIFF
--- a/youtube_dl/extractor/dailymotion.py
+++ b/youtube_dl/extractor/dailymotion.py
@@ -122,10 +122,13 @@ class DailymotionIE(DailymotionBaseInfoExtractor):
         description = self._og_search_description(webpage) or self._html_search_meta(
             'description', webpage, 'description')
 
-        view_count = str_to_int(self._search_regex(
+        view_count_str = self._search_regex(
             [r'<meta[^>]+itemprop="interactionCount"[^>]+content="UserPlays:(\d+)"',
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/dailymotion.py
```diff
@@ -122,10 +122,13 @@ def _real_extract(self, url):
         description = self._og_search_description(webpage) or self._html_search_meta(
             'description', webpage, 'description')
 
-        view_count = str_to_int(self._search_regex(
+        view_count_str = self._search_regex(
             [r'<meta[^>]+itemprop="interactionCount"[^>]+content="UserPlays:(\d+)"',
```
dstftw: This regex should also be fixed.

youtube_dl/extractor/dailymotion.py
```diff
@@ -122,10 +122,13 @@ def _real_extract(self, url):
...
-             r'video_views_count[^>]+>\s+([\d\.,]+)'],
-            webpage, 'view count', fatal=False))
+             r'video_views_count[^>]+>\s+([\d\.,\s]+)'],
+            webpage, 'view count', fatal=False)
+        if view_count_str and view_count_str.split:
+            view_count_str = ''.join(view_count_str.split())
```
dstftw: Just use `replace` or `re.sub`.

PR #8610, opened 2016-02-19
[mtv] Extract duration from each playlist item

wiiaboo:
RSS used instead of manifest files because it's exact to the millisecond
with the video I tested while in manifest it's only exact to the second.

### DIFF
--- a/youtube_dl/extractor/mtv.py
+++ b/youtube_dl/extractor/mtv.py
@@ -11,6 +11,7 @@ from ..utils import (
     ExtractorError,
     find_xpath_attr,
     fix_xml_ampersands,
+    float_or_none,
     HEADRequest,
     sanitized_Request,
     unescapeHTML,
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/mtv.py
```diff
@@ -158,6 +158,12 @@ def _get_video_info(self, itemdoc):
         if mtvn_id_node is not None:
             mtvn_id = mtvn_id_node.text
 
+        content_el = itemdoc.find('.//{http://search.yahoo.com/mrss/}content')
+        if content_el is not None:
+            duration = float(content_el.attrib['duration'])
+        else:
+            duration = None
```
remitamine: you can use `float_or_none`.

wiiaboo: Instead of `float` or instead of the whole if? In case of the latter doesn't looking for the attribute raise an error?

remitamine:
```
content_el = itemdoc.find(self._xpath_ns('.//content', 'http://search.yahoo.com/mrss/'))
duration = float_or_none(content_el.attrib.get('duration')) if content_el is not None else None
```
or
```
content_el = find_xpath_attr(itemdoc, self._xpath_ns('.//content', 'http://search.yahoo.com/mrss/'), 'duration')
duration = float_or_none(content_el.attrib['duration']) if content_el is not None else None
```

wiiaboo:
OK, I assume it's instead of the `float` since if either `.find` or ´.attrib` fail it'd error out. Now it just outputs null.

PR #8611, opened 2016-02-19
[downloader/external] Add FFmpegFD(fixes #622)

### LINKED ISSUE #622: Download range
Hi,

Is there any way to implement a download range? To download starting from 1m 30 to 1m 50?

Thanks

remitamine:
- replace HlsFD and RtspFD
- add basic support for downloading part of the video or audio


dstftw:
I guess it would be better to also keep the old rtsp code and select downloader based on ffmpeg/mplayer/mpv availability. Same could be done for rtmp if ffmpeg supports it properly enough.

youtube_dl/downloader/__init__.py
```diff
@@ -30,6 +31,9 @@ def get_suitable_downloader(info_dict, params={}):
     protocol = determine_protocol(info_dict)
     info_dict['protocol'] = protocol
 
+    if (info_dict.get('start_time') or info_dict.get('end_time')) and FFmpegFD.supports(info_dict):
```
dstftw: This will prevent downloading http URLs with time pointers without ffmpeg in `PATH`. ffmpeg availability should also be checked here.

youtube_dl/downloader/external.py
```diff
@@ -136,6 +141,66 @@ def _make_cmd(self, tmpfilename, info_dict):
...
+        ffpp.check_version()
+
+        args = [ffpp.executable, '-y']
+
+        start_time = info_dict.get('start_time', 0)
+        if start_time:
+            args += ['-ss', compat_str(start_time)]
+        end_time = info_dict.get('end_time')
+        if end_time:
+            args += ['-t', compat_str(end_time - start_time)]
```
dstftw: `start_time` may be `None` at this point.

youtube_dl/downloader/external.py
```diff
@@ -136,6 +141,66 @@ def _make_cmd(self, tmpfilename, info_dict):
...
+
+    def _call_downloader(self, tmpfilename, info_dict):
+        url = info_dict['url']
+        ffpp = FFmpegPostProcessor(downloader=self)
+        ffpp.check_version()
+
+        args = [ffpp.executable, '-y']
+
+        start_time = info_dict.get('start_time', 0)
+        if start_time:
```
dstftw: Technically, explicit 0 is also valid value for `start_time` but filtered with this check.

youtube_dl/downloader/external.py
dstftw: `end_time` could be < than `start_time` producing negative duration.

youtube_dl/downloader/__init__.py
dstftw: This is also questionable whether we want this behavior by default at all, since it may possibly cause out of sync media in case of DASH when extracted pieces have too approximate boundaries. Have you tried downloading DASH video+audio with this approach?

remitamine:
as i know mpv and mplayer are based on ffmpeg/avconv so if they are available then ffmpeg/avconv is also available.
i'm plaining to add other external downloader that support other protocols(VLC 3.0 will support F4M and MPD it also support time range)(MP4Box support MPD) i can them .
also i'm also have other ideas to improve this part of code:
- add posibility to select multiple extranal downladers and use the first downloader than can handle the link(ex: aria2c,ffmpeg,vlc)(i'm not sure how we can handle extranal-downloader-args in this case).
- add supports_features property(and rename supports to supports-protocol) to downloader so every downloader have a differante set of features(aria2c multi-connection)(ffmpeg,vlc time-range)...

dstftw:
>as i know mpv and mplayer are based on ffmpeg/avconv

Not necessarily, I've an exe of mpv but no ffmpeg/avconv.

In any case lets just do it step by step.

remitamine:
what if i did the same for mpv and mplayer(create an external downloader for them so they can be used for other protocols if the user want).

youtube_dl/downloader/__init__.py
remitamine: > Have you tried downloading DASH video+audio with this approach?

no, i will test it now.

remitamine:
there is a problem with -f fmt it doesn't always work with info_dict['ext'](aac is not an output format for ffmpeg).
ffmpeg depends on the ext from filename to detect what format to use.

youtube_dl/downloader/external.py
remitamine: > echnically, explicit 0 is also valid value for start_time but filtered with this check.

by default ffmpeg will download the video from the start.

youtube_dl/downloader/__init__.py
remitamine: i tested dash video+audio the video and audio are out of sync.
but it works with this command(at least for the tests that i run on youtube):
`ffmpeg -ss start -t end video_url -ss start -t end audio_url -c copy out.ext`

remitamine: > ffmpeg availability should also be checked here.

should work with 99cbe98ce8617c119c2fb6a567b0e6ef7eae8859.

youtube_dl/downloader/external.py
remitamine: > start_time may be None at this point.

fixed with f34294fa0c0097cea7f6388d5d691d5a54950491.
> end_time could be < than start_time producing negative duration.

this should be handled in general may be in YoutubeDL.process_ie_result.

remitamine:
i tried to add support for mpv as an external downloader but there is a problem when using --stream-dump(reported here https://github.com/mpv-player/mpv/issues/2848).
```diff
diff --git a/youtube_dl/downloader/__init__.py b/youtube_dl/downloader/__init__.py
index 67c2840..7047a3a 100644
--- a/youtube_dl/downloader/__init__.py
+++ b/youtube_dl/downloader/__init__.py
@@ -9,6 +9,7 @@ from .dash import DashSegmentsFD
 from .external import (
     get_external_downloader,
     FFmpegFD,
+    MPVFD,
 )
 
 from ..utils import (
@@ -31,8 +32,10 @@ def get_suitable_downloader(info_dict, params={}):
     protocol = determine_protocol(info_dict)
     info_dict['protocol'] = protocol
 
-    if (info_dict.get('start_time') or info_dict.get('end_time')) and FFmpegFD.available() and FFmpegFD.supports(info_dict):
-        return FFmpegFD
+    if (info_dict.get('start_time') or info_dict.get('end_time')):
+        for ed in (MPVFD, FFmpegFD):
+            if ed.available() and ed.supports(info_dict):
+                return ed
 
     external_downloader = params.get('external_downloader')
     if external_downloader is not None:
diff --git a/youtube_dl/downloader/external.py b/youtube_dl/downloader/external.py
index a4fdf1a..63334b8 100644
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -159,6 +159,26 @@ class HttpieFD(ExternalFD):
         return cmd
 
 
+class MPVFD(ExternalFD):
+    available_opt = ['-V']
+
+    @classmethod
+    def supports(cls, info_dict):
+        return info_dict['protocol'] in ('http', 'https', 'ftp', 'ftps', 'm3u8', 'rtsp', 'rtmp', 'mms')
+
+    def _make_cmd(self, tmpfilename, info_dict):
+        url = info_dict['url']
+        cmd = ['mpv', '--stream-dump', tmpfilename]
+        start_time = info_dict.get('start_time') or 0
+        if start_time:
+            cmd += ['--start', compat_str(start_time)]
+        end_time = info_dict.get('end_time')
+        if end_time:
+            cmd += ['--end', compat_str(end_time)]
+        cmd += ['--', url]
+        return cmd
+
+
 class FFmpegFD(ExternalFD):
     @classmethod
     def supports(cls, info_dict):
```

remitamine:
for mplayer m3u8 support was disabled:
```
Playlist parsing disabled for security reasons. Ignoring file.
```

remitamine:
> I guess it would be better to also keep the old rtsp code and select downloader based on ffmpeg/mplayer/mpv availability.

the old rtsp code(based on mpv and mplayer) didn't work at all.
```
python2 __main__.py 'http://dcndigital.ae/#/live/6/dubai-tv'
[dcn:live] 6: Downloading JSON metadata
[dcn:live] 6: Downloading webpage
WARNING: unable to extract m3u8 url; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
[download] Destination: Dubai Al Oula 2016-02-20 11_31-6.unknown_video
do_connect: could not connect to socket
connect: No such file or directory
[mplayer] 0 bytes
[download] 100% of 0.00B
[ffmpeg] There aren't any subtitles to convert
```
```
python2 __main__.py 'http://dcndigital.ae/#/live/6/dubai-tv'
[dcn:live] 6: Downloading JSON metadata
[dcn:live] 6: Downloading webpage
WARNING: unable to extract m3u8 url; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
[download] Destination: Dubai Al Oula 2016-02-20 11_37-6.unknown_video


ERROR: mpv exited with code 2
```

dstftw:
Old code works fine with mms (`http://www.npo.nl/de-nieuwe-mens-deel-1/21-07-2010/WO_VPRO_043706 -f 0`) and I also remember it worked for some rtsp URLs.

remitamine:
> Old code works fine with mms (http://www.npo.nl/de-nieuwe-mens-deel-1/21-07-2010/WO_VPRO_043706 -f 0) and I also remember it worked for some rtsp URLs.

i don't know why it didn't work in the example that i give but it works with ffmpeg(may because mplayer/mpv didn't support live rtsp).

remitamine:
> there is a problem with -f fmt it doesn't always work with info_dict['ext'](aac is not an output format for ffmpeg).
ffmpeg depends on the ext from filename to detect what format to use.

i think overiding `temp_name` and `undo_temp_name` for FFmpegFD to keep the ext as the last part in the filename will fix the problem(something like this `filename.part.ext`).

jaimeMF:
> > there is a problem with -f fmt it doesn't always work with info_dict['ext'](aac is not an output format for ffmpeg).
ffmpeg depends on the ext from filename to detect what format to use.

> i think overiding temp_name and undo_temp_name for FFmpegFD to keep the ext as the last part in the filename will fix the problem(something like this filename.part.ext).

Maybe build a function that in those cases return the appropriate value for the ffmpeg `-f` option. I think it's nice that all extractor use `*.part` and since some files can't be played until ffmpeg has finished (that's what happen for mp4 videos I think) desktop environments won't open them automatically with the video player app.

remitamine:
> Maybe build a function that in those cases return the appropriate value for the ffmpeg -f option.

this is the full list of the ext that have a diffrent format name:
```
{
    "264": "h264",
    "265": "hevc",
    "302": "daud",
    "A64": "a64",
    "aac": "adts",
    "afc": "aiff",
    "aif": "aiff",
    "aifc": "aiff",
    "bmp": "image2",
    "cavs": "cavsvideo",
    "chk": "webm_chunk",
    "dpx": "image2",
    "drc": "dirac",
    "ffmeta": "ffmetadata",
    "flm": "filmstrip",
    "h265": "hevc",
    "isma": "ismv",
    "jls": "image2",
    "jpeg": "image2",
    "jpg": "image2",
    "js": "jacosub",
    "jss": "jacosub",
    "lbc": "ilbc",
    "ljpg": "image2",
    "loas": "latm",
    "m1v": "mpeg1video",
    "m2a": "mp2",
    "m2t": "mpegts",
    "m2ts": "mpegts",
    "m2v": "mpeg2video",
    "m3u8": "hls",
    "m4a": "ipod",
    "mjpg": [
        "mpjpeg",
        "mjpeg"
    ],
    "mka": "matroska",
    "mkv": "matroska",
    "mpa": "mp2",
    "mpg": [
        "mpeg",
        "mpeg1video"
    ],
    "mts": "mpegts",
    "ogv": "ogg",
    "pam": "image2",
    "pbm": "image2",
    "pcx": "image2",
    "pgm": "image2",
    "pgmyuv": "image2",
    "png": "image2",
    "ra": "rm",
    "rco": "g723_1",
    "rcv": "vc1test",
    "rgb": "rawvideo",
    "sf": "ircam",
    "ssa": "ass",
    "sub": "microdvd",
    "tco": "g723_1",
    "thd": "truehd",
    "ts": "mpegts",
    "vc2": "dirac",
    "vtt": "webvtt",
    "wma": [
        "asf",
        "asf_stream"
    ],
    "wmv": [
        "asf",
        "asf_stream"
    ],
    "xml": "webm_dash_manifest",
    "y4m": "yuv4mpegpipe",
    "yuv": "rawvideo"
}
```

jaimeMF:
I would reduce that list to only the values we actually found in our extractors, for example I don't think we need the formats for images files.

remitamine:
there some obvious values that i know we don't need(xml, ffmeta) and also image formats but even for video and audio it's difficult to enumerate the actual formats that the extractors realy need.

remitamine:
i will add this list(i don't if all of them are used in the extractor).
```
{
    "aac": "adts",
    "m4a": "ipod",
    "mka": "matroska",
    "mkv": "matroska",
    "mpg": "mpeg",
    "ogv": "ogg",
    "ts": "mpegts",
    "wma": "asf",,
    "wmv": "asf",
}
```
should i add the code for mpv, mplayer, cvlc in this pull request or keep it for another pull request.

remitamine:
> This is also questionable whether we want this behavior by default at all, since it may potentially cause out of sync media in case of DASH when extracted pieces have too approximate boundaries. Have you tried downloading DASH video+audio with this approach?

the downloader won't be selected  when multiple formats requested 634415ca1791b22ddb8a272266084d8b22199d09.
and the problem with ext has been fixed with a755f82549cb6bffd8ff9545e0b0a4883b7422d2.

youtube_dl/downloader/external.py
```diff
@@ -46,6 +52,10 @@ def exe(self):
         return self.params.get('external_downloader')
 
     @classmethod
+    def available(cls):
+        return check_executable(cls.get_basename(), cls.available_opt)
```
dstftw: I suppose it's better to wrap into list here and store bare option string in `available_opt`. This is also more idiomatic to type it in uppercase.

youtube_dl/downloader/rtsp.py
```diff
@@ -1,45 +0,0 @@
...
-                'downloaded_bytes': fsize,
-                'total_bytes': fsize,
-                'filename': filename,
-                'status': 'finished',
-            })
-            return True
-        else:
-            self.to_stderr('\n')
-            self.report_error('%s exited with code %d' % (args[0], retval))
-            return False
```
dstftw: I've still suggest at least keeping the code of mplayer downloader.

remitamine: > should i add the code for mpv, mplayer, cvlc in this pull request or keep it for another pull request.

i was asking if i can add them in this PR so they can be used for other protocols.

dstftw: Just keep the old `RtspFD` now and replace it with generalized external fd in further PRs in order to have clear vision of changes: what was removed and what was the replacement in a single commit.

youtube_dl/downloader/__init__.py
```diff
@@ -30,14 +31,17 @@ def get_suitable_downloader(info_dict, params={}):
     protocol = determine_protocol(info_dict)
     info_dict['protocol'] = protocol
 
+    if (info_dict.get('start_time') or info_dict.get('end_time')) and FFmpegFD.available() and FFmpegFD.supports(info_dict):
```
dstftw: Some convenience method like `can_download(info_dict)` checking for `available() and supports(info_dict)` would be good (this is used twice in the code).

youtube_dl/downloader/rtsp.py
remitamine: i'm not sure how i will keep it in the current branch.
when i try to rebase choose to edit the first commit it stops me after the changes.
i think if i add the file again the previous commit history will be erased.

dstftw: I guess single commit would be fine. There are not that much changes.

youtube_dl/downloader/external.py
```diff
@@ -46,9 +52,17 @@ def exe(self):
...
+    def available(cls):
+        return check_executable(cls.get_basename(), [cls.AVAILABLE_OPT])
+
+    @classmethod
     def supports(cls, info_dict):
         return info_dict['protocol'] in ('http', 'https', 'ftp', 'ftps')
 
+    @classmethod
+    def can_download(cls, info_dict):
+        return cls.available() and cls.supports()
```
dstftw: You've forgot to pass `info_dict` to `supports()`.

youtube_dl/downloader/rtsp.py
dstftw: Alternatively you can just restore it after PR is merged.

remitamine:
i added back the rtsp downloader.

dstftw:
One thing we should also decide is whether we want [start; end] fragment downloading scenario by default and how we will notify the user that we are downloading a fragment rather that the whole video (something like `--no-playlist` message analogy).
As for me we have the following options:
1. Download fragment by default when possible and output a warning message that is only fragment is being downloaded and suggest using a `--no-fragment-download` option (bad name, will need a better one) to download the whole video (this will obviously require adding this new option).
2. Download whole video by default and output a warning message suggesting using `--fragment-download` to download the whole video.

From my POV for now we can just simply disable downloading fragments keeping the old behavior and move this discussing in separate issue.

youtube_dl/downloader/__init__.py
```diff
@@ -30,14 +32,17 @@ def get_suitable_downloader(info_dict, params={}):
     protocol = determine_protocol(info_dict)
     info_dict['protocol'] = protocol
 
+    if (info_dict.get('start_time') or info_dict.get('end_time')) and FFmpegFD.can_download(info_dict):
```
dstftw: This should go after the `external_downloader` check. Now when `--external-downlaoder` is specified and there is a [start; end] range ffmpeg is used anyway.

remitamine: i think this would be the correct behaviour.
the extractor that support all features needed is the one that will be selected(for now it's the only one that support this feature).

dstftw: I don't think so. If I explicitly specify an `--external-downloader` I do this intentionally and express the interntion to use exactly this extractor thus all the consequences (that the whole video will be downloaded instead of a fragment) are my responsibility.

remitamine:
> From my POV for now we can just simply disable downloading fragments keeping the old behavior and move this discussing in separate issue.

the opened feature request will be a good place.

i will comment the code to select the ffmpeg downloader for video with time range.
this PR started to:
> replace HlsFD and RtspFD
add basic support for downloading part of the video or audio

it's disappointed to end up with almost nothing new.

dstftw:
Well, there is actually lots of new and very useful functionality and we are on the right way - it's only the matter of time when we'll eventually activate it. You're doing great, thanks!

youtube_dl/downloader/__init__.py
remitamine: the feature will be disabled for now.
but for a later time it will be better to add something to specify downloaders priority as it will be cases when multiple downloaders support the same protocols https://github.com/rg3/youtube-dl/pull/8611#issuecomment-186361517.

remitamine:
> we can just simply disable downloading fragments keeping the old behavior

done in da1973a0383ddc8125702e74a88d3a35bbfcf06d.

youtube_dl/downloader/external.py
```diff
@@ -130,12 +152,80 @@ def _make_cmd(self, tmpfilename, info_dict):
...
         cmd = ['http', '--download', '--output', tmpfilename, info_dict['url']]
         for key, val in info_dict['http_headers'].items():
             cmd += ['%s:%s' % (key, val)]
         return cmd
 
+
+class FFmpegFD(ExternalFD):
+    @classmethod
+    def supports(cls, info_dict):
+        return info_dict['protocol'] in ('http', 'https', 'ftp', 'ftps', 'm3u8', 'rtsp', 'rtmp', 'mms') and not info_dict.get('requested_formats')
```
dstftw: `and not info_dict.get('requested_formats')` part should be removed here since it blocks ffmpeg from choosing as downloader via `--external-downloader` when single video/audio media file is requested.

youtube_dl/downloader/external.py
```diff
@@ -130,12 +152,80 @@ def _make_cmd(self, tmpfilename, info_dict):
...
+
+    @classmethod
+    def available(cls):
+        return FFmpegPostProcessor().available
+
+    def _call_downloader(self, tmpfilename, info_dict):
+        url = info_dict['url']
+        ffpp = FFmpegPostProcessor(downloader=self)
+        ffpp.check_version()
```
dstftw: When we have an m3u8 download and no ffmpeg installed we'll get unhandled exception and crash with a traceback. Previously there were an additional check before this line that has been reporting an error and returning quietly. For now lets just restore it and later think how to move these checks outside actual downloaders to phase when we select downloader.

youtube_dl/downloader/external.py
remitamine: the conditions should be moved:
```
    # if (info_dict.get('start_time') or info_dict.get('end_time') and not info_dict.get('requested_formats')) and FFmpegFD.can_download(info_dict):
    #     return FFmpegFD
```

dstftw: Yep.

remitamine:
both error has been fixed with 6ae27bed01111ddd3c533fb94ae187306e7cf940 and 77dea16ac8bb38d735ac6f002eaba7570a67d361.

dstftw:
`--external-downloader-args` are not passed to ffmpeg downloader.

remitamine:
for ffmpeg there are multiple types of arguments(global, for every input stream, output) evey type should be placed differently.

dstftw:
They should be passed at the same place where `--postprocessor-args` are passed to ffmpeg postprocessor.

remitamine:
this part will still executed if ffmpeg was called when it download an m3u8 file or requested explicitly.
should i comment it?
```python
        start_time = info_dict.get('start_time') or 0
        if start_time:
            args += ['-ss', compat_str(start_time)]
        end_time = info_dict.get('end_time')
        if end_time:
            args += ['-t', compat_str(end_time - start_time)]
```

dstftw:
Lets comment it.

remitamine:
done in 2f76e201f948a35350f084c894a67e827a9e5e5d and e72cae45e90e9869815b39d469a4f96ee2fec245.

dstftw:
`.\youtube_dl\downloader\external.py:10:1: F401 'compat_str' imported but unused`
LGTM. I guess this can be merged now.

dstftw:
I've opened an issue for further discussions on range download design: #8851.

PR #8628, opened 2016-02-21
[powerwatch] Add new extractor

Yurifag:


dstftw:
Instead of separate extractor `_VALID_URL` of `XFileShare` extractor should be extended.

PR #8639, opened 2016-02-22
[shush] Add support for Shush.se.

mbirth:


dstftw:

mbirth:
Sorry, didn't read that. (You should add a short note/link to the "Adding support for a new site" section as that was where I was looking at the whole time.)

Issue #8350, opened 2016-01-28
Vimeo: Chokes on --add-metadata


vxbinaca:
```
youtube-dl  --ignore-config --verbose  --download-archive ~/.ytdlarchive --no-overwrites --call-home --continue --write-info-json --write-description --write-thumbnail --merge-output-format mkv--all-subs --sub-format srt --convert-subs srt --write-sub --add-metadata https://vimeo.com/70668043 https://vimeo.com/70666333
```

```
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--ignore-config', u'--verbose', u'--download-archive', u'/home/vxbinaca/.ytdlarchive', u'--no-overwrites', u'--call-home', u'--continue', u'--write-info-json', u'--write-description', u'--write-thumbnail', u'--merge-output-format', u'mkv--all-subs', u'--sub-format', u'srt', u'--convert-subs', u'srt', u'--write-sub', u'--add-metadata', u'https://vimeo.com/70668043', u'https://vimeo.com/70666333']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.01.27
[debug] Python version 2.7.10 - Linux-4.2.0-25-generic-x86_64-with-Ubuntu-15.10-wily
[debug] exe versions: ffmpeg 2.7.5-0ubuntu0.15.10.1, ffprobe 2.7.5-0ubuntu0.15.10.1, rtmpdump 2.4
[debug] Proxy map: {}
[debug] Public IP address: 76.101.221.232
[vimeo] 70668043: Downloading webpage
[vimeo] 70668043: Extracting information
[vimeo] 70668043: Downloading webpage
[vimeo] 70668043: Downloading JSON metadata
[vimeo] 70668043: Downloading m3u8 information
[info] Video description is already present
[info] Video description metadata is already present
[vimeo] 70668043: Thumbnail is already present
[debug] Invoking downloader on u'https://10-lvl3-hls.vimeocdn.com/1453990861-28223b02a7d6053983227f4b64333f85d0240957/01/4133/2/70668043/178317076.mp4.m3u8'
[download] Ask Ash No. 1-70668043.mp4 has already been downloaded
[download] 100% of 9.83MiB
[ffmpeg] Adding metadata to 'Ask Ash No. 1-70668043.mp4'
[...]

PR #8643, opened 2016-02-23
[MDR] Why m3u8_native instead of m3u8?

TheConstructor:
I recently used youtube-dl to download a KiKa-broadcast-recording like this: `youtube-dl -f HLS-3771 "http://www.kika.de/checker-tobi/sendungen/sendung76752.html"`. The resulting file played fine in VLC but would not play in QuickTime Player. I could fix the file by doing `ffmpeg -i Der\ THW-Check-76752.mp4 -c copy -f mp4 -bsf:a aac_adtstoasc Der\ THW-Check-76752.2.mp4`.

I then realized that there are two implementations of m3u8 in [youtube_dl/downloader/hls.py](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/downloader/hls.py). One difference between the two is that the step I took to fix the file is already done while downloading the file.

I locally changed [youtube_dl/extractor/mdr.py](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/mdr.py#L102) to read

                if ext == 'm3u8':
                    url_formats = self._extract_m3u8_formats(
                        video_url, video_id, 'mp4', entry_protocol='m3u8_native',
                        preference=0, m3u8_id='HLS', fatal=False)

and it seems to fix my problem. Also it seems that by using the `hls_prefer_native`-setting clients without `ffmpeg` installed are still supported.

I currently don't have the time to verify this globally at hand, but to me it seems, that extractors generally should be switched from `m3u8_native` to `m3u` and let `hls_prefer_native` do the magic.

### DIFF
--- a/youtube_dl/extractor/mdr.py
+++ b/youtube_dl/extractor/mdr.py
@@ -99,7 +99,7 @@ class MDRIE(InfoExtractor):
                 ext = determine_ext(url_el.text)
                 if ext == 'm3u8':
                     url_formats = self._extract_m3u8_formats(
-                        video_url, video_id, 'mp4', entry_protocol='m3u8_native',
+                        video_url, video_id, 'mp4', entry_protocol='m3u8',
                         preference=0, m3u8_id='HLS', fatal=False)
                 elif ext == 'f4m':
                     url_formats = self._extract_f4m_formats(


remitamine:
`m3u8_native` was used by a lot of extractors and it has some features that ffmpeg doen't support(resume download) and it didn't require external dependency.
the solution is not to make the ffmpeg the default for all extrators while it can handle some of them.
the is to post processing the result file if ffmpeg is available(something like FFmpegFixup*PP classes).

TheConstructor:
@remitamine sound's ok for me. But it will be April before I can start writing a FFmpegFixup*PP class. Then again I wonder why it is a bad idea to switch to `m3u` and change the default value of `hls_prefer_native` after a FFmpegFixup*PP class exists.

dstftw:
Switching to m3u8 will require ffmpeg installed.

TheConstructor:
@dstftw only if `hls_prefer_native` is false. See the switch in [youtube_dl/downloader/__init__.py](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/downloader/__init__.py#L39)

dstftw:
It's `False` by default, average user won't bother with options.

TheConstructor:
@dstftw I agree that it should have a dynamic default. On the same topic: if `hls_prefer_native` was `True` by default [youtube_dl/postprocessor/ffmpeg.py](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/postprocessor/ffmpeg.py#L395) #8350 would have probably taken care of the conversion.

dstftw:
It should not be default at least due to it does not support live streams

PR #8497, opened 2016-02-10
Add experimenta lazy loading of info extractors

jaimeMF:
Inspired by @remitamine's comment (https://github.com/rg3/youtube-dl/issues/3029#issuecomment-181532852)
In my computer the difference is not super spectacular. `youtube-dl --version` goes from 0.6s to 0.3-0.4s, in the case of the zipped executable it goes from 1s to 0.4s; and `youtube-dl 'http://youtube.com/watch?v=BaW_jenozKcj'` goes from 2.1s to 1.6s. On slower devices like the Raspberry Pi (#3029) the difference may be more noticeable.

Since a lot of things may break, it requires to run `make lazy-extractors` first.

PR #8680, opened 2016-02-27
[closertotruth] Add new extractor

stevengos:
Solves issue https://github.com/rg3/youtube-dl/issues/8435

### DIFF
--- a/youtube_dl/extractor/closertotruth.py
+++ b/youtube_dl/extractor/closertotruth.py
@@ -1,0 +1,61 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+import itertools
+import hashlib
+
+from .common import InfoExtractor
+from ..utils import (
+    int_or_none,
+    unified_strdate,
+)
+
+class CloserToTruthIE(InfoExtractor):
+    _VALID_URL = r'http?://(?:www\.)?closertotruth\.com/series/\S+#video-(?P<id>\w+)'
+    _TESTS = [{
+        'url': 'http://closertotruth.com/series/solutions-the-mind-body-problem#video-3688',
+        'md5': '2aa5b8971633d86fe32152827846a5b4',
+        'info_dict': {
+            'id': '3688',
+            'ext': 'mov',
+            'title': 'Solutions to the Mind-Body Problem? -  Dean W.Zimmerman '
+        }
+    },{
+        'url': 'http://closertotruth.com/series/solutions-the-mind-body-problem#video-4048',
+        'md5': 'a3882bb6e453720d8a7a3983f58abd04',
+        'info_dict': {
+            'id': '4048',
+            'ext': 'mov',
+            'title': 'Solutions to the Mind-Body Problem? -  John Searle '
+        }
+    }]
+
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        
+        #compose title for video
+        video_title = self._search_regex(r'<title>(.+) \|.+</title>', webpage, 'video title')
+
+        entry_id = self._search_regex(r'<a href="\S+" id="video-'+video_id+'" data-kaltura="(\w+)">.+<span.+<\/a>', webpage, "video entry_id")
+        interviewee_name = re.sub(r'(<[^>]+>)', '',self._search_regex(r'<a href="\S+" id="video-'+video_id+'" data-kaltura="\w+">(.+)<span.+<\/a>', webpage, "video interviewee_name"))
+
+        video_title = video_title + ' - ' + interviewee_name
+
+        #extract the partner id for kaltura.com
+        p_id = self._search_regex(r'<script src="http://cdnapi\.kaltura\.com/p/(?P<p>\w+)/sp/\w+/\S+/partner_id/\w+"></script>', webpage, "kaltura partner_id")
+        
+        #request video url at kaltura API
+        #from: http://knowledge.kaltura.com/faq/how-retrieve-download-or-streaming-url-using-api-calls
+        api_request_url = 'http://www.kaltura.com/p/'+p_id+'/sp/0/playManifest/entryId/'+entry_id+'/protocol/HTTPS/flavorParamId/0/video.mp4'
+        api_response = self._download_webpage(api_request_url, video_id)
+        
+        video_url =  self._search_regex(r'<media url="(\S+)"', api_response, "video url")
+        
+        return {
+            'url': video_url,
+            'id': video_id,
+            'title': video_title,
+        }
--- a/youtube_dl/extractor/closertotruth.py
+++ b/youtube_dl/extractor/closertotruth.py
@@ -1,13 +1,11 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
-import re
-
 from .common import InfoExtractor
 
 
 class CloserToTruthIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?closertotruth\.com/series/[^#]+#video-(?P<id>\w+)'
+    _VALID_URL = r'https?://(?:www\.)?closertotruth\.com/(series|interviews)/(?:[^#]+#video-)?(?P<id>\d+)'
     _TESTS = [{
         'url': 'http://closertotruth.com/series/solutions-the-mind-body-problem#video-3688',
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/closertotruth.py
```diff
@@ -0,0 +1,61 @@
...
+
+        video_title = video_title + ' - ' + interviewee_name
+
+        #extract the partner id for kaltura.com
+        p_id = self._search_regex(r'<script src="http://cdnapi\.kaltura\.com/p/(?P<p>\w+)/sp/\w+/\S+/partner_id/\w+"></script>', webpage, "kaltura partner_id")
+        
+        #request video url at kaltura API
+        #from: http://knowledge.kaltura.com/faq/how-retrieve-download-or-streaming-url-using-api-calls
+        api_request_url = 'http://www.kaltura.com/p/'+p_id+'/sp/0/playManifest/entryId/'+entry_id+'/protocol/HTTPS/flavorParamId/0/video.mp4'
+        api_response = self._download_webpage(api_request_url, video_id)
```
dstftw: Extraction should be delegated to [kaltura extractor](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/ellentv.py#L81-L83).

youtube_dl/extractor/closertotruth.py
```diff
@@ -0,0 +1,61 @@
...
+            'title': 'Solutions to the Mind-Body Problem? -  John Searle '
+        }
+    }]
+
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        
+        #compose title for video
```
dstftw: Avoid senseless comments.

youtube_dl/extractor/closertotruth.py
```diff
@@ -0,0 +1,61 @@
...
+            'title': 'Solutions to the Mind-Body Problem? -  Dean W.Zimmerman '
+        }
+    },{
+        'url': 'http://closertotruth.com/series/solutions-the-mind-body-problem#video-4048',
+        'md5': 'a3882bb6e453720d8a7a3983f58abd04',
+        'info_dict': {
+            'id': '4048',
+            'ext': 'mov',
+            'title': 'Solutions to the Mind-Body Problem? -  John Searle '
+        }
```
dstftw: If both test the same extraction scenario leave only one.

dstftw:
Check code with flake8 and fix all issues.

stevengos:
how about this? :)

youtube_dl/extractor/closertotruth.py
```diff
@@ -0,0 +1,38 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+
+
+class CloserToTruthIE(InfoExtractor):
+    _VALID_URL = r'http?://(?:www\.)?closertotruth\.com/series/\S+#video-(?P<id>\w+)'
```
dstftw: It should be `https?` not `http?`. Change `\S+` to `[^#]+`.

youtube_dl/extractor/closertotruth.py
```diff
@@ -0,0 +1,38 @@
...
+
+        entry_id = self._search_regex(r'<a href="\S+" id="video-' + video_id + '" data-kaltura="(\w+)">.+<span.+<\/a>', webpage, "video entry_id")
+        interviewee_name = re.sub(r'(<[^>]+>)', '', self._search_regex(r'<a href="\S+" id="video-' + video_id + '" data-kaltura="\w+">(.+)<span.+<\/a>', webpage, "video interviewee_name"))
+
+        video_title = video_title + ' - ' + interviewee_name
+
+        # extract the partner id for kaltura.com
+        p_id = self._search_regex(r'<script src="http://cdnapi\.kaltura\.com/p/(?P<p>\w+)/sp/\w+/\S+/partner_id/\w+"></script>', webpage, "kaltura partner_id")
+
+        return self.url_result('kaltura:%s:%s' % (p_id, entry_id), 'Kaltura', entry_id, video_title)
```
dstftw: Passing title here makes no sense for non playlists since it will be overridden by delegated extractor. If you want to pass metadata that won't be overridden you should return info dict of [`url_transparent` type](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/makertv.py#L28-L31).

youtube_dl/extractor/closertotruth.py
```diff
@@ -0,0 +1,38 @@
...
+
+        video_title = self._search_regex(r'<title>(.+) \|.+</title>', webpage, 'video title')
+
+        entry_id = self._search_regex(r'<a href="\S+" id="video-' + video_id + '" data-kaltura="(\w+)">.+<span.+<\/a>', webpage, "video entry_id")
+        interviewee_name = re.sub(r'(<[^>]+>)', '', self._search_regex(r'<a href="\S+" id="video-' + video_id + '" data-kaltura="\w+">(.+)<span.+<\/a>', webpage, "video interviewee_name"))
+
+        video_title = video_title + ' - ' + interviewee_name
+
+        # extract the partner id for kaltura.com
+        p_id = self._search_regex(r'<script src="http://cdnapi\.kaltura\.com/p/(?P<p>\w+)/sp/\w+/\S+/partner_id/\w+"></script>', webpage, "kaltura partner_id")
```
dstftw: This can be simplified to just something like `<script[^>]+src=["\'].+?partner_id/(\d+)`. Also for beter robustness additional pattern for `"partner_id":1529571` can be added.

youtube_dl/extractor/closertotruth.py
```diff
@@ -0,0 +1,38 @@
...
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        video_title = self._search_regex(r'<title>(.+) \|.+</title>', webpage, 'video title')
+
+        entry_id = self._search_regex(r'<a href="\S+" id="video-' + video_id + '" data-kaltura="(\w+)">.+<span.+<\/a>', webpage, "video entry_id")
```
dstftw: This should be simplified to something like `r'<a[^>]+id="video-%s"[^>]+data-kaltura="([^"]+)' % video_id`

youtube_dl/extractor/closertotruth.py
```diff
@@ -0,0 +1,38 @@
...
+        interviewee_name = re.sub(r'(<[^>]+>)', '', self._search_regex(r'<a href="\S+" id="video-' + video_id + '" data-kaltura="\w+">(.+)<span.+<\/a>', webpage, "video interviewee_name"))
+
+        video_title = video_title + ' - ' + interviewee_name
+
+        # extract the partner id for kaltura.com
```
dstftw: I've already suggested avoiding senseless comments.

youtube_dl/extractor/closertotruth.py
sandervdo: I'm not entirely sure what you mean with your second sentence. Do you mean that we could simply match on `"partner_id":1529571`? Wouldn't that be too generic?

sandervdo:
I've processed your feedback, thanks! I believe everything has been fixed, except for one [comment](https://github.com/rg3/youtube-dl/pull/8680#discussion_r54336580). If you have any other feedback please let us know.

youtube_dl/extractor/closertotruth.py
dstftw: Of course when it appears inside `script`.

dstftw:
There are also different URL schema possible http://closertotruth.com/interviews/1725.
I've already suggested checking code with flake8. Also long lines should be carried.

dstftw:
Episodes: http://closertotruth.com/episodes/how-do-brains-work.

youtube_dl/extractor/closertotruth.py
stevengos: Could you please give more clarification on what you mean by adding an additional pattern for `"partner_id":1529571` ?

stevengos:
The URL http://closertotruth.com/episodes/how-do-brains-work you propose does not contain clear links that point to video's that appear to belong to a kind of playlist.

dstftw:
image

stevengos:
@dstftw thanks for the feedback, is there anything else I could have overseen?

yan12125:

Issue #8644, opened 2016-02-23
add support for LCP replay


rgenoud:
LCP (aka La Chaîne Parlementaire) is the french national assembly TV channel.
It has a replay web site that doesn't seems to be handled by youtube-dl.
e.g. http://www.lcp.fr/emissions/transportez-moi 
could youtube-dl support this web site ?

PR #8682, opened 2016-02-27
[LCP] Add new extractor

rvanbekkum:
Solves issue #8644

Information Extractor for LCP (La Chaîne parlementaire)
Checked against the PEP8 style guide and added tests for two different extraction scenarios.


youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,137 @@
...
+from __future__ import unicode_literals
+from .common import InfoExtractor
+from ..utils import (
+    int_or_none
+)
+
+
+class LcpIE(InfoExtractor):
+    IE_NAME = 'LCP'
+    _VALID_URL = r'https?:\/\/(?:www\.)?lcp\.fr\/(?:[^\/]+/)*(?P<id>[^/]+)'
```
dstftw: No need to escape `/`.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,137 @@
...
+            'ext': 'mp4',
+            'title': 'Politique Matin - Politique matin'
+        }
+    }]
+
+    def _real_extract(self, url):
+        """Extracts the information for a given url and returns it in a dictionary"""
+        display_id = self._match_id(url)
+
+        # Extract the web page
```
dstftw: Avoid senseless comments.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __get_mp4_video_formats(self, media_files_json):
+        """Retrieves all mp4 video formats contained in the media file info"""
+        formats = []
+        mp4_files_json = media_files_json.get('Mp4')
+        if mp4_files_json is not None:
+            for video_info in mp4_files_json:
+                bitrate = int_or_none(video_info.get('Bitrate'))
+                if bitrate is not None:
+                    bitrate /= 1000  # Set bitrate to KBit/s
```
dstftw: Use `scale` parameter of `int_or_none` instead.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,137 @@
...
+        display_id = self._match_id(url)
+
+        # Extract the web page
+        webpage = self._download_webpage(url, display_id)
+
+        # Extract the required info of the media files
+        media_files_info = self.__extract_from_webpage(display_id, webpage)
+        # Some web pages embed videos from other platforms like dailymotion, therefore we pass on these URL
+        if media_files_info is None:
+            return self.url_result(self.__extract_embed_url(webpage))
```
dstftw: If you are not detecting embeds yourself (you does not) you should just delegate to generic extractor `self.url_result(url, 'Generic')`.

rvanbekkum:
Thank you for your feedback. I have tried to process it carefully.

You are right about the `/` not needing to be escaped, so I updated the regular expressions in the extractor accordingly.
Further, I went over all comments in the code and tried to filter out all of the comments that were redundant (as naming of variables and functions might be self-explanatory).
The extractor should now directly delegate to the generic extractor when necessary for detecting embeds on the page.

Finally did some regression testing to verify that my changes in this commit did not break the extractor. Hopefully you are pleased with the adjustments, otherwise let me know.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,122 @@
...
+        """Extracts the media info JSON object for the video for the provided web page."""
+        embed_url = self.__extract_embed_url(webpage)
+        embed_regex = r'(?:[a-zA-Z0-9]+\.)?lcp\.fr/embed/(?P<clip_id>[A-za-z0-9]+)/(?P<player_id>[A-za-z0-9]+)/(?P<skin_name>[^\/]+)'
+
+        clip_id = self._search_regex(embed_regex, embed_url, 'clip id', group='clip_id', fatal=False)
+        player_id = self._search_regex(embed_regex, embed_url, 'player id', group='player_id', fatal=False)
+        skin_name = self._search_regex(embed_regex, embed_url, 'skin name', group='skin_name', fatal=False)
+
+        # Check whether the matches failed, which might be when dealing with other players (e.g., dailymotion stream)
+        if (clip_id is None) or (player_id is None) or (skin_name is None):
```
dstftw: Truth value check is enough here: `if not clip_id or not player ...`. Same for other occurences in the source.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,122 @@
...
+            'id': media_files_info['EntryName'],
+            'title': self._og_search_title(webpage),
+            'formats': video_formats,
+            'thumbnails': video_thumbnails
+        }
+
+    def __extract_from_webpage(self, display_id, webpage):
+        """Extracts the media info JSON object for the video for the provided web page."""
+        embed_url = self.__extract_embed_url(webpage)
+        embed_regex = r'(?:[a-zA-Z0-9]+\.)?lcp\.fr/embed/(?P<clip_id>[A-za-z0-9]+)/(?P<player_id>[A-za-z0-9]+)/(?P<skin_name>[^\/]+)'
```
dstftw: It should include `<iframe` part.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }]
+
+    def _real_extract(self, url):
+        display_id = self._match_id(url)
+        webpage = self._download_webpage(url, display_id)
+
+        # Extract the required info of the media files gathered in a dictionary
+        media_files_info = self.__extract_from_webpage(display_id, webpage)
+        # Some web pages embed videos from other platforms like dailymotion, therefore we pass on these URLs
```
dstftw: Add a test for such webpage.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,122 @@
...
+        # Check whether the matches failed, which might be when dealing with other players (e.g., dailymotion stream)
+        if (clip_id is None) or (player_id is None) or (skin_name is None):
+            return None
+
+        return self.__extract_from_player(display_id, clip_id, player_id, skin_name)
+
+    def __extract_embed_url(self, webpage):
+        return self._search_regex(
+            r'<iframe[^>]+src=(["\'])(?P<url>.+?)\1',
+            webpage, 'embed url', group='url')
```
dstftw: This will fail in case of no iframe. Generic embeds may be embedded differently so flow will never reach generic extraction delegation for such embeds.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def __extract_from_player(self, display_id, clip_id, player_id, skin_name):
+        """Extracts the JSON object containing the required media info from the embedded arkena player"""
+        arkena_url = 'http://play.arkena.com/config/avp/v1/player/media/{0}/{1}/{2}/?callbackMethod=?'.format(clip_id,
```
dstftw: Looks like arkena is a video hosting platform. It probably make sense to generalize it in separate extractor similarly to kaltura and others.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,122 @@
...
+            self._sort_formats(formats)
+
+        return formats
+
+    def __get_mp4_video_formats(self, media_files_json):
+        formats = []
+        mp4_files_json = media_files_json.get('Mp4')
+        if mp4_files_json is not None:
+            for video_info in mp4_files_json:
+                bitrate = int_or_none(video_info.get('Bitrate'), scale=0.001) # Scale bitrate to KBit/s
```
dstftw: It should be 1000 not 0.001.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,122 @@
...
+        info_json = self._parse_json(self._search_regex(arkena_info_regex, arkena_info, 'json', group='json'),
+                                     display_id)
+
+        # All videos are part of a playlist, a single video is also put in a playlist
+        media_files_info = info_json.get('Playlist')
+        if media_files_info is not None:
+            media_files_info = media_files_info[0]
+        return media_files_info
+
+    def __get_thumbnails(self, media_files_info):
```
dstftw: Extracting code in separate methods is questionable in this case since all of them are used only once and just make it harder to read the code due to jumping back and forth.

youtube_dl/extractor/lcp.py
rvanbekkum: Yes, noted that as well. Do you suggest I should make a separate extractor for it, and if so, can we include that in this PR?

youtube_dl/extractor/lcp.py
rvanbekkum: Good catch, should have checked that better and tested it (must have been because of the implementation including a `*`). I will change it.

youtube_dl/extractor/lcp.py
dstftw: As you wish. Note that looks like it would be impossible to craft a proper `_VALID_URL` URI template since embed URI is not domain specific (https://play.arkena.com/embed/avp/v2/player/media/b41dda37-d8e7-4d3f-b1b5-9a9db578bdfe/1/129411, http://play.lcp.fr/embed/327336/131064/darkmatter/0), so turns out it may only be available via shortcut like `arkena:clip_id:skin:player_id` or via base extractor implementation.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,122 @@
...
+        return formats
+
+    def __get_mp4_video_formats(self, media_files_json):
+        formats = []
+        mp4_files_json = media_files_json.get('Mp4')
+        if mp4_files_json is not None:
+            for video_info in mp4_files_json:
+                bitrate = int_or_none(video_info.get('Bitrate'), scale=0.001) # Scale bitrate to KBit/s
+                formats.append({
+                    'url': video_info.get('Url'),
```
dstftw: If `Url` is not in `video_info` a format with `None` `url` will be added. Since `url` is critical field no such formats should be added. Same for thumbnails.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def __get_video_formats(self, media_files_info):
+        formats = []
+        media_files = media_files_info.get('MediaFiles')
+
+        if media_files is not None:
+            formats.extend(self.__get_mp4_video_formats(media_files))
+            self._sort_formats(formats)
+
+        return formats
```
dstftw: There are also m3u8 and f4m formats available that should be extracted as well.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,122 @@
...
+            return self.url_result(url, 'Generic')
+
+        video_formats = self.__get_video_formats(media_files_info)
+        video_thumbnails = self.__get_thumbnails(media_files_info)
+
+        return {
+            'id': media_files_info['EntryName'],
+            'title': self._og_search_title(webpage),
+            'formats': video_formats,
+            'thumbnails': video_thumbnails
```
dstftw: `PublishDate` may be extracted as `timestamp` with `utils.parse_iso8601`. Description may be extracted from the `webpage`.

youtube_dl/extractor/lcp.py
rvanbekkum: Interesting. I looked at the example pages you posted and compared them and saw that they both contain a call to `ArkenaWidgetApi.Initialize` and both hold a URL from which you can obtain information about the media files in different format (e.g., https://play.arkena.com/config/avp/v2/player/media/b41dda37-d8e7-4d3f-b1b5-9a9db578bdfe/1/129411/?callbackMethod=?).
Maybe that would be a good starting point for setting up a separate extractor for this platform?

youtube_dl/extractor/lcp.py
rvanbekkum: Updated all occurences. (Commit I pushed now does not include all suggested changes *yet*)

youtube_dl/extractor/lcp.py
rvanbekkum: Added such a test. It delegates to the dailymotion extractor.

youtube_dl/extractor/lcp.py
dstftw: Sounds good.

youtube_dl/extractor/lcp.py
remitamine: it also have DASH formats but it won't work with the current implementaion(i didn't add support for `$Time$` in template because i thought it only used for live streams but it looks like it can be used in static streams).

youtube_dl/extractor/lcp.py
rvanbekkum: For some pages the `description` is missing, in that case it now extracts the title of the page. The `timestamp` is also added (tested for proper output as well).

youtube_dl/extractor/lcp.py
rvanbekkum: This is now taken care of. The formats/thumbnails without urls are skipped.

youtube_dl/extractor/lcp.py
dstftw: Missing in some cases description does not prevent it's extraction when it's present.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,149 @@
...
+        if not playlist_files_info:
+            return self.url_result(url, 'Generic')
+
+        media_files_info = playlist_files_info[0]
+        video_formats = self.__get_video_formats(media_files_info)
+        video_thumbnails = self.__get_thumbnails(media_files_info)
+        video_timestamp = parse_iso8601(media_files_info.get('MediaInfo', {}).get('PublishDate'))
+
+        title = self._og_search_title(webpage)
+        description = self._html_search_regex(self._meta_regex('description'), webpage, 'description', group='content', default=title)
```
dstftw: Use `_html_search_meta` instead. `default` should be `None` there is no sense duplicating metafields' values.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,149 @@
...
+        embed_url = self.__extract_embed_url(webpage)
+        embed_regex = r'(?:[a-zA-Z0-9]+\.)?lcp\.fr/embed/(?P<clip_id>[A-za-z0-9]+)/(?P<player_id>[A-za-z0-9]+)/(?P<skin_name>[^\/]+)'
+
+        clip_id = self._search_regex(embed_regex, embed_url, 'clip id', group='clip_id', default=None)
+        player_id = self._search_regex(embed_regex, embed_url, 'player id', group='player_id', default=None)
+        skin_name = self._search_regex(embed_regex, embed_url, 'skin name', group='skin_name', default=None)
+
+        # Check whether the matches failed, which might be when dealing with other players (e.g., dailymotion stream)
+        if not clip_id or not player_id or not skin_name:
+            return None
```
dstftw: Why not just replace (80-97) with single regex match?

youtube_dl/extractor/lcp.py
rvanbekkum: When it is present it is extracted from the webpage. When it is not present on the webpage, the title is used. Do you think this is solution is fine, or would you suggest something else to tackle this?

An example where the description is missing is the following page:
http://www.lcp.fr/emissions/politique-matin/271085-politique-matin

youtube_dl/extractor/lcp.py
rvanbekkum: Great! Processing this and the remaining comments might take a longer time. Will be working on it in the coming days. I will leave a comment as soon as it is ready so you can take a look at it.
Thank you for your comments and suggestions on it.

youtube_dl/extractor/lcp.py
```diff
@@ -0,0 +1,149 @@
...
+            'id': '327336',
+            'url': 're:http://httpod.scdn.arkena.com/11970/327336_[0-9]+.mp4',
+            'ext': 'mp4',
+            'title': 'Politique Matin - Politique matin',
+            'upload_date': '20160225',
+            'description': 'Politique Matin - Politique matin',
+            'timestamp': 1456391602
+        }
+    }, {
+        'url': 'http://www.lcp.fr/le-direct',
```
dstftw: Is there any other URL with generic embeds? It's probably easier to make `_VALID_URL` only match `http://www.lcp.fr/.../...` but not `http://www.lcp.fr/...`. This would allow drop all explicit redirections to generic extractor and make extractor more staightforward.

youtube_dl/extractor/lcp.py
dstftw: Thanks for your efforts.

youtube_dl/extractor/lcp.py
rvanbekkum: Good point. I could only find `http://www.lcp.fr/le-direct` and `http://www.lcp.fr/le-direct/24-24` with dailymotion embeds. For that people might just want to visit the [dailymotion page of LCP](http://www.dailymotion.com/LCP) though. The straightforward approach with matching `http://www.lcp.fr/.../...` might therefore not work, but I will think about an alternative.
If you already have suggestions, let me know.

youtube_dl/extractor/lcp.py
rvanbekkum: The default for `default` when calling `_html_search_meta` seems to be `NO_DEFAULT` though instead of `None`. I encountered this when testing for pages where the description is missing, so for this page the test also fails when using `_html_search_meta`, that is why I used the `_html_search_regex` here.
I agree with you that it would be better to use `_html_search_meta` here, but it causes problems when the description is missing.

Do you know any workarounds for cases in which the `description` is missing?

dstftw: Don't see any problem here.
1. Extract description with `_html_search_meta` passing `default=None`.
2. Remove description from second test.

rvanbekkum: Ah, I get what you mean, I was a little bit confused at first when looking at the signature of `_html_search_meta` (no `default` parameter in there). I see what I was missing now. Thank you.

dstftw: It's just passed through via keyworded arguments.

rvanbekkum:
Processed your feedback and therefore implemented an *Arkena Play* extractor that can be used either through a shortcut (similar to the kaltura extractor) or through the url of the embedded player.
The shortcut is based on the description that is provided [here](https://support.arkena.com/display/PLAY/Ways+to+embed+your+video) on how to embed a video using the PLAY player and looks like this:
```
arkena:<version>:<mediatype>:<mediaId>:<widgetsettingId>:<accountId>
```
(Here `<version>` is the version in the url (e.g., `/avp/v2/...`) so we can support multiple versions. I also added an example of the usage under the tests for the extractor.)

Support for multiple formats is now added (mp4, flv, f4m, m3u8, dash, webm).
Further, the code has been refactored such that it is not split up too much into separate functions anymore, but in such way that functions encapsulate separate functionality.

Hopefully this is what you have in mind, having the Arkena extractor that can be reused for other websites and/or extractors. If there is anything further I may do for this PR, let me know. I am glad to help.

youtube_dl/extractor/lcp.py
rvanbekkum: As noted in my comment below, I now have added support for multiple formats (including the ones you mentioned) in the *Arkena Play* extractor :smile:.

youtube_dl/extractor/lcp.py
rvanbekkum: Took this into account now. It should be not cluttered anymore in the current implementation.

yan12125:

rvanbekkum:
The imports for the extractors are now moved accordingly. Hopefully everything is to your wish now, just let me know :wink:

dstftw:
Cherry-picked in 4671dd41b27e39eb4682189fca44d0f4272a4751, thanks. Sorry for long wait.

rvanbekkum:
No problem, thanks for adding the changes.

PR #8705, opened 2016-03-01
Fix iqiyi extract function

hlx98007:
Iqiyi changed their flash player address. With this fix, youtube-dl is able to download from iqiyi again.


yan12125:
Thanks, but actually it's a dead code. See 1b541d8d6ee5e3543253501d725af7ce4a55cad8.

hlx98007:
I am still able to download from iqiyi with this fix though.

PR #8715, opened 2016-03-01
[downloader/fragment] Set 'total_bytes' in progress hook to None

### LINKED ISSUE #8711: progresshook on fragment.py: Inconsistent behavior on total_bytes
In [the documentation of progress_hooks](https://github.com/rg3/youtube-dl/blob/9fb556eef0c6e83594ca4a7fe4d924a6a2a43a60/youtube_dl/YoutubeDL.py#L223), it's said that that if total_bytes is unknown, the value would be None:
```
    progress_hooks:    A list of functions that get called on download
                       progress, with a dictionary with the entries
                       * status: One of "downloading", "error", or "finished".
                                 Check this first and ignore unknown values.
                       If status is one of "downloading", or "finished", the
                       following properties may also be present:
...
                       * total_bytes: Size of the whole file, None if unknown
```
However this is not the case for fragment.py, where the [key simply doesn't exist](https://github.com/rg3/youtube-dl/blob/9fb556eef0c6e83594ca4a7fe4d924a6a2a43a60/youtube_dl/downloader/fragment.py#L58):
```
        # This dict stores the download progress, it's updated by the progress
        # hook
        state = {
            'status': 'downloading',
            'downloaded_bytes': 0,
            'frag_index': 0,
            'frag_count': total_frags,
            'filename': ctx['filename'],
            'tmpfilename': ctx['tmpfilename'],
        }
```

This has caused downstream issues such as [KeyError: 'total_bytes'](https://github.com/Toollabs/video2commons/issues/8)

zhuyifei1999:
To be consistent with progress hook documentation.

Fixes #8711

### DIFF
--- a/youtube_dl/downloader/fragment.py
+++ b/youtube_dl/downloader/fragment.py
@@ -62,6 +62,7 @@ class FragmentFD(FileDownloader):
             'frag_count': total_frags,
             'filename': ctx['filename'],
             'tmpfilename': ctx['tmpfilename'],
+            'total_bytes': None,
         }
 
         start = time.time()


dstftw:
Current documentation is consistent. It clearly states these fields may not be present.

Issue #8622, opened 2016-02-20
Why prefer libav(avconv) over ffmpeg?


mikhaelkh:
ffmpeg is better in practically every way - explained [here](https://wiki.debian.org/Debate/libav-provider/ffmpeg)

jaimeMF:
Personally I would prefer ffmpeg.
But it seems that there could be [some problems in Ubuntu](https://github.com/rg3/youtube-dl/pull/353#issuecomment-6081888), I don't know if the last versions have it though.

remitamine:
> I don't know if the last versions have it though.

i should be available at least in the last two version:
http://packages.ubuntu.com/search?keywords=ffmpeg

i also prefer ffmpeg.
avconv is not available in the official repositoy of archlinux.
mpv prefer ffmpeg: https://github.com/mpv-player/mpv/wiki/FFmpeg-versus-Libav
the same for mplayer: http://www.mplayerhq.hu/design7/news.html
and i think vlc also prefer ffmpeg(they host a mirror for ffmpeg): http://git.videolan.org/?p=ffmpeg.git;a=summary

yan12125:
It's possible to detect whether ffmpeg is really ffmpeg or symlinked to avconv - either find "avconv" in ```ffmpeg -version``` or checking symlinks should work.

I prefer ffmpeg, too. Here are some issues relating ffmpeg/avconv:
Only avconv fails:
* #7238 file: protocol is problematic
* #7208 data: protocol not implemented
* #6737 MP3 Conversion error
* #6046 duplicated embedded image (need more checking)
* #2437 av_interleaved_write_frame(): Invalid argument
* #7374 Non-monotonous mpegts fragments not playable. Can't reproduce with avconv v12_dev0-2381-g8caadfc
* #8135 avconv issue (?)
* #8170 avconv not supporting AES-128
* avconv fails for test_NBC_2 and ffmpeg succeed

Only ffmpeg fails:
* #8350 avconv can handle streams without aac_adtstoasc
* https://github.com/rg3/youtube-dl/pull/8408#issuecomment-179163549 ffmpeg can't handle files from kaltura.com properly

> avconv is not available in the official repositoy of archlinux.

@remitamine I use libav-git-no-libs from AUR. It coexists fine with official ffmpeg.

Reino17:
[...]

PR #8720, opened 2016-03-01
Stop libav compatibility links from being considered as genuine.

e7appew:
Debian has abandoned libav for ffmpeg, but provides compatibility links to ffmpeg equivalents for programs that still require libav tools. youtube-dl doesn't differentiate between the two, so incorrectly complains about avconv being outdated and refuses to do any muxing. See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=798936.

This fix prevents registration of version numbers for compatibility links, so that youtube-dl is forced to deal with only tools that are actually installed.

### DIFF
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -85,11 +85,16 @@ class FFmpegPostProcessor(PostProcessor):
                 self._paths = dict(
                     (p, os.path.join(location, p)) for p in programs)
                 self._versions = dict(
-                    (p, get_exe_version(self._paths[p], args=['-version']))
+                    (p, get_exe_version(self._paths[p], args=['-version'],
+                                        version_re=r'%s\s+version\s+([-0-9._a-zA-Z]+)' % p,
+                                        unrecognized=False))
                     for p in programs)
         if self._versions is None:
             self._versions = dict(
-                (p, get_exe_version(p, args=['-version'])) for p in programs)
+                (p, get_exe_version(p, args=['-version'],
+                                    version_re=r'%s\s+version\s+([-0-9._a-zA-Z]+)' % p,
+                                    unrecognized=False))
+                for p in programs)
             self._paths = dict((p, p) for p in programs)
 
         if prefer_ffmpeg:


jaimeMF:
Probably it is not too common, but this would break if you symlink `avconv` to something like `avconv-11.4`. Why don't you just use `--prefer-ffmpeg`? Also, there is some discussion in #8622 for prefering ffmpeg over avconv.

dstftw:
Why not just parse the output of `-version` and recognize the actual app by `{avconv,ffmpeg,avprobe,ffprobe} version ...` pattern?

e7appew:
@jaimeMF I have updated the PR to address that type of symlink. I'm aware of the --prefer-ffmpeg option, but it's not unreasonable to expect software to behave correctly in spite of what the default preference is.

@dstftw Whatever is easier, as long as the problem is fixed. I just wasn't prepared to break the expected behaviour of get_exe_version().

dstftw:
It can be easily simlinked to ffmpeg under `arbitrary-name` and will fail to detect it at all. Thus this solution won't work in general.

jaimeMF:
@dstftw's proposal to look into the output of `-version` seems more reasonable.

e7appew:
I've updated the fix again. Hopefully the new regex satisfies all requirements and doesn't break anything.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -85,11 +85,14 @@ def _determine_executables(self):
                 self._paths = dict(
                     (p, os.path.join(location, p)) for p in programs)
                 self._versions = dict(
-                    (p, get_exe_version(self._paths[p], args=['-version']))
+                    (p, get_exe_version(self._paths[p], args=['-version'],
+                                        version_re=r'{}\s+version\s+([-0-9._a-zA-Z]+)'.format(p)))
```
dstftw: Unnamed placeholders are not supported in python 2.6.

dstftw:
No, it does not work at all. avconv simlinked to custom ffmpeg results in:
```
[debug] exe versions: avconv present, ffprobe 2.8.5, rtmpdump 2.4
```
It should detect it as ffmpeg 2.8.5.

e7appew:
@dstftw Would you be kind enough to show me the -version output of your custom ffmpeg? Thanks.

dstftw:
By custom I mean installed under custom name. The `-version` output is intact.

e7appew:
I can't reproduce what you have. If you have avconv symlinked to ffmpeg with an arbitrary name, wouldn't you also need an ffmpeg symlink to the arbitrary name for it to detect the presence of ffmpeg?

dstftw:
No, why? I've a valid name found via `$PATH` (`avconv`) that is simlinked to some arbitrary path where ffmpeg is located, it just should read the output of `avconv -version` and figure out it's actually ffmpeg.

e7appew:
That behaviour is as a result of existing code, not as a result of the changes I'm trying to introduce. Without complicating things, all the fix is trying to do is to prevent ffmpeg from being passed off as the real avconv, or vice versa. I'd be happy to close this ticket, if you would open another that fixes my original problem and the extra feature you want.

mintar:
This bug is still present in version 2017.08.18 (on Ubuntu 16.04, where avconv is symlinked to ffmpeg). A workaround is to specify `--prefer-ffmpeg` either on the command line or in `~/.config/youtube-dl/config`.

PR #8736, opened 2016-03-02
allows more possibilities for the match filter options

gillouche:
Only letters and numbers were accepted by the regular expression for the match filter. 
Youtube playlists' can also contain other characters like space, dash, underscore.

This change allows more possibilities for the matcher filter.


dstftw:
It does not properly parse strings with whitespace (trailing whitespace is erroneously captured). Following test fails:
```
        self.assertFalse(match_str(
            'playlist_title != some playlist with space & playlist_title != foobar',
            {'playlist_title': 'some playlist with space'}))
```

gillouche:
Indeed, thanks. The whitespace before the '&' is taken into account. I'll work on a fix.

gillouche:
I added more valid characters to the check and trim both part of the comparison if we deal with strings instead of just the match part.

Travis-ci seems to be dead though, it is still running for the last commit since yesterday.

gillouche:
Hi,

I've been using my version for a while now and have not encountered any problems.

Cheers.

dstftw:
Now you can match quotes string literals that is more generic solution, e.g. `--match uploader = '...'`.

PR #8739, opened 2016-03-03
[extractor/common] add update_url_params helper method to add or update query string params

remitamine:
it a common task to create or update query string params, it will be simpler to use this helper method.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -517,6 +517,13 @@ class InfoExtractor(object):
             else:
                 self.report_warning(errmsg + str(ve))
 
+    def update_url_params(self, url, params):
+        parsed_url = compat_urlparse.urlparse(url)
+        qs = compat_urlparse.parse_qs(parsed_url.query)
+        qs.update(params)
+        return compat_urlparse.urlunparse(
+            parsed_url._replace(query=compat_urllib_parse.urlencode(qs, True)))
+
     def report_warning(self, msg, video_id=None):
         idstr = '' if video_id is None else '%s: ' % video_id
         self._downloader.report_warning(
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -20,6 +20,7 @@ from ..compat import (
     compat_urllib_error,
     compat_urllib_parse,
     compat_urlparse,
+    compat_parse_qs,
     compat_str,
     compat_etree_fromstring,
 )
@@ -519,7 +520,7 @@ class InfoExtractor(object):
 
     def update_url_params(self, url, params):
         parsed_url = compat_urlparse.urlparse(url)
-        qs = compat_urlparse.parse_qs(parsed_url.query)
+        qs = compat_parse_qs(parsed_url.query)
         qs.update(params)
         return compat_urlparse.urlunparse(
             parsed_url._replace(query=compat_urllib_parse.urlencode(qs, True)))
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -20,7 +20,6 @@ from ..compat import (
     compat_urllib_error,
     compat_urllib_parse,
     compat_urlparse,
-    compat_parse_qs,
     compat_str,
     compat_etree_fromstring,
 )
@@ -517,13 +516,6 @@ class InfoExtractor(object):
                 raise ExtractorError(errmsg, cause=ve)
             else:
                 self.report_warning(errmsg + str(ve))
-
-    def update_url_params(self, url, params):
-        parsed_url = compat_urlparse.urlparse(url)
-        qs = compat_parse_qs(parsed_url.query)
-        qs.update(params)
-        return compat_urlparse.urlunparse(
-            parsed_url._replace(query=compat_urllib_parse.urlencode(qs, True)))
 
     def report_warning(self, msg, video_id=None):
         idstr = '' if video_id is None else '%s: ' % video_id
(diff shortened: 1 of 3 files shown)


youtube_dl/extractor/common.py
```diff
@@ -517,6 +517,13 @@ def _parse_json(self, json_string, video_id, transform_source=None, fatal=True):
             else:
                 self.report_warning(errmsg + str(ve))
 
+    def update_url_params(self, url, params):
+        parsed_url = compat_urlparse.urlparse(url)
+        qs = compat_urlparse.parse_qs(parsed_url.query)
```
yan12125: Use [```compat_parse_qs```](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/compat.py#L254-L287)

remitamine: the code was modified from `CrunchyrollBaseIE._add_skip_wall` which use `compat_urlparse.parse_qs`.

dstftw:
1. This should better go to `utils`.
2. Should be renamed to `update_url_query` or so since `params` in terms of `ParseResult` is different part of URI (`<scheme>://<netloc>/<path>;<params>?<query>#<fragment>`).
3. Test should be added.

remitamine:
the tests didn't work the order of the query params is not know in advance.

dstftw:
I did not mention the order of query params.

remitamine:
the tests fail because the order change:
```
self.assertEqual(update_url_query('http://example.com/path', {'quality': ['HD'], 'format': ['mp4']}), 'http://example.com/path?quality=HD&format=mp4')
```
i run the test twise it works the first time the second fail.
```
[amine@amine test]$ python test_utils.py TestUtil.test_update_url_query
.
Ran 1 test in 0.008s

OK
[amine@amine test]$ python test_utils.py TestUtil.test_update_url_query
F
FAIL: test_update_url_query (__main__.TestUtil)
  File "test_utils.py", line 462, in test_update_url_query
    self.assertEqual(update_url_query('http://example.com/path', {'quality': ['HD'], 'format': ['mp4']}), 'http://example.com/path?quality=HD&format=mp4')
AssertionError: 'http://example.com/path?format=mp4&quality=HD' != 'http://example.com/path?quality=HD&format=mp4'
- http://example.com/path?format=mp4&quality=HD
?                                   -----------
+ http://example.com/path?quality=HD&format=mp4
?                         +++++++++++


Ran 1 test in 0.012s

FAILED (failures=1)
```

remitamine:
the tests fail because the order change:
```
self.assertEqual(update_url_query('http://example.com/path', {'quality': ['HD'], 'format': ['mp4']}), 'http://example.com/path?quality=HD&format=mp4')
```
i run the test twise it works the first time the second fail.
```
[amine@amine test]$ python test_utils.py TestUtil.test_update_url_query
.
Ran 1 test in 0.008s

OK
[amine@amine test]$ python test_utils.py TestUtil.test_update_url_query
F
FAIL: test_update_url_query (__main__.TestUtil)
  File "test_utils.py", line 462, in test_update_url_query
    self.assertEqual(update_url_query('http://example.com/path', {'quality': ['HD'], 'format': ['mp4']}), 'http://example.com/path?quality=HD&format=mp4')
AssertionError: 'http://example.com/path?format=mp4&quality=HD' != 'http://example.com/path?quality=HD&format=mp4'
- http://example.com/path?format=mp4&quality=HD
?                                   -----------
+ http://example.com/path?quality=HD&format=mp4
?                         +++++++++++


Ran 1 test in 0.012s

FAILED (failures=1)
```

dstftw:
They **should not** be tested this way.

remitamine:
it's the first time i work on unit tests.
so i don't know how to fix this.

dstftw:
Compare parsed URLs or query dicts.

remitamine:
Comparing parsed URLs didn't work but it works with query dicts.
i added two tests one for updating and one for creating query, is this enough.
```python
    def test_update_url_query(self):
        def query_dict(url):
            return compat_parse_qs(compat_urlparse.urlparse(url).query)
        self.assertEqual(query_dict(update_url_query('http://example.com/path?id=ref%3ABC2996102916001&refurl=http%3A%2F%2Fwww.example.com%2Fen%2Fvideos%2F1331655630249&key=AQ%7E%7E%2CAAAApYJ7UqE%7E%2Cxqr_zXk0I-zzNndy8NlHogrCb5QdyZRf&id=1398061561001#__youtubedl_smuggle=%7B%22Referer%22%3A+%22http%3A%2F%2Fwww.example.com%2Fen%2Fvideos%2F1331655630249%22%7D', {'id': ['6102916001'], 'key': []})), query_dict('http://example.com/path?id=6102916001&refurl=http%3A%2F%2Fwww.example.com%2Fen%2Fvideos%2F1331655630249#__youtubedl_smuggle=%7B%22Referer%22%3A+%22http%3A%2F%2Fwww.example.com%2Fen%2Fvideos%2F1331655630249%22%7D'))
        self.assertEqual(query_dict(update_url_query('http://example.com/path', {'quality': ['HD'], 'format': ['mp4']})), query_dict('http://example.com/path?quality=HD&format=mp4'))
```

dstftw:
At least following should be tested: type other than string, several values in list under the key, non list under the key.

remitamine:
> 

    This should better go to utils.
    Should be renamed to update_url_query or so since params in terms of ParseResult is different part of URI (<scheme>://<netloc>/<path>;<params>?<query>#<fragment>).
    Test should be added.

> At least following should be tested: type other than string, several values in list under the key, non list under the key.

done.

dstftw:
LGTM.
Some non-ASCII and integers, floats would also be useful.
Also options not to overwrite existing keys/merge values under the keys may be useful.

remitamine:
> Some non-ASCII and integers, floats would also be useful.
i will check them now.

> Also options not to overwrite existing keys/merge values under the keys may be useful as further improvement.

i don't know if this can be useful, it's better to add if it's needed.

remitamine:
i added tests for int, float, and non ASCII strings.

PR #8774, opened 2016-03-05
[jeuvideo.com] Modify the attribut data-src in data-srcset-video

kadogo:
Hello,

The attribut data-src have change in data-srcset-video in the last version of jeuvideo.com

Cheers.


youtube_dl/extractor/jeuxvideo.py
```diff
@@ -30,7 +30,7 @@ def _real_extract(self, url):
         webpage = self._download_webpage(url, title)
         title = self._html_search_meta('name', webpage) or self._og_search_title(webpage)
         config_url = self._html_search_regex(
-            r'data-src="(/contenu/medias/video.php.*?)"',
+            r'data-srcset-video="(/contenu/medias/video.php.*?)"',
```
dstftw: Keep both variants.

kadogo: Ah, because without that modification I cannot download the videos.

Issue #8783, opened 2016-03-06
Create an issue template for GitHub


rg3:
The repository already has a CONTRIBUTING.md file that mentions calling youtube-dl with the --verbose flag when reporting issues. This file is linked from the new issue page above the new issue form and is noticeable. I think it has helped reduce somewhat the number of bad bug reports. I don't know if you share that opinion.

Recenly, GitHub has introduced the concept of [issue templates](https://github.com/blog/2111-issue-and-pull-request-templates). Perhaps you could create an ISSUE_TEMPLATE.md file with an unmissable request to include the --verbose flag and the full output in a proper code block, and to make sure they're running the latest version of youtube-dl.

Feel free to close if you don't agree.

rg3:
Solved. Thanks!

Issue #7849, opened 2015-12-12
Display current latest version on GitHub when using the `--update` option if other software management solution is used on system


Vdragon:
Please support fetching & showing the latest release when using the `--update` option on systems using alternative software management solution, it's useful to check whether your youtube-dl distribution is the latest.

dstftw:
We have no control over alternative software management solution therefore they decide themselves how to update their packages.

Vdragon:
@dstftw 
Hi, I mean showing/printing only the latest upstream "version number" instead of updating the software, so it's not related to upgrading using alternative software management solution.

You can check the latest version by launching a web browser and browse the [Releases page](https://github.com/rg3/youtube-dl/releases) but it's better if youtube-dl could show it itself.

dstftw:
I don't think mixing official version and package version is a good idea since it can be misleading. Moreover most likely this will be patched out by package maintainers.

yan12125:
@Vdragon FYI: The latest version number can be fetched from https://rg3.github.io/youtube-dl/update/LATEST_VERSION.

Vdragon:
@dstftw 
IMO as long as the message isn't misleading its a good idea.

@yan12125 
Thanks, highly appreciated!

PR #8785, opened 2016-03-06
Add an ISSUE_TEMPLATE file

sandervdo:
I agree with @rg3 that an issue template could be useful (see issue #8783). I noticed that a lot of issues are being answered with requests to provide more information or clarifications. 

I've had a look at the existing [CONTRIBUTIONS.md](https://github.com/rg3/youtube-dl/blob/master/CONTRIBUTING.md) and several issues. I have tried to create an issue template that is as general as possible. Since there are issues containing bug reports, site support requests and feature requests I wanted to make them all fit in the template.

I think this template can be of use for the maintainers of the project/issues. If you have any remarks/feedback please let me know, I'd be happy to improve the template.

Regards,
Sander


yan12125:
For users install youtube-dl via package managers (apt-get, pacman, etc.), ```youtube-dl -U``` ask users to upgrade via package managers, while package managers tell users youtube-dl is up-to-date (it's usually not in fact). This can be misleading. Checking http://yt-dl.org/ is a possible solution, but I guess there's a better way than opening a webpage.

sandervdo:
Fair point. I guess we can rephrase this. Maybe we can point the users to the contributions.md or leave out the command for updating youtube-dl? Another suggestion would be to include this remark about the package managers. I think the first one would be best as it already contains a detailed explanation on the updating process. Thoughts?

yan12125:
There's a similar idea at #7849 but rejected due to misleading wording. I guess your proposal is much better. How do you think @dstftw?

dstftw:
We already have newer version availability check with `--call-home` the problem is that it can be patched out by package maintainers.
As for me the very very first emphasized wording should be a notice suggesting checking a version with `--version` and matching it with the latest released youtube-dl version according to us that should be printed right in `ISSUE_TEMPLATE.md`. Thus it should be autogenerated on each release similarly to `CONTRIBUTING.md`.

sandervdo:
Thanks for the feedback. Based on the feedback I've altered the template. Right now it's being generated through the Makefile (task `issue_template`). It fetches the current version number from `youtube_dl/version.py` and replaces the version number that was already in the template.

Also I've changed the version verification more towards the top of the form and I've made it a simple checkbox. 

Is this what you had in mind?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,25 @@
+**Kind of issue**
+- [ ] Site request (add support for a new website)
+- [ ] Feature request (request new functionality)
+- [ ] Bug report (encountered problems with youtube-dl)
+- [ ] Other, namely ...
+
+**Run `youtube-dl --version` to check whether you have the latest version**
+- [ ] I've verified that I'm running the latest version available (2016.03.06)
```
dstftw: This should be on the very top. I'd word it differently:

Make sure you are using the **latest** version of youtube-dl. Latest youtube-dl vesion at the moment is **2016.03.06**. Read [our FAQ](https://yt-dl.org/update) if you have troubles updating.
[ ] I've verified that I'm running youtube-dl version 2016.03.06

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,25 @@
...
+*In case you have suggestions for a solution or any other relevant information you can write it here*
+
+**Example URL**
+
+*Provide the url of the file that you were trying to download.*
+
+**Output of running youtube-dl with `--verbose` or `-v`**
+```
+Replace the contents between the backticks (`) with the output of youtube-dl when running with the --verbose or -v flag.
+```
```
dstftw: This should go right after version check. There is no need to mention both `--verbose` and `-v` single `-v` is enough. It should be worded similarly to [this](https://github.com/rg3/youtube-dl#bugs) with example log between triple `'s. It should also mention early that applier is not required to fill this field it the issue's purpose is just a question or something else that does not imply providing a log.

sandervdo:
The sections that you mentioned are moved to the top of the file. I've added a line about when the log is not required. Also adopted the make file so that it replaces both occurrences of the date. 

For the dates I know match on `**[0-9\.]**`, I'm not sure if this is a safe approach, if the template uses a similar notion somewhere else it might be replaced as well. I can also change the regex to match on a specific amount of digits (say `**[xxxx.yy.zz]**`, 4 digits, period, 2 digits, etc.). Would that be a better approach? I, personally, like the current version as it looks cleaner to me.

If you have any other feedback please let me know.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+[debug] Encodings: locale cp1251, fs mbcs, out cp866, pref cp1251
+[debug] youtube-dl version 2015.12.06
+[debug] Git HEAD: 135392e
+[debug] Python version 2.6.6 - Windows-2003Server-5.2.3790-SP2
+[debug] exe versions: ffmpeg N-75573-g1d0487f, ffprobe N-75573-g1d0487f, rtmpdump 2.4
+[debug] Proxy map: {}
+...
+```
+
+**Kind of issue**
```
dstftw: I guess this should better be rephrased to something like: What is your request? (put an `x` into relevant box).

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
+Make sure you are using the **latest** version of youtube-dl. Latest youtube-dl version at this moment is **2016.03.06**. Read [our FAQ](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) if you have troubles updating.
+- [ ] I've verified that I'm running youtube-dl version **2016.03.06**
+
+*The log (next question) is not required when asking a question or when the purpose of this issue does not imply the need of a log*
```
dstftw: I was wrong. It's really more logical when it goes after *kind-of-issue* section. In this case we can spell it like: If your request is a site request or a bug report or you are unsure please include the log as described below.

sandervdo: *If the purpose of this issue is a site request, bug report or you are unsure about its relevance please include a log as described below.*

Like this? We are speaking of the purpose of an issue before, that's why I phrased it like this (to keep it consistent).

ISSUE_TEMPLATE.md
sandervdo: Agree, changed it. Will wait a bit if you have any further comments, then I'll push the changes. :+1:

ISSUE_TEMPLATE.md
dstftw: Yes. The main point is emphasize each case when it's required since average user usually has no idea whether he should or not.

sandervdo: I agree. I tried to keep everything as 'to-the-point' as possible as, in my experience, users won't read long texts/explanations. You can find the changes in the latest commit.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
+Make sure you are using the **latest** version of youtube-dl. Latest youtube-dl version at this moment is **2016.03.06**. Read [our FAQ](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) if you have troubles updating.
+- [ ] I've verified that I'm running youtube-dl version **2016.03.06**
+
+**What is the purpose of this issue? Put an `x` to the relevant box**
+- [ ] Site request (add support for a new website)
```
dstftw: We usually call it *site support request*.

sandervdo: Fixed. I was also thinking it might be better to have the example URL directly after the 'kind-of-issue' section. Now it's the latest section of the template, people might forget/miss it after all. Thoughts?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+
+**Brief description of the problem/request**
+
+*I am having a problem with ... I have tried to do ... and ... I expected that ... would happen, but instead ... happened. Example: I tried to download a file but the site was not supported. Please add support for site xyz. Another example: I encountered a bug when downloading a video from xyz. I have tried to do a and b.*
+
+**Suggested solution or other information**
+
+*In case you have suggestions for a solution or any other relevant information you can write it here*
+
+**Example URL**
```
dstftw: This can also be worded like: If the purpose of this issues is a site support request please provide at least one example URL of a video or a playlist you are trying to download.

ISSUE_TEMPLATE.md
dstftw: Sure it should be above. Let's make it this way. Put *example-url* section after *kind-of-issue* section and remove mentioning site support request from *verbose-log* section.

jaimeMF:
Note that this can be placed inside a [`.github` directory](https://help.github.com/articles/creating-an-issue-template-for-your-repository/#adding-an-issue-template) to not add noise to the root directory. I prefer using the special directory, but others may prefer to make it visible at the root.

sandervdo:
In that case it would make sense to move CONTRIBUTING.md as well (together with other supported files that are not necessary in the root of the repo).

rvanbekkum:
I think it might be convenient to keep the `CONTRIBUTING.md` file in the root so that people who want to contribute can find it quickly. (Even though the contents of that file are also in the `README.md`.

sandervdo:
Yeah, that's actually exactly why I suggested to move it to the `.gitignore` folder. The contents are taken from the `README.md`, and when you create a new issue or a new pull request it'll show a link to the file regardless of its location. So to me it makes sense to move that file as well.

jaimeMF:
I would keep `CONTRIBUTING.md` at its current location, since as @rvanbekkum people can quickly see it when they are locally browsing the source code while `ISSUE_TEMPLATE.md` is pretty much only useful for GitHub.

sandervdo:
I've moved the `ISSUE_TEMPLATE.md` into a `.github` folder. Any other feedback on the issue template as it is right now? :)

Makefile
```diff
@@ -67,6 +67,9 @@ README.md: youtube_dl/*.py youtube_dl/*/*.py
 CONTRIBUTING.md: README.md
 	$(PYTHON) devscripts/make_contributing.py README.md CONTRIBUTING.md
 
+issue_template: ISSUE_TEMPLATE.md youtube_dl/version.py
+	$(PYTHON) devscripts/make_issue_template.py ISSUE_TEMPLATE.md
```
jaimeMF: Must be updated to `.github/ISSUE_TEMPLATE.md`.

sandervdo: Whoops, so sorry, totally forgot about the makefile. Fixed now, ran `make` and it ran without errors.

sandervdo:
@jaimeMF I have changed the makefile, thanks for that. Anything else that requires change? :smile:

Issue #8731, opened 2016-03-02
[Feature Suggestion - twitch.tv] Enable Audio_Only format


simonics:
Adding the request parameter `'allow_audio_only': 'true'` alongside the two instances of `'allow_source': 'true'` in the twitch extractor module appears to enable an `Audio_Only` format, which is regularly restricted to the twitch.tv mobile app.

After implementing this change, `youtube-dl -F twitch.tv/[channel]` on a random channel returns the following additional format summary:

    Audio_Only   mp4        audio only  128k , mp4a.40.2

Would this be worth adding in? I'd imagine that the format might be useful in low-bandwidth situations.

buchno:
> Would this be worth adding in? I'd imagine that the format might be useful in low-bandwidth situations.

Would be great for the [music channels](http://www.twitch.tv/directory/game/Music).

benjamincongdon:
Addressed in #8786.

dstftw:
audio only formats will be supported in the next version.

PR #8786, opened 2016-03-07
[extractor/twitch] Added "Audio_Only" format

benjamincongdon:
Added the `allow_audio_only` tag to queries to support Audio_Only format for the Twitch extractor.

Requested in #8731.


dstftw:
Check your code with flake8.

benjamincongdon:
Thanks for the feedback. I addressed these points.

dstftw:
Squash commits into single one with proper name.

benjamincongdon:
Thanks for your patience. Followed your advise and fixed this in #8791.

PR #8791, opened 2016-03-08
[twitch] Support for "Audio_Only" format

benjamincongdon:
Added the `allow_audio_only` tag to queries to support Audio_Only format for the Twitch extractor.

Requested in #8731.


dstftw:
Thanks.

PR #8805, opened 2016-03-09
[cda] Add new extractor

### LINKED ISSUE #8760: The biggest movie site in Poland
Hello,
Could you add support for http://www.cda.pl/ site?

kasper93:
Fixes #8760 

Not much to say, simple extractor for current cda.pl page. Premium accounts are not supported, I don't have one nor I intend to buy it so someone else will need to do that part.

### DIFF
--- a/youtube_dl/extractor/cda.py
+++ b/youtube_dl/extractor/cda.py
@@ -1,0 +1,96 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from ..utils import (
+    decode_packed_codes,
+    ExtractorError,
+    parse_duration
+)
+
+
+class CDAIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:(?:www|ebd)\.)?cda\.pl/(?:video|[0-9]+x[0-9]+)/(?P<id>[0-9a-z]+)'
+    _TESTS = [
+        {
+            'url': 'http://www.cda.pl/video/5749950c',
+            'md5': '6f844bf51b15f31fae165365707ae970',
+            'info_dict': {
+                'id': '5749950c',
+                'ext': 'mp4',
+                'height': 720,
+                'title': 'Oto dlaczego przed zakrętem należy zwolnić.',
+                'duration': 39
+            }
+        },
+        {
+            'url': 'http://www.cda.pl/video/57413289',
+            'md5': 'a88828770a8310fc00be6c95faf7f4d5',
+            'info_dict': {
+                'id': '57413289',
+                'ext': 'mp4',
+                'title': 'Lądowanie na lotnisku na Maderze',
+                'duration': 137
+            }
+        }
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage('http://ebd.cda.pl/0x0/' + video_id, video_id)
+
+        if 'Ten film jest dostępny dla użytkowników premium' in webpage:
+            raise ExtractorError('This video is only available for premium users.', expected=True)
+
+        title = self._html_search_regex(r'<title>(.+?)</title>', webpage, 'title', fatal=False)
+
+        def _get_format(page, version=''):
+            unpacked = decode_packed_codes(page)
+            duration = self._search_regex(r"duration:\\'(.+?)\\'", unpacked, 'duration', fatal=False)
+            format_id = None
+            height = None
+
+            m = re.search(r'<a data-quality="(?P<format_id>[^"]+)" href="[^"]+" class="quality-btn quality-btn-active">(?P<height>[0-9]+)p<\/a>', page)
+            if m:
+                format_id = m.group('format_id')
+                height = int(m.group('height'))
+
+            url = self._search_regex(r"url:\\'(.+?)\\'", unpacked, version + ' url', fatal=False)
+            if url is None:
+                return None
+
+            return {
+                'format_id': format_id,
+                'height': height,
+                'url': url
+            }, parse_duration(duration)
+
+        formats = []
+
+        format_desc, duration = _get_format(webpage) or (None, None)
+        if format_desc is not None:
+            formats.append(format_desc)
+
+        pattern = re.compile(r'<a data-quality="[^"]+" href="([^"]+)" class="quality-btn">([0-9]+p)<\/a>')
+        for version in re.findall(pattern, webpage):
+            webpage = self._download_webpage(version[0], video_id, 'Downloading %s version information' % version[1], fatal=False)
+            if not webpage:
+                # Manually report warning because empty page is returned when invalid version is requested.
+                self.report_warning('Unable to download %s version information' % version[1])
+                continue
+
+            format_desc, duration_ = _get_format(webpage, version[1]) or (None, None)
+            duration = duration or duration_
+            if format_desc is not None:
+                formats.append(format_desc)
+
+        self._sort_formats(formats)
+
+        return {
+            'id': video_id,
+            'title': title,
+            'formats': formats,
+            'duration': duration
+        }
--- a/youtube_dl/extractor/__init__.py
+++ b/youtube_dl/extractor/__init__.py
@@ -107,6 +107,7 @@ from .cbsnews import (
 )
 from .cbssports import CBSSportsIE
 from .ccc import CCCIE
+from .cda import CDAIE
 from .ceskatelevize import CeskaTelevizeIE
 from .channel9 import Channel9IE
 from .chaturbate import ChaturbateIE


youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,97 @@
...
+                          page, re.DOTALL)
+            unpacked = unpack(m.group('payload'), m.group('symbol_table').split('|'), int(m.group('base')))
+
+            try:
+                duration = self._search_regex(r"duration:\\'(.+?)\\'", unpacked, 'duration')
+                t = time.strptime(duration, '%H:%M:%S')
+                durations.append(datetime.timedelta(hours=t.tm_hour,
+                                                    minutes=t.tm_min,
+                                                    seconds=t.tm_sec).total_seconds())
+            except:
```
dstftw: Never use bare except.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,97 @@
...
+        unpack = Unpacker()
+        durations = []
+
+        def _get_format(page):
+            m = re.search(r"}\('(?P<payload>.*)', *(?P<base>[0-9]+), *[0-9]+, *'(?P<symbol_table>.*)'\.split\('\|'\)",
+                          page, re.DOTALL)
+            unpacked = unpack(m.group('payload'), m.group('symbol_table').split('|'), int(m.group('base')))
+
+            try:
+                duration = self._search_regex(r"duration:\\'(.+?)\\'", unpacked, 'duration')
```
dstftw: Use `utils.parse_duration` instead.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,97 @@
...
+            word = match.group(0)
+            return symbol_table[self._translate(word, base)] or word
+
+        return re.sub(r'\b\w+\b', lookup, payload)
+
+    def _translate(self, word, base):
+        ret = 0
+        for index, cipher in enumerate(word[::-1]):
+            ret += (base ** index) * self.dictionary[cipher]
+        return ret
```
dstftw: This looks similar to `utils.decode_packed_codes`.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,97 @@
...
+        if self._html_search_regex(r'(<a href="https:\/\/premium\.cda\.pl\/rejestracja" target="_blank" onclick="return premiumInfo\(\)")',
+                                   webpage, 'premium_check', default=None):
+            raise ExtractorError('This video is only available for premium users.', expected=True)
+
+        title = self._html_search_regex(r'<title>(.+?)<\/title>', webpage, 'title', fatal=False)
+        unpack = Unpacker()
+        durations = []
+
+        def _get_format(page):
+            m = re.search(r"}\('(?P<payload>.*)', *(?P<base>[0-9]+), *[0-9]+, *'(?P<symbol_table>.*)'\.split\('\|'\)",
```
dstftw: This looks similar to `utils.decode_packed_codes`.

youtube_dl/extractor/cda.py
kasper93: Sure.

youtube_dl/extractor/cda.py
kasper93: Thanks for the tip, I really should have check utils before :)

kasper93:
New version uploaded.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,71 @@
...
+            'title': 'Oto dlaczego przed zakrętem należy zwolnić.',
+            'duration': 39,
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage("http://ebd.cda.pl/0x0/" + video_id, video_id)
+
+        if self._html_search_regex(r'(<a href="https:\/\/premium\.cda\.pl\/rejestracja" target="_blank" onclick="return premiumInfo\(\)")',
```
dstftw: This can be simplified with just `if 'abc' in webpage`.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,71 @@
...
+                durations.append(parse_duration(duration))
+
+            m = re.search(r'<a data-quality="(?P<format_id>[^"]+)" href="[^"]+" class="quality-btn quality-btn-active">(?P<resolution>[0-9]+p)<\/a>',
+                          page, re.DOTALL)
+
+            return {
+                'url': self._search_regex(r"url:\\'(.+?)\\'", unpacked, 'url'),
+                'format_id': m.group('format_id'),
+                'resolution': m.group('resolution'),
+                'preference': int(m.group('resolution').rstrip('p'))
```
dstftw: Instead of `resolution` and `preference` it should be extracted as `height`.

youtube_dl/extractor/cda.py
kasper93: Done.

youtube_dl/extractor/cda.py
kasper93: Done.

kasper93:
New version uploaded.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,69 @@
...
+
+        pattern = re.compile(r'<a data-quality="[^"]+" href="([^"]+)" class="quality-btn">([0-9]+p)<\/a>')
+        for version in re.findall(pattern, webpage):
+            webpage = self._download_webpage(version[0], video_id, 'Fetching %s version information' % version[1])
+            formats.append(_get_format(webpage))
+
+        self._sort_formats(formats)
+
+        if (len(durations) <= 0) or (durations.count(durations[0]) != len(durations)):
+            durations.insert(0, 0)
```
dstftw: What is the rationale behind this? Isn't it easier to just initialize duration with `None` and then update it with first single duration seen when processing formats?

kasper93: My idea was to validate if returned duration is the same for each format, and set it to zero otherwise (or when there is no duration). But it is indeed not needed. Duration is not critical. I've changed it not so that it will remember first valid duration it find or set to `None` if not found  as you suggested.

kasper93:
New version uploaded. Apart from duration extraction change I've added `height` value to the test.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,66 @@
...
+
+        title = self._html_search_regex(r'<title>(.+?)<\/title>', webpage, 'title', fatal=False)
+        duration = [None]
+
+        def _get_format(page):
+            unpacked = decode_packed_codes(page)
+            if duration[0] is None:
+                duration[0] = self._search_regex(r"duration:\\'(.+?)\\'", unpacked, 'duration', default=None)
+
+            m = re.search(r'<a data-quality="(?P<format_id>[^"]+)" href="[^"]+" class="quality-btn quality-btn-active">(?P<height>[0-9]+)p<\/a>',
```
dstftw: This will break the entire extraction if there is no match for some format.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,66 @@
...
+from .common import InfoExtractor
+from ..utils import (
+    decode_packed_codes,
+    ExtractorError,
+    parse_duration,
+)
+
+
+class CDAIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:(www|ebd)\.)?cda\.pl/(video|[0-9]+x[0-9]+)/(?P<id>[0-9a-z]+)'
```
dstftw: Do not capture groups if you are not going to use them.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,66 @@
...
+            'ext': 'mp4',
+            'height': 720,
+            'title': 'Oto dlaczego przed zakrętem należy zwolnić.',
+            'duration': 39,
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage("http://ebd.cda.pl/0x0/" + video_id, video_id)
```
dstftw: Use single quotes consistently.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,66 @@
...
+
+        if 'Ten film jest dostępny dla użytkowników premium' in webpage:
+            raise ExtractorError('This video is only available for premium users.', expected=True)
+
+        title = self._html_search_regex(r'<title>(.+?)<\/title>', webpage, 'title', fatal=False)
```
dstftw: No need to escape forward slash.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,66 @@
...
+        duration = [None]
```
dstftw: This *lack-of-`nonlocal`* workaround is clumsy. Why not just return tuple with duration from `_get_format`?

youtube_dl/extractor/cda.py
kasper93: This is expected. `_get_format` is called from within the loop where there are already matched buttons for each format so we know that they are available. We just "select" format we know it's available.

I can add safe check here if you prefer, but the point is to call `_get_format` on well-formed pages and if this is not the case it is likely that the issue is somewhere else.

youtube_dl/extractor/cda.py
kasper93: I've just added duration to each format instead.

kasper93:
Uploaded new version with test for the case when only one format is available.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,80 @@
...
+            format_id = None
+            height = None
+
+            m = re.search(r'<a data-quality="(?P<format_id>[^"]+)" href="[^"]+" class="quality-btn quality-btn-active">(?P<height>[0-9]+)p<\/a>', page)
+            if m:
+                format_id = m.group('format_id')
+                height = int(m.group('height'))
+
+            return {
+                'duration': parse_duration(duration),
```
dstftw: `duration` should not appear in format dict.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,80 @@
...
+                'format_id': format_id,
+                'height': height,
+                'url': self._search_regex(r"url:\\'(.+?)\\'", unpacked, 'url')
```
dstftw: This will break the entire extraction if url is not found for some format.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+        formats = [_get_format(webpage)]
+
+        pattern = re.compile(r'<a data-quality="[^"]+" href="([^"]+)" class="quality-btn">([0-9]+p)<\/a>')
+        for version in re.findall(pattern, webpage):
+            webpage = self._download_webpage(version[0], video_id, 'Fetching %s version information' % version[1])
```
dstftw: This will break the entire extraction if some format's page fails to download. `_download_webpage` should be passed `fatal=False`.

youtube_dl/extractor/cda.py
kasper93: Yes, it will. `_get_format` is to be called on well-formed pages if this is not the case the code should be fixed elsewhere instead of hiding error here.

youtube_dl/extractor/cda.py
kasper93: Not sure what general strategy is. But if this one fails for any reason it means that whole extraction is not good. We know for sure that the format is available and the page is available. So only thing that can really fail if it will be intermediate error. I personally don''t see a point in returning incomplete results instead of error in case of such.

youtube_dl/extractor/cda.py
kasper93: Why not?

youtube_dl/extractor/cda.py
dstftw: It's not about on what to be called. It's about on what it will be called. Malformed webpage is very possible case and even one single format webpage is malformed the **whole** extraction will be broken. The error should not be hidden rather a warning printed (this is what you'll get with `fatal=False`) but it should not break everything.

youtube_dl/extractor/cda.py
dstftw: The general strategy is not to fail if you have enough data to extract at least one format and title. You don't know for sure until you've successfull downloaded the format webpage and ensured it's well formed. You you break on any minor stuff like this you'll get the whole extraction broken. In all the distros and packages where youtube-dl is presented some of which are non-rolling release and never updated. Thus you'll get broken extractor in such distros although could easily prevent it from this by correctly handling minor breakages.

youtube_dl/extractor/cda.py
dstftw: Because it's a [field of info dict](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L180), not format dict.

youtube_dl/extractor/cda.py
kasper93: I don't see format extraction as a minor breakage. But I understand that only one format is enough so indeed it shouldn't fail whole extraction. I guess I'm too used to all or nothing strategy, but this clearly is not good here.

youtube_dl/extractor/cda.py
kasper93: Sure.

youtube_dl/extractor/cda.py
kasper93: Than it should work in the first place. Clearly format dict and info dict fields are mixed later on.Anyway I will make duration back to info dict.

youtube_dl/extractor/cda.py
dstftw: *all-or-nothing* strategy is reasonable for cases where there are no dependencies on 3rd parties or communication between parties is a well defined protocol. Here we don't have such privileges thus are forces to make it as flexible and adaptive as possible.

kasper93:
New version is up.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,96 @@
...
+        format_desc, duration = _get_format(webpage) or (None, None)
+        if format_desc is not None:
+            formats.append(format_desc)
+
+        pattern = re.compile(r'<a data-quality="[^"]+" href="([^"]+)" class="quality-btn">([0-9]+p)<\/a>')
+        for version in re.findall(pattern, webpage):
+            webpage = self._download_webpage(version[0], video_id, 'Fetching %s version information' % version[1], fatal=False)
+            if not webpage:
+                # Manually report warning because empty page is returned when invalid version is requested.
+                self.report_warning('Unable to fetch %s version information' % version[1])
```
dstftw: `errnote` of `_download_webpage` should be used instead.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,96 @@
...
+
+        formats = []
+
+        format_desc, duration = _get_format(webpage) or (None, None)
+        if format_desc is not None:
+            formats.append(format_desc)
+
+        pattern = re.compile(r'<a data-quality="[^"]+" href="([^"]+)" class="quality-btn">([0-9]+p)<\/a>')
+        for version in re.findall(pattern, webpage):
+            webpage = self._download_webpage(version[0], video_id, 'Fetching %s version information' % version[1], fatal=False)
```
dstftw: I've already suggested using `Downloading` as idiomatic wording.

youtube_dl/extractor/cda.py
```diff
@@ -0,0 +1,96 @@
...
+            duration = self._search_regex(r"duration:\\'(.+?)\\'", unpacked, 'duration', fatal=False)
+            format_id = None
+            height = None
+
+            m = re.search(r'<a data-quality="(?P<format_id>[^"]+)" href="[^"]+" class="quality-btn quality-btn-active">(?P<height>[0-9]+)p<\/a>', page)
+            if m:
+                format_id = m.group('format_id')
+                height = int(m.group('height'))
+
+            url = self._search_regex(r"url:\\'(.+?)\\'", unpacked, 'url', fatal=False)
```
dstftw: It would be useful to mention version in `'url'` to know which has failed. Version can be passed outside to `_get_format`.

youtube_dl/extractor/cda.py
kasper93: The reason is in comment. When we request wrong version empty page is returned. And there isn't any warning that's why I made it this way.

youtube_dl/extractor/cda.py
kasper93: Ok. I must have missed that.

kasper93:
Forgot to say yesterday that new version is up. I explictly don't set error note when downloading the page, because empty page should be an error too in our case. So that we can see in log if the page was downloaded but was empty or was downloaded with error.

dstftw:
Cherry-picked in 8b0d7a66ef5451556bb8ae5b085c7bef4c992f8b, thanks.

setaur:
Stopped working...
```
youtube-dl --verbose http://www.cda.pl/video/4816766c
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--verbose', u'http://www.cda.pl/video/4816766c']
[debug] youtube-dl version 2016.06.14
[debug] Python version 2.7.9 - Linux-3.16.0-4-amd64-x86_64-with-debian-8.5
[debug] exe versions: avconv 2.6.9, avprobe 2.6.9, ffmpeg 2.6.9, ffprobe 2.6.9
[debug] Proxy map: {}
[CDA] 4816766c: Downloading webpage
WARNING: unable to extract default url; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
ERROR: No video formats found; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 676, in extract_info
    ie_result = ie.extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 342, in extract
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/cda.py", line 94, in _real_extract
    self._sort_formats(formats)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 863, in _sort_formats
    raise ExtractorError('No video formats found')
ExtractorError: No video formats found; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

yan12125:
@setaur Please open a new issue for that.

Issue #8796, opened 2016-03-08
audioBoom site request


canamcy:
It's possible to use the generic extractor by adding .mp3 to the url.

https://audioboom.com/boos/4233286-ai-safety-and-the-legacy-of-bletchley-park
dl url =>
https://audioboom.com/boos/4233286-ai-safety-and-the-legacy-of-bletchley-park.mp3

[generic] 4233286-ai-safety-and-the-legacy-of-bletchley-park: Requesting header
[redirect] Following redirect to https://d15mj6e6qmt1na.cloudfront.net/attachmen
ts/20081845/ai-safety-and-the-legacy-of-bletchley-park.mp3?audio_clip_id=4233286
&Expires=1457469549&Signature=cZ0sizg2bgY8eloGuohilrF4fMzgtzfleVaUnv81UChf1Ua0XL
X%7ECY0CoHrIJ-wVHB6bagg%7EdP%7E0SvTTqX20cUzMcLskDm3s66sJ1-Ntt5TSnFxo7RdSpUV7fxa4
vgR6VQQyybVPHIV9SNmdEcN3HW0ImXJ8axGJ8S4ce-q%7E4Lg24lO7UcmImyfzyh9Iw8nHSxb0KAyF5X
zDHktgdl9EEjO-ezQ4hWSSfHAAV33dLYVZooB2xvkVEtX0GdIbJ-YT0HM5kI-iGXD77ZVG-a5IoEYntX
IAH193QzVM3iF2CmQAXfbp%7EDeKcMQmA4wrjp0cXW5nTBjSjEH5MbaDDaFpWw__&Key-Pair-Id=APK
AJS5FX6XHKL62XLYQ
[generic] ai-safety-and-the-legacy-of-bletchley-park: Requesting header
[info] Available formats for ai-safety-and-the-legacy-of-bletchley-park:
format code  extension  resolution note
mpeg         mp3        audio only
..\..\..\yo0\M\yodl\youtube-dl\generic\ai-safety-and-the-legacy-of-bletchley-par
k-ai-safety-and-the-legacy-of-bletchley-park.mp3
[generic] 4233286-ai-safety-and-the-legacy-of-bletchley-park: Requesting header
[redirect] Following redirect to https://d15mj6e6qmt1na.cloudfront.net/attachmen
ts/20081845/ai-safety-and-the-legacy-of-bletchley-park.mp3?audio_clip_id=4233286
[...]

PR #8812, opened 2016-03-10
[AudioBoom] Added extractor for AudioBoom

benjamincongdon:
Added simple extractor for AudioBoom.com as requested in #8796.

### DIFF
--- a/youtube_dl/extractor/audioboom.py
+++ b/youtube_dl/extractor/audioboom.py
@@ -1,0 +1,39 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import float_or_none
+
+
+class AudioBoomIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?audioboom\.com/boos/(?P<id>[0-9]+)'
+    _TEST = {
+        'url': 'https://audioboom.com/boos/4279833-3-09-2016-czaban-hour-3?t=0',
+        'md5': '63a8d73a055c6ed0f1e51921a10a5a76',
+        'info_dict': {
+            'id': '4279833',
+            'ext': 'mp3',
+            'title': '3/09/2016 Czaban Hour 3',
+            'description': 'Guest:   Nate Davis - NFL free agency,   Guest:   Stan Gans',
+            'duration': 2245.72
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._og_search_title(webpage)
+
+        download_url = self._og_search_property('audio', webpage, 'url')
+
+        duration = float_or_none(self._html_search_meta(
+            'weibo:audio:duration', webpage, fatal=False))
+
+        return {
+            'id': video_id,
+            'title': title,
+            'url': download_url,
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/audioboom.py
```diff
@@ -0,0 +1,32 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class AudioBoomIE(InfoExtractor):
+    _VALID_URL = r'https?:\/\/(?:www\.)?audioboom\.com\/boos\/(?P<id>[0-9]+)'
```
dstftw: No need to escape forward slash.

benjamincongdon:
Fixed unnecessarily escaped forward slashes.

youtube_dl/extractor/audioboom.py
```diff
@@ -0,0 +1,32 @@
...
+            'title': 'AI Safety and The Legacy of Bletchley Park',
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._og_search_title(webpage)
+        _id = self._search_regex(r'https?://(?:www\.)?audioboom\.com/boos/(?P<id>[0-9]+)', url, 'id')
```
dstftw: This matches just any plain text that meet this pattern. Instead it should extract it via pattern that has embed semantics. This also won't work if actual media is stored under different URI.

youtube_dl/extractor/audioboom.py
```diff
@@ -0,0 +1,32 @@
...
+
+        download_url = self._search_regex(r'^(?P<url>.*?)(\?|$)', url, 'url') + '.mp3'
```
dstftw: What if it's not an mp3?

dstftw:
Description and duration should also be extracted.

benjamincongdon:
Description and duration now extracted. Now get download URL from `<meta>` tags.

youtube_dl/extractor/audioboom.py
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._og_search_title(webpage)
+
+        download_url = self._search_regex(
+            r'<meta content=\'(?P<url>[A-Za-z0-9.\-\:/]+)\' property=\'og:audio\'>', webpage, 'url')
```
dstftw: Use `self._og_search_property` instead.

youtube_dl/extractor/audioboom.py
```diff
@@ -0,0 +1,40 @@
...
+
+        duration = float_or_none(self._search_regex(
+            r'<meta content=\'(?P<duration>[0-9.]+)\' property=\'weibo:audio:duration\'>', webpage, 'duration'))
```
dstftw: Use `self._og_search_property` instead. Duration extraction must not be `fatal`.

benjamincongdon:
Duration extraction now not `fatal`. URL extracted now using `_og_search_property()`

dstftw:
Cherry-picked in 61f317c24c040d051ec6652cb0d4c650c1fc0361, thanks.

PR #8819, opened 2016-03-10
[extractor/common] simplify using data, headers and query params with _download_* methods

remitamine:


youtube_dl/extractor/common.py
```diff
@@ -354,6 +355,17 @@ def _request_webpage(self, url_or_request, video_id, note=None, errnote=None, fa
...
             else:
                 self.to_screen('%s: %s' % (video_id, note))
+        if data or headers or query:
+            headers = headers or {}
+            if isinstance(url_or_request, compat_str):
+                if query:
+                    url_or_request = update_url_query(url_or_request, query)
+                url_or_request = sanitized_Request(url_or_request, data, headers)
+            else:
+                url_or_request.full_url = update_url_query(url_or_request.get_full_url(), query)
```
yan12125: ```full_url``` does not exist on Python 2.x

remitamine: `full_url` are available from python 3.1.
`_Request__original` works for python 2(at least version 2.7).
i will try to look at the source code of python 3.0 and see if one of them can work for it.

yan12125:
I have a different idea: if ```url_or_request``` is a ```Request``` object, ignore ```data```, ```headers``` and ```query``` fields (or raise an ```ExtractorError``` if anyone if given). I guess it simplifies the codes and is suitable for current use cases.

remitamine:
> I have a different idea: if url_or_request is a Request object, ignore data, headers and query fields (or raise an ExtractorError if anyone if given). I guess it simplifies the codes and is suitable for current use cases.

thanks for the idea done in 8b9da7408eaee3d3497e4a069b4b953ae93865fb.
after making the first changes i forget at all why i started this change and tried to make it compatible with both types. 
the initial idea was to avoid creating Request objects manually in the extractors as most the times we're creating the `Request` objects mostly for passing `headers` or `data` to the `urllib`.

yan12125:
Looks good. Could you squash commits and add documentation for the new behavior of ```_request_webpage```?

remitamine:
i will squash the commit.
and about the documentation, the headers and data params are named with the same names used by `Request` so the python documentation has details on this params.

yan12125:
Sorry for misleading. I was talking about the behavor "```data```, ```headers``` and ```query``` are ignored if ```url_or_request``` is already a ```Request``` object".

remitamine:
like adding a comment like this above the code or after function description:
```
# data, headers and query params will be ignored for `Request` objects
```
is this enough?

yan12125:
That's fine. Just for unneeded confusion in case of an extractor fails silently.

remitamine:
may be it's better add this part directly to `sanitized_Request`(add `query` param to `sanitized_Request`):
```
if query:
                url_or_request = update_url_query(url_or_request, query)
```
and for optimization. will it be better to create a `Request` object when the function receive `query` without `headers` and `data`.

yan12125:
If you're going yo do that, chage the name of ```sanitized_Request```. The original name should do nothing except making the request sane. Another option: just merge the functionality of ```sanitized_Request``` into ```_request_webpage```?

remitamine:
for this PR, i will merge only the changes made to _download_* method.

PR #8821, opened 2016-03-10
[YoutubeDL] check for --list-thumbnails immediately after processing them

remitamine:


youtube_dl/YoutubeDL.py
```diff
@@ -1237,6 +1237,10 @@ def process_video_result(self, info_dict, download=True):
         if thumbnails and 'thumbnail' not in info_dict:
             info_dict['thumbnail'] = thumbnails[-1]['url']
 
+        if self.params.get('list_thumbnails'):
+            self.list_thumbnails(info_dict)
+            return
```
dstftw: This can even go before `if` block at 1237 if we remove references to `thumbnail` (1905-1908) in `list_thumbnails`. Since `thumbnails` will always be present at that point when at least one thumbnail is extracted.

remitamine:
> This can even go before if block at 1237 if we remove references to thumbnail (1905-1908) in list_thumbnails.

if `list_thumbnails` is meant only to be called by youtube-dl then it will be safer to change the code to:
```python
    def list_thumbnails(self, info_dict):
        thumbnails = info_dict.get('thumbnails')
        if not thumbnails:
            self.to_screen('[info] No thumbnails present for %s' % info_dict['id'])
            return
```
and
```python
info_dict['thumbnails'] = thumbnails = [{'id': '0', 'url': thumbnail}]
```

youtube_dl/YoutubeDL.py
remitamine: > This can even go before if block at 1237 if we remove references to thumbnail (1905-1908) in list_thumbnails.

if `list_thumbnails` is meant only to be called only by youtube-dl then it will be safer to change the code to:
```python
    def list_thumbnails(self, info_dict):
        thumbnails = info_dict.get('thumbnails')
        if not thumbnails:
            self.to_screen('[info] No thumbnails present for %s' % info_dict['id'])
            return
```
and
```python
info_dict['thumbnails'] = thumbnails = [{'id': '0', 'url': thumbnail}]
```

dstftw:
This is what I've suggested. Assuming the second snipped to go at 1226 you don't really need to specify `id` since will be autonumerated anyway below.

remitamine:
i put the code just to confirm if this is the way to achieve it.
> Assuming the second snipped to go at 1226 you don't really need to specify id since will be autonumerated anyway below.

this is right, i didn't notice that.

remitamine:
> This can even go before if block at 1237 if we remove references to thumbnail (1905-1908) in list_thumbnails.

done in b7b72db9ad248e16a7e26489fee93e52b70948d7.

dstftw:
Looks fine.

Issue #8822, opened 2016-03-10
Broken BBC download (http://www.bbc.co.uk/schoolreport/35744779)


mcepl:
```
matej@mitmanek: ~$ youtube-dl --verbose 'http://www.bbc.co.uk/schoolreport/35744779'
[debug] System config: [u'--prefer-free-formats']
[debug] User config: []
[debug] Command-line args: [u'--verbose', u'http://www.bbc.co.uk/schoolreport/35744779']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.03.06
[debug] Python version 2.7.5 - Linux-3.10.0-357.el7.x86_64-x86_64-with-redhat-7.3-Maipo
[debug] exe versions: ffmpeg 2.6.5, ffprobe 2.6.5, rtmpdump 2.4
[debug] Proxy map: {}
[bbc] 35744779: Downloading webpage
[bbc] p03lsz19: Downloading media selection XML
[bbc] p03lsz19: Downloading m3u8 information
[bbc] p03lsz19: Downloading m3u8 information
ERROR: Unable to extract OpenGraph title; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/youtube_dl/YoutubeDL.py", line 668, in extract_info
    ie_result = ie.extract(url)
  File "/usr/lib/python2.7/site-packages/youtube_dl/extractor/common.py", line 319, in extract
    return self._real_extract(url)
  File "/usr/lib/python2.7/site-packages/youtube_dl/extractor/bbc.py", line 800, in _real_extract
    playlist_title = playlist_title or remove_end(self._og_search_title(webpage), ' - BBC News')
  File "/usr/lib/python2.7/site-packages/youtube_dl/extractor/common.py", line 708, in _og_search_title
    return self._og_search_property('title', html, **kargs)
  File "/usr/lib/python2.7/site-packages/youtube_dl/extractor/common.py", line 696, in _og_search_property
    escaped = self._search_regex(self._og_regexes(prop), html, name, flags=re.DOTALL, **kargs)
  File "/usr/lib/python2.7/site-packages/youtube_dl/extractor/common.py", line 614, in _search_regex
[...]

PR #8826, opened 2016-03-11
[bbc] Make failure to find playlist_title non-fatal

benjamincongdon:
Addresses #8822 by allowing `playlist_title` to be blank if none exists.

### DIFF
--- a/youtube_dl/extractor/bbc.py
+++ b/youtube_dl/extractor/bbc.py
@@ -10,7 +10,6 @@ from ..utils import (
     int_or_none,
     parse_duration,
     parse_iso8601,
-    remove_end,
     unescapeHTML,
 )
 from ..compat import (
@@ -796,9 +795,15 @@ class BBCIE(BBCCoUkIE):
                             entries.append(self._extract_from_playlist_sxml(
                                 playlist.get('progressiveDownloadUrl'), playlist_id, timestamp))
 
+        playlist_title = self._og_search_title(webpage, default=None)
+        playlist_title = playlist_title or self._html_search_regex(
+            r'<title>(.*?)</title>', webpage, 'playlist title')
+
+        playlist_title = self._search_regex(r'(.+)\s*-\s*BBC', playlist_title, 'title', default=playlist_title)
+
+        playlist_description = self._og_search_description(webpage, default=None)
+
         if entries:
-            playlist_title = playlist_title or remove_end(self._og_search_title(webpage), ' - BBC News')
-            playlist_description = playlist_description or self._og_search_description(webpage, default=None)
             return self.playlist_result(entries, playlist_id, playlist_title, playlist_description)
 
         # single video story (e.g. http://www.bbc.com/travel/story/20150625-sri-lankas-spicy-secret)
@@ -828,10 +833,6 @@ class BBCIE(BBCCoUkIE):
                 'formats': formats,
                 'subtitles': subtitles,
             }
-
-        playlist_title = self._html_search_regex(
-            r'<title>(.*?)(?:\s*-\s*BBC [^ ]+)?</title>', webpage, 'playlist title')
-        playlist_description = self._og_search_description(webpage, default=None)
 
         def extract_all(pattern):
             return list(filter(None, map(


youtube_dl/extractor/bbc.py
```diff
@@ -797,7 +797,8 @@ def _real_extract(self, url):
                                 playlist.get('progressiveDownloadUrl'), playlist_id, timestamp))
 
         if entries:
-            playlist_title = playlist_title or remove_end(self._og_search_title(webpage), ' - BBC News')
+            playlist_title = playlist_title or remove_end(
+                self._og_search_title(webpage, default=' - BBC News'), ' - BBC News')
```
dstftw: There is a playlist title that can be extracted. Defaulting to ` - BBC News` in order to replace it then by ` - BBC News` looks clumsy.

benjamincongdon:
I agree, this is a clumsy solution.

In the example provided in #8822, there is no `og:title` tag. Nor does it appear that there is any playlist title given in the page.

Would it be best to default the `playlist_title` to be the `title` of the first entry in `entries`, then?

dstftw:
`School which breaks down barriers in Jerusalem` is the title.

benjamincongdon:
Using the `<title>` tag to extract the title you gave (no `<meta>` tags or references to the media itself had this title).

Couldn't use `remove_end` because some titles have ` - BBC News` suffix while others have ` - BBC School Report`. Using regex instead.

dstftw:
It's already extracted in 837-839 you should only tweak it a little and move before `if entries:`.

benjamincongdon:
Ah. Missed that because of the way it was originally failing. I think this implementation should be good.

youtube_dl/extractor/bbc.py
```diff
@@ -796,9 +795,11 @@ def _real_extract(self, url):
                             entries.append(self._extract_from_playlist_sxml(
                                 playlist.get('progressiveDownloadUrl'), playlist_id, timestamp))
 
+        playlist_title = self._html_search_regex(
+            r'<title>(.*?)(?:\s*-\s*BBC [^ ]+)?</title>', webpage, 'playlist title')
+        playlist_description = self._og_search_description(webpage, default=None)
+
         if entries:
-            playlist_title = playlist_title or remove_end(self._og_search_title(webpage), ' - BBC News')
```
dstftw: `_og_search_title` should be preserved as well.

benjamincongdon:
Alright. Fixed to preserve `_og_search_title`.

youtube_dl/extractor/bbc.py
```diff
@@ -829,10 +831,6 @@ def _real_extract(self, url):
                 'subtitles': subtitles,
             }
 
-        playlist_title = self._html_search_regex(
-            r'<title>(.*?)(?:\s*-\s*BBC [^ ]+)?</title>', webpage, 'playlist title')
-        playlist_description = self._og_search_description(webpage, default=None)
```
dstftw: `playlist_description` is now not extracted on this branch.

youtube_dl/extractor/bbc.py
```diff
@@ -796,9 +795,12 @@ def _real_extract(self, url):
                             entries.append(self._extract_from_playlist_sxml(
                                 playlist.get('progressiveDownloadUrl'), playlist_id, timestamp))
 
+        playlist_title = self._html_search_regex(
```
dstftw: Place right here in the following order:
1. If no title extract with `_og_search_title`.
2. If still no title extract from `<title>`.
3. Extract `playlist_description`.

benjamincongdon:
Fixed.

youtube_dl/extractor/bbc.py
```diff
@@ -796,9 +796,16 @@ def _real_extract(self, url):
                             entries.append(self._extract_from_playlist_sxml(
                                 playlist.get('progressiveDownloadUrl'), playlist_id, timestamp))
 
+        playlist_title = self._og_search_title(webpage, fatal=False)
```
dstftw: Should be `default=None`.

youtube_dl/extractor/bbc.py
```diff
@@ -796,9 +796,16 @@ def _real_extract(self, url):
...
+        if playlist_title:
+            playlist_title = remove_end(playlist_title, ' - BBC News')
+        else:
+            playlist_title = self._html_search_regex(
+                r'<title>(.*?)(?:\s*-\s*BBC [^ ]+)?</title>', webpage, 'playlist title')
```
dstftw: Should be `default=None`. Suffix part is not cropped:
```
[download] Downloading playlist: School which breaks down barriers in Jerusalem - BBC School Report
```

benjamincongdon:
Right. Fixed.

youtube_dl/extractor/bbc.py
```diff
@@ -796,9 +795,15 @@ def _real_extract(self, url):
                             entries.append(self._extract_from_playlist_sxml(
                                 playlist.get('progressiveDownloadUrl'), playlist_id, timestamp))
 
+        playlist_title = self._og_search_title(webpage, default=None, fatal=False)
```
dstftw: `default` implies non fatal.

youtube_dl/extractor/bbc.py
```diff
@@ -796,9 +795,15 @@ def _real_extract(self, url):
...
+        playlist_title = playlist_title or self._html_search_regex(
+            r'<title>(.*?)(?:\s*-\s*BBC [^ ]+)?</title>', webpage, 'playlist title')
```
dstftw: Now ` - BBC ...` part of regex here makes no sense since it's handled below.

youtube_dl/extractor/bbc.py
```diff
@@ -796,9 +795,15 @@ def _real_extract(self, url):
...
+
+        playlist_title = self._search_regex(r'(.*?) - BBC', playlist_title, 'title')
```
dstftw: Add `default=playlist_title` in order not to fail when `<title>` does not contain ` - BBC`. Regex should match at least one characted and should be greedy `r'(.+)\s*-\s*BBC'`.

benjamincongdon:
Changed the regex & added the default.

PR #8842, opened 2016-03-12
[WDR] complete overhaul after relaunch of the site

Boris-de:
The WDR relaunched their site on 2016-02-23 which not only changed the
URL-schema completely but also the layout of their pages.

Apparently the whole "mediathek" now runs on the wdr-domain, so no
separate URL for funkhauseuropa anymore.
There seems to be no explicit handling of video-sizes on the page or in
the URLs anymore. There seems to be only one size for HTML5, but still
several sizes for flash. The new extractor adds all to the list of formats.

There is no metadata for the HTML5-stream, so that the best flash-stream
will always be considered as the "best" format. At least in my tests
this seemed to be true anyway.

As this is my first extractor please let my know if I did anything wrong or
missed something.
I was not sure about the tests, currently they are of course working, but
because of the weird laws in Germany I know that they will be deleted
soonish, so I already added the skips to avoid that they'll break the build
at the time that happens.

### DIFF
--- a/youtube_dl/extractor/wdr.py
+++ b/youtube_dl/extractor/wdr.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-import itertools
 import re
 
 from .common import InfoExtractor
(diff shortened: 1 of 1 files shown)


dstftw:
Duplicate of #8843.

Issue #8831, opened 2016-03-11
mixcloud: unable to extract like count


wvanderp:
root@me-second:~# youtube-dl --verbose https://www.mixcloud.com/djshusta/shusta-special-k-trap/
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--verbose', u'https://www.mixcloud.com/djshusta/shusta-special-k-trap/']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.03.06
[debug] Python version 2.7.6 - Linux-3.13.0-52-generic-x86_64-with-Ubuntu-14.04-trusty
[debug] exe versions: ffmpeg N-77455-g4707497, ffprobe N-77455-g4707497
[debug] Proxy map: {}
[mixcloud] djshusta-shusta-special-k-trap: Downloading webpage
[mixcloud] djshusta-shusta-special-k-trap: Trying mp3 URL
[mixcloud] djshusta-shusta-special-k-trap: Trying m4a URL
WARNING: unable to extract like count; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
[debug] Invoking downloader on u'https://stream13.mixcloud.com/c/m4a/64/9/c/4/6/5864-8702-45ef-92fc-66de3db5ff00.m4a'
[download] Shusta & Special K - TRAP-djshusta-shusta-special-k-trap.m4a has already been downloaded
[download] 100% of 33.57MiB

PR #8845, opened 2016-03-12
[mixcloud] Fix like_count extraction

benjamincongdon:
Update regex / extractor for extracting `like count` from Mixcloud.

Addresses #8831.

### DIFF
--- a/youtube_dl/extractor/mixcloud.py
+++ b/youtube_dl/extractor/mixcloud.py
@@ -85,9 +85,15 @@ class MixcloudIE(InfoExtractor):
         uploader_id = self._search_regex(
             r'\s+"profile": "([^"]+)",', webpage, 'uploader id', fatal=False)
         description = self._og_search_description(webpage)
-        like_count = str_to_int(self._search_regex(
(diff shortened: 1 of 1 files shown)


benjamincongdon:
Address possibility of calling `strip()` on `None`.

youtube_dl/extractor/mixcloud.py
```diff
@@ -85,9 +85,15 @@ def _real_extract(self, url):
...
-            r'\bbutton-favorite\b[^>]+m-ajax-toggle-count="([^"]+)"',
-            webpage, 'like count', fatal=False))
+
+        like_count = self._search_regex(
+            r'\bbutton-favorite[^>]+><[^>]+>([^<]+)',
+            webpage, 'like count', fatal=False)
+
+        if like_count:
+            like_count = like_count.strip()
+            like_count = 1000 * str_to_int(like_count[:-1]) if like_count.endswith('k') else str_to_int(like_count)
```
dstftw: This calculation is wrong. `str_to_int` just removes dots resulting in 11000 instead of 1100.

Issue #8704, opened 2016-02-29
[twitter] Wrong URLs reported for DASH formats


montanaflynn:
When using youtube-dl to find twitter video URL's sometimes it fails, it seems random and fails roughly half the time. Here are the logs of two identical commands running right after each other where one fails and the next one works.

```
montanaflynn@youtube-dl-server:~$ youtube-dl --verbose -g https://twitter.com/jaydingeer/status/700207533655363584
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--verbose', u'-g', u'https://twitter.com/jaydingeer/status/700207533655363584']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.02.27
[debug] Python version 2.7.6 - Linux-3.19.0-49-generic-x86_64-with-Ubuntu-14.04-trusty
[debug] exe versions: ffmpeg N-78050-g209f50e-static, ffprobe N-78050-g209f50e-static
[debug] Proxy map: {}
ERROR: There's no video in this tweet.; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 666, in extract_info
    ie_result = ie.extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 316, in extract
    return self._real_extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/twitter.py", line 237, in _real_extract
    raise ExtractorError('There\'s no video in this tweet.')
ExtractorError: There's no video in this tweet.; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

```
montanaflynn@youtube-dl-server:~$ youtube-dl --verbose -g https://twitter.com/jaydingeer/status/700207533655363584
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--verbose', u'-g', u'https://twitter.com/jaydingeer/status/700207533655363584']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.02.27
[...]

PR #8848, opened 2016-03-13
return mpd_url for dash segmented formats(#8704)

remitamine:
mpd_url construction rules can be found in Annex C section 4 in ISO/IEC 23009-1

- separate downloader specific params from other info_dict keys
- the mpd_url can be used with external programs like MP4Box and VLC 3
- it's needed to be passed to dash downloader to support live stream download

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -23,6 +23,7 @@ from ..compat import (
     compat_urllib_error,
     compat_urllib_parse,
     compat_urlparse,
+    compat_urllib_parse_urlparse,
 )
 from ..utils import (
     NO_DEFAULT,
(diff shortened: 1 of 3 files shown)


youtube_dl/extractor/common.py
```diff
@@ -1509,18 +1513,31 @@ def extract_multisegment_info(element, ms_parent_info):
                             media_template = re.sub(r'\$(Number|Bandwidth)(?:%(0\d+)d)?\$', r'%(\1)\2d', media_template)
                             media_template.replace('$$', '$')
                             representation_ms_info['segment_urls'] = [media_template % {'Number': segment_number, 'Bandwidth': representation_attrib.get('bandwidth')} for segment_number in range(representation_ms_info['start_number'], representation_ms_info['total_number'] + representation_ms_info['start_number'])]
-                        if 'segment_urls' in representation_ms_info:
+                        segment_urls = representation_ms_info.get('segment_urls')
+                        if segment_urls:
+                            fragment_mpd_url = segment_urls[0]
+                            if parsed_mpd_url:
+                                fragment_parts = ((period, 'id', 'pariod'), (adaptation_set, 'id', 'as'), (adaptation_set, 'group', 'track'))
```
yan12125: A typo? ```pariod```

remitamine: yes it is.

PR #8868, opened 2016-03-15
[Udemy] Fix issue 8753

quanhua92:
Fix issue in https://github.com/rg3/youtube-dl/issues/8753
The web page changes the HTML in new version of video player so it causes the error of "Unable to extract course id". I change the regex and everything works now.

### DIFF
--- a/youtube_dl/extractor/udemy.py
+++ b/youtube_dl/extractor/udemy.py
@@ -142,9 +142,12 @@ class UdemyIE(InfoExtractor):
         lecture_id = self._match_id(url)
 
         webpage = self._download_webpage(url, lecture_id)
-
-        course_id = self._search_regex(
+        try:
+            course_id = self._search_regex(
             r'data-course-id=["\'](\d+)', webpage, 'course id')
+        except ExtractorError as e:
+            course_id = self._search_regex(
+               r'&quot;id&quot;: (\d+)', webpage, 'course id')
 
         try:
             lecture = self._download_lecture(course_id, lecture_id)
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/udemy.py
```diff
@@ -142,9 +142,12 @@ def _real_extract(self, url):
         lecture_id = self._match_id(url)
 
         webpage = self._download_webpage(url, lecture_id)
-
-        course_id = self._search_regex(
+        try:
+            course_id = self._search_regex(
             r'data-course-id=["\'](\d+)', webpage, 'course id')
```
dstftw: Pass tuple of regexes to `_search_regex` instead.

Cmakota:
Thank you sir!

youtube_dl/extractor/udemy.py
quanhua92: Sorry. I am new to this project and don't know how to do that.

dstftw:
```python
course_id = self._search_regex(
    (r'data-course-id=["\'](\d+)', r'&quot;id&quot;: (\d+)'),
    webpage, 'course id')
```

quanhua92: Thanks. Just fix this.

quanhua92:
@Cmakota Just fix it with tuble search_regex. It is working right now. Until this pull request is merged into a release, you can change the udemy.py as you said to use this fix.

PR #8876, opened 2016-03-16
[extractor/common] add helper method to extract html5 media entries

remitamine:
as @dstftw suggested https://github.com/rg3/youtube-dl/pull/6811#issuecomment-139282386, a separate method to extract html5 media entries.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -78,6 +78,7 @@ from youtube_dl.utils import (
     cli_option,
     cli_valueless_option,
     cli_bool_option,
+    parse_codecs,
 )
 from youtube_dl.compat import (
     compat_chr,
@@ -578,6 +579,29 @@ class TestUtil(unittest.TestCase):
         self.assertTrue(
             limit_length('foo bar baz asd', 12).startswith('foo bar'))
         self.assertTrue('...' in limit_length('foo bar baz asd', 12))
+
+    def test_parse_codecs(self):
+        self.assertEqual(parse_codecs(''), {})
+        self.assertEqual(parse_codecs('avc1.77.30, mp4a.40.2'), {
+            'vcodec': 'avc1.77.30',
+            'acodec': 'mp4a.40.2',
+        })
+        self.assertEqual(parse_codecs('mp4a.40.2'), {
+            'vcodec': 'none',
+            'acodec': 'mp4a.40.2',
+        })
+        self.assertEqual(parse_codecs('mp4a.40.5,avc1.42001e'), {
+            'vcodec': 'avc1.42001e',
+            'acodec': 'mp4a.40.5',
+        })
+        self.assertEqual(parse_codecs('avc3.640028'), {
+            'vcodec': 'avc3.640028',
+            'acodec': 'none',
+        })
+        self.assertEqual(parse_codecs(', h264,,newcodec,aac'), {
+            'vcodec': 'h264',
+            'acodec': 'aac',
+        })
 
     def test_escape_rfc3986(self):
         reserved = "!*'();:@&=+$,/?#[]"
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -54,6 +54,8 @@ from ..utils import (
     update_Request,
     update_url_query,
     parse_m3u8_attributes,
+    extract_attributes,
+    parse_codecs,
 )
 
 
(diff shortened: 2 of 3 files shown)


dstftw:
1. Tests for `parse_codecs` should be added (e.g. from aforementioned RFC).
2. Tests for html5 video/audio should be added.
3. This should be adopted into the generic extractor and replace the current implementation.

youtube_dl/extractor/common.py
```diff
@@ -1539,6 +1541,59 @@ def extract_multisegment_info(element, ms_parent_info):
         self._sort_formats(formats)
         return formats
 
+    def _extract_html5_media_entries(self, url, webpage):
```
dstftw: This should be called `parse` in order to match the naming convention of the other parse/extract methods since it does not download the webpage but just parses it.

remitamine:
sorry for late response.
> Tests for parse_codecs should be added (e.g. from aforementioned RFC).

may be putting the RFC url is misleading, the code comply only with the common use of this RFC(vcodec, acodec or acodec) while the RFC doesn't specify the order or the number of codecs.
examples of the cases that can be handled by this code https://www.w3.org/TR/html/embedded-content-0.html#the-source-element.

yan12125:
> the RFC doesn't specify the order or the number of codecs.

Tests can be examples in the wild, and this line should be mentioned in ```parse_codecs```.

youtube_dl/extractor/common.py
```diff
@@ -1539,6 +1541,59 @@ def extract_multisegment_info(element, ms_parent_info):
         self._sort_formats(formats)
         return formats
 
+    def _parse_html5_media_entries(self, url, webpage):
```
yan12125: I guess something like ```base_url``` is a better name than ```url```.

youtube_dl/extractor/common.py
```diff
@@ -1539,6 +1541,59 @@ def extract_multisegment_info(element, ms_parent_info):
...
+        def absolute_url(video_url):
+            return compat_urlparse.urljoin(url, video_url)
+
+        def parse_content_type(ct):
```
yan12125: I guess using ```content_type``` instead of ```ct``` improves readability.

youtube_dl/extractor/common.py
```diff
@@ -1539,6 +1541,59 @@ def extract_multisegment_info(element, ms_parent_info):
...
+            ctr = re.search(r'(?P<mimetype>[^/]+/[^;]+)(?:;\s*codecs="?(?P<codecs>[^"]+))?', ct)
+            if ctr:
+                mimetype, codecs = ctr.groups()
+                f = parse_codecs(codecs)
+                f['ext'] = mimetype2ext(mimetype)
+                return f
+            return {}
+
+        entries = []
+        for media_attributes, media_type, media_content in re.findall(r'(?s)(<(video|audio)[^>]*>)(.*?)</\2>', webpage):
```
yan12125: ```r'(?s)(<(?P<tag>video|audio)[^>]*>)(.*?)</(?P=tag)>'``` is better as numbers may change in the future.

remitamine:
> Tests can be examples in the wild, and this line should be mentioned in parse_codecs.

i will add the line in a comment and maybe a TODO to improve it to detect when the codecs are not in order(somtimes i see the codecs not detected correctly for m3u8 formats so it can be useful to improve it and use it whenever we have to parse the codecs string).

youtube_dl/extractor/common.py
```diff
@@ -1539,6 +1541,59 @@ def extract_multisegment_info(element, ms_parent_info):
...
+                'subtitles': {},
+            }
+            if media_attributes:
+                media_attributes = extract_attributes(media_attributes)
+                src = media_attributes.get('src')
+                if src:
+                    media_info['formats'].append({'url': absolute_url(src)})
+                media_info['thumbnail'] = media_attributes.get('poster')
+            if media_content:
+                for source_attributes in re.findall(r'(?s)<source[^>]*>', media_content):
```
yan12125: ```(?s)``` has no effects in regular expressions without ```.```

youtube_dl/extractor/common.py
```diff
@@ -1539,6 +1541,59 @@ def extract_multisegment_info(element, ms_parent_info):
...
+            ctr = re.search(r'(?P<mimetype>[^/]+/[^;]+)(?:;\s*codecs="?(?P<codecs>[^"]+))?', content_type)
+            if ctr:
+                mimetype, codecs = ctr.groups()
+                f = parse_codecs(codecs)
+                f['ext'] = mimetype2ext(mimetype)
+                return f
+            return {}
+
+        entries = []
+        for media_attributes, media_type, media_content in re.findall(r'(?s)(<(?P<tag>video|audio)[^>]*>)(.*?)</(?P=tag)>', webpage):
```
yan12125: ```media_attributes``` is somewhat misleading. It's the whole tag but not just attributes.

yan12125: And seems ```media_type``` is not used at all?

remitamine:
> (?s) has no effects in regular expressions without 
> media_attributes is somewhat misleading. It's the whole tag but not just attributes.
> And seems media_type is not used at all?

done(used `media_type` for detecting audio only formats).

yan12125:
Could you add tests for ```parse_codecs```?

remitamine:
> Could you add tests for parse_codecs?
Tests can be examples in the wild, and this line should be mentioned in parse_codecs.

done.

test/test_utils.py
```diff
@@ -579,6 +580,22 @@ def test_limit_length(self):
...
+        self.assertEqual(parse_codecs('avc1.77.30, mp4a.40.2'), {
+            'vcodec': 'avc1.77.30',
+            'acodec': 'mp4a.40.2',
+        })
+        self.assertEqual(parse_codecs('mp4a.40.2'), {
+            'vcodec': 'none',
+            'acodec': 'mp4a.40.2',
+        })
+        # http://once.unicornmedia.com/now/master/playlist/bb0b18ba-64f5-4b1b-a29f-0ac252f06b68/77a785f3-5188-4806-b788-0893a61634ed/93677179-2d99-4ef4-9e17-fe70d49abfbf/content.m3u8
+        # self.assertEqual(parse_codecs('mp4a.40.5,avc1.42001e'), {
```
yan12125: Such cases should be handled, too. I guess a possible approach is creating a table with common video and audio codecs. If given codecs are not on the table, fallback to ```video,audio``` order.

remitamine: this are some codecs that i know about:
vcodecs: avc1, vp9, vp8, avc3, hev1, h263, h264, mp4v
acodecs: mp4a, opus, vorbis, mp3, aac
is there other codecs to add?

yan12125: Some more: avc2, avc3, avc4. These would be enough.

remitamine:
> Such cases should be handled, too. I guess a possible approach is creating a table with common video and audio codecs. If given codecs are not on the table, fallback to video,audio order.

done.

yan12125:
Thanks @remitamine the next step is to replace the current implementation in ```generic.py```. I'll fix some extractors that uses HTML5 first.

PR #8891, opened 2016-03-18
Added format_id to the filers on -f.

JohnPeel:
Useful for certain websites (e.g. Funimation) where the formats are episode specific but all have something in common.

I wanted to be able to specify only the japanese formats, and this allows it.

Example
`youtube-dl -f best[format_id*=JPN] "url"`


dstftw:
Cherry-picked in d5aacf9a90e0855976401b6085ac56b66ca09d12.

Issue #3101, opened 2014-06-18
request: Per service configuration


gh0stl4b:
I love having the option to configure youtube-dl globaly, but it would be great to be able to specify options per service, so like one directory for YouTube, another for SoundCloud and another for Twitch.  
I would presume it should look something like this:  

```
[all]
--no-part

[youtube]
-o "%(uploader)s\%(title)s.%(ext)s"

[youtube:playlists]
-o "%(uploader)s\%(playlist)s\%(title)s.%(ext)s"
--ignore-errors

[soundcloud]
--audio-format m4a

[udemy]
--username MyUsername
--password MyPassword

[twitch]
--username MyUsername
--password MyPassword
```

I hope you get the basic gist.  
Thank you :)

yan12125:
Would it be a good idea to use a Python script as the configuration file? For example:
```Python
# file ~/.config/youtube-dl/config.py

def YoutubeDLFlags(url, ie_key, other_flags):
    flags = []
    if ie_key == 'YouTube':
        flags.extend(['-f', 'bestvideo[ext!=webm]+bestaudio[ext!=webm]'])
    elif ie_key == 'Youku':
        flags.extend(['--proxy', 'hostname:port'])
    return flags
```
There are indeed lots of implementation issues, but I think it's a good start. A fixed set of flags is simply not flexible enough.

jaimeMF:
@yan12125 Could you explain why a simple config file with sections is not enough? For the example you posted all you would need is:
```ini
[YouTube]
format = bestvideo[ext!=webm]+bestaudio[ext!=webm]

[Youku]
proxy = hostname:port
```
(I like not using leading dashes, but it could be made optional to support the old format).

yan12125:
Sorry for a bad example. Python scripts could solve #6745 as well. In my case, I use a different set of flags when developing youtube-dl. For example:
```Python
import os.path

def YoutubeDLFlags(url, ie_key, other_flags):
    if os.path.exists('./youtube_dl/__main__.py'):
        return ['-v', '--write-pages']
    flags = []
    if ie_key == 'YouTube':
        flags.extend(['-f', 'bestvideo[ext!=webm]+bestaudio[ext!=webm]'])
    elif ie_key == 'Youku':
        flags.extend(['--proxy', 'hostname:port'])
    return flags
```

jaimeMF:
[...]

PR #8901, opened 2016-03-19
Add support for website specific configuration

jaimeMF:
This is for #3101.

Still to be done:
- [x] Add filenames for Windows
- [x] Read a global configuration file, porbably `/etc/youtube-dl.ini`
- [x] Postprocessors support (-x, --embed-sub and others).
- [x] Add documentation to the manual
- [ ] Deprecate old config files and add migration guide?

Example config:

``` ini
extract-audio=

[youtube]
audio-format=mp3

[youtube:playlist]
output=%(playlist)s/%(title)s.%(ext)s
```

Options that accept no arguments must still be written as `write-thumbnail=` because the  `allow_no_value`  option for `ConfigParser`  is not supported in python 2.6.

There's probably some parameter that is still global which should be local, I'll try to look for them, but point them out if you find one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,6 +427,24 @@ Note that options in configuration file are just the same options aka switches u
 
 You can use `--ignore-config` if you want to disable the configuration file for a particular youtube-dl run.
 
+### Per extractor configuration
+
+In addition to globally setting options, you can also set different options for each extractor.
+A different set of files is used: `~/.config/youtube-dl/config.ini` and `/etc/youtube-dl.ini` on Unix, `%APPDATA%\youtube-dl\config.ini` or `C:\Users\<user name>\youtube-dl.ini` on Windows.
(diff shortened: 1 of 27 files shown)


jaimeMF:
I have added what I've planned and added some initial documentation.

Do you think using the short names for the options should be supported? It's probably trivial to add, but I personally prefer using the long in config files to help me remembering what they do.

Hrxn:
> Do you think using the short names for the options should be supported?

For the sake of consistency, yes..

jaimeMF:
> > Do you think using the short names for the options should be supported?
> 
> For the sake of consistency, yes..

Done.

remitamine:
there currently a problem with passing external `--external-downloader-args` in the config #9268.
is it possible to add downloader specific args to the config file.
somthing like:
```
[aria2c]
external-downloader-args="-x 16"

[ffmpeg]
external-downloader-args="-loglevel 16'
```

alimony:
@jaimeMF Are you still working on this?

PR #8902, opened 2016-03-19
Makefile: make it compatible with bmake

jaimeMF:
It's the portable version of BSD make: http://crufty.net/help/sjg/bmake.html
The syntax for conditionals is different in GNU make and BSD make, so we use the shell

Since we don't use any other special feature of GNU make, I think it's nice to support both.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,15 +12,7 @@ SHAREDIR ?= $(PREFIX)/share
 PYTHON ?= /usr/bin/env python
 
 # set SYSCONFDIR to /etc if PREFIX=/usr or PREFIX=/usr/local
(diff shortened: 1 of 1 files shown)


yan12125:
Maybe better to document the change in ```README.md```? Something like:
```
If you want to create a build of youtube-dl yourself, you'll need

* python
* make (GNU make and bmake are supported)
* pandoc
* zip
* nosetests

```

jaimeMF:
@yan12125 thanks for the suggestion, added in 920d318d3cf70ee0b80cc67e7d3f85b5d45a20a7.

PR #8924, opened 2016-03-21
[tunein] fix Stream parsing (for good)

bubbleguuum:


bubbleguuum:
The CI build fails but all Tunein tests pass. WTF ?

PR #8956, opened 2016-03-24
Ora extractor fixes

fstirlitz:
The OraTV extractor broke recently ("Unable to extract current video"); one of the patches in this PR fixes that and additionally adds support for unsafespeech.com URLs. The other improves the `js_to_json` function to handle more JS object literals (and handle others correctly).

### DIFF
--- a/youtube_dl/extractor/ora.py
+++ b/youtube_dl/extractor/ora.py
@@ -6,13 +6,14 @@ from .common import InfoExtractor
 from ..compat import compat_urlparse
 from ..utils import (
     get_element_by_attribute,
+    js_to_json,
     qualities,
     unescapeHTML,
 )
 
 
 class OraTVIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?ora\.tv/([^/]+/)*(?P<id>[^/\?#]+)'
+    _VALID_URL = r'https?://(?:www\.)?(ora\.tv|unsafespeech\.com)/([^/]+/)*(?P<id>[^/\?#]+)'
     _TEST = {
         'url': 'https://www.ora.tv/larrykingnow/2015/12/16/vine-youtube-stars-zach-king-king-bach-on-their-viral-videos-0_36jupg6090pq',
         'md5': 'fa33717591c631ec93b04b0e330df786',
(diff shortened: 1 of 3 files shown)


youtube_dl/utils.py
```diff
@@ -1847,24 +1847,26 @@ def fix_kv(m):
...
-                "\\'": "'",
+        elif v.startswith('/*') or v == ',':
+            return ""
+
+        if v[0] in ("'", '"'):
+            v = re.sub(r'(?s)\\.|"', lambda m: {
                 '"': '\\"',
-            }[m.group(0)], v)
+                "\\'": "'",
+                '\\\n': '',
```
yan12125: Should be ```'\n'```. Removing newlines is the work of individual extractors.

fstirlitz: No, in JavaScript
```js
"a\
b" === "ab"
````
evaluates as true. Try it yourself.

Comment (yan12125) [inline youtube_dl/utils.py]: You're right. Could you add a comment here?

File: youtube_dl/extractor/ora.py [hunk 2]
```diff
@@ -6,13 +6,14 @@
...
     get_element_by_attribute,
+    js_to_json,
     qualities,
     unescapeHTML,
 )
 
 
 class OraTVIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?ora\.tv/([^/]+/)*(?P<id>[^/\?#]+)'
+    _VALID_URL = r'https?://(?:www\.)?(ora\.tv|unsafespeech\.com)/([^/]+/)*(?P<id>[^/\?#]+)'
```
yan12125: Don't use capturing groups here; should be ```(?:ora\.tv|unsafespeech\.com)```

fstirlitz: There's a capturing group right after the backlash... is it really such a huge problem?

yan12125: Capturing groups increases memory usage. It also brings extra fields to the result of ```re.match(self._VALID_URL, url).groups()```. In some extractors the correctness relies on the number of capturing groups, so unnecassary groups should not be captured. Now it's a convention in youtube-dl: don't use capturing groups unless they're necessary.

dstftw:
Cherry-picked, thanks.

PR #8958, opened 2016-03-24
[mnet] Add new extractor

Kagami:


youtube_dl/extractor/mnet.py
```diff
@@ -0,0 +1,72 @@
...
+        file_url = rtmp_info['serverurl'] + rtmp_info['fileurl']
+
+        return {
+            'id': video_id,
+            'title': title,
+            'duration': duration,
+            'thumbnail': thumbnail,
+            'url': file_url,
+            # RTMP URL contains ".mp4" but resulting container is
+            # actually FLV.
```
dstftw: rtmp's container is always flv by design. This comment is superfluous.

youtube_dl/extractor/mnet.py
```diff
@@ -0,0 +1,72 @@
...
+            'url': 'http://www.mnet.com/tv/vod/vod_view.asp?clip_id=172790&tabMenu=',
+            'md5': '1552891a585b9f640a64e4ba89d51f6e',
+            'info_dict': {
+                'id': '172790',
+                'ext': 'flv',
+                'title': '[직캠]일대일아이컨택ㅣ최유정 -  ♬Yum-Yum(얌얌) @ 콘셉트 평가',
+                'thumbnail': 're:^https?://.*\.jpg$',
+                'duration': 209,
+            },
+        },
```
dstftw: Leave only one test if all test the same extraction scenario. For different URL schemas use `only_matching` test.

youtube_dl/extractor/mnet.py
```diff
@@ -0,0 +1,72 @@
...
+        info_url = 'http://content.api.mnet.com/player/vodConfig?id=%s' % video_id
+        info = self._download_json(info_url, video_id)
+        info = info['data']['info']
+
+        title = info['title']
+        duration = parse_duration(info.get('time'))
+        try:
+            thumbnail = info['cover']['maxres']['url']
+        except KeyError:
+            thumbnail = None
```
dstftw: We usually use default fallbacks instead for shorter code: `info.get('cover', {}).get('maxres', {}).get('url')`

Kagami:
@dstftw addressed all comments and squashed.

youtube_dl/extractor/mnet.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        info_url = 'http://content.api.mnet.com/player/vodConfig?id=%s' % video_id
+        info = self._download_json(info_url, video_id)
+        info = info['data']['info']
+
+        title = info['title']
+        duration = parse_duration(info.get('time'))
+        thumbnail = info.get('cover', {}).get('maxres', {}).get('url')
```
dstftw: There are multiple thumbnails available that can all be extracted.

dstftw:
There is also an upload timestamp in `data/info/date` and age limit data in `data/info/adult`.

Kagami:
@dstftw added few more fields from info JSON.

youtube_dl/extractor/mnet.py
```diff
@@ -0,0 +1,73 @@
...
+        description = info.get('ment')
+        duration = parse_duration(info.get('time'))
+        upload_date = unified_strdate(info.get('date'))
+        age_limit = 0 if info.get('adult', 'N') == 'N' else 18
+        thumbnails = [
+            {
+                'id': thumb_format,
+                'url': thumb.get('url'),
+                'width': int(thumb.get('width', 0)),
+                'height': int(thumb.get('height', 0)),
```
dstftw: `utils.int_or_none` should be used instead.

youtube_dl/extractor/mnet.py
```diff
@@ -0,0 +1,73 @@
...
+        video_id = self._match_id(url)
+        info_url = 'http://content.api.mnet.com/player/vodConfig?id=%s' % video_id
+        info = self._download_json(info_url, video_id)
+        info = info['data']['info']
+
+        title = info['title']
+        description = info.get('ment')
+        duration = parse_duration(info.get('time'))
+        upload_date = unified_strdate(info.get('date'))
+        age_limit = 0 if info.get('adult', 'N') == 'N' else 18
```
dstftw: `age_limit` should be `None` when `adult` is missing.

youtube_dl/extractor/mnet.py
```diff
@@ -0,0 +1,73 @@
...
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        info_url = 'http://content.api.mnet.com/player/vodConfig?id=%s' % video_id
+        info = self._download_json(info_url, video_id)
+        info = info['data']['info']
+
+        title = info['title']
+        description = info.get('ment')
+        duration = parse_duration(info.get('time'))
+        upload_date = unified_strdate(info.get('date'))
```
dstftw: `utils.parse_iso8601` should be used instead and extracted as `timestamp`.

youtube_dl/extractor/mnet.py
```diff
@@ -0,0 +1,73 @@
...
+                'url': thumb.get('url'),
+                'width': int(thumb.get('width', 0)),
+                'height': int(thumb.get('height', 0)),
+            }
+            for (thumb_format, thumb) in info.get('cover', {}).items()
+        ]
+
+        rtmp_info_url = info['cdn'] + 'CLIP'
+        rtmp_info = self._download_json(rtmp_info_url, video_id)
+        file_url = rtmp_info['serverurl'] + rtmp_info['fileurl']
```
dstftw: Video URL should be extracted the very first right after the title.

Kagami: Explain please. What do you want to see?

dstftw: Move in right after `title = info['title']`.

Kagami:
Addressed all comments.

dstftw:
Cherry-picked in e0317686666f9de4a6eca3fc26ede32e664f2bec, thanks.

Kagami:
Thanks. "Kagami Hiiragi" will be fine.

PR #9025, opened 2016-03-29
screencast.com: support missing www

mart-e:
- The "www." part of the URL is not mandatory. Modified a test to test both URL formats
- use the `<title>` tag of the page to find filename

### DIFF
--- a/youtube_dl/extractor/screencast.py
+++ b/youtube_dl/extractor/screencast.py
@@ -12,7 +12,7 @@ from ..utils import (
 
 
 class ScreencastIE(InfoExtractor):
-    _VALID_URL = r'https?://www\.screencast\.com/t/(?P<id>[a-zA-Z0-9]+)'
+    _VALID_URL = r'https?://(?:www\.)?screencast\.com/t/(?P<id>[a-zA-Z0-9]+)'
     _TESTS = [{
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/screencast.py
```diff
@@ -34,7 +34,7 @@ class ScreencastIE(InfoExtractor):
             'thumbnail': 're:^https?://.*\.(?:gif|jpg)$',
         }
     }, {
-        'url': 'http://www.screencast.com/t/aAB3iowa',
+        'url': 'http://screencast.com/t/aAB3iowa',
```
dstftw: Keep this test as it was and add another one `only_matching`.

youtube_dl/extractor/screencast.py
```diff
@@ -95,7 +95,8 @@ def _real_extract(self, url):
         if title is None:
             title = self._html_search_regex(
                 [r'<b>Title:</b> ([^<]*)</div>',
-                 r'class="tabSeperator">></span><span class="tabText">(.*?)<'],
+                 r'class="tabSeperator">></span><span class="tabText">(.*?)<',
+                 r'<title>([^<]*)</title>'],
```
dstftw: Add a test for this scenario.

PR #9028, opened 2016-03-29
Update crunchyroll.py

DarkstaIkers:
Yes: http://i.imgur.com/DxFESLp.png
No: http://i.imgur.com/0znlYaz.png

### DIFF
--- a/youtube_dl/extractor/crunchyroll.py
+++ b/youtube_dl/extractor/crunchyroll.py
@@ -184,7 +184,7 @@ class CrunchyrollIE(CrunchyrollBaseIE):
         output += 'WrapStyle: %s\n' % sub_root.attrib['wrap_style']
         output += 'PlayResX: %s\n' % sub_root.attrib['play_res_x']
         output += 'PlayResY: %s\n' % sub_root.attrib['play_res_y']
-        output += """ScaledBorderAndShadow: yes
+        output += """ScaledBorderAndShadow: no
 
 [V4+ Styles]
 Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding


DarkstaIkers:
Yes: http://i.imgur.com/DxFESLp.png
No: http://i.imgur.com/0znlYaz.png

lachs0r:
This is wrong and breaks typesetting in a lot of scripts. Please revert.

lachs0r:
Rationale: Crunchyroll pretty much always uses `ScaleBorderAndShadow: yes`, deliberately so (although the player API doesn’t seem to export that property). It just so happens that the Flash-based player on the website doesn’t know how to scale borders, and it really doesn’t help readability, especially on high-resolution displays. This is evicenced by the fact that the hardsubbed encodes (e.g. for consoles or the Windows app) have the same border size at any resolution. If you want hairline borders, override it using video player settings.

yan12125:
> This is wrong and breaks typesetting in a lot of scripts.

Sorry I didn't get it. Subtitles downloaded by youtube-dl are unrelated to players on web pages, are they? Maybe some screenshots explain things better.

The reason for this change is that youtube-dl should add no decorations to videos/audios/subtitles unless requested. Just like what you've said, you can also override ScaledBorderAndShadow in video players.

lachs0r:
The point is that this *does* make youtube-dl change the styling of Crunchyroll’s scripts. `ScaledBorderAndShadow: yes` is very much intended. It is used in a very large percentage of the original ASS scripts that are uploaded to CR’s site (`no` is only used accidentally). Some subtitling teams use a subset of ASS’ more advanced features to get prettier signs in the hardsubbed streams (which are rendered using libass) while retaining Flash player compatibility. Setting this to `no` affects more than just the border width. It also affects blur scaling, and, more importantly, 3D transforms. In opaque box border mode, the size is pretty wrong as well. Expect all of these to get used much more frequently as soon as the Flash player is replaced (which has been on CR’s TODO list for quite some time and is very likely to happen this year).

yan12125:
> It is used in a very large percentage of the original ASS scripts that are uploaded to CR’s site (no is only used accidentally)

Thanks for this information. It can be a strong reason for reverting previous changes. Does this change work for you?
```Diff
diff --git a/youtube_dl/extractor/crunchyroll.py b/youtube_dl/extractor/crunchyroll.py
index f811c7f33..109d1c5a8 100644
--- a/youtube_dl/extractor/crunchyroll.py
+++ b/youtube_dl/extractor/crunchyroll.py
@@ -255,8 +255,7 @@ class CrunchyrollIE(CrunchyrollBaseIE):
         output += 'WrapStyle: %s\n' % sub_root.attrib['wrap_style']
         output += 'PlayResX: %s\n' % sub_root.attrib['play_res_x']
         output += 'PlayResY: %s\n' % sub_root.attrib['play_res_y']
-        output += """ScaledBorderAndShadow: no
-
+        output += """
 [V4+ Styles]
 Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
 """
```
As libass [uses yes as the default for ScaledBorderAndShadow](https://github.com/libass/libass/blob/e54c123/libass/ass.c#L1344), I guess leaving rendering configurations to libass and video players is better than handling them in youtube-dl.

lachs0r:
Yes, that should do the trick.

Issue #9106, opened 2016-04-07
[youtube] hangs with HLS format


shaicoleman:
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.04.06**
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)

When downloading a youtube video with HLS, it just hangs

```
$ youtube-dl -v -f 94 https://www.youtube.com/watch?v=Gjv9LrRPo0Q

[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--verbose', u'-f', u'94', u'https://www.youtube.com/watch?v=Gjv9LrRPo0Q']
[debug] Encodings: locale ANSI_X3.4-1968, fs ANSI_X3.4-1968, out None, pref ANSI_X3.4-1968
[debug] youtube-dl version 2016.04.06
[debug] Python version 2.7.11+ - Linux-4.4.0-14-generic-x86_64-with-Ubuntu-16.04-xenial
[debug] exe versions: ffmpeg 2.8.6-1ubuntu2, ffprobe 2.8.6-1ubuntu2
[debug] Proxy map: {}
[youtube] Gjv9LrRPo0Q: Downloading webpage
[youtube] Gjv9LrRPo0Q: Downloading video info webpage
[youtube] Gjv9LrRPo0Q: Extracting video information
[youtube] Gjv9LrRPo0Q: Downloading formats manifest
[youtube] Gjv9LrRPo0Q: Downloading MPD manifest
[...]

PR #9109, opened 2016-04-07
move mpd multisegement parsing into dash downloader

remitamine:
related to #9106.
as a first step move the code to parse mpd segements into dash downloader which will simplify adding support for live streams by:
- update the segment urls by downloading and parsing the mpd every `MPD@minimumUpdatePeriod`.
- checking for segement availabiliy
...


yan12125:
Is there a case needs live streaming in DASH?

yan12125:
Oh I see #9106.

PR #9110, opened 2016-04-07
[utils] imporove parse_duration to handle more formats

remitamine:
related to the previous work on dash support https://github.com/rg3/youtube-dl/pull/8408#issuecomment-179783768.

### DIFF
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -1540,44 +1540,46 @@ def parse_duration(s):
 
     s = s.strip()
 
-    m = re.match(
-        r'''(?ix)(?:P?T)?
-        (?:
-            (?P<only_mins>[0-9.]+)\s*(?:mins?\.?|minutes?)\s*|
-            (?P<only_hours>[0-9.]+)\s*(?:hours?)|
-
-            \s*(?P<hours_reversed>[0-9]+)\s*(?:[:h]|hours?)\s*(?P<mins_reversed>[0-9]+)\s*(?:[:m]|mins?\.?|minutes?)\s*|
-            (?:
+    days, hours, mins, secs, ms = [None] * 5
+    m = re.match(r'(?:(?:(?:(?P<days>[0-9]+):)?(?P<hours>[0-9]+):)?(?P<mins>[0-9]+):)?(?P<secs>[0-9]+)(?P<ms>\.[0-9]+)?$', s)
+    if m:
+        days, hours, mins, secs, ms = m.groups()
+    else:
+        m = re.match(
+            r'''(?ix)(?:P?T)?
                 (?:
-                    (?:(?P<days>[0-9]+)\s*(?:[:d]|days?)\s*)?
-                    (?P<hours>[0-9]+)\s*(?:[:h]|hours?)\s*
+                    (?P<days>[0-9]+)\s*d(?:ays?)?\s*
                 )?
-                (?P<mins>[0-9]+)\s*(?:[:m]|mins?|minutes?)\s*
-            )?
-            (?P<secs>[0-9]+)(?P<ms>\.[0-9]+)?\s*(?:s|secs?|seconds?)?
-        )$''', s)
-    if not m:
-        return None
-    res = 0
-    if m.group('only_mins'):
-        return float_or_none(m.group('only_mins'), invscale=60)
-    if m.group('only_hours'):
-        return float_or_none(m.group('only_hours'), invscale=60 * 60)
-    if m.group('secs'):
-        res += int(m.group('secs'))
-    if m.group('mins_reversed'):
-        res += int(m.group('mins_reversed')) * 60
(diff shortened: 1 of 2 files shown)


youtube_dl/utils.py
```diff
@@ -1540,44 +1540,35 @@ def parse_duration(s):
...
-        res += int(m.group('days')) * 24 * 60 * 60
-    if m.group('ms'):
-        res += float(m.group('ms'))
-    return res
+    days, hours, mins, secs, ms = None, None, None, None, None
+    m = re.match(r'(?:(?:(?:(?P<days>[0-9]+):)?(?P<hours>[0-9]+):)?(?P<mins>[0-9]+):)?(?P<secs>[0-9]+)(?P<ms>\.[0-9]+)?$', s)
+    if m:
+        days, hours, mins, secs, ms = m.groups()
+    else:
+        m = re.match(r'(?i)(?:P?T)?(?:(?P<days>[0-9]+)\s*d(?:ays?)?\s*)?(?:(?P<hours>[0-9]+)\s*h(?:ours?)?\s*)?(?:(?P<mins>[0-9]+)\s*m(?:in(?:ute)?s?)?\s*)?(?:(?P<secs>[0-9]+)(?P<ms>\.[0-9]+)?\s*s(?:ec(?:ond)?s?)?\s*)?$', s)
```
dstftw: This regex should be split into multiple lines for bettercode perception.

youtube_dl/utils.py
```diff
@@ -1540,44 +1540,35 @@ def parse_duration(s):
...
+    if mins:
+        duration += float(mins) * 60
+    if hours:
+        duration += float(hours) * 60 * 60
+    if days:
+        duration += float(days) * 24 * 60 * 60
+    if ms:
+        duration += float(ms)
+    if duration.is_integer():
+        duration = int(duration)
```
dstftw: I guess it's ok to always return stable float here.

youtube_dl/utils.py
remitamine: like that?
```python
        m = re.match(r'''(?ix)(?:P?T)?
            (?:(?P<days>[0-9]+)\s*d(?:ays?)?\s*)?
            (?:(?P<hours>[0-9]+)\s*h(?:ours?)?\s*)?
            (?:(?P<mins>[0-9]+)\s*m(?:in(?:ute)?s?)?\s*)?
            (?:(?P<secs>[0-9]+)(?P<ms>\.[0-9]+)?\s*s(?:ec(?:ond)?s?)?\s*)?$''', s)
```

dstftw: I would prefer indent similar to the former code.

youtube_dl/utils.py
```diff
@@ -1540,44 +1540,35 @@ def parse_duration(s):
...
-    if m.group('hours'):
-        res += int(m.group('hours')) * 60 * 60
-    if m.group('hours_reversed'):
-        res += int(m.group('hours_reversed')) * 60 * 60
-    if m.group('days'):
-        res += int(m.group('days')) * 24 * 60 * 60
-    if m.group('ms'):
-        res += float(m.group('ms'))
-    return res
+    days, hours, mins, secs, ms = None, None, None, None, None
```
dstftw: This can be simplified to `[None] * 5` or `(None, ) * 5`.

youtube_dl/utils.py
remitamine: i don't how exactly you're expect it to look like(it not the same regex).
```
        m = re.match(
            r'''(?i)(?:P?T)?
                (?:
                    (?P<days>[0-9]+)\s*d(?:ays?)?\s*
                )?
                (?:
                    (?P<hours>[0-9]+)\s*h(?:ours?)?\s*
                )?
                (?:
                    (?P<mins>[0-9]+)\s*m(?:in(?:ute)?s?)?\s*
                )?
                (?:
                    (?P<secs>[0-9]+)(?P<ms>\.[0-9]+)?\s*s(?:ec(?:ond)?s?)?\s*
                )?$''', s)
```
or this:
```
        m = re.match(
            r'''(?i)(?:P?T)?
                (?:
                    (?P<days>[0-9]+)\s*d
                    (?:ays?)?\s*
                )?
                (?:
                    (?P<hours>[0-9]+)\s*h
                    (?:ours?)?\s*
                )?
                (?:
                    (?P<mins>[0-9]+)\s*m
                    (?:in(?:ute)?s?)?\s*
                )?
                (?:
                    (?P<secs>[0-9]+)
                    (?P<ms>\.[0-9]+)?\s*s
                    (?:ec(?:ond)?s?)?\s*
                )?$''', s)
```
or something else.

dstftw: 1 is ok.

remitamine:
all done.

dstftw:
Thanks.

Issue #8467, opened 2016-02-08
[SITE-SUPPORT Request] for www.streamable.com


abdhan:
First of all thanks to everyone for this software.

I tried to make an extractor myself but could not figure out.

Example url: https://streamable.com/dnd1

They support downloads in two qualities: 

Low : https://cdn.streamable.com/video/mp4-mobile/dnd1.mp4

High: https://cdn.streamable.com/video/mp4/dnd1.mp4

The default extractor seems to work for this particular link but it gets only the High Quality version.

title is in `<h1> TITLE </h1>`

Any help is appreciated.

yan12125:
Already supported in the latest version.

PR #9122, opened 2016-04-08
[streamable] Add new extractor

remixz:
This fixes GH-8467.

The generic extractor does work for Streamable, but it wouldn't extract the video title, plus only exposed one video quality. This uses the Streamable API to grab all of the available formats, plus a bunch of other metadata.

### DIFF
--- a/youtube_dl/extractor/streamable.py
+++ b/youtube_dl/extractor/streamable.py
@@ -1,0 +1,76 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    float_or_none
+)
+
+
+class StreamableIE(InfoExtractor):
+    _VALID_URL = r'https?://streamable\.com/(?P<id>[\w]+)'
+    _TESTS = [
+        {
+            'url': 'https://streamable.com/dnd1',
+            'md5': '3e3bc5ca088b48c2d436529b64397fef',
+            'info_dict': {
+                'id': 'dnd1',
+                'ext': 'mp4',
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/streamable.py
```diff
@@ -0,0 +1,76 @@
...
+                'height': info.get('height'),
+                'filesize': info.get('size'),
+                'fps': info.get('framerate'),
+                'vbr': float_or_none(info.get('bitrate'), 1000)
+            })
+        self._sort_formats(formats)
+
+        return {
+            'id': video_id,
+            'title': video.get('result_title'),
```
dstftw: Title is mandatory.

youtube_dl/extractor/streamable.py
```diff
@@ -0,0 +1,76 @@
...
+
+        formats = []
+        for key, info in video.get('files').items():
+            formats.append({
+                'format_id': key,
+                'url': info['url'],
+                'width': info.get('width'),
+                'height': info.get('height'),
+                'filesize': info.get('size'),
+                'fps': info.get('framerate'),
```
dstftw: All numeric metafields should be wrapped in `int_or_none`.

youtube_dl/extractor/streamable.py
```diff
@@ -0,0 +1,76 @@
...
+        if status != 2:
+            raise ExtractorError(
+                'This video is currently unavailable. It may still be uploading or processing.',
+                expected=True)
+
+        formats = []
+        for key, info in video.get('files').items():
+            formats.append({
+                'format_id': key,
+                'url': info['url'],
```
dstftw: According to `status` in format dict particular files may also fail. Extraction should be tolerate to such files and missing `url`.

yan12125:

PR #9195, opened 2016-04-14
[downloader/external] enable piping for FFmpegFD(closes #2124)

### LINKED ISSUE #2124: Stream to vlc
Streaming video to vlc.

remitamine:
proposed the solution 14 days ago in https://github.com/rg3/youtube-dl/issues/2124#issuecomment-204072217.
as i understand from the issue the problem was that HLS live or AES encrypted streams can't be streamed to external software this should fix this problem.

### DIFF
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -235,10 +235,7 @@ class FFmpegFD(ExternalFD):
             args += ['-f', EXT_TO_OUT_FORMATS.get(info_dict['ext'], info_dict['ext'])]
 
         args = [encodeArgument(opt) for opt in args]
-        if tmpfilename == '-':
-            args.append('pipe:1')
-        else:
-            args.append(encodeFilename(ffpp._ffmpeg_filename_argument(tmpfilename), True))
+        args.append(encodeFilename(ffpp._ffmpeg_filename_argument(tmpfilename), True))
 
         self._debug_cmd(args)
 
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -225,7 +225,7 @@ class FFmpegFD(ExternalFD):
 
         args += ['-i', url, '-c', 'copy']
         if protocol == 'm3u8':
-            if self.params.get('hls_use_mpegts', False):
+            if self.params.get('hls_use_mpegts', False) or tmpfilename == '-':
                 args += ['-f', 'mpegts']
             else:
                 args += ['-f', 'mp4', '-bsf:a', 'aac_adtstoasc']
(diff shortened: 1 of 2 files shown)


yan12125:
Is this approach working on non-Unix systems (e.g. Windows)? If not, it would be better to add some documents.

remitamine:
i tested only on linux. i don't have a Windows OS to test with.

youtube_dl/downloader/external.py
```diff
@@ -235,7 +235,10 @@ def _call_downloader(self, tmpfilename, info_dict):
             args += ['-f', EXT_TO_OUT_FORMATS.get(info_dict['ext'], info_dict['ext'])]
 
         args = [encodeArgument(opt) for opt in args]
-        args.append(encodeFilename(ffpp._ffmpeg_filename_argument(tmpfilename), True))
+        if tmpfilename == '-':
+            args.append('pipe:1')
```
dstftw: I guess this should better go right to `_ffmpeg_filename_argument`.

dstftw:
It does work on Windows.

remitamine:
> I guess this should better go right to _ffmpeg_filename_argument. It does also work with just -.

done in d868f43c588b9929208823e86c12669d69e4e2cd.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -175,7 +175,7 @@ def _ffmpeg_filename_argument(self, fn):
         # Always use 'file:' because the filename may contain ':' (ffmpeg
         # interprets that as a protocol) or can start with '-' (-- is broken in
         # ffmpeg, see https://ffmpeg.org/trac/ffmpeg/ticket/2127 for details)
-        return 'file:' + fn
+        return 'file:' + fn if fn != '-' else fn
```
dstftw: Short rationale comment on not touching `-` here won't hurt.

remitamine: adding `- used to pipe the media stream to the standard output` to the comment.
is this good as an explanation?

dstftw: Something like: `Also leave '-' intact in order not to break streaming to stdout`.

remitamine: add the rationale comment in b9f2fdd37fe2085deb09710a2084c940e9920304.

dstftw:
LGTM.

bertho-zero:
It works with a conversion on the fly?

yan12125:
Currently no.

bertho-zero:
And now ? 6 month later

PR #9204, opened 2016-04-15
[la7] Update to fix the extractor to reflect the current reality

iamleot:
La7.it no longer uses the richplayer and the parsing logic is now
completely changed. Completely rework it to fix that.

XXX: Unfortunately I was able to test it only using a connection from Italy.
XXX: I am not sure if it is georestricted or not.

### DIFF
--- a/youtube_dl/extractor/la7.py
+++ b/youtube_dl/extractor/la7.py
@@ -1,60 +1,37 @@
+# encoding: utf-8
 from __future__ import unicode_literals
 
 from .common import InfoExtractor
-from ..utils import (
-    parse_duration,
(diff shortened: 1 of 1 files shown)


yan12125:
la7.it has changed again and I have fixed it in 712b0b5b7016202885a51c2a8af79921bf0fce2b.

iamleot:
Hello Yen!

Yen Chi Hsuan writes:
> la7.it has changed again and I have fixed it in 712b0b5b7016202885a51c2a8af79921bf0fce2b.
Thank you very much for updating and fixing it!

Issue #9157, opened 2016-04-11
XML error on www.rai.it download


pkirk:
Both on my OSX and on Linux Debian 7, the same error.

youtube-dl http://www.report.rai.it/dl/Report/puntata/ContentItem-2af9e702-92c6-4612-8139-33aed9979ef8.html
[Rai] 2af9e702-92c6-4612-8139-33aed9979ef8: Downloading webpage
[RaiTV] 78a62133-1ae6-46ba-be73-b7e7a3c6dea3: Downloading video JSON
[RaiTV] 78a62133-1ae6-46ba-be73-b7e7a3c6dea3: Downloading XML
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/usr/local/bin/youtube-dl/__main__.py", line 19, in <module>
  File "/usr/local/bin/youtube-dl/youtube_dl/__init__.py", line 419, in main
  File "/usr/local/bin/youtube-dl/youtube_dl/__init__.py", line 409, in _real_main
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 1725, in download
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 680, in extract_info
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 732, in process_ie_result
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 669, in extract_info
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 323, in extract
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/rai.py", line 111, in _real_extract
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 507, in _download_xml
  File "/usr/local/bin/youtube-dl/youtube_dl/compat.py", line 279, in compat_etree_fromstring
  File "/usr/local/bin/youtube-dl/youtube_dl/compat.py", line 269, in _XML
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1655, in close
    self._raiseerror(v)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1507, in _raiseerror
    raise err
[...]

PR #9232, opened 2016-04-17
fix issue #9157

LeoLamCY:


yan12125:
Rai has changed again and the latest format comes with output=45. I've adapted your changes and pushed to them master.

PR #9262, opened 2016-04-21
Add pornstars to pornhub extractor

burnersauce:
Adds 'pornstars' as a dict item returned by _real_extract in pornhub.py.
Returns a list of unicode strings which are the names of credited pornstars.

### DIFF
--- a/youtube_dl/extractor/pornhub.py
+++ b/youtube_dl/extractor/pornhub.py
@@ -133,6 +133,8 @@ class PornHubIE(InfoExtractor):
                 'height': height,
             })
         self._sort_formats(formats)
+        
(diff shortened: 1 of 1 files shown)


dstftw:
1. Just random metafield can't be added to info dict it should be [introduced](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L149-L251) first.
2. All introduced fields should be generic enough to be used with any kind of video. `pornstars` can only be used for porn videos thus not enough generic. `cast` or something sounds more generic.

burnersauce:
A fair criticism, what is the process to request a new field? Cast would seem an appropriate suggestion. Might this be considered an official request for the cast field or would a pull request with the field name changed work? Thank you.

dstftw:
For accepting a new field a PR that explains the field [here](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L75-L257) should be opened for discussion and opinions.

PR #9265, opened 2016-04-21
[francetv] Default m3u8 downloader

Maxime-J:
Back to the default downloader ( ffmpeg / avconv )
The native HLS downloader should only be used with --hls-prefer-native

### DIFF
--- a/youtube_dl/extractor/francetv.py
+++ b/youtube_dl/extractor/francetv.py
@@ -64,7 +64,7 @@ class FranceTVBaseInfoExtractor(InfoExtractor):
                         video_id, f4m_id=format_id, fatal=False))
             elif ext == 'm3u8':
                 formats.extend(self._extract_m3u8_formats(
-                    video_url, video_id, 'mp4', entry_protocol='m3u8_native',
+                    video_url, video_id, 'mp4',
                     m3u8_id=format_id, fatal=False))
             elif video_url.startswith('rtmp'):
                 formats.append({


yan12125:
Anything wrong with the native downloader? Our current policy is using the native downloader unless in cases it's not working (encrypted videos, etc.)

Maxime-J:
Two reasons:
- It's not really coherent to use the native way by default because it is listed as experimental and there is an option to trigger it.
- There are a lot of file operations:
all fragments are downloaded, after that they are concatened and finally there is the aac fixup which creates a copy before deleting the original.

yan12125:
Good points. Personally I prefer using ffmpeg, too. How do you think? @phihag @dstftw @jaimeMF @remitamine

dstftw:
I'd vote againt this change since it requires ffmpeg installed and does not support resuming.

yan12125:
Then we should remove the experimental flag and add something like ```--hls-prefer-ffmpeg```.

dstftw:
Done in 870d525848d2535be23ea04fcd3fbf826d03cae1 and bf09af3acbafd37e8327b8bab118066bd4c23d31.

Maxime-J:
Ok thanks.
I close my request.

Maxime-J:
There is an issue,
when you force ffmpeg with --hls-prefer-ffmpeg, info_dict protocol still is m3u8_native and the correct args are not passed to ffmpeg in [downloader/external.py#L227](downloader/external.py#L227https://github.com/rg3/youtube-dl/blob/master/youtube_dl/downloader/external.py#L227)

Could you correct this?

PR #9285, opened 2016-04-23
Odatv

skacurt:
A new extractor for Turkish news website odatv.com

### DIFF
--- a/youtube_dl/extractor/odatv.py
+++ b/youtube_dl/extractor/odatv.py
@@ -1,0 +1,60 @@
+# coding: utf-8
+
+from __future__ import unicode_literals
+from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    remove_start
+)
+import re
+
+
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/odatv.py
```diff
@@ -0,0 +1,59 @@
...
+            'title': 'md5:69654805a16a16cf9ec9d055e079831c'
+        }
+    }, {
+        'url': 'http://odatv.com/mob_video.php?id=8E388',
+        'md5': 'dc61d052f205c9bf2da3545691485154',
+        'info_dict': {
+            'id': '8E388',
+            'ext': 'mp4',
+            'title': 'md5:69654805a16a16cf9ec9d055e079831c'
+        }
```
dstftw: Remove duplicating tests and tests that test the same scenario.

skacurt: it's done.

Issue #9063, opened 2016-04-02
FFmpeg error while embedding subtitles in lynda.com videos


xiatianyu:
### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2016.04.01*.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.04.01**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

```
$ youtube-dl -v -o '%(playlist_index)s - %(title)s.%(ext)s' --write-sub --embed-subs --cookies cookies.txt 'https://www.lynda.com/JavaScript-tutorials/JavaScript-Essential-Training/81266-2.html'
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'-o', u'%(playlist_index)s - %(title)s.%(ext)s', u'--write-sub', u'--embed-subs', u'--cookies', u'cookies.txt', u'https://www.lynda.com/JavaScript-tutorials/JavaScript-Essential-Training/81266-2.html']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.04.01
[debug] Python version 2.7.9 - Linux-3.19.0-42-generic-x86_64-with-Ubuntu-15.04-vivid
[debug] exe versions: ffmpeg 2.5.10-0ubuntu0.15.04.1, ffprobe 2.5.10-0ubuntu0.15.04.1, rtmpdump 2.4
[debug] Proxy map: {}
[lynda:course] 81266: Downloading course JSON
[download] Downloading playlist: JavaScript Essential Training
[lynda:course] playlist JavaScript Essential Training: Collected 56 video ids (downloading 56 of them)
[...]

PR #9288, opened 2016-04-23
[ffmpeg] Fix embedding subtitles (#9063)

reyyed:
Change command line parameters for ffmpeg when embedding subtitles.
Change to ‘-map 0:v -c:v copy -map 0:a -c:a copy’

### DIFF
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -363,8 +363,10 @@ class FFmpegEmbedSubtitlePP(FFmpegPostProcessor):
         input_files = [filename] + sub_filenames
 
         opts = [
-            '-map', '0',
-            '-c', 'copy',
+            '-map', '0:v',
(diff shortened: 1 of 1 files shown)


yan12125:
Sorry for the late response. Any broken example before this fix?

reyyed:
Hi, sorry, this is to fix [#9063](https://github.com/rg3/youtube-dl/issues/9063). As discussed in that issue, this is caused by certain lynda videos having empty streams.
The changes to the parameters enables ffmpeg to pick the correct working streams in those lynda videos, including other videos that come along with empty streams.

yan12125:
From http://ffmpeg.org/ffmpeg.html:
```
stream_type is one of following: ’v’ or ’V’ for video, ’a’ for audio, ’s’ for subtitle, ’d’ for data, and ’t’ for attachments.
```
Honestly I don't know what are data and attachments in media files, and I haven't seen any example with them. I guess not copying them does not cause problems :)

PR #9295, opened 2016-04-24
[myspace] Added support for downloading all songs of an artist

AllanDaemon:
With this, you can download all music from a artist page songs, including the ones outside an album.

### DIFF
--- a/youtube_dl/extractor/myspace.py
+++ b/youtube_dl/extractor/myspace.py
@@ -206,8 +206,8 @@ class MySpaceAlbumIE(InfoExtractor):
         return self.playlist_result(entries, playlist_id, self._og_search_title(webpage))
 
 
-class MySpaceArtistSongsIE(InfoExtractor):
-    IE_NAME = 'MySpace:artist:songs'
+class MySpaceArtistIE(InfoExtractor):
+    IE_NAME = 'MySpace:artist'
     _VALID_URL = r'https?://myspace\.com/(?P<artist>[^/]*)/music/songs'
 
     _TEST = {
--- a/youtube_dl/extractor/myspace.py
+++ b/youtube_dl/extractor/myspace.py
@@ -210,3 +210,46 @@ class MySpaceAlbumIE(InfoExtractor):
             'title': self._og_search_title(webpage),
             'entries': entries,
         }
+
+
+class MySpaceArtistSongsIE(InfoExtractor):
+    IE_NAME = 'MySpace:artist:songs'
+    _VALID_URL = r'https?://myspace\.com/(?P<artist>[^/]*)/music/songs'
+
+    _TESTS = [{
+        'url': 'https://myspace.com/starset2/music/album/transmissions-19455773',
+        'url': 'https://myspace.com/ryanlewismusic/music/songs',
+        'info_dict': {
+            'title': 'RYAN LEWIS',
+        },
+        'playlist_count': 8,
+    }, {
+        'url': 'https://myspace.com/studio99/music/songs/',
+        'info_dict': {
+            'title': 'Studio 99 IS CLOSED!! R.I.P.',
+        },
+        'playlist_count': 4,
+    }]
+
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/myspace.py
```diff
@@ -204,3 +204,46 @@ def _real_extract(self, url):
...
             'entries': entries,
         }
+
+
+class MySpaceArtistSongsIE(InfoExtractor):
+    IE_NAME = 'MySpace:artist:songs'
+    _VALID_URL = r'https?://myspace\.com/(?P<artist>[^/]*)/music/songs'
+
+    _TESTS = [{
+        'url': 'https://myspace.com/starset2/music/album/transmissions-19455773',
```
dstftw: This should be removed.

youtube_dl/extractor/myspace.py
```diff
@@ -204,3 +204,46 @@ def _real_extract(self, url):
...
+            'title': 'RYAN LEWIS',
+        },
+        'playlist_count': 8,
+    }, {
+        'url': 'https://myspace.com/studio99/music/songs/',
+        'info_dict': {
+            'title': 'Studio 99 IS CLOSED!! R.I.P.',
+        },
+        'playlist_count': 4,
+    }]
```
dstftw: If both tests do test the same scenario leave only one.

youtube_dl/extractor/myspace.py
```diff
@@ -204,3 +204,46 @@ def _real_extract(self, url):
...
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        display_id = mobj.group('artist')
+        webpage = self._download_webpage(url, display_id)
+        tracks_paths = re.findall(r'<meta itemprop="url" content ="(.*?)">', webpage)
+        if not tracks_paths:
+            raise ExtractorError(
+                '%s: No songs found, try using proxy' % display_id,
```
dstftw: How are missing songs related to using proxy? https://myspace.com/mekowilliamssho/music/songs have no songs but suggests using proxy. This check should be removed completely. youtube-dl will handle this on it's own.

youtube_dl/extractor/myspace.py
```diff
@@ -204,3 +204,46 @@ def _real_extract(self, url):
...
+            for t_path in tracks_paths if t_path.startswith('/music/song/')]
+        #   if we invert this if, we get album urls. but not all music are in alba.
+        #   Also the musics in alba are already here individually
+
+        return {
+            '_type': 'playlist',
+            'display_id': display_id,
+            'title': self._og_search_title(webpage),
+            'entries': entries,
+        }
```
dstftw: Use `self.playlist_result` instead.

youtube_dl/extractor/myspace.py
AllanDaemon: Well. This IE is copy-based from the Album one (MySpaceAlbumIE). I just copied and adapted the code artist instead album. I guess the original author put this due geoblocking giving different results. If change this part, it's better to change the other too, to keep coherent.

Commits to this part:
95c673a148f1a6cd8571b4dfa931d98092500fc2

AllanDaemon:
Hi, is something else needed in order to accept the pull request?

JohnKacz:
Checking on this?

AllanDaemon:
I closed accidentally.

AllanDaemon:
Is there anything else needed to merge this request?

AllanDaemon:
bump

AllanDaemon:
BUMP

AllanDaemon:
**bump**

llacb47:
This will never be merged, look into contributing to a fork. I can't link you any suggestions or my comment will get deleted.

AllanDaemon:
_bump_

AllanDaemon:
BuMp

dstftw:
I've already pointed out what should be fixed to get this accepted.

PR #8820, opened 2016-03-10
[extractor/common] improve m3u8 info extraction

remitamine:
parts from the documention that explain some of the changes.
> If an EXT-X-STREAM-INF tag contains a RESOLUTION attribute and a
VIDEO attribute, then every alternative video Rendition MUST have
an optimal display resolution matching the value of the RESOLUTION
attribute.

```python
                group = stream_info.get('VIDEO')
                if group:
                    groups[group] = r
...
                group = groups.get(media_info['GROUP-ID'])
                if group:
                    v.update(group)
```
> If the media TYPE is AUDIO and the URI attribute is
missing, clients MUST assume that the audio data for this Rendition
is present in every video Rendition specified by the EXT-X-STREAM-INF
tag.

```python
lang = media_info.get('LANGUAGE')
...
'language': lang,
```

PR #9324, opened 2016-04-26
[vlive] Add support for live videos

### LINKED ISSUE #9321: [vlive] Unable to download live videos
### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2016.04.24*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.04.24**

- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

- [x] Site support request (request for adding support for a new site)
- [x] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

(Note that this live might end when you check it.)

```
$ youtube-dl -v 'http://www.vlive.tv/video/7779/퇴근-저흰출근방송'
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['-v', 'http://www.vlive.tv/video/7779/퇴근-저흰출근방송']
[debug] youtube-dl version 2016.04.24
[debug] Python version 3.4.3 - Linux-4.5.0-gentoo-x86_64-Intel-R-_Core-TM-_i7-3820_CPU_@_3.60GHz-with-gentoo-2.2
[debug] exe versions: ffmpeg N-79272-gbd1dcaf, ffprobe N-79272-gbd1dcaf, rtmpdump 2.4
[debug] Proxy map: {}
[vlive] 7779: Downloading webpage
ERROR: Unable to extract long video id; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
  File "/home/kagami/.local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 673, in extract_info
    ie_result = ie.extract(url)
  File "/home/kagami/.local/bin/youtube-dl/youtube_dl/extractor/common.py", line 341, in extract
  File "/home/kagami/.local/bin/youtube-dl/youtube_dl/extractor/vlive.py", line 36, in _real_extract
    webpage, 'long video id')
  File "/home/kagami/.local/bin/youtube-dl/youtube_dl/extractor/common.py", line 644, in _search_regex
youtube_dl.utils.RegexNotFoundError: Unable to extract long video id; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

---

### If the purpose of this *issue* is a *site support request* please provide all kinds of example URLs support for which should be included (replace following example URLs by **yours**):
- Live video: has the exact same type of URLs, just marked as live videos at the site; check fresh lives [here](http://www.vlive.tv/home)

---

### Description of your *issue*, suggested solution and other information

(Not sure which category is best suited for this issue, sorry.)

Lives are served with Flash through HLS streams, there is JSON with streams parameters at the code page:

```
vlive.tv.video.ajax.request.handler.init("7779", "{\"liveStatus\":\"LIVE\",\"resolutions\":[
{\"name\":\"audio\",\"width\":null,\"height\":null,\"cdnUrl\":\"http:\/\/vlive.hls.edgesuite.net\/kr002\/alow.stream\/playlist.m3u8?__gda__=1461682617_f30c9b77b4d56a116b0e436d36b7b1d1\",\"additionalProperties\":{}},
{\"name\":\"250\",\"width\":320,\"height\":184,\"cdnUrl\":\"http:\/\/vlive.hls.edgesuite.net\/kr002\/250.stream\/playlist.m3u8?__gda__=1461682617_5bd3a23663a5cd54dc204da558ff00e5\",\"additionalProperties\":{}},
{\"name\":\"400\",\"width\":640,\"height\":368,\"cdnUrl\":\"http:\/\/vlive.hls.edgesuite.net\/kr002\/400.stream\/playlist.m3u8?__gda__=1461682617_62f2ced6228651d617e7a7b9fff0da6a\",\"additionalProperties\":{}},
{\"name\":\"1200\",\"width\":640,\"height\":368,\"cdnUrl\":\"http:\/\/vlive.hls.edgesuite.net\/kr002\/1200.stream\/playlist.m3u8?__gda__=1461682617_88ce3b5362dc63e50439be087164e98b\",\"additionalProperties\":{}}
],\"additionalProperties\":{}}", "", "", "", "");
```

Kagami:
Fixes #9321 

This is WIP, I don't know how to handle HLS subtitles yet. But otherwise everything seems to be working.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -1061,7 +1061,7 @@ class InfoExtractor(object):
     def _extract_m3u8_formats(self, m3u8_url, video_id, ext=None,
                               entry_protocol='m3u8', preference=None,
                               m3u8_id=None, note=None, errnote=None,
-                              fatal=True):
+                              fatal=True, live=False):
 
         formats = [{
             'format_id': '-'.join(filter(None, [m3u8_id, 'meta'])),
@@ -1139,7 +1139,11 @@ class InfoExtractor(object):
                 if m3u8_id:
                     format_id.append(m3u8_id)
                 last_media_name = last_media.get('NAME') if last_media and last_media.get('TYPE') != 'SUBTITLES' else None
-                format_id.append(last_media_name if last_media_name else '%d' % (tbr if tbr else len(formats)))
+                # Bandwidth of live streams may differ over time thus making
+                # format_id unpredictable. So it's better to keep provided
+                # format_id intact.
+                if last_media_name and not live:
+                    format_id.append(last_media_name if last_media_name else '%d' % (tbr if tbr else len(formats)))
                 f = {
                     'format_id': '-'.join(format_id),
                     'url': format_url(line.strip()),
--- a/youtube_dl/extractor/vlive.py
+++ b/youtube_dl/extractor/vlive.py
@@ -1,8 +1,11 @@
 # coding: utf-8
-from __future__ import unicode_literals
+from __future__ import division, unicode_literals
 
+import re
+import time
 from .common import InfoExtractor
 from ..utils import (
+    ExtractorError,
     dict_get,
     float_or_none,
     int_or_none,
@@ -31,16 +34,77 @@ class VLiveIE(InfoExtractor):
         webpage = self._download_webpage(
             'http://www.vlive.tv/video/%s' % video_id, video_id)
 
-        long_video_id = self._search_regex(
-            r'vlive\.tv\.video\.ajax\.request\.handler\.init\(\s*"[0-9]+"\s*,\s*"[^"]*"\s*,\s*"([^"]+)"',
-            webpage, 'long video id')
+        # UTC+x - UTC+9 (KST)
+        tz = time.altzone if time.localtime().tm_isdst == 1 else time.timezone
+        tz_offset = -tz // 60 - 9 * 60
+        self._set_cookie('vlive.tv', 'timezoneOffset', '%d' % tz_offset)
 
-        key = self._search_regex(
-            r'vlive\.tv\.video\.ajax\.request\.handler\.init\(\s*"[0-9]+"\s*,\s*"[^"]*"\s*,\s*"[^"]+"\s*,\s*"([^"]+)"',
-            webpage, 'key')
+        status_params = self._download_json(
+            'http://www.vlive.tv/video/status?videoSeq=%s' % video_id,
+            video_id, 'Downloading JSON status',
+            headers={'Referer': url})
+        status = status_params.get('status')
+        air_start = status_params.get('onAirStartAt', '')
+        is_live = status_params.get('isLive')
 
+        video_params = self._search_regex(
+            r'vlive\.tv\.video\.ajax\.request\.handler\.init\((.+)\)',
+            webpage, 'video params')
+        live_params, long_video_id, key = re.split(
+            r'"\s*,\s*"', video_params)[1:4]
+
+        if status == 'LIVE_ON_AIR' or status == 'BIG_EVENT_ON_AIR':
+            live_params = self._parse_json('"%s"' % live_params, video_id)
+            live_params = self._parse_json(live_params, video_id)
+            return self._live(video_id, webpage, live_params)
+        elif status == 'VOD_ON_AIR' or status == 'BIG_EVENT_INTRO':
+            if long_video_id and key:
+                return self._replay(video_id, webpage, long_video_id, key)
+            elif is_live:
+                status = 'LIVE_END'
+            else:
+                status = 'COMING_SOON'
+
+        if status == 'LIVE_END':
+            raise ExtractorError('Uploading for replay. Please wait...',
+                                 expected=True)
+        elif status == 'COMING_SOON':
+            raise ExtractorError('Coming soon! %s' % air_start, expected=True)
+        elif status == 'CANCELED':
+            raise ExtractorError('We are sorry, '
+                                 'but the live broadcast has been canceled.',
+                                 expected=True)
+        else:
+            raise ExtractorError('Unknown status %s' % status)
+
+    def _get_common_fields(self, webpage):
         title = self._og_search_title(webpage)
+        creator = self._html_search_regex(
+            r'<div[^>]+class="info_area"[^>]*>\s*<a\s+[^>]*>([^<]+)',
+            webpage, 'creator', fatal=False)
+        thumbnail = self._og_search_thumbnail(webpage)
+        return {
+            'title': title,
+            'creator': creator,
+            'thumbnail': thumbnail,
+        }
 
+    def _live(self, video_id, webpage, live_params):
+        formats = []
+        for vid in live_params.get('resolutions', []):
+            formats.extend(self._extract_m3u8_formats(
+                vid['cdnUrl'], video_id, 'mp4',
+                m3u8_id=vid.get('name'),
+                fatal=False, live=True))
+        self._sort_formats(formats)
+
+        return dict(self._get_common_fields(webpage),
+            id=video_id,
+            formats=formats,
+            is_live=True,
+        )
+
+    def _replay(self, video_id, webpage, long_video_id, key):
         playinfo = self._download_json(
             'http://global.apis.naver.com/rmcnmv/rmcnmv/vod_play_videoInfo.json?%s'
             % compat_urllib_parse_urlencode({
@@ -62,11 +126,6 @@ class VLiveIE(InfoExtractor):
         } for vid in playinfo.get('videos', {}).get('list', []) if vid.get('source')]
         self._sort_formats(formats)
 
-        thumbnail = self._og_search_thumbnail(webpage)
-        creator = self._html_search_regex(
-            r'<div[^>]+class="info_area"[^>]*>\s*<a\s+[^>]*>([^<]+)',
-            webpage, 'creator', fatal=False)
-
         view_count = int_or_none(playinfo.get('meta', {}).get('count'))
 
         subtitles = {}
@@ -77,12 +136,9 @@ class VLiveIE(InfoExtractor):
                     'ext': 'vtt',
                     'url': caption['source']}]
 
-        return {
-            'id': video_id,
-            'title': title,
-            'creator': creator,
-            'thumbnail': thumbnail,
-            'view_count': view_count,
-            'formats': formats,
-            'subtitles': subtitles,
-        }
+        return dict(self._get_common_fields(webpage),
+            id=video_id,
+            formats=formats,
+            view_count=view_count,
+            subtitles=subtitles,
+        )


Kagami:
Add correct sub URLs, though both ffmpeg with `-i url.m3u8 out.vtt` and youtube-dl with `--write-sub` can't currently download them. Any ideas?

Kagami:
Get it, ffmpeg actually downloads HLS WebVTT fine, just in a bit strange way. Though `youtube-dl --write-sub` currently can't handle such streams. So, in order to support it properly, we need to:

* Indicate HLS/stream in subtitle entry, e.g. `{'ext': 'vtt', 'url': 'http://', 'hls': True}`
* Add support for dumping HLS to `--write-sub` routine
* Add option to not download video in case of `--write-sub` because it doesn't make sense in our case: subs writing will work until stream is ended, so user will need to run youtube-dl with different options twice; or maybe spawn `--write-sub` in separate process?

What do you think?

yan12125:
FYI: Seems #6144 implements downloading subtitles from m3u8 playlists.

Kagami:
Thanks, didn't know about it. Though it seems to be stalled? If that would be easier for you, I can comment subtitle field for now and make separate PR. Grabbing live video streams is useful even without subs.

yan12125:
In view-source:http://static.vlive.tv/desktop/resources/generated/js/page/2016033117/video.min.js:
```Javascript
            case "VOD_ON_AIR":
                if (a.vodId && a.vodInKey) {
                    this._vodOnAir(a)
                } else {
                    if (a.isLive) {
                        this._liveEnd()
                    } else {
                        this._comingSoon()
                    }
                }
                break
```
Maybe youtube-dl should follow the same logic?

youtube_dl/extractor/vlive.py
```diff
@@ -31,16 +34,73 @@ def _real_extract(self, url):
...
+        self._sort_formats(formats)
+
+        subtitles = {
+            'eng': [{'ext': 'vtt', 'url': self._get_sub_url(f['url'], 'eng')}
+                    for f in formats],
+            'zcn': [{'ext': 'vtt', 'url': self._get_sub_url(f['url'], 'zcn')}
+                    for f in formats],
+            'vie': [{'ext': 'vtt', 'url': self._get_sub_url(f['url'], 'vie')}
+                    for f in formats],
+        }
```
yan12125: Not all live videos have subtitles. For example http://www.vlive.tv/video/7871/1996-%EC%95%88%EC%8A%B9%EC%A4%80-%EC%9C%A4%EC%A0%95%EC%9E%AC-%EB%9D%BC%EC%9D%B4%EB%B8%8C does not have a subtitle when I just checked. By the way, I guess ```_get_live_sub_url``` is a better function name.

youtube_dl/extractor/vlive.py
```diff
@@ -31,16 +34,73 @@ def _real_extract(self, url):
...
-        key = self._search_regex(
-            r'vlive\.tv\.video\.ajax\.request\.handler\.init\(\s*"[0-9]+"\s*,\s*"[^"]*"\s*,\s*"[^"]+"\s*,\s*"([^"]+)"',
-            webpage, 'key')
+        video_params = self._search_regex(
+            r'vlive\.tv\.video\.ajax\.request\.handler\.init\((.*)\)',
+            webpage, 'video params')
+        live, long_video_id, key = re.split(r'"\s*,\s*"', video_params)[1:4]
 
+        if status == 'LIVE_ON_AIR' or status == 'BIG_EVENT_ON_AIR':
+            live_params = json.loads(json.loads('"%s"' % live))
```
yan12125: Use ```self._parse_json```.

yan12125:
For your previous question on how to download live subtitles, there are two options:

* Implemented it in pure Python [HlsFD](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/downloader/hls.py)
* Create a thread for downloading subtitles.

There's an issue on ihe second approach. We need to add an option to disable subtitle downloading in ffmpeg to prevent duplicated traffic.
If you already have an implementation, please don't include it in this PR. Open a new one - it's not a good idea to add too many features in a single PR.

youtube_dl/extractor/vlive.py
Kagami: I haven't found anything similar in Live API, seems like the only way to determine subtitle status is to download `playlist.m3u8` and parse it manually. It looks like this:

```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-MEDIA:TYPE=SUBTITLES,GROUP-ID="subs",NAME="Vietnamese",FORCED=NO,AUTOSELECT=YES,URI="subtitlelist_lvie.m3u8",LANGUAGE="vie"
#EXT-X-MEDIA:TYPE=SUBTITLES,GROUP-ID="subs",NAME="English",FORCED=NO,AUTOSELECT=YES,URI="subtitlelist_leng.m3u8",LANGUAGE="eng"
#EXT-X-MEDIA:TYPE=SUBTITLES,GROUP-ID="subs",NAME="zcn",FORCED=NO,AUTOSELECT=YES,URI="subtitlelist_lzcn.m3u8",LANGUAGE="zcn"
#EXT-X-STREAM-INF:BANDWIDTH=1464974,CODECS="avc1.66.22,mp4a.40.2",RESOLUTION=640x368,SUBTITLES="subs"
chunklist.m3u8?__agda__=1461853927_d2d91835e251112f931019235adf7e99
```

Are you ok with that?

Kagami:
>Implemented it in pure Python HlsFD
>Create a thread for downloading subtitles

If we won't spawn another thread for `--write-sub` then we won't be able to download video, so I think we need to spawn thread anyway?

>We need to add an option to disable subtitle downloading in ffmpeg to prevent duplicated traffic

What do you mean by duplicated? ffmpeg only follows `chunklist.m3u8` stream which doesn't contain subtitles as far as I can tell.

>Open a new one

Do you want me to leave subtitle URLs handling in this PR? `--write-sub` won't work, though someone might use those URLs with `youtube-dl --dump-json`.

Also, I'm not sure regarding subtitles streams for various qualities. Flash player will use `1200.stream/subtitlelist_l<lang>.m3u8` with `1200.stream/playlist.m3u8` and `400.stream/subtitlelist_l<lang>.m3u8` with `400.stream/playlist.m3u8`, though `1200.stream/subtitlelist_l<lang>.m3u8` and `400.stream/subtitlelist_l<lang>.m3u8` seems to be the same. Do we need to provide them all or only one?

youtube_dl/extractor/vlive.py
yan12125: Parsing should be the way. It's better than hard-coded values in general. An approach is modifying [```_extract_m3u8_formats```](https://github.com/rg3/youtube-dl/blob/d671237/youtube_dl/extractor/common.py#L1061-L1179) so that it can handle subtitles, too.

yan12125:
Haven't looked into the spec of M3U8 manifests. If each video segment corresponds to exactly one subtitle segment, then the subtitle can be downloaded immediately after the video segment.

> What do you mean by duplicated?

I have seen "ffmpeg can handle subtitles" somewhere. I found it's not merged yet (https://trac.ffmpeg.org/ticket/2833), so the poster may be using a custom fork of ffmpeg.

> Do you want me to leave subtitle URLs handling in this PR?

Comment it out and add some comments.

youtube_dl/extractor/vlive.py
remitamine: >  An approach is modifying _extract_m3u8_formats so that it can handle subtitles, too.

i already proposed this in #8820 as a first step to handle WEBVTT m3u8 subtitles.

Kagami: It would be very cool if we can get it merged, don't want to reimplement already written code.

Kagami:
Btw, ffmpeg HLS dumper seems to be very slow, audio goes out of sync for high-res streams. I get much better results with `livestreamer "hlsvariant://..." -o out.ts`. We might want to implement `--hls-prefer-livestreamer`. Or improve `hls.py` though I don't know protocol well and don't have that much time so I would like to use external downloaders if you don't mind.

yan12125:
Could you try again by ```--hls-prefer-native``` to youtube-dl? By default youtube-dl uses ffmpeg to download HLS streams, and ```--hls-prefer-native``` select the embedded pure Python (incomplete) implementation of the HLS protocol.

Kagami:
I tried it, but native downloader seems to be not working on vlive streams. You could try it right away with `python -m youtube_dl http://www.vlive.tv/video/7880 --hls-prefer-native` on my branch. Output:

```
$ python -m youtube_dl 'http://www.vlive.tv/video/7880' --hls-prefer-native -v
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['http://www.vlive.tv/video/7880', '--hls-prefer-native', '-v']
[debug] youtube-dl version 2016.04.24
[debug] Git HEAD: 14b6e54
[debug] Python version 3.4.3 - Linux-4.5.0-gentoo-x86_64-Intel-R-_Core-TM-_i7-3820_CPU_@_3.60GHz-with-gentoo-2.2
[debug] exe versions: ffmpeg N-79663-ge639f50, ffprobe N-79663-ge639f50, rtmpdump 2.4
[debug] Proxy map: {}
[vlive] 7880: Downloading webpage
[vlive] 7880: Downloading JSON status
[debug] Invoking downloader on 'http://vlive.hls.edgesuite.net/hk002/400.stream/playlist.m3u8?__gda__=1461877422_547a10c5467c1b3aabe2b6d13b08be52'
[hlsnative] Total fragments: 1
[download] Destination: [V LIVE] MTV - Minh Hang on the stage-7880.mp4
[download] 100% of 211.00B in 00:00
[ffmpeg] Fixing malformated aac bitstream in "[V LIVE] MTV - Minh Hang on the stage-7880.mp4"
[debug] ffmpeg command line: ffmpeg -y -i 'file:[V LIVE] MTV - Minh Hang on the stage-7880.mp4' -c copy -f mp4 -bsf:a aac_adtstoasc 'file:[V LIVE] MTV - Minh Hang on the stage-7880.temp.mp4'
^C
ERROR: Interrupted by user
```

Kagami:
If I pass it `chunklist.m3u8` URL instead it downloads only 3 fragments (~5 seconds of video) and exits. Seems like it's not suited for live translations?

yan12125:
Yep ```HlsFD``` does not support live streams yet. A thread is necessary.

youtube_dl/extractor/vlive.py
```diff
@@ -31,16 +34,73 @@ def _real_extract(self, url):
...
+        }
+
+    def _get_sub_url(self, video_url, lang):
+        url = video_url.split('/playlist.m3u8')[0]
+        url += '/subtitlelist_l%s.m3u8' % lang
+        return url
+
+    def _live(self, video_id, webpage, live_params):
+        formats = [{
+            'url': vid['cdnUrl'],
```
yan12125: Should extract chunklists via ```self._extract_m3u8_formats``` here.

youtube_dl/extractor/vlive.py
```diff
@@ -31,16 +34,73 @@ def _real_extract(self, url):
...
+            webpage, 'video params')
+        live, long_video_id, key = re.split(r'"\s*,\s*"', video_params)[1:4]
 
+        if status == 'LIVE_ON_AIR' or status == 'BIG_EVENT_ON_AIR':
+            live_params = json.loads(json.loads('"%s"' % live))
+            return self._live(video_id, webpage, live_params)
+        elif status == 'VOD_ON_AIR' or status == 'BIG_EVENT_INTRO':
+            return self._replay(video_id, webpage, long_video_id, key)
+        elif status == 'LIVE_END':
+            raise ExtractorError('Uploading for replay. Please wait…',
```
yan12125: Use ... instead of … here. Non-ascii outputs may break random things.

Kagami:
A bit unrelated question. Is it possible to get subtitle URL without downloading it, similar to `--get-url` for video?

yan12125:
No direct options. You can dump the JSON dict with ```-j``` and everything is in it.

Kagami:
@yan12125 addressed all comments. Removed subtitles handling because it depends on #8820 (#6144 also has similar code).

youtube_dl/extractor/vlive.py
```diff
@@ -1,8 +1,12 @@
 # coding: utf-8
+from __future__ import division
```
yan12125: Merge them: ```from __future__ import division, unicode_literals```

youtube_dl/extractor/vlive.py
```diff
@@ -31,16 +35,74 @@ def _real_extract(self, url):
...
+            elif is_live:
+                status = 'LIVE_END'
+            else:
+                status = 'COMING_SOON'
+
+        if status == 'LIVE_END':
+            raise ExtractorError('Uploading for replay. Please wait...',
+                                 expected=True)
+        elif status == 'COMING_SOON':
+            raise ExtractorError('Coming soon! %s' % air_start, expected=True)
```
yan12125: If ```air_start``` is ```None```, the message should be ```Coming soon!``` instead of ```Coming soon! None```.

youtube_dl/extractor/vlive.py
```diff
@@ -31,16 +35,74 @@ def _real_extract(self, url):
...
+            headers={'Referer': url, 'Cookie': 'timezoneOffset=%d' % tz_offset})
+        status = status_params.get('status')
+        air_start = status_params.get('onAirStartAt')
+        is_live = status_params.get('isLive')
 
-        key = self._search_regex(
-            r'vlive\.tv\.video\.ajax\.request\.handler\.init\(\s*"[0-9]+"\s*,\s*"[^"]*"\s*,\s*"[^"]+"\s*,\s*"([^"]+)"',
-            webpage, 'key')
+        video_params = self._search_regex(
+            r'vlive\.tv\.video\.ajax\.request\.handler\.init\((.*)\)',
```
yan12125: Better to use ```([^;]+)``` instead of ```(.*)```. It reduces the probability of matching something unexpected.

youtube_dl/extractor/vlive.py
```diff
@@ -34,26 +35,44 @@ def _real_extract(self, url):
...
 
-        status = self._download_json(
+        # UTC+x - UTC+9 (KST)
+        tz = time.altzone if time.localtime().tm_isdst == 1 else time.timezone
+        tz_offset = -tz // 60 - 9 * 60
+        status_params = self._download_json(
             'http://www.vlive.tv/video/status?videoSeq=%s' % video_id,
             video_id, 'Downloading JSON status',
-            headers={'Referer': url}).get('status')
+            headers={'Referer': url, 'Cookie': 'timezoneOffset=%d' % tz_offset})
```
yan12125: Use ```self._add_cookies``` to add global cookies. It's better to use the same timezone offset for all requests to get consistent results.

youtube_dl/extractor/vlive.py
Kagami: Semicolon might appear in live params, it's not safe to match like that.

yan12125: OK Then ˙˙˙(.+)˙˙˙

Kagami:
@yan12125 addressed all comments. Tested on few lives, seems like it works as expected.

Kagami:
There is issue with default format ids generated by `_extract_m3u8_formats`. It appends bitrate after the dash but it constantly changes in case of vlive. I think we should delete everything after `-` because there is only single video stream for each m3u8.

yan12125:
You can override ```format_id```.

Kagami:
Extended `_extract_m3u8_formats`.

yan12125:
What's the meaning of ```media_name_fallback```? Not quite intuitive. Maybe ```live```, which defaults to ```False```, is better.

Kagami:
Fixed.

youtube_dl/extractor/common.py
```diff
@@ -1139,7 +1139,8 @@ def _extract_m3u8_formats(self, m3u8_url, video_id, ext=None,
                 if m3u8_id:
                     format_id.append(m3u8_id)
                 last_media_name = last_media.get('NAME') if last_media and last_media.get('TYPE') != 'SUBTITLES' else None
-                format_id.append(last_media_name if last_media_name else '%d' % (tbr if tbr else len(formats)))
+                if last_media_name or not live:
```
yan12125: Should be ```and```.

yan12125: Also add some comments for why not adding bitrates for live streams.

Kagami: >Should be and

We may result in identical format ids in that case. Though there might be several streams without `NAME` too. What about simple counter? E.g. `fid`, `fid-1`, `fid-2`.

yan12125: Such a feature is already implemented in ```YoutubeDL```: https://github.com/rg3/youtube-dl/blob/bf09af3/youtube_dl/YoutubeDL.py#L1318-L1322

remitamine: > Though there might be several streams without NAME too.

according to standard `NAME` attribute is REQUIRED.

Kagami: >Such a feature is already implemented

Ah, that's cool, fixed.

>according to the standard NAME attribute is REQUIRED

Not in case of vlive... (See example playlist above.)

yan12125:
Looks good. The final bit: add ```'is_live'``` field to the returned dict.

Kagami:
Fixed and squashed.

yan12125:
Merged. Thanks for the work and sorry for the rushes.

Kagami:
Thanks!

PR #9328, opened 2016-04-27
[mwave] Add Mwave Meet & Greet extractor

pmrowla:
Adds support for alternate URLs to the existing mwave.interest.me extractor


yan12125:
Thanks merged. Would you like to be listed in [AUTHORS](https://github.com/rg3/youtube-dl/blob/master/AUTHORS) under the name Peter Rowlands?

pmrowla:
@yan12125 that's not necessary, thanks though

PR #9341, opened 2016-04-28
[wrzuta.pl - playlist] wrzuta playlist extraction

venth:
I've extended wrzuta extractor by adding playlist extraction capability.

### DIFF
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -979,6 +979,7 @@ from .wimp import WimpIE
 from .wistia import WistiaIE
 from .worldstarhiphop import WorldStarHipHopIE
 from .wrzuta import WrzutaIE
+from .wrzuta import WrzutaPlaylistIE
 from .wsj import WSJIE
 from .xbef import XBefIE
 from .xboxclips import XboxClipsIE
--- a/youtube_dl/extractor/wrzuta.py
+++ b/youtube_dl/extractor/wrzuta.py
@@ -80,3 +80,77 @@ class WrzutaIE(InfoExtractor):
             'description': self._og_search_description(webpage),
             'age_limit': embedpage.get('minimalAge', 0),
         }
+
+
+_ENTRY_PATTERN = r'<a href="(?P<playlist_entry_url>[^"]+)" target="_blank" class="playlist\-file\-page">'
+_PLAYLIST_SIZE_PATTERN = r'<div class="playlist-counter">[0-9]+/([0-9]+)</div>'
+
+
+class WrzutaPlaylistIE(InfoExtractor):
+    """
+        this class covers extraction of wrzuta playlist entries
+        the extraction process bases on following steps:
+        * collect information of playlist size
+        * download all entries provided on
+          the playlist webpage (the playlist is split
+          on two pages: first directly reached from webpage
+          second: downloaded on demand by ajax call and rendered
+          using the ajax call response)
+        * in case size of extracted entries not reached total number of entries
+          use the ajax call to collect the remaining entries
+    """
+
+    IE_NAME = 'wrzuta.pl:playlist'
+
+    _VALID_URL = r'https?://(?P<uploader>[0-9a-zA-Z]+)\.wrzuta\.pl/playlista/' \
+                 '(?P<id>[0-9a-zA-Z]+)/.*'
+
+    _TESTS = [{
+        'url': 'http://miromak71.wrzuta.pl/playlista/7XfO4vE84iR/moja_muza',
+        'playlist_mincount': 14,
+        'info_dict': {
+            'id': '7XfO4vE84iR',
+            'title': 'Moja muza',
+        },
+    }, {
+        'url': 'http://heroesf70.wrzuta.pl/playlista/6Nj3wQHx756/lipiec_-_lato_2015_muzyka_swiata',
+        'playlist_mincount': 144,
+        'info_dict': {
+            'id': '6Nj3wQHx756',
+            'title': 'Lipiec - Lato 2015 Muzyka Świata',
+        },
+    }]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        playlist_id = mobj.group('id')
+        uploader = mobj.group('uploader')
+
+        entries = []
+
+        webpage = self._download_webpage(url, playlist_id)
+
+        playlist_size = self._html_search_regex(_PLAYLIST_SIZE_PATTERN, webpage, 'Size of the playlist')
+        playlist_size = int(playlist_size) if playlist_size else 0
+
+        playlist_title = self._og_search_title(webpage).replace('Playlista: ', '', 1)
+
+        if playlist_size:
+            entries = list(map(
+                lambda entry_url: self.url_result(entry_url),
+                re.findall(_ENTRY_PATTERN, webpage)
+            ))
+
+            if playlist_size > len(entries):
+                playlist_content = self._download_json(
+                    'http://{uploader_id}.wrzuta.pl/xhr/get_playlist_offset/{playlist_id}'.format(
+                        uploader_id=uploader,
+                        playlist_id=playlist_id,
+                    ),
+                    playlist_id,
+                    'Downloading playlist content as JSON metadata',
+                    'Unable to download playlist content as JSON metadata',
+                )
+                entries += [self.url_result(entry['filelink']) for entry in playlist_content['files']]
+
+        return self.playlist_result(entries, playlist_id, playlist_title)


setup.py
```diff
@@ -111,6 +111,11 @@ def run(self):
     # test_suite = 'nose.collector',
     # test_requires = ['nosetest'],
 
+    tests_require=[
+        # used to provide unit test capabilities
+        'mock >= 2.0.0',
```
dstftw: Current policy is to add additional dependencies only in case of uttermost necessity. All provided tests can be expressed via standard youtube-dl's `_TESTS` mechanism without any problem.

youtube_dl/extractor/wrzuta.py
```diff
@@ -37,23 +38,57 @@ class WrzutaIE(InfoExtractor):
...
     def _real_extract(self, url):
         mobj = re.match(self._VALID_URL, url)
-        video_id = mobj.group('id')
+        extracted_id = mobj.group('id')
         typ = mobj.group('typ')
         uploader = mobj.group('uploader')
+        media_title = mobj.group('media_title')
 
-        webpage = self._download_webpage(url, video_id)
+        if typ == 'playlista':
```
dstftw: Instead of this playlist extractor should be separated into `WrzutaPlaylistIE`.

dstftw:
One more thing is that mocking makes little sense here at all since there is no contract between the parties. Tests should be performed on real data to be able to detect breakages and format changes by remote party immediately.

setup.py
venth: OK, I'll investigate it. I've added mock because it's convenient to use especially when there isn't any possibility to reach Internet - like during flight.

I'll look at other extractors to see how it's used or use pure Python to mock in worst case scenario.

youtube_dl/extractor/wrzuta.py
venth: Got it. Looking at others extractors I got an impression that you want to keep all logic related to one site within single extractor. If creating a new one for extracting a playlist is allowed then it's good. I'll revert my changes done to wrzuta and create playlist extractor.

venth:
You're right. Usually, these are long ones, so I add only few of them. Nevertheless, I'll add them as well 😄

venth:
OK, I've applied agreed changes. What's next?

yan12125:
Please put all tests in ```_TESTS``` and modify ```test/test_download.py``` if needed.

venth:
OK

youtube_dl/extractor/extractors.py
```diff
@@ -972,6 +972,7 @@
 from .wistia import WistiaIE
 from .worldstarhiphop import WorldStarHipHopIE
 from .wrzuta import WrzutaIE
+from .wrzuta_playlist import WrzutaPlaylistIE
```
yan12125: ```WrzutaPlaylistIE``` should be place in ```wrzuta.py```.

youtube_dl/extractor/wrzuta_playlist.py
```diff
@@ -0,0 +1,135 @@
...
+        * download all entries provided on
+          the playlist webpage (the playlist is split
+          on two pages: first directly reached from webpage
+          second: downloaded on demand by ajax call and rendered
+          using the ajax call response)
+        * in case size of extracted entries not reached total number of entries
+          use the ajax call to collect the remaining entries
+    """
+
+    IE_NAME = 'wrzuta.pl - playlist'
```
yan12125: Should be ```wrzuta.pl:playlist```.

youtube_dl/extractor/wrzuta_playlist.py
```diff
@@ -0,0 +1,135 @@
...
+
+            number_of_extracted_entries_still_lower_than_list_size = playlist_size > len(entries)
+            if number_of_extracted_entries_still_lower_than_list_size:
+                entries += self._collect_and_extract_remaining_entries(playlist_id, uploader)
+
+        return self.playlist_result(
+            entries=entries,
+            playlist_id=playlist_id,
+            playlist_title=playlist_title,
+            playlist_description=playlist_title,
```
yan12125: There's no need to include a description if it's identical to the title.

youtube_dl/extractor/wrzuta_playlist.py
```diff
@@ -0,0 +1,135 @@
...
+            'title': 'andziaoox_moja_playlista_1',
+            'description': 'andziaoox_moja_playlista_1',
+        },
+    }]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        extracted_id = mobj.group('id')
+        uploader = mobj.group('uploader')
+        media_title = mobj.group('playlist_title')
```
yan12125: Title should be human readable. Take http://andziaoox.wrzuta.pl/playlista/1TOgrSX1hFR/andziaoox_moja_playlista_1 for example, the title can be ```andziaoox / moja playlista #1``` or ```moja playlista #1```.

youtube_dl/extractor/wrzuta_playlist.py
```diff
@@ -0,0 +1,135 @@
...
+
+    @staticmethod
+    def _extract_html_entries(webpage):
+        _entry_pattern = r'<a href="(?P<playlist_entry_url>[^"]+)" target="_blank" class="playlist\-file\-page">'
+        return re.findall(_entry_pattern, webpage)
+
+    @staticmethod
+    def _extract_playlist_size(webpage):
+        _size_pattern = r'<div class="playlist-counter">[0-9]+/(?P<playlist_size>[0-9]+)</div>'
+        matched_obj = re.search(_size_pattern, webpage)
```
yan12125: Use ```_html_search_regex()``` instead. And in general there's no need to create functions for short codes that are used only once.

.gitignore
```diff
@@ -34,6 +34,9 @@ test/testdata
 test/local_parameters.json
 .tox
 youtube-dl.zsh
+
+# IntelliJ related files
 .idea
-.idea/*
+*.iml
```
yan12125: Please open a separate pull request for this.

venth: OK

youtube_dl/extractor/wrzuta_playlist.py
venth: Got it. I'll correct it.

youtube_dl/extractor/wrzuta_playlist.py
venth: OK

youtube_dl/extractor/wrzuta_playlist.py
venth: OK

youtube_dl/extractor/extractors.py
venth: OK

youtube_dl/extractor/wrzuta_playlist.py
venth: > Use _html_search_regex() instead
For `_extract_html_entries()` might be difficult to replace regexp with `_html_search_regex() `. I had quick look at the method and didn't see a way to return collection (or iterator) of matched groups. I could use `clean_html()` and `strip()` functions along with the regular expression instead of `_extract_html_entries()` function or you could help me by giving a hint of usage - perhaps I missed something.

yan12125: Currently there's no good replacement for ```re.findall``` - just leave it alone.

venth: OK

venth:
Changes applied

venth:
Guys, is it good or shall I correct something?

youtube_dl/extractor/wrzuta.py
```diff
@@ -80,3 +80,135 @@ def _real_extract(self, url):
...
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        extracted_id = mobj.group('id')
+        uploader = mobj.group('uploader')
+
+        return self._extracted_playlist_content(
+            url=url,
+            uploader=uploader,
+            playlist_id=extracted_id,
```
dstftw: What's the point of such verbosity? It can be safely written on a single line skipping arg names - they are not keyworded anyway.

youtube_dl/extractor/wrzuta.py
```diff
@@ -80,3 +80,135 @@ def _real_extract(self, url):
...
+
+        webpage = self._download_webpage(
+            url_or_request=url,
+            video_id=playlist_id,
+        )
+        playlist_size = self._extracted_playlist_size(webpage)
+        playlist_title = self._extracted_playlist_title(webpage)
+
+        there_are_entries_for_extraction = playlist_size > 0
+        if there_are_entries_for_extraction:
```
dstftw: `if playlist_size:` is enough.

youtube_dl/extractor/wrzuta.py
```diff
@@ -80,3 +80,135 @@ def _real_extract(self, url):
...
+            entries = self._extracted_entries_from_webpage(webpage)
+
+            number_of_extracted_entries_still_lower_than_list_size = playlist_size > len(entries)
+            if number_of_extracted_entries_still_lower_than_list_size:
```
dstftw: `if playlist_size > len(entries)` is enough.

youtube_dl/extractor/wrzuta.py
```diff
@@ -80,3 +80,135 @@ def _real_extract(self, url):
...
+        )
+
+    @staticmethod
+    def _extracted_html_entries(webpage):
+        _entry_pattern = r'<a href="(?P<playlist_entry_url>[^"]+)" target="_blank" class="playlist\-file\-page">'
+        return re.findall(_entry_pattern, webpage)
+
+    def _extracted_playlist_size(self, webpage):
+        playlist_size = self._html_search_regex(
+            pattern=r'<div class="playlist-counter">[0-9]+/(?P<playlist_size>[0-9]+)</div>',
```
dstftw: When there is only one capture group just use unnamed one.

youtube_dl/extractor/wrzuta.py
```diff
@@ -80,3 +80,135 @@ def _real_extract(self, url):
...
+            uploader=uploader,
+            playlist_id=extracted_id,
+        )
+
+    def _extracted_playlist_title(self, webpage):
+        return self._html_search_meta(
+            name='og:title',
+            html=webpage,
+            display_name='playlist title'
+        ).replace('Playlista: ', '', 1)
```
dstftw: Use `self._og_search_title`.

youtube_dl/extractor/wrzuta.py
```diff
@@ -80,3 +80,135 @@ def _real_extract(self, url):
...
+            video_id=playlist_id,
+        )
+        playlist_size = self._extracted_playlist_size(webpage)
+        playlist_title = self._extracted_playlist_title(webpage)
+
+        there_are_entries_for_extraction = playlist_size > 0
+        if there_are_entries_for_extraction:
+            entries = self._extracted_entries_from_webpage(webpage)
+
+            number_of_extracted_entries_still_lower_than_list_size = playlist_size > len(entries)
```
dstftw: Does not work on python 3:
```
[debug] System config: []
[debug] User config: ['-o', '%(title)s-%(id)s-f%(format_id)s.%(ext)s']
[debug] Command-line args: ['http://heroesf70.wrzuta.pl/playlista/6Nj3wQHx756/lipiec_-_lato_2015_muzyka_swiata', '-F', '-v']
[debug] youtube-dl version 2016.06.03
[debug] Git HEAD: c173b4e
[debug] Python version 3.5.1 - Windows-10-10.0.10240-SP0
[debug] exe versions: ffmpeg 3.0, ffprobe 3.0, rtmpdump 2.4
[debug] Proxy map: {}
[wrzuta.pl:playlist] 6Nj3wQHx756: Downloading webpage
  File ".\youtube_dl\__main__.py", line 19, in <module>
    youtube_dl.main()
  File "C:\Dev\youtube-dl\master\youtube_dl\__init__.py", line 421, in main
    _real_main(argv)
  File "C:\Dev\youtube-dl\master\youtube_dl\__init__.py", line 411, in _real_main
  File "C:\Dev\youtube-dl\master\youtube_dl\YoutubeDL.py", line 1736, in download
  File "C:\Dev\youtube-dl\master\youtube_dl\YoutubeDL.py", line 676, in extract_info
    ie_result = ie.extract(url)
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\common.py", line 341, in extract
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\wrzuta.py", line 149, in _real_extract
    playlist_id=extracted_id,
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\wrzuta.py", line 173, in _extracted_playlist_content
    number_of_extracted_entries_still_lower_than_list_size = playlist_size > len(entries)
TypeError: object of type 'map' has no len()
```

youtube_dl/extractor/wrzuta.py
```diff
@@ -80,3 +80,135 @@ def _real_extract(self, url):
...
+
+    def _extracted_playlist_content(self, url, uploader, playlist_id):
+        entries = []
+
+        webpage = self._download_webpage(
+            url_or_request=url,
+            video_id=playlist_id,
+        )
+        playlist_size = self._extracted_playlist_size(webpage)
+        playlist_title = self._extracted_playlist_title(webpage)
```
dstftw: What's the point extracting one line routine in separate method that is used only once? This just complicates readability.

youtube_dl/extractor/wrzuta.py
```diff
@@ -80,3 +80,135 @@ def _real_extract(self, url):
...
+            'id': '6Nj3wQHx756',
+            'title': 'Lipiec - Lato 2015 Muzyka Świata',
+        },
+    }, {
+        'url': 'http://andziaoox.wrzuta.pl/playlista/1TOgrSX1hFR/andziaoox_moja_playlista_1',
+        'playlist_mincount': 29,
+        'info_dict': {
+            'id': '1TOgrSX1hFR',
+            'title': 'andziaoox / moja playlista #1',
+        },
```
dstftw: From what I can see all these tests test the same extraction scenarios thus. If so leave only two: one that test playlist without additional ajax and one with. That's enough.

youtube_dl/extractor/wrzuta.py
venth: Readability :) I believe, it's matter of taste. I'll convert it to a single line, without keywords.

youtube_dl/extractor/wrzuta.py
venth: OK

youtube_dl/extractor/wrzuta.py
venth: Instead of putting a comment, I extracted a method. I'll remove the extracted method and embed the content into the caller one.

youtube_dl/extractor/wrzuta.py
venth: It will.... after some changes ;) sorry my bad. I need to install python3

youtube_dl/extractor/wrzuta.py
venth: It is, but for me was too terse. Instead of putting a comment, I extracted the variable. I'll revert this.

youtube_dl/extractor/wrzuta.py
venth: OK

youtube_dl/extractor/wrzuta.py
venth: They cover following scenarios:
- happy path without,
- happy path with ajax call,
- one entry above the limit, when ajax call will come to the play,
- big number of entries
- one entry below limit, when ajax call will come to the play

The additional thing is verification of special characters in title and description.

I'll shrink them down to two.

youtube_dl/extractor/wrzuta.py
venth: Yes, it is. For me it was too terse. I'll revert it to your proposal.

venth:
Changes applied.

venth:
Is there anything else, what should I correct?

dstftw:
Not all changed applied.

youtube_dl/extractor/wrzuta.py
dstftw: This.

youtube_dl/extractor/wrzuta.py
```diff
@@ -80,3 +80,100 @@ def _real_extract(self, url):
...
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        playlist_id = mobj.group('id')
+        uploader = mobj.group('uploader')
+
+        entries = []
+
+        webpage = self._download_webpage(
+            url_or_request=url,
+            video_id=playlist_id,
```
dstftw: I've already pointed out superfluous verbosity.

venth: I don't agree that is superfluous 😀 Nevertheless it's my bad. I thought I have removed it. Sorry, next push will have it changed.

dstftw:
Cherry-picked in fea55ef4a95d226668bd63742c4731832de93a79, thanks.

venth:
Thanks guys 😀 Next time will be quicker 😜

PR #9342, opened 2016-04-28
Flipagram add new extractor

jerryaalvarado84:
Added extractor for Flipagram videos

### DIFF
--- a/youtube_dl/extractor/flipagram.py
+++ b/youtube_dl/extractor/flipagram.py
@@ -1,0 +1,58 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from ..utils import get_element_by_attribute
+
+
+class FlipagramIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?flipagram\.com/f/(?P<id>[^/?_]+)'
+    _TESTS = [{
+        'url': 'https://flipagram.com/f/mSxPSOFyid',
+        'info_dict': {
+            'url': 'https://d2fab04skj7pig.cloudfront.net/4bfd9a5d4733f6a34b2301af3bae0bb402c5a299_967905053_1458804346377.mp4',
+            'id': 'mSxPSOFyid',
+            'ext': 'mp4',
+            'title': 'Video by Irene M Retno Widiati',
+        }
+    }, {
+        'url': 'https://flipagram.com/f/nm44HumIuD',
+        'info_dict': {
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/flipagram.py
```diff
@@ -0,0 +1,58 @@
...
+        'info_dict': {
+            'url': 'https://d2fab04skj7pig.cloudfront.net/e6dd15c6b49d7b66306c44790a31f722d46a2322_2128787897_1460930458561.mp4',
+            'id': 'nm44HumIuD',
+            'ext': 'mp4',
+            'title': 'Video by Sarah Willems'
+        }
+    }]
+
+    @staticmethod
+    def _extract_embed_url(webpage):
```
dstftw: This method is unused. Either use somewhere or remove.

youtube_dl/extractor/flipagram.py
```diff
@@ -0,0 +1,58 @@
...
+        mobj = re.search(
+            r'<a[^>]+href=[\'"])(?P,link.[^\'"]+)\1', blockquote_el)
+        if mobj:
+            return mobj.group('link')
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+
+        video_id = mobj.group('id')
+        webpage_url = 'https://flipagram.com/f/' + video_id
```
dstftw: Whats the point reconstructing the URL? You already have it in `url`.

youtube_dl/extractor/flipagram.py
```diff
@@ -0,0 +1,58 @@
...
+
+        mobj = re.search(
+            r'<a[^>]+href=[\'"])(?P,link.[^\'"]+)\1', blockquote_el)
+        if mobj:
+            return mobj.group('link')
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+
+        video_id = mobj.group('id')
```
dstftw: Use `self._match_id` instead.

youtube_dl/extractor/flipagram.py
```diff
@@ -0,0 +1,58 @@
...
+        webpage = self._download_webpage(webpage_url, video_id)
+
+        self.report_extraction(video_id)
+
+        video_url = self._html_search_regex(r'"contentUrl":"https://d2fab04skj7pig.cloudfront.net/(.+?)"', webpage, u'video_URL', fatal=False)
+
+        return{
+            'id': video_id,
+            'ext': 'mp4',
+            'url': "https://d2fab04skj7pig.cloudfront.net/" + video_url,
```
dstftw: Again, why reconstruct rather than extract whole URL in `video_url`?

dstftw:
More metadata for extraction in `window.reactH2O`.

dstftw:
Superseded by #9538.

Issue #9076, opened 2016-04-04
FFmpegMetadataPP does not process `creator` field from info extractor


blahgeek:
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.04.01**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)

---

In FFmpegMetadataPP:

```
        if info.get('artist') is not None:
            metadata['artist'] = info['artist']
```

Only `artist` field is used. 

But in the documentation of `InfoExtractor`:

```
creator:        The main artist who created the video.
```

There's no `artist` field from any info extractor.

Shouldn't it be something like

```
        if info.get('creator') is not None:
            metadata['artist'] = info['creator']
```

?

Thanks!

blahgeek:
Also, `FFmpegMetadataPP` uses `album` field as metadata, which is also not mentioned in the documentation of `InfoExtractor`

PR #9357, opened 2016-05-01
[FFmpeg postprocess] use `creator' field as artist

blahgeek:
fix issue #9076


dstftw:
I guess `creator` should take precedence over uploader fields.

PR #8201, opened 2016-01-10
[downloader/hls] Add support for AES-128 encrypted segments in hlsnative downloader

remitamine:

PR #9358, opened 2016-05-01
[downloader/hls] Delegate extraction to ffmpeg when unsupported features detected

dstftw:
This PR introduces the delegation feature for native hls downloader: i.e. native hls downloader will delegate extraction to `FFmpegFD` when it realizes that it's not capable of downloading requested stream.

This can be useful when:
 - Some extractor forces `entry_protocol='m3u8_native'` in `InfoExtractor._extract_m3u8_formats` but eventually (e.g. due to some change on remote side) incoming manifest start using features hlsnative does not support.
 - When one forces `--hls-prefer-native` (e.g. in configuration file) and tries to download stream hlsnative is not capable to.

If we carefully craft the set of unsupported features we can even force `m3u8_native=True` in `InfoExtractor._extract_m3u8_formats` though this will require additional request to download manifest all the time when hlsnative is not capable to handle it.

### DIFF
--- a/youtube_dl/downloader/hls.py
+++ b/youtube_dl/downloader/hls.py
@@ -4,6 +4,7 @@ import os.path
 import re
 
 from .fragment import FragmentFD
+from .external import FFmpegFD
 
 from ..compat import compat_urlparse
 from ..utils import (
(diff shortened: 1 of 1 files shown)


remitamine:
> If we carefully craft the set of unsupported features we can even force m3u8_native=True in InfoExtractor._extract_m3u8_formats though this will require additional request to download manifest all the time when hlsnative is not capable to handle it.

i think it's better to keep the default to reduce request if it's possible and enable it by default in places when the first point happen:
examples of the places that it will be usfull to use it:
- sites that uses like tf1, tubitv.
- sites that depend on other extractors that serve both types(encrypted and non encrypted) like sites that depend on brightcove and theplatform.
so in general if we know that the site always serve encrypted/live manifests than it's better to use ffmpeg by default else use HlsFD downloader.

remitamine:
this will be also helpful for #8201(if pycrypto is not available than it fall back to ffmpeg).

PR #9367, opened 2016-05-02
Feature: --restrict-filenames: replace accented characters by their unaccented counterpart instead of "_"

codesparkle:
Fixed https://github.com/rg3/youtube-dl/issues/9347.

### DIFF
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -89,6 +89,11 @@ KNOWN_EXTENSIONS = (
     'wav',
     'f4f', 'f4m', 'm3u8', 'smil')
 
+# needed for sanitizing filenames in restricted mode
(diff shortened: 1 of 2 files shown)


youtube_dl/utils.py
```diff
@@ -365,6 +365,11 @@ def sanitize_filename(s, restricted=False, is_id=False):
     Set is_id if this is not an arbitrary string, but an ID that should be kept if possible
     """
     def replace_insane(char):
+        accents = dict(zip('ÂÃÄÀÁÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõöøùúûüýþÿ',
+                           itertools.chain('AAAAAA', ['AE'], 'CEEEEIIIIDNOOOOOOUUUUYP', ['ss'],
+                                           'aaaaaa', ['ae'], 'ceeeeiiiionoooooouuuuypy')))
+        if restricted and char in accents:
+            return accents[char]
```
dstftw: It should not be in `replace_insane`. Accents table should be build once and not every time.

codesparkle:
@dstftw you're right, I measured the difference using timeit. fixed by moving the dict creation outside of the function.

yan12125:
Thanks.

PR #9371, opened 2016-05-02
[neteasemusic] Add album info

blahgeek:
For a complete ID3 information.


youtube_dl/extractor/neteasemusic.py
```diff
@@ -85,6 +85,8 @@ class NetEaseMusicIE(NetEaseMusicBaseIE):
             'ext': 'mp3',
             'title': 'Bad Blood (feat. Kendrick Lamar)',
             'creator': 'Taylor Swift / Kendrick Lamar',
+            'album': 'Bad Blood (feat. Kendrick Lamar)',
+            'album_artist': 'Taylor Swift',
```
dstftw: `album_artist` is only used for compilation/splits/etc.

blahgeek: OK... so no `album_artist` then.

dstftw: It should be `artist` for artist of this particular track.

youtube_dl/extractor/neteasemusic.py
```diff
@@ -84,7 +84,8 @@ class NetEaseMusicIE(NetEaseMusicBaseIE):
             'id': '32102397',
             'ext': 'mp3',
             'title': 'Bad Blood (feat. Kendrick Lamar)',
-            'creator': 'Taylor Swift / Kendrick Lamar',
```
dstftw: `creator` should be preserved for backward compatibility.

blahgeek: So, are you suggesting that,

for tracks with single artist, the information should be:

```
'creator': 'A',
'artist': 'A',
```

and for tracks with multiple artists, the information should be

```
'creator': 'A',
'artist': 'A / B'
```

?

That's seems redundant.

---

Technically speaking, as `creator` is `The main artist who created the video` and `artist` is `Artist(s) of the track`, the `creator` field should be extracted from `artist` field automatically by youtube-dl core.

dstftw:
It's also make sense to extract `track` and `track_number` and change `title`'s format to `$artist - $track`.

youtube_dl/extractor/neteasemusic.py
dstftw: `creator` should not necessarily contain artist in terms of a person or band. It can be a studio/channel/whatever. The  meaning of `creator` has expanded from the initial description wording.

dstftw: Redundant or not it should be preserved either by providing manually or by copying from `artist` automatically when not present (that is not implemented) since it can be used by someone already.

PR #9394, opened 2016-05-04
[Downloader] --hls-prefer-ffmpeg fix

Maxime-J:
Changing hls protocol check so that correct parameters are passed to ffmpeg in [downloader/external.py#L227](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/downloader/external.py#L227)

### DIFF
--- a/youtube_dl/downloader/__init__.py
+++ b/youtube_dl/downloader/__init__.py
@@ -30,6 +30,10 @@ PROTOCOL_MAP = {
 def get_suitable_downloader(info_dict, params={}):
     """Get the downloader class that can handle the info dict."""
     protocol = determine_protocol(info_dict)
+    if protocol == 'm3u8' and params.get('hls_prefer_native') is True:
+        protocol = 'm3u8_native'
+    elif protocol == 'm3u8_native' and params.get('hls_prefer_native') is False:
+        protocol = 'm3u8'
     info_dict['protocol'] = protocol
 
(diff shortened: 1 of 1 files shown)


youtube_dl/downloader/__init__.py
```diff
@@ -30,6 +30,10 @@
 def get_suitable_downloader(info_dict, params={}):
     """Get the downloader class that can handle the info dict."""
     protocol = determine_protocol(info_dict)
+    if protocol == 'm3u8' and params.get('hls_prefer_native') is True:
+        protocol = 'm3u8_native'
+    elif protocol == 'm3u8_native' and params.get('hls_prefer_native') is False:
+        protocol = 'm3u8'
```
dstftw: `protocol` should not be touched. `protocol` is protocol selected according to extracted info dictionary not according to provided command line options' configuration.

Maxime-J: I don't see the problem,
we change protocol before storing it to info_dict['protocol'].

There's two advantages:
-We fix the current problem whith hls-prefer-ffmpeg without breaking anything.
-We have a cleaner code without multiple returns.

dstftw: I've already said `protocol` as well as any other value in `info` dict **must be original**. There are conventional `requested_*` fields for bookkeeping purposed.

Maxime-J: Ok, so if you don't want to do it this way,
Could you add `m3u8_native` in the [external.py#L227](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/downloader/external.py#L227) check because --hls-prefer-ffmpeg is currently broken.

Issue #3611, opened 2014-08-28
[site-support-request] afreeca.com


ubless607:
<pre>C:\rtmp>youtube-dl --verbose http://www.afreeca.com/player/Player.swf?nTitleNo=12851241&szBjId=mistyj&nBbsNo=12387661&nStationNo=12049126
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['--verbose', 'http://www.afreeca.com/player/Player.swf?nTitleNo=12851241']
[debug] Encodings: locale cp949, fs mbcs, out cp949, pref cp949
[debug] youtube-dl version 2014.08.29
[debug] Python version 2.7.8 - Windows-7-6.1.7601-SP1
[debug] Proxy map: {}
[generic] Player: Requesting header
WARNING: Falling back on generic information extractor.
[generic] Player: Downloading webpage
[generic] Player: Extracting information
ERROR: Unsupported URL: http://www.afreeca.com/player/Player.swf?nTitleNo=12851241; please report this issue on https://yt-dl.org/bug . Be sure to call youtube-dl with the --verbose flag and include its complete output. Make sure you are using the latest version; type  youtube-dl -U  to update.
Traceback (most recent call last):
  File "youtube_dl\extractor\generic.pyo", line 530, in _real_extract
  File "youtube_dl\utils.pyo", line 1469, in parse_xml
  File "xml\etree\ElementTree.pyo", line 1300, in XML
  File "xml\etree\ElementTree.pyo", line 1642, in feed
  File "xml\etree\ElementTree.pyo", line 1506, in _raiseerror
ParseError: syntax error: line 1, column 0
Traceback (most recent call last):
  File "youtube_dl\YoutubeDL.pyo", line 521, in extract_info
  File "youtube_dl\extractor\common.pyo", line 178, in extract
  File "youtube_dl\extractor\generic.pyo", line 869, in _real_extract
ExtractorError: Unsupported URL: http://www.afreeca.com/player/Player.swf?nTitleNo=12851241; please report this issue on https://yt-dl.org/bug . Be sure to call youtube-dl with the --verbose flag and include its complete output. Make sure you are using the latest version; type  youtube-dl -U  to update.

'szBjId'은(는) 내부 또는 외부 명령, 실행할 수 있는 프로그램, 또는
배치 파일이 아닙니다.
'nBbsNo'은(는) 내부 또는 외부 명령, 실행할 수 있는 프로그램, 또는
배치 파일이 아닙니다.
[...]

Issue #2054, opened 2013-12-28
Please, add support for afreeca.com


AleXoundOS:
Example: http://afbbs.afreeca.com:8080/app/read_ucc_bbs.cgi?szBjId=sogoodtt&nStationNo=55614&nBbsNo=124&nTitleNo=3448726&szSkin=

Afreeca is a Korean video streaming service. The best StarCraft: Brood War players stream there.

yan12125:
Thanks @pmrowla (#9395), afreeca will be supported in the next version.

AleXoundOS:
Thank you

PR #9395, opened 2016-05-04
[afreecatv] Add new extractor for afreecatv.com VODs

### LINKED ISSUE #2026: please support afreeca.com
Here is one example

http://live.afreeca.com:8079/app/index.cgi?szType=read_ucc_bbs&nStationNo=479436&nBbsNo=11989006&szBjId=tthmy&nTitleNo=5032602&

pmrowla:
afreecatv.com is a Korean streaming service. This adds support for handling afreeca VODs but it does not handle any live stream URLs.

This resolves #2026, #3611, #2054, but the videos for all three of those issues no longer exist.

### DIFF
--- a/youtube_dl/extractor/afreecatv.py
+++ b/youtube_dl/extractor/afreecatv.py
@@ -1,6 +1,8 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
+import re
+
 from .common import InfoExtractor
 from ..compat import (
     compat_urllib_parse_urlparse,
@@ -34,6 +36,15 @@ class AfreecaTVIE(InfoExtractor):
         }
     }
 
+    @staticmethod
+    def parse_video_key(key):
+        video_key = {'upload_date': None, 'part': '0'}
+        m = re.match(r'^(?P<upload_date>\d{8})_\w+_(?P<part>\d+)$', key)
+        if m:
+            video_key['upload_date'] = m.group('upload_date')
+            video_key['part'] = m.group('part')
+        return video_key
+
     def _real_extract(self, url):
         video_id = self._match_id(url)
         parsed_url = compat_urllib_parse_urlparse(url)
@@ -54,9 +65,11 @@ class AfreecaTVIE(InfoExtractor):
 
         entries = []
         for video_file in video_xml.findall('./track/video/file'):
+            video_key = self.parse_video_key(video_file.get('key'))
             entries.append({
-                'id': video_file.get('key'),
+                'id': '%s_%s' % (video_id, video_key['part']),
                 'title': title,
+                'upload_date': video_key['upload_date'],
                 'duration': int_or_none(video_file.get('duration')),
                 'url': video_file.text,
             })
@@ -75,6 +88,7 @@ class AfreecaTVIE(InfoExtractor):
             info['entries'] = entries
         elif len(entries) == 1:
             info['url'] = entries[0]['url']
+            info['upload_date'] = entries[0]['upload_date']
         else:
             raise ExtractorError(
                 'No files found for the specified AfreecaTV video, either'
--- a/youtube_dl/extractor/afreecatv.py
+++ b/youtube_dl/extractor/afreecatv.py
@@ -30,7 +30,7 @@ class AfreecaTVIE(InfoExtractor):
             'id': '36164052',
             'ext': 'mp4',
             'title': '데일리 에이프릴 요정들의 시상식!',
-            'thumbnail': 're:^https?://videoimg.afreecatv.com/.*$',
+            'thumbnail': 're:^https?://(?:video|st)img.afreecatv.com/.*$',
             'uploader': 'dailyapril',
             'uploader_id': 'dailyapril',
             'upload_date': '20160503',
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/afreecatv.py
```diff
@@ -0,0 +1,84 @@
...
+
+        track = video_xml.find('track')
+        if track.find('flag').text != 'SUCCEED':
+            raise ExtractorError('Specified AfreecaTV video does not exist',
+                                 expected=True)
+        title = track.find('title').text
+        uploader = track.find('nickname').text
+        uploader_id = track.find('bj_id').text
+        duration = int_or_none(track.find('duration').text)
+        thumbnail = track.find('titleImage').text
```
yan12125: Use ```xpath_text``` for all ```track.find('XXX').text``` occurrences. This function provides more information for debugging.

youtube_dl/extractor/afreecatv.py
```diff
@@ -0,0 +1,84 @@
...
+                                          'duration'))
+        thumbnail = xpath_text(video_xml, './track/titleImage', 'thumbnail')
+
+        entries = []
+        for video_file in video_xml.findall('./track/video/file'):
+            entries.append({
+                'id': video_file.get('key'),
+                'title': title,
+                'duration': int_or_none(video_file.get('duration')),
+                'formats': [{'url': video_file.text}]
```
yan12125: If there's only one format, just use ```'url'```.

youtube_dl/extractor/afreecatv.py
```diff
@@ -0,0 +1,126 @@
...
+                                          'duration'))
+        thumbnail = xpath_text(video_xml, './track/titleImage', 'thumbnail')
+
+        entries = []
+        for video_file in video_xml.findall('./track/video/file'):
+            video_key = self.parse_video_key(video_file.get('key'))
+            entries.append({
+                'id': '%s_%s' % (video_id, video_key['part']),
+                'title': title,
+                'upload_date': video_key['upload_date'],
```
yan12125: Should not fail if ```upload_date``` does not exist.

youtube_dl/extractor/afreecatv.py
```diff
@@ -0,0 +1,126 @@
...
+)
+
+
+class AfreecaTVIE(InfoExtractor):
+    IE_DESC = 'afreecatv.com'
+    _VALID_URL = r'''(?x)^
+        https?://(?:(live|afbbs|www)\.)?afreeca(?:tv)?\.com(?::\d+)?
+        (?:
+            /app/(?:index|read_ucc_bbs)\.cgi|
+            /player/[Pp]layer\.(?:swf|html))
```
yan12125: Any /player/ URL example? Please add it to ```_TESTS``` with ```'only_matching': True```.

youtube_dl/extractor/extractors.py
```diff
@@ -16,6 +16,7 @@
     AdobeTVVideoIE,
 )
 from .adultswim import AdultSwimIE
+from .afreecatv import AfreecaTVIE
```
yan12125: Should be after ```.aenetworks```.

yan12125:
English, Japanese and Taiwan sites use a different URL scheme. For example:
http://afreeca.tv/36864156/v/31700
http://afreecatv.jp/35845951/v/255590
http://afreecatv.com.tw/36865941/v/18590
Seems these videos use a different workflow, too. Do you want to implement them as well?

pmrowla:
@yan12125 for the global afreeca sites the API looks a lot different. I'll look into implementing them when I get the chance, but I'd prefer to leave them alone as far as this pull request goes

youtube_dl/extractor/afreecatv.py
```diff
@@ -0,0 +1,129 @@
...
+                                 expected=True)
+        title = xpath_text(video_xml, './track/title', 'title')
+        uploader = xpath_text(video_xml, './track/nickname', 'uploader')
+        uploader_id = xpath_text(video_xml, './track/bj_id', 'uploader id')
+        duration = int_or_none(xpath_text(video_xml, './track/duration',
+                                          'duration'))
+        thumbnail = xpath_text(video_xml, './track/titleImage', 'thumbnail')
+
+        entries = []
+        for i, video_file in enumerate(video_xml.findall('./track/video/file[@key]')):
```
dstftw: Selection by attribute does not work in python 2.6, use `find_xpath_attr` instead.

youtube_dl/extractor/afreecatv.py
```diff
@@ -0,0 +1,131 @@
...
+        video_id = self._match_id(url)
+        parsed_url = compat_urllib_parse_urlparse(url)
+        info_url = compat_urlparse.urlunparse(parsed_url._replace(
+            netloc='afbbs.afreecatv.com:8080',
+            path='/api/video/get_video_info.php'))
+        video_xml = self._download_xml(info_url, video_id)
+
+        if xpath_text(video_xml, './track/flag', default='FAIL') != 'SUCCEED':
+            raise ExtractorError('Specified AfreecaTV video does not exist',
+                                 expected=True)
```
yan12125: Sorry for the late response. This check may give a false alert if someday afreeca.tv decides to use a different name than ```./track/flag```. ```./track/video/file``` is more reliable. The overall extraction workflow should be:
1. Check ```./track/video/file```
2. Raise an error if no entries
3. Check other fields

yan12125:
Thanks and sorry for the late response!

PR #9400, opened 2016-05-05
[liveleak] Adds support for thumbnails and updates tests

inondle:
Added support for fetching the video thumbnails, the only issue is concerning the third extractor test. When I run youtube-dl on that url using the --get-thumbnail flag it warns me that it is falling back to the default extractor. I'm not sure if this is an issue with the extractor or if this is expected. I didn't add the thumbnail url into that test but it seems the rest work just fine.

### DIFF
--- a/youtube_dl/extractor/liveleak.py
+++ b/youtube_dl/extractor/liveleak.py
@@ -17,7 +17,8 @@ class LiveLeakIE(InfoExtractor):
             'ext': 'flv',
             'description': 'extremely bad day for this guy..!',
             'uploader': 'ljfriel2',
-            'title': 'Most unlucky car accident'
+            'title': 'Most unlucky car accident',
+            'thumbnail': 're:^https?://.*\.jpg$'
         }
     }, {
         'url': 'http://www.liveleak.com/view?i=f93_1390833151',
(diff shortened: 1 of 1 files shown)


yan12125:
In general thumbnail in tests are matched with regular expressions. See [this example](https://github.com/rg3/youtube-dl/blob/master/README.md#adding-support-for-a-new-site).

inondle:
@yan12125 How complex should this regex be? I have a pretty long one but looking through some other extractors I've noticed that many have the example regex `'re:^https?://.*\.jpg$'`. Does the thumbnail matching need to be more complex than that?

inondle:
@yan12125 How complex should this regex be? I've noticed that a lot of extractors only have the example regex `^https?://.*\.jpg$`. Would this be sufficient here?

inondle:
Sorry about the double comment. My laptop isnt showing the new comments and i thought it somehow didnt submit them :P

yan12125:
> Would this be sufficient here?

Yes.

inondle:
@yan12125 changed the tests, feel free to check it out

youtube_dl/extractor/liveleak.py
```diff
@@ -72,6 +75,8 @@ def _real_extract(self, url):
         age_limit = int_or_none(self._search_regex(
             r'you confirm that you are ([0-9]+) years and over.',
             webpage, 'age limit', default=None))
+        video_thumbnail = self._og_search_thumbnail(webpage)
+
```
dstftw: Check code with flake8 and fix all issues.

inondle:
@dstftw deleted the extra line

dstftw:

inondle:
@dstftw Sure! Could I just go by Inondle?

PR #9421, opened 2016-05-06
Adds installation of zsh completions from the setup.py file

inondle:


setup.py
```diff
@@ -49,6 +49,7 @@
     files_spec = [
         ('etc/bash_completion.d', ['youtube-dl.bash-completion']),
         ('etc/fish/completions', ['youtube-dl.fish']),
+        ('etc/zsh/completions', ['youtube-dl.zsh']),
```
dstftw: zsh completions are not stored at this location. `youtube-dl.zsh` does not exist at this point.

PR #9424, opened 2016-05-07
[beatport] Add new extractor

### LINKED ISSUE #5219: Add support for www.beatport.com
example url: https://www.beatport.com/flamingorecordingsbps/tracks/qpwhxukynk7r/burnin-up-original-mix

Thanks!

Thor77:
Fix #5219

### DIFF
--- a/youtube_dl/extractor/beatport.py
+++ b/youtube_dl/extractor/beatport.py
@@ -1,0 +1,78 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from youtube_dl.utils import int_or_none
+
+from .common import ExtractorError, InfoExtractor
+
+
+class BeatportIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?beatport\.com/(?P<uploader_id>\w+)/tracks/(?P<id>.+)/(?P<display_id>.+)'
+    _TESTS = [{
+        'url': 'https://www.beatport.com/musicalfreedom/tracks/l2axnqcd6azu/secrets-feat-vassy-original-mix',
+        'md5': '11e330f6236be07f61fcc4669e972cd8',
+        'info_dict': {
+            'id': 'l2axnqcd6azu',
+            'display_id': 'secrets-feat-vassy-original-mix',
+            'uploader': 'Musical Freedom',
+            'uploader_id': 'musicalfreedom',
+            'uploader_url': 'http://www.musicalfreedom.com',
+            'ext': 'mp4',
+            'title': 'Secrets feat. Vassy',
+            'duration': 267,
+            'artist': 'Tiësto, VASSY, KSHMR',
+            'genre': 'Electro House',
+            'release_year': '2015',
+            'thumbnail': 're:^https?://.*\.jpg',
+            'view_count': int,
+            'like_count': int
+        }
+    }, {
+        'url': 'https://www.beatport.com/flamingorecordingsbps/tracks/qpwhxukynk7r/burnin-up-original-mix',
+        'md5': '67bb3421c43f92b5a625c2065255925f',
+        'info_dict': {
+            'id': 'qpwhxukynk7r',
+            'display_id': 'burnin-up-original-mix',
+            'uploader': 'Flamingo Recordings',
+            'uploader_id': 'flamingorecordingsbps',
+            'uploader_url': 'http://www.flamingorecordings.com',
+            'ext': 'mp4',
+            'title': "Burnin' Up",
+            'duration': 272,
+            'artist': 'Flatdisk',
+            'genre': 'Progressive House',
+            'release_year': '2015',
+            'thumbnail': 're:^https?://.*\.jpg',
+            'view_count': int,
+            'like_count': int
+        }
+    }
+    ]
+
+    def _real_extract(self, url):
+        url_data = self._VALID_URL_RE.match(url)
+        track_id = url_data.group('id')
+        api_base = 'https://apix.beatport.com'
+        api_data = self._download_json('%s/sounds/%s?format=json' % (api_base, track_id),
+                                       track_id)
+        if api_data['is_streamable']:
+            url = self._download_json('%s/sounds/%s/stream?format=json' % (api_base, track_id),
+                                      track_id, note='Downloading JSON streamdata')['url']
+        else:
+            raise ExtractorError('Not streamable', video_id=track_id, expected=True)
+        return {
+            'url': url,
+            'id': track_id,
+            'display_id': url_data.group('display_id'),
+            'uploader': api_data.get('owner_profile', {}).get('display_name'),
+            'uploader_url': api_data.get('owner_profile', {}).get('website_url'),
+            'uploader_id': url_data.group('uploader_id'),
+            'title': api_data['name'],
+            'duration': int_or_none(api_data.get('duration'), scale=1000),
+            'view_count': int_or_none(api_data.get('play_count')),
+            'like_count': int_or_none(api_data.get('heart_count')),
+            'artist': ', '.join(a['display_name'] for a in api_data.get('artists', [])),
+            'genre': ', '.join(g['name'] for g in api_data.get('genres', [])),
+            'release_year': api_data.get('pro_released_date', '').split('-')[0],
+            'thumbnail': api_data.get('image')
+        }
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/beatport.py
```diff
@@ -0,0 +1,78 @@
...
+            'display_id': 'secrets-feat-vassy-original-mix',
+
+            'uploader': 'Musical Freedom',
+            'uploader_id': 'musicalfreedom',
+            'uploader_url': 'http://www.musicalfreedom.com',
+
+            'ext': 'mp4',
+            'title': 'Secrets feat. Vassy',
+            'duration': 267441,
+            'artist': ['Tiësto', 'VASSY', 'KSHMR'],
```
dstftw: Must be string.

youtube_dl/extractor/beatport.py
```diff
@@ -0,0 +1,78 @@
...
+            'genre': ['Electro House'],
```
dstftw: Must be string.

youtube_dl/extractor/beatport.py
```diff
@@ -0,0 +1,78 @@
...
+            'id': 'l2axnqcd6azu',
+            'display_id': 'secrets-feat-vassy-original-mix',
+
+            'uploader': 'Musical Freedom',
+            'uploader_id': 'musicalfreedom',
+            'uploader_url': 'http://www.musicalfreedom.com',
+
+            'ext': 'mp4',
+            'title': 'Secrets feat. Vassy',
+            'duration': 267441,
```
dstftw: Must be in seconds.

youtube_dl/extractor/beatport.py
Thor77: Uh, is there no option to provide multiple artists(/genres)?

youtube_dl/extractor/beatport.py
```diff
@@ -0,0 +1,78 @@
...
+            'duration': 272822,
+            'artist': ['Flatdisk']
+        }
+    }
+    ]
+
+    def _real_extract(self, url):
+        url_data = self._VALID_URL_RE.match(url)
+        api_base = 'https://apix.beatport.com'
+        api_data = self._download_json('{}/sounds/{}?format=json'.format(api_base, url_data.group('id')),
```
dstftw: `{}` does not work on python 2.6.

youtube_dl/extractor/beatport.py
```diff
@@ -0,0 +1,78 @@
...
+        return {
+            'url': url,
+            'id': url_data.group('id'),
+            'display_id': url_data.group('display_id'),
+
+            'uploader': api_data.get('owner_profile', {}).get('display_name'),
+            'uploader_url': api_data.get('owner_profile', {}).get('website_url'),
+            'uploader_id': url_data.group('uploader_id'),
+
+            'title': api_data.get('name'),
```
dstftw: `title` is mandatory.

youtube_dl/extractor/beatport.py
```diff
@@ -0,0 +1,78 @@
...
+            'duration': api_data.get('duration'),
+
+            'view_count': api_data.get('play_count'),
```
dstftw: All integer fields must be wrapped in `utils.int_or_none`.

youtube_dl/extractor/beatport.py
dstftw: Multiple artists/genres should be provided as string.

Thor77: I seperated them by `,` now.
During my research on that topic, i didn't really find any official information in id3's design-document. Some people said, it depends on the program to parse that information correctly, others suggested using `/`, `,` and `;`...

dstftw: It should have the same format as on the particular website.

Thor77: Are you reffering to this part of the specification?
```
All text information frames supports multiple strings, stored as a null
separated list, where null is reperesented by the termination code
for the charater encoding.
```
=> Use `\0`?

dstftw: I'm not referring to any specification. It should be formatted the same way it formatted on beatport.com website.

Thor77: [artist_list](https://www.beatport.com/musicalfreedom/tracks/l2axnqcd6azu/secrets-feat-vassy-original-mix)
Okey, i will add a space to the seperator.

youtube_dl/extractor/beatport.py
```diff
@@ -0,0 +1,87 @@
...
+    def _real_extract(self, url):
+        url_data = self._VALID_URL_RE.match(url)
+        api_base = 'https://apix.beatport.com'
+        api_data = self._download_json('%s/sounds/%s?format=json' % (api_base, url_data.group('id')),
+                                       url_data.group('id'))
+        if api_data['is_streamable']:
+            url = self._download_json('%s/sounds/%s/stream?format=json' % (api_base, url_data.group('id')),
+                                      url_data.group('id'), note='Downloading JSON streamdata')['url']
+        else:
+            url = None
```
dstftw: What is the point of this? If video url can't be extracted the whole further extraction makes no sense.

Thor77: I don't want to execute foreseeable failing requests.
Additionally i didn't find a proper way to abort the extraction-process immediately (`InfoExtractor.report_warning` could've been useful here though) and so opted for this solution.

dstftw: Raise `ExtractorError`.

youtube_dl/extractor/beatport.py
```diff
@@ -0,0 +1,87 @@
...
+
+class BeatportIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?beatport\.com/(?P<uploader_id>\w+)/tracks/(?P<id>.+)/(?P<display_id>.+)'
+    _TESTS = [{
+        'url': 'https://www.beatport.com/musicalfreedom/tracks/l2axnqcd6azu/secrets-feat-vassy-original-mix',
+        'md5': '11e330f6236be07f61fcc4669e972cd8',
+        'info_dict': {
+            'id': 'l2axnqcd6azu',
+            'display_id': 'secrets-feat-vassy-original-mix',
+
```
dstftw: Remove unnecessary blank lines.

youtube_dl/extractor/beatport.py
```diff
@@ -0,0 +1,87 @@
...
+            'uploader_id': 'musicalfreedom',
+            'uploader_url': 'http://www.musicalfreedom.com',
+
+            'ext': 'mp4',
+            'title': 'Secrets feat. Vassy',
+            'duration': 267,
+            'artist': 'Tiësto, VASSY, KSHMR',
+            'genre': 'Electro House',
+            'release_year': '2015',
+            'thumbnail': 'https://api-media.beatport.com/sounds/l2axnqcd6azu/1431513641.jpg',
```
dstftw: Replace with just `'thumbnail': 're:^https?://.*\.jpg',`.

youtube_dl/extractor/beatport.py
```diff
@@ -0,0 +1,87 @@
...
+            'view_count': int,
+            'like_count': int
+        }
+    }
+    ]
+
+    def _real_extract(self, url):
+        url_data = self._VALID_URL_RE.match(url)
+        api_base = 'https://apix.beatport.com'
+        api_data = self._download_json('%s/sounds/%s?format=json' % (api_base, url_data.group('id')),
```
dstftw: You are calculating `url_data.group('id')` over and over again. Extract it into a variable.

dstftw:
Also squash commits when finished.

Thor77:
Done. Thanks for your review!

dstftw:
Apparently they are shutting down.
image
Does it work with pro.beatport.com?

Thor77:
There's already an [extractor](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/beatportpro.py) for BeatportPro

Thor77:
I guess you can close this, as there's no sense to include an extractor for a closed platform.

yan12125:
Now both https://www.beatport.com/musicalfreedom/tracks/l2axnqcd6azu/secrets-feat-vassy-original-mix and https://www.beatport.com/flamingorecordingsbps/tracks/qpwhxukynk7r/burnin-up-original-mix redirect to https://pro.beatport.com/. Closing.

PR #9431, opened 2016-05-07
[ceskatelevize] Live stream may not have title or description

sleep-walker:
This is the only thing which prevented run on live stream of Czech TV.

### DIFF
--- a/youtube_dl/extractor/ceskatelevize.py
+++ b/youtube_dl/extractor/ceskatelevize.py
@@ -33,14 +33,13 @@ class CeskaTelevizeIE(InfoExtractor):
             'skip_download': True,
         },
     }, {
-        'url': 'http://www.ceskatelevize.cz/ivysilani/10532695142-prvni-republika/bonus/14716-zpevacka-z-duparny-bobina',
+        # live stream
+        'url': 'http://www.ceskatelevize.cz/ivysilani/zive/ct4/',
         'info_dict': {
-            'id': '61924494876844374',
+            'id': 402,
             'ext': 'mp4',
-            'title': 'První republika: Zpěvačka z Dupárny Bobina',
-            'description': 'Sága mapující atmosféru první republiky od r. 1918 do r. 1945.',
-            'thumbnail': 're:^https?://.*\.jpg',
-            'duration': 88.4,
+            'title': 're:ČT Sport.*',
+            'is_live': True,
         },
         'params': {
             # m3u8 download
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/ceskatelevize.py
```diff
@@ -118,8 +119,12 @@ def _real_extract(self, url):
         req = sanitized_Request(compat_urllib_parse_unquote(playlist_url))
         req.add_header('Referer', url)
 
-        playlist_title = self._og_search_title(webpage)
-        playlist_description = self._og_search_description(webpage)
+        try:
+            playlist_title = self._og_search_title(webpage)
+            playlist_description = self._og_search_description(webpage)
```
dstftw: Use `default=` instead. Also add some test.

dstftw:
1. Live streams does not work properly - only current part of the stream is downloaded:
```
PS C:\dev\youtube-dl\master> py -2.7 .\youtube_dl\__main__.py http://www.ceskatelevize.cz/ivysilani/zive/ct4/
[CeskaTelevize] ct4: Downloading webpage
[CeskaTelevize] ct4: Downloading JSON metadata
[CeskaTelevize] ct4: Downloading JSON metadata
[CeskaTelevize] ct4: Downloading m3u8 information
[CeskaTelevize] ct4: Downloading m3u8 information
[download] Downloading playlist: ct4
[CeskaTelevize] playlist ct4: Collected 1 video ids (downloading 1 of them)
[hlsnative] Total fragments: 3
[download] Destination: T Sport.mp4
[download] 100% of 10.77MiB in 00:09
[ffmpeg] Fixing malformated aac bitstream in "T Sport.mp4"
[download] Finished downloading playlist: ct4
```

2. Live streams should use `self._live_title` for title and set `is_live` in info dict.
3. `Živé vysílání ČT sport` and `Sledujete živé vysílání kanálu ČT sport.` can be used for title and description of playlist.

sleep-walker:
Live streams does not work properly - only current part of the stream is downloaded.
Getting URL and passing to player works, which is my primary use. It would need some continuous read of playlist (it seems to change over the time). I tried to peek into other extractors but I haven't seen similar case (yet I know a very little about m3u8 or HLS). Any hint?

    Live streams should use self._live_title for title and set is_live in info dict.
Good point, fixed.

    Živé vysílání ČT sport and Sledujete živé vysílání kanálu ČT sport. can be used for title and description of playlist.
I don't think this is necessary, parsed `title` contains TV channel name and with `self._live_title` it gets timestamp so it is good as it is.

dstftw:
>Any hint?

Just use `entry_protocol='m3u8'` in `self._extract_m3u8_formats` call for live streams to enable ffmpeg.

sleep-walker:
> Just use entry_protocol='m3u8' in self._extract_m3u8_formats call for live streams to enable ffmpeg.

It helped, thanks.

youtube_dl/extractor/ceskatelevize.py
```diff
@@ -77,6 +76,7 @@ class CeskaTelevizeIE(InfoExtractor):
     }]
 
     def _real_extract(self, url):
+        is_live = re.search('/ivysilani/zive/', url) != None
```
dstftw: Live playlist items have `"type":"LIVE"` in JSON.

dstftw:
Also squash commits when finished.

sleep-walker:
Thanks for your guidance on this.

PR #9453, opened 2016-05-11
[philharmonie] Add new extractor

cklein:


youtube_dl/extractor/philharmonie.py
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    IE_NAME = 'philharmonie.tv'
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, video_id)
+        source_url = self._search_regex(
+            r'<iframe id="vidframe".*?data-src="(https://playout.3qsdn.com/[\w-]+)"',
```
dstftw: It's a generic embed. It should be handled by generic extractor that should delegate extraction to separate 3qsdn extractor.

PR #9454, opened 2016-05-11
Bugfix: Allow colons in --add-header values.

teemuy:
Simple bug fix for off-by-one in the second argument to str.split().


dstftw:
Cherry-picked in e73b9c65e279f283b28d14be5b7173eae46d4364, thanks.

PR #9455, opened 2016-05-11
Adding additional hooks during the process

uri247:
I'm trying to build a GUI application that will take a YouTube playlist of music videos, and create an album ready to be uploaded to iPhone or Android. A special care is taken to correct metadata (thumbnail, artist, song name). For details, see http://uri247.github.io/ot/

I would like the UI to be asynchronous, so as information is coming from the extractor I would update the GUI. For this reason, I would like to have a hook when YoutubeDL finishes to extract information.

For this, in addition to: `downloading`, `error`, and `finished`, I would like to introduce a new type of hook, named **extract_info**, and, in addition to the 'status' attribute, is also have `ie_result` attribute, with the result of the extractor.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -408,6 +408,10 @@ class YoutubeDL(object):
         self._pps.append(pp)
         pp.set_downloader(self)
 
+    def _hook_progress(self, status):
+        for ph in self._progress_hooks:
+            ph(status)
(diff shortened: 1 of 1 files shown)


dstftw:
Progress hook is only supposed to report downloading progress. Just call `YoutubeDL.extract_info` on each playlist entry separately to get response after each item's info extraction.

PR #9487, opened 2016-05-13
add "--no-generic-extractor" flag 

JBuzzard:
Exclude usage of the generic extractor when processing urls.
The generic extractor sends two http requests for each url its used on.
Using the --no-generic-extractor switch facilitates using YTDL in an automated fashion against many urls without spamming those sites with two requests each from the generic extractor.


youtube_dl/YoutubeDL.py
```diff
@@ -664,6 +665,8 @@ def extract_info(self, url, download=True, ie_key=None, extra_info={},
             ies = self._ies
 
         for ie in ies:
+            if(ie.ie_key() == "Generic" and no_generic_extractor):
```
dstftw: It's idiomatic not to use parentheses in `if` statement. Also use single quotes consistently.

youtube_dl/options.py
```diff
@@ -155,6 +155,10 @@ def _hide_login_info(opts):
         action='store_true', dest='force_generic_extractor', default=False,
         help='Force extraction to use the generic extractor')
     general.add_option(
+        '--no-generic-extractor',
+        action='store_true', dest='no_generic_extractor', default=False,
+        help='prevent extraction from using the generic extractor')
```
dstftw: Should start with capital letter.

JBuzzard:
Are we deadlocked?  hopefully you are not waiting on me?    I have committed what I think are the appropriate changes to address earlier comments.  Please comment further if there are additional concerns or uncertainties.

dstftw:
FYI there are not notifications sent on new commits in PR.

dstftw:
There is one ambiguity: concrete extractors may delegate extraction back to generic extractor. When `--no-generic-extractor` is used such extractors will still successfully delegate extraction to generic extractor thus it actually does not completely prevent extraction from using generic extractor.
For example:
```
[debug] System config: []
[debug] User config: [u'-o', u'%(title)s-%(id)s-f%(format_id)s.%(ext)s']
[debug] Command-line args: [u'https://www.kickstarter.com/projects/597507018/pebble-e-paper-watch-for-iphone-and-android/posts/659178', u'-F', u'--no-generic-extractor', u'-v']
[debug] youtube-dl version 2016.05.21.2
[debug] Git HEAD: 16872aa
[debug] Python version 2.6.2 - Windows-post2008Server-6.2.9200
[debug] exe versions: ffmpeg 3.0, ffprobe 3.0, rtmpdump 2.4
[debug] Proxy map: {}
[KickStarter] 597507018: Downloading webpage
to generic
[generic] 659178: Requesting header
WARNING: Could not send HEAD request to https://www.kickstarter.com/projects/597507018/pebble-e-paper-watch-for-iphone-and-android/posts/659178: IncompleteRead(0 bytes read)
[generic] 659178: Downloading webpage
[generic] 659178: Extracting information
[vimeo] 78704821: Downloading webpage
[vimeo] 78704821: Extracting information
[vimeo] 78704821: Downloading JSON metadata
[vimeo] 78704821: Downloading m3u8 information
[info] Available formats for 78704821:
hls-meta     mp4        multiple   Quality selection URL
hls-515      mp4        480x270     515k , avc1.42C01E, mp4a.40.2
http-270p    mp4        480x270    24fps
hls-475      mp4        640x360     475k , avc1.4D401F, mp4a.40.2
http-360p    mp4        640x360    24fps
hls-1047     mp4        1280x720   1047k , avc1.4D401F, mp4a.40.2
http-720p    mp4        1280x720   24fps (best)
```

JBuzzard:
hmm...yes, I see.  Looking at the kickstarter extractor as an example, I see that the generic extractor is used only after a url matching the _VALID_URL expression has been found.  This behavior meets 'my' requirements - of not including the generic extractor when searching for an extractor to use against a given URL.  Once a URL has been found to match on of the non-generic extractors, using the generic extractor is not a problem.  I'm open to suggestions if there is perhaps a different approach that might be better. Maybe an adjustment of the flag name, or its description to help clarify its meaning would be an acceptable approach?  Perhaps changing the description to something like this 'Omit generic extractor from extractor search' ?

JBuzzard:
I've updated the --no-generic-extractor flag description to be more specific about its behavior.

chytanyak:
We are using the YTDL tool to provide a simple and easy way for our end users to detect, download and enjoy web videos. While the tool has been working great with its ever expanding list of extractors (some of which are contributed by us), we did receive some feedback from the users and partners around the excessive requests to the sites that do not have videos. This is being reported as either causing website interoperability issues or too many requests. Hence, we are proposing the above change that allows us to specify the tool via an optional command line to resolve the input URL only if it matches the patterns supported by one of the specific extractors. We plan to use this flag in our application to  address the some of the user feedback and provide a better experience overall.

Would really appreciate if you can review and approve the change and/or let us know if you have any additional comments to the latest set of diffs provided by Jeff.

Thanks,
Chytanya

PR #9503, opened 2016-05-15
Update collegehumor.py

TRox1972:
updated embed test URL and removed md5, as --test returns one audio and one video file

### DIFF
--- a/youtube_dl/extractor/collegehumor.py
+++ b/youtube_dl/extractor/collegehumor.py
@@ -35,14 +35,14 @@ class CollegeHumorIE(InfoExtractor):
             },
         }, {
             # embedded youtube video
-            'url': 'http://www.collegehumor.com/embed/6950306',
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/collegehumor.py
```diff
@@ -35,14 +35,13 @@ class CollegeHumorIE(InfoExtractor):
             },
         }, {
             # embedded youtube video
-            'url': 'http://www.collegehumor.com/embed/6950306',
+            'url': 'http://www.collegehumor.com/post/6950306/jfk-john-f-kennedy-died-this-morning-interview-prank',
```
dstftw: All test URLs **must** be matched by `_VALID_URL` of corresponding extractor.

TRox1972:
Forgot to include that in the pull request. The embed test url, /embed/, redirects to /post/, and running youtube-dl on the /embed/-url does no longer work.

dstftw:
It does not change the fact that http://www.collegehumor.com/post/6950306/jfk-john-f-kennedy-died-this-morning-interview-prank is not matched by `_VALID_URL`.

PR #9505, opened 2016-05-15
Update jpopsukitv.py

TRox1972:


dstftw:
Use proper commit messages and don't mix unrelated changes into single PR.

yan12125:
As @dstftw said, don't mix commits for different sites in a single pull request.

PR #9574, opened 2016-05-22
[Pix11] add new extractor

TRox1972:


dstftw:
There may be several videos on a page: http://pix11.com/2016/05/25/multiple-people-shot-at-concert-at-irving-plaza/.

PR #9575, opened 2016-05-22
[extractor/generic] fix RollingStone test and improve Brightcove embed extraction

TRox1972:
- Updated the RollingStone test in extractor/generic which switched from Ooyla to Brightcove.
- Brightcove ids seems to always be alphanumeric; improved regex

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -1658,10 +1658,10 @@ class GenericIE(InfoExtractor):
 
         # Look for Ooyala videos
         mobj = (re.search(r'player\.ooyala\.com/[^"?]+[?#][^"]*?(?:embedCode|ec)=(?P<ec>[^"&]+)', webpage) or
-                re.search(r'OO\.Player\.create\([\'"].*?[\'"],\s*[\'"](?P<ec>\w{32})[\'"]', webpage) or
-                re.search(r'SBN\.VideoLinkset\.ooyala\([\'"](?P<ec>\w{32})[\'"]\)', webpage) or
-                re.search(r'data-ooyala-video-id\s*=\s*[\'"](?P<ec>\w{32})[\'"]', webpage) or
-                re.search(r'"ooyalaplayer-1",\s*"(?P<ec>\w{32})"', webpage))
+                re.search(r'OO\.Player\.create\([\'"].*?[\'"],\s*[\'"](?P<ec>.{32})[\'"]', webpage) or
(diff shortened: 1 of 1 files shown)


dstftw:
>Brightcove ids seems to always be alphanumeric
It's wrong. It may contain `-` and `_` and it's ooyala not brightcove.

youtube_dl/extractor/generic.py
```diff
@@ -1659,9 +1658,10 @@ def _playlist_from_matches(matches, getter=None, ie=None):
 
         # Look for Ooyala videos
         mobj = (re.search(r'player\.ooyala\.com/[^"?]+[?#][^"]*?(?:embedCode|ec)=(?P<ec>[^"&]+)', webpage) or
-                re.search(r'OO\.Player\.create\([\'"].*?[\'"],\s*[\'"](?P<ec>.{32})[\'"]', webpage) or
-                re.search(r'SBN\.VideoLinkset\.ooyala\([\'"](?P<ec>.{32})[\'"]\)', webpage) or
-                re.search(r'data-ooyala-video-id\s*=\s*[\'"](?P<ec>.{32})[\'"]', webpage))
+                re.search(r'OO\.Player\.create\([\'"].*?[\'"],\s*[\'"](?P<ec>\w{32})[\'"]', webpage) or
+                re.search(r'SBN\.VideoLinkset\.ooyala\([\'"](?P<ec>\w{32})[\'"]\)', webpage) or
+                re.search(r'data-ooyala-video-id\s*=\s*[\'"](?P<ec>\w{32})[\'"]', webpage) or
+                re.search(r'"ooyalaplayer-1",\s*"(?P<ec>\w{32})"', webpage))
```
dstftw: This regex does not look like generic embed. Provide several examples that use this embedding.

TRox1972: I was working on an extractor for pix11.com, which uses ooyala. I've added a pull request for the Pix11 extractor, but also opened this because I am unsure of what the best thing to do is: create an own extractor for Pix11, which only extracts the ooyala ID, or extend on the generic extractor.

Pix11 is the only site I've found for which the regex, works, so if there should be more than one site it works for, I'll close the pull request. Else, I'll close the Pix11 request.

Specific examples for which the regex works:
http://pix11.com/2016/05/25/beth-howland-vera-on-tv-sitcom-alice-dies-at-74/
http://pix11.com/2016/05/05/donald-trump-on-cinco-de-mayo-i-love-hispanics/
http://pix11.com/2016/05/24/no-place-to-call-home-homeless-born-into-the-system/

dstftw: Generic extractor is for generic embed templates used across multiple sites.

PR #9593, opened 2016-05-23
[xhamster] url regex fix for videos with empty title.

mexican-porn-commits:
apparently xhamster allows for videos with an empty title, so the url turns out like url/.html
this made the regex (and therefore the download) to fail, so i just made a quick fix for that.


dstftw:
Cherry-picked in a4690b3244a42a833146c406e622c96045b23df5, thanks.

PR #9595, opened 2016-05-24
[vlive] Address site update

Kagami:
Changes:
* Fix video params extraction
* Don't make status request since status info now available on the page
* Remove unneeded code
* Fix test

### DIFF
--- a/youtube_dl/extractor/vlive.py
+++ b/youtube_dl/extractor/vlive.py
@@ -1,8 +1,7 @@
 # coding: utf-8
-from __future__ import division, unicode_literals
+from __future__ import unicode_literals
 
 import re
-import time
 
 from .common import InfoExtractor
 from ..utils import (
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/vlive.py
```diff
@@ -61,16 +47,14 @@ def _real_extract(self, url):
...
-                status = 'LIVE_END'
             else:
                 status = 'COMING_SOON'
 
         if status == 'LIVE_END':
             raise ExtractorError('Uploading for replay. Please wait...',
                                  expected=True)
         elif status == 'COMING_SOON':
-            raise ExtractorError('Coming soon! %s' % air_start, expected=True)
+            raise ExtractorError('Coming soon!', expected=True)
```
yan12125: Seems ```onAirStartAt``` is available via ```/video/status``` only. I guess it's useful for others, but I can't speak for others as I don't use ```VLiveIE``` myself. Would you like to call ```/video/status``` if the status is ```'COMING_SOON'```?

Kagami: Extra request and 12 lines of code (including timezone offset cookie) isn't worth it in my opinion.

yan12125:
Thanks!

PR #9598, opened 2016-05-24
[blick] Add new extractor (blick.ch is a swiss newspaper platform, wh…

goggle:
…ich provides also videos)

### DIFF
--- a/youtube_dl/extractor/blick.py
+++ b/youtube_dl/extractor/blick.py
@@ -84,10 +84,6 @@ class BlickIE(InfoExtractor):
 
         self._sort_formats(entry_info_dict['formats'])
 
-        # Remove entries containing a url to an index.m3u8 file
-        cleaned_formats = [x for x in entry_info_dict['formats'] if x.get('format_id') != 'meta']
-        entry_info_dict['formats'] = cleaned_formats
-
         duration_found = False
         duration = None
         attr = ''
--- a/youtube_dl/extractor/blick.py
+++ b/youtube_dl/extractor/blick.py
@@ -1,0 +1,123 @@
+# encoding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+import re
+
+
+class BlickIE(InfoExtractor):
+    _VALID_URL = r'(?:https?://)?(?:www\.)?blick\.ch/.*-id(?P<id>\d+).*\.html'
+
+    _TESTS = [{
+        'url': 'http://www.blick.ch/sport/uli-forte-vor-dem-abstiegs-showdown-ich-gehe-davon-aus-dass-der-fussball-gott-fcz-fan-ist-id5070813.html',
+        'info_dict': {
+            'id': '5070813',
+            'ext': 'mp4',
+            'title': 'uli-forte-vor-dem-abstiegs-showdown-ich-gehe-davon-aus-dass-der-fussball-gott-fcz-fan-ist',
+            'thumbnail': 'http://blick.simplex.tv/content/51/52/70062/simvid_1.jpg',
+            'description': 'Am Mittwochabend entscheidet sich, ob der FCZ oder der FC Lugano aus der Super League absteigt. Uli Forte schwört dabei auf den Fussball-Gott und zündet in der Kirche eine Kerze an.'
+        }
+    }, {
+        'url': 'http://www.blick.ch/sport/tennis/nominiert-fuer-musik-preis-in-schweden-so-toll-singt-guenthardts-tochter-alessandra-id5066863.html',
+        'info_dict': {
+            'id': '5066863',
+            'ext': 'mp4',
+            'title': 'nominiert-fuer-musik-preis-in-schweden-so-toll-singt-guenthardts-tochter-alessandra',
+            'thumbnail': 'http://f.blick.ch/img/incoming/crop5066860/5146024130-csquare-w300-h300/Bildschirmfoto-2016-05-23-um-14.jpg',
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/blick.py
```diff
@@ -0,0 +1,123 @@
...
+            'description': 'Da ist Papa Heinz mächtig stolz. Seine Tochter Alessandra Günthardt ist für einen schwedischen Musik-Preis unter den drei Nominierten. Die Abstimmung läuft noch bis 7. Juni.'
+        }
+    }]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('id')
+        webpage = self._download_webpage(url, video_id)
+
+        found_videos_og = re.findall(r'<meta.*?property="og:video".*?content="(.*?)"', webpage)
```
TRox1972: Does [`_og_search_property`](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L723-L729) work?

yan12125: [```_og_search_video_url```](https://github.com/rg3/youtube-dl/blob/7b2fcbf/youtube_dl/extractor/common.py#L740-L744) may be even better.

goggle: I'm using now `_of_regexes` to do this.

youtube_dl/extractor/blick.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('id')
+        webpage = self._download_webpage(url, video_id)
+
+        found_videos = []
+        regex_og = self._og_regexes('video')
+        regex_ogs = self._og_regexes('video:secure_url')
+        video_og = self._html_search_regex(regex_og, webpage, name=None, default=None, fatal=False)
```
dstftw: `default=None` implies `fatal=False`.

youtube_dl/extractor/blick.py
```diff
@@ -0,0 +1,138 @@
...
+            'ext': 'mp4',
+            'title': 'totomat-fehler-in-sion-fcz-buff-stinksauer-wegen-falschem-lugano-resultat',
+            'thumbnail': 'http://f.blick.ch/img/incoming/crop5063475/820602933-csquare-w300-h300/Bildschirmfoto-2016-05-22-um-19.jpg',
+            'description': 'Der FC Zürich bleibt das Schlusslicht der Raiffeisen Super League. Einen dicken Hals bekommen Buff und Co. aber wegen einer falschen Resultatanzeige aus dem Ländle.',
+        }
+    }]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('id')
```
dstftw: Use `self._match_id` instead.

youtube_dl/extractor/blick.py
```diff
@@ -0,0 +1,138 @@
...
+                video_url = video
+                break
+
+        if not video_url:
+            return []
+
+        video_title = str(url)
+        b_ind = video_title.rfind('/') + 1
+        e_ind = video_title.rfind('-id')
+        video_title = video_title[b_ind:e_ind]
```
dstftw: Extract human readable title from the `webpage`.

youtube_dl/extractor/blick.py
```diff
@@ -0,0 +1,138 @@
...
+            'thumbnail': thumbnail,
+            'duration': None,
+        }
+        entry_info_dict['formats'] = self._extract_m3u8_formats(
+            video_url,
+            video_id,
+            ext='mp4',
+            entry_protocol='m3u8_native')
+
+        if entry_info_dict.get('formats'):
```
dstftw: This check does not make any sense. If there are no formats extraction should stop immediately.

youtube_dl/extractor/blick.py
```diff
@@ -0,0 +1,138 @@
...
+            self._sort_formats(entry_info_dict['formats'])
+
+            # Remove entries containing a url to an index.m3u8 file
+            cleaned_formats = [x for x in entry_info_dict['formats'] if x.get('format_id') != 'meta']
+            entry_info_dict['formats'] = cleaned_formats
```
dstftw: These formats should not be removed.

youtube_dl/extractor/blick.py
```diff
@@ -0,0 +1,138 @@
...
+        video_og = self._html_search_regex(regex_og, webpage, name=None, default=None, fatal=False)
+        video_ogs = self._html_search_regex(regex_ogs, webpage, name=None, default=None, fatal=False)
+        video_meta = self._html_search_meta('contentURL', webpage, fatal=False, default=None)
+        for elem in [video_og, video_ogs, video_meta]:
+            if elem:
+                found_videos.append(elem)
+
+        video_url = ''
+        for video in found_videos:
+            if re.match(r'.*detect\.mp4', video):
```
dstftw: Direct URLs should also be extracted.

goggle: The problem is, that _html_search_regex only gives me the first found result, which normally contains the url to a file called "detect.mp4". But this file is not a real mp4 file like the name suggests, so I cannot use this file. I need the url to a file called "index.m3u8". I could write my own regular expression to find this file, but yan12125 suggested to use the builtin function _og_regexes, which I did...

So what should I do here?

PR #9600, opened 2016-05-24
[eporner] fix for the new URL layout

wankerer:
Recently eporner slightly changed the URL layout, the ID that used to
be digits only are now digits and letters, so youtube-dl falls back to the
generic extractor that doesn't work.

Fix the matching regex to allow letters in ID.

### DIFF
--- a/youtube_dl/extractor/eporner.py
+++ b/youtube_dl/extractor/eporner.py
@@ -11,8 +11,8 @@ from ..utils import (
 
 
 class EpornerIE(InfoExtractor):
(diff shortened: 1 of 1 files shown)


dstftw:
Add a test.

wankerer:
Commit updated, added a test:

$ nosetests -v test/test_download.py -m porner
test_Eporner (test.test_download.TestDownload) ... ok
test_Eporner_1 (test.test_download.TestDownload) ... ok

Ran 2 tests in 2.139s

OK

yan12125:
Thanks! Do you want to be listed in https://github.com/rg3/youtube-dl/blob/master/AUTHORS? If so under what name?

PR #9602, opened 2016-05-25
[tvp] Add support for TVP Wiadomości

JakubAdamWieczorek:
This is the main state-funded daily news programme in Poland.

### DIFF
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -833,7 +833,7 @@ from .tvc import (
 )
 from .tvigle import TvigleIE
 from .tvland import TVLandIE
-from .tvp import TvpIE, TvpSeriesIE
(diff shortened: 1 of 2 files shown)


dstftw:
Instead of separating extractor just expand `_VALID_URL` of `TvpIE`.

JakubAdamWieczorek:
@dstftw Thanks for taking a look. I did try that at first but could not come up with one expression to handle both the wiadomosci.tvp.pl URLs and the other ones. Notice that in the former the video ID comes at first, in the latter it comes at the end:

    http://wiadomosci.tvp.pl/25169746/24052016-1200
    http://www.tvp.pl/there-can-be-anything-so-i-shortened-it/17916176

If you have an idea, it'd be appreciated.

JakubAdamWieczorek:
Thanks! :-)

PR #9615, opened 2016-05-28
[GoogleDrive] support downloading original video file

tellowkrinkle:
Adds the original video to the format list in the google drive downloader (the actual file that was uploaded, rather than one that was reencoded by Google).  Useful for playing videos from google drive using mpv.

Also updates the GoogleDrive test cases, which were probably broken by a google drive encoding change.


remitamine:
supported since v2017.09.02(fea82c1780cc751267fd2b9b4145996bfc0c1994).

PR #9620, opened 2016-05-28
Add support for defining subtitle language preference order

weinimo:
Previously defining preference order as described in
https://github.com/rg3/youtube-dl#format-selection was not possible
for the --sub-lang option. This commit adds support for this feature.

I've implemented this feature by adding a new helper method
_get_first_match(). I hope it can be reused at other locations as well
in order to simplify the often very complex code.


weinimo:
Should I close this? Is there really no interest in this PR?

yan12125:
This PR is useful. It's not examined yet because we're examining some more older PRs...

PR #9663, opened 2016-06-01
Issue template: Add syntax highlighting

rugk:
:smile:


dstftw:
This does not look any better.

PR #9669, opened 2016-06-02
Added sanitization support for Hungarian letters Ő and Ű

bzc6p:
The recently introduced support for replacing accented characters with non-accented ones covers Hungarian language almost completely, missed only these letters.

A test case would be (with --restrict-filenames, of course): https://www.youtube.com/watch?v=XygJiqezhU8


yan12125:
Thanks!

PR #9720, opened 2016-06-07
[vlive] Acknowledge vlive+ streams statuses

Kagami:
Same as common statuses just with "PRODUCT_" prefix:
PRODUCE_LIVE_END, PRODUCT_COMING_SOON, etc.

### DIFF
--- a/youtube_dl/extractor/vlive.py
+++ b/youtube_dl/extractor/vlive.py
@@ -9,6 +9,7 @@ from ..utils import (
     ExtractorError,
     float_or_none,
     int_or_none,
+    remove_start,
 )
 from ..compat import compat_urllib_parse_urlencode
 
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/vlive.py
```diff
@@ -40,6 +40,9 @@ def _real_extract(self, url):
         status, _, _, live_params, long_video_id, key = re.split(
             r'"\s*,\s*"', video_params)[2:8]
 
+        if status.startswith('PRODUCT_'):
+            status = status[8:]
```
dstftw: Use `utils.remove_start` instead.

Kagami: Fixed.

PR #9736, opened 2016-06-09
[openload] Added support for links starting with an underscore

N1k145:
Links from openload can start whith an underscore

### DIFF
--- a/youtube_dl/extractor/openload.py
+++ b/youtube_dl/extractor/openload.py
@@ -14,7 +14,7 @@ from ..utils import (
 
 
 class OpenloadIE(InfoExtractor):
-    _VALID_URL = r'https://openload.(?:co|io)/(?:f|embed)/(?P<id>[a-zA-Z0-9-]+)'
+    _VALID_URL = r'https://openload.(?:co|io)/(?:f|embed)/(?P<id>[a-zA-Z0-9-\_]+)'
 
     _TESTS = [{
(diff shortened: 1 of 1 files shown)


yan12125:
An example URL?

N1k145:
For example https://openload.co/f/_-ztPaZtMhM/ but you find many more by googleing "openload.co/f/_"

yan12125:
Good catch. Please add it to ```_TESTS``` with ```'only_matching': True```.

youtube_dl/extractor/openload.py
```diff
@@ -14,7 +14,7 @@
 
 
 class OpenloadIE(InfoExtractor):
-    _VALID_URL = r'https://openload.(?:co|io)/(?:f|embed)/(?P<id>[a-zA-Z0-9-]+)'
+    _VALID_URL = r'https://openload.(?:co|io)/(?:f|embed)/(?P<id>[a-zA-Z0-9-\_]+)'
```
yan12125: _ is not a special character in Python's regular expressions. There's no need to escape it.

TRox1972:
@N1k145 you could simplify the regular expression by using `\w-` instead

dstftw:
Cherry-picked in 21efee5f8bc8daf0cbb5fc3408a1fc5b9d5eadcb, thanks.

Issue #9073, opened 2016-04-04
skipping video download also skips non-video post-processing actions


cemysce:
### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2016.04.01*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.04.01**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add `-v` flag to **your command line** you run youtube-dl with, copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):
```
$ youtube-dl -v --skip-download --write-sub --convert-subs srt --all-subs https://www.youtube.com/watch?v=sLEdi49XEhA
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'--skip-download', u'--write-sub', u'--convert-subs', u'srt', u'--all-subs', u'https://www.youtube.com/watch?v=sLEdi49XEhA']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.04.01
[debug] Python version 2.7.6 - Linux-3.13.0-83-generic-x86_64-with-Ubuntu-14.04-trusty
[debug] exe versions: none
[debug] Proxy map: {}
[youtube] sLEdi49XEhA: Downloading webpage
[youtube] sLEdi49XEhA: Downloading video info webpage
[youtube] sLEdi49XEhA: Extracting video information
[youtube] sLEdi49XEhA: Downloading MPD manifest
[...]

PR #9738, opened 2016-06-09
Postprocessor dependencies

yan12125:
Try to solve #9073 by introducing "Download targets" and check them in `YoutubeDL.post_process()`.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -48,6 +48,7 @@ from .utils import (
     determine_ext,
     determine_protocol,
     DownloadError,
+    DownloadTarget,
     encode_compat_str,
     encodeFilename,
     error_to_compat_str,
@@ -1462,6 +1463,7 @@ class YoutubeDL(object):
         self._num_downloads += 1
 
         info_dict['_filename'] = filename = self.prepare_filename(info_dict)
+        info_dict['_downloaded_targets'] = []
 
         # Forced printings
         if self.params.get('forcetitle', False):
@@ -1541,6 +1543,9 @@ class YoutubeDL(object):
             # that way it will silently go on when used with unsupporting IE
             subtitles = info_dict['requested_subtitles']
             ie = self.get_info_extractor(info_dict['extractor_key'])
+
+            all_subtitles_ok = True
+
             for sub_lang, sub_info in subtitles.items():
                 sub_format = sub_info['ext']
                 if sub_info.get('data') is not None:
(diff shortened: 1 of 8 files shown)


youtube_dl/YoutubeDL.py
```diff
@@ -1563,8 +1569,12 @@ def process_info(self, info_dict):
                             subfile.write(sub_data)
                 except (OSError, IOError):
                     self.report_error('Cannot write subtitles file ' + sub_filename)
+                    all_subtitles_ok = False
                     return
```
yan12125: This ```return``` is introduced in a0432a1e8086e2318b122744c7afa5145abdf663. I guess failures in downloading subtitles should be non-fatal, just like thumbnails?

youtube_dl/YoutubeDL.py
```diff
@@ -48,6 +48,7 @@
     determine_ext,
     determine_protocol,
     DownloadError,
+    DownloadTarget,
```
dstftw: Probably an `asset` fits better to denote the notion.

youtube_dl/YoutubeDL.py
```diff
@@ -1462,6 +1463,7 @@ def process_info(self, info_dict):
         self._num_downloads += 1
 
         info_dict['_filename'] = filename = self.prepare_filename(info_dict)
+        info_dict['_downloaded_targets'] = []
```
dstftw: It's probably makes sense to remove it from resulting output JSON.

youtube_dl/__init__.py
```diff
@@ -252,7 +252,7 @@ def parse_retries(retries):
             'key': 'EmbedThumbnail',
             'already_have_thumbnail': already_have_thumbnail
         })
-        if not already_have_thumbnail:
+        if not already_have_thumbnail and not opts.skip_download:
```
dstftw: This changes the behavior of the following use case:
1. Media was downloaded previously (file exists).
2. youtube-dl is run with `--embed-thumbnail --skip-download`.
3. No thumbnail is downloaded and embedded (it does both with current code).
Is this intentional to match `--embed-thumbnail --skip-download` behavior?

youtube_dl/YoutubeDL.py
dstftw: Sounds reasonable.

youtube_dl/YoutubeDL.py
yan12125: Did you mean naming the class as ```Asset```?

youtube_dl/__init__.py
yan12125: Thanks. Will change this.

PR #9744, opened 2016-06-10
[indavideo] Fix extractor

nyuszika7h:
No idea why I had this redundant code at all.

### DIFF
--- a/youtube_dl/extractor/indavideo.py
+++ b/youtube_dl/extractor/indavideo.py
@@ -51,16 +51,9 @@ class IndavideoEmbedIE(InfoExtractor):
             video_urls.append(video_file)
         video_urls = list(set(video_urls))
 
-        video_prefix = video_urls[0].rsplit('/', 1)[0]
-
-        for flv_file in video.get('flv_files', []):
-            flv_url = '%s/%s' % (video_prefix, flv_file)
-            if flv_url not in video_urls:
-                video_urls.append(flv_url)
-
         formats = [{
             'url': video_url,
-            'height': self._search_regex(r'\.(\d{3,4})\.mp4$', video_url, 'height', default=None),
+            'height': int_or_none(self._search_regex(r'\.(\d+)\.mp4', video_url, 'height', default=None)),
         } for video_url in video_urls]
         self._sort_formats(formats)
 


youtube_dl/extractor/indavideo.py
```diff
@@ -60,8 +60,12 @@ def _real_extract(self, url):
 
         formats = [{
             'url': video_url,
-            'height': self._search_regex(r'\.(\d{3,4})\.mp4$', video_url, 'height', default=None),
+            'height': int_or_none(self._search_regex(r'\.(\d{3,4})\.mp4$', video_url, 'height', default=None)),
         } for video_url in video_urls]
+        
+        # Filter out redundant null height format
+        formats = [fmt for fmt in formats if fmt['height']]
```
dstftw: It should not be included in `formats` at all rather than filtering.

nyuszika7h: Oh, I didn't check carefully enough, the URL does have the height as it's the same apart from the query string, but the regex match is returning None since it's anchored to the end.

nyuszika7h: Looks like all this FLV stuff (which is what the API calls them but they are MP4s) is redundant, since it already provides the URLs in `video_url`. I have no idea why I did that, but getting rid of that seems to solve the problem.

youtube_dl/extractor/indavideo.py
```diff
@@ -51,16 +51,9 @@ def _real_extract(self, url):
             video_urls.append(video_file)
         video_urls = list(set(video_urls))
 
-        video_prefix = video_urls[0].rsplit('/', 1)[0]
-
-        for flv_file in video.get('flv_files', []):
-            flv_url = '%s/%s' % (video_prefix, flv_file)
-            if flv_url not in video_urls:
-                video_urls.append(flv_url)
-
```
dstftw: With this you drop 720p.

nyuszika7h: Oh, I see, was that code added later? I don't remember testing with HD.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #9758, opened 2016-06-11
"MiTele" May not have a title

DarkZeros:
Example episode:
http://www.mitele.es/programas-tv/cuarto-milenio/temporada-6/programa-226/

Has no title, and trying to download just gives a crash error. The error
is not there because something actually failed, but just because the
video has no title. Setting it to NO_TITLE default makes it work
properly, and only points out that it is missing in a download list.

### DIFF
--- a/youtube_dl/extractor/mitele.py
+++ b/youtube_dl/extractor/mitele.py
@@ -15,7 +15,7 @@ class MiTeleIE(InfoExtractor):
     IE_DESC = 'mitele.es'
     _VALID_URL = r'https?://www\.mitele\.es/[^/]+/[^/]+/[^/]+/(?P<id>[^/]+)/'
 
-    _TEST = {
+    _TESTS = [{
         'url': 'http://www.mitele.es/programas-tv/diario-de/la-redaccion/programa-144/',
         # MD5 is unstable
         'info_dict': {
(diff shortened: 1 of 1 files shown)


dstftw:
Title can be constructed from
```
				<h1>
					<span>Cuarto Milenio</span>
					<span>Temporada 6</span>
					<span>Programa 226</span>
				</h1>
```
Also add a test.

DarkZeros:
Ok, I am not a Python programmer, but let me guess it out and update this commit with a proper test.

What do you mean by:

>                 <h1>
>                     <span>Cuarto Milenio</span>
>                     <span>Temporada 6</span>
>                     <span>Programa 226</span>
>                 </h1>

?

dstftw:
This is the excerpt from HTML source near the original title that is missing for this video.
To add a test rename `_TEST` to `_TESTS` and wrap in a list (see any other extractor).

DarkZeros:
Updated with a test and a proper default value.
Hope this changes are OK, I tested them myself.

It is very easy to develop here!
I may add another site that is not supported now that I know how.

dstftw:
I would prefer extracting it from place I've pointed out. Also squash commits when done.

PR #9774, opened 2016-06-13
[YoutubeDL] Fix wrong hanging identation

stepshal:
From pylint.


dstftw:
This breaks PEP 8:
```
.\youtube_dl\YoutubeDL.py:1755:5: E129 visually indented line with same indent as next logical line
```

PR #9778, opened 2016-06-13
[nick.de] add new extractor

leopard91:
add support for nick.de

tested on 2.7.11 and 3.5.1

### DIFF
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -513,6 +513,7 @@ from .nhl import (
     NHLIE,
 )
 from .nick import NickIE
+from .nickde import NickDeIE
 from .niconico import NiconicoIE, NiconicoPlaylistIE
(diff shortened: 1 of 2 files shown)


dstftw:
I'm almost sure you've just reinvented the wheel. Extractor for mtvnservices is already implemented and [nick.com](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/nick.py) in implemented in terms of it. nick.de should be implemeted similarly.

leopard91:
hi dstftw, the german site of nick.com have a total different structur, i have looked at the implemention for nick.com and mtvnservices but my opinion was it create much extra work to force subclassing the existing mtv classes without breaking nick.com

dstftw:
There are 10+ extractors that based on `MTVServicesInfoExtractor` that co-exists without any problem or breakage.
Brief look at nick.de confirms that it uses the same feed based content delivery.

PR #9781, opened 2016-06-14
Added note about installing avconv for Mac users

NetOperatorWibby:


yan12125:
As per #8622, ffmpeg is recommended over libav. Does homebrew provide ffmpeg?

NetOperatorWibby:
I have no idea, but I found that command immediately and it worked.

On Tue, Jun 14, 2016 at 12:32 AM, Yen Chi Hsuan <notifications@github.com> wrote:
As per #8622 [https://github.com/rg3/youtube-dl/issues/8622] , ffmpeg is recommended over libav. Does homebrew provide ffmpeg?

yan12125:
Seems they have ffmpeg: https://github.com/Homebrew/homebrew-core/blob/master/Formula/ffmpeg.rb

PR #9813, opened 2016-06-16
[polskieradio] Add support for Polskie Radio.

JakubAdamWieczorek:
Polskie Radio is the main Polish state-funded radio broadcasting service.

### DIFF
--- a/youtube_dl/extractor/polskieradio.py
+++ b/youtube_dl/extractor/polskieradio.py
@@ -1,0 +1,74 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import int_or_none
+
+import calendar
+from datetime import datetime
+
+
+class PolskieRadioIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?polskieradio\.pl/[0-9]+/[0-9]+/Artykul/(?P<id>[0-9]+),.+'
+    _TESTS = [{
+        'url': 'http://www.polskieradio.pl/7/5102/Artykul/1587943,Prof-Andrzej-Nowak-o-historii-nie-da-sie-myslec-beznamietnie',
+        'md5': '2984ee6ce9046d91fc233bc1a864a09a',
+        'info_dict': {
+            'id': '1587943',
+            'ext': 'mp3',
+            'title': 'Prof. Andrzej Nowak: o historii nie da się myśleć beznamiętnie',
+            'description': 'md5:12f954edbf3120c5e7075e17bf9fc5c5',
+            'release_date': '20160227',
+            'upload_date': '20160227',
+            'timestamp': 1456594200,
+            'duration': 2364
+        }
+    }, {
+        'url': 'http://polskieradio.pl/9/305/Artykul/1632955,Bardzo-popularne-slowo-remis',
+        'md5': '68a393e25b942c1a76872f56d303a31a',
+        'info_dict': {
+            'id': '1632955',
+            'ext': 'mp3',
+            'title': 'Bardzo popularne słowo: remis',
+            'description': 'md5:3b58dfae614100abc0f175a0b26d5680',
+            'release_date': '20160617',
+            'upload_date': '20160617',
+            'timestamp': 1466184900,
+            'duration': 393
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        metadata_string = self._html_search_regex(r'<span class="play pr-media-play" data-media=(\{.+\})>', webpage, 'metadata')
+        metadata = self._parse_json(metadata_string, video_id)
+
+        title = self._og_search_title(webpage)
+        if title is not None:
+            title = title.strip()
+
+        description = self._og_search_description(webpage)
+        if description is not None:
+            description = description.strip()
+
+        release_date = self._html_search_regex(r'Data emisji:[^0-9]+([0-9]{1,2}\.[0-9]{2}\.[0-9]{4})', webpage, 'release date', fatal=False)
+        if release_date is not None:
+            release_date = datetime.strptime(release_date, '%d.%m.%Y').strftime('%Y%m%d')
+
+        upload_datetime = self._html_search_regex(r'<span id="datetime2" class="time">\s+(.+)\s+</span>', webpage, 'release time', fatal=False)
+        if upload_datetime is not None:
+            timestamp = calendar.timegm(datetime.strptime(upload_datetime, '%d.%m.%Y %H:%M').timetuple())
+        else:
+            timestamp = None
+
+        return {
+            'id': video_id,
+            'title': title,
+            'description': description,
+            'display_id': metadata.get('id'),
+            'duration': int_or_none(metadata.get('length')),
+            'url': self._proto_relative_url(metadata.get('file'), 'http:'),
+            'release_date': release_date,
+            'timestamp': timestamp
+        }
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/polskieradio.py
```diff
@@ -0,0 +1,44 @@
...
+        metadata_string = self._html_search_regex(r'data-media=\{([^\}]+)\}', webpage, 'metadata')
+        metadata = json.loads('{' + metadata_string + '}')
+
+        return {
+            'id': video_id,
+            'title': self._og_search_title(webpage).strip(),
+            'description': self._og_search_description(webpage).strip(),
+            'display_id': metadata.get("id"),
+            'duration': metadata.get("length"),
+            'url': "http:" + metadata["file"]
```
TRox1972: Use `self._proto_relative_url` instead.

youtube_dl/extractor/polskieradio.py
```diff
@@ -0,0 +1,44 @@
...
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        metadata_string = self._html_search_regex(r'data-media=\{([^\}]+)\}', webpage, 'metadata')
+        metadata = json.loads('{' + metadata_string + '}')
+
+        return {
+            'id': video_id,
+            'title': self._og_search_title(webpage).strip(),
+            'description': self._og_search_description(webpage).strip(),
+            'display_id': metadata.get("id"),
```
TRox1972: Single quotes consistency

youtube_dl/extractor/polskieradio.py
```diff
@@ -0,0 +1,44 @@
...
+            'ext': 'mp3',
+            'title': 'Wokół "Historyji o Chwalebnym Zmartwychwstaniu Pańskim"',
+            'description': '„Wokół Historyji o Chwalebnym Zmartwychwstaniu Pańskim" - audycja Iwony Malinowskiej i Bogumiły Prządki z udziałem historyka literatury - profesora Krzysztofa Mrowcewicza, historyka teatru - profesora Andrzeja Kruczyńskiego, reżysera teatralnego, który wystawiał misterium Mikołaja z Wilkowiecka na scenie - Piotra Cieplaka i reżysera radiowego - Zdzisława Dąbrowskiego, który zrealizował ten utwór dla radia w 1977 roku.',
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        metadata_string = self._html_search_regex(r'data-media=\{([^\}]+)\}', webpage, 'metadata')
```
TRox1972: Expand the group to include the two curly braces. That way you can just do `json.loads(metadata_string)` below.

youtube_dl/extractor/polskieradio.py
```diff
@@ -0,0 +1,44 @@
...
+            'description': '- O polskiej historii nie da się myśleć bez wchodzenia w ten emocjonujący spór, co było dobre, a co złe, co było błędem, a co było słuszną decyzją - mówi profesor Andrzej Nowak.',
+        }
+    }, {
+        'url': 'http://polskieradio.pl/8/295/Artykul/213411,Wokol-Historyji-o-Chwalebnym-Zmartwychwstaniu-Panskim',
+        'md5': 'e925764839c67820b4fbdcd7e2785c08',
+        'info_dict': {
+            'id': '213411',
+            'ext': 'mp3',
+            'title': 'Wokół "Historyji o Chwalebnym Zmartwychwstaniu Pańskim"',
+            'description': '„Wokół Historyji o Chwalebnym Zmartwychwstaniu Pańskim" - audycja Iwony Malinowskiej i Bogumiły Prządki z udziałem historyka literatury - profesora Krzysztofa Mrowcewicza, historyka teatru - profesora Andrzeja Kruczyńskiego, reżysera teatralnego, który wystawiał misterium Mikołaja z Wilkowiecka na scenie - Piotra Cieplaka i reżysera radiowego - Zdzisława Dąbrowskiego, który zrealizował ten utwór dla radia w 1977 roku.',
```
TRox1972: Use `'md5:'` instead

youtube_dl/extractor/polskieradio.py
TRox1972: Expand regex-group to include the two curly braces, to avoid the awkward 'plus' statement below. And use `self._parse_json()` instead of `json.loads()`.

It's better to parse the regex into `metadata` via `self._parse_json()` instead of creating an own variable for it, since you're using it only once.

JakubAdamWieczorek:
@TRox1972, thank you for the review. I have addressed your concerns.

youtube_dl/extractor/polskieradio.py
```diff
@@ -0,0 +1,42 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class PolskieRadioIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?polskieradio\.pl/[0-9]+/[0-9]+/Artykul/(?P<id>[0-9]+),.*'
```
dstftw: `.*` at the end makes no sense.

youtube_dl/extractor/polskieradio.py
```diff
@@ -0,0 +1,42 @@
...
+        webpage = self._download_webpage(url, video_id)
+        metadata_string = self._html_search_regex(r'data-media=(\{[^\}]+\})', webpage, 'metadata')
+        metadata = self._parse_json(metadata_string, video_id)
+
+        return {
+            'id': video_id,
+            'title': self._og_search_title(webpage).strip(),
+            'description': self._og_search_description(webpage).strip(),
+            'display_id': metadata.get('id'),
+            'duration': metadata.get('length'),
```
dstftw: Should be wrapped in `int_or_none`.

youtube_dl/extractor/polskieradio.py
```diff
@@ -0,0 +1,42 @@
...
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        metadata_string = self._html_search_regex(r'data-media=(\{[^\}]+\})', webpage, 'metadata')
+        metadata = self._parse_json(metadata_string, video_id)
+
+        return {
+            'id': video_id,
+            'title': self._og_search_title(webpage).strip(),
+            'description': self._og_search_description(webpage).strip(),
```
dstftw: Breaks the extraction if description is `None`.

youtube_dl/extractor/polskieradio.py
JakubAdamWieczorek: Why?

youtube_dl/extractor/polskieradio.py
```diff
@@ -0,0 +1,42 @@
...
+            'ext': 'mp3',
+            'title': 'Wokół "Historyji o Chwalebnym Zmartwychwstaniu Pańskim"',
+            'description': 'md5:bfffb8a938ef087c5f3b261c0f4e9593',
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        metadata_string = self._html_search_regex(r'data-media=(\{[^\}]+\})', webpage, 'metadata')
```
dstftw: This will capture invalid JSON if it contains inner `{}`.

youtube_dl/extractor/polskieradio.py
dstftw: Cause `*` means zero or more characters allowing string to end right at `,`.

JakubAdamWieczorek:
@dstftw, thank you for the review. I have made the suggested changes.

JakubAdamWieczorek:
Is this good to go now?

JakubAdamWieczorek:
@dstftw, @TRox1972 – any objections?

JakubAdamWieczorek:
Could we move further with this, please? It would be much appreciated.

JakubAdamWieczorek:
Thank you. :)

PR #9827, opened 2016-06-18
Unlicense

TRox1972:
It's recommended by http://unlicense.org to create an `UNLICENSE` fie with the 'license' so it'll be easier for people to find public domain projects like this one.


phihag:
The widely supported de-facto standard is `LICENSE`, even if it is the license called "unlicense". To find out whether a project is put in the public domain, simply match the contents of `LICENSE` against a list of known public domain licenses.

PR #9828, opened 2016-06-18
Added UNLICENSE file

TRox1972:
It's recommended by http://unlicense.org to create an UNLICENSE fie with the 'license' so it'll be easier for people to find public domain projects like this one.

### DIFF
--- a/UNLICENSE
+++ b/UNLICENSE
@@ -1,0 +1,24 @@
+This is free and unencumbered software released into the public domain.
+
+Anyone is free to copy, modify, publish, use, compile, sell, or
+distribute this software, either in source code form or as a compiled
+binary, for any purpose, commercial or non-commercial, and by any
+means.
+
+In jurisdictions that recognize copyright laws, the author or authors
+of this software dedicate any and all copyright interest in the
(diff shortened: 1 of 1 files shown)


dstftw:
It's already [created](https://github.com/rg3/youtube-dl/blob/master/LICENSE).

TRox1972:
Yes, but it's called 'LICENSE', but public domain isn't really a license. Creating a new file called 'UNLICENSE' would maybe make it easier for people searching for public domain software projects.

dstftw:
>make it easier for people searching for public domain software projects

How's that? Searching by license text works much better than search for filename.

TRox1972:
From their website

> You would traditionally put the above statement into a file named COPYING or LICENSE. However, to explicitly distance yourself from the whole concept of copyright licensing, we recommend that you put your unlicensing statement in a file named UNLICENSE. Doing so also means that your project can more easily be found on e.g. GitHub or Bitbucket, enabling others to reuse your code in their own unencumbered public domain projects.

yan12125:
> make it easier for people searching for public domain software projects

Well, I guess most people here was trying to find a video downloader but not to find a public domain project.

PR #9829, opened 2016-06-18
Avoid double key lookup for setting new key

lucasmoura:
In order to add a new key to both __objects and __functions dicts on jsinterp.py, it is
necessary to first verify if a key was present and if not, create the key and
assign it to a value.

However, this can be done with a single step using dict setdefault method.

### DIFF
--- a/youtube_dl/jsinterp.py
+++ b/youtube_dl/jsinterp.py
@@ -131,9 +131,8 @@ class JSInterpreter(object):
             if variable in local_vars:
                 obj = local_vars[variable]
             else:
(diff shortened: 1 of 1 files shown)


dstftw:
Can you check with flake8 and squash?

lucasmoura:
oh sorry for that, I will fix the warnings right away

lucasmoura:
@dstftw Done

dstftw:
Cherry-picked in 7c05097633138459e9bdf7e10738e021b04689a7, thanks.

PR #9832, opened 2016-06-19
[utils] Add try error type

daineseh:
Hi, I using youtube-dl to get youtube video link, and showed TypeError message.

> r Type: <type 'exceptions.TypeError'>
> Error Contents: attribute of type 'NoneType' is not callable
> Traceback (most recent call last):
>   File "addon.py", line 56, in <module>
>     info = ydl.extract_info(url)
>   File "youtube_dl/YoutubeDL.py", line 676, in extract_info
>     ie_result = ie.extract(url)
>   File "youtube_dl/extractor/common.py", line 341, in extract
>     return self._real_extract(url)
>   File "youtube_dl/extractor/youtube.py", line 1422, in _real_extract
>     upload_date = unified_strdate(upload_date)
>   File "youtube_dl/utils.py", line 1054, in unified_strdate
>     upload_date = datetime.datetime.strptime(date_str, expression).strftime('%Y%m%d')
> TypeError: attribute of type 'NoneType' is not callable

### DIFF
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -1063,7 +1063,7 @@ def unified_strdate(date_str, day_first=True):
     for expression in format_expressions:
         try:
             upload_date = datetime.datetime.strptime(date_str, expression).strftime('%Y%m%d')
-        except ValueError:
+        except (ValueError, TypeError):
             pass
     if upload_date is None:
         timetuple = email.utils.parsedate_tz(date_str)


dstftw:
**Post the full output of youtube-dl when run with `-v`**, i.e. **add** `-v` flag to **your command line**, copy the **whole** output and post it in the issue body wrapped in \`\`\` for better formatting. It should look similar to this:
```
$ youtube-dl -v <your command line>
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://www.youtube.com/watch?v=BaW_jenozKcj']
[debug] youtube-dl version 2015.12.06
[debug] Git HEAD: 135392e
[debug] Python version 2.6.6 - Windows-2003Server-5.2.3790-SP2
[debug] exe versions: ffmpeg N-75573-g1d0487f, ffprobe N-75573-g1d0487f, rtmpdump 2.4
[debug] Proxy map: {}
...
```
**Do not post screenshots of verbose log only plain text is acceptable.**

The output (including the first lines) contains important debugging information. Issues without the full output are often not reproducible and therefore do not get solved in short order, if ever.

PR #9850, opened 2016-06-21
Vimeo: When parsing vimeo user playlist, include video id in url result.

purdeaandrei:
This will allow us to decide much faster that we don't want an already archived video,
and will allow us NOT to download webpages for each video that has already been downloaded,
thus significantly speeding up the archival of channels that have no new content.


dstftw:
Cherry-picked in dfc8f46e1c0e47a3b080d2e38d7d6da279f18fd2, thanks.

Issue #9854, opened 2016-06-22
Bug report：Cannot find video formats


JackieXie168:
I use this program to download facebook video then got the following error message.

$ youtube-dl https://www.facebook.com/englishbad/videos/557269211145445/
[facebook] 557269211145445: Downloading webpage
ERROR: Cannot find video formats; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.

Note：I've already download and run the latest version of youtube-dl. It's no problem to  download facebook video yesterday, but today is not...

OS : Mac OS X - mountain lion v10.8.5

$ youtube-dl https://www.facebook.com/englishbad/videos/557269211145445/ -v
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'https://www.facebook.com/englishbad/videos/557269211145445/', u'-v']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.06.20
[debug] Python version 2.7.11 - Darwin-12.6.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg 3.0.2, ffprobe 3.0.2
[debug] Proxy map: {}
[facebook] 557269211145445: Downloading webpage
ERROR: Cannot find video formats; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/opt/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 676, in extract_info
    ie_result = ie.extract(url)
  File "/opt/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 343, in extract
    return self._real_extract(url)
  File "/opt/local/bin/youtube-dl/youtube_dl/extractor/facebook.py", line 290, in _real_extract
    webpage, info_dict = self._extract_from_url(real_url, video_id, fatal_if_no_video=False)
  File "/opt/local/bin/youtube-dl/youtube_dl/extractor/facebook.py", line 261, in _extract_from_url
    raise ExtractorError('Cannot find video formats')
[...]

PR #9856, opened 2016-06-22
Fix issue #9854 with downloading Facebook videos

shaicoleman:
youtube-dl expects the format items to be returned as a list,
but when there's only one item Facebook returns a dict instead,
this wraps the dict in a list if necessary


phihag:
Thanks! Merged.

PR #9858, opened 2016-06-22
[xnxx] fix url regex

rr-:
The pattern has changed from "video123412" to "video-o8xa19".
The changes maintain backwards compatibility with old-style URLs.

### DIFF
--- a/youtube_dl/extractor/xnxx.py
+++ b/youtube_dl/extractor/xnxx.py
@@ -6,17 +6,20 @@ from ..compat import compat_urllib_parse_unquote
 
 
 class XNXXIE(InfoExtractor):
-    _VALID_URL = r'^https?://(?:video|www)\.xnxx\.com/video(?P<id>[0-9]+)/(.*)'
-    _TEST = {
-        'url': 'http://video.xnxx.com/video1135332/lida_naked_funny_actress_5_',
-        'md5': '0831677e2b4761795f68d417e0b7b445',
+    _VALID_URL = r'^https?://(?:video|www)\.xnxx\.com/video-?(?P<id>[0-9a-z]+)/(.*)'
+    _TESTS = [{
+        'url': 'http://www.xnxx.com/video-6gqggeb/hd_star-581_sam',
+        'md5': '6a2a6aff3f10467d94e572edb7b7deb6',
         'info_dict': {
-            'id': '1135332',
+            'id': '6gqggeb',
             'ext': 'flv',
-            'title': 'lida » Naked Funny Actress  (5)',
+            'title': 'HD STAR-581 sam',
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/xnxx.py
```diff
@@ -6,14 +6,14 @@
 
 
 class XNXXIE(InfoExtractor):
-    _VALID_URL = r'^https?://(?:video|www)\.xnxx\.com/video(?P<id>[0-9]+)/(.*)'
+    _VALID_URL = r'^https?://(?:video|www)\.xnxx\.com/video-?(?P<id>[0-9a-z]+)/(.*)'
     _TEST = {
-        'url': 'http://video.xnxx.com/video1135332/lida_naked_funny_actress_5_',
+        'url': 'http://www.xnxx.com/video-43s6xfe/javso.com_-_beautiful_body_asian_softcore_idol',
```
TRox1972: change `www.` to `(?:www\.)?`

rr-: @TRox1972 Are you sure you haven't mistaken this line with `_VALID_URL` which is regex indeed?

TRox1972: @rr- Sorry, misread. You're correct.

youtube_dl/extractor/xnxx.py
```diff
@@ -6,17 +6,26 @@
...
+    _VALID_URL = r'^https?://(?:video|www)\.xnxx\.com/video-?(?P<id>[0-9a-z]+)/(.*)'
+    _TESTS = [{
+        'url': 'http://www.xnxx.com/video-43s6xfe/javso.com_-_beautiful_body_asian_softcore_idol',
+        'md5': '6675d4424033b7f27b34dc103b3ad7ae',
         'info_dict': {
+            'id': '43s6xfe',
+            'ext': 'flv',
+            'title': 'javso.com - Beautiful Body Asian Softcore Idol',
+            'age_limit': 18,
+        },
```
dstftw: One test is enough remove this and update the old one.

rr-: Please note that the old URL is still valid and reachable though

dstftw: I'm aware of that. Old format URLs are redirected to the new one. Testing both will only introduce unnecessary network I/O. Make it `only_matching` if you don't want to delete.

rr-: Done - added `only_matching`

dstftw:
Can you also replace test video with one that does not violate any copyrights?

rr-:
I hope trailers are OK

PR #9860, opened 2016-06-22
[voissa] Add new extractor

tiburce:
Add support for voissa.com site


tiburce:
No error with this extractor in travis logs.
are there any things to do ?

PR #9864, opened 2016-06-23
[RoosterTeeth] Add new extractor

nehalvpatel:

- [ ] Bug fix
- [x] New extractor
- [ ] New feature

---

Added a new extractor for all [Rooster Teeth](https://www.roosterteeth.com/) sites. In addition to the video, it extracts the description and the thumbnail. It can download both public and sponsor-only videos (login).

### DIFF
--- a/youtube_dl/extractor/roosterteeth.py
+++ b/youtube_dl/extractor/roosterteeth.py
@@ -1,0 +1,111 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    urlencode_postdata,
+)
+
+
+class RoosterTeethIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:(www|achievementhunter|funhaus|screwattack|theknow)\.)?roosterteeth\.com/episode/(?P<id>[^/?#&]+)'
+    _LOGIN_URL = 'https://roosterteeth.com/login'
+    _NETRC_MACHINE = 'roosterteeth'
+    _TESTS = [{
+        'url': 'http://roosterteeth.com/episode/million-dollars-but-season-2-million-dollars-but-the-game-announcement',
+        'info_dict': {
+            'id': '26576',
+            'ext': 'mp4',
+            'title': 'Million Dollars, But... The Game Announcement',
+            'thumbnail': 're:^https?://.*\.png$',
+            'description': 'Introducing Million Dollars, But... The Game! Available for pre-order now at www.MDBGame.com ',
+            'creator': 'Rooster Teeth',
+            'series': 'Million Dollars, But...',
+            'episode': 'Million Dollars, But... The Game Announcement',
+            'episode_id': '26576',
+        },
+        'params': {
+            'skip_download': True,  # m3u8 downloads
+        },
+    }, {
+        'url': 'http://achievementhunter.roosterteeth.com/episode/off-topic-the-achievement-hunter-podcast-2016-i-didn-t-think-it-would-pass-31',
+        'only_matching': True,
+    }, {
+        'url': 'http://funhaus.roosterteeth.com/episode/funhaus-shorts-2016-austin-sucks-funhaus-shorts',
+        'only_matching': True,
+    }, {
+        'url': 'http://screwattack.roosterteeth.com/episode/death-battle-season-3-mewtwo-vs-shadow',
+        'only_matching': True,
+    }, {
+        'url': 'http://theknow.roosterteeth.com/episode/the-know-game-news-season-1-boring-steam-sales-are-better',
+        'only_matching': True,
+    }]
+
+    def _login(self):
+        (username, password) = self._get_login_info()
+        if username is None or password is None:
+            return False
+
+        # token is required to authenticate request
+        login_page = self._download_webpage(self._LOGIN_URL, None, 'Getting login token', 'Unable to get login token')
+        hidden_inputs = self._hidden_inputs(login_page)
+        token = hidden_inputs['_token']
+
+        login_data = {
+            '_token': token,
+            'username': username,
+            'password': password,
+        }
+
+        login_payload = urlencode_postdata(login_data)
+
+        # required for proper responses
+        login_headers = {
+            'Referer': 'https://roosterteeth.com/login',
+        }
+
+        login_request = self._download_webpage(self._LOGIN_URL, None, note='Logging in as %s' % username, data=login_payload, headers=login_headers)
+
+        if 'Authentication failed' in login_request:
+            raise ExtractorError(
+                'Login failed (invalid username/password)', expected=True)
+
+    def _real_initialize(self):
+        self._login()
+
+    def _real_extract(self, url):
+        match_id = self._match_id(url)
+        webpage = self._download_webpage(url, match_id)
+
+        episode_id = self._html_search_regex(r"commentControls\('#comment-([0-9]+)'\)", webpage, 'episode id', match_id, False)
+
+        self.report_extraction(episode_id)
+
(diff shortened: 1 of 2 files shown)


nehalvpatel:
> m3u8 download detected but ffmpeg or avconv could not be found. Please install one.

Seems like there was an error with ffmpeg? It extracted the m3u8 URL just fine.

youtube_dl/extractor/roosterteeth.py
```diff
@@ -81,6 +84,6 @@ def _real_extract(self, url):
             'title': title,
             'thumbnail': self._og_search_thumbnail(webpage),
             'description': self._og_search_description(webpage),
-            'creator': self._og_search_property("site_name", webpage, "site"),
+            'creator': self._html_search_regex(r'body id="episode" class="site-([^"]*)"', webpage, "site", fatal=False),
```
TRox1972: "site" should be in single quotes

nehalvpatel: Fixed in e8b6c00567ad716d6efa558a112b25307a801720

youtube_dl/extractor/roosterteeth.py
```diff
@@ -0,0 +1,89 @@
...
+from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    sanitized_Request,
+    urlencode_postdata,
+)
+
+
+class RoosterTeethIE(InfoExtractor):
+    _VALID_URL = r'https?:\/\/(?:(.+)\.)?roosterteeth\.com\/episode\/(?P<id>[^\r\n]+)'
```
dstftw: No need to escape `/`.

youtube_dl/extractor/roosterteeth.py
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    def _login(self):
+        (username, password) = self._get_login_info()
+        if username is None or password is None:
+            return False
+
+        # token is required to authenticate request
+        login_page = self._download_webpage(self._LOGIN_URL, None, 'Getting login token', 'Unable to get login token')
+        token = self._html_search_regex(r'name="_token" type="hidden" value="([^"]*)"', login_page, 'login token')
```
dstftw: Use `self._hidden_inputs` instead.

youtube_dl/extractor/roosterteeth.py
```diff
@@ -0,0 +1,89 @@
...
+            '_token': token,
+            'username': username,
+            'password': password,
+        }
+
+        payload = urlencode_postdata(form_data)
+
+        request = sanitized_Request(self._LOGIN_URL, payload)
+        request.add_header('Referer', 'https://roosterteeth.com/login')
+        request.add_header('Content-Type', 'application/x-www-form-urlencoded')
```
dstftw: 41-51 can be passed directly to `_download_webpage`.

youtube_dl/extractor/roosterteeth.py
```diff
@@ -0,0 +1,89 @@
...
+
+        self.report_login()
```
dstftw: As well as this.

youtube_dl/extractor/roosterteeth.py
```diff
@@ -0,0 +1,89 @@
...
+                'Login failed (invalid username/password)', expected=True)
+
+    def _real_initialize(self):
+        self._login()
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        self.report_extraction(video_id)
```
dstftw: Superfluous message.

nehalvpatel: @dstftw Fixed in a22bbb1adffda4b1c04b01a7983c759544ca76ca

youtube_dl/extractor/roosterteeth.py
```diff
@@ -0,0 +1,86 @@
...
+
+from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    urlencode_postdata,
+)
+
+
+class RoosterTeethIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:(.+)\.)?roosterteeth\.com/episode/(?P<id>[^/?#&]+)'
```
dstftw: Do not capture group if you don't use it. Add `only_matching` tests for subdomains. Subdomain regex should not be greedy.

dstftw:
Squash commits when done.

nehalvpatel:
@dstftw Relaxed the regex and squashed the commits: 65855424d3117d971b087935d18bf77b1b0ee0b1

dstftw:
Not actually. You've restricted it. And still capture the group you don't use.

youtube_dl/extractor/roosterteeth.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def _login(self):
+        (username, password) = self._get_login_info()
+        if username is None or password is None:
+            return False
+
+        # token is required to authenticate request
+        login_page = self._download_webpage(self._LOGIN_URL, None, 'Getting login token', 'Unable to get login token')
+        hidden_inputs = self._hidden_inputs(login_page)
+        token = hidden_inputs['_token']
```
dstftw: I've already suggested using this whole dictionary. `_token` is already extracted you don't need to access it explicitly.

youtube_dl/extractor/roosterteeth.py
```diff
@@ -0,0 +1,111 @@
...
+            '_token': token,
+            'username': username,
+            'password': password,
+        }
+
+        login_payload = urlencode_postdata(login_data)
+
+        # required for proper responses
+        login_headers = {
+            'Referer': 'https://roosterteeth.com/login',
```
dstftw: Code duplication. Replace with `self._LOGIN_URL`.

youtube_dl/extractor/roosterteeth.py
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        login_request = self._download_webpage(self._LOGIN_URL, None, note='Logging in as %s' % username, data=login_payload, headers=login_headers)
```
dstftw: Carry long lines.

youtube_dl/extractor/roosterteeth.py
nehalvpatel: To be honest, I'm new to Python and I didn't know what you meant. Please explain.

dstftw: Add `username` and `password` to `hidden_inputs` and use it as payload.

nehalvpatel:
@dstftw Right now, it's not capturing the subdomain group, just matching it. The `?:` prevents the capture. Besides, how else can it detect all the possible subdomains?

dstftw:
Wrong. You have inner group that is captured `(www|achievementhunter|funhaus|screwattack|theknow)`.

nehalvpatel:
If you take a look [here](https://regex101.com/r/tH2tR2/1), only the episode id is matched. Unless I'm misunderstood, the `?:` marks the inner group as non-capturing.

youtube_dl/extractor/roosterteeth.py
nehalvpatel: Thanks, will do.

dstftw:
If you take a look at the string you'are matching you may notice that you have nothing to capture by your first group. Change it to `http://www.roosterteeth.com/episode/day-5-day-5-episode-1-waking-nightmare` and see.

nehalvpatel:
Oh okay I see.

How about this: https://regex101.com/r/tH2tR2/2

dstftw:
Just change to ``(?:[\da-zA-Z-]+\.)?roosterteeth\.com` or `(?:.+?\.)?roosterteeth\.com`.

nehalvpatel:
Thanks, changed it.

PR #9883, opened 2016-06-23
Update syntax highlighting

benwa:
bash and bat where appropriate


dstftw:
There is no need in syntax highlighting for command listings.

PR #9890, opened 2016-06-25
Added extractor for trailers on Comingsoon.it

MaurizioPz:


youtube_dl/extractor/comingsoon_it.py
```diff
@@ -0,0 +1,44 @@
...
+            'title': title,
+            'description': description,
+
+            'formats': [
+                {
+                    'url': 'http://video.comingsoon.it/MP4/' + video_id + '.mp4',
+                    'format': 'Standard Definition'
+                },
+                {
+                    'url': 'http://video.comingsoon.it/MP4/' + video_id + '.mp4',
```
dstftw: Download URL for both formats are equal.

MaurizioPz: Fixed

youtube_dl/extractor/comingsoon_it.py
```diff
@@ -0,0 +1,64 @@
...
+            'description': description,
+
+            'formats': [
+                {
+                    'url': 'http://video.comingsoon.it/MP4/' + video_id + '.mp4',
+                    'format': 'Standard Definition',
+                    'format_id': 'sd'
+                },
+                {
+                    'url': 'http://video.comingsoon.it/MP4/' + video_id + 'HD.mp4',
```
remitamine: not all video has HD quality and even some of them doesn't have the same url format so it better to follow the logic in the browser:
downloading the embed page which contain video urls: `http://www.comingsoon.it/videoplayer/embed/?idv=<id>`
extract `vLwRes` and `vHiRes`, if both urls are the same then only sd quality is available else both sd and hd format is available.
also don't hardcode `ext` it can be detected automaticaly from the url(some video in the web site served as flv files).

jasan-s:
@MaurizioPz  is the extractor working for comingsoon?

PR #9893, opened 2016-06-25
[imgur] Updating album resolution endpoint

okofish:
The old album endpoint, `http://imgur.com/gallery/%s/album_images/hit.json?all=true`, worked for some older albums but not recent ones. The new one, `http://imgur.com/ajaxalbums/getimages/%s/hit.json?all=true`, should work for both old and new albums.

It might be beneficial to eventually change to the [official Imgur API](http://api.imgur.com/) - it requires registration to receive a key, but we won't have to worry about unannounced API changes.

### DIFF
--- a/youtube_dl/extractor/imgur.py
+++ b/youtube_dl/extractor/imgur.py
@@ -132,7 +132,7 @@ class ImgurAlbumIE(InfoExtractor):
         album_id = self._match_id(url)
 
         album_images = self._download_json(
-            'http://imgur.com/gallery/%s/album_images/hit.json?all=true' % album_id,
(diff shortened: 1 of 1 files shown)


dstftw:
Add a test.
youtube-dl will never use any API that requires a developer key.

okofish:
Could you clarify what you mean by "add a test?" I didn't add any behavior that didn't have a test already; the pre-existing _TESTS value covers this.

dstftw:
>The old album endpoint, http://imgur.com/gallery/%s/album_images/hit.json?all=true, worked for some older albums but not recent ones

>but not recent ones

That is what you should test.

Issue #9286, opened 2016-04-23
--hls-use-mpegts makes postprocessors fail


fstirlitz:
Using the `--hls-use-mpegts` option makes the metadata postprocessor fail:

```
[ffmpeg] Adding metadata to 'blah.mp4'
[debug] ffmpeg command line: ffmpeg -y -i file:blah.mp4 -c copy -metadata purl=[REDACTED] -metadata date=[REDACTED] -metadata title=[REDACTED] file:blah.temp.mp4
ERROR: Conversion failed!
Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/youtube_dl/YoutubeDL.py", line 1765, in post_process
    files_to_delete, info = pp.run(info)
  File "/usr/lib/python3.5/site-packages/youtube_dl/postprocessor/ffmpeg.py", line 425, in run
    self.run_ffmpeg(filename, temp_filename, options)
  File "/usr/lib/python3.5/site-packages/youtube_dl/postprocessor/ffmpeg.py", line 172, in run_ffmpeg
    self.run_ffmpeg_multiple_files([path], out_path, opts)
  File "/usr/lib/python3.5/site-packages/youtube_dl/postprocessor/ffmpeg.py", line 168, in run_ffmpeg_multiple_files
    raise FFmpegPostProcessorError(msg)
youtube_dl.postprocessor.ffmpeg.FFmpegPostProcessorError: Conversion failed!
```

When I run the `ffmpeg` command manually, I get:

```
[mp4 @ 0x56077bd1bac0] Malformed AAC bitstream detected: use the audio bitstream filter 'aac_adtstoasc' to fix it ('-bsf:a aac_adtstoasc' option with ffmpeg)
av_interleaved_write_frame(): Operation not permitted
```

Adding the `-bsf:a aac_adtstoasc` option as suggested fixes the problem.

By the way, there is another way to make partial downloads playable: just pass `-movflags empty_moov` to FFmpeg in the downloader.

(youtube-dl version: 2016.03.26; not that I think it should matter).

yan12125:
Which URL are you trying? Post the full verbose log.

fstirlitz:
It doesn't really matter. Any m3u8 URL should do; at least each one I tried sufficed. But if you like being flooded with irrelevant nonsense, fine:
```
[...]

PR #9910, opened 2016-06-26
Add -movflags empty_moov when downloading an MP4 file by FFmpegFD

fstirlitz:
With this flag, partial MP4 downloads become playable.

### DIFF
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -313,7 +313,7 @@ class FFmpegFD(ExternalFD):
             if self.params.get('hls_use_mpegts', False) or tmpfilename == '-':
                 args += ['-f', 'mpegts']
             else:
-                args += ['-f', 'mp4']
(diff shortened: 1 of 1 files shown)


dstftw:
Post [an example URL](https://github.com/rg3/youtube-dl/blob/master/README.md#is-the-description-of-the-issue-itself-sufficient).

fstirlitz:
FFS, have you even read the title or the patch?! Anything downloaded by FFmpegFD will do, which usually means `m3u8` streams. You don't even need a URL at all, just grab any large local video file and observe whether it is playable after being partially stuffed into an mp4 container by ffmpeg with and without the switch.

Related: #9286. Where I also basically solved the problem for you, and you people didn't even notice, instead demanding pointless bug reporting theater.

dstftw:
I did. All partial mp4 downloads I've tried play fine for me.

PR #9918, opened 2016-06-26
Made the generation of supported sites repeatable

muffato:
sorted() was not called on the string that's actually printed, but on its prefix (IE_NAME), resulting in some strings being printed in a random order.
This commit makes the sorting repeatable.

### DIFF
--- a/devscripts/make_supportedsites.py
+++ b/devscripts/make_supportedsites.py
@@ -33,10 +33,10 @@ def main():
                 ie_md += ' (Currently broken)'
             yield ie_md
 
-    ies = sorted(youtube_dl.gen_extractors(), key=lambda i: i.IE_NAME.lower())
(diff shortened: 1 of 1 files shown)


dstftw:
Extractor names are supposed to be unique thus clashes are not possible as well as described scenario.

muffato:
Nop: "ARD:mediathek" and "skynewsarabia:video" are each in two extractors (resp. srmediathek.SRMediathekIE and ard.ARDMediathekIE, and skynewsarabia.SkyNewsArabiaArticleIE and skynewsarabia.SkyNewsArabiaIE)
You can check that by printing ie.IE_NAME on each element of the list generated by gen_extractors().

dstftw:
These are obviously copy-paste errors.

PR #9920, opened 2016-06-26
New test-case: extractor names are supposed to be unique

muffato:
@dstftw explained in https://github.com/rg3/youtube-dl/pull/9918#issuecomment-228625878 that extractor names are supposed to be unique.
@dstftw has fixed the two offending extractors, and here I add a test to ensure this does not happen in the future.


dstftw:
Cherry-picked in e3a6747d8f19ad0ba8aee7c3214cdb64903beba0, thanks.

Issue #5185, opened 2015-03-12
Pad %(playlist_index)s with zeroes 2


Romain-Revel:
I am reopening this issue because not fixed entirely:
https://github.com/rg3/youtube-dl/issues/4157

(Tested with version 2015.03.03.1)

Here is a context where a padding would be useful:
I want to download this playlist https://www.youtube.com/playlist?list=PLE30F9d9cZayucP7UqdO37kjo3gtf989s
composed of 121 vidéos at the moment. So I start downloading it with a filename containing %(playlist_index)s. So this program creates the files 001, 002 ... Then there is an error, or the time is too long so I have to stop the download and restart it later with --playlist-start NUMBER. But if I use --playlist-start 22 or more, then there are less than 99 épisodes to download, so the %(playlist_index)s is written on 2 digits.
The first problem is that there is no coherence between the name of the first episodes and the last ones.
The second problem is that as I stopped the first download, it most probably stopped in the middle of the download of one episode, so when it restarts it doesn't find the .part file and starts the download of the current episode from the beginning.

It might be useful to be able to give the format in the --playlist-start in the following way:
--playlist-start 1 => file 1, 2 ... 10 ... 222 ... 99999
--playlist-start 01 => file 01, 02 ... 10 ... 222 ... 99999
--playlist-start 001 => file 001, 002 ... 010 ... 222 ... 99999

yan12125:
FYI, not related to this issue: 2015.03.09 is out

Romain-Revel:
I've just tested with version 2015.03.09, and there is still the same problem.

Try the 2 following commands (provided there are still 121 vidéos in the palylist and not any new one):
youtube-dl --get-filename --playlist-start 22 -o "%(playlist)s - %(playlist_index)s - %(title)s.%(ext)s" https://www.youtube.com/playlist?list=PLE30F9d9cZayucP7UqdO37kjo3gtf989s
youtube-dl --get-filename --playlist-start 23 -o "%(playlist)s - %(playlist_index)s - %(title)s.%(ext)s" https://www.youtube.com/playlist?list=PLE30F9d9cZayucP7UqdO37kjo3gtf989s

jaimeMF:
Instead of deciding automatically how many zeros to use, we could just allow in the output template `%(playlist_index)04d`, which would produce `000x`.

caddor:
download the source and open file "youtube_dl\YoutubeDL.py" 
find function             def prepare_filename(self, info_dict)
[...]

PR #9929, opened 2016-06-27
Add --playlist-index-size option

Moredread:
It works similarly to the --autonumber-size option, i.e. it determines
the length of the '%(playlist_index)s' format string, with '0' as padding. When
it isn't specified it works like before, i.e. the size is determined from the
number of entries to download.

Would be an option for issue #5185.


dstftw:
Instead of introducing unnecessary options [this](https://github.com/rg3/youtube-dl/issues/5185#issuecomment-78540517) generic approach should be implemented.

Moredread:
Damn, I hoped I could get away with easier to write option :p

I'll write something in the issue

Moredread:
Superseeded with PR #9966

PR #9933, opened 2016-06-28
setup.py: add missing blank line.

stepshal:


dstftw:
There is no difference. This file is not checked by flake8.

stepshal:
It's from latest pycodestyle check, one of flake8's tools.

yan12125:
FYI: If you are using pycodestyle, change ```[flake]``` to ```[pycodestyle]``` in https://github.com/rg3/youtube-dl/blob/master/setup.cfg

Issue #9512, opened 2016-05-16
Allow removing fragments for incomplete HLS downloads (was: hls fragment downloader does not clean up after itself)


RealDolos:
- [X] I've **verified** and **I assure** that I'm running youtube-dl **2016.05.10**
- [X] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [X] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
### What is the purpose of your _issue_?
- [X] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [X] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

Not sure if it counts as a bug or a feature request, to me it is a small annoyance tho.

youtube-dl may download tons of fragments via the hls dowloader to individual files.
https://github.com/rg3/youtube-dl/blob/master/youtube_dl/downloader/hls.py#L74-L88

If there is an unrecoverable error, e.g. when a stream is removed, then these files will never get removed; they are only ever removed once all fragments have been retrieved successfully, which will never happen once the stream is gone, no matter how hard one tries to resume.

To fix this properly, there needs to be a distinction between unrecoverable errors (e.g. stream is gone) and recoverable errors (e.g. user aborted, but may resume later).
And maybe even an option to remove fragment files always.

Since I personally use youtube-dl in an automated fashion, my local fix does not have to deal potentially resuming later, so I just moved the `os.remove`-loop into a `finally`.

yan12125:
These files are kept in failure for restarting the download. I guess it's OK to always delete these files if ```--no-continue``` is specified.

dstftw:
This is not actual now since fragments are not stored on disk anymore.

PR #9937, opened 2016-06-28
[fragment,hls,f4m,dash,ism] improve fragment downloading

remitamine:
- resume immediately
- no need to concatenate segments and decrypt them on every resume
- no need to save temp files for segments

and for hls downloader:
- no need to download keys for segments that already downloaded

### DIFF
--- a/youtube_dl/downloader/common.py
+++ b/youtube_dl/downloader/common.py
@@ -327,21 +327,22 @@ class FileDownloader(object):
             os.path.exists(encodeFilename(filename))
         )
 
-        continuedl_and_exists = (
-            self.params.get('continuedl', True) and
-            os.path.isfile(encodeFilename(filename)) and
-            not self.params.get('nopart', False)
-        )
+        if not hasattr(filename, 'write'):
+            continuedl_and_exists = (
+                self.params.get('continuedl', True) and
+                os.path.isfile(encodeFilename(filename)) and
+                not self.params.get('nopart', False)
+            )
 
-        # Check file already present
-        if filename != '-' and (nooverwrites_and_exists or continuedl_and_exists):
-            self.report_file_already_downloaded(filename)
-            self._hook_progress({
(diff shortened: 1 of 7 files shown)


remitamine:
this should fix the problem in #9512.

weedy:
Please don't let these ideas die devs

yan12125:
Allowing the ```filename``` field to be a stream object can be misleading... How about "filename_or_stream"?

youtube_dl/downloader/common.py
```diff
@@ -327,21 +327,22 @@ def download(self, filename, info_dict):
...
-        # Check file already present
-        if filename != '-' and (nooverwrites_and_exists or continuedl_and_exists):
-            self.report_file_already_downloaded(filename)
-            self._hook_progress({
-                'filename': filename,
-                'status': 'finished',
-                'total_bytes': os.path.getsize(encodeFilename(filename)),
-            })
-            return True
+        if type(filename).__name__ != 'BytesIO':
```
yan12125: Should be better with ```isinstanceof(filename, BytesIO)```

remitamine: the `BytesIO` is not used directly in the file, so i decided not to import it.

yan12125: Or ```if not hasattr(filename, 'write')``` or ```if isinstance(filename, compat_str)```? Checking for an exact type makes the codes less flexible.

remitamine:
> Allowing the filename field to be a stream object can be misleading... How about "filename_or_stream"?

> Or if not hasattr(filename, 'write') or if isinstance(filename, compat_str)? Checking for an exact type makes the codes less flexible.

done.

remitamine:
is there something still missing in the code?

yan12125:
That looks good to me. I've also tested several scenarios and all looked fine. Feel free to merge or rebase.

Issue #9940, opened 2016-06-29
Request: support spaces in a format string


keybounce:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like that [x])
- Use *Preview* tab to see how your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2016.06.27*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [ x] I've **verified** and **I assure** that I'm running youtube-dl **2016.06.27**

### Before submitting an *issue* make sure you have:
- [x ] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x ] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [ ] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ x] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

I'd like to see support for spaces in a format string.

Reason: Here is my config file, before I pack it:

```
--no-mtime
-o %(title)s.%(ext)s
-f "
best     [exp=mp4][height=480]
bestvideo[ext=mp4][height=480]+bestaudio[ext=m4a]
best     [exp=mp4][height>360][height<=480]
bestvideo[ext=mp4][height>360][height<=480]+bestaudio
best     [height>360][height<=480]
bestvideo[height>360][height<=480]+bestaudio
best     [exp=mp4][height=>280][height<=360]
best     [height<=480]
worst"
--write-sub
--embed-subs
--recode-video mp4
```
That's understandable, right? Well, guess what that format string will look like with the spaces removed and slashes instead of newlines :-)

yan12125:
I guess keeping slashes is easier to implement.

jaimeMF:
I have opened #9945 for handling multiline quoted strings.
You haven't added `/` between the lines, is that intentional or do you want it to be added implicitly by youtube-dl? 

[...]

PR #9945, opened 2016-06-29
[options] Accept quoted string across multiple lines (#9940)

jaimeMF:
I think there shouldn't be any downside for passing the whole file to shlex, but if there are, please tell me.


yan12125:
Maybe better to add an multiline examples to ```test_compat_shlex_split()```.

jaimeMF:
> Maybe better to add an multiline examples to test_compat_shlex_split().

Added a test in https://github.com/rg3/youtube-dl/pull/9945/commits/3a7e7ed55c3b6d0babde5740aa3a0c1979d16339.

dstftw:
Cherry-picked, thanks.

keybounce:
Thank you

PR #9953, opened 2016-06-30
[XTube] Fix extraction

paulguy:
XTube seems to have changed their video page quite a bit again, this fixes extraction for this site.

I'm a bit rusty at python and have very little experience with this so it's more or less just me trying stuff until it works, but it seems to work.  Had to remove a few fields that couldn't easily be gleaned from the page.

### DIFF
--- a/youtube_dl/extractor/xtube.py
+++ b/youtube_dl/extractor/xtube.py
@@ -26,7 +26,6 @@ class XTubeIE(InfoExtractor):
             'title': 'strange erotica',
             'description': 'contains:an ET kind of thing',
             'uploader': 'greenshowers',
-            'duration': 450,
             'age_limit': 18,
         }
     }, {
@@ -51,38 +50,39 @@ class XTubeIE(InfoExtractor):
         req.add_header('Cookie', 'age_verified=1; cookiesAccepted=1')
         webpage = self._download_webpage(req, display_id)
 
-        flashvars = self._parse_json(
-            self._search_regex(
-                r'xt\.playerOps\s*=\s*({.+?});', webpage, 'player ops'),
-            video_id)['flashvars']
+        sources = self._search_regex(r'sources:\s*({.+?}),', webpage, 'sources')
 
-        title = flashvars.get('title') or self._search_regex(
-            r'<h1>([^<]+)</h1>', webpage, 'title')
-        video_url = compat_urllib_parse_unquote(flashvars['video_url'])
-        duration = int_or_none(flashvars.get('video_duration'))
+        sourcesdict = self._parse_json(sources, video_id)
+
+        formats = []
+        for formatkey in sourcesdict.keys():
+            curformat = {}
+            curformat['height'] = int(formatkey)
+            curformat['url'] = sourcesdict[formatkey]
+            formats.append(curformat)
+            
+        formats = self._sort_formats(formats)
+
+        title = self._search_regex(
+            r'videoTitle:\s*\'(.+?)\',', webpage, 'title')
 
         uploader = self._search_regex(
-            r'<input[^>]+name="contentOwnerId"[^>]+value="([^"]+)"',
+            r'<span class=\"nickname\">([^<]+)',
             webpage, 'uploader', fatal=False)
         description = self._search_regex(
             r'</h1>\s*<p>([^<]+)', webpage, 'description', fatal=False)
         view_count = str_to_int(self._search_regex(
             r'<dt>Views:</dt>\s*<dd>([\d,\.]+)</dd>',
             webpage, 'view count', fatal=False))
-        comment_count = str_to_int(self._html_search_regex(
-            r'>Comments? \(([\d,\.]+)\)<',
-            webpage, 'comment count', fatal=False))
 
         return {
             'id': video_id,
             'display_id': display_id,
-            'url': video_url,
+            'formats': formats,
             'title': title,
             'description': description,
             'uploader': uploader,
-            'duration': duration,
             'view_count': view_count,
-            'comment_count': comment_count,
             'age_limit': 18,
         }
 


youtube_dl/extractor/xtube.py
```diff
@@ -51,38 +50,62 @@ def _real_extract(self, url):
...
+                continue
+            elif line[0].isalpha(): # key
+                newjsonstr += '\"' + line[0:line.index(':')] + '\":'
+                if line[line.index(':') + 2] == '\'': # string value
+                    newjsonstr += ' \"' + line[line.index(':') + 3:line.rindex('\'')]\
+                                  + '\",'
+                else:
+                    newjsonstr += line[line.index(':') + 1:]
+            else:
+                newjsonstr += line
```
dstftw: This looks like `utils.js_to_json`.

paulguy: It mostly works, but in this case, the site outputs a single line comment in the JSON (`//`) which _parse_json() isn't OK with.

yan12125: Extract ```sources``` and ```videoTitle``` separately.

youtube_dl/extractor/xtube.py
```diff
@@ -51,38 +50,44 @@ def _real_extract(self, url):
         req.add_header('Cookie', 'age_verified=1; cookiesAccepted=1')
         webpage = self._download_webpage(req, display_id)
 
-        flashvars = self._parse_json(
-            self._search_regex(
-                r'xt\.playerOps\s*=\s*({.+?});', webpage, 'player ops'),
-            video_id)['flashvars']
+#        jsonstr = self._search_regex(
+#                r'MHP1138\.createPlayer\(\'xtubePlayer\',\s*{.*medias:\s*({.+?}),\s*menu:',
+#                webpage, 'player ops', flags=re.DOTALL, group=1)
```
yan12125: Remove unused codes.

youtube_dl/extractor/xtube.py
```diff
@@ -51,38 +50,44 @@ def _real_extract(self, url):
...
 
-        title = flashvars.get('title') or self._search_regex(
-            r'<h1>([^<]+)</h1>', webpage, 'title')
-        video_url = compat_urllib_parse_unquote(flashvars['video_url'])
-        duration = int_or_none(flashvars.get('video_duration'))
+        sources = self._search_regex(r'sources:\s*({.+?}),', webpage, 'sources')
+
+        sourcesdict = self._parse_json(sources, video_id)
+        print(str(sourcesdict))
```
yan12125: Remove debugging codes.

youtube_dl/extractor/xtube.py
```diff
@@ -51,38 +50,44 @@ def _real_extract(self, url):
...
+
+        formats = []
+        for formatkey in sourcesdict.keys():
+            curformat = {}
+            curformat['height'] = int(formatkey)
+            curformat['url'] = sourcesdict[formatkey]
+            formats.append(curformat)
+            
+        formats = sorted(formats, key=lambda x: x['height'])
```
yan12125: Use ```self._sort_formats(formats)``` instead.

youtube_dl/extractor/xtube.py
paulguy: Oops.

yan12125:
Thanks. Could you squash all commits into a single one?

paulguy:
I'm not sure how...

yan12125:
Well, the easiest way is to re-commit changes.

youtube_dl/extractor/xtube.py
```diff
@@ -51,38 +50,39 @@ def _real_extract(self, url):
...
 
         return {
             'id': video_id,
             'display_id': display_id,
-            'url': video_url,
+            'formats': formats,
             'title': title,
             'description': description,
             'uploader': uploader,
-            'duration': duration,
```
dstftw: It's available and should be extracted.

youtube_dl/extractor/xtube.py
```diff
@@ -51,38 +50,39 @@ def _real_extract(self, url):
...
             'view_count': view_count,
-            'comment_count': comment_count,
```
dstftw: This as well.

youtube_dl/extractor/xtube.py
```diff
@@ -51,38 +50,39 @@ def _real_extract(self, url):
...
-        title = flashvars.get('title') or self._search_regex(
-            r'<h1>([^<]+)</h1>', webpage, 'title')
-        video_url = compat_urllib_parse_unquote(flashvars['video_url'])
-        duration = int_or_none(flashvars.get('video_duration'))
+        sourcesdict = self._parse_json(sources, video_id)
+
+        formats = []
+        for formatkey in sourcesdict.keys():
+            curformat = {}
+            curformat['height'] = int(formatkey)
```
dstftw: Use `int_or_none` instead.

youtube_dl/extractor/xtube.py
```diff
@@ -51,38 +50,39 @@ def _real_extract(self, url):
...
+        sources = self._search_regex(r'sources:\s*({.+?}),', webpage, 'sources')
 
-        title = flashvars.get('title') or self._search_regex(
-            r'<h1>([^<]+)</h1>', webpage, 'title')
-        video_url = compat_urllib_parse_unquote(flashvars['video_url'])
-        duration = int_or_none(flashvars.get('video_duration'))
+        sourcesdict = self._parse_json(sources, video_id)
+
+        formats = []
+        for formatkey in sourcesdict.keys():
```
dstftw: Use `for key, value in sourcesdict.items()` instead.

paulguy:
I guess I'll just leave it up to people who know what they're doing next time.

PR #9964, opened 2016-06-30
[Pornhub] Fix error detection

kidol:
Regex was outdated. Example url's below.

Deleted by uploader: http://www.pornhub.com/view_video.php?viewkey=ph572716d15a111
Copyright claim: http://www.pornhub.com/view_video.php?viewkey=ph57151b9f16dba


dstftw:
Cherry-picked in 044e3d91b5715f7aa63c578097b77fd510ed0f73, thanks.

PR #9966, opened 2016-06-30
Output template now supports '0*d' zero padding

Moredread:
Might be an option for issue #5185.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -33,6 +33,7 @@ from youtube_dl.utils import (
     ExtractorError,
     find_xpath_attr,
     fix_xml_ampersands,
+    auto_str_to_int_fmt,
     InAdvancePagedList,
     intlist_to_bytes,
     is_html,
@@ -984,6 +985,17 @@ The first line
     def test_urshift(self):
         self.assertEqual(urshift(3, 1), 1)
         self.assertEqual(urshift(-3, 1), 2147483646)
+
+    def test_auto_str_to_int_fmt(self):
+        input = {"foo": "bar",
+                 "number": 1234,
+                 "s_number": "001234"}
+
+        self.assertEqual(auto_str_to_int_fmt("%(foo)s", input), "bar")
+        self.assertEqual(auto_str_to_int_fmt("%(number)d", input), "1234")
+        self.assertEqual(auto_str_to_int_fmt("%(s_number)d", input), "1234")
+        self.assertEqual(auto_str_to_int_fmt("%(s_number)07d", input), "0001234")
+        self.assertEqual(auto_str_to_int_fmt("%(foo)s%(number)d-%(s_number)d", input), "bar1234-1234")
 
 if __name__ == '__main__':
     unittest.main()
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -41,6 +41,7 @@ from .compat import (
 from .utils import (
     age_restricted,
     args_to_str,
+    auto_str_to_int_fmt,
     ContentTooShortError,
     date_from_str,
     DateRange,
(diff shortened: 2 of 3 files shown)


Moredread:
I'm open for a better name for the fmt function. :p

dstftw:
1. Complete python 2 [string formatting operations specification](https://docs.python.org/2/library/stdtypes.html#string-formatting) should be supported.
2. Current code has lots of issues. FYI, I've an implementation started several months ago that covers 1 but postponed due some subtleties I don't actually remember now. Here are the test set (integer only) I've used to test own implementation against (everything passes). Yours should pass as well:
```python
    def test_prepare_filename(self):
        info = {
            'id': '1234',
            'ext': 'mp4',
            'width': None,
            'height': 1080,
        }

        def fname(templ):
            ydl = YoutubeDL({'outtmpl': templ})
            return ydl.prepare_filename(info)
        self.assertEqual(fname('%(id)s.%(ext)s'), '1234.mp4')
        self.assertEqual(fname('%(id)s-%(width)s.%(ext)s'), '1234-NA.mp4')
        # Replace missing fields with 'NA'
        self.assertEqual(fname('%(uploader_date)s-%(id)s.%(ext)s'), 'NA-1234.mp4')
        self.assertEqual(fname('%(abr)d.%(ext)s'), 'NA.mp4')
        self.assertEqual(fname('%(height)d.%(ext)s'), '1080.mp4')
        self.assertEqual(fname('%(height)d.%(ext)s'), '1080.mp4')
        self.assertEqual(fname('%(height)6d.%(ext)s'), '  1080.mp4')
        self.assertEqual(fname('%(height)-6d.%(ext)s'), '1080  .mp4')
        self.assertEqual(fname('%(height)06d.%(ext)s'), '001080.mp4')
        self.assertEqual(fname('%(height) 06d.%(ext)s'), ' 01080.mp4')
        self.assertEqual(fname('%(height)   06d.%(ext)s'), ' 01080.mp4')
        self.assertEqual(fname('%(height)0 6d.%(ext)s'), ' 01080.mp4')
        self.assertEqual(fname('%(height)0   6d.%(ext)s'), ' 01080.mp4')
        self.assertEqual(fname('%(height)   0   6d.%(ext)s'), ' 01080.mp4')
        self.assertEqual(fname('%%(height)06d.%(ext)s'), '%(height)06d.mp4')
        self.assertEqual(fname('%(width)06d.%(ext)s'), 'NA.mp4')
        self.assertEqual(fname('%(width)06d.%%(ext)s'), 'NA.%(ext)s')
        self.assertEqual(fname('%%(width)06d.%(ext)s'), '%(width)06d.mp4')
```

jaimeMF:
I think it can be achieved in a simpler way, see this example:
```python
class StringInt(str):
    def __int__(self):
        return int(str(self))

kw = {
    'id': 'foo',
    'number': '001234',
}

for k, v in kw.items():
    if isinstance(v, str) and v.isdigit():
        kw[k] = StringInt(v)

formats = [
    '%(id)s-%(number)s',
    '%(id)s-%(number)d',
    '%(id)s-%(number)06d',
    '%(id)s-%(number)-6d.ext',
]

for f in formats:
    print('==> f=', f)
    print(f % kw)
```

dstftw:
@jaimeMF this approach fails on missing keys at least. And seems there is no way to stuff `NA` into non-string format specifier without rewriting it to `%(field)s`.

Moredread:
The solution by @jaimeMF is much cleaner than mine, as no string manipulation is necessary (and I'm really not happy about parsing an informal spec (that changed between Python versions) with regexes).

I'll adapt it to my PR, and also add the tests, but IMO supporting missing fields should be handled in a different PR, as it definitely involves string parsing as @dstftw notes, and is a different feature, that is useful with or without full formatting string support.

dstftw:
This should be a single PR since not handling missing fields breaks current behavior.

jaimeMF:
@dstftw Note that `playlist_index` is a string, so it's not affected with your patch. I'm fine with making it a number, but keep in mind that it won't be automatically padded with he number of entries in the playlist.

dstftw:
If we introduce string formatting operations padding as well as any other formatting is the responsibility of the user imo.

jaimeMF:
> If we introduce string formatting operations padding as well as any other formatting is the responsibility of the user imo.

I think as you, but I don't personally use them so real users may disagree.

Issue #9930, opened 2016-06-28
Allow Random Values for --sleep-interval


ap4125:
The `--sleep-interval` option is great. I think that this is the main reason why we can run downloads from heavily monitored sites.

Is there a way to modify this option so that it can accept random values? For example, instead of setting it to `--sleep-interval 300` to set it to sleep for 300 seconds or 5 minutes between each download, make it like so:

`--sleep-interval 120-300`

so that the sleep interval would be set between 120 and 300 seconds.

The rationale is to simulate more realistic browsing behavior.

shan3k:
Having a browse here to see what I could actually work on. 
Could we approach it like this?

`        sleep_interval = random.randrange(int(self.params.get('sleep_interval').split('-')[0]), int(self.params.get('sleep_interval').split('-')[1]))
`

I guess thats not ideal because we are taking a string as input and instead we should take two ints (and upper and a lower)? 

Maybe @yan12125 can provide feedback on my noob approach :+1:

PR #9980, opened 2016-07-01
Update sleep-interval to allow rand range in addition to int value #9930

scopeviable:
Adds random number feature to sleep_interval as suggested in #9930

### DIFF
--- a/youtube_dl/downloader/common.py
+++ b/youtube_dl/downloader/common.py
@@ -4,7 +4,6 @@ import os
 import re
 import sys
 import time
-import random
 
 from ..compat import compat_os_name
 from ..utils import (
@@ -343,7 +342,7 @@ class FileDownloader(object):
             })
             return True
 
-        sleep_interval = random.randrange(self.params.get('sleep_interval'))
+        sleep_interval = self.params.get('sleep_interval')
         if sleep_interval:
             self.to_screen('[download] Sleeping %s seconds...' % sleep_interval)
             time.sleep(sleep_interval)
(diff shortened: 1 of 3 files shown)


dstftw:
It should not change current behavior.

shan3k:
ack, maybe I can update with --random_interval flag?

expenses:
IMO I think it should work as default for `--sleep_interval NUM`, and randomly for `--sleep_internal NUM-NUM`.

dstftw:
`--sleep-interval` supporting both variants and documented should be enough.

shan3k:
@dstftw @expenses ack, let me have a go at that so. Still learning :+1:

youtube_dl/downloader/common.py
```diff
@@ -343,6 +344,12 @@ def download(self, filename, info_dict):
             return True
 
         sleep_interval = self.params.get('sleep_interval')
+        if '-' in sleep_interval:
+            try:
+                start, stop = sleep_interval.split('-')
+                sleep_interval = random.randrange(int(start), int(stop))
```
dstftw: floating point must be supported.

dstftw:
Well, `start-end` format seems to be a not so good idea. Floating point numbers can be written in scientific notation, e.g. `3e-1`.

expenses:
@dstftw Agreed, perhaps you could have `--sleep_interval time` for a constant time value, and `--sleep_interval start end` for a random time value?

shan3k:
So, I guess in that case if we use `--sleep_interval time`` --sleep_interval start end` with support for floats, we should take input as a list and then parse?

expenses:
@shan3k Seems to be the way to go.

PR #9987, opened 2016-07-02
[Pornhub] Fix private video detection

kidol:
Forgot to test private video in #9964, so it got broke...

Example url: http://www.pornhub.com/view_video.php?viewkey=ph56fd731fce6b7

### DIFF
--- a/youtube_dl/extractor/pornhub.py
+++ b/youtube_dl/extractor/pornhub.py
@@ -107,7 +107,7 @@ class PornHubIE(InfoExtractor):
         webpage = self._download_webpage(req, video_id)
 
         error_msg = self._html_search_regex(
-            r'(?s)<div[^>]+class=(["\']).*?\bremoved\b.*?\1[^>]*>(?P<error>.+?)</div>',
+            r'(?s)<div[^>]+class=(["\']).*?\b(removed|userMessageSection)\b.*?\1[^>]*>(?P<error>.+?)</div>',
             webpage, 'error message', default=None, group='error')
         if error_msg:
             error_msg = re.sub(r'\s+', ' ', error_msg)
--- a/youtube_dl/extractor/pornhub.py
+++ b/youtube_dl/extractor/pornhub.py
@@ -107,7 +107,7 @@ class PornHubIE(InfoExtractor):
         webpage = self._download_webpage(req, video_id)
 
         error_msg = self._html_search_regex(
-            r'(?s)<div[^>]+class=(["\']).*?\b(removed|userMessageSection)\b.*?\1[^>]*>(?P<error>.+?)</div>',
+            r'(?s)<div[^>]+class=(["\']).*?\b(?:removed|userMessageSection)\b.*?\1[^>]*>(?P<error>.+?)</div>',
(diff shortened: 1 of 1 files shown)


dstftw:
Add a test.

kidol:
`python test/test_download.py TestDownload.test_PornHub`
only runs the first test... How to run all?

>Ran 1 test in 5.155s

dstftw:
`TestDownload.test_PornHub_<number>`.

youtube_dl/extractor/pornhub.py
```diff
@@ -107,7 +107,7 @@ def _real_extract(self, url):
         webpage = self._download_webpage(req, video_id)
 
         error_msg = self._html_search_regex(
-            r'(?s)<div[^>]+class=(["\']).*?\bremoved\b.*?\1[^>]*>(?P<error>.+?)</div>',
+            r'(?s)<div[^>]+class=(["\']).*?\b(removed|userMessageSection)\b.*?\1[^>]*>(?P<error>.+?)</div>',
```
dstftw: Do not capture group if you don't use it.

kidol:
Thanks but I still don't know how the test works. Test 1 & 2 work with `TestDownload.test_PornHub_<number>`, how do the rest work?

i.e.:

```python
{
    # removed at the request of the copyright owner
    'url': 'http://www.pornhub.com/view_video.php?viewkey=788152859',
    'only_matching': True,
},
```

What does only_matching mean?

dstftw:
`only_matched` means exactly that: only test URL to match `_VALID_URL`. It's tested in `test_all_urls.py`.

kidol:
>only_matched means exactly that: ...

Ok the comments confused me, they suggest something else.

Anyway, when I look at other tests (YouTube), I don't see any tests which check for specific error messages like "this video is private" etc.

What do you want me to do?

dstftw:
Add only matching test.

kidol:
I don't get it.

This url structure

>view_video.php?viewkey=phxxxxxxxxxxx

is already covered in a matching test.

?

dstftw:
So what?

kidol:
You want me to add this:

```python
{
        # video is private
        'url': 'http://www.pornhub.com/view_video.php?viewkey=ph56fd731fce6b7',
        'only_matching': True,
}
```

But why? It's same as existing test:

```python
{
        # removed by uploader
        'url': 'http://www.pornhub.com/view_video.php?viewkey=ph572716d15a111',
        'only_matching': True,
}
```

dstftw:
Because
>Forgot to test private video in #9964, so it got broke...

kidol:
Yes this was related to the regex. Has nothing to do with url structure.

kidol:
Now there are 2 matching tests which are equivalent and for what?
I just want to understand... :weary:

dstftw:
All these tests are for reference to keep track of possible "unavailable" URLs to do manual testing after some changes applied in extractor's code. Doing real testing for them during travis build is just a time waste since it does not really matter whether unavailability will be detected or not. What really matters is to test "available" URLs are really available.

kidol:
Ok thanks.

Issue #10009, opened 2016-07-05
Wrong YouTube channel downloaded when grabbing geoblocked channel


ivan:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like that [x])
- Use *Preview* tab to see how your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2016.07.05*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.07.05**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add `-v` flag to **your command line** you run youtube-dl with, copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):

```
# PYTHONPATH=. python3 -m youtube_dl -v 'https://www.youtube.com/user/kananishinoSMEJ/videos'
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['-v', 'https://www.youtube.com/user/kananishinoSMEJ/videos']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.07.05
[debug] Git HEAD: 4080530
[debug] Python version 3.5.1+ - Linux-4.4.8-std-1-armv7l-with-Ubuntu-16.04-xenial
[...]

PR #10011, opened 2016-07-05
[youtube] Don't extract the wrong channel playlist for geoblocked channel

ivan:
This is a potential fix for #10009 that extracts the channel ID from two different places and makes sure that they match.  This should avoid filling up directories with videos from an incorrect channel, at the expense of possibly having to update the channel-playlist-extraction code more often.  In the case that this double-extraction fails for some channels, I expect a bug report to appear pretty soon :-)


ivan:
@dstftw Do you know of any channels that actually required `data-(?:channel-external-|yt)id="([^"]+)"` because they had no `channelId`?  (If you do, my PR is bad.)

ivan:
This is buggy; don't merge

PR #10025, opened 2016-07-07
[BrainPOP] Add new extractor (Closes #10000)

### LINKED ISSUE #10000: Site support request: Brainpop
Example URL: https://www.brainpop.com/english/freemovies/williamshakespeare/

nehalvpatel:
- [ ] Bug fix
- [x] New extractor
- [ ] New feature

---

Added a new extractor for [BrainPOP](https://www.brainpop.com/). In addition to the video, it extracts the description and the thumbnail. It should work on any video on the site, even if it's in a different language.

### DIFF
--- a/youtube_dl/extractor/brainpop.py
+++ b/youtube_dl/extractor/brainpop.py
@@ -1,0 +1,47 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class BrainPOPIE(InfoExtractor):
+    _VALID_URL = r'https?:\/\/(?:(.+)\.)?brainpop\.com\/(?P<id>[^\r\n]+)'
+    _TEST = {
+        'url': 'https://www.brainpop.com/english/freemovies/williamshakespeare/',
+        'md5': '676d936271b628dc05e4cec377751919',
+        'info_dict': {
+            'id': 'english/freemovies/williamshakespeare/',
+            'ext': 'mp4',
+            'title': 'William Shakespeare - BrainPOP',
+            'thumbnail': 're:^https?://.*\.png$',
+            'description': 'He could do comedies, tragedies, histories and poetry.  Learn about the greatest playwright in the history of the English language!',
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        self.report_extraction(video_id)
+
+        ec_token = self._html_search_regex(r'ec_token : \'(.+)\'', webpage, "token")
+        movie_cdn_path = self._html_search_regex(r'movie_cdn_path : \'(.+)\'', webpage, "cdn path")
+        mp4 = self._html_search_regex(r'mp4":"([^"]*)', webpage, "mp4")
+
+        url = movie_cdn_path + mp4.replace("\\", "") + "?" + ec_token
+
+        title = self._html_search_regex(r'type":"Movie","name":"([^"]*)"', webpage, "title")
+
+        thumbnail_cdn = self._html_search_regex(r"'cdn_path' : '([^']*)'", webpage, "thumbnail cdn")
+        thumbnail_image = self._html_search_regex(r'type":"Movie","name":"[^"]*","image":"([^"]*)"', webpage, "thumbnail")
+        thumbnail = thumbnail_cdn + thumbnail_image.replace("\\", "")
+
+        description = self._html_search_regex(r'type":"Movie","name":"[^"]*","image":"[^"]*","description":"([^"]*)"', webpage, "description")
+
+        return {
+            'id': video_id,
+            'url': url,
+            'title': title,
+            'thumbnail': thumbnail,
+            'description': description,
+        }
--- a/youtube_dl/extractor/brainpop.py
+++ b/youtube_dl/extractor/brainpop.py
@@ -24,24 +24,43 @@ class BrainPOPIE(InfoExtractor):
 
         self.report_extraction(video_id)
 
-        ec_token = self._html_search_regex(r'ec_token : \'(.+)\'', webpage, 'token')
-        movie_cdn_path = self._html_search_regex(r'movie_cdn_path : \'(.+)\'', webpage, 'cdn path')
-        mp4 = self._html_search_regex(r'mp4":"([^"]*)', webpage, 'mp4')
+        ec_token = self._html_search_regex(r"ec_token : '([^']*)'", webpage, 'token')
 
-        url = movie_cdn_path + mp4.replace('\\', '') + '?' + ec_token
+        settings = self._parse_json(self._html_search_regex(r'var settings = ([^;]*)', webpage, 'settings'), video_id)
+        title = settings['title']
+        description = settings['description']
 
-        title = self._html_search_regex(r'type":"Movie","name":"([^"]*)"', webpage, 'title') or self._html_search_regex(r'<title>(.+?)</title>', webpage, 'title')
+        global_content = self._parse_json(self._html_search_regex(r'var global_content = ([^;]*)', webpage, 'global content').replace("'", '"'), video_id)
+        cdn_path = global_content['cdn_path']
+        movie_cdn_path = global_content['movie_cdn_path']
 
-        thumbnail_cdn = self._html_search_regex(r"'cdn_path' : '([^']*)'", webpage, 'thumbnail cdn', fatal=False)
-        thumbnail_image = self._html_search_regex(r'type":"Movie","name":"[^"]*","image":"([^"]*)"', webpage, 'thumbnail', fatal=False)
-        thumbnail = thumbnail_cdn + thumbnail_image.replace('\\', '')
+        content = self._parse_json(self._html_search_regex(r'var content = ([^;]*)', webpage, 'content'), video_id)
+        movies = content['category']['unit']['topic']['movies']
+        screenshots = content['category']['unit']['topic']['screenshots']
 
-        description = self._html_search_regex(r'type":"Movie","name":"[^"]*","image":"[^"]*","description":"([^"]*)"', webpage, 'description', fatal=False)
+        formats = []
+        formats.append({
+            'url': movie_cdn_path + movies['mp4'] + '?' + ec_token,
+            'height': 768,
+            'width': 768,
+        })
+        formats.append({
+            'url': movie_cdn_path + movies['mp4_small'] + '?' + ec_token,
+            'height': 480,
+            'width': 480,
+        })
+        self._sort_formats(formats)
+
+        thumbnails = []
+        for (i, screenshot) in enumerate(screenshots):
+            thumbnails.append({
+                'url': cdn_path + screenshot,
+            })
 
         return {
             'id': video_id,
-            'url': url,
             'title': title,
-            'thumbnail': thumbnail,
+            'formats': formats,
+            'thumbnails': thumbnails,
             'description': description,
         }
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/brainpop.py
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        self.report_extraction(video_id)
+
+        ec_token = self._html_search_regex(r'ec_token : \'(.+)\'', webpage, 'token')
```
TRox1972: Its easier to parse the JSON instead of searching through it manually.

youtube_dl/extractor/brainpop.py
```diff
@@ -0,0 +1,47 @@
...
+
+        thumbnail_cdn = self._html_search_regex(r"'cdn_path' : '([^']*)'", webpage, 'thumbnail cdn', fatal=False)
+        thumbnail_image = self._html_search_regex(r'type":"Movie","name":"[^"]*","image":"([^"]*)"', webpage, 'thumbnail', fatal=False)
+        thumbnail = thumbnail_cdn + thumbnail_image.replace('\\', '')
+
+        description = self._html_search_regex(r'type":"Movie","name":"[^"]*","image":"[^"]*","description":"([^"]*)"', webpage, 'description', fatal=False)
+
+        return {
+            'id': video_id,
+            'url': url,
```
TRox1972: There are multiple formats

youtube_dl/extractor/brainpop.py
nehalvpatel: @TRox1972 Right, but the JSON is mixed in with the HTML. How should I go about extracting it?

youtube_dl/extractor/brainpop.py
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        self.report_extraction(video_id)
+
+        ec_token = self._html_search_regex(r"ec_token : '([^']*)'", webpage, 'token')
```
TRox1972: It this needed? Does it work without it?

youtube_dl/extractor/brainpop.py
```diff
@@ -0,0 +1,66 @@
...
+        })
+        formats.append({
+            'url': movie_cdn_path + movies['mp4_small'] + '?' + ec_token,
+            'height': 480,
+            'width': 480,
+        })
+        self._sort_formats(formats)
+
+        thumbnails = []
+        for (i, screenshot) in enumerate(screenshots):
```
TRox1972: Just loop through `screenshots`; you aren't using `i` anyway

TRox1972: Also you can shorten this to a block statement, I guess

youtube_dl/extractor/brainpop.py
nehalvpatel: The token? Yea, you can't access the videos without it.

youtube_dl/extractor/brainpop.py
```diff
@@ -0,0 +1,66 @@
...
+        title = settings['title']
+        description = settings['description']
+
+        global_content = self._parse_json(self._html_search_regex(r'var global_content = ([^;]*)', webpage, 'global content').replace("'", '"'), video_id)
+        cdn_path = global_content['cdn_path']
+        movie_cdn_path = global_content['movie_cdn_path']
+
+        content = self._parse_json(self._html_search_regex(r'var content = ([^;]*)', webpage, 'content'), video_id)
+        movies = content['category']['unit']['topic']['movies']
+        screenshots = content['category']['unit']['topic']['screenshots']
```
TRox1972: This should be non-fatal

youtube_dl/extractor/brainpop.py
nehalvpatel: Sorry, I'm new to Python. Could you show me how?

youtube_dl/extractor/brainpop.py
```diff
@@ -0,0 +1,66 @@
...
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        self.report_extraction(video_id)
+
+        ec_token = self._html_search_regex(r"ec_token : '([^']*)'", webpage, 'token')
+
+        settings = self._parse_json(self._html_search_regex(r'var settings = ([^;]*)', webpage, 'settings'), video_id)
+        title = settings['title']
+        description = settings['description']
```
TRox1972: It's better to return `title` and `description` directly since you are not using the variables anywhere else. Also, description extraction should be non-fatal

youtube_dl/extractor/brainpop.py
nehalvpatel: Which part? The settings JSON?

youtube_dl/extractor/brainpop.py
TRox1972:
```
thumbnails = [{'url': cdn_path + screenshot} for screenshot in screenshots]

File: youtube_dl/extractor/brainpop.py (hunk 5 above)
Comment (TRox1972) [inline youtube_dl/extractor/brainpop.py]: Nope, screenshots, e.g. `screenshots = content['category']['unit']['topic'].get('screenshots', {}`

File: youtube_dl/extractor/brainpop.py (hunk 6 above)
Comment (TRox1972) [inline youtube_dl/extractor/brainpop.py]: You can also remove the '` - BrainPOP`' string from the title by using `from ..utils import remove_end`

File: youtube_dl/extractor/brainpop.py [hunk 7]
```diff
@@ -0,0 +1,66 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class BrainPOPIE(InfoExtractor):
+    _VALID_URL = r'https?:\/\/(?:(.+)\.)?brainpop\.com\/(?P<id>[^\r\n]+)'
```
TRox1972: Only the last part of the URL should be the `id`

youtube_dl/extractor/brainpop.py
```diff
@@ -0,0 +1,66 @@
...
+            'id': 'english/freemovies/williamshakespeare/',
+            'ext': 'mp4',
+            'title': 'William Shakespeare - BrainPOP',
+            'thumbnail': 're:^https?://.*\.png$',
+            'description': 'He could do comedies, tragedies, histories and poetry.  Learn about the greatest playwright in the history of the English language!',
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
```
TRox1972: Use `display_id` instead; the 'real' `ID` seems to be stored in the JSON as `EntryID`

youtube_dl/extractor/brainpop.py
```diff
@@ -0,0 +1,66 @@
...
+        self.report_extraction(video_id)
+
+        ec_token = self._html_search_regex(r"ec_token : '([^']*)'", webpage, 'token')
+
+        settings = self._parse_json(self._html_search_regex(r'var settings = ([^;]*)', webpage, 'settings'), video_id)
+        title = settings['title']
+        description = settings['description']
+
+        global_content = self._parse_json(self._html_search_regex(r'var global_content = ([^;]*)', webpage, 'global content').replace("'", '"'), video_id)
+        cdn_path = global_content['cdn_path']
```
TRox1972: This should be non-fatal

youtube_dl/extractor/brainpop.py
```diff
@@ -50,17 +52,14 @@ def _real_extract(self, url):
...
+        settings = self._parse_json(self._html_search_regex(r'var settings = ([^;]*)', webpage, 'settings'), display_id)
 
         return {
-            'id': video_id,
-            'title': title,
-            'formats': formats,
+            'id': content['category']['unit']['topic']['EntryID'],
+            'display_id': display_id,
+            'title': remove_end(settings['title'], ' - BrainPOP'),
+            'description': settings['description'],
```
TRox1972: This should be non-fatal

youtube_dl/extractor/brainpop.py
```diff
@@ -2,42 +2,44 @@
...
+        global_content = self._parse_json(self._html_search_regex(r'var global_content = ([^;]*)', webpage, 'global content').replace("'", '"'), display_id)
+        cdn_path = global_content.get('cdn_path', 'https://cdn.brainpop.com')
+        movie_cdn_path = global_content.get('movie_cdn_path', 'https://svideos.brainpop.com')
         ec_token = self._html_search_regex(r"ec_token : '([^']*)'", webpage, 'token')
 
-        settings = self._parse_json(self._html_search_regex(r'var settings = ([^;]*)', webpage, 'settings'), video_id)
-        title = settings['title']
-        description = settings['description']
+        screenshots = content['category']['unit']['topic'].get('screenshots', {})
+        thumbnails = [{'url': cdn_path + screenshot} for screenshot in screenshots]
```
TRox1972: You can cut this down to one statement

youtube_dl/extractor/brainpop.py
```diff
@@ -2,42 +2,44 @@
...
     }
 
     def _real_extract(self, url):
-        video_id = self._match_id(url)
-        webpage = self._download_webpage(url, video_id)
+        display_id = self._match_id(url)
+        webpage = self._download_webpage(url, display_id)
 
-        self.report_extraction(video_id)
+        content = self._parse_json(self._html_search_regex(r'var content = ([^;]*)', webpage, 'content'), display_id)
```
TRox1972: Just get `['category']['unit']['topic']` right away, so you don't need to specify it every time

youtube_dl/extractor/brainpop.py
```diff
@@ -52,14 +52,14 @@ def _real_extract(self, url):
...
+
+        settings = self._parse_json(self._html_search_regex(r'var settings = ([^;]*)', webpage, 'settings JSON', '{}'), display_id)
 
         return {
-            'id': content['category']['unit']['topic']['EntryID'],
+            'id': topic['EntryID'],
             'display_id': display_id,
-            'title': remove_end(settings['title'], ' - BrainPOP'),
-            'description': settings['description'],
+            'title': remove_end(settings.get('title', display_id), ' - BrainPOP'),
```
TRox1972: The `title` _should_ be mandatory

yan12125: Such a pattern is OK if ```title``` is missing in ```settings```. @nehalvpatel Do you have an example?

TRox1972: `title` can also be extracted from `<title>(.*)</title>` also. One of them should be fatal, right?

yan12125: Either way is fine. In general ```<title>``` is better than extracting from embedded data as the former is less likely to change. If there's an example that ```<title>``` or ```settings['title']``` is missing, a fallback should be provided, otherwise the extraction should be fatal.

youtube_dl/extractor/brainpop.py
```diff
@@ -0,0 +1,65 @@
...
+            'width': 480,
+        })
+        self._sort_formats(formats)
+
+        settings = self._parse_json(self._html_search_regex(r'var settings = ([^;]*)', webpage, 'settings JSON', '{}'), display_id)
+
+        return {
+            'id': topic['EntryID'],
+            'display_id': display_id,
+            'title': remove_end(settings.get('title', display_id), ' - BrainPOP'),
```
TRox1972: Also extract title from `<title>(.+)</title>` in case current extraction fails

EpicCodeWizard:
Does this thing still work?

PR #10028, opened 2016-07-07
[polskieradio] Add thumbnails.

JakubAdamWieczorek:
- [x] Bug fix
- [ ] New extractor
- [ ] New feature

---

This is a simple change to extract thumbnails from Polskie Radio auditions.

### DIFF
--- a/youtube_dl/extractor/polskieradio.py
+++ b/youtube_dl/extractor/polskieradio.py
@@ -33,6 +33,7 @@ class PolskieRadioIE(InfoExtractor):
                 'timestamp': 1456594200,
                 'upload_date': '20160227',
                 'duration': 2364,
+                'thumbnail': 're:^https?://static.prsa.pl/images/.*\.jpg$'
             },
         }],
     }, {
@@ -68,6 +69,8 @@ class PolskieRadioIE(InfoExtractor):
             r'(?s)<span[^>]+id="datetime2"[^>]*>(.+?)</span>',
             webpage, 'timestamp', fatal=False))
 
+        thumbnail_url = self._og_search_thumbnail(webpage)
+
         entries = []
 
         media_urls = set()
@@ -87,6 +90,7 @@ class PolskieRadioIE(InfoExtractor):
                 'duration': int_or_none(media.get('length')),
                 'vcodec': 'none' if media.get('provider') == 'audio' else None,
                 'timestamp': timestamp,
+                'thumbnail': thumbnail_url
             })
 
         title = self._og_search_title(webpage).strip()


yan12125:
Usually the ```thumbnail``` field in ```_TESTS``` uses regular expressions instead of full URLs. See https://github.com/rg3/youtube-dl#adding-support-for-a-new-site for an example.

JakubAdamWieczorek:
@yan12125 Well, it's important we check that it finds the right image. Especially on Polskie Radio the page often contains, aside from the main image, multiple thumbnails for related materials.

yan12125:
To avoid any possible false-positive, the extraction method itself should be precise. In this case, it's unlikely that ```_og_search_thumbnail``` finds a wrong image as in normal cases there's only one ```og:image``` tag in a web page.

JakubAdamWieczorek:
Agree, yet the test should not be concerned with the implementation details.

yan12125:
You want to make sure that the extractor always gives the correct thumbnail, but using a full URL does not increase the detection rate of website changes. Thumbnail URLs are constantly changing. If there's a failed test, we can't know, without further investigation, whether it's due to wrong files or just moved/renamed files. Though I don't have a concrete statistical result, moved files are more common than wrong files, and in this case, a moved/renamed file is almost the only possible reason. As a result, if there's a failed test in polskieradio in the future, we will just update the thumbnail URL. It wastes time of developers, so I recommend you to use a regular expression.

If you really want a strict check, the correct way is to check the MD5 checksum of thumbnails, which is much less unlikely to change.

JakubAdamWieczorek:
That's a fair point on the MD5, thank you. I made the suggested change. :)

youtube_dl/extractor/polskieradio.py
```diff
@@ -87,6 +90,7 @@ def _real_extract(self, url):
                 'duration': int_or_none(media.get('length')),
                 'vcodec': 'none' if media.get('provider') == 'audio' else None,
                 'timestamp': timestamp,
+                'thumbnail': thumbnail_url
```
TRox1972: Just pass in `self._og_search_thumbnail()` directly here, since it is not used anywhere else

yan12125:
Well ```md5:``` checks for the field value but not the corresponding file contents. You'll have to extend ```test/test_download.py```

JakubAdamWieczorek:
@yan12125 That seems like a good idea, sadly I'm rather time constrained now. I've gone with the regular expression approach for now.

Is this good to go now?

yan12125:
Thanks. Merged with minor fixes.

JakubAdamWieczorek:
Ah, of course. Ought to have escaped the dots. Thanks!

PR #10129, opened 2016-07-19
[aqstream] New extractor

Kagami:
- [ ] Bug fix
- [x] New extractor
- [ ] New feature

---

Some initial support for aqstream.com live streams.
It doesn't work with all streams yet but works for some.

### DIFF
--- a/youtube_dl/extractor/aqstream.py
+++ b/youtube_dl/extractor/aqstream.py
@@ -1,0 +1,112 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import base64
+import hashlib
+import random
+from .common import InfoExtractor
+from ..aes import aes_cbc_decrypt
+from ..utils import (
+    ExtractorError,
+    bytes_to_intlist,
+    intlist_to_bytes,
+    decode_pkcs7,
+    urlencode_postdata,
+    remove_start,
+)
+
+
+class AqstreamIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?aqstream\.com/[-\w]+/(?P<id>[-\w]+)'
+    _TESTS = [{
+        'url': 'http://aqstream.com/jtbc/JTBC',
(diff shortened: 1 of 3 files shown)


youtube_dl/extractor/aqstream.py
```diff
@@ -0,0 +1,112 @@
...
+            self._sort_formats(formats)
+        else:
+            formats = [{'url': link}]
+
+        descriptions = streams['info'][self._STREAMS_CAT].get(
+            'descriptions', {})
+
+        return {
+            'id': video_id,
+            'title': video_id,
```
dstftw: For live streams `self._live_title` should be used.

youtube_dl/extractor/aqstream.py
```diff
@@ -0,0 +1,112 @@
...
+    decode_pkcs7,
+    urlencode_postdata,
+    remove_start,
+)
+
+
+class AqstreamIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?aqstream\.com/[-\w]+/(?P<id>[-\w]+)'
+    _TESTS = [{
+        'url': 'http://aqstream.com/jtbc/JTBC',
```
dstftw: Any particular reason for not making it regular test without download?

youtube_dl/extractor/aqstream.py
```diff
@@ -0,0 +1,112 @@
...
+    bytes_to_intlist,
+    intlist_to_bytes,
+    decode_pkcs7,
+    urlencode_postdata,
+    remove_start,
+)
+
+
+class AqstreamIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?aqstream\.com/[-\w]+/(?P<id>[-\w]+)'
```
dstftw: Looks like at least `+` and `&` are also possible: `http://aqstream.com/mbcst/MBC-Star&`.

PR #10130, opened 2016-07-20
Optionally process sys and user config in embedded app

pincorrect:

---
- [ ] Bug fix
- [ ] New extractor
- [x] New feature

---

Explanation of your _pull request_ in arbitrary form goes here. Please make sure the description explains the purpose and effect of your _pull request_ and is worded well enough to be understood. Provide as much context and examples as possible.

The option parsing code in options.py provides a very nice feature to have both system and user config files that specify options without needing to enter them on the command line. This feature was not provided for python apps that embed youtube-dl and call main() with an argv list. My modification allows apps that embed youtube-dl to use system and user config files, if desired.

Although embedding apps that do not want this feature could easily disable it by adding '--ignore-config' as one of the supplied argv args, the behavior of existing apps could change unless so modified. To prevent this, I added an additional argument to main(), ignoreConfig, with a default value True. This ensures that existing apps will run without any change in behavior. To use the new feature, you need to specify ignoreConfig=False.

### DIFF
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -382,6 +382,8 @@ def _real_main(argv=None, ignoreConfig=True):
         'external_downloader_args': external_downloader_args,
         'postprocessor_args': postprocessor_args,
         'cn_verification_proxy': opts.cn_verification_proxy,
+        'geo_verification_proxy': opts.geo_verification_proxy,
+
     }
 
     with YoutubeDL(ydl_opts) as ydl:
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -43,7 +43,7 @@ from .extractor import gen_extractors, list_extractors
 from .YoutubeDL import YoutubeDL
 
 
-def _real_main(argv=None):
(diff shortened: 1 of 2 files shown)


youtube_dl/__init__.py
```diff
@@ -382,8 +382,6 @@ def parse_retries(retries):
         'external_downloader_args': external_downloader_args,
         'postprocessor_args': postprocessor_args,
         'cn_verification_proxy': opts.cn_verification_proxy,
-        'geo_verification_proxy': opts.geo_verification_proxy,
```
yan12125: Don't you like it?

pincorrect: Sorry, I did not mean to delete that line. Later today when I can get to my dev system, I'll correct that.

pincorrect: Odd. I did not delete that line. I never had it. But it looks like it was added 17 days ago and I only forked yesterday. I will update my branch from the latest in master, then check the diffs carefully to make sure my branch only has the change I intended.

pincorrect: This is all straightened out now.

PR #10158, opened 2016-07-24
Added random sleep for given interval range. Fixes #9930

### LINKED ISSUE #9930: Allow Random Values for --sleep-interval

singh-pratyush96:

---

- [ ] Bug fix
- [ ] New extractor
- [x] New feature

### Added implementation that enables `--sleep-interval` to take interval range input

The input can be in the form `<lower limit>to<upper limit>` and `<upper limit>to<lower limit>` (where limits are in `seconds`), while keeping the original behavior (constant sleep) intact.

```
youtube-dl https://www.youtube.com/playlist?list=PLsPUh22kYmNDnezZ15MkicBGVDm2ypthG --sleep-interval 0to33e-1
```

As discusses at #9980, it is clear that the format `--sleep-interval a-b` won't work because of interval values like `2e-1`. Therefore using something like `to` would be a better option.

Right now, I have added `ValueError` to handle the case of faulty argument, which would in turn raise `UnavailableVideoError`. Though the name of the error is irrelevant to the actual error, I still used it because the `IOError` (raised from `time.sleep()`) for negative sleep interval in current version (`2016.07.24`) also raises it.

It would be better to add `SleepIntervalError` in `utils.py` that would provide a better verbose for user upon where things went wrong, instead of just showing `ERROR: unable to download video`.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -250,6 +250,7 @@ class YoutubeDL(object):
     call_home:         Boolean, true iff we are allowed to contact the
                        youtube-dl servers for debugging.
     sleep_interval:    Number of seconds to sleep before each download.
+    max_sleep_interval:Upper bound of number of seconds for random sleep.
     listformats:       Print an overview of available video formats and exit.
     list_thumbnails:   Print a table of all thumbnails and exit.
     match_filter:      A function that gets called with the info_dict of
--- a/youtube_dl/downloader/common.py
+++ b/youtube_dl/downloader/common.py
@@ -4,6 +4,7 @@ import os
 import re
 import sys
 import time
+import random
 
 from ..compat import compat_os_name
 from ..utils import (
@@ -344,8 +345,16 @@ class FileDownloader(object):
 
         sleep_interval = self.params.get('sleep_interval')
         if sleep_interval:
-            self.to_screen('[download] Sleeping %s seconds...' % sleep_interval)
-            time.sleep(sleep_interval)
+            min_sleep_interval = sleep_interval
+            max_sleep_interval = self.params.get('max_sleep_interval') or min_sleep_interval
+
+            sleep_time = random.uniform(
+                min_sleep_interval,
+                max_sleep_interval
+            )
+
+            self.to_screen('[download] Sleeping %s seconds...' % sleep_time)
+            time.sleep(sleep_time)
 
         return self.real_download(filename, info_dict)
 
--- a/youtube_dl/downloader/common.py
+++ b/youtube_dl/downloader/common.py
@@ -346,7 +346,7 @@ class FileDownloader(object):
         sleep_interval = self.params.get('sleep_interval')
         if sleep_interval:
             min_sleep_interval = sleep_interval
-            max_sleep_interval = self.params.get('max_sleep_interval') or min_sleep_interval
+            max_sleep_interval = self.params.get('max_sleep_interval') if self.params.get('max_sleep_interval') else min_sleep_interval
 
             sleep_time = random.uniform(
                 min_sleep_interval,
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -6,8 +6,8 @@
 
 ---
 
-### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2016.07.22*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
-- [ ] I've **verified** and **I assure** that I'm running youtube-dl **2016.07.22**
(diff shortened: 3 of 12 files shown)


youtube_dl/YoutubeDL.py
```diff
@@ -250,6 +250,10 @@ class YoutubeDL(object):
     call_home:         Boolean, true iff we are allowed to contact the
                        youtube-dl servers for debugging.
     sleep_interval:    Number of seconds to sleep before each download.
+                        Accepts range for random sleep in formats:
+                        - <lower limit>to<upper limit>
+                        - <upper limit>to<lower limit>
+                            where limits are in seconds
```
dstftw: No, AtoB is definitely not how it should look like. It should be pair of options instead.

youtube_dl/YoutubeDL.py
```diff
@@ -1684,7 +1688,7 @@ def compatible_formats(formats):
             except (compat_urllib_error.URLError, compat_http_client.HTTPException, socket.error) as err:
                 self.report_error('unable to download video data: %s' % error_to_compat_str(err))
                 return
-            except (OSError, IOError) as err:
+            except (OSError, IOError, ValueError) as err:
```
dstftw: No.

youtube_dl/downloader/common.py
```diff
@@ -344,8 +345,16 @@ def download(self, filename, info_dict):
...
         if sleep_interval:
-            self.to_screen('[download] Sleeping %s seconds...' % sleep_interval)
-            time.sleep(sleep_interval)
+            sleep_interval_lower = sleep_interval.lower()
+
+            if 'to' in sleep_interval_lower:
+                lower_sleep_limit, upper_sleep_limit = map(float, sleep_interval_lower.split('to'))
+                sleep_time = random.uniform(lower_sleep_limit, upper_sleep_limit)
+            else:
+                sleep_time = float(sleep_interval)
```
dstftw: All parsing and error reporting should take place in `__init__.py`.

README.md
```diff
@@ -329,8 +329,11 @@ which means you can modify it, redistribute it or use it however you like.
     --bidi-workaround                Work around terminals that lack
                                      bidirectional text support. Requires bidiv
                                      or fribidi executable in PATH
-    --sleep-interval SECONDS         Number of seconds to sleep before each
-                                     download.
+    --sleep-interval SECONDS/RANGE   Number of seconds to sleep before each
+                                     download. Also accepts random sleep in 
+                                     format <lower limit>to<upper limit> and
+                                     <upper limit>to<lower limit>, where limits
+                                     are in seconds.
```
dstftw: This is autogenerated.

singh-pratyush96:
Instead of having two options, will having another argument like `--sleep-offset` useful?

This offset would be used to select a random interval in `[a, a+b]`, where `a` is actual sleep interval and `b` is offset. This argument will be considered only if `--sleep-interval` is passed.

dstftw:
By two options I mean `--sleep-interval` to be used as lower bound and to be mapped to `--min-sleep-interval`. The seconds option is `--max-sleep-interval` denotes the upper bound.

singh-pratyush96:
HI!

I have implemented `--max-sleep-interval`. Please take a look.

youtube_dl/downloader/common.py
```diff
@@ -344,8 +346,22 @@ def download(self, filename, info_dict):
...
+                        float(min_sleep_interval),
+                        float(max_sleep_interval)
+                    )
+                else:
+                    sleep_time = float(min_sleep_interval)
+
+                self.to_screen('[download] Sleeping %s seconds...' % sleep_time)
+                time.sleep(sleep_time)
+            except (ValueError, IOError) as err:
+                raise InvalidSleepTimeError()
```
dstftw: I've already clarified there **must be no runtime checks here.** Sleep intervals are static information thus **must be validated on start**.

dstftw:
Fix issues mentioned earlier and squash commits when done.

singh-pratyush96:
Hi again!

I hope that this does well, except that I couldn't squash the commits into one from my side (failed as you can see). I guess that these commits can be squashed while merging.

dstftw:
>Sleep intervals are static information thus **must be validated on start**.

https://www.google.ru/search?q=how+to+squash+commits

singh-pratyush96:
> Sleep intervals are static information thus *must be validated on start*.

Doesn't the type of option as `float` validates the intervals on start? If no, what other checks are needed to be conducted?

singh-pratyush96:
The solution provided for squashing was `rebase` which didn't removed the previous commits, but just added one more commit which reflected the changes from all other commits. Is this what was intended?

dstftw:
You must ensure values are denoting a valid range 0<=a<=b.
What intended is a single commit without all these non relevant mess.

singh-pratyush96:
Thanks for discussion. Will send another PR.

PR #10223, opened 2016-08-04
Added option --max-sleep-interval to support random sleep between downloads. Fixes #9930

### LINKED ISSUE #9930: Allow Random Values for --sleep-interval

singh-pratyush96:

---

- [ ] Bug fix
- [ ] New extractor
- [x] New feature

---

### Added option `--max-sleep-interval` to support random sleep between downloads.

The new option allows user to perform sleep between consecutive downloads in the interval `[a, b]` where `0 <= a <= b` and are set by the options `--sleep-interval` and `--max-sleep-interval` respectively.

#### Example usage :
```bash
youtube-dl https://www.youtube.com/playlist?list=PLsPUh22kYmNDnezZ15MkicBGVDm2ypthG --sleep-interval 10 --max-sleep-interval 20
```

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -250,6 +250,7 @@ class YoutubeDL(object):
     call_home:         Boolean, true iff we are allowed to contact the
                        youtube-dl servers for debugging.
     sleep_interval:    Number of seconds to sleep before each download.
+    max_sleep_interval:Max number of random seconds to sleep.
     listformats:       Print an overview of available video formats and exit.
     list_thumbnails:   Print a table of all thumbnails and exit.
     match_filter:      A function that gets called with the info_dict of
--- a/youtube_dl/downloader/common.py
+++ b/youtube_dl/downloader/common.py
@@ -4,6 +4,7 @@ import os
 import re
 import sys
 import time
+import random
 
 from ..compat import compat_os_name
 from ..utils import (
@@ -342,8 +343,10 @@ class FileDownloader(object):
             })
             return True
 
-        sleep_interval = self.params.get('sleep_interval')
-        if sleep_interval:
+        sleep_lower_bound = self.params.get('sleep_interval')
+        if sleep_lower_bound:
+            sleep_upper_bound = self.params.get('max_sleep_interval') if self.params.get('max_sleep_interval') else sleep_lower_bound
+            sleep_interval = random.uniform(sleep_lower_bound, sleep_upper_bound)
             self.to_screen('[download] Sleeping %s seconds...' % sleep_interval)
             time.sleep(sleep_interval)
 
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -502,6 +502,11 @@ def parseOpts(overrideArguments=None):
         '--sleep-interval', metavar='SECONDS',
         dest='sleep_interval', type=float,
         help='Number of seconds to sleep before each download.')
+    workarounds.add_option(
+        '--max-sleep-interval', metavar='SECONDS',
+        dest='max_sleep_interval', type=float,
+        help='Max number of random seconds to sleep.'
+    )
 
     verbosity = optparse.OptionGroup(parser, 'Verbosity / Simulation Options')
     verbosity.add_option(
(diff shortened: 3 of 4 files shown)


youtube_dl/downloader/common.py
```diff
@@ -342,8 +343,10 @@ def download(self, filename, info_dict):
             })
             return True
 
-        sleep_interval = self.params.get('sleep_interval')
-        if sleep_interval:
+        sleep_lower_bound = self.params.get('sleep_interval')
+        if sleep_lower_bound:
+            sleep_upper_bound = self.params.get('max_sleep_interval') if self.params.get('max_sleep_interval') else sleep_lower_bound
```
dstftw: Just `sleep_upper_bound = self.params.get('max_sleep_interval', sleep_lower_bound)`.

youtube_dl/options.py
```diff
@@ -502,6 +502,11 @@ def _scrub_eq(o):
         '--sleep-interval', metavar='SECONDS',
```
dstftw: `--min-sleep-interval` should be added for symmetry.

youtube_dl/options.py
```diff
@@ -502,6 +502,11 @@ def _scrub_eq(o):
...
         dest='sleep_interval', type=float,
         help='Number of seconds to sleep before each download.')
```
dstftw: Description should be clarified for the case when using with `--max-sleep-interval`.

singh-pratyush96:
I have made the changes.

PS: Please excuse me for the silly mistakes I make being a newbie.

dstftw:
Could you squash commits?

youtube_dl/YoutubeDL.py
```diff
@@ -249,7 +249,11 @@ class YoutubeDL(object):
     source_address:    (Experimental) Client-side IP address to bind to.
     call_home:         Boolean, true iff we are allowed to contact the
                        youtube-dl servers for debugging.
-    sleep_interval:    Number of seconds to sleep before each download.
+    sleep_interval:    Minimum number of seconds to sleep before each download.
+                       Sleep will be for a random interval if --max-sleep-interval is also passed.
+    max_sleep_interval:Max number of seconds to sleep before each download.
+                       Sleep will be for a random interval if passed along with --min-sleep-interval
+                       or --sleep-interval, otherwise ignored.
```
dstftw: These are programmatic options not command line. CLI should not even be mentioned here.

singh-pratyush96:
No, I couldn't squash. Rebasing merges all the changes into single commit but the intermediate commits still exist.

But as I saw while merging some PR, Github provides an option to squash all the commits in the PR before merging.

youtube_dl/YoutubeDL.py
singh-pratyush96: Does 'these' refer to `--max-sleep-interval` and `--min-sleep-interval`?

Where are they supposed to be mentioned in the case (`options.py`)?

dstftw:
We try to avoid GitHub merge due to merge commits it introduces. Anyway if GitHub can do that obviously this can be done by hand and you just doint it wrong.

youtube_dl/YoutubeDL.py
dstftw: They are already mentioned in `options.py`.

singh-pratyush96: In the case, shall I remove `max_sleep_interval:Max number...` comment from here?

dstftw: You should remove mention of any command line switches but not programmatic options as I've already said.

dstftw:
Cherry-picked in 065bc354894f1d35592529455d9eb685470124b9, thanks.

PR #10238, opened 2016-08-05
Change the mkv warning into an info

maxnordlund:

---

- [x] Bug fix
- [ ] New extractor
- [ ] New feature

---

Because there are so many issues with this warning, and that it works just
fine, this turns the warning about merging into an mkv to an informational
message.

It is also about not crying wolf when there is nothing to worry about.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1659,8 +1659,8 @@ class YoutubeDL(object):
                     requested_formats = info_dict['requested_formats']
                     if self.params.get('merge_output_format') is None and not compatible_formats(requested_formats):
                         info_dict['ext'] = 'mkv'
-                        self.report_warning(
-                            'Requested formats are incompatible for merge and will be merged into mkv.')
+                        self.to_screen(
+                            '[info] Requested formats are incompatible for merge and will be merged into mkv.')
                     # Ensure filename always has a correct extension for successful merge
                     filename = '%s.%s' % (filename_wo_ext, info_dict['ext'])
                     if os.path.exists(encodeFilename(filename)):


dstftw:
As info message it will be never noticed and there will be even more confusion about merging to mkv.

maxnordlund:
So instead you prefer to cry wolf and teach the users to ignore warnings? That doesn't sound right to me.

maxnordlund:
To make it clearer what I mean by crying wolf, let me explain it a bit more.

The users of this tools should be, at least a bit, familiar with the command line. As such they are most like used to warnings being a strong indication that something gone wrong. But merging the two files, which the user has requested, into one file format or another is business as usual. So by continuously emitting warnings when there is nothing wrong you teach the user to ignore _all_ warnings.

Therefore this change into an info, so that the user still sees that the file will be an mkv, but without the blinding effect.

maxnordlund:
https://github.com/rg3/youtube-dl/issues?utf8=%E2%9C%93&q=is%3Aissue%20Requested%20formats%20are%20incompatible%20for%20merge%20and%20will%20be%20merged%20into%20mkv%20

47 hits, with a fair amount being confused because of the warning and immediately closed. Just saying that this is more confusing then the info. Please humor me for a moment, where can I find the users confused over mkv files?

maxnordlund:
I'm trying to help, you know. What does @yan12125 say?

yan12125:
For me two formats merge into a third one is reasonable, so there's no need to issue a warning. However I'm highly biased as I know how containers and codecs work. I don't know how others think and what's their reaction in face of MKV files.

By the way, the number 47 is inaccurate. Most issues there are not about this warning.

maxnordlund:
Yeah, the inside of mkv vs mp4 for example is very different. But as a user I just want to watch a video, and as longs as my video player understands that format it doesn't matter. Personally I use VLC, which I think is fairly common, and it can definitely play the mkvs produced by youtube-dl.

The number is from the search query, and yes not all of those mention it explicitly as a problem. But I saw at least three that was confused and had their issues immediately closed.

yan12125:
> as longs as my video player understands that format it doesn't matter

It's a good point. Nevertheless consensus needed.

PR #10240, opened 2016-08-05
Make metadata-from-title parsing failure non-critical

nyorain:

- [x] Bug fix
- [ ] New extractor
- [ ] New feature

---

__Behaviour at the moment__
If the metadata can't be parsed from the title (no match for the string passed with --metadata-from-title) a critical error will be raised that will cause the program to exit.

__Proposed change with this pull request__
IMO failing to parse the metadata from the title is not a critical error (one simply cannot assure that all titles have the same format) and it can be really annoying that it is treated as one e.g. when downloading a playlist. Therefore i would suggest to just output a warning instead of raising an error and just write the default metadata.

### DIFF
--- a/youtube_dl/postprocessor/metadatafromtitle.py
+++ b/youtube_dl/postprocessor/metadatafromtitle.py
@@ -38,7 +38,8 @@ class MetadataFromTitlePP(PostProcessor):
         title = info['title']
         match = re.match(self._titleregex, title)
         if match is None:
-            raise MetadataFromTitlePPError('Could not interpret title of video as "%s"' % self._titleformat)
+            self._downloader.to_screen('[fromtitle] Could not interpret title of video as "%s"' % self._titleformat)
+            return [], info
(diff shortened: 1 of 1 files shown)


youtube_dl/postprocessor/metadatafromtitle.py
```diff
@@ -38,7 +38,8 @@ def run(self, info):
         title = info['title']
         match = re.match(self._titleregex, title)
         if match is None:
-            raise MetadataFromTitlePPError('Could not interpret title of video as "%s"' % self._titleformat)
+            self._downloader.to_screen('[fromtitle]  Could not interpret title of video as "%s"' % self._titleformat)
```
dstftw: Remove superfluous whitespace after bracket.

nyorain:
I'm sorry, should be fixed now.

dstftw:
Also squash commits.

nyorain:
Alright, done.

Issue #10268, opened 2016-08-09
Fix line endings handling for downloaded web pages


lkho:
### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2016.08.07*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.08.07**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### Description of your *issue*, suggested solution and other information

This error is found while trying to download a subtitle from http://d2anahhhmp1ffz.cloudfront.net/1828211116/0de60692c14fea8784203ca9f696a6be874beb52, format: SRT, encoding: utf-8, line ending: CRLF. the first 70 bytes of the response is
`b'\xef\xbb\xbf1\r\n00:00:20,173 --> 00:00:21,303\r\n(Episode 1)\r\n\r\n2\r\n00:00:50,559 --'`

However the written subtitle file has all '\r\n' replaced to '\r\r\n', which breaks the SRT format and caused some players not rendering the correct subtitle lines.

#### Suggested solution

fix the decoding of downloaded webpage content with io streams instead of bytes.decode(), so python will handle the normalization of line endings, at
`C:\Users\Me\git\youtube-dl\youtube_dl\extractor\common.py: def _webpage_read_content()`

yan12125:
Which URL are you trying to downloading?

lkho:
I was downloading `https://www.viu.com/ott/hk/zh-hk/vod/16061/`, with a custom extractor (https://github.com/rg3/youtube-dl/pull/8131) not built in the current release.

yan12125:
Fixed as  #10269 merged. Thanks!

PR #10269, opened 2016-08-09
Fix line endings handling for downloaded web pages #10268

lkho:

- [x] Bug fix
- [ ] New extractor
- [ ] New feature

---

Suggested fix for #10268

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1603,7 +1603,7 @@ class YoutubeDL(object):
                         self.to_screen('[info] Video subtitle %s.%s is already_present' % (sub_lang, sub_format))
                     else:
                         self.to_screen('[info] Writing video subtitles to: ' + sub_filename)
-                        with io.open(encodeFilename(sub_filename), 'w', encoding='utf-8') as subfile:
+                        with io.open(encodeFilename(sub_filename), 'w', encoding='utf-8', newline='') as subfile:
                             subfile.write(sub_data)
                 except (OSError, IOError):
                     self.report_error('Cannot write subtitles file ' + sub_filename)


yan12125:
Line endings should never be changed as ```_webpage_read_content``` can be used to read arbitrary binary data:
```
$ python
Python 3.5.2 (default, Jun 28 2016, 08:46:01) 
[GCC 6.1.1 20160602] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import io
>>> webpage_bytes = b'\r\n'
>>> encoding = 'utf-8'
>>> with io.TextIOWrapper(io.BytesIO(webpage_bytes), encoding, 'replace') as intxt:
...     content = intxt.read()
... 
>>> content
'\n'
```

lkho:
How about replacing '\r\n', '\r' with '\n' before writing the subtitle file, such as in https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L1594 ?

yan12125:
I guess the problem is this line: https://github.com/rg3/youtube-dl/blob/7aa589a/youtube_dl/YoutubeDL.py#L1606. Could you try it to:
```
with io.open(encodeFilename(sub_filename), 'w', encoding='utf-8', newline='') as subfile:
```
And check whether subtitles work or not?

lkho:
It works, the line endings are written untranslated. But will it cause OS compatibility issues?

yan12125:
> will it cause OS compatibility issues?

Did you mean files with UNIX line endings are written as is, so some Windows applications can't handle it?

lkho:
yes, or vice versa

yan12125:
Good question. Take subtitles for example. Assume there's a subtitle file on the Internet with UNIX line endings. I download it with my browser. I would expect it can be handled by media players on Windows without line endings conversion. I guess keeping all contents un-changed is the safest way.

lkho:
>  I would expect it can be handled by media players on Windows without line endings conversion.
good point, good players should handle that well.

yan12125:
There are lots of ```io.open``` usages in this project. Would you like to replace all occurrences?

youtube_dl/extractor/common.py
```diff
@@ -11,6 +11,7 @@
 import sys
 import time
 import math
+import io
```
yan12125: Remove this line.

lkho: fixed

yan12125:
Could you squash all commits into a single one?

lkho:
like this?

yan12125:
Merged. Thanks!

Issue #10153, opened 2016-07-24
[douyutv] API Changed


dojima:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like that [x])
- Use *Preview* tab to see how your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2016.07.24*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.07.24**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add `-v` flag to **your command line** you run youtube-dl with, copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):
```
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['http://www.douyu.com/105025', '-v']
[debug] Encodings: locale cp1252, fs mbcs, out cp437, pref cp1252
[debug] youtube-dl version 2016.07.24
[debug] Python version 3.4.4 - Windows-7-6.1.7600-SP0
[debug] exe versions: ffmpeg N-79440-g0efafc5, rtmpdump 2.4
[debug] Proxy map: {}
[DouyuTV] 105025: Downloading webpage
ERROR: Server reported error 1002: api??????
Traceback (most recent call last):
[...]

PR #10318, opened 2016-08-13
[douyutv] Use new api

steven7851:
use lapi for flv info, and html5 api for room info
#10153

### DIFF
--- a/youtube_dl/extractor/douyutv.py
+++ b/youtube_dl/extractor/douyutv.py
@@ -3,9 +3,10 @@ from __future__ import unicode_literals
 
 import hashlib
 import time
+import uuid
 from .common import InfoExtractor
-from ..utils import (ExtractorError, unescapeHTML)
-from ..compat import (compat_str, compat_basestring)
+from ..utils import (ExtractorError, unescapeHTML, sanitized_Request)
+from ..compat import (compat_str, compat_basestring, compat_urllib_parse_urlencode)
 
 
 class DouyuTVIE(InfoExtractor):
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/douyutv.py
```diff
@@ -75,74 +73,74 @@ def _real_extract(self, url):
...
         # Douyu API sometimes returns error "Unable to load the requested class: eticket_redis_cache"
         # Retry with different parameters - same parameters cause same errors
         for i in range(5):
-            prefix = 'room/%s?aid=android&client_sys=android&time=%d' % (
-                room_id, int(time.time()))
-            auth = hashlib.md5((prefix + '1231').encode('ascii')).hexdigest()
+            tt = int(time.time() / 60)
+            did = uuid.uuid4().hex.upper()
+
+            sign_content = '{room_id}{did}A12Svb&%1UUmf@hC{tt}'.format(room_id = room_id, did = did, tt = tt)
```
yan12125: How did you get this key? Better to add some comments so that others can follow it if the key is changed.

youtube_dl/extractor/douyutv.py
```diff
@@ -75,74 +73,74 @@ def _real_extract(self, url):
...
+            sign = hashlib.md5((sign_content).encode('utf-8')).hexdigest()
+
+            payload = {'cdn': 'ws', 'rate': '0', 'tt': tt, 'did': did, 'sign': sign}
+            flv_data = compat_urllib_parse_urlencode(payload)
 
-            config_page = self._download_webpage(
-                'http://www.douyutv.com/api/v1/%s&auth=%s' % (prefix, auth),
-                video_id)
+            flv_request_url = 'http://www.douyu.com/lapi/live/getPlay/%s' % room_id
+            flv_request = sanitized_Request(flv_request_url, flv_data,
```
yan12125: Use the ```query``` parameter of ```_download_webpage``` instead of ```sanitized_Request```.

youtube_dl/extractor/douyutv.py
TRox1972: `.format()` is not supported for some Python versions

yan12125: It's added in Python 2.6, so it's OK to use it as youtube-dl supports Python 2.6+ only.

youtube_dl/extractor/douyutv.py
steven7851: When I use ```_download_webpage``` with ```query``` parameter , it always returns 403 error.
I don't know how to solve, so I use ```sanitized_Request``` that is more easier to me.

remitamine: i think that `flv_data` should be used with `data` param and `content_type` dict used with `headers` param of _download_webpage.

remitamine: and the function that normally used to encode postdata is `urlencode_postdata`.

yan12125:
Sorry for being late. Could you squash all commits into a single one?

steven7851:
@yan12125 open a new PR?

yan12125:
Either pushing to this PR with ```--force``` or opening a new one is fine.

PR #10389, opened 2016-08-19
[generic] Make generic extractor use hlsnative for m3u8 files

nehalvpatel:

- [x] Bug fix
- [ ] New extractor
- [x] New feature

---

The generic extractor should use the built-in `hlsnative` utility instead of ffmpeg because it supports resuming downloads and reduces dependency on external downloaders. I've also noticed that m3u8 videos downloaded through ffmpeg are buggy, tend to desync, and cause weird playback issues.


dstftw:
No, it should not. What may produce buggy files is actually hlsnative. It also does not support live streams.

Issue #4437, opened 2014-12-11
Embedded video clip NHK...html download fails - site support requesting


eufft-z:
These are the results when trying to download the site with option verbose:
======================================
euffth@cd368ubu1404:~$ youtube-dl www.nhk.or.jp/news/html/20141210/k10013872831000.html --verbose[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['www.nhk.or.jp/news/html/20141210/k10013872831000.html', '--verbose']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2014.12.10.3
[debug] Python version 2.7.6 - Linux-3.13.0-40-generic-i686-with-Ubuntu-14.04-trusty
[debug] exe versions: avconv 9.16-6, avprobe 9.16-6, rtmpdump 2.4
[debug] Proxy map: {}
WARNING: The url doesn't specify the protocol, trying with http
[generic] k10013872831000: Requesting header
[redirect] Following redirect to http://www3.nhk.or.jp/news/html/20141210/k10013872831000.html
[generic] k10013872831000: Requesting header
WARNING: Falling back on generic information extractor.
[generic] k10013872831000: Downloading webpage
[generic] k10013872831000: Extracting information
ERROR: Unsupported URL: http://www3.nhk.or.jp/news/html/20141210/k10013872831000.html; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 554, in extract_info
    ie_result = ie.extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 242, in extract
    return self._real_extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/generic.py", line 1044, in _real_extract
    raise ExtractorError('Unsupported URL: %s' % url)
ExtractorError: Unsupported URL: http://www3.nhk.or.jp/news/html/20141210/k10013872831000.html; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
:~$ 

Comment: This is the case with any of the news video clips from NHK.
[...]

PR #10424, opened 2016-08-24
[nhk] Add extractor for VoD.

Xender:

---

- [ ] Bug fix
- [x] New extractor
- [ ] New feature

---

Extractor for NHK VoD.
Was requested by at least 2 people in #4437 and by myself.

Heavily based on Vice extractor, which also sometimes uses Ooyala player.

I haven't find a suitable test case, as all videos I've visited are available only for a limited period of time.

Example URL: http://www3.nhk.or.jp/nhkworld/en/vod/tokyofashion/20160815.html

This is my first contribution to youtube-dl. Please point me out if I missed something important.

### DIFF
--- a/youtube_dl/extractor/nhk.py
+++ b/youtube_dl/extractor/nhk.py
@@ -1,0 +1,29 @@
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class NhkVodIE(InfoExtractor):
+    _VALID_URL = r'http://www3\.nhk\.or\.jp/nhkworld/en/vod/(?P<id>.+)\.html'
+    _TESTS = [{
+        'url': 'http://www3.nhk.or.jp/nhkworld/en/vod/tokyofashion/20160815.html',
+        'info_dict': {
+            'id': 'A1bnNiNTE6nY3jLllS-BIISfcC_PpvF5',
+            'ext': 'flv',
+            'title': '[nhkworld]VOD;2009-251-2016;TOKYO FASHION EXPRESS;The Kimono as Global Fashion;en',
+        },
+        'params': {
+            'skip_download': True  # Videos available only for a limited period of time.
+        },
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        embed_code = self._search_regex(
+            r'''nw_vod_ooplayer\('movie-area', '([^']+)'\);''',
+            webpage,
+            'ooyala embed code')
+
+        return self.url_result('ooyala:' + embed_code, 'Ooyala')
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -541,6 +541,7 @@ from .nextmedia import (
 )
 from .nfb import NFBIE
 from .nfl import NFLIE
+from .nhk import NhkVodIE
 from .nhl import (
     NHLVideocenterIE,
     NHLNewsIE,


youtube_dl/extractor/nhk.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class NhkVodIE(InfoExtractor):
+    _VALID_URL = r'http://www3.nhk.or.jp/nhkworld/en/vod/(?P<id>.+)\.html'
+    # _TESTS = []  # TODO videos I've found are all available only for a limited period of time. That automatically makes such test cases obsolete. :/
```
TRox1972: If you can't find any permanent videos, add a temporary video and `'skip_download'` to `True`. This at least tests the `_VALID_URL`.

dstftw:
Provide [an example URL](https://github.com/rg3/youtube-dl/blob/master/README.md#is-the-description-of-the-issue-itself-sufficient).

youtube_dl/extractor/nhk.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        embed_code = self._search_regex(
+            r"nw_vod_ooplayer\('movie-area', '([^']+)'\);",
```
TRox1972: Single quotes consistency

youtube_dl/extractor/nhk.py
Xender: Fixed in later commit. The test downloads the metadata, though, so it will become outdated anyway.

Xender:
@dstftw - Edited first post. Will 404 in a few days anyway, though.

youtube_dl/extractor/nhk.py
```diff
@@ -0,0 +1,29 @@
...
+            'skip_download': True  # Videos available only for a limited period of time.
+        },
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        embed_code = self._search_regex(
+            r'''nw_vod_ooplayer\('movie-area', '([^']+)'\);''',
```
dstftw: Usage of single quotes only makes sense when string itself does not contain single quotes.

youtube_dl/extractor/nhk.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class NhkVodIE(InfoExtractor):
+    _VALID_URL = r'http://www3.nhk.or.jp/nhkworld/en/vod/(?P<id>.+)\.html'
```
dstftw: Dots must be escaped.

Xender:
I've tried to fix the mentioned issues and squashed the whole thing back into single commit as it started getting too much commits for something which is an initial request for a new feature/extractor.

The unsquashed version is on [Xender:nhk-vod-unsquashed](https://github.com/Xender/youtube-dl/tree/nhk-vod-unsquashed), but I quess it'd be unnecessary clutter to merge that.

youtube_dl/extractor/nhk.py
Xender: The original version had double quotes, which was commented on by @TRox1972 (I received the notification email from Github). I think he removed the comment later, though.

After all I think I prefer the triple single quotes over double quotes here anyway, as they more ambiguously indicate that the string contains special characters (be it line break or normal delimiter).

Would you consider it acceptable or do I have to change it back to double quotes or single quotes with every inner single quote escaped (later being IMHO more disruptive to read)?

PR #10443, opened 2016-08-25
[polskieradio] Add support for downloading whole programmes

JakubAdamWieczorek:
This extends the Polskie Radio (the Polish national radio) extractor to enable the user to download all the broadcasts of a single programme at once.

### DIFF
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -646,7 +646,7 @@ from .pluralsight import (
 )
 from .podomatic import PodomaticIE
 from .pokemon import PokemonIE
-from .polskieradio import PolskieRadioIE
+from .polskieradio import PolskieRadioIE, PolskieRadioProgrammeIE
 from .porn91 import Porn91IE
 from .porncom import PornComIE
 from .pornhd import PornHdIE
--- a/youtube_dl/extractor/polskieradio.py
+++ b/youtube_dl/extractor/polskieradio.py
@@ -7,12 +7,91 @@ from .common import InfoExtractor
 from ..compat import (
     compat_str,
     compat_urllib_parse_unquote,
+    compat_urlparse
 )
 from ..utils import (
     int_or_none,
     strip_or_none,
     unified_timestamp,
 )
+
+
+class PolskieRadioProgrammeIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?polskieradio\.pl/\d+(,[^/]+)?/(?P<id>\d+)'
+    _TESTS = [{
+        'url': 'http://www.polskieradio.pl/7/5102,HISTORIA-ZYWA',
+        'info_dict': {
+            'id': '5102',
+            'title': 'HISTORIA ŻYWA',
+        },
+        'playlist_mincount': 34,
+    }, {
+        'url': 'http://www.polskieradio.pl/7/4807',
+        'info_dict': {
+            'id': '4807',
+            'title': 'Vademecum 1050. rocznicy Chrztu Polski'
+        },
+        'playlist_mincount': 5
+    }, {
+        'url': 'http://www.polskieradio.pl/7/129,Sygnaly-dnia?ref=source',
+        'only_matching': True
+    }, {
+        'url': 'http://www.polskieradio.pl/37,RedakcjaKatolicka/4143,Kierunek-Krakow',
+        'info_dict': {
+            'id': '4143',
+            'title': 'Kierunek Kraków',
+        },
+        'playlist_mincount': 61
+    }, {
+        'url': 'http://www.polskieradio.pl/7,Jedynka/5102,HISTORIA-ZYWA',
+        'only_matching': True
+    }]
+
+    def _get_entries_from_page_content(self, base_url, content):
+        entries = []
+
+        articles = re.findall(
+            r'<article class="ID-(\d+) article">\s+<a href="([^"]+)"( data-layer="[^"]*")? class="[^"]*" title="([^"]+)">',
+            content)
+        for article_id, article_url, _, article_title in articles:
+            resolved_article_url = compat_urlparse.urljoin(base_url, article_url)
+            entries.append(self.url_result(
+                resolved_article_url,
+                ie='PolskieRadio',
+                video_id=article_id,
+                video_title=article_title))
+
+        return entries
+
+    @classmethod
+    def suitable(cls, url):
+        return False if PolskieRadioIE.suitable(url) else super(PolskieRadioProgrammeIE, cls).suitable(url)
+
+    def _real_extract(self, url):
+        programme_id = self._match_id(url)
+        webpage = self._download_webpage(url, programme_id)
+
+        title = self._html_search_regex(
+            r'<a href="[^"]+" id=".*_linkCategory" title="[^"]+">(.+?)</a>',
+            webpage, 'title', fatal=False)
+        description = None
+
+        entries = self._get_entries_from_page_content(url, webpage)
+
+        pages = re.findall(r'<a( href="([^"]+/Strona/)\d+")? id="[^"]+" title="strona&#32;(\d+)"', webpage)
+        page_count = max(int(page_number) for _, _, page_number in pages) if pages else 1
+
+        if page_count > 1:
+            page_url_root = next(url for _, url, _ in pages if len(url) > 0)
+            for page_number in range(2, page_count + 1):
+                page_url = page_url_root + str(page_number)
+                resolved_page_url = compat_urlparse.urljoin(url, page_url)
+                page_content = self._download_webpage(
+                    resolved_page_url, programme_id,
+                    note="Downloading page number %d" % page_number)
+                entries.extend(self._get_entries_from_page_content(url, page_content))
+
+        return self.playlist_result(entries, programme_id, title, description)
 
 
 class PolskieRadioIE(InfoExtractor):


youtube_dl/compat.py
```diff
@@ -45,6 +45,11 @@
     import urlparse as compat_urlparse
 
 try:
+    from urllib.parse import urljoin as compat_urllib_join
+except ImportError:  # Python 2
+    from urlparse import urljoin as compat_urllib_join
```
dstftw: What's the point of that? There is already `compat_urlparse.urljoin`.

dstftw:
```
  File ".\test\test_all_urls.py", line 98, in test_no_duplicates
AssertionError: PolskieRadioProgrammeIE should not match URL u'http://www.polskieradio.pl/7/5102/Artykul/1587943,Prof-Andrzej-Nowak-o-historii-nie-da-sie-myslec-beznamietnie' . That URL belongs to PolskieRadio.

```

youtube_dl/extractor/polskieradio.py
```diff
@@ -96,4 +151,6 @@ def _real_extract(self, url):
         title = self._og_search_title(webpage).strip()
         description = strip_or_none(self._og_search_description(webpage))
 
-        return self.playlist_result(entries, playlist_id, title, description)
+        result = self.playlist_result(entries, playlist_id, title, description)
+        result['_type'] = 'multi_video'
```
dstftw: What's the point of this?

JakubAdamWieczorek: The `multi_video` type seemed to me to be more appopriate for describing a list of media files that belong to a single article/broadcast, as noted in `common.py`. The `playlist` type seems to better fit the case of multiple broadcasts of a particular programme.

youtube_dl/compat.py
JakubAdamWieczorek: Oh, right. Of course. :)

youtube_dl/extractor/polskieradio.py
dstftw: `multi_video` is used for parts of complete video not for videos belonging to an article.

JakubAdamWieczorek: Oh, okay. I got this change out. :)

JakubAdamWieczorek:
Thank you for the instant review. I've made some changes to address the issues that have been raised.

youtube_dl/extractor/polskieradio.py
```diff
@@ -15,6 +16,60 @@
...
+        articles = re.findall(
+            r'<article class="ID-(\d+) article">\s+<a href="([^"]+)" class="[^"]*" title="([^"]+)">',
+            content)
+        for article in articles:
+            article_id, article_url, article_title = article
+            resolved_article_url = compat_urlparse.urljoin(base_url, article_url)
+            entries.append(self.url_result(
+                resolved_article_url,
+                ie='PolskieRadio',
+                video_id=int_or_none(article_id),
```
dstftw: Must be string.

youtube_dl/extractor/polskieradio.py
```diff
@@ -15,6 +16,60 @@
 )
 
 
+class PolskieRadioProgrammeIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?polskieradio\.pl/(?P<id>\d+/\d+)(,.*)?$'
```
dstftw: This does not match URLs like `http://www.polskieradio.pl/7/5102?foo=bar` that work fine in browser.

JakubAdamWieczorek: Fixed.

youtube_dl/extractor/polskieradio.py
JakubAdamWieczorek: Fixed.

youtube_dl/extractor/polskieradio.py
```diff
@@ -15,6 +16,63 @@
 )
 
 
+class PolskieRadioProgrammeIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?polskieradio\.pl/(?P<id>\d+/\d+)(,.*)?(\?.*)?$'
```
dstftw: Why not just `https?://(?:www\.)?polskieradio\.pl/(?P<id>\d+/\d+)`?

JakubAdamWieczorek: That'd be too ambiguous as it would match the URLs that are to be handled by the other extractor.

dstftw: `suitable` can be overridden for such cases.

JakubAdamWieczorek: Do you really think that's a cleaner solution? It seems to me that there should be a single canonical way to define which extractor handles which URLs, as is with the `_VALID_URL` constant, and I think the assumption that no two extractors collide in that regard is fundamental.

dstftw: `suitable` is exactly this canonical way. It's always feasible to handle this with regexes only.

JakubAdamWieczorek: What exactly would I put into that method that's any different from what the regular expression states?

dstftw: https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/acast.py#L64

JakubAdamWieczorek: So that makes two extractors interdependent with one another. This doesn't seem like a good practice to me when one extractor's ability to work with an URL is defined as a function of another extractor's suitability, versus a situation where each has a cleanly defined URL scope that it can process, as is the case now.

I'll make this change so that we can move forward but this is what I'm thinking.

dstftw: They are **already interdependent** since you delegate extraction from one extractor to another. Even if there were no such delegation, with complicating regexes you just express the same dependency but in less maintainable way.

JakubAdamWieczorek: Sorry, in what way "I delegated extraction from one extractor to another" before this change?

Anyway, let me know if there are any other problems or if this is good to go now. Thank you for the review!

dstftw: https://github.com/rg3/youtube-dl/pull/10443/commits/22ca2fc5c57c6db0d1b5c4e62e0df72b4a4b3b35#diff-7c05c78acf6e26a2d439e5e8e6678031R51

JakubAdamWieczorek: Okay, I see what you mean. I'd argue that's different as I could (should?) have just as well left it out.

JakubAdamWieczorek:
Are we good to go now? :-)

youtube_dl/extractor/polskieradio.py
```diff
@@ -15,6 +16,67 @@
 )
 
 
+class PolskieRadioProgrammeIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?polskieradio\.pl/(?P<id>\d+/\d+)'
```
dstftw: There is also another URL format: http://www.polskieradio.pl/7,Jedynka/5102,HISTORIA-ZYWA.

youtube_dl/extractor/polskieradio.py
```diff
@@ -15,6 +16,67 @@
...
+    }]
+
+    def _get_entries_from_page_content(self, base_url, content):
+        entries = []
+
+        articles = re.findall(
+            r'<article class="ID-(\d+) article">\s+<a href="([^"]+)" class="[^"]*" title="([^"]+)">',
+            content)
+        for article in articles:
+            article_id, article_url, article_title = article
```
dstftw: `article_id`, `article_url` and `article_title` can all be moved right under `for` eliminating superfluous `articles`.

youtube_dl/extractor/polskieradio.py
```diff
@@ -15,6 +16,67 @@
...
+    def _real_extract(self, url):
+        programme_id = self._match_id(url)
+        webpage = self._download_webpage(url, programme_id)
+
+        title = self._html_search_regex(
+            r'<h2 id=".*_lblTitle">(.+?)</h2>',
+            webpage, 'title', fatal=False)
+        description = None
+
+        page_urls = re.findall(r'<a href="(/' + programme_id + '/Strona/\d+)"', webpage)[:-1]
```
dstftw: This approach does not work when there are too much pages to be displayed at once: http://www.polskieradio.pl/7/1551,Kultura. Correct approach is to follow next link on each page.

youtube_dl/extractor/polskieradio.py
```diff
@@ -15,6 +16,67 @@
...
+        entries = self._get_entries_from_page_content(url, webpage)
+        for page_url in page_urls:
+            resolved_page_url = compat_urlparse.urljoin(url, page_url)
+            page_content = self._download_webpage(resolved_page_url, programme_id)
```
dstftw: There is no indication of page number being downloaded.

JakubAdamWieczorek: Fixed.

youtube_dl/extractor/polskieradio.py
JakubAdamWieczorek: Fixed. :)

youtube_dl/extractor/polskieradio.py
JakubAdamWieczorek: Fixed.

youtube_dl/extractor/polskieradio.py
JakubAdamWieczorek: Fixed.

JakubAdamWieczorek:
Thank you for the review, I have addressed your comments.

youtube_dl/extractor/extractors.py
```diff
@@ -646,7 +646,7 @@
 )
 from .podomatic import PodomaticIE
 from .pokemon import PokemonIE
-from .polskieradio import PolskieRadioIE
+from .polskieradio import PolskieRadioIE, PolskieRadioProgrammeIE
```
TRox1972: Enclose in parentheses

JakubAdamWieczorek: That does not seem to be the standing style in that file. Parentheses are only used when the imports are broken down into separate lines, and that's only done when the line would otherwise be too long.

JakubAdamWieczorek:
Is this good to go now? :)

JakubAdamWieczorek:
@dstftw, sorry, could we move forward with this? That would be appreciated. :)

JakubAdamWieczorek:
Hm… it looks like this had been merged in but I can't see it in `master`. Is that intentional?

yan12125:
There's something wrong. There's no ```PolskieRadioProgrammeIE``` in the latest [polskieradio.py](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/polskieradio.py)

JakubAdamWieczorek:
It seems the merge hadn't succeeded. @dstftw, could you please try again? Thank you.

JakubAdamWieczorek:
@dstftw, any idea what may have happened?

yan12125:
Hi @JakubAdamWieczorek could you open a new pull request? Seems there's no way to push the same commit again.

dstftw:
@JakubAdamWieczorek it was accidentally merged by hand and then reverted. GitHub has tracked the merge.

JakubAdamWieczorek:
@dstftw @yan12125 here's the new pull request: #10576.

Issue #10455, opened 2016-08-28
Double quote characters turn entire playlist name into "\" [bandcamp:album]


bryc:
This occurs when attempting to download an album on Bandcamp with names such as **"Entropy" EP** that contain quotes. I'd expect an output folder name such as **Entropy EP** or **\_Entropy\_ EP**, however for some reason it gives only `\` instead, is substituted as `_` when written as a folder.

Using this command line:
`youtube-dl -v -a urls.txt -i -o "%(playlist)s/%(playlist_index)s - %(title)s.%(ext)s"`

	[Bandcamp:album] entropy-ep: Downloading webpage
	[download] Downloading playlist: \
	[Bandcamp:album] playlist \: Collected 3 video ids (downloading 3 of them)
	[download] Downloading video 1 of 3
	[Bandcamp] entropy: Downloading webpage
	[debug] Invoking downloader on 'http://popplers5.bandcamp.com/download/track?enc=mp3-128&fsig=a3f285d778873f2f74921589a40c209d&id=1317106790&stream=1&ts=1472354346.0'
	[download] _\1 - Entropy.mp3 has already been downloaded
	[download] 100% of 2.74MiB
	[download] Downloading video 2 of 3
	[Bandcamp] entropy-roughmath-remix: Downloading webpage
	[debug] Invoking downloader on 'http://popplers5.bandcamp.com/download/track?enc=mp3-128&fsig=623cbb6b90d0f5935c3825026f0f6e6f&id=507884965&stream=1&ts=1472354353.0'
	[download] _\2 - Entropy (RoughMath Remix).mp3 has already been downloaded
	[download] 100% of 5.11MiB
	[download] Downloading video 3 of 3
	[Bandcamp] entropy-aaronic-remix: Downloading webpage
	[debug] Invoking downloader on 'http://popplers5.bandcamp.com/download/track?enc=mp3-128&fsig=228f38eecf6c54ce81db0b3bb774d798&id=2350421365&stream=1&ts=1472354365.0'
	[download] _\3 - Entropy (Aaronic Remix).mp3 has already been downloaded
	[download] 100% of 4.75MiB
	[download] Finished downloading playlist: \

It creates a folder called "_" from the "\" character, and puts all downloaded files there, so any affected album names will end up there. However on the very first line of the output, it gets it right: "[Bandcamp:album] entropy-ep" Why isn't it using that?

URL to the album is here: https://jstrecords.bandcamp.com/album/entropy-ep

[...]

PR #10495, opened 2016-08-30
removed all quotes from extracted webpage to solve issue 10455

PedroLopes:
Regarding issue #10455 

I had the same issue, hence **here's one possible fix**:
`webpage = webpage.replace('\\"', "")` (added to the file youtube_dl/extractor/bandcamp.py)

Note: possibly this could be done _right before any extraction is performed on the webpage_.

**Cause of the problem** is the `\"` on the extracted webpage (see below)
webpage

... makes the regex exit immediately (see below)
regexfail

**Alternative Fix**: use the regex itself to jump the quotes

**Details:**
Tested in Mac OSX
Last pull from git (as of Tue Aug 30 02:56:46 CEST 2016)

**Cloned into**: https://github.com/PedroLopes/youtube-dl 

P.s.: @bryc I think your text command should have quotes too (at least I needed those to replicate your issue), hence:
`youtube-dl -v -a urls.txt -i -o "%(playlist)s/%(playlist_index)s - %(title)s.%(ext)s"`

---

- [x] Bug fix
- [ ] New extractor
- [ ] New feature

---

### DIFF
--- a/youtube_dl/extractor/bandcamp.py
+++ b/youtube_dl/extractor/bandcamp.py
@@ -176,6 +176,7 @@ class BandcampAlbumIE(InfoExtractor):
         entries = [
             self.url_result(compat_urlparse.urljoin(url, t_path), ie=BandcampIE.ie_key())
             for t_path in tracks_paths]
+        webpage = webpage.replace('\\"', "")
         title = self._search_regex(
             r'album_title\s*:\s*"(.*?)"', webpage, 'title', fatal=False)
         return {


dstftw:
This is just another hack. Instead regex should be fixed and inner quotes should be extracted as part of title.

PedroLopes:
Hey @dstftw , here's a (possible) regex solution (there might be a nicer formulation out there):

`album_title\s*:\s*"((\\\")*(.*)?|(.*)?)"` that seems to work for both the standard case and the new case (added this to my fork (at: https://github.com/PedroLopes/youtube-dl). Works for:

1. normal case (no "\):
screen shot 2016-08-30 at 2 12 23 pm

2. special case with quotes
screen shot 2016-08-30 at 2 12 16 pm

However, when tested, although it matched correctly, when it produces an incorrect folder name:
`'Entropy_' EP` (I'm not sure where the _ is coming from)

I don't think this is a better output than with the other solution, but we can also explore where the _ is coming from and find a better/nicer fix.

PR #10517, opened 2016-09-01
[extractor] added info extractor for CBA (https://cba.fro.at)

equinox0815:

- [ ] Bug fix
- [x] New extractor
- [ ] New feature

---

This adds an info-extractor for the cultural broadcasting archive aka. CBA. It is mostly used as platform to share radio shows and snippets of shows between community radio stations in Austria. Normal users are only allowed to stream copyrighted material and therefore will not be able to download this material with youtube-dl. However there are several posts which are free to download because they don't contain any copyrighted material.
Radio Stations may have an API key which allows them to download all material for the purpose of rebraodcasts. The API key must be supplied via environment variable. This is intended to be used by automatic importing scripts of radio shows between stations.


equinox0815:
If preferred I can send a separate pull request for the change in _webpage_read_content.

youtube_dl/extractor/cba.py
```diff
@@ -0,0 +1,104 @@
...
+    }
+    _FORMATS = {
+        'audio/ogg': {'id': '1', 'ext': 'ogg', 'preference': 100},
+        'audio/mpeg': {'id': '2', 'ext': 'mp3', 'preference': 50}
+    }
+    _API_KEY = None
+
+    def __init__(self, *args, **kwargs):
+        try:
+            self._API_KEY = os.environ["CBA_API_KEY"]
```
dstftw: `--username`/`--password` should be used instead.

youtube_dl/extractor/cba.py
```diff
@@ -0,0 +1,104 @@
...
+        api_posts_url = "https://cba.fro.at/wp-json/wp/v2/posts/%s" % video_id
+        api_media_url = "https://cba.fro.at/wp-json/wp/v2/media?media_type=audio&parent=%s" % video_id
+
+        title = 'unknown'
+        description = ''
+        formats = []
+
+        posts_result = self._download_json(api_posts_url, video_id, 'query posts api-endpoint', 'unable to query posts api-endpoint')
+        try:
+            title = clean_html(posts_result['title']['rendered'])
```
dstftw: Title should be mandatory.

youtube_dl/extractor/cba.py
```diff
@@ -0,0 +1,104 @@
...
+                                           'unable to qeury media api-endpoint%s' % api_key_str)
+        for media in media_result:
+            try:
+                url = media['source_url']
+                if url == "":
+                    continue
+
+                ft = media['mime_type']
+                f = { 'url': url, 'format': ft, 'format_id': self._FORMATS[ft]['id'], 'preference': self._FORMATS[ft]['preference'] }
+                self._add_optional_parameter(f, 'filesize', media['media_details'], 'filesize')
```
dstftw: `utils.int_or_none` or `utils.float_or_none` instead.

youtube_dl/extractor/cba.py
```diff
@@ -0,0 +1,104 @@
...
+        if self._API_KEY:
+            api_key_str =  " (using API_KEY '%s')" % self._API_KEY
+            api_media_url = update_url_query(api_media_url, {'c': self._API_KEY})
+
+        media_result = self._download_json(api_media_url, video_id, 'query media api-endpoint%s' % api_key_str,
+                                           'unable to qeury media api-endpoint%s' % api_key_str)
+        for media in media_result:
+            try:
+                url = media['source_url']
+                if url == "":
```
dstftw: Pythonic is `if not url:`.

youtube_dl/extractor/cba.py
```diff
@@ -0,0 +1,104 @@
...
+        api_key_str = " (without API_KEY)"
+        if self._API_KEY:
+            api_key_str =  " (using API_KEY '%s')" % self._API_KEY
+            api_media_url = update_url_query(api_media_url, {'c': self._API_KEY})
+
+        media_result = self._download_json(api_media_url, video_id, 'query media api-endpoint%s' % api_key_str,
+                                           'unable to qeury media api-endpoint%s' % api_key_str)
+        for media in media_result:
+            try:
+                url = media['source_url']
```
dstftw: Shouldn't fail if key `source_url` does not exist.

youtube_dl/extractor/cba.py
```diff
@@ -0,0 +1,104 @@
...
+                f = { 'url': url, 'format': ft, 'format_id': self._FORMATS[ft]['id'], 'preference': self._FORMATS[ft]['preference'] }
+                self._add_optional_parameter(f, 'filesize', media['media_details'], 'filesize')
+                self._add_optional_parameter(f, 'abr', media['media_details'], 'bitrate', lambda x: x/1000)
+                self._add_optional_parameter(f, 'asr', media['media_details'], 'sample_rate')
+
+                formats.append(f)
+            except KeyError:
+                pass
+
+        if len(formats) == 0:
```
dstftw: `if not formats:`.

youtube_dl/extractor/common.py
```diff
@@ -438,6 +439,10 @@ def _guess_encoding_from_content(content_type, webpage_bytes):
     def _webpage_read_content(self, urlh, url_or_request, video_id, note=None, errnote=None, fatal=True, prefix=None, encoding=None):
         content_type = urlh.headers.get('Content-Type', '')
         webpage_bytes = urlh.read()
+        webpage_bytes, bom_enc = parse_strip_bom(webpage_bytes)
```
dstftw: This change does not look related to this PR. There is also no explanation what this changes is supposed to solve and lacks concrete examples and tests.

equinox0815: You are right it is not directly related to the PR. That's what i meant with the above comment about the separate pull. 
The reason i need this is that the site sends unicode POMs which lead to a json parser error. My first attempt (see ed10b1c) was to strip the prefix using  `transform_source` but it turned out to be trickier than i thought. The best i could come up with was the function `strip_bom_utf8`. But apperently this does not work with python 3.2.
This latest change is an attempt to solve the issue in a more general way. As i think i'm not the only one that has to work around json parser errors and unicode POMs. I'm open to suggestions as to how to fix the problem.

youtube_dl/extractor/cba.py
equinox0815: Ok. Haven't found any other extractors that need API keys - apparently i didn't look well enough. Do you mind if it falls back to the environment varaible if `--password` ist empty? 
This makes it imho much easier to integrate into bigger applications.

youtube_dl/extractor/common.py
dstftw: 1. Explicit `encoding` may be added to `_parse_json` and forwarded there from `_download_json`. With this one will be able to specify encoding manually, `utf-8-sig` in this case.
2. Drop BOM in `_parse_json` leaving webpage intact.

youtube_dl/extractor/cba.py
dstftw: AFAIR there are no extractors that use personal API keys since there is almost always a way to workaround mimicking browser's behavior without it.
It should not use env at all. Any integration should use generic interface but not a zoo of system wide env variables.

youtube_dl/extractor/cba.py
equinox0815: So you prefer KeyError exceptions over a title set to unknown?

dstftw: Yes. Though you may provide fallback sources for it (if any). But overall title should be fatal. `unknown` hardcode may be used if absence of title is allowed by source provider or in case there is no title is expected at all (in this case video id is usually used).

youtube_dl/extractor/cba.py
equinox0815: Hm ok - i will drop the environment variable.
Concerning the API keys: this is the peferred way by the people who run the platform. Of course there would be the possibility to mimick the browser behaviour but at least for us (a community radio station which has an API key) it is of course much less error prone to just use the key.
I will take a look into how to download the file via the flash player. Don't know whether the platfrom operators will like this though.

dstftw: youtube-dl is wide targeted. Absence of any authentication is preferred since not everybody may have/want to bother to obtain an API key.

youtube_dl/extractor/common.py
equinox0815: Thanks! using `utf-8-sig` is indeed **much** easier. Didn't know it existed and searching the web didn't bring it up... :(

equinox0815:
Besides the download-without-api key feature i think i've fixed all issues that came up at your review. Would be nice if you could review the PR again.
Should i do a new PR with all changes squashed into one commit?

equinox0815:
Hi again, haven't heard anything for a while. Are there still things that i should adapt? Shall i do a new PR with all the changes stashed into one commit?

PR #10536, opened 2016-09-02
[drtv] Add support for dr.dk/nyheder

C4K3:

---

- [ ] Bug fix
- [x] (sort of) New extractor
- [ ] New feature

---

Edit the drtv extractor to also support urls under dr.dk/nyheder.

The only difference is that the video player is loaded differently, and certain metadata (title and description) is not available under dr.dk/mu, so make it by default get that from some of the html meta tags.

### DIFF
--- a/youtube_dl/extractor/drtv.py
+++ b/youtube_dl/extractor/drtv.py
@@ -5,13 +5,14 @@ from .common import InfoExtractor
 from ..utils import (
     ExtractorError,
     parse_iso8601,
+    remove_end,
 )
 
 
 class DRTVIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?dr\.dk/tv/se/(?:[^/]+/)*(?P<id>[\da-z-]+)(?:[/#?]|$)'
+    _VALID_URL = r'https?://(?:www\.)?dr\.dk/(?:tv/se|nyheder)/(?:[^/]+/)*(?P<id>[\da-z-]+)(?:[/#?]|$)'
 
-    _TEST = {
+    _TESTS = [{
         'url': 'https://www.dr.dk/tv/se/boern/ultra/panisk-paske/panisk-paske-5',
         'md5': 'dc515a9ab50577fa14cc4e4b0265168f',
         'info_dict': {
@@ -23,7 +24,20 @@ class DRTVIE(InfoExtractor):
             'upload_date': '20150322',
             'duration': 1455,
         },
-    }
+        'skip': 'Video is no longer available',
+    }, {
+        'url': 'https://www.dr.dk/nyheder/indland/live-christianias-rydning-af-pusher-street-er-i-gang',
+        'md5': '2ada5074f9e79afc0d324a8e9784d850',
+        'info_dict': {
+            'id': 'christiania-pusher-street-ryddes-drdkrjpo',
+            'ext': 'mp4',
+            'title': 'LIVE Christianias rydning af Pusher Street er i gang',
+            'description': '- Det er det fedeste, der er sket i 20 år, fortæller christianit til DR Nyheder.',
+            'timestamp': 1472800279,
+            'upload_date': '20160902',
+            'duration': 131.4,
+        }
+    }]
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/drtv.py
```diff
@@ -34,17 +47,28 @@ def _real_extract(self, url):
...
         programcard = self._download_json(
             'http://www.dr.dk/mu/programcard/expanded/%s' % video_id,
             video_id, 'Downloading video JSON')
         data = programcard['Data'][0]
 
-        title = data['Title']
-        description = data['Description']
+        title = self._html_search_regex(
+            r'property="og:title"[^>]*content="([^">]+)"',
+            webpage, 'title', default=data['Title'])
```
dstftw: `_og_search_title` instead. Why is that preferred over the old source?

youtube_dl/extractor/drtv.py
```diff
@@ -34,17 +47,28 @@ def _real_extract(self, url):
...
-            r'data-(?:material-identifier|episode-slug)="([^"]+)"',
-            webpage, 'video id')
+        if ("nyheder" in url) and ("tv/se" not in url):
+            video_id = self._search_regex(
+                r'data-resource="[^>"]+mu/programcard/expanded/([^>"]+)"',
+                webpage, 'video id')
+        else:
+            video_id = self._search_regex(
+                r'data-(?:material-identifier|episode-slug)="([^"]+)"',
+                webpage, 'video id')
```
dstftw: Just put both regexes under the same `_search_regex`.

youtube_dl/extractor/drtv.py
```diff
@@ -9,9 +9,9 @@
 
 
 class DRTVIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?dr\.dk/tv/se/(?:[^/]+/)*(?P<id>[\da-z-]+)(?:[/#?]|$)'
+    _VALID_URL = r'https?://(?:www\.)?dr\.dk/(tv/se|nyheder)/(?:[^/]+/)*(?P<id>[\da-z-]+)(?:[/#?]|$)'
```
dstftw: Don't capture groups you does not use.

youtube_dl/extractor/drtv.py
C4K3: It's preferred because some videos (for example the one in the test I added) in their /mu data json file make the title the same as the video ID, rather than the one published on the site, and the description as just a blank string. Whereas every video I've looked at so far the og:title and description have properly matched the video.

dstftw:
Also squash commits when done.

youtube_dl/extractor/drtv.py
```diff
@@ -35,16 +48,17 @@ def _real_extract(self, url):
                 'Video %s is not available' % video_id, expected=True)
 
         video_id = self._search_regex(
-            r'data-(?:material-identifier|episode-slug)="([^"]+)"',
+            r'(?:data-(?:material-identifier|episode-slug)="|data-resource="[^>"]+mu/programcard/expanded/)([^"]+)"',
```
dstftw: What I actually meant is to put them in tuple.

youtube_dl/extractor/drtv.py
```diff
@@ -35,16 +48,18 @@ def _real_extract(self, url):
...
             webpage, 'video id')
 
         programcard = self._download_json(
             'http://www.dr.dk/mu/programcard/expanded/%s' % video_id,
             video_id, 'Downloading video JSON')
         data = programcard['Data'][0]
 
-        title = data['Title']
-        description = data['Description']
+        title = self._og_search_title(webpage) or data['Title']
```
dstftw: For tv/se `og:title` contains unnecessary suffix.

C4K3: Oh, you're absolutely right.

But the data['Title'] still uses ugly naming at least under nyheder. Should I make it prefer one or the other depending on which part of the site it's downloading from, or should I use string.replace to remove the suffix?

dstftw: Just cut it with `utils.remove_end`.

dstftw:
Cherry-picked in 5e9e3d0f6bf2055c557f360758d6d7eb146edcba, thanks.

C4K3:
Cool, thanks @dstftw for taking your time to review this.

PR #10554, opened 2016-09-04
Provide MPD url and segment durations in JSON

wiiaboo:
- [x] New feature

---

mpv is planning to support segmented dash support that would need potentially both these commits.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -1549,6 +1549,16 @@ class InfoExtractor(object):
                 initialization = segment_list.find(_add_ns('Initialization'))
                 if initialization is not None:
                     ms_info['initialization_url'] = initialization.attrib['sourceURL']
+                segment_timeline = segment_list.find(_add_ns('SegmentTimeline'))
+                if segment_timeline is not None:
+                    s_e = segment_timeline.findall(_add_ns('S'))
+                    if s_e:
+                        ms_info['total_number'] = 0
+                        ms_info['segment_durations'] = []
+                        for s in s_e:
+                            r = int(s.get('r', 0))
+                            ms_info['total_number'] += 1 + r
+                            ms_info['segment_durations'].append(int(s.attrib['d']))
             else:
                 segment_template = element.find(_add_ns('SegmentTemplate'))
                 if segment_template is not None:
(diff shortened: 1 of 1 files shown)


yan12125:
Hi, could you split the two changes into two pull requests? The idea of ```dash_mpds``` is not so good. I suggest adding a new field, maybe ```manifest_url```, to ```formats```, so that f4m and hls can use it as well. Also, don't forget to [document it](https://github.com/rg3/youtube-dl/blob/0def758/youtube_dl/extractor/common.py#L85-L145).

remitamine:
i already proposed to return the mpd_url in #8848 and #9109 the will return the url according to the standard and without in introducing a new field in the info dict.

yan12125:
I see both PRs closed. Are they incomplete yet?

remitamine:
no, #9109 replaces #8848 with more improvments, but maintaining it for a long time was difficult(handle lots of merge conlicts) and also some of the code in the pull request was introduced in the master(handling `$Time` in media_template) branch which make some of duplication in the functionality.

wiiaboo:
I don't see the dash mpd urls or segment duration anywhere on -J from master.

yan12125:
> mpv needs at least one of these to support http_dash_fragments.

Then could you remove the line ```dash_mpds```? Other changes are less controversial.

remitamine:
i don't think that it's a good idea for mpv to depend on the segement urls in the info dict, in my opinion the only information that should be returned about dash formats is the url of the mpd manifest.

yan12125:
@remitamine I guess your concern is that fields like ```segment_urls``` are for youtube-dl internal uses only and may change at any time? Then the way to go is #9109.

remitamine:
> I guess your concern is that fields like segment_urls are for youtube-dl internal uses only and may change at any time?

yes.
> Then the way to go is #9109.

it should be created again as there was a lot of changes since the #9109 was created, for now i think that #9937 is more important to be processed than #9109 , if it's acceptable i can push the code to return the mpd url directly to master branch, than i will try to open another PR for the other changes after #9937 merged.

yan12125:
OK I'll look at #9937 first. By the way, I've opened a milestone at https://github.com/rg3/youtube-dl/milestone/6 to track all relevant pull requests and issues. Hope that reduces some headaches.

dstftw:
Will be available in the next version as:
```json
"formats" : {
  ...
  "manifest_url": "http://manifest/url.mpd",
  "fragments": [
    {
      "url": "http://fragment1/url",
      "duration": 0,
    },
    {
      "url": "http://fragment2/url",
      "duration": 0,
    },
    ...
  ]
  ...
}
```

PR #10567, opened 2016-09-05
Fragments interface generalization

dstftw:
- [x] New feature

---

Closes #10554.

This PR introduces a generalization of an interface for exposing fragments data for fragmented media.
The general idea is to generalize all fragment based protocols to return data in this format to be consumed by clients in a uniform way.

Pros:
* Finally fragments interface for clients (`segment_urls` is an implementation detail since it's not documented)
* Transparency for client applications: no need to parse manifests, you've just got final list of fragments with complete fragments' URLs, all available metadata and possibly some additional data needed for downloading, e.g. HTTP headers (to be discussed).
* Possibility to merge all dumb *download-fragment-and-concatenate* fragment based downloaders into single one or at least generalize in future

Cons:
* Possible growth in size of `-J` due to storing full fragments' URLs (especially for YouTube). This one may be worked around by introducing `base_fragment_url` and assuming in this case all fragments to be relative to `base_fragment_url`.

Note this is the first step. I've intentionally tried to make this PR to be as small and single-featured as possible. The seconds step would be to eliminate `segment_urls` from dashsegments downloader and using `fragments` directly.
I'm aware of some code duplication - this is intentional to ease the second step refactoring.
`url` for dashsegments is not really a manifest URL currently, it will after the second step.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -86,7 +86,9 @@ class InfoExtractor(object):
                     from worst to best quality.
 
                     Potential fields:
-                    * url        Mandatory. The URL of the video file
+                    * url        Mandatory. The URL of the video file or URL of
+                                 the manifest file in case of fragmented media
+                                 (DASH, hls, hds).
                     * ext        Will be calculated from URL if missing
                     * format     A human-readable description of the format
                                  ("mp4 container with h264/opus").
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/common.py
```diff
@@ -1541,54 +1548,67 @@ def is_drm_protected(element):
 
         def extract_multisegment_info(element, ms_parent_info):
             ms_info = ms_parent_info.copy()
+
+            # As per [1, 5.3.9.2.2] SegmentList and SegmentTemplate share some
+            # common attributes and elements.  We will only extract relevant
```
TRox1972: 'extract what is relevant to us'

remitamine:
how it will handle this cases:
- live streams(update `fragments` frequently).
- hls formats(additional requests to extract `fragments`).
- fragment post processing(AES-128 decryption for m3u8 fragments).

dstftw:
1. Handling live streams with `-J` does not make much sense since a lot of unrelated data may be downloaded on each run. So I don't expect it to be used for live streams. It can simply return fragments of current window (manifest specific identifiers may be included in fragment data if needed) everything else is up to the client.
2. By default it may extract only requested format's fragments, that's enough in most cases. There also may be an option to force fragments extraction for all formats.
3. Key URI, encryption method and other data can be included in fragments/fragment data. Client decides then how he is going to used it.

dstftw:
Cherry-picked.

PR #10576, opened 2016-09-06
[polskieradio] Add support for downloading whole programmes.

JakubAdamWieczorek:
A continuation of #10443.

This extends the Polskie Radio (the Polish national radio) extractor to enable the user to download all the broadcasts of a single programme at once.

### DIFF
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -667,7 +667,7 @@ from .pluralsight import (
 )
 from .podomatic import PodomaticIE
 from .pokemon import PokemonIE
-from .polskieradio import PolskieRadioIE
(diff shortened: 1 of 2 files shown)


dstftw:
Some playlist don't work: http://www.polskieradio.pl/10,czworka/214,muzyka.
Playlist extraction should use generators or `OnDemandPagedList/InAdvancePagedList` in order to stop appropriately when `--playlist-end` is provided since there are large playlists http://www.polskieradio.pl/43/279 available.

JakubAdamWieczorek:
@dstftw, I'd planned on looking at it tomorrow, but thank you very much for taking it on. :)

JakubAdamWieczorek:
@dstftw, also, your solution seems much nicer!

PR #10597, opened 2016-09-08
Fix _VALID_URLs even further

stepshal:

---

- [ ] Bug fix
- [ ] New extractor
- [ ] New feature

---

### DIFF
--- a/youtube_dl/extractor/openload.py
+++ b/youtube_dl/extractor/openload.py
@@ -13,7 +13,7 @@ from ..utils import (
 
 
 class OpenloadIE(InfoExtractor):
-    _VALID_URL = r'https://openload.(?:co|io)/(?:f|embed)/(?P<id>[a-zA-Z0-9-_]+)'
+    _VALID_URL = r'https?://openload\.(?:co|io)/(?:f|embed)/(?P<id>[a-zA-Z0-9-_]+)'
 
     _TESTS = [{
         'url': 'https://openload.co/f/kUEfGclsU9o',
--- a/youtube_dl/extractor/abc.py
+++ b/youtube_dl/extractor/abc.py
@@ -13,7 +13,7 @@ from ..utils import (
 
 class ABCIE(InfoExtractor):
     IE_NAME = 'abc.net.au'
-    _VALID_URL = r'https?://www\.abc\.net\.au/news/(?:[^/]+/){1,2}(?P<id>\d+)'
+    _VALID_URL = r'https?://(?:www\.)?abc\.net\.au/news/(?:[^/]+/){1,2}(?P<id>\d+)'
 
     _TESTS = [{
         'url': 'http://www.abc.net.au/news/2014-11-05/australia-to-staff-ebola-treatment-centre-in-sierra-leone/5868334',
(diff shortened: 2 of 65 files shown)


yan12125:
Hi @stepshal did you use any checker to find those bugs? It would be great to integrate it into youtube-dl.

stepshal:
No, i just used this command: `grep -r "_VALID_URL = r'https" youtube-dl > urls`, that search and prints lines with this pattern from directory and then looked at urls, i probably be able to listed only `something.com` without backslash, but i don't know how.

youtube_dl/extractor/openload.py
```diff
@@ -13,7 +13,7 @@
 
 
 class OpenloadIE(InfoExtractor):
-    _VALID_URL = r'https://openload.(?:co|io)/(?:f|embed)/(?P<id>[a-zA-Z0-9-_]+)'
+    _VALID_URL = r'https://openload\.(?:co|io)/(?:f|embed)/(?P<id>[a-zA-Z0-9-_]+)'
```
stepshal: Should `r'https://` be `r'https?://'? Or if the site is only in https, thats how it should be?

TRox1972:
Another improvement could be to replace `www\.` with `(?:www\.)`. Furthermore, wouldn't it be good to have a script that can check for things like this? Could be integrated with PRs.

yan12125:
> Furthermore, wouldn't it be good to have a script that can check for things like this?

It's great to have, though I'm not sure whether checking regular expressions is easy or not.

youtube_dl/extractor/openload.py
yan12125: ```https?://``` is better in this case. openload supports both HTTP and HTTPS.

stepshal:
@TRox1972 are you shure it's not `(?:www\.)?` ?

stepshal:
@yan12125  For now i fixed all cases, check please.

dstftw:
```
...F.............
  File ".\test\test_all_urls.py", line 98, in test_no_duplicates
AssertionError: True is not false : VimeoUserIE should not match URL 'http://vimeo.com/moogaloop.swf?clip_id=2539741' . That URL belongs to Vimeo.

Ran 17 tests in 5.449s

FAILED (failures=1)
```

stepshal:
@dstftw don't know that's the problem, so vimeo fixes removed for now.

stepshal:
@dstftw look into it again please.

youtube_dl/extractor/nhk.py
```diff
@@ -4,7 +4,7 @@
 
 
 class NhkVodIE(InfoExtractor):
-    _VALID_URL = r'https?://www3\.nhk\.or\.jp/nhkworld/en/vod/(?P<id>.+?)\.html'
+    _VALID_URL = r'https?://(?:www3\.)?nhk\.or\.jp/nhkworld/en/vod/(?P<id>.+?)\.html'
```
dstftw: `nhk.or.jp` does not even resolves.

stepshal:
@dstftw thiught about it, deleted.

dstftw:
Cherry-picked in 925194022cd661747771e58bad41e5f7ae118999, thanks.

PR #10610, opened 2016-09-10
add libsecret support

dffischer:
Instead of asking a user to type the password, libsecret can be asked for a password corresponding to a username, when present.

Several web browsers, most notably Firefox and Chrome, support storing passwords via libsecret out of the box or via Add-Ons. When these facilities are in use, the password is most probably already available.

### DIFF
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -129,7 +129,15 @@ def _real_main(argv=None):
     if opts.usetitle and opts.useid:
         parser.error('using title conflicts with using video ID')
     if opts.username is not None and opts.password is None:
-        opts.password = compat_getpass('Type account password and press [Return]: ')
+        try:
+            from gi import require_version
+            require_version('Secret', '1')
+            from gi.repository import Secret
+            opts.password = Secret.Service.get_sync(
+                    Secret.ServiceFlags.NONE, None).lookup_sync(None, {
+                        "username": opts.username}).get_text()
+        except:
+            opts.password = compat_getpass('Type account password and press [Return]: ')
     if opts.ratelimit is not None:
         numeric_limit = FileDownloader.parse_bytes(opts.ratelimit)
         if numeric_limit is None:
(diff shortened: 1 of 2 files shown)


youtube_dl/__init__.py
```diff
@@ -129,7 +129,15 @@ def _real_main(argv=None):
...
     if opts.username is not None and opts.password is None:
-        opts.password = compat_getpass('Type account password and press [Return]: ')
+        try:
+            from gi import require_version
+            require_version('Secret', '1')
+            from gi.repository import Secret
+            opts.password = Secret.Service.get_sync(
+                    Secret.ServiceFlags.NONE, None).lookup_sync(None, {
+                        "username": opts.username}).get_text()
+        except:
```
yan12125: Don't use bare ```except:```

yan12125:
I feel it better to have an option for libsecret rather than use it implicitly. I'm not sure how other developers think. Lets wait.

youtube_dl/__init__.py
```diff
@@ -129,7 +129,15 @@ def _real_main(argv=None):
...
         parser.error('using title conflicts with using video ID')
     if opts.username is not None and opts.password is None:
-        opts.password = compat_getpass('Type account password and press [Return]: ')
+        try:
+            from gi import require_version
+            require_version('Secret', '1')
+            from gi.repository import Secret
+            opts.password = Secret.Service.get_sync(
+                    Secret.ServiceFlags.NONE, None).lookup_sync(None, {
+                        "username": opts.username}).get_text()
```
dstftw: This restricts password to the only single one while it should be on per extractor basis. As a result this breaks credentials input mechanism completely, one are enable to have different credentials per extractor anymore.

dstftw:
As already mentioned by @yan12125, it should not touch the current behavior and should only activate explicitly when an option is specified. Using of `--password` is explicit user's intention to provide password himself either on command line or interactive.

PR #10616, opened 2016-09-11
[tvanouvelles] New extractor

jnbdz:

---
- [ ] Bug fix
- [x] New extractor
- [ ] New feature

---

I am just sending youtube-dl a new extractor. It's similar to AlJazeera.


youtube_dl/extractor/tvanouvelles.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class TVANouvellesIE(InfoExtractor):
+    _VALID_URL = r'https?://(www\.|)tvanouvelles\.(ca|com|qc)/.*.?/(?P<id>[^/]+)'
```
dstftw: Don't capture groups you don't use. `(www\.|)` that's incorrect. `.*.?` does not make much sense.

youtube_dl/extractor/tvanouvelles.py
```diff
@@ -0,0 +1,30 @@
...
+
+    _TEST = {
+        'url': 'http://www.tvanouvelles.ca/videos/5117035533001',
+        'info_dict': {
+            'id': '5117035533001',
+            'ext': 'mp4',
+            'title': 'L\u2019industrie du taxi d\xe9nonce l\u2019entente entre Qu\xe9bec et Uber: explications',
```
dstftw: Use original unicode characters.

youtube_dl/extractor/tvanouvelles.py
```diff
@@ -0,0 +1,30 @@
...
+        'add_ie': ['BrightcoveNew'],
+        'skip': 'Not accessible from Travis CI server',
+    }
+    BRIGHTCOVE_URL_TEMPLATE = 'http://players.brightcove.net/1741764581/default_default/index.html?videoId=%s'
+
+    def _real_extract(self, url):
+        program_name = self._match_id(url)
+        webpage = self._download_webpage(url, program_name)
+        brightcove_id = self._search_regex(
+            r'data-video-id\=(.+[0-9]?)', webpage, 'brightcove id')
```
dstftw: That's not correct.

dstftw:
Squash commits.

youtube_dl/extractor/tvanouvelles.py
jnbdz: Could you elaborate?

youtube_dl/extractor/tvanouvelles.py
jnbdz: I am not very familiar to Python regex. I just wanted to give a possibility that it might have the www. or not. for .*.? was to ignore everything until you see the last /. The reason is because this can also be the URL: [http://www.tvanouvelles.ca/2016/09/11/les-etats-unis-commemorent-le-traumatisme-du-11-septembre](http://www.tvanouvelles.ca/2016/09/11/les-etats-unis-commemorent-le-traumatisme-du-11-septembre).

youtube_dl/extractor/tvanouvelles.py
jnbdz: It's UTF-8 and my editor was complaining because Python if I understand correctly is ASCII. I would of had to set it as UTF-8. Should have I done that?

youtube_dl/extractor/tvanouvelles.py
jnbdz: I just want to say I am not familiar to Python. So any suggestion are very much welcome.

TRox1972: The regex should be improved

jnbdz:
@dstftw For some reason, it's failing on travis-ci. But I did test on my machine and everything seems to be fine. I was able to download videos from TVA Nouvelles, without any issues.

I also changed the regex like it was recommended. I hope it's good enough.

dstftw:
>Squash commits.

PR #10645, opened 2016-09-14
Apply fixups to individual streams (fixes #10644/#9913)

### LINKED ISSUE #10644: Conversion failed because of missing aac_adtstoasc/FFmpegFixupM3u8PP
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.09.11.1**

When only a single format uses the m3u8 protocol, the fixup fails to be applied.

Example failing URL: http://www.tf1.fr/tmc/quotidien-avec-yann-barthes/videos/didier-deschamps-souhaite-bonne-chance-a-quotidien.html

johnpatch:
This isn't very pretty as it calls post_process multiple times, so _pps needs to be cleared temporarily to avoid duplicate processing. Maybe someone has a more elegant solution?

In the case where `info_dict['protocol'] == 'm3u8'` in `process_info`, FFmpegFixupM3u8PP may be applied multiple times: once for each stream, and once for the whole file. I don't know if it can happen, or if it's a problem if it happens. Perhaps we don't need to apply it to the whole file.

Note that in the example from #10644, `new_info['protocol'] == 'm3u8'` for the audio stream but `info_dict['protocol']` is not set, so that problem does not appear.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1486,6 +1486,70 @@ class YoutubeDL(object):
             subs[lang] = f
         return subs
 
+    def add_fixup_post_processors(self, filename, info_dict, protocol_only=False):
+        # Fixup content
+        fixup_policy = self.params.get('fixup')
+        if fixup_policy is None:
+            fixup_policy = 'detect_or_warn'
+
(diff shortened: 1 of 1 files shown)


xantares:
+1, this fixes my bug too, someone for a review ?

Issue #10174, opened 2016-07-27
[Question] How can i disable some hosts


yas9999:
Hello,
i wonder how can i disable some hosts to be downloaded (p0rn) for my case

Best regards,

yan12125:
It's currently impossible with youtube-dl. See #6497 and #6501.

yas9999:
thank you bro, but i thought that i can disable some hosts by using /etc/youtube-dl.conf file !
i'm using HTML GUI from so i guess i can disable some by using php (array function )

regards,

yan12125:
Disable extractors via configuration files is not implemented yet. So far you can remove some lines from ```youtube_dl/extractor/extractors.py```

phihag:
You can disable videos restricted to age 18 and above by passing in `--age-limit 18` in `/etc/youtube-dl.conf`. That is by far the most sensible solution.

phihag:
You can disable videos restricted to age 18 and above by passing in `--age-limit 17` in `/etc/youtube-dl.conf`. That is by far the most sensible solution, and does cover pornographic videos from hosts that are not exclusively distributing pornography such as YouTube. It even allows you a more fine-grained control. For instance, if you deploy youtube-dl in a school with kids as young as 13, pass in `--age-limit 12` to ensure that they don't get videos unsuitable for their age.

yas9999:
thank you @yan12125  & @phihag,
porn was just an example i may need to remove for example vimeo !
so if there is no bugs i will edit youtube_dl/extractor/extractors.py

yan12125:
That depends. If you download youtube-dl via ```wget``` or ```curl```, you need to extract all files with ```unzip```, modify them and re-package all files with ```zip``` again. If you use pip, ```pip show youtube-dl | grep Location``` will tell you where it is.

yas9999:
@yan12125
i still don't get the idea ! there is no zip files :
$ pip show youtube-dl | grep Location
Location: /usr/local/lib/python2.7/dist-packages
$ ls /usr/local/lib/python2.7/dist-packages

backports                                      functools32
backports.ssl_match_hostname-3.5.0.1.egg-info  functools32-3.2.3_2.egg-info
cached_property-1.3.0.dist-info                ipaddress-1.0.17.dist-info
cached_property.py                             ipaddress.py
cached_property.pyc                            ipaddress.pyc
compose                                        jsonschema
docker                                         jsonschema-2.5.1.dist-info
[...]

Issue #6497, opened 2015-08-08
Move porn sites extractors to a separate branch or repo or simple way to delete them


remitamine:
some way to make the presence of this type of extractors optional and not installed by default.

dstftw:
I don't see much sense in this since this will result in additional headache and bogus bugreports. There is a `--age-limit` option that can disable sites for adults.

remitamine:
i know about --age-limit.
but i think it's better to get them in a seperate place and not shipped by default.

dstftw:
No. Not shipping them by default will definitely result in tons of bogus reports for "unsupported" URLs, that are actually supported but not included in default package.

remitamine:
i don't mean remove them all, i mean if someone want them he can get them from a separate place(for example in separate repo).
and may be let the issue open for discusion as feature request.
may be others have ideas about better way to do this.

dstftw:
Once again. An average user will **never ever bother** doing anything more complex than grabbing the binary or copy pasting the install command needless to say of some additional actions. Especially now when adult sites are incorporated in youtube-dl for a bunch of years already imagine what will happen if we stop shipping them by default - we will drown in bogus bugreports (same as gentoo *no-offensive-sites* package but on much larger scale). Probably it make some sense to make a non-default *lite-version* but it still does not resolve aforementioned problem since one may mistakenly grab wrong package.

remitamine:
> simple way to delete them
than what about ship a small script to remove them, if someone don't need them just execute it.

dstftw:
I don't know if this will work for py2exe binary.

remitamine:
if it is possible to add it also as a compiling flag? so it is possible to make version with and without this type of sites.
but i can't test it, i'm working in linux.

remitamine:
i created this script to remove the uneeded extractors(it's not complete i need to change also the generic extractor):
```
from youtube_dl.extractor import *
import re
import os

def list_non_suitable_extractors(age_limit):
    """
    Return a list of extractors that are suitable for the given age,
    sorted by extractor ID.
    """

    return sorted(
        filter(lambda ie: not ie.is_suitable(age_limit), gen_extractors()),
        key=lambda ie: ie.IE_NAME.lower())

non_suitable_extractors = list_non_suitable_extractors(0)

extractors_dir = 'youtube_dl/extractor/'

with open(extractors_dir + '__init__.py') as f:
    content = f.read()
    for extractor in non_suitable_extractors:
[...]

PR #10691, opened 2016-09-18
Add CLI options to filter usable extractors

danielkza:

- [ ] Bug fix
- [ ] New extractor
- [x] New feature

---

Add `--enable-extractors` and `--disable-extractors` options, which make
it possible to restrict the set of extractors to be considered when
downloading. This is useful to handle URLs that match multiple
extractors (although this should be rare), or only using particular
modes of some extractors (for example, only live videos for Twitch,
enabling only `twitch:stream`).

Both options can be specified multiple times, and each argument is
interpreted as a comma-separated list of fnmatch patterns, to allow the
use of wildcards. Comparisons to extractor names are case-insensitive.
The order of the arguments is not relevant - matching always proceeds as
follows:

- Initialize the set of considered extractors to all available
- If --enable-extractors is specified, remove all extractors that
*don't* match those patterns from consideration
- If --disable-extractors is specified, remove all extractors that *do*
match those patterns from consideration
- If --age-limit is specified, remove all extractors that are not
suitable from consideration

Therefore, disables and the age limit take precedende over enables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,14 @@ which means you can modify it, redistribute it or use it however you like.
                                      extractors
     --force-generic-extractor        Force extraction to use the generic
                                      extractor
+    --enable-extractors EXTRACTORS   Enable only the chosen extractors. Comma-
+                                     separated list of patterns, wildcards
+                                     allowed. Example:
+                                     "twitch:*,youtube:*,vimeo"
+    --disable-extractors EXTRACTORS  Disable the chosen extractors. Comma-
+                                     separated list of patterns, wildcards
+                                     allowed. Example:
+                                     "twitch:*,youtube:*,vimeo"
     --default-search PREFIX          Use this prefix for unqualified URLs. For
                                      example "gvsearch2:" downloads two videos
                                      from google videos for youtube-dl "large
--- a/test/test_execution.py
+++ b/test/test_execution.py
@@ -11,6 +11,7 @@ import subprocess
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
 from youtube_dl.utils import encodeArgument
+from youtube_dl.extractor import gen_extractors, get_info_extractor
 
 rootDir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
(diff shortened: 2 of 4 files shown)


dstftw:
This won't work in general since many extractors cross delegate extraction to another extractors.

For example, I only want to be able to download from AlJazeera thus enable it:
```
> py -2.7 .\youtube_dl\__main__.py http://www.aljazeera.com/programmes/the-slum/2014/08/deliverance-201482883754237240.html --enable-extractors aljazeera
[AlJazeera] deliverance-201482883754237240: Downloading webpage
[brightcove:new] 3792260579001: Downloading webpage
[brightcove:new] 3792260579001: Downloading JSON metadata
[brightcove:new] 3792260579001: Downloading m3u8 information
[download] Destination: The Slum - Episode 1 - Deliverance-3792260579001.mp4
[download]   0.1% of 1.71GiB at  1.53MiB/s ETA 19:07
ERROR: Interrupted by user
```
It does work even though it delegates to brightcove.

Then I want to download from Discovery and enable it:
```
> py -2.7 .\youtube_dl\__main__.py http://www.discovery.com/tv-shows/mythbusters/videos/mission-impossible-outtakes/ -F --enable-extractors discovery
[Discovery] mission-impossible-outtakes: Downloading JSON metadata
[download] Downloading playlist: Mission Impossible Outtakes
[Discovery] playlist Mission Impossible Outtakes: Collected 1 video ids (downloading 1 of them)
ERROR: no suitable InfoExtractor for URL http://players.brightcove.net/103207/default_default/index.html?videoId=ref:585f8ee5a885fcaa59a868f970721423d68dbe2f
```
But it does not work despite it uses the same brightcove as embed but via playlist indirection.

In this case I'm forced to enable all the extractors that Discovery may use in similar way. This is misleading and requires me as end user to know the implementation details of the particular extractor in some cases but does not in others.

danielkza:
That will never be fixable without some explicit metadata for dependencies,
and adding it would be a much larger undertaking. As-is this already
fulfills relevant use cases for many extractors without such "hidden"
requirements.

Em dom, 18 de set de 2016 08:28, Sergey M. <notifications@github.com>
escreveu:

dstftw:
>That will never be fixable without some explicit metadata for dependencies

That's why disabling or enabling extractors does not make sense.

danielkza:
Of course it does. My use case is only wanting to download live Twitch
streams. There are others, such as restricting the set of desirable
extractors for redirects, avoiding slow extractors that try to fetch large
amounts of metadata (of which Twitch is another example), or disabling
extractors that are broken for reasons external to yt-dl (such as a
corporate proxy).

Em dom, 18 de set de 2016 08:46, Sergey M. <notifications@github.com>
escreveu:

yan12125:
Not getting it. It's possible to feed youtube-dl the URL of a Twitch live stream and it starts downloading. What's the problem?

Anyway features known to be broken shouldn't be landed.

dstftw:
It does not until it works in all cases in uniform way. At least I'm not going to accept semi-broken options.
In case of twitch all that should be done is removing fallback to profile extraction in stream extractor that AFAIR has been added because previously profile videos has been displayed on the same page as stream (now they are not).

danielkza:
> Not getting it. It's possible to feed youtube-dl the URL of a Twitch live stream and it starts downloading. What's the problem?

If the stream isn't live the Twitch extractor will download VODs, which I don't want. And it expends a long time parsing dozens of videos I don't care about to even do that.

> Anyway features known to be broken shouldn't be landed.

I made an explicit effort to easily allow multiple extractors to be chosen and wildcards to be used to make it possible to manually handle cases such as the dependency one.

danielkza:
> It does not until it works in all cases in uniform way.

Do you plan on adding dependency metadata to every extractor? 

> At least I'm not going to accept semi-broken options.

The option is not broken. It does exactly what it promises,

> In case of twitch all that should be done is removing fallback to profile extraction in stream extractor that AFAIR has been added because previously profile videos has been displayed on the same page as stream (now they are not).

That doesn't solve any of the other use cases I mentioned.

You cannot project your issues with the wildness of the different extractors into this particular feature.

yan12125:
> Do you plan on adding dependency metadata to every extractor?

Dependency list is not a solution. For example, ```TumblrIE``` delegates ```GenericIE``` if it can't find a Tumblr video. In this case lots of extractors are added. I guess it's not what ```--enable-extractors``` tries to achieve.

danielkza:
@yan12125 That seems like a perfect example of the usefulness of this feature. What if I want the tumblr extractor but *not* the generic fallback that includes dozens of other video platforms?

yan12125:
> I want the tumblr extractor but _not_ the generic fallback that includes dozens of other video platforms?

Is there a _real_ need to achieve this?

yan12125:
Just looked at twitch.py. I guess you don't want this fallback?
```Python
        # Fallback on profile extraction if stream is offline
        if not stream:
            return self.url_result(
                'http://www.twitch.tv/%s/profile' % channel_id,
                'TwitchProfile', channel_id)
```

danielkza:
@yan12125 Is there a real reason why I should need to use *all* extractors or none whatsoever? Why is only wanting to download from some particular sources that problematic?

> Just looked at twitch.py. I guess you don't want this fallback?

Yes, but being able to choose extractors can be useful for other different cases. Specially when there are hundreds of extractors mixing with each other without much care.

danielkza:
@dstftw Also, why be so quick to close instead of providing me feedback and giving me the chance to improve/fix any issues?

dstftw:
Because there is no easy way to fix this.

yan12125:
About the problem of Twitch: this fallback was added in 240b9b7a5cd5f19f50ae28fb3ef991fede216ef6. My idea is removing the fallback. Personally I expect youtube-dl tell me there's currently no live stream (and better if youtube-dl tells me when the live stream will start) rather than downloading previous videos of the publisher. If a user want all previous recordings, they should use the profile page.

dstftw:
I've already explained why this fallback was added previously and it's already removed.

danielkza:
@dstftw  Bullshit. Your original example only fails because one case delegates directly to an extractor and the other to an URL. There is no reason why the filters could not be kept in the YoutubeDL object and used when dealing with dependent extractors. I had to go and look that up myself, since you chose to reject the feature from the start instead of giving me information so that I could improve the PR.

dstftw:
Then go and fix it. Note that we are not going to maintain additional explicit dependencies per extractor just in order to add an option that won't be used by 99% of users.

yan12125:
@dstftw Sorry I didn't notice there's a new commit.

@danielkza If you're going to continue this work, here are a few suggestions:
* ```--enable-extractors``` should never include any other extractors implicitly. In the example of AlJazeera, the download should fail. 
* Check related issues/PRs: #9487, #6501, #10174, #6497 (maybe more)
* Add tests. From what I can see this will be a fragile feature.

PR #10693, opened 2016-09-18
[soundcloud] Added soundcloud license detection for file name formatt…

coolsa:

- [ ] Bug fix
- [ ] New extractor
- [x] New feature

---

Added the ability for a file formatted with -o and downloaded from soundcloud to use the %(license)s tag, and when used prefixes it with the footers license.

Available soundcloud licenses are

all-rights-reserved
cc
cc-by
etc...

### DIFF
--- a/youtube_dl/extractor/soundcloud.py
+++ b/youtube_dl/extractor/soundcloud.py
@@ -138,13 +138,13 @@ class SoundcloudIE(InfoExtractor):
     def _resolv_url(cls, url):
         return 'http://api.soundcloud.com/resolve.json?url=' + url + '&client_id=' + cls._CLIENT_ID
 
-    def _extract_info_dict(self, info, license, full_title=None, quiet=False, secret_token=None):
+    def _extract_info_dict(self, info, full_title=None, quiet=False, secret_token=None):
         track_id = compat_str(info['id'])
         name = full_title or track_id
         if quiet:
             self.report_extraction(name)
-
         thumbnail = info['artwork_url']
+        track_license = info['license']
         if thumbnail is not None:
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/soundcloud.py
```diff
@@ -222,6 +228,14 @@ def _real_extract(self, url):
 
         track_id = mobj.group('track_id')
         token = None
+
+        #Get the tracks license, could be improved.
+        track_webpage = self._download_webpage(url, track_id) 
+        track_license = self._html_search_regex(
+            r'</ul>\s*License: (.*)\s*</footer>',
+            track_webpage, 'license', default="all-rights-reserved")
```
dstftw: There is no such licence `all-rights-reserved` exist. Lack of information should be expressed as `None`.

youtube_dl/extractor/soundcloud.py
```diff
@@ -222,6 +228,14 @@ def _real_extract(self, url):
 
         track_id = mobj.group('track_id')
         token = None
+
+        #Get the tracks license, could be improved.
+        track_webpage = self._download_webpage(url, track_id)
```
dstftw: This should not break extraction if it fails.

youtube_dl/extractor/soundcloud.py
```diff
@@ -222,6 +228,14 @@ def _real_extract(self, url):
 
         track_id = mobj.group('track_id')
         token = None
+
+        #Get the tracks license, could be improved.
```
dstftw: This comment is pointless.

dstftw:
Cherry-picked in4bfd294e2f83301921494c02e497cccf1a26cfd5, thanks.

PR #10697, opened 2016-09-18
[Antenna] Add new extractor

vagnum08:

---
- [ ] Bug fix
- [x] New extractor
- [ ] New feature

---

Added an extractor for the webtv of  [Antenna](http://www.antenna.gr/webtv/).
The service is geo-restricted  to Greece.

### DIFF
--- a/youtube_dl/extractor/antenna.py
+++ b/youtube_dl/extractor/antenna.py
@@ -1,0 +1,67 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import ExtractorError
+
+
+class AntennaIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?antenna\.gr/webtv/watch\?cid=(?P<id>[\w_%]+)(&p=[0-9]+)?'
+    _TESTS = [
+        {
+            'url': 'http://www.antenna.gr/webtv/watch?cid=otn_f_jvi5_e_z_i%3d&p=1',
+            'info_dict': {
+                'id': 'otn_f_jvi5_e_z_i%3d',
+                'ext': 'mp4',
+                'title': 'Αραχτοί και λάιτ (επεισ.16)',
+                'thumbnail': 're:^https?://.*\.jpg$',
+                'description': '"Αραχτές και... λάιτ" καταστάσεις στον Ant1.',
+            },
+            'params': {
+                'skip_download': True,
+            },
+            'skip': 'Content not available outside Greece.'
+        },
+        {
+            'url': 'http://www.antenna.gr/webtv/watch?cid=%2fn_u4_x_n79i_d_i%3d',
+            'info_dict': {
+                'id': '%2fn_u4_x_n79i_d_i%3d',
+                'ext': 'mp4',
+                'title': 'Της Ελλάδος τα παιδιά (επεισ.38 - Ιφιγένεια εν Τατοϊω)',
+                'thumbnail': 're:^https?://.*\.jpg$',
+                'description': 'Τρείς σμηνίτες, η κυρία Μπoυμπoύ και o Σμήvαρχoς τoυ γραφείoυ Κάκαλoς, συνθέτουν μια '
+                               'από τις πιο αστείες συντροφιές της ελληνικής τηλεόρασης.',
+            },
+            'params': {
+                'skip_download': True,
+            },
+            'skip': 'Content not available outside Greece.'
+        }]
+
+    def _real_extract(self, url):
+        error = 'Content not available outside Greece.'
+        desc = r'<description><!\[CDATA\[(.+?)\]\]></description>'
+        thumb = r'<jwplayer:image>(.+?)</jwplayer:image>'
+        formats = []
+        link = 'http://www.antenna.gr/templates/data/jplayer?d=m&cid='
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        rsspage = self._download_webpage(link + video_id, video_id)
+        title = self._og_search_title(webpage) or self._search_regex(r'<title>(.+?)</title>', rsspage, 'title')[9:-3]
+        video_url = self._search_regex(r'file="(.+?)"', rsspage, 'url')
+        if video_url == 'http://extranet.antenna.gr/flvsteaming/GR.flv':
+            raise ExtractorError('%s returned error: %s' % (self.IE_NAME, error), expected=True)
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/antenna.py
```diff
@@ -0,0 +1,115 @@
...
+        return {
+            'id': video_id,
+            'formats': formats,
+            'title': title,
+            'thumbnail': thumbnail,
+            'description': desc.split(' | ')[-1],
+        }
+
+    def _real_extract(self, url):
+        extractors = [self._extract_info_json, self._extract_info_rss]
```
dstftw: Is there any point in this? Can you provide example URL that is only available via JSON and only via RSS?

youtube_dl/extractor/antenna.py
```diff
@@ -0,0 +1,115 @@
...
+        jplayer_url = 'http://www.antenna.gr/templates/data/jplayer?d=m&cid='
+        desc_re = r'<description><!\[CDATA\[(.+?)\]\]></description>'
+        rss = self._download_webpage(jplayer_url + video_id, video_id)
+        video_url = self._search_regex(r'file="(.+?)"', rss, 'url')
+        if video_url == 'http://extranet.antenna.gr/flvsteaming/GR.flv':
+            raise AntennaIE.MediaSelectionError('Content not available outside Greece.')
+        formats.extend(self._extract_akamai_formats(video_url, video_id))
+        for video_format in formats:
+            if video_format.get('format_note') == 'Quality selection URL':
+                formats.remove(video_format)
```
dstftw: This should not be touched.

youtube_dl/extractor/antenna.py
```diff
@@ -0,0 +1,115 @@
...
+        if len(formats) == 0:
+            raise AntennaIE.MediaSelectionError('No formats available')
+        title = self._og_search_title(webpage).split(' | ')[-1] or self._search_regex(r'<title>(.+?)</title>', rss,
+                                                                                      'title')[9:-3]
+        thumbnail = self._og_search_thumbnail(webpage)
+        desc = self._og_search_description(webpage).split(' | ')[-1] or self._search_regex(desc_re, rss, 'description')
```
dstftw: In both extraction functions you extract these fields from `webpage` (as fallbacks in the second case) but you do it differently (split on ` | ` here and as is in the second case). What's the rationale?
`_search_regex` is fatal by default and will break extraction if it's not found for optional fields.

youtube_dl/extractor/antenna.py
```diff
@@ -0,0 +1,115 @@
...
+        desc_re = r'<description><!\[CDATA\[(.+?)\]\]></description>'
+        rss = self._download_webpage(jplayer_url + video_id, video_id)
+        video_url = self._search_regex(r'file="(.+?)"', rss, 'url')
+        if video_url == 'http://extranet.antenna.gr/flvsteaming/GR.flv':
+            raise AntennaIE.MediaSelectionError('Content not available outside Greece.')
+        formats.extend(self._extract_akamai_formats(video_url, video_id))
+        for video_format in formats:
+            if video_format.get('format_note') == 'Quality selection URL':
+                formats.remove(video_format)
+        if len(formats) == 0:
```
dstftw: `if not formats` is enough.

youtube_dl/extractor/antenna.py
vagnum08: On the webpage they have two meta tags like this: 
```html
<meta property="og:description" content="ANT1 WEB TV / ANT1 NEWS 19:00 | ΔΕΛΤΙΑ ΕΙΔΗΣΕΩΝ | Μετά από αλλεπάλληλες αναβολές ξεκίνησε η δίκη για την τραγωδία της Marfin.">
<meta name="description" content="ANT1 WEB TV / ANT1 NEWS 19:00 | ΔΕΛΤΙΑ ΕΙΔΗΣΕΩΝ | Μετά από αλλεπάλληλες αναβολές ξεκίνησε η δίκη για την τραγωδία της Marfin.">
```
On the `_extract_info_json` function I use `og_search_description` and `html_search_meta` on the webpage, as the json doesn't contain a description, only thumb, url and title.

Both `og_search_description` and `html_search_meta` return the following for the example above:

> "ANT1 WEB TV / ANT1 NEWS 19:00 | ΔΕΛΤΙΑ ΕΙΔΗΣΕΩΝ | Μετά από αλλεπάλληλες αναβολές ξεκίνησε η δίκη για την τραγωδία της Marfin." 

I am using a single split on `|` in line 99 to remove "ANT1 WEB TV / ANT1 NEWS 19:00 | ΔΕΛΤΙΑ ΕΙΔΗΣΕΩΝ | " and keep only the description.

On the other hand in `_extract_info_rss` function the rss contains a description.
```xml
<rss version="2.0" xmlns:jwplayer="http://rss.jwpcdn.com/">
    <channel>
    <item>
    <title><![CDATA[ANT1 News 19-09-2016 στις 19:00 
]]></title>
     
    <description><![CDATA[Μετά από αλλεπάλληλες αναβολές ξεκίνησε η δίκη για την τραγωδία της Marfin.]]></description>
    <jwplayer:image>http://img.antenna.gr/images/b90d73af-4ca7-432e-9103-476b0c9ca434.jpg</jwplayer:image>
    <jwplayer:source file="http://antennatvvod-vh.akamaihd.net/i/vods2/,044ef991-d45e-41f4-9f29-9a9d824ac80b,.mp4.csmil/master.m3u8"  />
    </item>
    </channel>
</rss>
```
Because the description from the rss feed doesn't contain extra stuff, I am using the split only on the first one as it uses the webpage to get it. 
```python
desc = self._og_search_description(webpage).split(' | ')[-1] or self._search_regex(desc_re, rss, 'description')
```

That way both extractor functions will correctly return "Μετά από αλλεπάλληλες αναβολές ξεκίνησε η δίκη για την τραγωδία της Marfin." as the description.

As for _search_regex since is for an optional field, probably will be best to use `fatal=false`, right?

youtube_dl/extractor/antenna.py
vagnum08: Frankly, I have yet to find a URL that only uses only one method either JSON or RSS.
I added both methods, as it seems to be redundant in case they remove one of JSON or RSS in the future. 

The JSON based extractor is first as it is more straight forward and if that give results it doesn't run the second one (RSS).

Would it be better if the RSS extractor was removed and only using JSON?

vagnum08:
The website has changed completely since my last commit and the code, as is, doesn't work anymore.

I will make changes to make it work and commit again.

PR #10716, opened 2016-09-20
[YoutubeDL] Include `playlist_description` in post-processing

TRox1972:
`playlist_id` and `playlist_title` are already supported, and there doesn't seem to be any other use for `playlist_description`, even though it's often returned by extractors

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -846,6 +846,7 @@ class YoutubeDL(object):
                     'playlist': playlist,
                     'playlist_id': ie_result.get('id'),
                     'playlist_title': ie_result.get('title'),
+                    'playlist_description': ie_result.get('description'),
                     'playlist_index': i + playliststart,
                     'extractor': ie_result['extractor'],
                     'webpage_url': ie_result['webpage_url'],


dstftw:
What's the point in the first place?

TRox1972:
@dstftw It's returned by some extractors, but doesn't seem to be used anywhere.

yan12125:
I guess @dstftw asked for a real use case for this addition. Currently no post processor uses it either.

TRox1972:
@yan12125 I can't think of any real use case this could have for the post processors atm, but it's returned by some extractors, which is pointless if it doesn't have any use case other than that. I know that at least some users would be interested in getting the playlist description, so why not return it?

yan12125:
Well, end users can get playlist descriptions from ```-j```. Also, users have no direct access to post-processors, do they?

yan12125:
From https://github.com/rg3/youtube-dl/blob/master/README.md#is-anyone-going-to-need-the-feature:

> Only post features that you (or an incapacitated friend you can personally talk to) require. Do not post features because they seem like a good idea. If they are really useful, they will be requested by someone who requires them.

So closing.

TRox1972:
@yan12125 OK. Thanks for your time, anyway :)   But do you have an example URL where I can get the playlist description using `-j`?

yan12125:
It was my wrong that I didn't test it. Playlists should use ```-J```, for example:
```
youtube-dl test:LePlaylist -J --flat-playlist | json_reformat
```

PR #10718, opened 2016-09-21
[meipai] Add new extractor

pyx:

---
- [ ] Bug fix
- [x] New extractor
- [ ] New feature

---

Explanation of your _pull request_ in arbitrary form goes here. Please make sure the description explains the purpose and effect of your _pull request_ and is worded well enough to be understood. Provide as much context and examples as possible.

This is the extractor for http://www.meipai.com/ (Meipai 美拍). The site allows users to share videos shoot from their mobile apps and also live streaming.

### DIFF
--- a/youtube_dl/extractor/meipai.py
+++ b/youtube_dl/extractor/meipai.py
@@ -1,0 +1,114 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from ..utils import parse_iso8601, unescapeHTML
+from .common import InfoExtractor
+
+# This is a sloppy fallback, except there is currently no other more reliable
+# ways to get the title if the page does not contain OpenGraph data, which
+# might be the case for some of the videos, the HTML output of the server is
+# not consistent.
+RE_TITLE = r'<title[^>]*>(.+)</title>'
+
+# This pattern will only showed up when this video is a playback of previous
+# live streaming
+RE_M3U8_URL = r'file:\s*encodeURIComponent\(["\'](.+)["\']\)'
+
+# This is the word in Simplified Chinese meaning 'Live Streaming Playback', it
+# will showed up in the keyword meta tag if the video is a recorded playback,
+# we use it to avoid false positives in the future as the RE_M3U8_URL pattern
+# is not specific enough.
+KEYWORD_PLAYBACK = '直播回放'
+
+
+class MeipaiIE(InfoExtractor):
+    IE_DESC = '美拍'
+    _VALID_URL = r'https?://(?:www\.)?meipai.com/media/(?P<id>[0-9]+)'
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/meipai.py
```diff
@@ -0,0 +1,114 @@
...
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._og_search_title(webpage, default=None)
+        if title is None:
+            # fall back to text used in title
+            title = unescapeHTML(
```
dstftw: `_html_search_regex` already calls `unescapeHTML`.

youtube_dl/extractor/meipai.py
```diff
@@ -0,0 +1,114 @@
...
+                self._html_search_regex(RE_TITLE, webpage, 'title'))
+
+        release_date = self._og_search_property(
+            'video:release_date', webpage, 'release date', default=None)
+        if release_date:
```
dstftw: No need for that check it's already checked in `parse_iso8601`.

youtube_dl/extractor/meipai.py
```diff
@@ -0,0 +1,114 @@
...
+
+# This pattern will only showed up when this video is a playback of previous
+# live streaming
+RE_M3U8_URL = r'file:\s*encodeURIComponent\(["\'](.+)["\']\)'
+
+# This is the word in Simplified Chinese meaning 'Live Streaming Playback', it
+# will showed up in the keyword meta tag if the video is a recorded playback,
+# we use it to avoid false positives in the future as the RE_M3U8_URL pattern
+# is not specific enough.
+KEYWORD_PLAYBACK = '直播回放'
```
dstftw: There is no need in these constants since they are only used once. Some comments are also pointless since it's obvious from code.

youtube_dl/extractor/meipai.py
```diff
@@ -0,0 +1,114 @@
...
+            release_date = parse_iso8601(release_date)
+
+        tags = self._og_search_property(
+            'video:tag', webpage, 'tags', default='').split(',')
+
+        info = {
+            'id': video_id,
+            'title': title,
+            'thumbnail': self._og_search_thumbnail(webpage, default=None),
+            'description': self._og_search_description(webpage, default=None),
```
dstftw: With `default` you say that it's an expected scenario when field is missing. According to tests - it's not thus `fatal=False` should be used instead. Or provide test these fields.

pyx:
Okay. Done. Please review, thanks.

pyx:
Any update on this?

PR #10748, opened 2016-09-24
[mwave] added support for URL language flags

cheshyre:
- [x] Bug fix
- [ ] New extractor
- [ ] New feature

---

Fixes a bug that caused the mwave extractor to fail on URLs with language tags (as mentioned in [this issue](https://github.com/rg3/youtube-dl/issues/10735)).

### DIFF
--- a/youtube_dl/extractor/mwave.py
+++ b/youtube_dl/extractor/mwave.py
@@ -9,7 +9,7 @@ from ..utils import (
 
 
 class MwaveIE(InfoExtractor):
-    _VALID_URL = r'https?://mwave\.interest\.me/mnettv/videodetail\.m\?searchVideoDetailVO\.clip_id=(?P<id>[0-9]+)'
+    _VALID_URL = r'https?://mwave\.interest\.me/(([a-z][a-z]\/)?)mnettv/videodetail\.m\?searchVideoDetailVO\.clip_id=(?P<id>[0-9]+)'
     _URL_TEMPLATE = 'http://mwave.interest.me/mnettv/videodetail.m?searchVideoDetailVO.clip_id=%s'
     _TEST = {
         'url': 'http://mwave.interest.me/mnettv/videodetail.m?searchVideoDetailVO.clip_id=168859',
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/mwave.py
```diff
@@ -28,6 +28,14 @@ class MwaveIE(InfoExtractor):
     def _real_extract(self, url):
         video_id = self._match_id(url)
 
+        # Check that language (if given) is valid
+        languages = ['en/', 'jp/', 'cn/', 'tw/', 'kr/', 'es/']
+        start = url.index('mwave.interest.me/') + len('mwave.interest.me/')
+        end = url.index('mnettv/', start)
+        if url[start:end] and not url[start:end] in languages:
+            raise Exception('[mwave] invalid language detected: {}'.format(url[start:end]))
```
dstftw: This should be detected in `_VALID_URL` if reasonable. But there is also no sense checking it all since new languages may appear in future. There is also a code duplicate. `{}` won't work in python 2.6. For extractor errors `ExtractorError` should be used.

youtube_dl/extractor/mwave.py
```diff
@@ -9,7 +9,7 @@
 
 
 class MwaveIE(InfoExtractor):
-    _VALID_URL = r'https?://mwave\.interest\.me/mnettv/videodetail\.m\?searchVideoDetailVO\.clip_id=(?P<id>[0-9]+)'
+    _VALID_URL = r'https?://mwave\.interest\.me/(([a-z][a-z]\/)?)mnettv/videodetail\.m\?searchVideoDetailVO\.clip_id=(?P<id>[0-9]+)'
```
dstftw: Outer parentheses are superfluous. Also don't capture groups you don't use.

youtube_dl/extractor/mwave.py
cheshyre: so by checking it in _VALID_URL do you mean changing the regex to something like `(((en)|(kr)|(jp))\/)?` ? or something like capturing the group and comparing it to some languages list in-line (I'm not sure if this is possible)?

dstftw: You don't need to check languages. Read the whole post.

PR #10755, opened 2016-09-25
[vk] Use inline video page to extract info.

kasper93:
- [x] I am the original original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

By requesting "show_inline" we can extract information for more videos which would be not available in normal video page. This is common when video is available only on user timeline.

Also some mirror changes in tests so that they all pass. (can be split into two commits if needed)

### DIFF
--- a/youtube_dl/extractor/toggle.py
+++ b/youtube_dl/extractor/toggle.py
@@ -200,7 +200,7 @@ class ToggleIE(InfoExtractor):
 
 class MeWatchIE(InfoExtractor):
     IE_NAME = 'mewatch'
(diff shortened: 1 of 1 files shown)


dstftw:
Do not mix unrelated changes in single PR.

dstftw:
Another side effect is that this approach reports wrong error messages for some videos.
Dropping `upload_date` is not acceptable.
The correct solutions would be to request show API when login data is provided and to request both otherwise.

kasper93:
Ok, so I fixed date and view count extraction and make use of inline page only if "need login" error is encountered. This works fine for all my test cases. What do you think?

kasper93:
bump

PR #10777, opened 2016-09-27
Improve regression detection with continous testing

### LINKED ISSUE #8496: Make continuous testing great again
It seems the youtube-dl project has abandoned the idea of using tests with travis-ci to keep all IEs always working (last passed build was 2 years ago, [#2571](https://travis-ci.org/rg3/youtube-dl/builds/99421929)). Not that it's a bad idea, since it's hard to keep up with all those sites. But a side effect of that is that it's harder to detect test fails that are due to a new commit, and not a website change.

What's I'd like to propose, is to keep the current testing infrastructure as a dashboard for working/non-working IEs, and add another "testsuite" that would test each commit for regressions: if a test is failing, the previous commit would be tested as well, and if it fails, it's not considered a regression.

I've made a simple proof of concept https://gist.github.com/anisse/6093f8b5814ab3ce7140 in bash. This could be improved/rewritten, in order to be integrated with travis-ci.

What do you guys think ?

anisse:

---
- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

travis: Add new regression detection script as default test suite

```
It runs tests and parses nosetests output to detect failures and test
them for regressions against a reference version. If it finds a
regression, it is automatically bisected.

Unstable or flaky tests are detected and ignored automatically by
running them multiple times in a row.

We keep the original test suite around, but mark it as allowed to fail.
It serves as a dashboard of current test statuses, but since the test
can fail out of our control (this is the essence of this project), we
don't want it to be blocking.

Using the regression detection as a fail means that any failing build
need to be examined. Even if the next build is "fixed", it does not mean
that the regression has been fixed. This is a change in semantics when
analyzing build history.

We map/reduce by splitting the test suite in 7 parts by abusing travis'
matrix feature. Reduce is done by hand, by analyzing the dashboard of
travis, any failing test being critical.

Because nosetests --processes just doesn't work with youtube-dl test
suite (yet), we route around it by first enumerating tests. This is a
bit long because nosetests needs to find and run all tests files in
order to enumerate them, but it should be at most 30 seconds, while the
test suite can take more than 2 hours on travis' infrastructure.

By doing this, we ensure we'll be able to run the tests faster, since
they are mostly I/O (network) bound due to the nature of the project.

Closes #8496
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,20 @@ python:
   - "3.4"
   - "3.5"
 sudo: false
-script: nosetests test --verbose
+env:
+  - TESTS=complete
+  - TESTS=regression_1-of-7
+  - TESTS=regression_2-of-7
+  - TESTS=regression_3-of-7
+  - TESTS=regression_4-of-7
+  - TESTS=regression_5-of-7
+  - TESTS=regression_6-of-7
+  - TESTS=regression_7-of-7
+matrix:
+  allow_failures:
+    - env: TESTS=complete
+
+script: devscripts/travis.sh
 notifications:
   email:
     - filippo.valsorda@gmail.com
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ notifications:
   email:
     - filippo.valsorda@gmail.com
     - yasoob.khld@gmail.com
+    - anisse@astier.eu
 #  irc:
 #    channels:
 #      - "irc.freenode.org#youtube-dl"
--- a/devscripts/regdetect.py
+++ b/devscripts/regdetect.py
@@ -1,0 +1,238 @@
+#!/usr/bin/env python3
+
+from __future__ import print_function, unicode_literals
+
+import subprocess
+import sys
+import os
+import time
+
+NOSECOMMAND="nosetests"
+
+
+def process(test):
+    # Parse nose output to get test statuses
+    # Format: test_<test_name> (<test_path>) ... [multi-line-error-message?] {ok,ERROR,FAIL}
+    if not test or len(test) == 0:
+        print("Received empty line")
+        return (None, None)
+    s = test.split()
+    if len(s) < 4:
+        print("Bad line passed, not enough elements:", test)
+    # We try to convert this nose output:
+    #   test_opengraph (test.test_InfoExtractor.TestInfoExtractor) ... ok
+    # into this test handle:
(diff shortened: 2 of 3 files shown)


yan12125:
This is a good idea. My concern is that it takes too much time. For example, https://travis-ci.org/rg3/youtube-dl/builds/163077852 takes 14 hrs 28 sec in total (if not paralleled), and other tests takes only 5~6 hours. It's likely that regression tests stall. #7267 should be landed first.

anisse:
> This is a good idea. My concern is that it takes too much time. For
example, https://travis-ci.org/rg3/youtube-dl/builds/163077852 takes 14 hrs
28 sec in total (if not paralleled), and other tests takes only 5~6 hours.
It's likely that regression tests stall. #7267 should be landed first.

The time it takes is actually a function of how many failing tests there
are, since those are run at least twice to see if they are regressions.

Also, once a regression is detected there is 60 seconds cooldown to prevent
temporary network issues or rate limiting to interfere. We could reduce
that sleep time.

It might also be a good idea to add timestamps to see what 's taking so
long.

I don't think we should have this depend on #7267.

yan12125:
If I understand it correctly, each test is run at least 2 times? The time to run the whole test suite is not smaller than 10~12 hours. Let it be 11 hours, and Travis CI builders are perfectly parallelized. That is, each push takes 11/5 hours in average. If the number of commits and pull requests goes beyond 24 / (11 / 5), Travis CI never finishes its jobs. Having more then 10.9 pushes a day is common in youtube-dl, so this PR would break tests.

anisse:
No, it only runs failed tests twice, to see if they are regressions.

The test suite might get behind, but it would catch up eventually. I don't think it will become a problem. If it does, simply fixing failing tests would help reduce the running time.

yan12125:
I may be unclear. I was referring to the original tests and the new regression tests. Each single test case is run in two test suites.

> it would catch up eventually

If it takes several days to catch up, it's a problem.

> simply fixing failing tests would help reduce the running time.

Fixing falling tests is not easy. Otherwise I won't leave more than 400 errors and failures there :(

anisse:
You are correct, I forgot about the original test suite. Tests are indeed ran at least twice. I wonder if we could ask travis-ci.org for more workers ?

yan12125:
I guess Travis CI won't allow it :(

PR #10796, opened 2016-09-29
[Morningstar] Modernize

TRox1972:

---
- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Simplifies metadata extraction for Morningstar.

### DIFF
--- a/youtube_dl/extractor/morningstar.py
+++ b/youtube_dl/extractor/morningstar.py
@@ -1,9 +1,11 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
-import re
-
 from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    get_element_by_id,
+)
 
 
 class MorningstarIE(InfoExtractor):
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/morningstar.py
```diff
@@ -16,32 +14,27 @@ class MorningstarIE(InfoExtractor):
...
 
     def _real_extract(self, url):
-        mobj = re.match(self._VALID_URL, url)
-        video_id = mobj.group('id')
+        video_id = self._match_id(url)
 
         webpage = self._download_webpage(url, video_id)
-        title = self._html_search_regex(
-            r'<h1 id="titleLink">(.*?)</h1>', webpage, 'title')
+        title = get_element_by_id('titleLink', webpage)
```
yan12125: ```title``` is mandatory while ```get_element_by_id``` fails silently if nothing is found.

youtube_dl/extractor/morningstar.py
```diff
@@ -1,13 +1,11 @@
...
 from __future__ import unicode_literals
 
-import re
-
 from .common import InfoExtractor
+from ..utils import get_element_by_id
 
 
 class MorningstarIE(InfoExtractor):
-    IE_DESC = 'morningstar.com'
```
yan12125: Why remove it?

youtube_dl/extractor/morningstar.py
TRox1972: Should we change `get_element_by_id` to handle `fatal=True` cases? Or should I just raise an `ExtractorError` whenever title extraction fails?

youtube_dl/extractor/morningstar.py
TRox1972: Other extractors don't usually specify `IE_DESC` to include `.com`, and it isn't part of Morningstar's name.

yan12125: Some extractors use the domain name as ```IE_DESC```, so I guess it's OK.

TRox1972: OK. Reversed.

youtube_dl/extractor/morningstar.py
TRox1972: fixed

youtube_dl/extractor/morningstar.py
```diff
@@ -16,32 +18,30 @@ class MorningstarIE(InfoExtractor):
...
     def _real_extract(self, url):
-        mobj = re.match(self._VALID_URL, url)
-        video_id = mobj.group('id')
-
+        video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
-        title = self._html_search_regex(
-            r'<h1 id="titleLink">(.*?)</h1>', webpage, 'title')
+        
+        title = get_element_by_id('titleLink', webpage) or ''
```
yan12125: Is ```or ''``` necessary?

TRox1972: removed

youtube_dl/extractor/morningstar.py
```diff
@@ -16,32 +18,30 @@ class MorningstarIE(InfoExtractor):
...
-        video_id = mobj.group('id')
-
+        video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
-        title = self._html_search_regex(
-            r'<h1 id="titleLink">(.*?)</h1>', webpage, 'title')
+        
+        title = get_element_by_id('titleLink', webpage)
+        if not title:
+            raise ExtractorError('Unable to extract title for %s.' % video_id)
```
dstftw: This should be done in generic way similar to `_search_regex`.

PR #10830, opened 2016-10-01
[kotnrtube] Improve metadata extraction

TRox1972:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

### DIFF
--- a/youtube_dl/extractor/kontrtube.py
+++ b/youtube_dl/extractor/kontrtube.py
@@ -1,4 +1,4 @@
-# encoding: utf-8
+# coding: utf-8
 from __future__ import unicode_literals
 
 import re
@@ -6,12 +6,12 @@ import re
 from .common import InfoExtractor
 from ..utils import (
     int_or_none,
+    js_to_json,
     parse_duration,
 )
 
 
 class KontrTubeIE(InfoExtractor):
-    IE_NAME = 'kontrtube'
     IE_DESC = 'KontrTube.ru - Труба зовёт'
     _VALID_URL = r'https?://(?:www\.)?kontrtube\.ru/videos/(?P<id>\d+)/(?P<display_id>[^/]+)/'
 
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/kontrtube.py
```diff
@@ -31,16 +31,16 @@ class KontrTubeIE(InfoExtractor):
...
-        display_id = mobj.group('display_id')
+        video_id, display_id = mobj.groups()
 
-        webpage = self._download_webpage(
-            url, display_id, 'Downloading page')
+        webpage = self._download_webpage(url, display_id)
+
+        video_data = self._parse_json(self._search_regex(
+            r'var\s+flashvars\s*=\s*({[^}]+});', webpage, 'video data'),
+        video_id, transform_source=js_to_json)
```
dstftw: I've already pointed out: replacing one time-proven working data extraction approach with another is not an improvement and I will reject all such further PRs. All additional scenarios should be added as fallbacks.
Again, this code is not PEP 8 compliant.

TRox1972: My apologies. I understand. Changed to fallbacks.

youtube_dl/extractor/kontrtube.py
```diff
@@ -31,16 +31,25 @@ class KontrTubeIE(InfoExtractor):
...
+        video_id, display_id = mobj.groups()
 
-        webpage = self._download_webpage(
-            url, display_id, 'Downloading page')
+        webpage = self._download_webpage(url, display_id)
+
+        video_data = self._parse_json(
+            self._search_regex(
+                r'var\s+flashvars\s*=\s*({[^}]+});', webpage, 'video data', fatal=False),
+            video_id, transform_source=js_to_json)
```
dstftw: You seem misunderstand the point completely. This piece of code will break extraction if failed and it can fail easily due to inner dict in JSON (since you don't allow `}`), due to parts `js_to_json` can't fix or due to some other reason that will eventually produce unparsable JSON.

TRox1972: OK. Closing

Issue #10780, opened 2016-09-27
[generic] The Guardian.com downloads javascript file (not video)


RingoTheDog:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like that [x])
- Use *Preview* tab to see how your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2016.09.27*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.09.27**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add `-v` flag to **your command line** you run youtube-dl with, copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):
```
>youtube-dl.py -v "https://www.theguardian.com/football/2016/sep/26/leicester-defence-claudio-ranieri-porto-champions-league"
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['-v', 'https://www.theguardian.com/football/2016/sep/26/leicester-defence-claudio-ranieri-porto-champions-league']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2016.09.27
[debug] Python version 3.5.1 - Windows-7-6.1.7601-SP1
[...]

PR #10854, opened 2016-10-06
[Iqm2] new extractor for Accela/IQM2

johnhawkinson:
---
- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

New extractor for .iqm2.com videos.

IQM2 aka Accela is a municipal meeting management platform that
(among other things) stores livestreamed video from municipal
meetings.  After a hefty (several-hour) processing time, that video
is available in easily downloadable form from their web portal, but
prior to that, the video can only be watched in realtime through
JWPlayer. This extractor is designed to download the realtime video
prior to download links being available.


youtube_dl/extractor/iqm2.py
```diff
@@ -0,0 +1,109 @@
...
+    @classmethod
+    def suitable(cls, url):
+        """Receives a URL and returns True if suitable for this IE."""
+
+        # This does not use has/getattr intentionally - we want to know whether
+        # we have cached the regexp for *this* class, whereas getattr would also
+        # match the superclass
+        if '_VALID_URL_RE' not in cls.__dict__:
+            cls._VALID_URL_RE = re.compile(cls._VALID_URL, flags=re.IGNORECASE)
+        return cls._VALID_URL_RE.match(url) is not None
```
dstftw: Use `(?i)` in regex itself if you want case insensitivity.

youtube_dl/extractor/iqm2.py
```diff
@@ -0,0 +1,109 @@
...
+        webpage = self._download_webpage(url, video_id)
+
+        # print "Original URL is", url
+
+        # We want to extract an inner URL like this:
+        # <div id="VideoPanel" class="LeftTopContent">
+        #   <div id="VideoPanelInner" onselectstart="javascript:return false;"
+        #     style="overflow:hidden"
+        #     src="/Citizens/VideoScreen.aspx?MediaID=1563&Frame=SplitView"
+        #     quality="hd">
```
dstftw: This comment is pointless. It's already clear from the regex itself what you are trying to extract and from where.

youtube_dl/extractor/iqm2.py
```diff
@@ -0,0 +1,109 @@
...
+# is really the only time this extractor is especially important).
+
+# This is also a relatively braindead extractor. It parses a given page like
+#   http://cambridgema.iqm2.com/Citizens/SplitView.aspx?Mode=Video&MeetingID=1679
+# to determine the location of an inner div defined by a URL of the form
+#   http://cambridgema.iqm2.com/Citizens/VideoScreen.aspx?MediaID=1563&Frame=SplitView
+
+# and then simply hands that URL to the GenericIE generic extractor,
+# which matches it under the "Broaden the findall a little bit:
+# JWPlayer JS loader" (line 2372 as of 6 Oct 2016).
```
dstftw: That's not a place for essays and repeating the code itself.

youtube_dl/extractor/iqm2.py
```diff
@@ -0,0 +1,109 @@
...
+        #     quality="hd">
+        inner_url_rel = self._html_search_regex(
+            r'<div id="VideoPanelInner".*src="([^"]+)"',
+            webpage, 'url');
+        # print "inner_URL is", inner_url_rel
+
+        inner_url = compat_urlparse.urljoin(url, inner_url_rel)
+        # print "Joined URL is", inner_url
+
+        return GenericIE(self._downloader)._real_extract(inner_url)
```
dstftw: Use `self.url_result(inner_url, 'Generic')` instead.

youtube_dl/extractor/iqm2.py
johnhawkinson: Thanks.

What is the right place for documentation of extractors? I realize I personally tend to have a verbose comment style, though looking at the youtube-dl codebase there are almost no comments at all. In a lot of places it's tough to figure out the rationale for design choices, or even what an extractor is really for. My general assumption is "some long comments are better than no comments"; if that's wrong here, is there any better guidance than "no essays"?

I assume your comment about "repeating the code" is in re lines 26-29? No problems with removing that.
But I think some clarity about what part of the GenericIE extractor (e.g. it's the jwplayer js section) is helpful if someone is trying to trace out how this code is supposed to work in the future. No?

youtube_dl/extractor/iqm2.py
johnhawkinson: Good point! Though very few current extractors do that:
````
pb3:extractor jhawk$ fgrep -Rn '(?i)'  *.py|grep VALID
commonprotocols.py:11:    _VALID_URL = r'(?i)rtmp[est]?://.+'
commonprotocols.py:37:    _VALID_URL = r'(?i)mms://.+'
dailymotion.py:41:    _VALID_URL = r'(?i)(?:https?://)?(?:(www|touch)\.)?dailymotion\.[a-z]{2,3}/(?:(?:embed|swf|#)/)?video/(?P<id>[^/?_]+)'
```

would it not make sense to consider changing suitable() globally to make most regexps case insensitive, rather than trying to touch every regexp?
OTOH I guess it might break some things?

Either way, I guess it's outside the scope of this pull request.

youtube_dl/extractor/iqm2.py
johnhawkinson: Hmm.

It's pretty-well acknowledged that a lot of people have trouble reading regexps, and often write them in a buggy way. So my experience is that it's helpful to have a comment showing what you're trying to match to compare against the regexp. 

Also, I have to assume, going forward, that youtube-dl will switch from using regexps in cases like this to using some kind of DOM-based parser that does the XPath equivalent of `//div[@id="VideoPanelInner"]/src`, and that it'll be helpful when converting regexps to XPath to have a clear example of what's trying to be matched. No?

youtube_dl/extractor/iqm2.py
johnhawkinson: Doh! 

Thanks for all the comments and the timely review, dstftw.
I'll respin this and push further changes consistent with your comments this weekend.

youtube_dl/extractor/iqm2.py
dstftw: Well, basically that's the problem of these people not us. We don't care whether one can read regexp or not.
Moreover most likely next time this code is read by someone is when extractor breaks due to layout change. Chances are this snippet is already irrelevant by that time.

dstftw:
There is also an issue with format detection. Now it detects different qualities of the same video as playlist items instead of different formats of the single video:
```
[download] Downloading playlist: None
[IQM2] playlist None: Collected 2 video ids (downloading 2 of them)
[download] Downloading video 1 of 2
[info] Available formats for 1563_480:
0            mp4        unknown
[download] Downloading video 2 of 2
[info] Available formats for 1563_720:
0            mp4        unknown
[download] Finished downloading playlist: None
```
You may need to handle this on your own rather than delegating to generic extractor.

youtube_dl/extractor/iqm2.py
johnhawkinson: Probably this is another issue to open, but it's very easy to make most extractors fail by uppercasing something.
Compare:
```
pb3:Downloads jhawk$ youtube-dl -s  'http://abcnews.go.COM/ThisWeek/video/week-exclusive-irans-foreign-minister-zarif-20411932'
[generic] week-exclusive-irans-foreign-minister-zarif-20411932: Requesting header
[generic] week-exclusive-irans-foreign-minister-zarif-20411932: Downloading webpage
[generic] week-exclusive-irans-foreign-minister-zarif-20411932: Extracting information
pb3:Downloads jhawk$ youtube-dl -s  'http://abcnews.go.com/ThisWeek/video/week-exclusive-irans-foreign-minister-zarif-20411932'
[abcnews:video] Downloading Akamai AMP feed
[abcnews:video] 20411932: Downloading f4m manifest
[abcnews:video] 20411932: Downloading m3u8 information
pb3:Downloads jhawk$ 
```

Sticking to a "seen in the wild" standard masks a lot of bugs.
I'll open an issue and submit a pull to at least make the YourExtractor sample code case-insentitive.
But I do think the project probably can and should do better...

youtube_dl/extractor/iqm2.py
johnhawkinson: I think we care whether the maintainers of the code can read the regexp. What is the cost to having such a comment, it would seem to only benefit?

I think your argument cuts the other way as well. If the HTML has in fact changed, then someone looking to update the extractor may have a really tough time trying to figure out what the regexp was supposed to do, because they inspect the HTML and find it doesn't seem to be close to relevant. Having the example of the regexp makes it that much easier to see what kind of change was there, and to make the code more understandabe.

I'm certainly not going to die on this one, but I strongly believe more comments make things clearer, and that regexps are so bad for intelligibility that it's worth a little time to improve clarity.

Douglas Crockford has written a bit about this problem; in his book _Javascript: The Good Parts_, in Chapter 7, he wrote:

> Regular expressions tend to be extremely terse, even cryptic. They are easy to use in their simplest form, but they can quickly become bewildering. JavaScript's regular expressions are difficult to read in part because they do not allow comments or whitespace. All of the parts of a regular expression are pushed tightly together, making them almost indecipherable. This is a particular concern when they are used in security applications for scanning and validation. If you cannot read and understand a regular expression, how can you have confidence that it will work correctly for all inputs? Yet, despite their obvious drawbacks, regular expressions are widely used.

Now, of course, python's regexps are a bit better since they support (?x) and it looks like about 1/6 of the extractors do so. So at least there's the oportunity to write better regexps. But my takeaway is that one good way to mitigate the crypticness and errorprone nature of regexps is to try to use them as simply and clearly as possible, and that means showing a reader what they mean.

Because again, what's the cost? 

Obviously nobody's going to get confused by this particular regexp, which is about as boring as they come in youtube-dl _VALID_URL contexts. But I'm writing for the general principle...

youtube_dl/extractor/iqm2.py
johnhawkinson: I was a little reluctant to slow down the travis-ci runs with unnecessary tests.
I guess my listing URLs there was really more along the lines of my documentation of the purpose of the regexp, see our discussion about lines 95-100, below.
What's the standard for tests? If there are multiple forms of URLs that are handled but we anticipate them being fate-sharing (e.g. `/path/to/*.aspx` all works the same), is it appropriate to include multiple test cases?

johnhawkinson:
> Now it detects different qualities of the same video as playlist items instead of different formats of a single video

I did notice this after submission — it only occurs after the realtime video has gone away, so in practice it is a non-issue, as again the point of this extractor is to handle the time-limited realtime video at the same location.

That said, I think it's a consequence of self.url_result(), but I'll see.

youtube_dl/extractor/iqm2.py
dstftw: Such tests should be included as [only matching](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/abc.py#L52-L53).

youtube_dl/extractor/iqm2.py
dstftw: That's simulated. No such URLs is seen in the wild so far and no one will ever intentionally upper case some part of it.

youtube_dl/extractor/iqm2.py
dstftw: The cost is that one should read the same information twice. When there is a need to read a regexp, for me (and I suppose all other developers) it's easier to read just the regexp rather than read the comment and still read the regexp. Especially when the comment is much longer than the regexp itself. Sample snippet could be useful in case of very complex regexp. This one is trivial.

johnhawkinson:
Sergey:
> > Now it detects different qualities of the same video as playlist items instead of different formats of a single video

> In practice it is a non-issue, as again the point of this extractor is to handle the time-limited realtime video at the same location. …  I think it's a consequence of not using self.url_result(), but I'll see.

Unfortunately it's not.  Once the realtime video is processed, the HTML looks like this (pretty-printed):
```
<script type='text/javascript'> SetupJWPlayer(eval('
[{"file":"http://MediaHTTP.IQM2.com/CambridgeMA/1563_480.mp4",
  "label":"SD 480",
  "default":false,
  "tracks":[
      {"file":"/Services/TranscriptGet.aspx?MediaID=1563&format=vtt",
       "label":"English",
       "kind":"captions",
       "default":true}]
 },
 {"file":"http://MediaHTTP.IQM2.com/CambridgeMA/1563_720.mp4",
  "label":"HD 720",
  "default":true,
  "tracks":[
      {"file":"/Services/TranscriptGet.aspx?MediaID=1563&format=vtt",
       "label":"English",
       "kind":"captions",
       "default":true}]
}]')); </script>
```
and GenericIE is matching on the `file:` portions with ` r'[^A-Za-z0-9]?(?:file|video_url)["\']?:\s*["\'](http(?![^\'"]+\.[0-9]+[\'"])[^\'"]+)["\']'`

Pulling code from the generic extractor into my own to handle this case is a *lot* of code.
It's like 50 lines: duplicating stuff like check_video(), filter_video(); extracting video_uploader, video_id and video_title; building dictionaries, etc. Duplicating that quantity of code in my own extractor seems wrong (I did test this and have it stashed if it's worth bringing back.)

Other approaches are to modify the generic extractor, perhaps:

1. Presume that anything matching this RE is in a similar format and if indeed the multiple files are in a single JS array and they have consistent filenames (1563_*) then assume they're format specifiers and go from there. If you try to parse out the JS Array, it makes the parsing much more fragile and complicates the regexp...substantially.

1. As above, but don't worry about ensuring they're in a single JS array. Just check the filenames for consistency, or maybe obvious format keys like _480 and _720.

1. Pass in a flag (smuggle_url() anyone?) to GenericIE and if the flag is set, assume multiple files are formats rather than playlist entries.

1. Abstract out a lot of the generic functions in GenericIE into common.py, like check_video() and filter_video(); that only gets you halfway there…and then I guess move the metadata extraction operations into additional function that can move to common.py?

1. Something else?

Another complexity is I don't really know which of the _TESTS in generic.py cover this regexp and SetupJWPlayer situation. I suppose I could just wget all the test URLs and grep, so that's annoying but not insurmountable?

My inclination is just to punt on this and not worry about the minor bug that the extractor results in a playlist instead of format choices, again because if that is the case, there's no point in using youtube-dl here because the website offers direct plain old boring a href links to mp4 files. But I'm happy to do otherwise.

Guidance?

johnhawkinson:
Anyhow, aside from the playlist/format issues, I think all the stuff from the review is addressed now.

youtube_dl/extractor/iqm2.py
johnhawkinson: Discussion forked to issue #10882 .

youtube_dl/extractor/iqm2.py
```diff
@@ -0,0 +1,85 @@
...
+        # print "inner_URL is", inner_url_rel
+
+        inner_url = compat_urlparse.urljoin(url, inner_url_rel)
+
+        if self._downloader.params.get('verbose'):
+            self.to_screen('Invoking downloader on %s' % inner_url)
+
+        # Generic extractor matches this under the "Broaden the
+        # findall a little bit: JWPlayer JS loader" (line 2372 as of 6
+        # Oct 2016, dcdb292fddc82ae11f4c0b647815a45c88a6b6d5).
```
yan12125: I guess ```JWPlatform._extract_jwplayer_data``` can be used here.

youtube_dl/extractor/iqm2.py
```diff
@@ -0,0 +1,85 @@
...
+        # and feed the canonicalized src element to the generic extractor
+        inner_url_rel = self._html_search_regex(
+            r'<div id="VideoPanelInner".*src="([^"]+)"',
+            webpage, 'url');
+        # print "inner_URL is", inner_url_rel
+
+        inner_url = compat_urlparse.urljoin(url, inner_url_rel)
+
+        if self._downloader.params.get('verbose'):
+            self.to_screen('Invoking downloader on %s' % inner_url)
```
yan12125: It should be in ```YoutubeDL.process_ie_result``` instead, and it's better to open a new pull request for this change.

youtube_dl/extractor/iqm2.py
```diff
@@ -0,0 +1,85 @@
...
+        # Simple extractor: take, e.g.
+        #   http://cambridgema.iqm2.com/Citizens/SplitView.aspx?Mode=Video&MeetingID=1679
+        # and look for
+        #   <div id="VideoPanel" class="LeftTopContent">
+        #     <div id="VideoPanelInner" ... src="/Citizens/VideoScreen.aspx?MediaID=1563&Frame=SplitView">
+        # and feed the canonicalized src element to the generic extractor
+        inner_url_rel = self._html_search_regex(
+            r'<div id="VideoPanelInner".*src="([^"]+)"',
+            webpage, 'url');
+        # print "inner_URL is", inner_url_rel
```
yan12125: Comments for debugging should be removed, too.

youtube_dl/extractor/iqm2.py
johnhawkinson: Well, err…not easily. For one thing, it just doesn't work here:
```
>>> g1=re.findall(r'[^A-Za-z0-9]?(?:file|video_url)["\']?:\s*["\'](http(?![^\'"]+\.[0-9]+[\'"])[^\'"]+)["\']', w)
>>> g1
['http://MediaHTTP.IQM2.com/CambridgeMA/1563_480.mp4', 'http://MediaHTTP.IQM2.com/CambridgeMA/1563_720.mp4']
>>> g2 =re.findall(r'jwplayer\((?P<quote>[\'"])[^\'" ]+(?P=quote)\)\.setup\s*\((?P<options>[^)]+)\)', w)
>>> g2
[]
```
Where g1 is the regexp from generic.py:2372 and g2 is the regexp from JWPlatform._find_jwplayer_data(), a helper function for _extract_jwplayer_data(); and w is the webpage.

And even if it did work, why would it be better to assume this would stay JWPlayer data in the future, rather than letting GenericIE handle it via whatever mechanism is appropriate at the current time? (Maybe this is a philosophical question on extractors that I don't really understand; I definitely lack experience with youtube-dl.)

And then I'd need to subclass `JWPlatformBaseIE`, not sure if that has consequences?

On the other hand, [JWPlatformBaseIE has](https://github.com/rg3/youtube-dl/blob/176006a1202cc6ef3d0a768ace41f97516c76c6d/youtube_dl/extractor/jwplatform.py#L17-L20):
```
class JWPlatformBaseIE(InfoExtractor):
    @staticmethod
    def _find_jwplayer_data(webpage):
# TODO: Merge this with JWPlayer-related codes in generic.py
```
from you in June, so maybe you can advise of the future of the JW code in generic.py vs. jwplayer.py?

But this really gets to the issues in [this comment](https://github.com/rg3/youtube-dl/pull/10854#issuecomment-252465329), replying to @dstftw's suggestion of not using the Generic extractor to avoid confusion between whether multiple videos are formats and playlist entries.

Can you tell me what you think in about the issues in that comment?

youtube_dl/extractor/iqm2.py
yan12125: > Is there a good way to do that that I'm missing?

Currently no. You can add it in another pull request.

Don't be surprising that youtube-dl lacks such a basic utility. It's an old project from 2006.

youtube_dl/extractor/iqm2.py
yan12125: > it just doesn't work here

IQM2 does not use standard JWPlayer layout. In this case you need to extract the data manually and pass it to _parse_jwplayer_data

> And then I'd need to subclass ```JWPlatformBaseIE```, not sure if that has consequences?

That's fine.

About @dstftw's comment: again IQM2 does not follow JWPlayer's rule - they should use 'sources' instead of an array. Unfortunately I guess the only solution is handling it specially. Maybe modify the returned value from ```_parse_jwplayer_data```.

I propose to use jwplayer-specific helper functions in all new extractors because:

* GenericIE is not so reliable. It can extract something unexpected, for example #10780
* It can extract subtitles, thumbnails and more

johnhawkinson: OK, how's this? I had to make JWPlatformBaseIE a little more general, see 00253f8. I hope that's OK without a seperate pull request :)

youtube_dl/extractor/iqm2.py
johnhawkinson: Or not. I can't seem to find the exact reason that I wanted this.
I think [L1640](https://github.com/rg3/youtube-dl/blob/176006a1202cc6ef3d0a768ace41f97516c76c6d/youtube_dl/YoutubeDL.py#L1640) seems to display it under normal circumstances.

Maybe I wanted it to display when running test_download.py? I'm not sure, but that doesn't seem a good enough reason to modify `process_ie_result`.

Dropping this unless someone disagrees.

youtube_dl/extractor/iqm2.py
johnhawkinson: Opened Issue #10894 for this.

youtube_dl/extractor/jwplatform.py
```diff
@@ -50,7 +50,11 @@ def _parse_jwplayer_data(self, jwplayer_data, video_id=None, require_title=True,
             # JWPlayer backward compatibility: flattened sources
             # https://github.com/jwplayer/jwplayer/blob/v7.4.3/src/js/playlist/item.js#L29-L35
             if 'sources' not in video_data:
-                video_data['sources'] = [video_data]
+                if isinstance(video_data, list):
```
yan12125: This is incorrect and may break other websites. According to [1], an array _is_ a playlist. Codes for handling incorrect JWPlayer usage in IQM2 should be in ```iqm2.py```.

[1] https://support.jwplayer.com/customer/portal/articles/1481963-building-managing-playlists#fndtn-code

johnhawkinson: As it stands, its not going to break other websites. Without this patch, when an array is passed in, a type error exception gets thrown, I believe from L57 where it makes an array of an array.

That said, how would you suggest fixing it in `iqm2.py` without substantial code duplication?

johnhawkinson: Specifically, here's what we get without that hunk of the patch; it's not possible that this change breaks other websites since they would already generate this TypeError:
```
pb3:youtube-dl jhawk$ python test/test_download.py TestDownload.test_IQM2_1
[IQM2] 1679: Downloading webpage
[IQM2] 1563: Downloading webpage
E
ERROR: test_IQM2_1 (__main__.TestDownload)
  File "test/test_download.py", line 141, in test_template
    force_generic_extractor=params.get('force_generic_extractor', False))
  File "/Users/jhawk/src/youtube-dl/youtube_dl/YoutubeDL.py", line 694, in extract_info
    ie_result = ie.extract(url)
  File "/Users/jhawk/src/youtube-dl/youtube_dl/extractor/common.py", line 355, in extract
  File "/Users/jhawk/src/youtube-dl/youtube_dl/extractor/iqm2.py", line 88, in _real_extract
    webpage, video_id, require_title=False)
  File "/Users/jhawk/src/youtube-dl/youtube_dl/extractor/jwplatform.py", line 33, in _extract_jwplayer_data
    jwplayer_data, video_id, *args, **kwargs)
  File "/Users/jhawk/src/youtube-dl/youtube_dl/extractor/jwplatform.py", line 57, in _parse_jwplayer_data
    video_data['sources'] = [video_data]
TypeError: list indices must be integers, not unicode

Ran 1 test in 2.128s

FAILED (errors=1)
```

And here's a little context via pdb:
```
pb3:youtube-dl jhawk$ python  -c 'import pdb, youtube_dl; print youtube_dl; pdb.runcall(youtube_dl.main, ["-v","http://cambridgema.iqm2.com/Citizens/SplitView.aspx?Mode=Video&MeetingID=1679"])' youtube_dl 
<module 'youtube_dl' from 'youtube_dl/__init__.pyc'>
> /Users/jhawk/src/youtube-dl/youtube_dl/__init__.py(444)main()
-> try:
(Pdb) b youtube_dl/extractor/jwplatform.py:57
Breakpoint 1 at /Users/jhawk/src/youtube-dl/youtube_dl/extractor/jwplatform.py:57
(Pdb) c
...
[IQM2] 1563: Downloading webpage
> /Users/jhawk/src/youtube-dl/youtube_dl/extractor/jwplatform.py(57)_parse_jwplayer_data()
-> video_data['sources'] = [video_data]
(Pdb) p video_data
[{u'default': False, u'tracks': [{u'default': True, u'kind': u'captions', u'file': u'/Services/TranscriptGet.aspx?MediaID=1563&format=vtt', u'label': u'English'}], u'file': u'http://MediaHTTP.IQM2.com/CambridgeMA/1563_480.mp4', u'label': u'SD 480'}, {u'default': True, u'tracks': [{u'default': True, u'kind': u'captions', u'file': u'/Services/TranscriptGet.aspx?MediaID=1563&format=vtt', u'label': u'English'}], u'file': u'http://MediaHTTP.IQM2.com/CambridgeMA/1563_720.mp4', u'label': u'HD 720'}]
(Pdb) p video_data['sources']
*** TypeError: TypeError('list indices must be integers, not str',)
(Pdb) p jwplayer_data
{u'playlist': [[{u'default': False, u'tracks': [{u'default': True, u'kind': u'captions', u'file': u'/Services/TranscriptGet.aspx?MediaID=1563&format=vtt', u'label': u'English'}], u'file': u'http://MediaHTTP.IQM2.com/CambridgeMA/1563_480.mp4', u'label': u'SD 480'}, {u'default': True, u'tracks': [{u'default': True, u'kind': u'captions', u'file': u'/Services/TranscriptGet.aspx?MediaID=1563&format=vtt', u'label': u'English'}], u'file': u'http://MediaHTTP.IQM2.com/CambridgeMA/1563_720.mp4', u'label': u'HD 720'}]]}
(Pdb) pp jwplayer_data
{u'playlist': [[{u'default': False,
                 u'file': u'http://MediaHTTP.IQM2.com/CambridgeMA/1563_480.mp4',
                 u'label': u'SD 480',
                 u'tracks': [{u'default': True,
                              u'file': u'/Services/TranscriptGet.aspx?MediaID=1563&format=vtt',
                              u'kind': u'captions',
                              u'label': u'English'}]},
                {u'default': True,
                 u'file': u'http://MediaHTTP.IQM2.com/CambridgeMA/1563_720.mp4',
                 u'label': u'HD 720',
                 u'tracks': [{u'default': True,
                              u'file': u'/Services/TranscriptGet.aspx?MediaID=1563&format=vtt',
                              u'kind': u'captions',
                              u'label': u'English'}]}]]}
(Pdb) 
```

johnhawkinson: > That said, how would you suggest fixing it in iqm2.py without substantial code duplication?

Oops, for some reason I was thinking I'd need to override `_parse_jwplayer_data()` which is fairly long, but I shoehorned it into an override of `_extract_jwplayer_data()` instead. So done.

It still seems to me that this change would have benefited `jwplayform.py` and moving it into IQM2 is not advantageous, but up to you.

johnhawkinson:
Is there an outstanding issue here I should be correcting?

johnhawkinson:
Hi, @yan12125, is OK for me to "bump" this and ask if it needs anything else? I haven't heard anything in two weeks? Thanks. (Sorry, not sure on protocol.)

yan12125:
Sorry for the delay. In http://somervillecityma.iqm2.com/Scripts/jwplayer/player.js, ```SetupJWPlayer()``` is defined as:
```
function SetupJWPlayer(playSources) {
    jwplayer.key = "VO1kc/+YcrrjDA7x43ztygqLfHfiZ3Dc3Dd4VQblU0vtcglj";
    playerStartPosition = parseInt(GetQueryParameters(window.location.href)["MediaPosition"]);


    jwMediaPlayer = jwplayer("MediaPlayer1").setup({
        playlist: playSources,
        autostart: true,
        width: "100%",
        height: "100%",
        androidhls: true
    });
```
As only ```playlist``` is necessary, all the mess can be dropped:
```Python
        jwplayer_options = self._parse_json(self._search_regex(
            r'SetupJWPlayer\(eval\(\'(?P<options>.+)\'\)\);', webpage,
            'JWPlayer options', group='options'),
            video_id, transform_source=js_to_json)
        info_dict = self._parse_jwplayer_data(
            {'playlist': jwplayer_options}, video_id, require_title=False)
```

youtube_dl/extractor/iqm2.py
```diff
@@ -0,0 +1,113 @@
...
+        inner_url_rel = self._html_search_regex(
+            r'<div id="VideoPanelInner".*src="([^"]+)"',
+            webpage, 'url');
+
+        inner_url = compat_urlparse.urljoin(url, inner_url_rel)
+        mobj = re.match(
+            r'(?i)https?://(?:\w+\.)?iqm2\.com/Citizens/\w+.aspx\?.*MediaID=(?P<id>[0-9]+)',
+            inner_url)
+        video_id = mobj.group('id')
+        webpage = self._download_webpage(inner_url, video_id)
```
yan12125: Don't use the same variable name for different pages

youtube_dl/extractor/iqm2.py
```diff
@@ -0,0 +1,113 @@
...
+# (several-hour) processing time, that video is available in easily
+# downloadable form from their web portal, but prior to that, the
+# video can only be watched in realtime through JWPlayer. Other
+# (somervillecityma.iqm2.com) instances don't seem to ever offer a
+# downloadable form. This extractor is designed to download the
+# realtime video without the download links being available. For more
+# info on Accela, see:
+#   http://www.iqm2.com/About/Accela.aspx
+#   http://www.accela.com/
+#   https://github.com/Accela-Inc/leg-man-api-docs
```
yan12125: This code is for extracting IQM2 videos, not for IQM2 itself. Introduction texts to IQM2 are unnecessary.

johnhawkinson: Done. Sorry, this _really_ goes against the grain for me. I've been trained to write with extneisve comments, and I cannot at all understand why you don't want good comments in youtube-dl. Every time I read another extractor, I wish there was more context explaining what it was and why the choices were made and how it applies and where to look for more information. So it's very difficult for me to omit those things.  Sorry if it feels like I'm not honoring your coding standards, I'm really trying but it is hard for me.

youtube_dl/extractor/iqm2.py
johnhawkinson: fixed

johnhawkinson:
>  yan12125 commented on Nov 16, 2016
> Sorry for the delay. In http://somervillecityma.iqm2.com/Scripts/jwplayer/player.js, SetupJWPlayer() is > defined as:

Sorry, I'm confused by that comment. I was confused in November and set it aside and I'm afraid I lost track of it.

I don't see why the `SetupJWPlayer()` function in that one IQM2 site is important. The `SetupJWPlayer() ` is defined differently in different sites, e.g. http://cambridgema.iqm2.com/Scripts/jwplayer/player.js calls  `jwplayer("MediaPlayer1").setup` with slightly different parameters.

Are you saying I should stop redefining `_find_jwplayer_data` and `_extract_jwplayer_data` and just let them inherit from JWPlatformBaseIE? It doesn't seem like that's going to work, but maybe I am misunderstanding you? You're saying to your code:
```
 jwplayer_options = self._parse_json(self._search_regex(
            r'SetupJWPlayer\(eval\(\'(?P<options>.+)\'\)\);', webpage,
            'JWPlayer options', group='options'),
            video_id, transform_source=js_to_json)
        info_dict = self._parse_jwplayer_data(
            {'playlist': jwplayer_options}, video_id, require_title=False)
```
and then to also remove...what exactly?

Sorry, can you be a little more clear?

PR #10879, opened 2016-10-08
[FFmpegExtractAudioPP] info dict return audio filepath and ext instead of the video filepath if file already exists

mdawar:
If the converted audio file already exists, the `FFmpegExtractAudioPP.run` method returns the original video filepath instead of the converted audio filepath and ext which is the expected behavior, so if we set the filepath and the ext before checking if the file already exists it will return the correct values:

```python
# set the converted file path and ext before checking if it already exists
information['filepath'] = new_path
information['ext'] = extension

# now this will return the correct values
if (new_path == path or
        (self._nopostoverwrites and os.path.exists(encodeFilename(new_path)))):
    self._downloader.to_screen('[ffmpeg] Post-process file %s exists, skipping' % new_path)
    return [], information
```

### DIFF
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -278,6 +278,9 @@ class FFmpegExtractAudioPP(FFmpegPostProcessor):
 
         prefix, sep, ext = path.rpartition('.')  # not os.path.splitext, since the latter does not work on unicode in all setups
         new_path = prefix + sep + extension
+        
+        information['filepath'] = new_path
+        information['ext'] = extension
 
         # If we download foo.mp3 and convert it to... foo.mp3, then don't delete foo.mp3, silly.
         if (new_path == path or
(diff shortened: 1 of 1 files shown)


youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -278,6 +278,9 @@ def run(self, information):
 
         prefix, sep, ext = path.rpartition('.')  # not os.path.splitext, since the latter does not work on unicode in all setups
         new_path = prefix + sep + extension
+        
+        information['filepath'] = new_path
+        information['ext'] = extension
```
dstftw: This will return non existent path if conversion fails.

mdawar: But isn't `PostProcessingError` supposed to be raised if the converaion fails ? so it will never return on failure.

dstftw: Well, indeed, currently it will.

PR #10926, opened 2016-10-15
[nytimes] Support downloading podcasts

johnhawkinson:
- [x] Improvement

---

Support New York Times podcasts.
Since this parses an array of JS from inside the page, it's potentially fragile, so add a fallback with simpler parsing (no metadata) as well. I don't have a test case where it actually fails, so I don't have a test for the fallback code.

### DIFF
--- a/youtube_dl/extractor/nytimes.py
+++ b/youtube_dl/extractor/nytimes.py
@@ -125,11 +125,11 @@ class NYTimesBaseIE(InfoExtractor):
             'id': video_id,
             'title': title,
             'creator': audio_data['track'].get('credit'),
-            'series': audio_data['podcast']['title'],
+            'series': podcast_title,
             'episode': episode_title,
             'episode_number': episode_number,
             'url': audio_data['track']['source'],
-            'duration': audio_data['track']['duration'],
+            'duration': audio_data['track'].get('duration'),
             'description': description,
         }
         
--- a/youtube_dl/extractor/nytimes.py
+++ b/youtube_dl/extractor/nytimes.py
@@ -13,6 +13,7 @@ from ..utils import (
     parse_iso8601,
     mimetype2ext,
     determine_ext,
+    UnsupportedError,
 )
 
 
--- a/youtube_dl/extractor/nytimes.py
+++ b/youtube_dl/extractor/nytimes.py
@@ -105,7 +105,7 @@ class NYTimesBaseIE(InfoExtractor):
         if not description:
             description = self._html_search_meta(['og:description', 'twitter:description'], webpage)
 
-        episode_title = audio_data['track']['title'].strip("‘’") # strip curlyquotes
+        episode_title = audio_data['track']['title']
         episode_number = None
         episode = audio_data['podcast']['episode'].split()
         if episode:
@@ -182,7 +182,7 @@ class NYTimesArticleIE(NYTimesBaseIE):
         'md5': 'e0d52040cafb07662acf3c9132db3575',
         'info_dict': {
             'id': '20',
-            'title': "The Run-Up: He Was Like an Octopus",
+            'title': "The Run-Up: \u2018He Was Like an Octopus\u2019",
             'ext': 'mp3',
             'description': 'We go behind the story of the two women who told us that Donald Trump touched them inappropriately (which he denies) and check in on Hillary Clinton’s campaign.',
         }
--- a/youtube_dl/extractor/nytimes.py
+++ b/youtube_dl/extractor/nytimes.py
@@ -102,13 +102,13 @@ class NYTimesBaseIE(InfoExtractor):
         audio_data = self._parse_json(json, page_id, transform_source=js_to_json)['data']
         
         description = audio_data['track']['description']
-        if not len(description):
+        if not description:
             description = self._html_search_meta(['og:description', 'twitter:description'], webpage)
 
         episode_title = audio_data['track']['title'].strip("‘’") # strip curlyquotes
         episode_number = None
         episode = audio_data['podcast']['episode'].split()
-        if len(episode):
+        if episode:
             episode_number = int_or_none(episode[-1])
             video_id = episode[-1]
         else:
--- a/youtube_dl/extractor/nytimes.py
+++ b/youtube_dl/extractor/nytimes.py
@@ -125,47 +125,9 @@ class NYTimesIE(NYTimesBaseIE):
 
         return self._extract_video_from_id(video_id)
 
-class NYTimesPodcastIE(InfoExtractor):
-    _VALID_URL = r'(?i)https?://(?:www\.)?nytimes\.com/.*/podcasts/(?P<id>[^.]+)(?:\.html)?'
-    _TESTS = [{
-        'url': 'http://www.nytimes.com/2016/10/14/podcasts/revelations-from-the-final-weeks.html',
-        'md5': 'e0d52040cafb07662acf3c9132db3575',
-        'info_dict': {
-            'id': '20',
-            'title': "The Run-Up: He Was Like an Octopus",
-            'ext': 'mp3',
-            'description': 'We go behind the story of the two women who told us that Donald Trump touched them inappropriately (which he denies) and check in on Hillary Clinton’s campaign.',
-        }
-        } ]
-    
-
-    def _real_extract(self, url):
-        page_id = self._match_id(url)
-
-        webpage = self._download_webpage(url, page_id)
-
-        data_json = self._html_search_regex(r'NYTD.FlexTypes.push\(({[^)]*)\)', webpage, 'json data');
-        audio_data = self._parse_json(data_json, page_id, transform_source=js_to_json)['data']
-
-        episode_title = audio_data['track']['title'].strip(u"‘’") # strip curlyquotes
-        episode_number = audio_data['podcast']['episode'].split()[1]
-        
-        info_dict = {
-            'id': episode_number,
-            'title': "%s: %s" % (audio_data['podcast']['title'], episode_title),
-            'series': audio_data['podcast']['title'],
-            'episode': episode_title,
-            'episode_number': episode_number,
-            'url': audio_data['track']['source'],
-            'duration': audio_data['track']['duration'],
-            'description': self._html_search_meta(['og:description', 'twitter:description'], webpage)
-            }
-
-        return info_dict
-
         
 class NYTimesArticleIE(NYTimesBaseIE):
-    _VALID_URL = r'https?://(?:www\.)?nytimes\.com/(.(?<!video)(?<!podcast))*?/(?:[^/]+/)*(?P<id>[^.]+)(?:\.html)?'
+    _VALID_URL = r'https?://(?:www\.)?nytimes\.com/(.(?<!video))*?/(?:[^/]+/)*(?P<id>[^.]+)(?:\.html)?'
     _TESTS = [{
         'url': 'http://www.nytimes.com/2015/04/14/business/owner-of-gravity-payments-a-credit-card-processor-is-setting-a-new-minimum-wage-70000-a-year.html?_r=0',
         'md5': 'e2076d58b4da18e6a001d53fd56db3c9',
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/nytimes.py
```diff
@@ -96,6 +98,43 @@ def get_file_size(file_size):
...
         }
 
+    def _extract_podcast_from_json(self, json, page_id, webpage):
+        audio_data = self._parse_json(json, page_id, transform_source=js_to_json)['data']
+        
+        description = audio_data['track']['description']
+        if not len(description):
+            description = self._html_search_meta(['og:description', 'twitter:description'], webpage)
+
+        episode_title = audio_data['track']['title'].strip(u"‘’") # strip curlyquotes
```
yan12125: Python 3.2 doesn't like u-literals.

yan12125:
> Since this parses an array of JS from inside the page, it's potentially fragile, so add a fallback with simpler parsing (no metadata) as well. I don't have a test case where it actually fails, so I don't have a test for the fallback code.

Parsing JS is indeed fragile, while there's no need to keep un-tested codes.

johnhawkinson:
All set, thanks. I'm embarassed to have gotten used to ignoring the travis-ci output, which would have caught the u"" literal problem in Python 3.2.
I'm not really sure we're better off without a fallback, but hey, it works today, so your call.

youtube_dl/extractor/nytimes.py
```diff
@@ -96,6 +98,43 @@ def get_file_size(file_size):
             'thumbnails': thumbnails,
         }
 
+    def _extract_podcast_from_json(self, json, page_id, webpage):
+        audio_data = self._parse_json(json, page_id, transform_source=js_to_json)['data']
+        
+        description = audio_data['track']['description']
+        if not len(description):
```
dstftw: `if not description` is enough, same for other occurrences.

youtube_dl/extractor/nytimes.py
```diff
@@ -96,6 +98,43 @@ def get_file_size(file_size):
...
+        
+        info_dict = {
+            'id': video_id,
+            'title': title,
+            'creator': audio_data['track'].get('credit'),
+            'series': audio_data['podcast']['title'],
+            'episode': episode_title,
+            'episode_number': episode_number,
+            'url': audio_data['track']['source'],
+            'duration': audio_data['track']['duration'],
```
dstftw: Optional fields should not break extraction if missing.

youtube_dl/extractor/nytimes.py
```diff
@@ -139,15 +178,39 @@ class NYTimesArticleIE(NYTimesBaseIE):
...
+
+        video_id = self._html_search_regex(r'data-videoid="(\d+)"', webpage, 'video id', None, False)
+        if video_id is not None:
+            return self._extract_video_from_id(video_id)
+        
+        data_json = self._html_search_regex(r'NYTD.FlexTypes.push\(({[^)]*)\)', webpage, 'json data', None, False);
+        if data_json is not None:
+            return self._extract_podcast_from_json(data_json, page_id, webpage)
+        else:
+            raise UnsupportedError(url)
```
dstftw: Do not pass `default=None` to `_html_search_regex` instead.

johnhawkinson:
Done @dstftw. Although I would disagree with 001a30f. Without that, the extractor goes through multiple cases and if all of them fail, it raises UnsupportedError(). But doing what you request
> Do not pass default=None to _html_search_regex instead.

means that we get an error specific to the last test to fail, which may not really reflect the problem.

For instance, without that change, the error is:

```
pb3:youtube-dl jhawk$ python -m youtube_dl -v 'http://www.nytimes.com/2016/10/15/world/americas/obama-cuba-trade-embargo.html?hp&action=click&pgtype=Homepage&clickSource=story-heading&module=second-column-region&region=top-news&WT.nav=top-news'
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://www.nytimes.com/2016/10/15/world/americas/obama-cuba-trade-embargo.html?hp&action=click&pgtype=Homepage&clickSource=story-heading&module=second-column-region&region=top-news&WT.nav=top-news']
[debug] youtube-dl version 2016.10.12
[debug] Git HEAD: 9fb7e1b
[debug] Python version 2.7.10 - Darwin-14.5.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg 3.1.4, ffprobe 3.1.4, rtmpdump 2.4
[debug] Proxy map: {}
[NYTimesArticle] obama-cuba-trade-embargo: Downloading webpage
ERROR: Unsupported URL: http://www.nytimes.com/2016/10/15/world/americas/obama-cuba-trade-embargo.html?hp&action=click&pgtype=Homepage&clickSource=story-heading&module=second-column-region&region=top-news&WT.nav=top-news
  File "youtube_dl/YoutubeDL.py", line 694, in extract_info
    ie_result = ie.extract(url)
  File "youtube_dl/extractor/common.py", line 356, in extract
  File "youtube_dl/extractor/nytimes.py", line 217, in _real_extract
    raise UnsupportedError(url)
UnsupportedError: Unsupported URL: http://www.nytimes.com/2016/10/15/world/americas/obama-cuba-trade-embargo.html?hp&action=click&pgtype=Homepage&clickSource=story-heading&module=second-column-region&region=top-news&WT.nav=top-news
```

Whereas now, we get an error about `json data` which is probably not applicable:

```
pb3:youtube-dl jhawk$ python -m youtube_dl -v 'http://www.nytimes.com/2016/10/15/world/americas/obama-cuba-trade-embargo.html?hp&action=click&pgtype=Homepage&clickSource=story-heading&module=second-column-region&region=top-news&WT.nav=top-news'
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://www.nytimes.com/2016/10/15/world/americas/obama-cuba-trade-embargo.html?hp&action=click&pgtype=Homepage&clickSource=story-heading&module=second-column-region&region=top-news&WT.nav=top-news']
[debug] youtube-dl version 2016.10.12
[debug] Git HEAD: 3915b51
[debug] Python version 2.7.10 - Darwin-14.5.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg 3.1.4, ffprobe 3.1.4, rtmpdump 2.4
[debug] Proxy map: {}
[NYTimesArticle] obama-cuba-trade-embargo: Downloading webpage
ERROR: Unable to extract json data; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
  File "youtube_dl/YoutubeDL.py", line 694, in extract_info
    ie_result = ie.extract(url)
  File "youtube_dl/extractor/common.py", line 356, in extract
  File "youtube_dl/extractor/nytimes.py", line 212, in _real_extract
    data_json = self._html_search_regex(r'NYTD.FlexTypes.push\(({[^)]*)\)', webpage, 'json data')
  File "youtube_dl/extractor/common.py", line 668, in _html_search_regex
    res = self._search_regex(pattern, string, name, default, fatal, flags, group)
  File "youtube_dl/extractor/common.py", line 659, in _search_regex
RegexNotFoundError: Unable to extract json data; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

It also means anyone adding another case to this extractor (two so far) will have to add a test, essentially reverting 001a30f.

dstftw:
Once there is an extractor matching the URL it's already assumed "supported" by design. `UnsupportedError` is only thrown from generic extractor.
If you throw `UnsupportedError` from extractor and website's layout changes later you will get invalid `Unsupported URL` rather than `unable to extract X`, so possibly broken extractor will be masked with completely different error message.
This pattern is used in all extractors and I see no point changing it.

>It also means anyone adding another case to this extractor (two so far) will have to add a test, essentially reverting 001a30f.

No, what for? One should just pass default to last case and add another fatal case next to it. Same would be done in case of `UnsupportedError`.

johnhawkinson:
> No, what for? One should just pass default to last case and add another fatal case next to it. Same would be done in case of `UnsupportedError`.

I just meant that in order to add another case after the json parsing case, an `if` would have to be added and the _html_search_regex() params would have to change to add a default and make it non-fatal.

Anyhow, doesn't matter much either way, and all I think all your requested changes are set!

youtube_dl/extractor/nytimes.py
```diff
@@ -96,6 +98,43 @@ def get_file_size(file_size):
             'thumbnails': thumbnails,
         }
 
+    def _extract_podcast_from_json(self, json, page_id, webpage):
+        audio_data = self._parse_json(json, page_id, transform_source=js_to_json)['data']
+        
+        description = audio_data['track']['description']
```
dstftw: Should not break if missing.

youtube_dl/extractor/nytimes.py
```diff
@@ -96,6 +98,43 @@ def get_file_size(file_size):
...
+        if not description:
+            description = self._html_search_meta(['og:description', 'twitter:description'], webpage)
+
+        episode_title = audio_data['track']['title'].strip("‘’") # strip curlyquotes
```
dstftw: >.strip("‘’") # strip curlyquotes

What's the point?

youtube_dl/extractor/nytimes.py
```diff
@@ -139,15 +178,36 @@ class NYTimesArticleIE(NYTimesBaseIE):
...
 
-        video_id = self._html_search_regex(r'data-videoid="(\d+)"', webpage, 'video id')
+        webpage = self._download_webpage(url, page_id)
 
-        return self._extract_video_from_id(video_id)
+        video_id = self._html_search_regex(r'data-videoid="(\d+)"', webpage, 'video id', None, False)
+        if video_id is not None:
+            return self._extract_video_from_id(video_id)
+        
+        data_json = self._html_search_regex(r'NYTD.FlexTypes.push\(({[^)]*)\)', webpage, 'json data')
```
dstftw: Unescaped dots. Also `)` may be contained in a string within JSON so matching `[^)]*` may fail.

youtube_dl/extractor/nytimes.py
johnhawkinson: In this case curlyquotes seem to delimit the field for no apparent reason, and having quotes of any kind in the filesystem is quite a hassle, so stripping them seems warranted:

http://www.nytimes.com/2016/10/14/podcasts/revelations-from-the-final-weeks.html

```
var NYTD=NYTD || {}; NYTD.FlexTypes = NYTD.FlexTypes || []; NYTD.FlexTypes.push({
    "target":"FT100000004709062",
    "type":"NYT5-Podcast-Audio",
    "data":{
	"podcast":{
	    "title":"The Run-Up",
	    "landing":"",
	    "episode":"Episode 20",
	    "episodeurl":"http:\/\/www.nytimes.com\/2016\/10\/11\/podcasts\/on-sex-and-power-in-the-campaign-from-the-women-who-cover-it.html",
	    "subscribe":[{
		"name":"iTunes",
		"url":"https:\/\/geo.itunes.apple.com\/us\/podcast\/the-run-up\/id1142083165?mt=2"},
			 {
			     "name":"Google Play Music",
			     "url":"https:\/\/play.google.com\/music\/listen?authuser&u=0#\/ps\/Iotf2cvoqfvgkt4ywtmpdb3gjlu"}]},
	"track":{
	    "title":"‘He Was Like an Octopus’",
	    "description":"",
	    "credit":"",
	    "source":"https:\/\/rss.art19.com\/episodes\/b0684887-7f4b-4bae-a7af-de2e84d8f4f7.mp3",
	    "duration":2130}
    }});
```

dstftw: This looks like intentional quote or so.

youtube_dl/extractor/nytimes.py
johnhawkinson: done

youtube_dl/extractor/nytimes.py
johnhawkinson: In retrospect I think you're correct, done. This was the first test case I had and I overgeneralized.

youtube_dl/extractor/nytimes.py
johnhawkinson: Hopefully this is more to your liking now?

It would have been helpful to have had an example on how this is best used, perhaps a comment in `common.py` where `parse_json` is defined, or somewhere else.
I started looking for examples and first up was `abc.py` which used `[^)]+`:
```
    r'inline(?P<type>Video|Audio|YouTube)Data\.push\((?P<json_data>[^)]+)\);'
```

`adultswim.py` uses `({.*});`
```
    r'var bootstrappedData = ({.*});'
```

It'd be nice to know what the best practice is.
There are a lot of places in the codebase where the same thing is done many different ways and it's hard to know which is preferred…

dstftw:
>It'd be nice to know what the best practice is.
There are a lot of places in the codebase where the same thing is done many different ways and it's hard to know which is preferred…

That depends on a particular case. There are lots of legacy extractors that are not so well written and don't follow best practices. Generally what is preferred is adaptive code with fallbacks.
Cherry-picked in 74324a7ac2633869d80263f1c12d7f48928c06fe, thanks.

johnhawkinson:
> That depends on a particular case. There are lots of legacy extractors that are not so well written and don't follow best practices. Generally what is preferred is adaptive code with fallbacks.

Sure. I guess I'm saying it would be helpful if the [developer] documentation highlighted a few extractors as good models. Or, in the case of functions in common.py, if their docstrings or comments pointed to examples of good usage. Especially in cases where the alphabetically first extractor (`abc.py`) is not a great model.

Tanks!

PR #10944, opened 2016-10-17
MIME type

darkmorpher:
Missing .exe forces unknown MIME type in browser when clicked to download.


Hrxn:
??

That's not a PE binary?

PR #10945, opened 2016-10-17
[utils] add a base exception class and initialize exceptions properly

mdawar:
It's always a good practice to have a base exception class and inherit from it, this way when handling exceptions you can differentiate the library errors from other exceptions, for example handling `Exception` will catch all the exceptions of youtube-dl and any other exceptions based on `Exception`, but now you can handle `YoutubeDLError` and `Exception` separately:

``` python
try:
    youtubedl = YoutubeDL(url, options)
    # ...
except YoutubeDLError:
    # handle all YoutubeDL errors
except Exception:
    # handle all the other exceptions

```

Also 2 exception classes didn't call the parent's `__init__` which is recommended and will set the `args` attribute of the exception which is used by the `__str__` and `__repr__` methods.

### DIFF
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -780,7 +780,7 @@ class ContentTooShortError(YoutubeDLError):
 
     def __init__(self, downloaded, expected):
         super(ContentTooShortError, self).__init__(
-            'Downloaded {} bytes, expected {} bytes'.format(downloaded, expected)
+            'Downloaded {0} bytes, expected {1} bytes'.format(downloaded, expected)
         )
         # Both in bytes
         self.downloaded = downloaded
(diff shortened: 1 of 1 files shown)


mdawar:
What about moving the exceptions to their own `exceptions` module ? Then we have to edit all the imports also it might break others code if they're importing them from `utils`, so what do you think ?

dstftw:
>it might break others code

You've answered yourself.

youtube_dl/utils.py
```diff
@@ -773,12 +779,15 @@ class ContentTooShortError(Exception):
     """
 
     def __init__(self, downloaded, expected):
+        super(ContentTooShortError, self).__init__(
+            'Downloaded {} bytes, expected {} bytes'.format(downloaded, expected)
```
dstftw: `{}` won't work in python 2.6.

mdawar:
@dstftw are there any other fixes required ?

mdawar:
@dstftw the string formatting was fixed for Python 2.6, any other fixes required ?

dstftw:
Included in #11970.

PR #11185, opened 2016-11-13
[mediasite] new extractor

### LINKED ISSUE #5428: Support for NYU videos
My university uses Mediasite player to host video lectures. The videos can be manually downloaded by pulling a link from a javascript file in the page.

The system is set up exactly as with the existing Sandia extractor, with the exception that one must login before accessing the videos, and some minor case stuff (lowercase v uppercase of certain strings).

I modified the Sandia extractor to match the NYU videos, and tried various methods of authentication, including: -u/-pw; adding in the AUTH string from the cookie to the extractor file, and; importing the cookie with --cookies. 

Interestingly enough, the error I get is with regexp and not authentication. 

Here's the output I get:
```
[debug] System config: []
[debug] User config: []
[debug] Command-line args: ['-u', 'PRIVATE', '-p', 'PRIVATE', '--verbose', 'http
://nyulaw.mediasite.com/mediasite/Play/197180ea8f5c448e9ccf3290c5d855411d']
[debug] Encodings: locale cp1252, fs mbcs, out cp437, pref cp1252
[debug] youtube-dl version 2015.04.09
[debug] Python version 3.4.3 - Windows-8-6.2.9200
[debug] exe versions: none
[debug] Proxy map: {}
[Sandia] 197180ea8f5c448e9ccf3290c5d855411d: Downloading webpage
ERROR: Unable to extract JS code URL; please report this issue on https://yt-dl.
org/bug . Make sure you are using the latest version; see  https://yt-dl.org/upd
ate  on how to update. Be sure to call youtube-dl with the --verbose flag and in
clude its complete output.
  File "D:\Knowledge\NYU Lectures\youtube-dl-master\youtube_dl\YoutubeDL.py", li
ne 651, in extract_info
    ie_result = ie.extract(url)
  File "D:\Knowledge\NYU Lectures\youtube-dl-master\youtube_dl\extractor\common.
py", line 275, in extract
  File "D:\Knowledge\NYU Lectures\youtube-dl-master\youtube_dl\extractor\sandia.
py", line 33, in _real_extract
    r'<script type="text/javascript" src="(/mediasite/FileServer/Presentation/[^
"]+)"', webpage, 'JS code URL')
  File "D:\Knowledge\NYU Lectures\youtube-dl-master\youtube_dl\extractor\common.
py", line 549, in _search_regex
youtube_dl.utils.RegexNotFoundError: Unable to extract JS code URL; please repor
t this issue on https://yt-dl.org/bug . Make sure you are using the latest versi
on; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl
with the --verbose flag and include its complete output.'
```

fstirlitz:
A new extractor for sites based on the Sonic Foundry Mediasite player. This extractor subsumes Sandia and CollegeRama extractors. This PR may at least partially resolve #5428.

Some SmoothStreaming manifests referenced by Mediasite players don't specify the FourCC of the audio codec, making `_parse_ism_formats` skip them when enumerating streams in the manifest. Microsoft's documentation says this is not a violation; FourCC is indeed optional in `<QualityLevel>` elements corresponding to audio streams. This PR sets `AACL` as the default FourCC in SmoothStreaming audio streams. I'm not sure if this is valid, but it seems to work... smoothly.

Downloading slides is still not supported; but I think it should be possible to create a synthetic Matroska/MP4 stream from the individual slide images and timing information in the JSON data, by feeding them to FFmpeg's `concat` demuxer (see `ffmpeg-formats(1)`). I did some preliminary testing and it appears to be feasible. When the `-vcodec copy` option is used, FFmpeg will construct losslessly encoded files (MJPEG, that is), which appear to be playable in GStreamer-based and FFmpeg-based applications (although it requires putting a dummy zero-duration frame at the end, so that the stream has the right duration; also, seeking doesn't seem to work very well). A custom downloader will have to be written, should you choose to pursue this approach.


remitamine:
> Microsoft's documentation says this is not a violation; FourCC is indeed optional in <QualityLevel> elements corresponding to audio streams.

from TrackAttributes in 2.2.2.5 of revision 7:
> If the track is contained in a stream whose Type is audio, the following additional fields are also required and MUST be present in TrackAttributes: MaxWidthAttribute, MaxHeightAttribute, CodecPrivateDataAttribute, SamplingRateAttribute, ChannelsAttribute, BitsPerSampleAttribute, PacketSizeAttribute, AudioTagAttribute, and FourCCAttribute.

> Some SmoothStreaming manifests referenced by Mediasite players don't specify the FourCC of the audio codec, making _parse_ism_formats skip them when enumerating streams in the manifest.

in this case i think it would be more safe to check for AudioTag value to be equal to 255, instead of defaulting to AACL.

fstirlitz:
The [MSDN page](https://msdn.microsoft.com/en-us/library/ff728116%28v=vs.90%29.aspx) says otherwise. Perhaps an older version of SSTR did things differently.

> in this case i think it would be more safe to check for AudioTag value to be equal to 255, instead of defaulting to AACL.

I agree. Care to write it? I'm not sure where I should put it.

remitamine:
> Care to write it?

My netbook is broken, so I can't contribute to the project now.

> I'm not sure where I should put it.

I think that the check for `AudioTag` can be added after `FourCC` extraction and before the check for supported values.

fstirlitz:
The [downloader itself](/rg3/youtube-dl/blob/0aacd2deb1075e0d4d4b8b23b9a65b3967a1d658/youtube_dl/downloader/ism.py#L131-L132) also looks at the FourCC.

remitamine:
Yes I mean there no `FourCC` and `AudioTag` equal to 255 than set `FourCC` to `AACL`.

fstirlitz:
Anything else?

cawka:
Bump and +1 for this to be merged. I have tried the pull request locally and it worked perfectly for my needs.  I would also need slide extraction, but that is a separate story.

dstftw:
Cherry-picked. Thanks.

PR #11186, opened 2016-11-13
[stolaf] new extractor

fstirlitz:


youtube_dl/extractor/stolaf.py
```diff
@@ -0,0 +1,69 @@
...
+        playlist_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, playlist_id)
+
+        title = self._og_search_property('title', webpage)
+        description = self._og_search_property('description', webpage)
+
+        javascript = self._download_webpage(
+            'http://www.stolaf.edu/multimedia/components/eventlib.cfc'
+            '?method=getPlayerPlaylist&eventtype=e&eventid=%s&html5stream=%u' % (playlist_id, 3),
```
dstftw: Use `query` param.

youtube_dl/extractor/stolaf.py
```diff
@@ -0,0 +1,69 @@
...
+            playlist_id, 'Downloading playlist #%u' % (3))
+        thePlaylist = self._parse_json(
+            self._search_regex(r'(?s)thePlaylist\s*=\s*(\[.*?\]);', javascript, 'thePlaylist'),
+            playlist_id, transform_source=js_to_json)
+        # token = self._search_regex(r'n7kIjJed73\s*=\s*\'(.*?)\';', javascript, 'token')
```
dstftw: Any reason why it's here?

fstirlitz:
```js
	function streamPlayer6jw () {
		jwplayer('player').setup({
			/* ... */
			rtmp: {
				securetoken: n7kIjJed73
			},
			/* ... */
		});
	}
```

It needs to be passed to the `--token` option of `rtmpdump` for RTMP downloading to work.

rye:
Is this PR still not able to be merged? I'm not sure if the review is stale or not at this point.

Issue #11077, opened 2016-10-30
[Bilibili] Extraction broken


starkshaw:
## Please follow the guide below
- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your _issue_ (like that [x])
- Use _Preview_ tab to see how your issue will actually look like

---
### Make sure you are using the _latest_ version: run `youtube-dl --version` and ensure your version is _2016.10.31_. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.10.31**
### Before submitting an _issue_ make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
### What is the purpose of your _issue_?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---
### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your _issue_

---
### Description of your _issue_, suggested solution and other information

The download mission still can be done but the speed is limited at 10kB/s, but viewing from website is fully working. This issue appears shortly after Bilibili announced their membership plan early this month.

yan12125:
They switched to a new appkey.

starkshaw:
@yan12125 I remember Bilibili itself does not host videos all videos are ultimately hosted on other websites, can we track the origin and trigger the download from there?

yan12125:
> can we track the origin and trigger the download procedure from there?

If you can watch the video on the hosting website instead of bilibili, yes.

ztytian:
Bilibili now host some of their videos on their own servers instead of streaming from other servers

starkshaw:
@yan12125 That's the problem... I am not sure how to track its source.

starkshaw:
@ztytian What kind of videos are hosted on Bilibili-owned servers?

[...]

PR #11272, opened 2016-11-23
[jsinterp] Actual JS interpreter

sulyi:

---

- [ ] At least skimmed through [adding new extractor tutorial](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/rg3/youtube-dl#youtube-dl-coding-conventions) sections

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [x] New feature

---

I've started to implement an actual JavaScript syntax parser.
-- EDIT --
And moved on making an interpreter.


youtube_dl/jsinterp.py
```diff
@@ -18,13 +22,75 @@
...
+_ASSIGN_OPERATORS_RE = r'|'.join(re.escape(op) for op, opfunc in _ASSIGN_OPERATORS)
+
 _NAME_RE = r'[a-zA-Z_$][a-zA-Z_$0-9]*'
 
+_SINGLE_QUOTED = r"""'(?:[^'\\\\]*(?:\\\\\\\\|\\\\['"nurtbfx/\\n]))*[^'\\\\]*'"""
+_DOUBLE_QUOTED = r'''"(?:[^"\\\\]*(?:\\\\\\\\|\\\\['"nurtbfx/\\n]))*[^"\\\\]*"'''
+_STRING_RE = r'%s|%s' % (_SINGLE_QUOTED, _DOUBLE_QUOTED)
+
+_INTEGER_RE = r'%(hex)s|%(dec)s|%(oct)s' % {'hex': __HEXADECIMAL_RE, 'dec': __DECIMAL_RE, 'oct': __OCTAL_RE}
+_FLOAT_RE = r'%(dec)s\.%(dec)s' % {'dec': __DECIMAL_RE}
```
yan12125: .3

youtube_dl/jsinterp.py
```diff
@@ -18,13 +22,75 @@
...
+
+_BOOL_RE = r'true|false'
+# XXX: it seams group cannot be refed this way
+# r'/(?=[^*])[^/\n]*/(?![gimy]*(?P<reflag>[gimy])[gimy]*\g<reflag>)[gimy]{0,4}'
+_REGEX_RE = r'/(?=[^*])[^/\n]*/[gimy]{0,4}'
```
yan12125:
```
>>> re.match(r'/(?=[^*])[^/\n]*/[gimy]{0,4}', r'''/\/\/\//''')
<_sre.SRE_Match object; span=(0, 3), match='/\\/'>
```

youtube_dl/jsinterp.py
```diff
@@ -18,13 +22,75 @@
...
+    'int': _INTEGER_RE,
+    'float': _FLOAT_RE,
+    'str': _STRING_RE,
+    'bool': _BOOL_RE,
+    'regex': _REGEX_RE
+}
+
+# _ARRAY_RE = r'\[(%(literal)s\s*,\s*)*(%(literal)s\s*)?\]' % {'literal': _LITERAL_RE}
+# _VALUE_RE = r'(?:%(literal)s)|(%(array)s)' % {'literal': _LITERAL_RE, 'array': _ARRAY_RE}
+_CALL_RE = r'\.?%(name)s\s*\(' % {'name': _NAME_RE}  # function or method!
```
yan12125: Function calls are complex. For example:
```
from youtube_dl.jsinterp import JSInterpreter

jsi = JSInterpreter('''
function a(x) { return x; }
function b(x) { return x; }
function c()  { return [a, b][0](0); }
''')
print(jsi.call_function('c'))
```

youtube_dl/jsinterp.py
```diff
@@ -18,13 +22,75 @@
     ('+', operator.add),
     ('%', operator.mod),
     ('/', operator.truediv),
-    ('*', operator.mul),
+    ('*', operator.mul)
 ]
 _ASSIGN_OPERATORS = [(op + '=', opfunc) for op, opfunc in _OPERATORS]
 _ASSIGN_OPERATORS.append(('=', lambda cur, right: right))
 
+_RESERVED_RE = r'(?:function|var|(?P<ret>return))\s'
```
sulyi: I've added test, but I don't see any problem.
And what do you mean @yan12125  "not context-free"?
Didn't you wanted to say not regular?

youtube_dl/jsinterp.py
sulyi: Hopefully I've managed to improve on it a little.

youtube_dl/jsinterp.py
sulyi: I've added test.
Tokenizing seams to be fine, but I haven't migrated the interpreter and old one does not support this.

youtube_dl/jsinterp.py
yan12125: > They can't be multiline, can they?

Yep. According to [ECMA 262 5.1](http://www.ecma-international.org/ecma-262/5.1/), CR (U+000D), LF (U+000A), LS (U+2028) and PS (U+2029) are not allowed in RegExp literals

sulyi: Thx. I'll need to read that, a couple more times.

yan12125:
A notice: ```OrderedDict``` are not available in Python 2.6. There was a proposal to drop 2.6 but no consensus yet (#5697)

sulyi:
For some reason code this had been missed by code inspector of the IDE I'm working with. I'll try to come up a workaround. Thanks.

sulyi:
I've just realise my original idea, that **_next_statement** method would do the lexical analysis and **interpret_statement** the parsing is fraud. To yield a statement parsing has to had happened, since *Statement* is one of the symbols (along with *FunctionDeclaration*) replacing *SourceElement* in the syntactic grammar.

youtube_dl/jsinterp/__init__.py
```diff
@@ -1,3 +1,5 @@
 from .jsinterp import JSInterpreter
+from .jsgrammar import _NAME_RE
 
-__all__ = ['JSInterpreter']
\ No newline at end of file
+# ALERT stop usage of _NAME_RE!
```
sulyi: openload.py

yan12125: openload is broken again (#10408) and most likely it requires substantial rewriting. You can just remove _NAME_RE from openload.py.

sulyi: Ok, I'll remove _NAME_RE from jsinterp. It'll still exist in jsinterp.grammar, if one needs it.

youtube_dl/jsinterp/jsgrammar.py
```diff
@@ -1,12 +1,10 @@
 from __future__ import unicode_literals
 
 import re
-from enum import Enum
-# ALERT enum34 package dependency
-# it's backported
 
 
-class Token(Enum):
+class T(object):
```
sulyi: works fine, but debugging is a nightmare due to having only token values, I've changed it to namedtuple, after realizing it's supported in 2.6

sulyi:
Review request for: **test_jsinterp_parser.py**

youtube_dl/jsinterp/jsgrammar.py
```diff
@@ -0,0 +1,82 @@
...
+               'OP', 'AOP', 'UOP', 'LOP', 'REL',
+               'COMMENT', 'TOKEN', 'PUNCT',
+               'NULL', 'BOOL', 'ID', 'STR', 'INT', 'FLOAT', 'REGEX',
+               'REFLAGS', 'REBODY',
+               'BLOCK', 'VAR', 'EXPR', 'IF', 'ITER', 'CONTINUE', 'BREAK', 'RETURN', 'WITH', 'LABEL', 'SWITCH',
+               'THROW', 'TRY', 'DEBUG',
+               'ASSIGN', 'MEMBER', 'FIELD', 'ELEM', 'CALL', 'ARRAY', 'COND', 'OPEXPR',
+               'RSV')
+
+Token = namedtuple('Token', _token_keys)._make(_token_keys)
```
sulyi: Only technically an object (should be treated as a class)

test/test_jsinterp.py
```diff
@@ -5,7 +5,10 @@
 # Allow direct execution
 import os
 import sys
-import unittest
+if sys.version_info < (2, 7):
```
sulyi: Is it ok to use unittest2?

test/test_jsinterp.py
```diff
@@ -5,7 +5,10 @@
...
+    import unittest2 as unittest
+else:
+    import unittest
```
sulyi: Is it ok to use unittest2?

yan12125: That's fine. Don't forget to add the dependency in requirements.txt and install it in .travis.yml. See https://github.com/rg3/youtube-dl/pull/7852/files

sulyi: Done and done. Thx, version limit was very helpful.

yan12125:
https://github.com/rg3/youtube-dl/pull/11272/commits/9b5e55a45afea978fc963bc50bb074bddd258927 looks a good way. https://pypi.python.org/pypi/unittest-data-provider may be helpful for generating test cases automatically, or you may implement your own test case generator like what ```test_download.py``` does.

test/test_jsinterp_parser.py
```diff
@@ -505,29 +505,114 @@ def test_precedence(self):
...
+                 ])
+                  )
+             ])),
+            (Token.FUNC, 'z',
+             [],
+             (Token.BLOCK, [
+                 (Token.RETURN, (Token.EXPR, [
+                     (Token.ASSIGN, None, (Token.OPEXPR, [
+                         # Not sure about this one
+                         (Token.MEMBER, (Token.ID, 'y'), None, (Token.CALL, [
```
sulyi: If I'm not mistaken args is not **None** if MemberExpression is a NewExpresion. Therefore this is correct.

yan12125:
If you're looking for some dirty Javascripts, here's the latest site that may benefit from a real JS interpreter, Bilibili: http://static.hdslb.com/player/js/bilibiliPlayer.min.js?v=3db87237
Ref: #11077

sulyi:
At the moment I'm looking for very small scripts that, uses specific JS features. Like the ones in my comment,of which review I've called out.

yan12125:
I guess Mozilla's test suite is enough. Or V8's test suite, ChakraCore's test suite, etc. Spidermonkey does not provide a standalone repository, and V8 requires Google's mysterious build system, so maybe ChakraCore is a choice.

sulyi:
I'll check them out hopefully one of them be a suit~e~able.

sulyi:
Can we talk about some kind of milestone and/or review criteria?
I'm thinking a fully functional parser and passing current interpreter tests.

yan12125:
The first milestone can be when most ES5 syntaxes are parsed

sulyi:
Gotcha, tests included I reckon, but I'll postpone suite integration.

sulyi:
I'm done wit the parser. Right now I'm adding ASTs to tests, and fixing bugs.

test/test_jsinterp_parser.py
```diff
@@ -880,22 +880,56 @@ def test_label(self):
...
 
-    @unittest.skip('Test not yet implemented: missing ast')
     def test_function_expression(self):
-        # ASAP function expression test
         jsi = JSInterpreter('''
             function f() {
                 var add = (function () {
                     var counter = 0;
-                    return function () {return counter += 1;}
+                    return function () {return counter += 1;};
```
sulyi: automatic semicolon insertion

test/test_jsinterp_parser.py
```diff
@@ -842,19 +842,44 @@ def test_for_in(self):
         ast = []
         self.assertEqual(list(jsi.statements()), ast)
 
-    @unittest.skip('Test not yet implemented: missing ast')
     def test_do(self):
-        # ASAP do statement test
         jsi = JSInterpreter('''
                     function f(x){
-                        i = 1
+                        i = 1;
```
sulyi: automatic semicolon insertion

test/test_jsinterp_parser.py
```diff
@@ -882,19 +882,44 @@ def test_do(self):
         ]
         self.assertEqual(list(jsi.statements()), ast)
 
-    @unittest.skip('Test not yet implemented: missing ast')
     def test_while(self):
-        # ASAP while statement test
         jsi = JSInterpreter('''
             function f(x){
-                i = 1
+                i = 1;
```
sulyi: automatic semicolon insertion

sulyi:
Parser is done. Beside some features (automatic semicolon insertion, and in for initialization), I should have marked most of those. I've also marked some parenthesis checks I've missed, but those are less bold.
I've really hoped I can also tackle some operator issues (**prefix**, **postfix**, **delete**, **void**, **typeof**, **in**, **instanceof**). These are going to need some attention both in parsing and interpreting.
At the current state of the test suite that would start making things messy. I might do it anyway, but sorting that out is more and more pressing.

yan12125:
Hmm, what's "automatic semicolon insertion"?

sulyi:
A feature in JS since a long time. Some of the semicolons amended by the interpreter automagicly to statements. I've marked some examples in tests.
I've run the tests in `Node.js` or `V8` (whatever you call it) and `Rhino` first to see if they are actually a JS script, and got no trouble with omitted semicolons, that's how those got in the first place.

yan12125:
Thanks for the link. I guess that feature is necessary as lots of JS compressors remove tokens as much as possible.

sulyi:
True that. Yet that's entirely a parser feature, therefore it can be added later and there's also a lot to do on the interpreter (function call, property handle, built-in methods).

youtube_dl/jsinterp/tstream.py
```diff
@@ -137,6 +138,10 @@ def _next_token(self, pos=0):
                 elif token_id is Token.ID:
                     yield (token_id, token_value, pos)
                 elif token_id in _operator_lookup:
+                    # FIXME signed values
```
sulyi: I'm not sure this is a valid fix.

youtube_dl/jsinterp/tstream.py
```diff
@@ -137,6 +138,10 @@ def _next_token(self, pos=0):
...
+                    if (token_id is Token.OP and token_value in ('-', '+') and
+                            self._last[0] not in token_keys and self._last[0] is not Token.PCLOSE):
+                        token_id = Token.UOP
```
sulyi: I'm not sure this is a valid fix.

sulyi: I'm not sure this is a valid fix.

test/jstests/object_literal.py
```diff
@@ -0,0 +1,59 @@
+from youtube_dl.jsinterp.jsgrammar import Token
+from youtube_dl.jsinterp.tstream import _ASSIGN_OPERATORS, _OPERATORS
+
+skip = {'i': 'Interpreting object literals not yet implemented'}
+
+tests = [
+    {
```
sulyi: needs assert(s)

test/jstests/object_literal.py
```diff
@@ -0,0 +1,59 @@
...
+        'code': '''
+            function f() {
+                var o = {
+                    a: 7,
+                    get b() { return this.a + 1; },
+                    set c(x) { this.a = x / 2; }
+                };
+                return o;
+            }
+            ''',
```
sulyi: needs assert(s)

sulyi:
JSInterpreter extract_function is kept as a fallback, used solely by call_function method.
Usage of this should be discouraged, since calling function declared outside of function body will fail.

PiotrDabkowski:
Hey, I noticed this project and I think it's interesting. I wrote Js2Py - something very similar to the interpreter your are developing here - but since it works by translating Js to Python it is bound by the limits of Python syntax (like maximum number of arguments of a function is 256 or no continue iside finally statement) and may not be suitable due to security concerns. 

However, I also wrote [pyjsparser](https://github.com/PiotrDabkowski/pyjsparser) which is a JS parser based on esprima.js and it supports whole ECMA 5 and parts of 6. Esprima is hevily tested so pyjsparser is virtually flawless. I was wondering why don't you adapt the pyjparser in your project? 

Also, about 1 month ago I started implementing a proper interpreter with JsCode -> Bytecode conversion. But I temporarily abandoned the project after implementing the interesting parts ;) I don't think Js2Py will be useful here as we took a completely different approach.

Regarding built-in objects (they are certainly not a fun thing to implement :) )  - Js2Py has them all implemented so if you want I think it would be possible to use this implementation here.

sulyi:
That sound like a life saver, I'm going to definitely check it out.
I can't make any meaningful comment till then, but I can tell that I don't want to generate bytecode and I'm prepared to even sacrifice a lot of performance just to not do that.
I've used narcissus and rhino for inspiration and guidance when I had difficulty with the specs.
And thanks, I'm hoping it will be helpful.

test/jstests/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+    do_loop,
+    while_loop,
+    label,
+    func_expr,
+    object_literal,
+    try_statement,
+    with_statement,
+    debug,
+    unshift
+)
```
yan12125: In Python usually imports are sorted alphanumerically. Does the current order mean something?

test/test_jsinterp.py
```diff
@@ -2,116 +2,68 @@
 
 from __future__ import unicode_literals
 
-# Allow direct execution
```
yan12125: Direct execution is still possible. Accidental removal?

test/test_jsinterp.py
```diff
@@ -2,116 +2,68 @@
...
 
+class TestJSInterpreter(unittest.TestCase):
+    def setUp(self):
+        self.defs = defs
+
+
+def generator(test_case, name):
+    def test_template(self):
+        for test in test_case['subtests']:
+            jsi = JSInterpreter(test['code'], variables=None if 'globals' not in test else test['globals'])
```
yan12125: ```None if 'globals' not in test else test['globals']``` => ```test.get('globals')```

test/test_jsinterp.py
```diff
@@ -2,116 +2,68 @@
...
+                log.debug('No asserts, skipping subtest')
+
+    log = logging.getLogger('TestJSInterpreter.%s' % name)
+
+    if 'i' not in test_case['skip']:
+        reason = False
+    else:
+        reason = test_case['skip']['i']
+
+    return test_template if not reason else unittest.skip(reason)(test_template)
```
yan12125: unittest.skip(reason)(test_template) looks counter-intuitive. Maybe raising SkipTest or calling skipTest() in the test function is better?

test/test_jsinterp.py
```diff
@@ -2,116 +2,68 @@
...
+
+
+# And add them to TestJSInterpreter
+for n, tc in enumerate(defs):
+    if 'i' not in tc['skip'] or tc['skip']['i'] is not True:
```
yan12125: Similarly, ```if tc['skip'].get('i')```

test/test_jsinterp_parse.py
```diff
@@ -0,0 +1,76 @@
...
+else:
+    import unittest
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+from youtube_dl.jsinterp import JSInterpreter
+from .jstests import gettestcases
+
+
+def traverse(node, tree_types=(list, tuple)):
+    if type(node) == zip:
```
yan12125: ```isinstance(node, zip)```

I'm afraid this line is fragile as Python spec does not require ```zip``` as a class. Of course it's fine if there's no other solutions.

test/jstests/__init__.py
sulyi: It's kind of shows the progress. Most of the test at the end are skipped. At least for now.

sulyi: Well, it's kind of represents the progress. At least for now. Most of the tests at the end are skipped yet.

sulyi:
I've created a tag here, since interpreter could be even bigger.

sulyi:
Would refactoring parser into separate module help with size issue?

sulyi:
Hope refactoring parser a tiny bit helps with size issue.

sulyi:
I don't think there are any more strings having ambiguity issues, since I remember getting lazy with those two.

sulyi:
After some cool-off time things start to catch up on the `jsbuilt-ins` branch. Not quite ready to merge, but I'd like to make a friendly reminder that, even if a bit slower and not pushing on this branch, I'm still working on this.

yan12125:
Sorry but I have neither time nor knowledge to review this. I'd like to merge it as-is. Could you move new codes to another module? (e.g., ```jsinterp2```)

Kcchouette:
Hello @sulyi, are you still working on it?

Tatsh:
Would like to see this. It is related to my incomplete PR https://github.com/rg3/youtube-dl/pull/14943 . Also I created this number parser according to the ECMA specifications:

```python
from __future__ import print_function                                                                                                                        
import math                                                                                                                                                  
import re                                                                                                                                                    
                                                                                                                                                             
regex = re.compile(r'''(?:\s+)? # StrNumericLiteral.StrWhiteSpace (optional)                                                                                 
            ( # StringNumericLitral.StrDecimalLiteral                                                                                                        
                (?P<inf_sign>[\+\-])? # optional sign                                                                                                        
                (?P<inf>Infinity)|
                [\+\-]?(?P<nan>NaN)|  # not documented, but Node returns NaN
                [\+\-]?(?:  # DecimalDigits . DecimalDigits(opt) ExponentPart(opt)
                    (?:\d+)? # DecimalDigits
                    (?:\.)
                    (?:\d+)? # DecimalDigits
                    (?:[eE] # ExponentPart.ExponentIndicator
                        [\+\=]?  # optional sign
                        (?:\d+)
                    )?
                )|
                0(?P<binary>[bB])(?P<binary_number>[01]+)|
                0(?P<octal>[oO])(?P<octal_number>[01234567]+)|
                0(?P<hex>[xX])(?P<hex_number>[0-9a-fA-F]+)
            )(?:\s+)?$''', re.X)

def getnan():
    try:
        return math.nan
    except AttributeError:
        pass
    return float('nan')


def getinf(mult=1):
    try:
        ret = math.inf
    except AttributeError:
        ret = float('inf')
    return mult * ret


def can_be_int(inp, out):
    return str(out)[-2:] == '.0' and ('.0' not in inp and not inp.endswith('.'))


def conditional_int(inp):
    # Values with leading zeros like 01.0 and 01. should fail
    if len(inp) > 1 and inp[0] == '0':
        raise SyntaxError('unexpected number; given "{}"'.format(inp))
    out = float(inp)
    return int(str(out)[:-2], 10) if can_be_int(inp, out) else out


def parse_numeric(x):
    m = re.match(regex, x)
    if not m:
        raise SyntaxError('invalid nor unexpected token; given "{}"'.format(x))

    ret = m.group(0).strip()
    groups = m.groupdict()

    if groups['inf']:
        sign = -1 if groups['inf_sign'] == '-' else 1
        ret = getinf(sign)
    elif groups['nan']:
        ret = getnan()
    else:
        base = key = None
        if groups['binary']:
            base = 2
            key = 'binary_number'
        elif groups['octal']:
            base = 8
            key = 'octal_number'
        elif groups['hex']:
            base = 16
            key = 'hex_number'
        if key and base:
            ret = int(groups[key], base)
        else:
            try:
                ret = conditional_int(ret)
            except ValueError as e:
                raise SyntaxError('invalid or unexpected token: given "{}"'.format(x))
    return ret
```

Passes all these, exactly the same as how Chrome parses:

```python
parse_numeric('.1')
parse_numeric('1.')
parse_numeric('Infinity')
parse_numeric('-Infinity')
parse_numeric('+Infinity')
parse_numeric('NaN')
parse_numeric('-NaN')
parse_numeric('+NaN')
try:
    parse_numeric('01.0')
except SyntaxError as e:
    print('01.0 -> {}'.format(e))
try:
    parse_numeric('01.')
except SyntaxError as e:
    print('01. -> {}'.format(e))
parse_numeric('1.0')
parse_numeric('1.1')
parse_numeric('1.e5')
parse_numeric('1.E5')
parse_numeric('1.1e5')
parse_numeric('1.1E6')
parse_numeric('-1.1E6')
parse_numeric('+1.1')
parse_numeric('0b101')
parse_numeric('0B101')
parse_numeric('0o755')
parse_numeric('0O755')
parse_numeric('0xaf')
parse_numeric('0XDEADBEEF9')
```

sulyi:
Been a while. I'm not yet up to speed with my previous work, but managed to fix some issues with "jsbuilt-ins" and merge that branch which has some of the infrastructure ready, but most of it is still not implemented. And this needs some work beyond that before it's able to replace the functionality of the current solution, e.g. interpretation of member expression. As this [test result][2] shows, I've ran it with changing this [line][3] to:
```python
 from ..jsinterp2 import JSInterpreter
```

I've totally missed @yan12125's [this][1] reply. It's a shame, really if I'd known... anyway @Kcchouette now I am intend to work on it again, but I wasn't till now. I've even considered closing it.

[1]: https://github.com/rg3/youtube-dl/pull/11272#issuecomment-290069838
[2]: https://gist.github.com/sulyi/f902edc51210a1d8bc51e7287bb4c91a
[3]: https://github.com/sulyi/youtube-dl/blob/e44a25227e926ad9be242716590efd1f4b8f13e0/youtube_dl/extractor/youtube.py#L15

sulyi:
Note: I've added a [new test](https://github.com/sulyi/youtube-dl/blob/b856d55847307311e03e98a56075171bc86e19ea/test/js2tests/stringprototype.py), for slicing strings.

test/test_jsinterp.py
```diff
@@ -1,117 +1,89 @@
...
 
-from youtube_dl.jsinterp import JSInterpreter
+from youtube_dl.jsinterp2 import JSInterpreter
+from .js2tests import gettestcases
 
+defs = gettestcases()
+# set level to logging.DEBUG to see messages about missing assertions
+logging.basicConfig(stream=sys.stderr, level=logging.WARNING)
 
-class TestJSInterpreter(unittest.TestCase):
```
yan12125: Don't remove tests for the original JSInterpreter. Before jsinterp2 replaces jsinterp, tests should be there to guarantee nothing gets broken by accident.

test/test_jsinterp.py
```diff
@@ -17,7 +17,7 @@
     import unittest
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
-from youtube_dl.jsinterp2 import JSInterpreter
+from youtube_dl.jsinterp import JSInterpreter
 from .js2tests import gettestcases
 
 defs = gettestcases()
```
sulyi: You can use logging level to check skipped assertions

sulyi:
Interesting: in 50/72 rule LN is included!

sulyi:
Changing logging level [here](https://github.com/sulyi/youtube-dl/blob/1f40e3ef633e0a12b74c40e20cbbf7f7afb20b82/test/test_jsinterp.py#L25) you can log skipped assertions.

test/test_jsinterp.py
sulyi: Yes, fixed.

test/test_jsinterp.py
sulyi: As right now both `test_jsinterp_orig` and `test_jsinterp` exist later is using my suite, but should do the same.

yan12125: I'd prefer leaving `test_jsinterp.py` as-is and add new tests in `test_jsinterp2.py`, so that the test file name matches the interpreter module name.

sulyi: Here's a different idea: new tests: `test_js2test_legacy.py`, `test_js2test_jsinterp2.py`, `test_js2test_jsinterp2_parse.py`; and the original: `test_jsinterp.py` kept untouched.
The tests `test_js2test_legacy.py` performs tests on the original `jsinterp.py`, but using the new test suite.
I'm not a big fan of these names, but this is what I could come up with.
On one hand there's a suggestion to the method the testcases are generated, on the other they're bit convoluted.

youtube_dl/extractor/youtube.py
```diff
@@ -12,7 +12,7 @@
 import traceback
 
 from .common import InfoExtractor, SearchInfoExtractor
-from ..jsinterp import JSInterpreter
+from ..jsinterp2 import JSInterpreter
```
sulyi: Accidental changes, but oddly passed `test_youtube_signature`. For me it fails due to `JSArrayPrototype._slice` doesn't handle arguments correctly since not being implemented yet.

youtube_dl/jsinterp2/jsinterp.py
```diff
@@ -241,9 +241,16 @@ def interpret_expression(self, expr):
...
         elif name is Token.ID:
             # XXX error handling (unknown id)
-            ref = (self.this[expr[1]] if expr[1] in self.this else
-                   self.global_vars[expr[1]])
-        
+            id = expr[1]
+            try:
+                ref = (self.this[id] if id in self.this else
+                       self.global_vars[id])
+            except KeyError:
```
sulyi: I think `JSInterpreter#extract_function` is useful filtering code before execution, therefore I'd like to continue supporting it. But here it needs to get object from outer context, and that behaviour is **not** in spec. I'd like to suggest a flag that disables these kind of features.

Tatsh:
This is of course, very neat. But a lot of Chrome's (and maybe others) standard library for many things are implemented in JavaScript. Instead of making the built-ins (like `String.prototype.match`) in Python, why not write them in JavaScript (where possible)?

Tatsh:
Main thing is to get the basics of the interpreter in, which includes the built-in types, and then the JavaScript portions can be written very similarly to how polyfills are written today. So you would not need to implement `Array.isArray()` in Python if you have the `===` operator working for comparing function references and `.constructor` property working on all objects. Then code `Array.isArray = function (x) { return x.constructor === Array };` and make this load at runtime before anything else.

I can take a look later since this does interest me. Mainly this was needed for me to get past CloudFlare anti-DDoS without needing cookies.

sulyi:
Well, I've already implemented operator `===` and most constructors I believe, but neither have been tested properly.
Also spec of `isArray` states:
1. If `Type(arg)` is not Object, return false.
1. If the value of the [[Class]] internal property of `arg` is "Array", then return true.
1. Return false.

I'm not sure when 1. would execute or whether your solution takes care of it.
My solution for this particular function would look like this:
```python
from .internals import jstype, object_type

def _is_array(arg):
   if jstype(arg) is not object_type:
      return False
   if arg.jsclass == 'Array':
      return True
  return False
```
I'm not sticking to it or claim that it's elegant, but this is how I'm able to solve the task at hand.
I have way less xp programming in js than in py. So, if you could tell me what needs to be get done for this to work I can probably take care of it, but to actually implement it that would be much harder to do it like this on my own.

sulyi:
I've added feature to lexer (`tstream.TokenStream`) ability to handle line terminators. This is the first step in implementing correct line reporting.

This is also useful to have in my other plan to change the test suite using json files instead of py to generate test cases. My reasoning behind it is if `jsparser.Parser` would support converting  AST to [estree][1] or some pretty similar format, it would be possible to easily compare it against the output of [acorn][2] or some other parser.

[1]: https://github.com/estree/estree
[2]: https://github.com/acornjs/acorn

PR #11312, opened 2016-11-27
[spike] Extract MGID via custom mobile URL

malept:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Currently, trying to use `youtube-dl` (version 2016.11.27) to download a full episode from Spike results in the exception:

```RegexNotFoundError: Unable to extract video url```

 I've added some code so that if the conventional MGID detection fails, it falls back to searching for the mobile app's custom URL format (`spikenetworkapp://$video_type/$episode_uuid`) - both variables from the URL can be used to generate a valid MGID.

### DIFF
--- a/youtube_dl/extractor/mtv.py
+++ b/youtube_dl/extractor/mtv.py
@@ -13,6 +13,7 @@ from ..utils import (
     fix_xml_ampersands,
     float_or_none,
     HEADRequest,
+    NO_DEFAULT,
     RegexNotFoundError,
     sanitized_Request,
     strip_or_none,
@@ -201,7 +202,7 @@ class MTVServicesInfoExtractor(InfoExtractor):
             [self._get_video_info(item) for item in idoc.findall('.//item')],
             playlist_title=title, playlist_description=description)
 
-    def _extract_mgid(self, webpage, fatal=True):
+    def _extract_mgid(self, webpage, default=NO_DEFAULT):
         try:
             # the url can be http://media.mtvnservices.com/fb/{mgid}.swf
             # or http://media.mtvnservices.com/{mgid}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: youtube_dl/extractor/spike.py [hunk 1]
```diff
@@ -32,3 +34,12 @@ class SpikeIE(MTVServicesInfoExtractor):
...
     _FEED_URL = 'http://www.spike.com/feeds/mrss/'
     _MOBILE_TEMPLATE = 'http://m.spike.com/videos/video.rbml?id=%s'
+    _CUSTOM_URL_REGEX = re.compile(r'spikenetworkapp://([^/]+/[-a-fA-F0-9]+)')
+
+    def _extract_mgid(self, webpage):
+        mgid = super(SpikeIE, self)._extract_mgid(webpage, fatal=False)
+        if mgid is None:
+            url_parts = self._search_regex(self._CUSTOM_URL_REGEX, webpage, 'episode_id')
+            video_type, episode_id = url_parts.split('/', 1)
+            mgid = 'mgid:arc:{}:spike.com:{}'.format(video_type, episode_id)
```
dstftw: `{}` won't work in python 2.6.

malept: Fixed in https://github.com/rg3/youtube-dl/pull/11312/commits/53ebfd701f8915b99a54323107d4c84d3f2ad47f.

dstftw:
All existing spike tests work fine current master. Provide example URLs that fail and add tests,

malept:
Added a full episode to the list of Spike tests in https://github.com/rg3/youtube-dl/pull/11312/commits/d4b9a720dc839e0cba87519a21d00fcb7321246c, which fails on master.

youtube_dl/extractor/mtv.py
```diff
@@ -201,7 +201,7 @@ def _get_videos_info_from_url(self, url, video_id):
             [self._get_video_info(item) for item in idoc.findall('.//item')],
             playlist_title=title, playlist_description=description)
 
-    def _extract_mgid(self, webpage):
+    def _extract_mgid(self, webpage, fatal=True):
```
dstftw: Instead `default` should be used in order to mute the warning.

malept: That's funny, because I initially wrote this fix using `default`, then I found that `fatal` existed, and that seemed like a cleaner way to write it.

malept: Uses `default` as of https://github.com/rg3/youtube-dl/pull/11312/commits/9d60d1851a221f9369db40292f71cb56c12116d6.

PR #11366, opened 2016-12-05
Add extractor for NHL TV

joshka:


joshka:
Anyone care to comment?

joshka:
@dstftw care to comment?

Manouchehri:
Can't seem authenticate with my paid Rogers account.

```bash
nitro:youtube-dl dave$ python -m youtube_dl https://www.nhl.com/tv/2016020321/221-1003765/46561403 -u "<removed>@rogers.com" --verbose
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'https://www.nhl.com/tv/2016020321/221-1003765/46561403', u'-u', u'PRIVATE', u'--verbose']
Type account password and press [Return]:
[debug] youtube-dl version 2016.11.27
[debug] Git HEAD: d9218ed9a
[debug] Python version 2.7.13 - Darwin-16.4.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg git-2017-01-07-fd01040, ffprobe git-2017-01-07-fd01040
[debug] Proxy map: {}
[nhl.com:nhltv] Requesting OAuth access token
[nhl.com:nhltv] Logging in
ERROR: Unable to log in: HTTP Error 401: Unauthorized (caused by HTTPError()); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
  File "youtube_dl/extractor/common.py", line 404, in _request_webpage
    return self._downloader.urlopen(url_or_request)
  File "youtube_dl/YoutubeDL.py", line 2001, in urlopen
    return self._opener.open(req, timeout=self._socket_timeout)
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 435, in open
    response = meth(req, response)
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 548, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 473, in error
    return self._call_chain(*args)
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 407, in _call_chain
    result = func(*args)
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 556, in http_error_default
    raise HTTPError(req.get_full_url(), code, msg, hdrs, fp)
```

joshka:
@Manouchehri this currently is only handling login with NHL TV logins, not Rogers. It looks like Rogers login is handled fairly similarly (different url, slight different JSON login request). This should be a trivial change, but will require an extra option to be passed to the downloader. I'll take a look sometime soon.

@dstftw you can create a free account at https://account.nhl.com/ui/pub/CreateProfile
I've deliberately chosen a video for the test that free accounts are allowed to download. I figured that creating an account and publishing the credentials in this repo was not a good idea.

joshka:
@Manouchehri can you please give that a try for me. I don't have a rogers acct to test with

Manouchehri:
```bash
nitro:youtube-dl dave$ python -m youtube_dl --verbose "https://www.nhl.com/tv/2016020594/221-1004038" -F --auth-provider rogers --username "<removed>@rogers.com"
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--verbose', u'https://www.nhl.com/tv/2016020594/221-1004038', u'-F', u'--auth-provider', u'rogers', u'--username', u'PRIVATE']
Type account password and press [Return]:
[debug] youtube-dl version 2016.11.27
[debug] Git HEAD: affd0185c
[debug] Python version 2.7.13 - Darwin-16.4.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg git-2017-01-07-fd01040, ffprobe git-2017-01-07-fd01040
[debug] Proxy map: {}
[nhl.com:nhltv] Requesting OAuth access token
[nhl.com:nhltv] Logging in
ERROR: Unable to log in: HTTP Error 400: Bad Request (caused by HTTPError()); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
  File "youtube_dl/extractor/common.py", line 404, in _request_webpage
    return self._downloader.urlopen(url_or_request)
  File "youtube_dl/YoutubeDL.py", line 2002, in urlopen
    return self._opener.open(req, timeout=self._socket_timeout)
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 435, in open
    response = meth(req, response)
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 548, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 473, in error
    return self._call_chain(*args)
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 407, in _call_chain
    result = func(*args)
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 556, in http_error_default
    raise HTTPError(req.get_full_url(), code, msg, hdrs, fp)
```

joshka:
I fixed a typo in the JSON (rogers should be roger)

Manouchehri:
Seems to work now. Is there an easy way to automatically cache the OAuth access token in youtube-dl? I get locked out after only 3-4 tries.

joshka:
@Manouchehri thanks for testing this for me and confirming that it works. I'm not yet caching the oauth token, but it's next up on my list. I've been hit by the same problem during testing. It seems to lock out for an hour or so.

@dstftw any stylistic changes you want to the code before merging?

dstftw:
I'm not going to merge this until accounts for testing are provided. But looking at the code ,there are lots of code duplication and optional metadata extraction that may break the extraction if missing. Read coding conventions and fix all the issues.

joshka:
@dstftw can you provide some more information on how you'd like test accounts to be provided? I've given you the url to create an account above. https://account.nhl.com/ui/pub/CreateProfile

Manouchehri:
Here's a test account. 

Username: `uzg09977@dsiay.com`
Password: `uzg09977@dsiay.com`

joshka:
@dstftw I think I've addressed your points about coding conventions, but I could have missed something you're looking for. Can you please annotate the PR with your review notes when you get a chance?

@Manouchehri after implementing the cached credentials, it looks like the too many usages problem is based on the the number of video downloads rather than on the login attempts :(
I just locked out that test account after 4 attempts to download the same video (my account was already locked, so I used it for that test). The Web UI seems to still work when the account is locked if I recall correctly, so there must be something else going on here.

joshka:
@dstftw Can you please suggest any needed changes to meet the coding conventions? What do you need in order to get this merged?

joshka:
I'm looking for some feedback on what it would take to get this PR merged.

@dstftw seems to have gone silent on responding.
@remitamine or @yan12125 can you please take a look for me?

dstftw:
It does not even work:
```
PS C:\Dev\youtube-dl\master> py -3.6 .\youtube_dl\__main__.py -u <snip> -p <snip> -v https://www.nhl.com/tv/2016020321/221-1003765/46561403
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-u', 'PRIVATE', '-p', 'PRIVATE', '-v', 'https://www.nhl.com/tv/2016020321/221-1003765/46561403']
[debug] Encodings: locale cp1251, fs utf-8, out utf-8, pref cp1251
[debug] youtube-dl version 2017.02.10
[debug] Git HEAD: 46e33ed
[debug] Python version 3.6.0 - Windows-10-10.0.10240-SP0
[debug] exe versions: ffmpeg N-83152-gf7e9275, ffprobe N-83152-gf7e9275, rtmpdump 2.4
[debug] Proxy map: {}
[nhl.com:nhltv] Logging in
[nhl.com:nhltv] Requesting OAuth access token
  File ".\youtube_dl\__main__.py", line 19, in <module>
    youtube_dl.main()
  File "C:\Dev\youtube-dl\master\youtube_dl\__init__.py", line 453, in main
    _real_main(argv)
  File "C:\Dev\youtube-dl\master\youtube_dl\__init__.py", line 443, in _real_main
  File "C:\Dev\youtube-dl\master\youtube_dl\YoutubeDL.py", line 1797, in download
  File "C:\Dev\youtube-dl\master\youtube_dl\YoutubeDL.py", line 697, in extract_info
    ie_result = ie.extract(url)
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\common.py", line 368, in extract
    self.initialize()
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\common.py", line 362, in initialize
    self._real_initialize()
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\nhl.py", line 478, in _real_initialize
    self._login()
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\nhl.py", line 486, in _login
    self.api.login(username, password, auth_provider)
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\nhl.py", line 374, in login
    access_token = self._get_oauth_access_token()
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\nhl.py", line 426, in _get_oauth_access_token
    'Unable to get OAuth access token')
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\common.py", line 561, in _download_json
    encoding=encoding, data=data, headers=headers, query=query)
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\common.py", line 529, in _download_webpage
    res = self._download_webpage_handle(url_or_request, video_id, note, errnote, fatal, encoding=encoding, data=data, headers=headers, query=query)
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\common.py", line 436, in _download_webpage_handle
    urlh = self._request_webpage(url_or_request, video_id, note, errnote, fatal, data=data, headers=headers, query=query)
  File "C:\Dev\youtube-dl\master\youtube_dl\extractor\common.py", line 416, in _request_webpage
    return self._downloader.urlopen(url_or_request)
  File "C:\Dev\youtube-dl\master\youtube_dl\YoutubeDL.py", line 2007, in urlopen
    return self._opener.open(req, timeout=self._socket_timeout)
  File "C:\Python\Python36\lib\urllib\request.py", line 524, in open
    req = meth(req)
  File "C:\Python\Python36\lib\urllib\request.py", line 1248, in do_request_
    raise TypeError(msg)
TypeError: POST data should be bytes, an iterable of bytes, or a file object. It cannot be of type str.
```

youtube_dl/extractor/nhl.py
```diff
@@ -349,3 +354,201 @@ def _real_extract(self, url):
...
             'formats': formats,
         }
+
+
+class NHLApi:
+    def __init__(self, extractor, auth=None):
+        self.extractor = extractor
+        self.auth = auth
+        if auth:
+            extractor.to_screen("Using cached credentials. Use the --rm-cache-dir option to remove.")
```
dstftw: `--rm-cache-dir` wipes the whole cache thus should never be suggested to use.

youtube_dl/extractor/nhl.py
```diff
@@ -349,3 +354,201 @@ def _real_extract(self, url):
...
+
+    def _real_initialize(self):
+        auth = self._downloader.cache.load(self._NHLTV_CACHE, 'auth')
+        self.api = NHLApi(self, auth)
+        if auth is None:
+            self._login()
+
+    def _login(self):
+        (username, password) = self._get_login_info()
+        if username is None:
```
dstftw: This prevents from authenticating with `--cookies`.

youtube_dl/options.py
```diff
@@ -350,6 +350,10 @@ def _scrub_eq(o):
         '--video-password',
         dest='videopassword', metavar='PASSWORD',
         help='Video password (vimeo, smotri, youku)')
+    authentication.add_option(
+        '--auth-provider',
+        dest='auth_provider', metavar='AUTH_PROVIDER',
+        help='Authentication provider (nhl, rogers, ...)')
```
dstftw: If this auth provider thing is only related to nhl.com site then it should have `nhl` in it's name. Or if you suggest this to be a generic option to be used across extractors you have to make a complete documented proposal of how this is supposed to be used.

youtube_dl/extractor/nhl.py
```diff
@@ -349,3 +354,201 @@ def _real_extract(self, url):
...
+        video_id, game_id = mobj.group('id'), mobj.group('gameId')
+        # At this point we may have a game_id without a video_id. The next call
+        # ensures that we get a video_id for the desired game.
+        video_id, title, timestamp = self.extract_game_info(video_id, game_id)
+        m3u8_url, media_auth = self.extract_playlist_url_and_auth(video_id)
+        m3u8_url = self.get_60fps_playlist(m3u8_url)
+
+        # media auth cookie authenticates the specific download url
+        self._set_cookie('nhl.com', 'mediaAuth_v2', media_auth)
+        formats = self._extract_m3u8_formats(m3u8_url, video_id, 'ts', m3u8_id='hls')
```
dstftw: `ext` should be mp4.

youtube_dl/extractor/nhl.py
```diff
@@ -349,3 +354,201 @@ def _real_extract(self, url):
...
+    }, {
+        'url': 'https://www.nhl.com/tv/2016020373',
+        'only_matching': True,
+    }]
+
+    def _real_initialize(self):
+        auth = self._downloader.cache.load(self._NHLTV_CACHE, 'auth')
+        self.api = NHLApi(self, auth)
+        if auth is None:
+            self._login()
```
dstftw: You cache cookies for authentication but have no mechanism for invalidation. If this cached cookie expires or becomes invalid for some reason you'll get stuck trying to authenticate with it over and over again. And due to it is preferred over credentials, authentication won't work at all. This should be detected and cached data should be cleared.

README.md
```diff
@@ -357,6 +357,7 @@ which means you can modify it, redistribute it or use it however you like.
     -2, --twofactor TWOFACTOR        Two-factor auth code
     -n, --netrc                      Use .netrc authentication data
     --video-password PASSWORD        Video password (vimeo, smotri, youku)
+    --auth-provider PROVIDER         Authentication provider (nhl, rogers)
```
dstftw: This part is autogenerated and should not be touched by hand.

youtube_dl/extractor/nhl.py
joshka: I agree. Would you suggest adding a specific parameter like `--nhltv-clear-credentials`, or do you have a better suggestion for how to handle this that's more in line with your view of how youtube-dl should work?

youtube_dl/extractor/nhl.py
joshka: I think from memory, the interface through to the cookies made it difficult to get the cookie expiry out. I've noted two TODOs in the source code to come back and fix these points.

I'm almost inclined to remove the credential cache logic altogether as it doesn't address the problem described by @Manouchehri above where the site does not allow you to make multiple requests over an hour. What are your thoughts on this given your better understanding of how youtube-dl should work?

youtube_dl/extractor/nhl.py
joshka: This is true. Can you provide some information on how the authentication flow should work including which parts should override and how things should be cached? I've tried to address an issue that seems to be problematic as mentioned above, but it looks like my attempts to do so are not in line with your expectations.

youtube_dl/extractor/nhl.py
joshka: I disagree. The file created is not an mp4 container, it's a copy of an mpeg2ts stream. They are different formats. `.ts` is recognized by mpv and vlc and likely other media players as a stream format that may not yet be completely finished. I've had problems using the `.mp4` in the past with media players that expect headers to give information on stream length. This avoids that problem.

README.md
joshka: Will remove

youtube_dl/options.py
joshka: I was suggesting that it be a generic option, perhaps it would be nice to deprecate the similar `--ap-mso` option in favor of this more descriptive option.

Happy to make this only for the nhl plugin if that's more appropriate.

joshka:
Will fix for py3

joshka:
Thanks for the review @dstftw. A lot of this seems to me a lot like my problem is not understanding your vision for how this should work. I could iterate on this for a while without success, but perhaps instead you could provide some more general comments that would explain the auth flow in youtube-dl and how this should interact?

As I see it, there's quite a few pieces:
Cookies (stored / cached / command line provided / expiry?)
Login (username / password)
Do stored cookies override username / password or the other way around?

From a simplest MVP perspective, a user should be able to pass a username and password and optional provider and retrieve a video. Anything beyond that is gravy and in my mind does not need to be implemented to get the initial part of this over the line. Would this be your viewpoint, or do you want a complete fill implementation taking into account all ways that something may break as a first cut?

youtube_dl/extractor/nhl.py
dstftw: I would strongly avoid adding per extractor specific options. Ideally, end user should not even be aware of the cache.
Currently, there is no proper mechanism for caching sensitive data and usually when one needs to cache session he just uses `--cookies` on his own.

youtube_dl/extractor/nhl.py
dstftw: First, if extractor allows using authenticated cookies it should respect `--cookies`. Next it should process credentials. Usually youtube-dl does not cache sensitive data.

youtube_dl/extractor/nhl.py
dstftw: The file created in this particular case (assuming no `--hls-use-mpegts`) is mp4. Check it if you don't believe.

youtube_dl/options.py
dstftw: `--ap-mso` should not be touched since it's a separate stand-alone mechanism that has stable unique MSO identifiers for TV providers across all extractors while this options is not.

youtube_dl/extractor/nhl.py
dstftw: I did not test it much but if it does not actually solve the problem then I don't see much point in it either.

youtube_dl/extractor/nhl.py
joshka: Ah, I use that option generally so that I can watch streams while downloading.

Manouchehri:
Even if it's not perfect, it'd be nice to finally get something semi-working merged after close to a year..

yan12125:
@Manouchehri: IMO a major problem in this patch is that it stores sensitive data without notifying users. That can be dangerous.

@joshka:

> Cookies (stored / cached / command line provided / expiry?)

Cookies are provided via the command line option ```--cookies```.

When a cookie is no longer accepted for login, Most likely NHL will return an error. Simply raising the error is OK. Detect it and prompt users to login again is better.

> Do stored cookies override username / password or the other way around?

Currently when both --cookies and --username/--password are provided, the login procedure goes on with new credentials and overrides saved cookies.

niconico.py can be served as an example. When username/password are not provided, _login() is skipped and cookies are used to access videos. ```YoutubeDL``` handles cookies implicitly so there's nothing to do for extractors.

https://github.com/rg3/youtube-dl/blob/cd670befc4c823a38a88fffbaa6c493e539dd79d/youtube_dl/extractor/niconico.py#L168-L169

When cookies are no longer valid (or no cookies are provided at all), it's detected and an error is reported:

https://github.com/rg3/youtube-dl/blob/cd670befc4c823a38a88fffbaa6c493e539dd79d/youtube_dl/extractor/niconico.py#L300-L302

joshka:
I got busy with life and dropped this way back then. Hockey is back and I've moved to a different timezone than my team, so it's very likely that I'll pick this up soon. Thanks for the reminder.

@yan12125 my problem was balancing the need for simplicity (username / password) and not hitting the lockout condition (i.e. "Too many usages in a row causes the authentication to give an error indicating too many login attempts"). It's entirely possible that I missed the fact that `--cookie` also writes to the cookie jar, not just reads. I'll definitely drop the in built cookie caching as it didn't solve the problem. Better to have something partially work for this and sometimes fail auth, than not having anything working for this.

> YoutubeDL handles cookies implicitly so there's nothing to do for extractors.

There's a cookie that is tied to the specific video that this change explicitly handles. This cannot be implicitly handled by YoutubeDL or provided by the user without significant effort per video. Part of the changes in this PR are to use the API to generate this value.

yan12125:
I might be misleading. Of course extractors need to handle the login procedure and possibly write cookies. And extractors should identify when cookies are no longer valid. YoutubeDL handles caching of cookies (in a form of text files), so there's no need to explicitly load caches in extractors.

Issue #11023, opened 2016-10-25
[Pandora.tv] extracts but HTTPError: HTTP Error 404: Not Found


TMoneyZ:
## Please follow the guide below
- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your _issue_ (like that [x])
- Use _Preview_ tab to see how your issue will actually look like

---
### Make sure you are using the _latest_ version: run `youtube-dl --version` and ensure your version is _2016.10.25_. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x ] I've **verified** and **I assure** that I'm running youtube-dl **2016.10.25**
### Before submitting an _issue_ make sure you have:
- [x ] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x ] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
### What is the purpose of your _issue_?
- [x ] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---
### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your _issue_

---
### If the purpose of this _issue_ is a _bug report_, _site support request_ or you are not completely sure provide the full verbose output as follows:

Add `-v` flag to **your command line** you run youtube-dl with, copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):

```
$youtube-dl -v 'http://jp.channel.pandora.tv/channel/video.ptv?ch_userid=monako5&prgid=54293799'
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://jp.channel.pandora.tv/channel/video.ptv?ch_userid=monako5&prgid=54293799']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.10.25
[...]

PR #11377, opened 2016-12-06
Pandoratv extractor fix

Serkora:

- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/) (well, the code that I inserted is by me)

- [x] Bug fix

---

Apparently, pandora.tv has changed the way the data is retrieved, so previous extractor is not completely usable. This pull request adds the necessary steps to get the direct media link. Specifically, this link must have a timestamp and hash code which the code added now obtains. More info can be found here #11023

### DIFF
--- a/youtube_dl/extractor/pandoratv.py
+++ b/youtube_dl/extractor/pandoratv.py
@@ -12,7 +12,6 @@ from ..utils import (
     float_or_none,
     parse_duration,
     str_to_int,
-    urlencode_postdata,
 )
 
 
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/pandoratv.py
```diff
@@ -56,6 +57,15 @@ def _real_extract(self, url):
                 r'^v(\d+)[Uu]rl$', format_id, 'height', default=None)
             if not height:
                 continue
+
+            post_data = ('prgid=%s&runtime=%s&vod_url=%s' % (video_id, info.get('runtime'), format_url)).encode()
+            data = compat_urllib_request.urlopen("http://m.pandora.tv/?c=api&m=play_url", data=post_data).read()
+            encoding = self._guess_encoding_from_content('json', data)
+            play_url = self._parse_json(data.decode(encoding, errors='ignore'), video_id)
```
dstftw: Use `self._download_json` instead.

dstftw: https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/awaan.py#L167-L172

Serkora: Done. Thanks for the help.

dstftw:

PR #11462, opened 2016-12-16
[tistory] Add new extractor

qsniyg:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Supports both direct and swf-based tistory media urls, preserving the original filenames (if applicable).

The swf part of the code decompresses the swf file if applicable (based on pyswf), then searches for a url that, afaics, is used to check if the file still exists. This url contains the components necessary to construct the final download url, which is then sent back to this extractor to download directly.

### DIFF
--- a/youtube_dl/extractor/tistory.py
+++ b/youtube_dl/extractor/tistory.py
@@ -1,0 +1,113 @@
+# coding: utf-8
+
+from .common import InfoExtractor
+from ..utils import (
+    url_basename,
+    mimetype2ext,
+    HEADRequest,
+    ExtractorError
+)
+from ..compat import (
+    compat_urllib_request,
+    compat_urlparse,
+    compat_str
+)
+
+import os.path
+import cgi
+import re
+
+
+class TistoryIE(InfoExtractor):
+    _VALID_URL = r'https?://cfile[0-9]*.uf.tistory.com/(?:media|attach|attachment)/(?P<id>[A-Za-z0-9]*)'
+
+    _TEST = {
+        'url': 'http://cfile23.uf.tistory.com/media/111ED14A4FAEBC3C23AAE1',
+        'md5': '55c32cda7b1a091d75c32aeaaea47595',
+        'info_dict': {
+            'id': '207B594C4FAEBBC118096B',
+            'title': compat_str('함친.wmv-muxed', encoding="UTF-8"),
+            'ext': 'mp4'
+        }
+    }
+
+    def unquote(self, url):
+        return compat_urlparse.unquote(url)
+
+    def get_title(self, url, response):
+        _, params = cgi.parse_header(response.info().get('Content-Disposition', ''))
+        if "filename" in params:
+            filename = params["filename"]
+        else:
+            filename = url_basename(url)
+
+        retval = os.path.splitext(self.unquote(filename))[0]
+
+        if type(retval) != compat_str:
+            retval = retval.decode('UTF-8')
+
+        return retval
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        self.to_screen('%s: Downloading headers' % (video_id))
+        req = HEADRequest(url)
+
+        head = compat_urllib_request.urlopen(req)
+        content_type = head.info().get("content-type")
+
+        ret = {
+            "id": compat_str(video_id),
+            "url": url,
+            "title": self.get_title(url, head)
+        }
+
+        if content_type == "application/x-shockwave-flash":
+            swfreq = self._request_webpage(url, video_id, "Downloading SWF")
+            data = swfreq.read()
+
+            a = data[0]
+            b = data[1]
+            c = data[2]
+
+            if isinstance(a, str):
+                a = ord(a)
+                b = ord(b)
+                c = ord(c)
+
+            rawswfdata = data[8:]
+
+            if a not in [0x43, 0x46, 0x5A] or b != 0x57 or c != 0x53:
+                raise ExtractorError("Not a SWF file")
+
+            if a == 0x46:
+                swfdata = rawswfdata
+            elif a == 0x43:
+                import zlib
+                zip = zlib.decompressobj()
+                swfdata = str(zip.decompress(rawswfdata))
+            elif a == 0x5A:
+                import pylzma
+                rawswfdata = data[11:]
+                swfdata = str(pylzma.decompress(rawswfdata))
+
+            match = re.search("(https?://[A-Za-z0-9.]*/attachment/cfile[0-9]*.uf@[A-Za-z0-9.@%]*)",
+                              swfdata)
+            if not match:
+                raise ExtractorError("Unable to find check URL")
+
+            checkurl = match.group(1)
+
+            checkmatch = re.search("(cfile[0-9]*.uf)@([A-Z0-9]*)", checkurl)
+            if not checkmatch:
+                raise ExtractorError("Unable to find real URL in check URL")
+
+            cfile = checkmatch.group(1)
+            url = checkmatch.group(2)
+
+            ret["url"] = "http://" + cfile + ".tistory.com/attach/" + url
+            return self._real_extract(ret["url"])
+        else:
+            ret["ext"] = mimetype2ext(content_type)
+            return ret
--- a/youtube_dl/extractor/tistory.py
+++ b/youtube_dl/extractor/tistory.py
@@ -20,7 +20,7 @@ import re
 
 
 class TistoryIE(InfoExtractor):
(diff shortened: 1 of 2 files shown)


dstftw:
>- [x] At least skimmed through [adding new extractor tutorial](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions]

Now read again and fix all the issues.

qsniyg:
Ah ... I might have read through it a little too quickly. Fixed afaics :)

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+            title = video_id
+
+        return {
+            "id": compat_str(video_id),
+            "url": url,
+            "title": title,
+            "ext": ext
+        }
+
+    def _ti_read_swf(self, url, video_id, head):
```
yan12125: Lots of codes in this method duplicates swfinterp.py. Re-use existing codes instead.

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+import os.path
+import cgi
+import re
+
+
+class TistoryBaseIE(InfoExtractor):
+    _TI_MEDIA_URL = r'https?://cfile[0-9]*.uf.(tistory.com|daum.net)/(?:media|attach|attachment|original)/(?P<id>[A-Za-z0-9]*)'
+
+    def _ti_unquote(self, url):
+        return compat_urlparse.unquote(url)
```
yan12125: Use compat_urllib_parse_unquote instead. Also, there's no need to create a function with only one line.

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+        cfile = checkmatch.group("server")
+        url = checkmatch.group("id")
+        ext = None
+
+        if len(checkmatch.groups()) > 3:
+            ext = checkmatch.group("ext")
+
+        return ("http://" + cfile + "." + host + "/attach/" + url, ext)
+
+    def _ti_get_video_id(self, url):
```
yan12125: This is equivalent to InfoExtractor._match_id

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+            }
+        }
+    ]
+
+    def _real_extract(self, url):
+        return self._ti_dl(url)
+
+
+class TistoryPlaylistIE(TistoryBaseIE):
+    _VALID_URL = r'.*(?P<rurl>https?://cfs.tistory.com/custom/blog/.*/skin/images/(?P<id>.*)\.xml).*'
```
yan12125: No test URLs?

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def _ti_dl(self, url, ext=None, title=None, video_id=None):
+        if not video_id:
+            video_id = self._ti_get_video_id(url)
+
+        head = None
+
+        try:
+            head = self._ti_get_headers(url, video_id)
+        except compat_urllib_error.HTTPError:
```
yan12125: Use compat_HTTPError instead

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+            pass
+        except Exception:
```
yan12125: Never ignore generic exceptions

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+    compat_str
+)
+
+import os.path
+import cgi
+import re
+
+
+class TistoryBaseIE(InfoExtractor):
+    _TI_MEDIA_URL = r'https?://cfile[0-9]*.uf.(tistory.com|daum.net)/(?:media|attach|attachment|original)/(?P<id>[A-Za-z0-9]*)'
```
yan12125: Don't capture unused groups

yan12125: Also, there's already daum.py. Are tistory.com and daum.net related?

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+    def _ti_get_title(self, url, response):
+        _, params = cgi.parse_header(response.info().get('Content-Disposition', ''))
+        if "filename" in params:
+            filename = params["filename"]
+        else:
+            filename = url_basename(url)
+
+        retval = os.path.splitext(self._ti_unquote(filename))[0]
+
+        if type(retval) != compat_str:
```
yan12125: filename should already be a unicode object. Does cgi.parse_header return bytes-like objects? If so a wrapper is necessary.

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+            retval = retval.decode('UTF-8')
+
+        return retval
+
+    def _ti_get_ext(self, mime):
+        ext = mimetype2ext(mime)
+        if ext == "x-shockwave-flash":
```
yan12125: Just modify mimetype2ext rather than introducing hacks in individual extractors

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+            self._TI_MEDIA_URL_RE = re.compile(self._TI_MEDIA_URL)
+        m = self._TI_MEDIA_URL_RE.match(url)
+        assert m
+        return m.group('id')
+
+    def _ti_get_headers(self, url, video_id):
+        self.to_screen('%s: Downloading headers' % (video_id))
+        req = HEADRequest(url)
+
+        return compat_urllib_request.urlopen(req)
```
yan12125: Use _request_webpage instead of plain urlopen. The former includes fixes and custom features missing from urllib.request

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def _ti_detect_swf(self, head):
+        content_type = head.info().get("content-type")
+        content_length = int(head.info().get("content-length"))
+
+        if content_type == "application/x-shockwave-flash" and content_length < 200000:
```
yan12125: Are flash files always < 200K?

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+class TistoryPlaylistIE(TistoryBaseIE):
+    _VALID_URL = r'.*(?P<rurl>https?://cfs.tistory.com/custom/blog/.*/skin/images/(?P<id>.*)\.xml).*'
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        rurl = self._VALID_URL_RE.match(url).group("rurl")
+
+        return self._ti_dl(rurl, video_id=video_id)
```
yan12125: Maybe TistoryIE and TistoryPlaylistIE can be merged into a single extractor?

youtube_dl/extractor/tistory.py
```diff
@@ -0,0 +1,228 @@
...
+        return m.group('id')
+
+    def _ti_get_headers(self, url, video_id):
+        self.to_screen('%s: Downloading headers' % (video_id))
+        req = HEADRequest(url)
+
+        return compat_urllib_request.urlopen(req)
+
+    def _ti_detect_swf(self, head):
+        content_type = head.info().get("content-type")
```
yan12125: There are lots of Content-Type calls. Please merge them together

qsniyg:
I'm sorry for getting back to this so late, I had completely forgotten about this pull request. Thank you very much for spending the time to review it!

I found out later that the code here only applies to one specific site, and that tistory most likely allows any type of .swf (or other) to be uploaded. In other words, the code I wrote actually is specific to one player, and has nothing to do with tistory (other than the fact that that player happened to be used on a website that was on tistory).

I'm not too sure what would be the best course of action here. From what I understand of youtube-dl, it seems to go by a website-by-website basis. Should it be kept as a generic "tistory" extractor? Or would it be better to somehow autodetect the usage of that player in websites?

Second question: In TistoryPlaylistIE, I later discovered that it's not used as a playlist, but rather as parts of the same video, that can be concatenated via ffmpeg. What would be the correct way to invoke ffmpeg in the youtube-dl codebase?

yan12125:
For the first question: did you find other websites using a similar player? If not it's OK to keep the current approach - a single extractor.

For the second question - unfortunately there's no generic way to solve segmented videos. Currently we use '_type': 'multi_video' and download segments as-is. The user needs to merge segments manually.

PR #11482, opened 2016-12-19
Added a "filename_hook" parameter to Python interface

CRImier:

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

Hello! I need the to sanitize the filenames myself since the existing functions do not satisfy me. Thus, I've added a way to add a hook that receives the filename generated by YoutubeDL and returns a filename that'll be used further down the line.

**Reasons:** with "restricted filenames", filenames are sanitized too much to the point of making the filename very ugly, and without "restricted filenames" I get characters that aren't compatible with the software I use. Also I plan to further integrate YoutubeDL with my music collection organizing scripts and such a feature would help. I already use it in my local version but I figure others could benefit from this as well.

**Sidenotes:** The hook is very likely to get called 2 or more times (since  ``process_filename`` is called quite often), but I can't see this causing any problems if we assume the processing function to always return same x for same y (which is the case for any reasonable sanitization function).

Example usage:

```
def filename_hook(filename):
    #An example hook - manual editing
    return raw_input('Edit this: "{}" : '.format(filename)).strip()

opts = {
"filename_hook":filename_hook,
"match_filter":print_info,
"outtmpl":"%(title)s",
"postprocessors":[{"key":'FFmpegExtractAudio', "preferredcodec":"mp3"}]
}
```

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -151,6 +151,10 @@ class YoutubeDL(object):
     simulate:          Do not download the video files.
     format:            Video format code. See options.py for more information.
     outtmpl:           Template for output names.
+    filename_hook:     A function that gets called before creating a filename, 
+                       allowing to change it before the filename is created.
+                       It should accept filename as first positional argument
+                       and return the desired filename (no extensions added)
     restrictfilenames: Do not allow "&" and spaces in file names
     ignoreerrors:      Do not stop on download errors.
     force_generic_extractor: Force downloader to use the generic extractor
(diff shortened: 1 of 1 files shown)


dstftw:
Use `outtmpl` instead.

CRImier:
No! You didn't even understand the purpose of this. Check this out:
```
def filename_hook(filename):
    filename = filename.replace(u'\u3010', u'[').replace(u'\u3011', u']')
    return filename
```
Can you show me a way to make this with ``outtmpl``?

dstftw:
Easy:
```
def build_filename(info_dict):
    return 'whatever_filename_based_on_info_dict.%(ext)s'

ydl = youtube_dl.YoutubeDL()
info = ydl.extract_info('BaW_jenozKc', download=False)
for entry_info in info.get('entries', [info]):
    ydl.params['outtmpl'] = build_filename(entry_info)
    ydl.process_ie_result(entry_info)
```

Or just derive from `YoutubeDL` and override `prepare_filename`.

CRImier:
Thank you very much, the "derive from YoutubeDL and override" way worked!

PR #11483, opened 2016-12-19
Added a "filename_hook" parameter to Python interface

CRImier:

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

Hello! I need the to sanitize the filenames myself since the existing functions do not satisfy me. Thus, I've added a way to add a hook that receives the filename generated by YoutubeDL and returns a filename that'll be used further down the line.

**Reasons:** with "restricted filenames", filenames are sanitized too much to the point of making the filename very ugly, and without "restricted filenames" I get characters that aren't compatible with the software I use. Also I plan to further integrate YoutubeDL with my music collection organizing scripts and such a feature would help. I already use it in my local version but I figure others could benefit from this as well. 

More importantly, there are currently no ways to implement custom filtering of filenames. For example, when downloading videos straight from Youtube for the purpose of storing them, you'll sometimes want to use the title as your filename. However, the title can contain redundant or unnecessary elements, like the channel name or genre. Currently, I have to write scripts to do the postprocessing for my purposes, and that's very annoying. 

The ``outtmpl`` template can't be used to, for example, replace characters in the generated filename or shorten it (though they can be used together, as my example shows), or applying a regex to strip out useless parts of the filename. If there is a way to hook a Python function so that it can process and change the filename used, I'd be happy to hear about it - the code didn't show any.

**Sidenotes:** The hook is very likely to get called 2 or more times (since  ``process_filename`` is called quite often), but I can't see this causing any problems if we assume the processing function to always return same x for same y (which is the case for any reasonable sanitization function).

Example usage:

```
def filename_hook(filename):
    filename = filename.replace(u'\u3010', u'[').replace(u'\u3011', u']')
    return filename

opts = {
"filename_hook":filename_hook,
"match_filter":print_info,
"outtmpl":"%(title)s",
"postprocessors":[{"key":'FFmpegExtractAudio', "preferredcodec":"mp3"}]
}
```

Sorry for reopening this, but it was closed when the reviewer, apparently, did not fully understand the purpose of this feature, and from what I found on Google the pull requests can't be reopened after closing. I also improved my pull request by further describing the usefullness of this function.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -151,6 +151,10 @@ class YoutubeDL(object):
     simulate:          Do not download the video files.
     format:            Video format code. See options.py for more information.
     outtmpl:           Template for output names.
+    filename_hook:     A function that gets called before creating a filename, 
+                       allowing to change it before the filename is created.
+                       It should accept filename as first positional argument
+                       and return the desired filename (no extensions added)
     restrictfilenames: Do not allow "&" and spaces in file names
     ignoreerrors:      Do not stop on download errors.
     force_generic_extractor: Force downloader to use the generic extractor
@@ -612,7 +616,9 @@ class YoutubeDL(object):
             # to workaround encoding issues with subprocess on python2 @ Windows
             if sys.version_info < (3, 0) and sys.platform == 'win32':
                 filename = encodeFilename(filename, True).decode(preferredencoding())
-            return sanitize_path(filename)
+            filename = sanitize_path(filename)
+            filename = self.params.get('filename_hook', lambda x: x)(filename)
+            return compat_str(filename)
         except ValueError as err:
             self.report_error('Error in output template: ' + str(err) + ' (encoding: ' + repr(preferredencoding()) + ')')
             return None


yan12125:
> I get characters that aren't compatible with the software I use

It's surprising to hear this in 2016 that an application can't handle unicode characters. Apparently the problem is in your software, not youtube-dl

> Currently, I have to write scripts to do the postprocessing for my purposes, and that's very annoying.

Hmm, isn't writing a hook function the same effort as writing postprocessing scripts?

CRImier:
Well, I've decided to disable restricted filenames in my non-modified setup, download a video with a Russian filename and see what fails:

```
Noize MC – Питерские крыши   #filename                                                                 
WARNING: Requested formats are incompatible for merge and will be merged into mkv.            
ERROR: unable to open for writing: [Errno 22] invalid mode ('wb') or filename: 'Noize MC \xe2\
x80\x93 \xd0\x9f\xd0\xb8\xd1\x82\xd0\xb5\xd1\x80\xd1\x81\xd0\xba\xd0\xb8\xd0\xb5 \xd0\xba\xd1\
x80\xd1\x8b\xd1\x88\xd0\xb8.f137.part'                                                        
Exception in thread Thread-1:                                                                 
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner                      
    self.run()                                                                                
  File "/usr/lib/python2.7/threading.py", line 763, in run                                    
    self.__target(*self.__args, **self.__kwargs)                                              
  File "ssyoutube.py", line 53, in dl_thread                                                  
    download(link)                                                                            
  File "ssyoutube.py", line 59, in download                                                   
    ydl.download([link])                                                                      
  File "/usr/local/lib/python2.7/dist-packages/youtube_dl-2016.12.18-py2.7.egg/youtube_dl/Yout
ubeDL.py", line 1797, in download                                                             
  File "/usr/local/lib/python2.7/dist-packages/youtube_dl-2016.12.18-py2.7.egg/youtube_dl/Yout
ubeDL.py", line 711, in extract_info                                                          
...                                                         
  File "/usr/local/lib/python2.7/dist-packages/youtube_dl-2016.12.18-py2.7.egg/youtube_dl/Yout
ubeDL.py", line 542, in trouble                                                               
    raise DownloadError(message, exc_info)                                                    
DownloadError: ERROR: unable to open for writing: [Errno 22] invalid mode ('wb') or filename: 
'Noize MC \xe2\x80\x93 \xd0\x9f\xd0\xb8\xd1\x82\xd0\xb5\xd1\x80\xd1\x81\xd0\xba\xd0\xb8\xd0\xb
5 \xd0\xba\xd1\x80\xd1\x8b\xd1\x88\xd0\xb8.f137.part'                                         
```

I don't know what exactly is failing and why. Could be the FS on source drive (FAT32, not very likely to change that since I need to work with it both from Windows and Linux). Could be that I need to upgrade to Python 3, as usual. If you're interested in the problem I'm getting, I can help you debug this, but this is definitely not the point.

This is not the only problem this hook can solve. Say, you download videos in Russian - you can "transliterate" filenames (map Russian characters to their closest phonetical analog in English) to avoid having filenames in Russian all of a sudden, which trips up media players (I'm not too keen to change players/software if the experience is great otherwise) and virtual terminals (you need to set up encodings properly before you can see the filename on the terminal).

Effort writing postprocessing scripts is not effort well-spent. An issue like that is better solved when the file is created, not when "hang on, let me launch my script". Especially when you can, for example, use ``info_dict`` (available only at download time, unless you want to use even dirtier hacks) to generate an even better filename. I've got two or three Python scripts doing this "downloading, processing, recording into DB, reformatting and synchronization" thing - there's no sense to have more, honestly, not when a solution is a 3-line patch that doesn't affect anybody not using it.

So, the hook purposes:
1) Transliterating filenames
2) Filtering filenames
3) Generating better filenames

I think it's likely that other people will need something like this.

yan12125:
OK I see your point.

> Especially when you can, for example, use info_dict (available only at download time, unless you want to use even dirtier hacks) to generate an even better filename.

This sounds a better approach. Allowing more flexible filename generation is better than fixing bad filenames afterwards. Could you like to have a try?

CRImier:
Hello! I got some useful advice in my previous PR: https://github.com/rg3/youtube-dl/pull/11482#issuecomment-267991201 . To get to my goal, I can just derive from YoutubeDL class and override the prepare_filename function. Therefore, this functionality can be achieved in the current version, so this PR is unnecessary =) Thank you everybody for the help, especially @dstftw .

@yan12125 - how do you imagine that "more flexible filename generation"? I got my solution working for me, but it's possible I could implement what you're suggesting in my free time. 

The resulting (tested) implementation can be seen here: https://github.com/CRImier/ssyoutube/commit/4e549193348e86239b5f48feb4391dd69a274614

yan12125:
> To get to my goal, I can just derive from YoutubeDL class and override the prepare_filename function.

I was proposing something like ```'outtmpl_callback': my_prepare_filename```, but your solution is better :)

PR #11520, opened 2016-12-23
[StreamMe] Add new extractor

shizeeg:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

It's an extractor for a new streaming platform https://www.stream.me/
the extractor handles vod, live-streams and can download #archives section of any channel.

multi-streams are not tested at all and probably don't work. (not critical)

### DIFF
--- a/youtube_dl/extractor/streamme.py
+++ b/youtube_dl/extractor/streamme.py
@@ -16,7 +16,7 @@ class StreamMeIE(InfoExtractor):
     _API_CHANNEL = 'https://www.stream.me/api-user/v1/%s/channel'
     _API_ARCHIVE = 'https://www.stream.me/api-vod/v1/%s/archives'
     _VALID_URL_BASE = r'https?://(video-cdn|www).stream.me'
-    _VALID_URL = r'%s/archive\/(?P<channel_id>[^\#/]+)/[^\/]+/(?P<id>[^/]+)' % _VALID_URL_BASE
+    _VALID_URL = r'%s/archive/(?P<channel_id>[^\#/]+)/[^/]+/(?P<id>[^/]+)' % _VALID_URL_BASE
     _TEST = {
         'url': 'https://www.stream.me/archive/kombatcup/kombat-cup-week-8-sunday-open/pDlXAj6mYb',
         'md5': 'b32af6fad972d0bcf5854a416b5b3b01',
@@ -107,7 +107,7 @@ class StreamMeIE(InfoExtractor):
 
 class StreamMeLiveIE(StreamMeIE):
     IE_NAME = 'StreamIE:live'
-    _VALID_URL = r'%s\/(?P<id>[^\#\/]+$)' % StreamMeIE._VALID_URL_BASE
+    _VALID_URL = r'%s/(?P<id>[^\#/]+$)' % StreamMeIE._VALID_URL_BASE
     _TEST = {
         'url': 'https://www.stream.me/kombatcup',
         'info_dict': {
@@ -121,6 +121,7 @@ class StreamMeLiveIE(StreamMeIE):
             'dislike_count': int,
             'is_live': True,
         },
+        'skip': 'kombatcup is offline',
         'params': {
             # m3u8 download
             'skip_download': True,
--- a/youtube_dl/extractor/streamme.py
+++ b/youtube_dl/extractor/streamme.py
@@ -94,6 +94,16 @@ class StreamMeIE(InfoExtractor):
                     # I don't know all the possible protocols yet.
                     # 'protocol': 'm3u8_native' if fmt_tag == 'mp4-hls' else 'http'
                 })
+            if d.get('origin') is not None:
+                fmt_tag = d.get('origin').get('location').split(':')[0]
+                formats.append({
+                    'url': d.get('origin').get('location'),
+                    'acodec': d.get('origin').get('audioCodec'),
+                    'vcodec': d.get('origin').get('videoCodec'),
+                    'format_id': 'Source-' + fmt_tag,
+                    'ext': 'flv' if fmt_tag == 'rtmp' else 'mp4',
+                    'source_preference': 1,
+                })
         return formats
 
 
--- a/youtube_dl/extractor/streamme.py
+++ b/youtube_dl/extractor/streamme.py
@@ -1,7 +1,6 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
-import json
 import re
 
 from .common import InfoExtractor
@@ -47,7 +46,7 @@ class StreamMeIE(InfoExtractor):
         manifest_json = self._download_json(vod_info['_links']['manifest']['href'],
                                             video_id, note='Downloading video manifest')
 
-        formats = self._extract_formats(manifest_json.get('formats'))
+        formats = self._extract_formats(manifest_json['formats'])
         self._sort_formats(formats, 'vbr')
         info = self._extract_info(vod_info)
         info['formats'] = formats
@@ -78,7 +77,7 @@ class StreamMeIE(InfoExtractor):
             # skip websocket and mjpeg we can't handle them anyway
             if fmt_tag in ('mjpeg-lodef', 'mp4-ws',):
                 continue
-            for fmt_info in d.get('encodings'):
+            for fmt_info in d['encodings']:
                 formats.append({
                     'url': fmt_info.get('location'),
                     'width': int_or_none(fmt_info.get('videoWidth')),
@@ -93,7 +92,7 @@ class StreamMeIE(InfoExtractor):
                     # 'protocol': 'm3u8_native' if fmt_tag == 'mp4-hls' else 'http'
                 })
             if d.get('origin') is not None:
-                fmt_tag = d.get('origin').get('location').split(':')[0]
+                fmt_tag = d['origin']['location'].split(':')[0]
                 formats.append({
                     'url': d.get('origin').get('location'),
                     'acodec': d.get('origin').get('audioCodec'),
@@ -132,10 +131,10 @@ class StreamMeLiveIE(StreamMeIE):
         channel_id = self._match_id(url)
         apiurl = StreamMeIE._API_CHANNEL % channel_id
 
-        data = json.loads(self._download_webpage(apiurl, channel_id))
+        data = self._download_json(apiurl, channel_id)
         stream_info = []
         # search for a live stream...
-        for stream in data.get('_embedded').get('streams'):
+        for stream in data['_embedded']['streams']:
             stream_info = stream
             break   # TODO: add to a list (multi-streams?)
 
@@ -143,9 +142,9 @@ class StreamMeLiveIE(StreamMeIE):
             raise ExtractorError('%s is offline' % channel_id, expected=True)
 
         manifest_json = self._download_json(stream_info['_links']['manifest']['href'],
-                                            channel_id, 'Download video manifest')
+                                            channel_id, 'Downloading video manifest')
 
-        formats = self._extract_formats(manifest_json.get('formats'))
+        formats = self._extract_formats(manifest_json['formats'])
         self._sort_formats(formats, 'vbr')
         info = self._extract_info(stream_info)
         info['formats'] = formats
@@ -177,8 +176,9 @@ class StreamMeArchiveIE(StreamMeIE):
         playlist = []
 
         for vod in data['_embedded']['vod']:
-            manifest_json = self._download_json(vod['_links']['manifest']['href'], vod.get('urlId'))
-            formats = self._extract_formats(manifest_json.get('formats'))
+            manifest_json = self._download_json(vod['_links']['manifest']['href'],
+                                                vod['urlId'], note='Downloading video manifest')
+            formats = self._extract_formats(manifest_json['formats'])
             self._sort_formats(formats, 'vbr')
             info = self._extract_info(vod)
             info['formats'] = formats
--- a/youtube_dl/extractor/streamme.py
+++ b/youtube_dl/extractor/streamme.py
@@ -99,7 +99,25 @@ class StreamMeIE(InfoExtractor):
 
 class StreamMeLiveIE(StreamMeIE):
     IE_NAME = 'StreamIE:live'
-    _VALID_URL = r'%s/(?P<id>[^\#]+$)' % StreamMeIE._VALID_URL_BASE
+    _VALID_URL = r'%s\/(?P<id>[^\#\/]+$)' % StreamMeIE._VALID_URL_BASE
+    _TEST = {
+        'url': 'https://www.stream.me/kombatcup',
+        'info_dict': {
+            'id': 'live',  # see: StreamMeIE._extract_info()
+            'ext': 'mp4',
+            'title': 'KombatCup\'s Live Stream',
+            'age_limit': 13,
+            'uploader_id': 'kombatcup',
+            'uploader': 'KombatCup',
+            'like_count': int,
+            'dislike_count': int,
+            'is_live': True,
+        },
+        'params': {
+            # m3u8 download
+            'skip_download': True,
+        },
+    }
 
     def _real_extract(self, url):
         channel_id = self._match_id(url)
--- a/youtube_dl/extractor/streamme.py
+++ b/youtube_dl/extractor/streamme.py
@@ -35,13 +35,24 @@ class StreamMeIE(InfoExtractor):
         video_id = self._match_id(url)
         data = self._download_json(self._API_VOD % video_id, video_id)
 
-        if len(data['_embedded']['streams']) > 0:
+        if not data and data.get('_embedded'):
+            raise ExtractorError(
+                '{0} returns no data or data is incorrect'.format(video_id), expected=True)
+
+        if len(data['_embedded'].get('streams')) > 0:
             vod_info = data['_embedded']['streams'][0]
         else:
-            raise ExtractorError('Video "%s" not found' % video_id, expected=True)
+            raise ExtractorError('Video "{0}" not found'.format(video_id), expected=True)
 
-        manifest_json = self._download_json(vod_info['_links']['manifest']['href'],
-                                            video_id, note='Downloading video manifest')
+        if vod_info.get('_links') and vod_info['_links'].get('manifest'):
+                if vod_info['_links']['manifest'].get('href'):
+                    manifest_json = self._download_json(
+                        vod_info['_links']['manifest'].get('href'),
+                        video_id, note='Downloading video manifest')
+        else:
+            raise ExtractorError('JSON has unexpected format', expected=True)
+        if not manifest_json or not manifest_json.get('formats'):
+            raise ExtractorError('Video manifest has no formats information', expected=True)
 
         formats = self._extract_formats(manifest_json['formats'])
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,193 @@
...
+        },
+        'params': {
+            # m3u8 download
+            'skip_download': True,
+        },
+    }
+
+    def _real_extract(self, url):
+        channel_id = self._match_id(url)
+        apiurl = StreamMeIE._API_CHANNEL.replace('<channel_id>', channel_id)
```
dstftw: Use regular string format syntax instead.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,193 @@
...
+    ExtractorError,
+)
+
+
+class StreamMeIE(InfoExtractor):
+    IE_NAME = 'StreamMe:video'
+    _API_CHANNEL = 'https://www.stream.me/api-user/v1/<channel_id>/channel'
+    _API_ARCHIVE = 'https://www.stream.me/api-vod/v1/<channel_id>/archives'
+    _VALID_URL_BASE = r'https?://(video-cdn|www).stream.me'
+    _VALID_URL = r'%s\/archive\/(?P<channel_id>[^\#\/]+)\/[^\/]+\/(?P<id>[^\/]+)' % _VALID_URL_BASE
```
dstftw: No need to escape `/`.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,193 @@
...
+            'age_limit': 13,
+        }
+    }
+
+    def _real_extract(self, url):
+        m = re.match(self._VALID_URL, url)
+        video_id = self._match_id(url)
+        apiurl = self._API_ARCHIVE.replace('<channel_id>', m.group('channel_id'))
+
+        data = json.loads(self._download_webpage(apiurl, video_id))
```
dstftw: `_download_json`.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,193 @@
...
+
+        for vod in data.get('_embedded').get('vod'):
```
dstftw: There is no point in `get()` here. You will fail anyway if there is no corresponding keys in dict.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,193 @@
...
+            vod_info = []
+            if vod.get('urlId') == video_id:
+                vod_info = vod
+                break
+
+        manifest_json = self._download_json(vod_info
+                                            .get('_links')
+                                            .get('manifest')
+                                            .get('href'), video_id)
```
dstftw: Same here.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,193 @@
...
+        return info
+
+    def _extract_info(self, info):
+        return {
+            'id': info.get('urlId') or 'live',
+            # 'formats': self.formats,
+            'title': info.get('title'),
+            'age_limit': int_or_none(info.get('ageRating')),
+            'description': info.get('description') or None,
+            'dislike_count': int_or_none(info.get('stats').get('raw').get('dislikes')),
```
dstftw: Here and so on.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,193 @@
...
+        self._sort_formats(formats, 'vbr')
+        info = self._extract_info(vod_info)
+        info['formats'] = formats
+        return info
+
+    def _extract_info(self, info):
+        return {
+            'id': info.get('urlId') or 'live',
+            # 'formats': self.formats,
+            'title': info.get('title'),
```
dstftw: Title is mandatory.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,193 @@
...
+
+        formats = self._extract_formats(manifest_json.get('formats'))
+        self._sort_formats(formats, 'vbr')
+        info = self._extract_info(vod_info)
+        info['formats'] = formats
+        return info
+
+    def _extract_info(self, info):
+        return {
+            'id': info.get('urlId') or 'live',
```
dstftw: `'live'` is not an id.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,193 @@
...
+            # 'formats': self.formats,
+            'title': info.get('title'),
+            'age_limit': int_or_none(info.get('ageRating')),
+            'description': info.get('description') or None,
```
dstftw: There is no point in `or None` since `None` is already default.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,193 @@
...
+    def _extract_formats(self, fmts):
+        formats = []
+        for fmt_tag, d in fmts.items():
+            # skip websocket and mjpeg we can't handle them anyway
+            if fmt_tag in ('mjpeg-lodef', 'mp4-ws',):
+                continue
+            for fmt_info in d.get('encodings'):
+                formats.append({
+                    'url': fmt_info.get('location'),
+                    'width': fmt_info.get('videoWidth'),
```
dstftw: `int_or_none` and `float_or_none` for all numeric fields.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,193 @@
...
+        'params': {
+            'skip_download': True,
+        }
+    }
+
+    def _real_extract(self, url):
+        channel_id = self._match_id(url).split('#')[0]
+        apiurl = StreamMeIE._API_ARCHIVE.replace('<channel_id>', channel_id)
+        # TODO: implement paginated downloading
+        data = json.loads(self._download_webpage(apiurl + '?limit=%d&offset=0' % self._PLAYLIST_LIMIT, channel_id))
```
dstftw: `_download_json`. `query` parameter for query.

youtube_dl/extractor/streamme.py
shizeeg: it seems like StreamMe doesn't provide any id for live streams. What should I put there instead?

shizeeg: well, it seems like I have something like 'publicId' in JSON, it looks like an UUID: publicId: '786df926-37f7-49ec-b3bc-bee00e1f86a5' can I use it as a live stream id?

shizeeg:
@dstftw thank you for your review & suggestions. Here the fixes you requested.

shizeeg:
phew... I think I'm done fixing all this mess. Should I squash it into a single commit & reopen PR?

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,187 @@
...
+        return info
+
+    def _extract_info(self, info):
+        return {
+            'id': info.get('urlId') or info.get('publicId'),
+            # 'formats': self.formats,
+            'title': info.get('title') or 'Untitled Broadcast',
+            'age_limit': int_or_none(info.get('ageRating')),
+            'description': info.get('description'),
+            'dislike_count': int_or_none(info.get('stats').get('raw').get('dislikes')),
```
dstftw: Not fixed.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,187 @@
...
+            'display_id': info.get('titleSlug'),
+            'duration': int_or_none(info.get('duration')),
+            'like_count': int_or_none(info.get('stats').get('raw').get('likes')),
```
dstftw: Not fixed.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,187 @@
...
+            'thumbnail': info.get('_links').get('thumbnail').get('href'),
```
dstftw: Not fixed.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,187 @@
...
+            'timestamp': info.get('whenCreated'),
+            'uploader': info.get('username'),
+            'uploader_id': info.get('userSlug'),
+            'view_count': int_or_none(info.get('stats').get('raw').get('views')),
```
dstftw: Not fixed.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,187 @@
...
+            # 'formats': self.formats,
+            'title': info.get('title') or 'Untitled Broadcast',
+            'age_limit': int_or_none(info.get('ageRating')),
+            'description': info.get('description'),
+            'dislike_count': int_or_none(info.get('stats').get('raw').get('dislikes')),
+            'display_id': info.get('titleSlug'),
+            'duration': int_or_none(info.get('duration')),
+            'like_count': int_or_none(info.get('stats').get('raw').get('likes')),
+            'thumbnail': info.get('_links').get('thumbnail').get('href'),
+            'timestamp': info.get('whenCreated'),
```
dstftw: `int_or_none`.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,187 @@
...
+                    'abr': int_or_none(fmt_info.get('audioKbps')),
+                    'acodec': d.get('audioCodec'),
+                    'vcodec': d.get('videoCodec'),
+                    'format_id': "%s%sp" % (fmt_tag, fmt_info.get('videoHeight')),
+                    'ext': 'flv' if fmt_tag.split('-')[1] == 'rtmp' else 'mp4',
+                    # I don't know all the possible protocols yet.
+                    # 'protocol': 'm3u8_native' if fmt_tag == 'mp4-hls' else 'http'
+                })
+            if d.get('origin') is not None:
+                fmt_tag = d['origin']['location'].split(':')[0]
```
dstftw: Will break if no `location`.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,187 @@
...
+        formats = self._extract_formats(manifest_json['formats'])
+        self._sort_formats(formats, 'vbr')
+        info = self._extract_info(stream_info)
+        info['formats'] = formats
+        return info
+
+
+class StreamMeArchiveIE(StreamMeIE):
+    IE_NAME = 'StreamMe:archives'
+    _VALID_URL = r'%s/(?P<id>[^\#]+)(\#archive)$' % StreamMeIE._VALID_URL_BASE
```
dstftw: No need to escape `#`. No need to capture groups you don't use.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,187 @@
...
+    int_or_none,
+    ExtractorError,
+)
+
+
+class StreamMeIE(InfoExtractor):
+    IE_NAME = 'StreamMe:video'
+    _API_CHANNEL = 'https://www.stream.me/api-user/v1/%s/channel'
+    _API_ARCHIVE = 'https://www.stream.me/api-vod/v1/%s/archives'
+    _VALID_URL_BASE = r'https?://(video-cdn|www).stream.me'
```
dstftw: Don't capture unused groups.

youtube_dl/extractor/streamme.py
```diff
@@ -0,0 +1,187 @@
...
+
+        formats = self._extract_formats(manifest_json['formats'])
+        self._sort_formats(formats, 'vbr')
+        info = self._extract_info(vod_info)
+        info['formats'] = formats
+        return info
+
+    def _extract_info(self, info):
+        return {
+            'id': info.get('urlId') or info.get('publicId'),
```
dstftw: `id` by no means should be `None`.

shizeeg: it can't be None. publicId is always presented. (It's a unique broadcast id). Every VOD has its id, but live streams hasn't so I've decided to use publicId in that case.

youtube_dl/extractor/streamme.py
shizeeg: why? twitch.tv extractor has the exact same code.

youtube_dl/extractor/streamme.py
shizeeg: what exactly?

youtube_dl/extractor/streamme.py
shizeeg: what's wrong with this? I don't get it, I'm sorry.

youtube_dl/extractor/streamme.py
shizeeg: I'll wrap 'timestamp' with int_or_none() ok, but 'uploader_id' and 'uploader' are strings so what I should to with them exactly?

youtube_dl/extractor/streamme.py
shizeeg: okay...

youtube_dl/extractor/streamme.py
dstftw: Then don't use `get`.

youtube_dl/extractor/streamme.py
dstftw: No, it doesn't.

shizeeg:
I made some fixes for stuff you've mentioned & rewrite `StreamMe:video` extractor a bit. (I've found a way to request VOD by a `video_id` directly, so no need to traverse the whole VODs list there). Now I'm thinking on how to unify `livestream`, `vod` and `archive` extractors.

P.S.: Some of your complaints are still unclear to me, so I don't want to waste your time anymore & you may close this PR if you wish.

youtube_dl/extractor/streamme.py
shizeeg: ah, okay.. but in every other extractor I looked for examples 'id' requested by .get()

youtube_dl/extractor/streamme.py
shizeeg: there are always `stats` and `raw`. It appears on a stream webpage, so... it's required to be here.

youtube_dl/extractor/streamme.py
shizeeg: I have no idea what to put instead of 'title' here. So I looked up into https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/twitch.py#L156

see? So I don't understand why that code is okay and mine is not.

youtube_dl/extractor/streamme.py
shizeeg: so you don't like the situation then NoneType has not .get() bla-bla error? Well... I got you now, but I'm too lazy to write guards here.

shizeeg: but I did, regardless.

shizeeg:
@dstftw finally find a way to ensure nothing will brake because of any JSON change, plus now 'id' & 'title' are always there no matter what.

shizeeg:
strange behavior has observed recently with Python 3.6.0
(with Python 2.7.13 continue to work just fine)

To fix this I need to do a research which I don't have any intention to make ATM. So I close this PR.

```log 
[debug] System config: []
[debug] User config: ['--netrc']
[debug] Command-line args: ['https://www.stream.me/archive/kombatcup/kombatcups-archive/9nq7e4xQY7', '--verbose']
[debug] youtube-dl version 2016.12.22
[debug] Git HEAD: 78a461053
[debug] Python version 3.6.0 - Linux-4.8.13-1-ARCH-x86_64-with-arch
[debug] exe versions: ffmpeg 3.2.2, ffprobe 3.2.2, rtmpdump 2.4
[debug] Proxy map: {}
[StreamMe:video] 9nq7e4xQY7: Downloading JSON metadata
[StreamMe:video] 9nq7e4xQY7: Downloading video manifest
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/shizeeg/archbox/workspace/youtube-dl/youtube_dl/__main__.py", line 19, in <module>
    youtube_dl.main()
  File "/home/shizeeg/archbox/workspace/youtube-dl/youtube_dl/__init__.py", line 444, in main
    _real_main(argv)
  File "/home/shizeeg/archbox/workspace/youtube-dl/youtube_dl/__init__.py", line 434, in _real_main
  File "/home/shizeeg/archbox/workspace/youtube-dl/youtube_dl/YoutubeDL.py", line 1791, in download
  File "/home/shizeeg/archbox/workspace/youtube-dl/youtube_dl/YoutubeDL.py", line 705, in extract_info
  File "/home/shizeeg/archbox/workspace/youtube-dl/youtube_dl/YoutubeDL.py", line 751, in process_ie_result
    return self.process_video_result(ie_result, download=download)
  File "/home/shizeeg/archbox/workspace/youtube-dl/youtube_dl/YoutubeDL.py", line 1435, in process_video_result
    self.process_info(new_info)
  File "/home/shizeeg/archbox/workspace/youtube-dl/youtube_dl/YoutubeDL.py", line 1506, in process_info
    reason = self._match_entry(info_dict, incomplete=False)
  File "/home/shizeeg/archbox/workspace/youtube-dl/youtube_dl/YoutubeDL.py", line 638, in _match_entry
    if date not in dateRange:
  File "/home/shizeeg/archbox/workspace/youtube-dl/youtube_dl/utils.py", line 1240, in __contains__
    date = date_from_str(date)
  File "/home/shizeeg/archbox/workspace/youtube-dl/youtube_dl/utils.py", line 1203, in date_from_str
    return datetime.datetime.strptime(date_str, '%Y%m%d').date()
  File "/usr/lib/python3.6/_strptime.py", line 565, in _strptime_datetime
    tt, fraction = _strptime(data_string, format)
  File "/usr/lib/python3.6/_strptime.py", line 365, in _strptime
    data_string[found.end():])
ValueError: unconverted data remains: 15
```

shizeeg:
fixed and reworked a bit.

CeruleanSky:
@dstftw Since it has been 22 months, is this one dead, or perhaps maybe @shizeeg could give it one more try and make it suitable?

If not, I added my own stream.me to my local fork, so while I am sure there are issues(lacking vigorous use of .get() ), it is smaller (1/3rd the lines) so in theory less to fix, unfortunately it just targets archives and not the live streams, and it differs in that it doesn't depend on the stream.me api. [[streamme]](https://github.com/CeruleanSky/youtube-dl/commit/142a49e725e4b151f7208da159298bc3dde91d7c). Maybe it could be made suitable, but the above seems to be much better.

There are at least a couple issues waiting for some type of stream.me support to happen #17673 , #18226 , and stream.me has recently begun to rise up sharply in the web traffic indexes so more requests will likely happen soon.

CeruleanSky:
Perhaps [my method](https://github.com/CeruleanSky/youtube-dl/commit/142a49e725e4b151f7208da159298bc3dde91d7c#diff-dd23111e7fa93f0cc39800ffcb6846c9R26) that bypasses the api by pulling the context json from the webpage and then parsing that for the manifest is better then in this case.  I have no immediate plans of adding livestreams though, and [streamlink uses the api](https://github.com/streamlink/streamlink/blob/fb6a00c86bbf752b578ea317cdede242a2b4e836/src/streamlink/plugins/streamme.py) just fine for those.

PR #11565, opened 2016-12-31
[infoq] Add audio only format if available

wader:
Refactor cookie code into a function.
Renamed formats to http_video, http_audio, rtmp_video
Renamed extract functions to video instead of videos as they return
one or no video.

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Add audio only format for InfoQ extractor if available

### DIFF
--- a/youtube_dl/extractor/infoq.py
+++ b/youtube_dl/extractor/infoq.py
@@ -4,7 +4,10 @@ from __future__ import unicode_literals
 
 import base64
 
-from ..compat import compat_urllib_parse_unquote
+from ..compat import (
+    compat_urllib_parse_unquote,
+    compat_urlparse,
+)
 from ..utils import determine_ext
 from .bokecc import BokeCCBaseIE
 
@@ -35,7 +38,7 @@ class InfoQIE(BokeCCBaseIE):
         },
     }]
 
-    def _extract_rtmp_videos(self, webpage):
+    def _extract_rtmp_video(self, webpage):
         # The server URL is hardcoded
         video_url = 'rtmpe://video.infoq.com/cfx/st/'
 
@@ -47,25 +50,43 @@ class InfoQIE(BokeCCBaseIE):
         playpath = 'mp4:' + real_id
 
         return [{
-            'format_id': 'rtmp',
+            'format_id': 'rtmp_video',
             'url': video_url,
             'ext': determine_ext(playpath),
             'play_path': playpath,
         }]
 
-    def _extract_http_videos(self, webpage):
-        http_video_url = self._search_regex(r'P\.s\s*=\s*\'([^\']+)\'', webpage, 'video URL')
-
+    def _extract_cookie(self, webpage):
         policy = self._search_regex(r'InfoQConstants.scp\s*=\s*\'([^\']+)\'', webpage, 'policy')
         signature = self._search_regex(r'InfoQConstants.scs\s*=\s*\'([^\']+)\'', webpage, 'signature')
         key_pair_id = self._search_regex(r'InfoQConstants.sck\s*=\s*\'([^\']+)\'', webpage, 'key-pair-id')
+        return 'CloudFront-Policy=%s; CloudFront-Signature=%s; CloudFront-Key-Pair-Id=%s' % (
+            policy, signature, key_pair_id)
+
+    def _extract_http_video(self, webpage):
+        http_video_url = self._search_regex(r'P\.s\s*=\s*\'([^\']+)\'', webpage, 'video URL')
+        return [{
+            'format_id': 'http_video',
+            'url': http_video_url,
+            'ext': determine_ext(http_video_url),
+            'http_headers': {
+                'Cookie': self._extract_cookie(webpage)
+            },
+        }]
+
+    def _extract_http_audio(self, webpage):
+        http_audio_url = self._search_regex(r'<input[^>]*?name="filename"[^>]*?value="([^\"]+)"[^>]*?>', webpage, 'audio URL', fatal=False)
+        if http_audio_url is None:
+            return []
+        http_audio_url = compat_urlparse.urljoin('http://res.infoq.com/downloads/mp3downloads/', http_audio_url)
 
         return [{
-            'format_id': 'http',
-            'url': http_video_url,
+            'format_id': 'http_audio',
+            'url': http_audio_url,
+            'ext': determine_ext(http_audio_url, ""),
+            'vcodec': 'none',
             'http_headers': {
-                'Cookie': 'CloudFront-Policy=%s; CloudFront-Signature=%s; CloudFront-Key-Pair-Id=%s' % (
-                    policy, signature, key_pair_id),
+                'Cookie': self._extract_cookie(webpage)
             },
         }]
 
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/infoq.py
```diff
@@ -47,25 +50,43 @@ def _extract_rtmp_videos(self, webpage):
...
+    def _extract_http_audio(self, webpage):
+        http_audio_url = self._search_regex(r'<input[^>]*?name="filename"[^>]*?value="([^\"]+)"[^>]*?>', webpage, 'audio URL', fatal=False)
+        if http_audio_url is None:
+            return []
+        http_audio_url = compat_urlparse.urljoin('http://res.infoq.com/downloads/mp3downloads/', http_audio_url)
+
+        return [{
+            'format_id': 'http_audio',
+            'url': http_audio_url,
+            'ext': determine_ext(http_audio_url, ""),
```
wader: not sure what default should be for audio only format. `determine_ext` defaults is `unknown_video`

wader:
How do i write a test case for this? tried to find some other extractor that has multiple formats with audio only but none of them seems to test it.

I noticed that some audio URLs fail with 403 unauthorized but that seems to be the case if you try to download using a browser also.

wader:
@yan12125 hi! do you have time and knowledge to review this?

youtube_dl/extractor/infoq.py
```diff
@@ -47,25 +50,43 @@ def _extract_rtmp_videos(self, webpage):
...
+        return 'CloudFront-Policy=%s; CloudFront-Signature=%s; CloudFront-Key-Pair-Id=%s' % (
+            policy, signature, key_pair_id)
 
+    def _extract_http_video(self, webpage):
+        http_video_url = self._search_regex(r'P\.s\s*=\s*\'([^\']+)\'', webpage, 'video URL')
         return [{
-            'format_id': 'http',
+            'format_id': 'http_video',
             'url': http_video_url,
+            'ext': determine_ext(http_video_url),
```
yan12125: In this case you can just omit this line as determine_ext will be automatically called under the hood.

youtube_dl/extractor/infoq.py
```diff
@@ -47,25 +50,43 @@ def _extract_rtmp_videos(self, webpage):
...
+            'format_id': 'rtmp_video',
             'url': video_url,
             'ext': determine_ext(playpath),
             'play_path': playpath,
         }]
 
-    def _extract_http_videos(self, webpage):
-        http_video_url = self._search_regex(r'P\.s\s*=\s*\'([^\']+)\'', webpage, 'video URL')
-
+    def _extract_cookie(self, webpage):
```
yan12125: cookie => cookies. There are 3 items.

youtube_dl/extractor/infoq.py
```diff
@@ -47,25 +50,43 @@ def _extract_rtmp_videos(self, webpage):
...
+            'http_headers': {
+                'Cookie': self._extract_cookie(webpage)
+            },
+        }]
+
+    def _extract_http_audio(self, webpage):
+        http_audio_url = self._search_regex(r'<input[^>]*?name="filename"[^>]*?value="([^\"]+)"[^>]*?>', webpage, 'audio URL', fatal=False)
+        if http_audio_url is None:
+            return []
+        http_audio_url = compat_urlparse.urljoin('http://res.infoq.com/downloads/mp3downloads/', http_audio_url)
```
yan12125: Is 'http://res.infoq.com/downloads/mp3downloads/' hardcoded somewhere? If so it's better to point out where it is. (for example, xyz.html or abc.js)

youtube_dl/extractor/infoq.py
yan12125: The best way is modifying determine_ext and make it recognize whether it's audio or video. As of now you can just omit this line and let the ext be unknown_video.

yan12125:
> How do i write a test case for this? tried to find some other extractor that has multiple formats with video and audio only but none of them seems to test it.

Try to specify 'format' in 'params'. See youtube.py for some examples.

> I noticed that some audio URLs fail with 403 unauthorized but that seems to be the case if you try to download using a browser also.

You can use _is_valid_url to check URLs.

wader:
@yan12125 Thanks for the review! Fixed most of the stuff, will have a look at tests.

I tried to use `_is_valid_url` but not sure it will work in this case. Ideally i would like to do a HEAD requests just to probe if the file is there or not.

yan12125:
_is_valid_url connects to the specified URL and disconnect immediately. It doesn't use HEAD as some websites block HEAD requests. You can improve _is_valid_url if it doesn't fit your need.

yan12125:
Is there an example with audio tracks?

wader:
Ok then it should work. I will have a look at how it behaves when the response is 403.

Here is an example with audio only https://www.infoq.com/presentations/Simple-Made-Easy

youtube_dl/extractor/infoq.py
```diff
@@ -47,25 +50,43 @@ def _extract_rtmp_videos(self, webpage):
...
+            },
+        }]
+
+    def _extract_http_audio(self, webpage):
+        http_audio_url = self._search_regex(r'<input[^>]*?name="filename"[^>]*?value="([^\"]+)"[^>]*?>', webpage, 'audio URL', fatal=False)
+        if http_audio_url is None:
+            return []
+        # base URL is found in the Location header in the response returned by
+        # GET https://www.infoq.com/mp3download.action?filename=... when logged in.
+        http_audio_url = compat_urlparse.urljoin('http://res.infoq.com/downloads/mp3downloads/', http_audio_url)
```
yan12125: It's better to use the original ```mp3download.action``` URL here as redirected URL may change in the future.

youtube_dl/extractor/infoq.py
```diff
@@ -47,25 +50,43 @@ def _extract_rtmp_videos(self, webpage):
...
             'url': http_video_url,
             'http_headers': {
-                'Cookie': 'CloudFront-Policy=%s; CloudFront-Signature=%s; CloudFront-Key-Pair-Id=%s' % (
-                    policy, signature, key_pair_id),
+                'Cookie': self._extract_cookies(webpage)
+            },
+        }]
+
+    def _extract_http_audio(self, webpage):
+        http_audio_url = self._search_regex(r'<input[^>]*?name="filename"[^>]*?value="([^\"]+)"[^>]*?>', webpage, 'audio URL', fatal=False)
```
yan12125: Use _hidden_inputs or _form_hidden_inputs - they're better than the long regex pattern.

youtube_dl/extractor/infoq.py
wader: The problem is that you will only be redirected to the mp3 file if your logged in

youtube_dl/extractor/infoq.py
wader: Ok. Is there support for dom tree parsing? could'nt find anything

youtube_dl/extractor/infoq.py
yan12125: Got it. Sorry for not reading the comment carefully

youtube_dl/extractor/infoq.py
yan12125: An example is extract_attributes, which uses HTMLParser. The point here is that _hidden_inputs are much well tested. On the contrary, there are lots of tricks in HTMLParser on different Python versions.

wader: Ok i will have a look at _hidden_inputs

wader:
@yan12125 Think i'ved resolved all the issues now

yan12125:
Thanks!

wader:
@yan12125  Thanks for all help! 👍

PR #11578, opened 2017-01-01
Update Readme: Set HOME properly in install command

rpunkfu:
Hi! :)

It's not always a default behaviour, but when you use `sudo pip` you most likely may want to use `-H` flag to set HOME value to directory of the target user. :)


yan12125:
Thanks!

PR #11596, opened 2017-01-03
[Beam] Add new extractor

shizeeg:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

It's an extractor for https://beam.pro/ streaming service recently acquired my Microsoft. 

Well, here is another one extractor from me. It doesn't have all features supported yet, but `live streams` are fully functional. It's a smaller one ATM but more robust one. So I'm just trying to get used to coding standards that are common in here.

### DIFF
--- a/youtube_dl/extractor/beampro.py
+++ b/youtube_dl/extractor/beampro.py
@@ -3,11 +3,12 @@ from __future__ import unicode_literals
 
 from .common import InfoExtractor
 from ..utils import (
-    int_or_none,
     ExtractorError,
-    compat_str,
     clean_html,
+    compat_str,
+    int_or_none,
     parse_iso8601,
+    try_get,
 )
 
 
@@ -15,8 +16,7 @@ class BeamProLiveIE(InfoExtractor):
     IE_NAME = 'Beam:live'
     _VALID_URL = r'https?://(?:\w+.)?beam.pro/(?P<id>[^?]+)$'
     _API_CHANNEL = 'https://beam.pro/api/v1/channels/{0}'
-    _API_MANIFEST = 'https://beam.pro/api/v1/channels/{0}/manifest.{1}'
-    _VALID_MANIFESTS = ('smil', 'm3u8', 'light', 'light2', 'ftl', 'ftlOld')
+    _API_MANIFEST = 'https://beam.pro/api/v1/channels/{0}/manifest.m3u8'
     _RATINGS = {'family': 0, 'teen': 13, '18+': 18}
 
     _TEST = {
@@ -48,11 +48,9 @@ class BeamProLiveIE(InfoExtractor):
             raise ExtractorError('{0} is offline'.format(channel_id), expected=True)
 
         formats = self._extract_m3u8_formats(
-            self._API_MANIFEST.format(
-                chan_data.get('id'),
-                self._VALID_MANIFESTS[1]), channel_id, ext='mp4',
-        )
-        self._sort_formats(formats, 'vbr')
+            self._API_MANIFEST.format(chan_data.get('id')), channel_id, ext='mp4')
+
+        self._sort_formats(formats)
         info = {}
         info['formats'] = formats
         if chan_data:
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/beampro.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+        self._sort_formats(formats, 'vbr')
+        info = {}
+        info['formats'] = formats
+        if chan_data:
+            info.update(self._extract_info(chan_data))
+        if not info.get('title'):
+            info['title'] = self._live_title(channel_id)
+        if not info.get('id'):  # barely possible but just in case
+            info['id'] = compat_str(abs(hash('{0}/{1}'.format(channel_id, formats[0]))) % (10 ** 8))
```
dstftw: Just `channel_id`.

youtube_dl/extractor/beampro.py
```diff
@@ -0,0 +1,89 @@
...
+        'params': {
+            'skip_download': True,
+        },
+    }
+
+    def _real_extract(self, url):
+        channel_id = self._match_id(url)
+        chan_data = self._download_json(self._API_CHANNEL.format(channel_id), channel_id)
+
+        if not chan_data.get('online'):
```
dstftw: This will also apply when `online` key does not exist.

youtube_dl/extractor/beampro.py
```diff
@@ -0,0 +1,89 @@
...
+    parse_iso8601,
+)
+
+
+class BeamProLiveIE(InfoExtractor):
+    IE_NAME = 'Beam:live'
+    _VALID_URL = r'https?://(?:\w+.)?beam.pro/(?P<id>[^?]+)$'
+    _API_CHANNEL = 'https://beam.pro/api/v1/channels/{0}'
+    _API_MANIFEST = 'https://beam.pro/api/v1/channels/{0}/manifest.{1}'
+    _VALID_MANIFESTS = ('smil', 'm3u8', 'light', 'light2', 'ftl', 'ftlOld')
```
dstftw: Only m3u8 is used. What's the point storing others?

youtube_dl/extractor/beampro.py
```diff
@@ -0,0 +1,89 @@
...
+        if not info.get('id'):  # barely possible but just in case
+            info['id'] = compat_str(abs(hash('{0}/{1}'.format(channel_id, formats[0]))) % (10 ** 8))
+
+        return info
+
+    def _rating_to_age(self, rating):
+        return self._RATINGS[rating] if rating in self._RATINGS else None
+
+    def _extract_info(self, info):
+        thumbnail = info['thumbnail'].get('url') if info.get('thumbnail') else None
```
dstftw: Use `utils.try_get`.

youtube_dl/extractor/beampro.py
```diff
@@ -0,0 +1,89 @@
...
+
+        if not chan_data.get('online'):
+            raise ExtractorError('{0} is offline'.format(channel_id), expected=True)
+
+        formats = self._extract_m3u8_formats(
+            self._API_MANIFEST.format(
+                chan_data.get('id'),
+                self._VALID_MANIFESTS[1]), channel_id, ext='mp4',
+        )
+        self._sort_formats(formats, 'vbr')
```
dstftw: `field_preference` must be `list` or `tuple`. There is no need to touch this usually since default sorting works fine.

youtube_dl/extractor/beampro.py
```diff
@@ -0,0 +1,89 @@
...
+            info['title'] = self._live_title(channel_id)
+        if not info.get('id'):  # barely possible but just in case
+            info['id'] = compat_str(abs(hash('{0}/{1}'.format(channel_id, formats[0]))) % (10 ** 8))
+
+        return info
+
+    def _rating_to_age(self, rating):
+        return self._RATINGS[rating] if rating in self._RATINGS else None
+
+    def _extract_info(self, info):
```
dstftw: What's the point extracting it in method that is only used once?

youtube_dl/extractor/beampro.py
```diff
@@ -0,0 +1,89 @@
...
+        return {
+            'id': video_id,
+            'title': info.get('name'),
+            'description': clean_html(info.get('description')),
+            'age_limit': self._rating_to_age(info.get('audience')),
+            'is_live': True if info.get('online') else False,
+            'timestamp': parse_iso8601(info.get('updatedAt')),
+            # 'release_date': info.get('createdAt'),
+            # 'upload_date': info.get('updatedAt'),
+            # 'formats': formats,
```
dstftw: To be removed.

youtube_dl/extractor/beampro.py
shizeeg: I'm planning to expand this extractor later. There are lots of stuff to do: VODs, various playlists like channels, categories, etc., etc.

youtube_dl/extractor/beampro.py
shizeeg: but it is:
 `info.get('online')` returns `None if 'online' doesn't exist and `if not None` is True in that case.

Issue #11604, opened 2017-01-04
No Error for Private Youtube Playlists


david476:
---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2017.01.05*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [ x] I've **verified** and **I assure** that I'm running youtube-dl **2017.01.05**

### Before submitting an *issue* make sure you have:
- [ x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [ x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [ ] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ x] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

I was trying to download a playlist from youtube and youtube-dl said that it had 0 videos. It turns out the playlist was private (duh!) but it would be nice to have some more substantial (obvious) message to let you know.

PR #11633, opened 2017-01-08
[youtube:playlist] Fix 'this playlist doesn't exist/is private' error not working (#11604)

Grzechooo:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This pull request fixes bug #11604.
The issue at hand is the fact that YouTube's page structure has changed such that the previous method of checking whether the playlist exists/is private ceased working. The notice now looks like this:
```html
    <div class="yt-alert-message" tabindex="0">
          This playlist is private.
    </div>
```
the old method only checks for `<div class="yt-alert-message">` and thus broke.

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -1955,7 +1955,8 @@ class YoutubePlaylistIE(YoutubePlaylistBaseInfoExtractor):
         url = self._TEMPLATE_URL % playlist_id
         page = self._download_webpage(url, playlist_id)
 
-        for match in re.findall(r'<div class="yt-alert-message">([^<]+)</div>', page):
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/youtube.py
```diff
@@ -1955,7 +1955,8 @@ def _extract_playlist(self, playlist_id):
         url = self._TEMPLATE_URL % playlist_id
         page = self._download_webpage(url, playlist_id)
 
-        for match in re.findall(r'<div class="yt-alert-message">([^<]+)</div>', page):
+        # the yt-alert-message now has tabindex attribute (see https://github.com/rg3/youtube-dl/issues/11604)
+        for match in re.findall(r'<div class="yt-alert-message"[^>]+?>([^<]+)</div>', page):
```
yan12125: ```[^>]+?``` should be ```[^>]*```

PR #11635, opened 2017-01-08
Adjusted aria2 flags

Manouchehri:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

aria2c by default will attempt to reduce the amount of new connections needed for a parallel download, but this may be undesirable for streaming videos as the start of the file may not be finished in time.

Instead, we should use `geom`, which starts downloading the video at the beginning, but still attempts to reduce connections for later pieces of the file.

> https://aria2.github.io/manual/en/html/aria2c.html#cmdoption--stream-piece-selector


Manouchehri:
Ping?

PR #11643, opened 2017-01-08
[orf] Update OE1 extractor, merge with FM4.

phaer:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Since oe1.orf.at has been updated, both ORF radios supported by youtube_dl
use the same API. This commit honors this fact by merging both extractors
into one, improving code re-use and making sure that oe1.orf.at will continue to work
whenever the legacy API gets deprecated.

flake8 has no complaints and the tests are passing.

### DIFF
--- a/youtube_dl/extractor/orf.py
+++ b/youtube_dl/extractor/orf.py
@@ -143,38 +143,6 @@ class ORFTVthekIE(InfoExtractor):
 
 
 class ORFRadioIE(InfoExtractor):
-    _VALID_URL = r'https?://(?P<station>oe1|fm4)\.orf\.at/(?:7tage/?#|player/)(?P<date>[0-9]+)/(?P<show>\w+)'
-
-    _TESTS = [
-        {
-            'url': 'http://fm4.orf.at/player/20170107/CC',
-            'md5': '2b0be47375432a7ef104453432a19212',
-            'info_dict': {
-                'id': '2017-01-07_2100_tl_54_7DaysSat18_31295',
-                'ext': 'mp3',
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/orf.py
```diff
@@ -144,77 +142,60 @@ def quality_to_int(s):
...
-            'id': show_id,
-            'title': data['item']['title'],
-            'url': data['item']['url_stream'],
-            'ext': 'mp3',
-            'description': data['item'].get('info'),
-            'timestamp': unix_timestamp
-        }
-
-
-class ORFFM4IE(InfoExtractor):
```
dstftw: Original extractor classes (`ie_key`s) should be preserved. Otherwise you break existing download archives.

midas02: @dstftw I was looking at fixing this, but:

1. In this commit it seems to me you have removed a class yourself. What's the difference with this situation? https://github.com/rg3/youtube-dl/commit/6d1ded75021ca76f14c4fe8d3aa698704bae1dd9
2. What do you mean by 'you break existing download archives'?

dstftw: 1. There is no other way in this case. pluzz.francetv.fr does not exist anymore.
2. `ie_key`s in `--download-archive` file.

phaer: Sorry, it took me a while to come back to this pull request. My latest commit just re-introduces the old classes which are now simply extending their common base class. This should preserve `ie_key`s. What do you think, @dstftw ?

PR #11705, opened 2017-01-13
[brightcove] Fix misplaced backslash in a regexp

jwilk:
Found using [pydiatra](https://github.com/jwilk/pydiatra).


dstftw:
Thanks.

PR #11724, opened 2017-01-15
[Iwara] Fix download

hakatashi:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

[Iwara](http://www.iwara.tv/) now switched their player to get media URLs from API, not HTML. So the previous extractor is now [failing](https://travis-ci.org/rg3/youtube-dl/jobs/192039917#L892).

This patch fixes the extractor to use API to extract video URL and fixed some test cases as the following.

* Some MD5 are updated according to the source change. I manually confirmed the downloaded content is correct.
* The video of the last test case was re-rated to 18+, and I updated the field.
* Then there are no test cases for 0+ video (which has `www.iwara.tv` domain), and I added one.

### DIFF
--- a/youtube_dl/extractor/iwara.py
+++ b/youtube_dl/extractor/iwara.py
@@ -52,26 +52,49 @@ class IwaraIE(InfoExtractor):
         # ecchi is 'sexy' in Japanese
         age_limit = 18 if hostname.split('.')[0] == 'ecchi' else 0
 
-        entries = self._parse_html5_media_entries(url, webpage, video_id)
-
-        if not entries:
-            iframe_url = self._html_search_regex(
-                r'<iframe[^>]+src=([\'"])(?P<url>[^\'"]+)\1',
-                webpage, 'iframe URL', group='url')
-            return {
-                '_type': 'url_transparent',
-                'url': iframe_url,
-                'age_limit': age_limit,
-            }
-
         title = remove_end(self._html_search_regex(
             r'<title>([^<]+)</title>', webpage, 'title'), ' | Iwara')
 
-        info_dict = entries[0]
-        info_dict.update({
+        api_url = 'http://%s/api/video/%s' % (hostname, video_id)
+
+        raw_formats = self._download_json(
+            api_url, video_id,
+            note='Downloading formats')
+
+        formats = []
+
+        for raw_format in raw_formats:
+            new_format = {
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/iwara.py
```diff
@@ -52,26 +61,72 @@ def _real_extract(self, url):
...
-        info_dict.update({
+            if raw_format.get('resolution') == '1080p':
+                height = 1080
+            elif raw_format.get('resolution') == '720p':
+                height = 720
+            elif raw_format.get('resolution') == '540p':
+                height = 540
+            elif raw_format.get('resolution') == '360p':
+                height = 360
+            else:
```
dstftw: Code duplication.

youtube_dl/extractor/iwara.py
```diff
@@ -52,26 +61,72 @@ def _real_extract(self, url):
...
+            api_url, video_id,
+            note='Downloading formats')
+
+        formats = []
+
+        for raw_format in raw_formats:
+            new_format = {
+                'url': raw_format['uri'],
+                'resolution': raw_format['resolution'],
+                'format_id': raw_format['resolution'],
```
dstftw: Should not break if there is no `resolution` key.

youtube_dl/extractor/iwara.py
```diff
@@ -52,26 +61,72 @@ def _real_extract(self, url):
...
+                height = 360
+            else:
+                height = None
+
+            if height is not None:
+                new_format['width'] = int(height / 9.0 * 16.0)
+                new_format['height'] = height
+
+            if raw_format.get('mime') == 'video/mp4':
+                new_format['ext'] = 'mp4'
```
dstftw: `utils.mimetype2ext`.

youtube_dl/extractor/iwara.py
hakatashi: Thanks. How should I fix this?

yan12125:
@hakatashi Sorry I didn't notice here's a pull request and I've push my patch to caf0f5f8b7d0854caaf6778fe3a646ee0d7668fe and pick some codes from this PR in 2ab2c0d1f53f66614eda4fefb042e851e78097f0. Anyway thanks for the contribution!

PR #11134, opened 2016-11-07
[rai] changed the extractor for the new site and dismisison of the older one

timendum:
### Before submitting a *pull request* make sure you have:
- [x] At least skimmed through [adding new extractor tutorial](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/rg3/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

Rai recently changed its streaming website to raiplay.it and axed the old one, I heavy changed the extractor to reflect the new changes.

PR #11790, opened 2017-01-20
[rai] changed the extractor for the new site and dismisison of the old one

timendum:

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Rai changed its streaming website to raiplay.it and axed the old one, I heavy changed the extractor to reflect the new changes.

This is a follow up of #11134 , I've dropped the old one and fixed the conflicts.

All the tests pass in Italy, maybe some media can be geo-restricted.

### DIFF
--- a/youtube_dl/extractor/rai.py
+++ b/youtube_dl/extractor/rai.py
@@ -3,8 +3,8 @@ from __future__ import unicode_literals
 from .common import InfoExtractor
 from ..compat import compat_urlparse
 from ..utils import (
-    determine_ext,
     ExtractorError,
+    determine_ext,
     find_xpath_attr,
     fix_xml_ampersands,
     int_or_none,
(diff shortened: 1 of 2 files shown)


timendum:
@dstftw  can you check?

Thansk

dstftw:
Not working:
http://www.report.rai.it/dl/Report/puntata/ContentItem-0c7a664b-d0f4-4b2c-8835-3f82e46f433e.html
http://www.tg1.rai.it/dl/tg1/2010/edizioni/ContentSet-9b6e0cba-4bef-4aef-8cf0-9f7f665b7dfb-tg1.html?item=undefined
http://www.rainews.it/dl/rainews/live/ContentItem-3156f2f2-dc70-4953-8e2f-70d7489d4ce9.html

timendum:
@dstftw  
I've imported some code from the older extractor for better handling of old urls.

Can you check now?

Should I squash the commits?

Thanks.

dstftw:
>Should I squash the commits?

Yes.

timendum:
@dstftw
Commits squashed.

rabblac:
Still open? Is anything blocking this? @dstftw

dstftw:
Check code with flake8.

timendum:
Done and squashed.

There are still some lines longer then 100 chars, but they are test cases and regexps.

youtube_dl/extractor/rai.py
```diff
@@ -55,181 +55,157 @@ def _extract_relinker_formats(self, relinker_url, video_id):
...
+                    thumbnails.append({
+                        'url': value.replace('[RESOLUTION]', self._RESOLUTION)
+                    })
 
-    def _real_extract(self, url):
-        video_id = self._match_id(url)
+        formats = None
+        duration = None
+        if 'video' in media:
+            formats
```
dstftw: What's this for?

youtube_dl/extractor/rai.py
```diff
@@ -55,181 +55,157 @@ def _extract_relinker_formats(self, relinker_url, video_id):
...
+                if value:
+                    thumbnails.append({
+                        'url': value.replace('[RESOLUTION]', self._RESOLUTION)
+                    })
 
-    def _real_extract(self, url):
-        video_id = self._match_id(url)
+        formats = None
+        duration = None
+        if 'video' in media:
```
dstftw: This is pointless. If no formats can be extracted extraction should stop immediately.

youtube_dl/extractor/rai.py
```diff
@@ -55,181 +55,157 @@ def _extract_relinker_formats(self, relinker_url, video_id):
...
-            'title': media['name'],
-            'description': media.get('desc'),
-            'thumbnails': thumbnails,
-            'uploader': media.get('author'),
-            'upload_date': unified_strdate(media.get('date')),
-            'duration': parse_duration(media.get('length')),
-            'formats': formats,
-            'subtitles': subtitles,
-        }
+        canonical_url = self._og_search_url(webpage)
```
dstftw: What's the point of this? `canonical_url` is the same as `url`.

timendum: Not in the first test case and in every case where a query parameter is added to url.

youtube_dl/extractor/rai.py
timendum: It's an initialization of a variable, I see no harm in it, but I'm open to suggestions.

youtube_dl/extractor/rai.py
dstftw: Everything apart from query is the same and this unnecessary webpage downloading should be removed.

youtube_dl/extractor/rai.py
dstftw: It's a conditional expression that allows to skip formats extraction. As I've already said this is **pointless** since valid list of formats must be always present. `None` `formats` is **not allowed**.

timendum:
I've hopefully resolved all the issue:

* The canonical url is now obtained removing the query with urlparse
* An ExtractorError is thrown if no video is found

PR #11835, opened 2017-01-25
[Bandcamp] Extract additional fields

gkoelln:
* [bandcamp] Add additional extraction fields

Adding additional extraction fields

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Adding additional extraction fields for tracks, album, and artist information

### DIFF
--- a/youtube_dl/extractor/bandcamp.py
+++ b/youtube_dl/extractor/bandcamp.py
@@ -16,6 +16,7 @@ from ..utils import (
     int_or_none,
     parse_filesize,
     unescapeHTML,
+    unified_strdate,
     update_url_query,
 )
 
@@ -47,6 +48,18 @@ class BandcampIE(InfoExtractor):
         mobj = re.match(self._VALID_URL, url)
         title = mobj.group('title')
         webpage = self._download_webpage(url, title)
+        release_date = self._search_regex(
+            r'"?(?:album_)?release_date"?\s*?:\s*?"(.*?)"',
+            webpage, 'release date', fatal=False)
+        release_date = unified_strdate(release_date)
+        album = self._search_regex(
+            r'album_title\s*:\s*"((?:\\.|[^"\\])+?)"',
+            webpage, 'album', default=None)
+        artist = self._search_regex(
+            r'artist\s*:\s*"((?:\\.|[^"\\])+?)"',
+            webpage, 'artist', fatal=False)
+        if release_date:
+            release_year = release_date[:4]
         m_download = re.search(r'freeDownloadPage: "(.*?)"', webpage)
         if not m_download:
             m_trackinfo = re.search(r'trackinfo: (.+),\s*?\n', webpage)
@@ -76,7 +89,14 @@ class BandcampIE(InfoExtractor):
                     'id': track_id,
                     'title': data['title'],
                     'formats': formats,
+                    'release_date': release_date,
                     'duration': float_or_none(data.get('duration')),
+                    'track': data.get('title'),
+                    'track_number': int_or_none(data.get('track_num')),
+                    'track_id': track_id,
+                    'album': album,
+                    'artist': artist,
+                    'release_year': release_year,
                 }
             else:
                 raise ExtractorError('No free songs found')
@@ -85,6 +105,10 @@ class BandcampIE(InfoExtractor):
         video_id = self._search_regex(
             r'(?ms)var TralbumData = .*?[{,]\s*id: (?P<id>\d+),?$',
             webpage, 'video id')
+
+        track_number = self._search_regex(
+            r'"track_num":(?P<track_number>\d+),',
+            webpage, 'track number', fatal=False)
 
         download_webpage = self._download_webpage(
             download_link, video_id, 'Downloading free downloads page')
@@ -145,9 +169,15 @@ class BandcampIE(InfoExtractor):
             'title': title,
             'thumbnail': info.get('thumb_url'),
             'uploader': info.get('artist'),
+            'release_date': release_date,
             'artist': artist,
             'track': track,
             'formats': formats,
+            'track_number': int_or_none(track_number),
+            'track_id': video_id,
+            'album': album,
+            'artist': artist,
+            'release_year': release_year,
         }
 
 
@@ -233,5 +263,6 @@ class BandcampAlbumIE(InfoExtractor):
             'uploader_id': uploader_id,
             'id': playlist_id,
             'title': title,
+            'album': title,
             'entries': entries,
         }


youtube_dl/extractor/bandcamp.py
```diff
@@ -233,5 +256,6 @@ def _real_extract(self, url):
             'uploader_id': uploader_id,
             'id': playlist_id,
             'title': title,
+            'album': title,
             'entries': entries,
         }
```
dstftw: All these fields must be optional.

gkoelln: I'm afraid I don't know what that means.

dstftw: That's explained in code conventions you've ticked as read.

gkoelln: Fair enough.  So will adding fatal=False be enough here?

gkoelln: Added more edits.  Am I getting closer?

gkoelln: One more edit since the last comment. It all tested well, from what I can tell.

gkoelln: Added track_id to free download

youtube_dl/extractor/bandcamp.py
```diff
@@ -77,6 +86,12 @@ def _real_extract(self, url):
                     'title': data['title'],
                     'formats': formats,
                     'duration': float_or_none(data.get('duration')),
+                    'track': data.get('title'),
+                    'track_number': data.get('track_num'),
+                    'track_id': track_id,
```
dstftw: It's not a track id.

youtube_dl/extractor/bandcamp.py
```diff
@@ -77,6 +86,12 @@ def _real_extract(self, url):
...
+                    'album': album,
+                    'album_artist': album_artist,
```
dstftw: It's artist not an album artist.

youtube_dl/extractor/bandcamp.py
```diff
@@ -86,6 +101,10 @@ def _real_extract(self, url):
             r'(?ms)var TralbumData = .*?[{,]\s*id: (?P<id>\d+),?$',
             webpage, 'video id')
 
+        track_number = self._search_regex(
+            r'"track_num":(?P<track_number>\d+),',
+            webpage, 'track number', fatal=False)
```
dstftw: Must be int.

youtube_dl/extractor/bandcamp.py
```diff
@@ -77,6 +86,12 @@ def _real_extract(self, url):
                     'title': data['title'],
                     'formats': formats,
                     'duration': float_or_none(data.get('duration')),
+                    'track': data.get('title'),
+                    'track_number': data.get('track_num'),
```
dstftw: Must be int.

gkoelln: Understood.

youtube_dl/extractor/bandcamp.py
gkoelln: Then what's the difference between id and track_id?

youtube_dl/extractor/bandcamp.py
gkoelln: "album_artist: List of all artists appeared on the album"
Based on that definition, I guess you're right.  I was thinking album artist vs. track artist (like on a mixtape, or similar).

youtube_dl/extractor/bandcamp.py
gkoelln: Understood.

gkoelln:
Changes made.

dstftw:
Squash commits.

gkoelln:
Sorry. Still learning.

gkoelln:
Commits have been squashed, by the way.

youtube_dl/extractor/bandcamp.py
```diff
@@ -47,6 +47,15 @@ def _real_extract(self, url):
         mobj = re.match(self._VALID_URL, url)
         title = mobj.group('title')
         webpage = self._download_webpage(url, title)
+        album = self._search_regex(
+            r'(?ms).*?title\s*?:\s*?"(?P<album>.*?)",',
```
dstftw: This extracts first title found not necessarily an album title.
All captured groups must be at least one character length.

youtube_dl/extractor/bandcamp.py
```diff
@@ -47,6 +47,15 @@ def _real_extract(self, url):
...
+            webpage, 'album', fatal=False)
+        artist = self._search_regex(
+            r'(?ms)var EmbedData = .*?[{,]\s*artist:\s*?"(?P<artist>.*?)",$',
```
dstftw: What's the point of this noise `var EmbedData = .*?[{,]\s*`?

youtube_dl/extractor/bandcamp.py
```diff
@@ -47,6 +47,15 @@ def _real_extract(self, url):
...
+            webpage, 'artist', fatal=False)
+        release_year = self._search_regex(
+            r'(?ms).*?release_date"?:\s*?"\d+ \w+ (?P<release_year>\d+)\s*?.*?GMT",',
```
dstftw: `utils.unified_strdate`, `release_date`.

youtube_dl/extractor/bandcamp.py
gkoelln: So, something like this?
```
album = self._search_regex(
    r'(?ms).*?album_title\s*?:\s*?"(?P<album>[^"]+)",',
```
Also, if a track is just a single track and not part of an album (such as the youtube-dl test song), do we just want to leave album as NA or have the album name be the same as the track title?  (Perhaps that's just a personal preference of some?)

youtube_dl/extractor/bandcamp.py
gkoelln: I wanted to use the artist that was included in the EmbedData var (which I can now see is the first instance of artist, usually). In most cases, three of four instances of the artist are the same, with one being null (in TralbumData).  Should I just remove the artist field altogether?

youtube_dl/extractor/bandcamp.py
dstftw: It should extract the album title field reported by the webpage. No more no less.

youtube_dl/extractor/bandcamp.py
dstftw: If they are all the same extract any that is not null.

gkoelln:
Changes made.

Issue #11840, opened 2017-01-26
Site support for euronews.com 


shirishag75:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like that [x])
- Use *Preview* tab to see how your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2017.01.25*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2017.01.25**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [ ] Bug report (encountered problems with youtube-dl)
- [x] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

─[$] youtube-dl -vF "http://www.euronews.com/2017/01/25/naples-airport-growing-at-double-the-national-average"                      
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'-vF', u'http://www.euronews.com/2017/01/25/naples-airport-growing-at-double-the-national-average']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2017.01.25
[debug] Python version 2.7.13 - Linux-4.9.0-1-amd64-x86_64-with-debian-9.0
[debug] exe versions: ffmpeg 3.2.2-1, ffprobe 3.2.2-1, rtmpdump 2.4
[debug] Proxy map: {}
[generic] naples-airport-growing-at-double-the-national-average: Requesting header
WARNING: Falling back on generic information extractor.
[generic] naples-airport-growing-at-double-the-national-average: Downloading webpage
[generic] naples-airport-growing-at-double-the-national-average: Extracting information
[...]

PR #11844, opened 2017-01-26
[euronews] Add new information extractor

goggle:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

### Description

In [#11840](https://github.com/rg3/youtube-dl/issues/11840) someone suggested to add youtube-dl support for euronews.com. Here is a proposal.

It seems that the site uses jwplayer, but hosts its videos itself as a mp4 file. I did not find any other formats, except the same video URL with a '/med/' included somewhere, but it seems as though it's exactly the same mp4 file...

### DIFF
--- a/youtube_dl/extractor/euronews.py
+++ b/youtube_dl/extractor/euronews.py
@@ -1,0 +1,61 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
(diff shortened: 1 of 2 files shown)


dstftw:
This is already covered by the generic extractor.

PR #11880, opened 2017-01-30
[vimeo] Add support for extracting licenses

### LINKED ISSUE #8726: Feature request: Add video license to output dict of Vimeo videos
(Related ticket for YouTube: #8725)
There is no mentioning of video license in output dict of Vimeo videos:

```
>>> dl.extract_info('https://vimeo.com/157316669', download=False).keys()
WARNING: Parameter outtmpl is bytes, but should be a unicode string. Put  from __future__ import unicode_literals  at the top of your code file or consider switching to Python 3.x.
[debug] youtube-dl version 2016.03.01
[debug] Python version 2.7.6 - Linux-3.19.0-49-generic-x86_64-with-Ubuntu-14.04-trusty
[debug] exe versions: avconv 9.18-6, avprobe 9.18-6
[debug] Proxy map: {}
[vimeo] 157316669: Downloading webpage
[vimeo] 157316669: Extracting information
[vimeo] 157316669: Downloading webpage
[vimeo] 157316669: Downloading JSON metadata
[vimeo] 157316669: Downloading m3u8 information
[u'display_id', u'upload_date', u'protocol', u'description', u'extractor', u'requested_subtitles', u'format', u'tbr', u'height', u'like_count', u'uploader', u'duration', u'format_id', u'id', u'subtitles', u'view_count', u'playlist', u'thumbnails', u'fps', u'title', u'url', u'extractor_key', u'http_headers', u'ext', u'uploader_id', u'width', u'comment_count', u'playlist_index', u'webpage_url', u'formats', u'thumbnail', u'webpage_url_basename']
```

In a browser, the license is shown in the popup of "more" link beside the uploader.

Why is this needed? It is usually necessary to know the license in order to comply with the terms (see [downstream bug](https://github.com/Toollabs/video2commons/issues/6)).

legoktm:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This will extract the URL to the relevant Creative Commons license, if
the video is licensed using one of those. Otherwise it'll be `None`
since vimeo does not specify a license.

Fixes #8726

I wasn't really able to run the tests, they were spewing random http/url errors that seemed unrelated to my change. Also, my patch currently is returning a license URL rather than a human readable license name, is that acceptable?

### DIFF
--- a/youtube_dl/extractor/vimeo.py
+++ b/youtube_dl/extractor/vimeo.py
@@ -575,6 +575,9 @@ class VimeoIE(VimeoBaseInfoExtractor):
             like_count = None
             comment_count = None
 
+        # Extract license
+        license = self._search_regex(r'<link [^>]+rel="license" href="(.*?)">', webpage, 'license', default=None)
+
         formats = []
         download_request = sanitized_Request('https://vimeo.com/%s?action=load_download_config' % video_id, headers={
             'X-Requested-With': 'XMLHttpRequest'})
@@ -609,6 +612,7 @@ class VimeoIE(VimeoBaseInfoExtractor):
             'view_count': view_count,
             'like_count': like_count,
             'comment_count': comment_count,
+            'license': license,
         })
 
         return info_dict


legoktm:
From my reading of `self._search_regex` it looks like the default parameter takes precedent over fatal?

shizeeg:
@legoktm https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L643 Fatal is true by default.

dstftw:
`default` and `fatal` have different purposes:
- `default` should be used when missing value is okay and you provide a hardcoded default fallback (either `None` or another value) and [`fatal` is not considered](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L669-L672) in this case.
- `fatal` is used when missing value is not okay but not that match to break the extraction. Warning is emitted in this case.

Whether one should be used or not is defined by a source:
- If meta field is always present in a source and you can't field any counter example then use `fatal`. If warning starts to appear for this meta field some time later it clear that most likely the source has changed and the regex should be updated.
- If meta field is not always present in a source `default` with `None` should be used.

legoktm:
Thanks for explaining @dstftw. Vimeo only adds the license info if it's CC or PD, there's nothing copyrighted videos. So it seems like `default` is correct in this case, since the field isn't always there.

Issue #9974, opened 2016-07-01
Support member-only videos on VTM.be


MichaelDeBoey:
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.07.01**
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
### What is the purpose of your _issue_?
- [x] Bug report (encountered problems with youtube-dl)
- [x] Site support request (request for adding support for a new site)
- [x] Question

---

```
$ youtube-dl -v http://vtm.be/video?aid=144670
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://vtm.be/video?aid=144670']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.07.01
[debug] Python version 2.7.10 - Darwin-15.5.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg 3.0.1, ffprobe 3.0.1
[debug] Proxy map: {}
[generic] video?aid=144670: Requesting header
WARNING: Falling back on generic information extractor.
[generic] video?aid=144670: Downloading webpage
[generic] video?aid=144670: Extracting information
ERROR: Unsupported URL: http://vtm.be/video?aid=144670
Traceback (most recent call last):
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/generic.py", line 1481, in _real_extract
    doc = compat_etree_fromstring(webpage.encode('utf-8'))
  File "/usr/local/bin/youtube-dl/youtube_dl/compat.py", line 2524, in compat_etree_fromstring
    doc = _XML(text, parser=etree.XMLParser(target=_TreeBuilder(element_factory=_element_factory)))
  File "/usr/local/bin/youtube-dl/youtube_dl/compat.py", line 2513, in _XML
    parser.feed(text)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/xml/etree/ElementTree.py", line 1642, in feed
    self._raiseerror(v)
[...]

PR #11912, opened 2017-01-31
[vtm] Add extractor for member-only videos on VTM.be

mrBliss:
Implementation of the approach described in #9974.

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Straightforward implementation of the approach described in #9974.

### DIFF
--- a/youtube_dl/extractor/vtm.py
+++ b/youtube_dl/extractor/vtm.py
@@ -64,6 +64,9 @@ class VTMIE(InfoExtractor):
         },
     ]
 
+    def _real_initialize(self):
+        self._logged_in = False
+
     def _login(self):
         (username, password) = self._get_login_info()
         if username is None or password is None:
@@ -91,23 +94,26 @@ class VTMIE(InfoExtractor):
         self._uid_signature = auth_info['UIDSignature']
         self._signature_timestamp = auth_info['signatureTimestamp']
 
+        self._logged_in = True
+
     def _real_extract(self, url):
         mobj = re.match(self._VALID_URL, url)
         site_id = mobj.group('site_id')
 
-        webpage = self._download_webpage(url, None)
+        webpage = self._download_webpage(url, None, "Downloading webpage")
 
         # The URL sometimes contains the video id, but not always, e.g., test
-        # case 3. Fortunately, all webpages contain the video id.
+        # case 3. Fortunately, all webpages of videos requiring authentication
+        # contain the video id.
         video_id = self._search_regex(
             r'\\"vodId\\":\\"(.+?)\\"', webpage, 'video_id', default=None)
 
-        # When no video_id is found, it was most likely a video not requiring
-        # authentication, so just fall back to the generic extractor
+        # It was most likely a video not requiring authentication.
         if not video_id:
             return self.url_result(url, 'Generic')
 
-        self._login()
+        if not self._logged_in:
+            self._login()
 
         title = self._html_search_regex(
             r'\\"title\\":\\"(.+?)\\"', webpage, 'title', default=None)
--- a/youtube_dl/extractor/vtm.py
+++ b/youtube_dl/extractor/vtm.py
@@ -24,7 +24,6 @@ class VTMIE(InfoExtractor):
     _TESTS = [
         {
             'url': 'http://vtm.be/video/volledige-afleveringen/id/vtm_20170219_VM0678361_vtmwatch',
-            'md5': '9357926318cdfcb10079d85d56f4e04e',
             'info_dict': {
                 'id': 'vtm_20170219_VM0678361_vtmwatch',
                 'ext': 'mp4',
@@ -32,7 +31,27 @@ class VTMIE(InfoExtractor):
                 'description': 'md5:4be86427521e7b07e0adb0c9c554ddb2',
             },
             'skip_download': True,
-        }
+        },
+        {
+            'url': 'http://vtm.be/video/volledige-afleveringen/id/257107153551000',
+            'info_dict': {
+                'id': '257107153551000',
+                'ext': 'mp4',
+                'title': 'Blind Getrouwd afl.6',
+                'description': 'md5:9ed26e8486ad0c1ddade4b066d83100f',
+            },
+            'skip_download': True,
+        },
+        {
+            'url': 'http://vtm.be/video?aid=163157',
+            'info_dict': {
+                'id': '257068686507000',
+                'ext': 'mp4',
+                'title': 'Amigo\'s afl.8',
+                'description': 'md5:1db8963594a1829839dabd55ef4f6a0b',
+            },
+            'skip_download': True,
+        },
     ]
 
     def _login(self):
@@ -68,24 +87,8 @@ class VTMIE(InfoExtractor):
 
         webpage = self._download_webpage(url, None)
 
-        # There are two known types of URLs both pointing to the same video.
-        # The type of URL you get depends on how you navigate to the video.
-        # Unfortunately, only one type of URL contains the actual video id:
-        #
-        # http://vtm.be/video/volledige-afleveringen/id/257059160170000
-        #
-        # The video id can also be something like:
-        # "vtm_20170213_VM0678492_vtmwatch"
-        #
-        # The other type of URL looks like:
-        #
-        # http://vtm.be/video?aid=166380
-        #
-        # The 'aid' is not the video id we are looking for. There can also be
-        # a lot of very long parameters in addition to the 'aid' parameter.
-        #
-        # So to work in all cases, we extract the video_id from the webpage
-        # instead of the URL.
+        # The URL sometimes contains the video id, but not always, e.g., test
+        # case 3. Fortunately, all webpages contain the video id.
         video_id = self._search_regex(
             r'\\"vodId\\":\\"(.+?)\\"', webpage, 'video_id', default=None)
 
--- a/youtube_dl/extractor/vtm.py
+++ b/youtube_dl/extractor/vtm.py
@@ -52,6 +52,16 @@ class VTMIE(InfoExtractor):
             },
             'skip_download': True,
         },
+        {
+            'url': 'http://www.q2.be/video/volledige-afleveringen/id/2be_20170301_VM0684442_q2',
+            'info_dict': {
+                'id': '2be_20170301_VM0684442_q2',
+                'ext': 'mp4',
+                'title': 'Modern Family afl. 15',
+                'description': 'md5:9866dd440480ee5ad5195382dd33d068',
+            },
+            'skip_download': True,
+        },
     ]
 
     def _login(self):
--- a/youtube_dl/extractor/vtm.py
+++ b/youtube_dl/extractor/vtm.py
@@ -34,33 +34,23 @@ class VTMIE(InfoExtractor):
         },
         {
             'url': 'http://vtm.be/video/volledige-afleveringen/id/257107153551000',
-            'info_dict': {
-                'id': '257107153551000',
-                'ext': 'mp4',
-                'title': 'Blind Getrouwd afl.6',
-                'description': 'md5:9ed26e8486ad0c1ddade4b066d83100f',
-            },
-            'skip_download': True,
+            'only_matching': True,
         },
         {
             'url': 'http://vtm.be/video?aid=163157',
-            'info_dict': {
-                'id': '257068686507000',
-                'ext': 'mp4',
-                'title': 'Amigo\'s afl.8',
-                'description': 'md5:1db8963594a1829839dabd55ef4f6a0b',
-            },
-            'skip_download': True,
+            'only_matching': True,
         },
         {
             'url': 'http://www.q2.be/video/volledige-afleveringen/id/2be_20170301_VM0684442_q2',
+            'only_matching': True,
+        },
+        {
+            'url': 'http://vtm.be/video?aid=168332',
             'info_dict': {
-                'id': '2be_20170301_VM0684442_q2',
+                'id': 'video?aid=168332',
                 'ext': 'mp4',
-                'title': 'Modern Family afl. 15',
-                'description': 'md5:9866dd440480ee5ad5195382dd33d068',
+                'title': 'Videozone',
             },
-            'skip_download': True,
         },
     ]
 
--- a/youtube_dl/extractor/vtm.py
+++ b/youtube_dl/extractor/vtm.py
@@ -96,7 +96,8 @@ class VTMIE(InfoExtractor):
 
         self._login()
 
-        title = remove_end(self._og_search_title(webpage), ' - Volledige Afleveringen')
+        title = self._html_search_regex(
+            r'\\"title\\":\\"(.+?)\\"', webpage, 'title', default=None)
 
         description = self._html_search_regex(
             r'<div[^>]+class="field-item\s+even">\s*<p>(.+?)</p>',
--- a/youtube_dl/extractor/vtm.py
+++ b/youtube_dl/extractor/vtm.py
@@ -1,0 +1,126 @@
+from __future__ import unicode_literals
+
+import re
+
+from .generic import GenericIE
+from .common import InfoExtractor
+from ..utils import (
+    urlencode_postdata,
+    compat_urllib_parse_urlencode,
+    ExtractorError,
+    remove_end,
+)
+
+
+class VTMIE(InfoExtractor):
+    """Download full episodes that require an account from vtm.be or q2.be.
+
+    The generic extractor can be used to download clips that do no require an
+    account.
+    """
+    _VALID_URL = r'https?://(?:www\.)?(?P<site_id>vtm|q2)\.be/video[/?].+?'
+    _NETRC_MACHINE = 'vtm'
+    _APIKEY = '3_HZ0FtkMW_gOyKlqQzW5_0FHRC7Nd5XpXJZcDdXY4pk5eES2ZWmejRW5egwVm4ug-'
+    _TESTS = [
+        {
+            'url': 'http://vtm.be/video/volledige-afleveringen/id/vtm_20170219_VM0678361_vtmwatch',
+            'md5': '9357926318cdfcb10079d85d56f4e04e',
+            'info_dict': {
+                'id': 'vtm_20170219_VM0678361_vtmwatch',
+                'ext': 'mp4',
+                'title': 'Allemaal Chris afl. 6',
+                'description': 'md5:4be86427521e7b07e0adb0c9c554ddb2',
+            },
+            'skip_download': True,
+        }
+    ]
+
+    def _login(self):
+        (username, password) = self._get_login_info()
+        if username is None or password is None:
+            self.raise_login_required()
+
+        auth_data = {
+            'APIKey': self._APIKEY,
+            'sdk': 'js_6.1',
+            'format': 'json',
+            'loginID': username,
+            'password': password,
+        }
+
+        auth_info = self._download_json(
+            'https://accounts.eu1.gigya.com/accounts.login', None,
+            note='Logging in', errnote='Unable to log in',
+            data=urlencode_postdata(auth_data), fatal=True)
+
+        error_message = auth_info.get('errorDetails')
+        if error_message:
+            raise ExtractorError(
+                'Unable to login: %s' % error_message, expected=True)
+
+        self._uid = auth_info['UID']
+        self._uid_signature = auth_info['UIDSignature']
+        self._signature_timestamp = auth_info['signatureTimestamp']
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        site_id = mobj.group('site_id')
+
+        webpage = self._download_webpage(url, None)
+
+        # There are two known types of URLs both pointing to the same video.
+        # The type of URL you get depends on how you navigate to the video.
+        # Unfortunately, only one type of URL contains the actual video id:
+        #
+        # http://vtm.be/video/volledige-afleveringen/id/257059160170000
+        #
+        # The video id can also be something like:
+        # "vtm_20170213_VM0678492_vtmwatch"
+        #
+        # The other type of URL looks like:
+        #
+        # http://vtm.be/video?aid=166380
+        #
+        # The 'aid' is not the video id we are looking for. There can also be
+        # a lot of very long parameters in addition to the 'aid' parameter.
+        #
+        # So to work in all cases, we extract the video_id from the webpage
+        # instead of the URL.
+        video_id = self._search_regex(
+            r'\\"vodId\\":\\"(.+?)\\"', webpage, 'video_id', default=None)
+
+        # When no video_id is found, it was most likely a video not requiring
+        # authentication, so just fall back to the generic extractor
+        if not video_id:
+            return self.url_result(url, 'Generic')
+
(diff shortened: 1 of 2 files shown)


mrBliss:
As the stream is encrypted with AES, the `pycrypto` package is required for this to work. When this package is not available `youtube_dl/downloader/hls.py` will fall back to ffmpeg, which fails because it can't handle AES encrypted segments.

It think it would be more sensible for this extractor to report to the user that (when missing) the `pycrypto` package is required for this extractor to work. This will save us from false bug reports from users using this extracting without the `pycrypto` package.

MichaelDeBoey:
Thanks for handling this @mrBliss :-)
Q2-video's work the same way, the only difference is the `app_id` which is `q2` instead of `vtm_watch`

MichaelDeBoey:
@dstftw I think this can be merged :-)

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,125 @@
...
+
+    The generic extractor can be used to download clips that do no require an
+    account.
+    """
+    _VALID_URL = r'https?://(?:www\.)?(?P<site_id>vtm|q2)\.be/video[/?].+?'
+    _NETRC_MACHINE = 'vtm'
+    _APIKEY = '3_HZ0FtkMW_gOyKlqQzW5_0FHRC7Nd5XpXJZcDdXY4pk5eES2ZWmejRW5egwVm4ug-'
+    _TESTS = [
+        {
+            'url': 'http://vtm.be/video/volledige-afleveringen/id/257085630173000',
```
dstftw: Video is gone.

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,125 @@
...
+                'description': 'md5:6526d6203a76bc572fe971211006bb70',
+            },
+            'skip': 'This video is only available for registered users'
+        }
+    ]
+
+    def _login(self):
+        (username, password) = self._get_login_info()
+        if username is None or password is None:
+            self.raise_login_required()
```
dstftw: Breaks downloading of videos that does not require authentication.

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,125 @@
...
+        m3u8_data = {
+            'app_id': 'vtm_watch' if site_id == 'vtm' else 'q2',
+            'user_network': 'vtm-sso',
+            'UID': self._uid,
+            'UIDSignature': self._uid_signature,
+            'signatureTimestamp': self._signature_timestamp,
+        }
+
+        data_url = "http://vod.medialaan.io/api/1.0/item/%s/video?%s" % (
+            video_id, compat_urllib_parse_urlencode(m3u8_data))
```
dstftw: With your code `video_id` is allowed to be `None`. This is broken in this case.

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,125 @@
...
+        data = self._download_json(data_url, video_id)
+
+        formats = self._extract_m3u8_formats(
+            data["response"]["uri"], video_id, entry_protocol='m3u8_native',
```
dstftw: Use single quotes consistently.

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,125 @@
...
+            ext='mp4', preference=0, fatal=True, m3u8_id='HLS')
```
dstftw: 1. `preference` is pointless.
2. `fatal=True` is the default.
3. Should be lowercase `hls`.

dstftw:
1. Fix issues mentioned in review.
2. Fix merge conflicts.
3. Squash commits.
4. Provide account credentials for testing.

mrBliss:
> 4. Provide account credentials for testing.

Do you mean I have to make an account for _you_ to run the tests?

youtube_dl/extractor/vtm.py
mrBliss: Oops, I just copy-pasted this from another extractor without thinking.

youtube_dl/extractor/vtm.py
mrBliss: I'll remove the default argument to the `_search_regex` call that extracts the `video_id`.

youtube_dl/extractor/vtm.py
mrBliss: No, because videos not requiring authentication are not matched by the URL regex as they don't have '/video' in the URL. So youtube-dl falls back to the generic extractor which works fine for those videos. I'll add a comment.

youtube_dl/extractor/vtm.py
mrBliss: Ok, I'll replace it with another video. That video will go offline after a while too as all member-only videos do, but at least you'll be able to test it.

mrBliss:
When running the (updated) test, I get: 
```
...
  File "blabla/youtube-dl/youtube_dl/downloader/hls.py", line 149, in real_download
    decrypt_info['KEY'], AES.MODE_CBC, iv).decrypt(frag_content)
  File "/usr/lib/python3.6/site-packages/Crypto/Cipher/blockalgo.py", line 295, in decrypt
    return self._cipher.decrypt(ciphertext)
ValueError: Input strings must be a multiple of 16 in length
```
I suppose the length of the test fragment is not a multiple of 16 which is needed for AES decryption.
Shouldn't all other extractor tests that use AES decryption run into this error too? How should I proceed? Fix `hls.py` so it pads fragments to a length that is a multiple of 16?

youtube_dl/extractor/vtm.py
dstftw: Wrong: http://vtm.be/video?aid=168332.

dstftw:
>Do you mean I have to make an account for you to run the tests?

You have to provide working credentials. It does not matter will it be a new account's or yours.

>When running the (updated) test, I get: ...

Just add `skip_download` param for a test.

mrBliss:
`You have to provide working credentials. It does not matter will it be a new account's or yours.`

Ok, I made a test account, should I just post the credentials here, publicly, or email it to you?

Thank you for being patient with me.

dstftw:
Either way is fine. If it's possible to disable ability to change password it also may be included in test.

mrBliss:
You can reset the password via e-mail, so I'll send you the credentials via e-mail.

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,133 @@
...
+        if not video_id:
+            fallback = GenericIE()
+            fallback.set_downloader(self._downloader)
+            return fallback._real_extract(url)
+
+        self._login()
+
+
+        # Remove the string ' - Volledige Afleveringen' from the end of the
+        # title (Dutch for "Complete Episodes")
```
dstftw: This comment is pointless.

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,133 @@
...
+        title = self._og_search_title(webpage).replace(
+            ' - Volledige Afleveringen', '')
```
dstftw: Use `utils.remove_end`.

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,133 @@
...
+
+        m3u8_data = {
+            'app_id': 'vtm_watch' if site_id == 'vtm' else 'q2',
+            'user_network': 'vtm-sso',
+            'UID': self._uid,
+            'UIDSignature': self._uid_signature,
+            'signatureTimestamp': self._signature_timestamp,
+        }
+
+        data_url = "http://vod.medialaan.io/api/1.0/item/%s/video?%s" % (
```
dstftw: Single quotes.

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,133 @@
...
+            video_id, compat_urllib_parse_urlencode(m3u8_data))
```
dstftw: Pass `m3u8_data` as `query` to download method.

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,133 @@
...
+            data=urlencode_postdata(auth_data), fatal=True)
+
+        error_message = auth_info.get('errorDetails')
+        if error_message:
+            raise ExtractorError(
+                'Unable to login: %s' % error_message, expected=True)
+
+        self._uid = auth_info.get('UID')
+        self._uid_signature = auth_info.get('UIDSignature')
+        self._signature_timestamp = auth_info.get('signatureTimestamp')
```
dstftw: These looks like mandatory fields.

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,133 @@
...
+        #
+        # The other type of URL looks like:
+        #
+        # http://vtm.be/video?aid=166380
+        #
+        # The 'aid' is not the video id we are looking for. There can also be
+        # a lot of very long parameters in addition to the 'aid' parameter.
+        #
+        # So to work in all cases, we extract the video_id from the webpage
+        # instead of the URL.
```
dstftw: This should be shortened to one sentence. Example URLs should be added as tests.

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,133 @@
...
+        video_id = self._search_regex(
+            r'\\"vodId\\":\\"(.+?)\\"', webpage, 'video_id', default=None)
+
+        # When no video_id is found, it was most likely a video not requiring
+        # authentication, so just fall back to the generic extractor
```
dstftw: This comment is pointless.

MichaelDeBoey:
Hopefully you can work this out @mrBliss :-)

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,133 @@
...
+        if not video_id:
+            fallback = GenericIE()
+            fallback.set_downloader(self._downloader)
+            return fallback._real_extract(url)
```
mrBliss: Thanks, I was looking for this.

youtube_dl/extractor/vtm.py
mrBliss: Ok. Maybe you can help me with the following: when I add multiple tests, only the first one seems to be run. I use `python test/test_download.py TestDownload.test_VTM` to run the tests. When I try running the tests of another extractor, e.g. `Youtube`, the same happens.

mrBliss: I just amended my commit to address most of your comments. I haven't fixed this comment yet, but I will as soon as I can get the multiple tests to work correctly.

youtube_dl/extractor/vtm.py
mrBliss: I disagree, certainly now that the `if`-clause just contains `return self.url_result(url, 'Generic')`. 

To me it is not obvious what it means (it was slightly clearer with the previous code, but that doesn't matter). More importantly, it is not obvious _why_ we fall back to the generic extractor. Without the comment, it seems as if we fall back when the extractor fails, as if it were a last resort in case the extractor is broken because the website changed. Which is not true, the check is deliberate because when no `video_id` is found, it is (most likely) a video not requiring authentication.

youtube_dl/extractor/vtm.py
dstftw: See https://github.com/rg3/youtube-dl#adding-support-for-a-new-site, point 6.

mrBliss: It's already `_TESTS` and also a list of dictionaries. This also wouldn't explain why I run into the same problem with other extractors, e.g. `Youtube`, `ABC`, ...

dstftw: Did you even read what I've posted?

mrBliss: Sorry, my bad, I stopped reading to soon.

mrBliss:
I have addressed your comments. I'll just squash the commits when you say everything is ok.

youtube_dl/extractor/vtm.py
MichaelDeBoey: @mrBliss you have checked this also?

mrBliss: @MichaelDeBoey Yes, `_login` is no longer called by default, but only when needed. I had typed a comment acknowledging that dstfw is right, but I guess I forgot to submit it.

MichaelDeBoey: @mrBliss OK, just to make sure everything is good :-) I'm really waiting for this one for a long time so... ;-)

MichaelDeBoey:
I think @mrBliss has done everything you asked for then @dstftw :-)

MichaelDeBoey:
Travis CI failed, but maybe you can check it out @dstftw?

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,140 @@
...
+        self._signature_timestamp = auth_info['signatureTimestamp']
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        site_id = mobj.group('site_id')
+
+        webpage = self._download_webpage(url, None)
+
+        # The URL sometimes contains the video id, but not always, e.g., test
+        # case 3. Fortunately, all webpages contain the video id.
```
dstftw: This comment contradicts the comment at 105-106. You claim video id is always available and provide a fallback to generic extractor at the same time.

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,140 @@
...
+
+        self._uid = auth_info['UID']
+        self._uid_signature = auth_info['UIDSignature']
+        self._signature_timestamp = auth_info['signatureTimestamp']
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        site_id = mobj.group('site_id')
+
+        webpage = self._download_webpage(url, None)
```
dstftw: Some `display_id` should be extracted from `url` and shown instead of `None`.

youtube_dl/extractor/vtm.py
dstftw: You can leave note about video not requiring authentication but you should not repeat the code itself:
>When no video_id is found

>so just fall back to the generic extractor

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,140 @@
...
+        # case 3. Fortunately, all webpages contain the video id.
+        video_id = self._search_regex(
+            r'\\"vodId\\":\\"(.+?)\\"', webpage, 'video_id', default=None)
+
+        # When no video_id is found, it was most likely a video not requiring
+        # authentication, so just fall back to the generic extractor
+        if not video_id:
+            return self.url_result(url, 'Generic')
+
+        self._login()
```
dstftw: If n URLs are downloaded with single command login will take place n times. You should store login flag or use `_real_initialize`.

youtube_dl/extractor/vtm.py
mrBliss: There is no single form of id that all URLs contain, so instead of trying to handle the multiple kinds of ids just to print something, wouldn't it be ok to just print "Downloading webpage", by passing `url, None, "Downloading webpage"` as arguments?

youtube_dl/extractor/vtm.py
```diff
@@ -0,0 +1,140 @@
...
+    _TESTS = [
+        {
+            'url': 'http://vtm.be/video/volledige-afleveringen/id/vtm_20170219_VM0678361_vtmwatch',
+            'info_dict': {
+                'id': 'vtm_20170219_VM0678361_vtmwatch',
+                'ext': 'mp4',
+                'title': 'Allemaal Chris afl. 6',
+                'description': 'md5:4be86427521e7b07e0adb0c9c554ddb2',
+            },
+            'skip_download': True,
```
mrBliss: Just to be clear: with the "same extraction scenario" you mean tests 1-2 or 1-3 (or even 1-4)? Say you mean 1-3, I suppose I should make tests 2-3 into `only_matching` tests, right?

dstftw: There are two scenarios: video without auth and video with auth. The rest are matching only.

youtube_dl/extractor/vtm.py
```diff
@@ -34,33 +34,23 @@ class VTMIE(InfoExtractor):
...
             'url': 'http://vtm.be/video/volledige-afleveringen/id/257107153551000',
-            'info_dict': {
-                'id': '257107153551000',
-                'ext': 'mp4',
-                'title': 'Blind Getrouwd afl.6',
-                'description': 'md5:9ed26e8486ad0c1ddade4b066d83100f',
-            },
-            'skip_download': True,
+            'only_matching': True,
         },
```
mrBliss: After grepping the `test` dir, `only_matching` seems to be only used in `test_no_duplicates`. At first sight, the actual extraction is no longer tested, which is much more important. Could you please clarify?

youtube_dl/extractor/vtm.py
```diff
@@ -34,33 +34,23 @@ class VTMIE(InfoExtractor):
...
+        },
+        {
+            'url': 'http://vtm.be/video?aid=168332',
             'info_dict': {
-                'id': '2be_20170301_VM0684442_q2',
+                'id': 'video?aid=168332',
                 'ext': 'mp4',
-                'title': 'Modern Family afl. 15',
-                'description': 'md5:9866dd440480ee5ad5195382dd33d068',
+                'title': 'Videozone',
```
mrBliss: I would like this to be the actual title of the video. I tried passing the title to `self.url_result` and even manually setting it on the dictionary returned, but neither approach works, as the returned result is not the final result.

youtube_dl/extractor/vtm.py
dstftw: `only_matching` is only matched against `_VALID_URL`, such tests are for reference purposes. There is no sense in several similar tests that cover one scenario.

youtube_dl/extractor/vtm.py
dstftw: https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/fxnetworks.py#L62-L70

youtube_dl/extractor/vtm.py
mrBliss: Ok, thanks.

youtube_dl/extractor/vtm.py
mrBliss: Ah, I now understand how delegation to another extractor works. However, it seems that `title` and any other custom fields I add are still discarded. Maybe it has something to do with the fact that the generic extractor thinks the page is a playlist?

dstftw: Yes. This is due to a playlist.

MichaelDeBoey:
Any progress on this @mrBliss @dstftw?

youtube_dl/extractor/vtm.py
mrBliss: For me it's fine this way.

mrBliss:
Thanks dstftw!

MichaelDeBoey:
Thanks @dstftw for imbedding this one and offcourse a great thank you to @mrBliss for all the hard work! :-)

johnhawkinson:
@dstftw, Commiting this broke Travis's `YTDL_TEST_SET=core` tests https://travis-ci.org/rg3/youtube-dl/builds/212310040, https://travis-ci.org/rg3/youtube-dl/jobs/212310041:
```

FAIL: test_no_duplicates (test.test_all_urls.TestAllURLsMatching)
  File "/home/travis/build/rg3/youtube-dl/test/test_all_urls.py", line 94, in test_no_duplicates
    self.assertTrue(ie.suitable(url), '%s should match URL %r' % (type(ie).__name__, url))
AssertionError: MedialaanIE should match URL u'http://vitaya.be/de-jurk/precies-je-hebt-geen-borsten'
Ran 193 tests in 6.432s
FAILED (failures=1)
```

dstftw:
@johnhawkinson fixed in f68ef1e2abd876ffca65544fd3e42756f9c33be3.
@mrBliss do you want to be credited as one of the authors of youtube-dl? If so under what name?

MichaelDeBoey:
@dstftw & @johnhawkinson Vitaya is a Medialaan channel tho, so I think they use the same script for showing video's

mrBliss:
@dstftw yes please, as "Thomas Winant".

midas02:
@dstftw Can you double-check the code? I came across some links which make youtube-dl crash (so not a normal programme exit).

> C:\>youtube-dl.exe "http://vtmkzoom.be/video/?f[0]=sm_field_video_origin_cms_longform%3AVolledige%20afleveringen&aid=45365"
[Medialaan] 45365: Downloading webpage
  File "__main__.py", line 19, in <module>
  File "C:\Users\dst\AppData\Roaming\Build archive\youtubedl\rg3\tmpje2_cfdl\build\youtube_dl\__init__.py", line 464, in main  
File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\rg3\tmpje2_cfdl\build\youtube_dl\__init__.py", line 454, in _real_main  
File "C:\Users\dst\AppData\Roaming\Build archive\youtubedl\rg3\tmpje2_cfdl\build\youtube_dl\YoutubeDL.py", line 1884, in download
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\rg3\tmpje2_cfdl\build\youtube_dl\YoutubeDL.py", line 772, in extract_info
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\rg3\tmpje2_cfdl\build\youtube_dl\YoutubeDL.py", line 826, in process_ie_result
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\rg3\tmpje2_cfdl\build\youtube_dl\YoutubeDL.py", line 1461, in process_video_result
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\rg3\tmpje2_cfdl\build\youtube_dl\YoutubeDL.py", line 1311, in _calc_headers
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\rg3\tmpje2_cfdl\build\youtube_dl\YoutubeDL.py", line 1323,
 in _calc_cookies  File "C:\Users\dst\AppData\Roaming\Build archive\youtubedl\rg3\tmpje2_cfdl\build\youtube_dl\utils.py", line 540, in sanitized_Request
  File "C:\Python\Python34\lib\urllib\request.py", line 267, in __init__
  File "C:\Python\Python34\lib\urllib\request.py", line 293, in full_url
  File "C:\Python\Python34\lib\urllib\request.py", line 322, in _parse
ValueError: unknown url type: 'http'

VTMKzoom.be uses different link styles. I've assembled a couple of them:
CRASH http://vtmkzoom.be/video/?f[0]=sm_field_video_origin_cms_longform%3AVolledige%20afleveringen&aid=45365
CRASH http://vtmkzoom.be/video/?aid=45425&f[0]=sm_field_video_origin_cms_longform%3AVolledige%20afleveringen
CRASH http://vtmkzoom.be/video?aid=45425
OK http://vtmkzoom.be/video/volledige-afleveringen/id/257105164068000
OK http://vtmkzoom.be/k3-dansstudio/warm-mee-op-met-hanne
OK http://vtmkzoom.be/bab-backstage/segway-singing-met-joost

PR #11931, opened 2017-02-01
update extra '\' in pattern

Justsoos:
http://www.douyu.com/t/lpl
`this room source get extra '\' with "room_id\" (from js coding)`


youtube_dl/extractor/douyutv.py
```diff
@@ -82,7 +82,7 @@ def _real_extract(self, url):
         else:
             page = self._download_webpage(url, video_id)
             room_id = self._html_search_regex(
-                r'"room_id"\s*:\s*(\d+),', page, 'room id')
+                r'"room_id\\*"\s*:\s*(\d+),', page, 'room id')
```
dstftw: Add a test.

Issue #11953, opened 2017-02-03
iqiyi not downloading properly


Stepin2:
Log is as follows:
```
$ ./youtube-dl http://www.iqiyi.com/a_19rrhanifd.html#vfrm=2-4-0-1 --verbose
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'http://www.iqiyi.com/a_19rrhanifd.html#vfrm=2-4-0-1', u'--verbose']
[debug] Encodings: locale ISO-8859-1, fs ISO-8859-1, out ISO-8859-1, pref ISO-8859-1
[debug] youtube-dl version 2017.02.01
[debug] Python version 2.7.3 - Linux-2.6.32_1-18-0-0-x86_64-with-redhat-4.3-Final
[debug] exe versions: none
[debug] Proxy map: {}
[iqiyi] temp_id: download video page
ERROR: Unable to extract tvid; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "./youtube-dl/youtube_dl/YoutubeDL.py", line 694, in extract_info
    ie_result = ie.extract(url)
  File "./youtube-dl/youtube_dl/extractor/common.py", line 369, in extract
    return self._real_extract(url)
  File "./youtube-dl/youtube_dl/extractor/iqiyi.py", line 345, in _real_extract
    r'data-player-tvid\s*=\s*[\'"](\d+)', webpage, 'tvid')
  File "./youtube-dl/youtube_dl/extractor/common.py", line 672, in _search_regex
    raise RegexNotFoundError('Unable to extract %s' % _name)
RegexNotFoundError: Unable to extract tvid; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

ThomasChr:
I can confirm the problem. It seems to be that the extractor is assuming the URL is a Video-URL - but this is not the case. The url seems to be a Playlist-URL.
(It's a little bit tricky for me, because I don't speak any chinese.)
Trying to download the video URL ( http://www.iqiyi.com/v_19rramot48.html ) seems to result in just the opposite - the extractor thinks its a playlist and fails.

Stepin2:
[...]

PR #11956, opened 2017-02-03
Fix for Issue #11953

ThomasChr:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Fix for #11953 - had problems with iqiyi giving me only empty json with code A00004 in it. Not sure if it's some kind of blocking/error, or simply the "new way" in saying: There are no more items on the playlist.
Also possbile that this kind of playlist fetching is not supported anymore.
All tests I had runned fine. But checking for the size of the playlist (and checking if youtube-dl fetches the whole playlist) is not possible for me: I can't read chinese!

### DIFF
--- a/youtube_dl/extractor/iqiyi.py
+++ b/youtube_dl/extractor/iqiyi.py
@@ -302,7 +302,7 @@ class IqiyiIE(InfoExtractor):
         PAGE_SIZE = 50
 
         links = re.findall(
-            r'<a[^>]+class="site-piclist_pic_link"[^>]+href="(http://www\.iqiyi\.com/.+\.html)"',
+            r'<a href="(http://www\.iqiyi\.com/.+\.html)"[^>]+?class="site-piclist_pic_link"[^>]+?>',
             webpage)
         if not links:
             return
--- a/youtube_dl/extractor/iqiyi.py
+++ b/youtube_dl/extractor/iqiyi.py
@@ -323,10 +323,13 @@ class IqiyiIE(InfoExtractor):
                 errnote='Failed to download playlist page %d' % page_num)
             pagelist = self._parse_json(
                 remove_start(pagelist_page, 'var tvInfoJs='), album_id)
-            vlist = pagelist['data']['vlist']
-            for item in vlist:
-                entries.append(self.url_result(item['vurl']))
-            if len(vlist) < PAGE_SIZE:
+            if 'data' in pagelist:
+                vlist = pagelist['data']['vlist']
+                for item in vlist:
+                    entries.append(self.url_result(item['vurl']))
+                if len(vlist) < PAGE_SIZE:
+                    break
+            else:
                 break
 
         return self.playlist_result(entries, album_id, album_title)
--- a/youtube_dl/extractor/iqiyi.py
+++ b/youtube_dl/extractor/iqiyi.py
@@ -302,7 +302,7 @@ class IqiyiIE(InfoExtractor):
         PAGE_SIZE = 50
 
         links = re.findall(
-            r'<a href="(http://www\.iqiyi\.com/.+\.html)"[^>]+?class="site-piclist_pic_link"[^>]+?>',
+            r'<a[^>]+?href="(http://www\.iqiyi\.com/.+\.html)"[^>]+?class="site-piclist_pic_link"[^>]+?>',
             webpage)
         if not links:
             return


youtube_dl/extractor/iqiyi.py
```diff
@@ -302,7 +302,7 @@ def _extract_playlist(self, webpage):
         PAGE_SIZE = 50
 
         links = re.findall(
-            r'<a[^>]+class="site-piclist_pic_link"[^>]+href="(http://www\.iqiyi\.com/.+\.html)"',
+            r'<a[^>]+?href="(http://www\.iqiyi\.com/.+\.html)"[^>]+?class="site-piclist_pic_link"[^>]+?>',
```
dstftw: `?` makes no sense here. Arbitrary attribute position should be supported in case this is dynamic.

ThomasChr: @dstftw 
I removed the ? - I thought maybe lazy matching would be better.
This regex now should capture the URL in all cases where 'href' ist before 'sit-piclist_pic_link'. If it's the other way around (making "href" not the first <a>-Attribute is valid but seldom used) it would fail.

I suppose I can make the regex so that the other way around is also matched - but this would make it more complex - do you think thats a good idea?

dstftw: The most simple way is to use `utils.extract_attributes`.

ThomasChr: I'll read through it and try! Thanks!

youtube_dl/extractor/iqiyi.py
```diff
@@ -301,10 +302,14 @@ def get_raw_data(self, tvid, video_id):
     def _extract_playlist(self, webpage):
         PAGE_SIZE = 50
 
-        links = re.findall(
-            r'<a[^>]+class="site-piclist_pic_link"[^>]+href="(http://www\.iqiyi\.com/.+\.html)"',
-            webpage)
-        if not links:
+        links = []
+        for link in re.findall(r'<a[^>]+class="[^"]*site-piclist_pic_link[^"]*"[^>]*>', webpage):
```
yan12125: It would be better to extend ```utils.get_element_by_class``` and make it return multiple results

ThomasChr: @yan12125 Good idea! I'll see to it.

ThomasChr: @yan12125 Any hints about how I should modify get_element_by_class without breaking anything?
a) Add a new parameter to get_element_by_class() 'multiple=false' which changes the behavior of get_element_by_class?
b) Add a new function get_element_by_class_multiple() / get_elements_by_class() ? (thats no big change because get_element_by_class() itself relies solely on get_element_by_attribute.)

I have searched through the code but I don't see any example function which returns only one element and changes it's behavior to returning multiple when the caller wants it.

Of course there is the third option of simply changing get_element_by_class() to return multiple resulsts, but:
1. We have to make sure that all calls to it are working fine afterwards (about 9 calls).
2. The name of the function is then wrong because it returns not only one result anymore.

And when I'm up to it, what's with:
- get_element_by_id()
- get_element_by_attribute()

All of them can return multiple resulsts :-(

yan12125: This is my idea:

1. get_element_by_id is not touched (by HTML standards IDs are unique)
2. get_element_by_attribute => get_elements_by_attribute
3. get_element_by_class => get_elements_by_class

And replace all usages in extractors. It's better to open a new pull request for this change.

ThomasChr: 1. I'm not really sure if all webpages confirm to the standard. Also it's maybe better to make even the ones which return only single results to return lists...

yan12125: Javascript (HTML DOM API) has ```getElementById``` only. If a webpage has multiple elements with the same ID, website developers have to do parsing stuffs if they want to access all elements of that ID. It would be suprising to me if there's really such a site.

Also, it's better to make function in utils match Javascript.

ThomasChr: @yan12125 Okay!

My suggestion is that we incoperate the changes in this pull request so that the Extractor is working again and we can close the issue, even though it's not using the new function get_elements_by_class.

And afterwards I'm building the "new" functions as you suggested and making all needed changes in a new pull request.

ThomasChr: @yan12125 I've created PR #12034

ThomasChr:
Hello @yan12125 
I can't use get_element_by_class/get_elements_by_class here because I don't want the contents of the element. I want the other attributes!
We have:
`<a href="xxx" class="site">My Link</a>`

I want:
`xxx`

By using get_elements_by_class I get:
`My Link`

So I don't think that the function would be beneficial here. The PR has also been checked by the creator of the issue and is working. If you haven't got any more changes to my code could you please commit this PR?

yan12125:
Break existing test case:
```
$ youtube-dl -vF "http://www.iqiyi.com/a_19rrhb8ce1.html"
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-vF', 'http://www.iqiyi.com/a_19rrhb8ce1.html']
[debug] youtube-dl version 2017.02.11
[debug] Git HEAD: 68c22c4c1
[debug] Python version 3.6.0 - Linux-4.9.6-1-ARCH-x86_64-with-arch
[debug] exe versions: ffmpeg 3.2.2, ffprobe 3.2.2
[debug] Proxy map: {}
[iqiyi] temp_id: download video page
ERROR: Unable to extract tvid; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
  File "/home/yen/Projects/youtube-dl/youtube_dl/YoutubeDL.py", line 696, in extract_info
    ie_result = ie.extract(url)
  File "/home/yen/Projects/youtube-dl/youtube_dl/extractor/common.py", line 369, in extract
  File "/home/yen/Projects/youtube-dl/youtube_dl/extractor/iqiyi.py", line 354, in _real_extract
    r'data-player-tvid\s*=\s*[\'"](\d+)', webpage, 'tvid')
  File "/home/yen/Projects/youtube-dl/youtube_dl/extractor/common.py", line 672, in _search_regex
youtube_dl.utils.RegexNotFoundError: Unable to extract tvid; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

> So I don't think that the function would be beneficial here

It would be better to have some new functions return the whole element. Not a required change, though.

ThomasChr:
@yan12125 Yeah, a few test cases are broken. The problems are on another part of the Extractor. I'll see if I can fix these problems too.

OmerJog:
What is the status of this PR?

aaronliu0130:
@ThomasChr What's the progress? Getting a little impatient here.

aaronliu0130:
Hey! Why'd the branch/repo get deleted?

PR #11969, opened 2017-02-04
[Piksel] app token in <video> (uscourts.gov)

johnhawkinson:
Support alternative api token found as
  <video ... data-de-api-key="bd1863e1-a59e-11e3-bfa8-005056865f49" ...>
in addition to extraction from the flashvars JSON.

This supports the player as invoked from uscourts.gov, e.g. [http://www.uscourts.gov/cameras-courts/state-washington-vs-donald-j-trump-et-al](http://www.uscourts.gov/cameras-courts/state-washington-vs-donald-j-trump-et-al)

Add test case for same.

### DIFF
--- a/youtube_dl/extractor/piksel.py
+++ b/youtube_dl/extractor/piksel.py
@@ -30,7 +30,8 @@ class PikselIE(InfoExtractor):
             }
         },
         {
-            'url': 'http://www.uscourts.gov/cameras-courts/state-washington-vs-donald-j-trump-et-al',
+            # Original source: http://www.uscourts.gov/cameras-courts/state-washington-vs-donald-j-trump-et-al
+            'url': 'https://player.piksel.com/v/v80kqp41',
             'md5': '753ddcd8cc8e4fa2dda4b7be0e77744d',
             'info_dict': {
                 'id': 'v80kqp41',
--- a/youtube_dl/extractor/piksel.py
+++ b/youtube_dl/extractor/piksel.py
@@ -16,18 +16,32 @@ from ..utils import (
 
 class PikselIE(InfoExtractor):
     _VALID_URL = r'https?://player\.piksel\.com/v/(?P<id>[a-z0-9]+)'
-    _TEST = {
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/piksel.py
```diff
@@ -16,18 +16,32 @@
...
+                'id': 'nv60p12f',
+                'ext': 'mp4',
+                'title': 'فن الحياة  - الحلقة 1',
+                'description': 'احدث برامج الداعية الاسلامي " مصطفي حسني " فى رمضان 2016علي النهار نور',
+                'timestamp': 1465231790,
+                'upload_date': '20160606',
+            }
+        },
+        {
+            'url': 'http://www.uscourts.gov/cameras-courts/state-washington-vs-donald-j-trump-et-al',
```
dstftw: All tests placed here must have URLs that match extractor's `_VALID_URL`. Tests for generic embeds should be placed into generic extractor.
```
PS C:\Dev\youtube-dl\master> py -3.5 .\test\test_all_urls.py
...F.............
  File ".\test\test_all_urls.py", line 94, in test_no_duplicates
    self.assertTrue(ie.suitable(url), '%s should match URL %r' % (type(ie).__name__, url))
AssertionError: False is not true : PikselIE should match URL 'http://www.uscourts.gov/cameras-courts/state-washington-vs-donald-j-trump-et-al'

Ran 17 tests in 8.066s

FAILED (failures=1)
```

johnhawkinson: Would it be better to put this test in the generic extractor, or to change this test to the URL that gets resolved by the generic extractor, [https://player.piksel.com/v/v80kqp41](https://player.piksel.com/v/v80kqp41)?

dstftw: I'd prefer the latter. Generic extractor is already too big.

johnhawkinson: Done. Thanks.

dstftw:
Thanks.

PR #11974, opened 2017-02-04
[twitch] Add basic support for two-factor authentication

puxlit:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This adds basic handling of two-factor authentication during login to the base Twitch extractor, in a manner similar to #3534.

Further work would be required to facilitate requesting of SMSes or phone calls, but I'd wager this isn't a common case, and can thus be deferred.

### DIFF
--- a/youtube_dl/extractor/twitch.py
+++ b/youtube_dl/extractor/twitch.py
@@ -12,7 +12,6 @@ from ..compat import (
     compat_str,
     compat_urllib_parse_urlencode,
     compat_urllib_parse_urlparse,
-    compat_urlparse,
 )
 from ..utils import (
     clean_html,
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/twitch.py
```diff
@@ -64,47 +64,62 @@ def fail(message):
...
-                response = self._parse_json(
-                    e.cause.read().decode('utf-8'), None)
-                fail(response['message'])
-            raise
-
-        if response.get('redirect'):
-            self._download_webpage(
-                response['redirect'], None, 'Downloading login redirect page',
-                headers=headers)
+        if not redirect_res:
```
dstftw: This is impossible outcome since `False` can only be returned when `self._download_*` call is not `fatal`, all calls in this code are `fatal`.

Issue #11221, opened 2016-11-17
indavideo.hu support


szunyi:
Hello,

Can you fix or add the the indavideo.hu site for download videos?
This site can be the biggest hungarian site.

This site use subdomains for each big category, so please support the subdomains too.

Example video for generic video: http://indavideo.hu/video/A_leggazabb_cipoproba
Another example for subdomain: http://auto.indavideo.hu/video/Ket_Nissan_torestesztje
And yes, please dont forget about erotic side too - NSFW link: http://erotika.indavideo.hu/video/MILF_11

Thanks!
Bye

PR #11989, opened 2017-02-06
[indavideo] Bugfix, support embedding

BagiraHun:

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature
---
- Bugfix - Issue #11221
- Site structure changed
- Code simplify
- Support embedded videos

### DIFF
--- a/youtube_dl/extractor/indavideo.py
+++ b/youtube_dl/extractor/indavideo.py
@@ -1,6 +1,8 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
+import re
+
 from .common import InfoExtractor
 from ..utils import (
     int_or_none,
@@ -9,8 +11,14 @@ from ..utils import (
 )
 
 
-class IndavideoEmbedIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:(?:embed\.)?indavideo\.hu/player/video/|assets\.indavideo\.hu/swf/player\.swf\?.*\b(?:v(?:ID|id))=)(?P<id>[\da-f]+)'
+class IndavideoIE(InfoExtractor):
+    _VALID_URL = r'''(?x)
+                    (?:https?:)//
+                    (?:
+                        (?:.+?\.)?indavideo\.hu/(?:player/)?video/|
+                        assets\.indavideo\.hu/swf/(?:inda)player\.swf\?.*\b(?:v(?:ID|id))=
+                    )
+                    (?P<id>[a-zA-Z0-9-_]+)'''
     _TESTS = [{
         'url': 'http://indavideo.hu/player/video/1bdc3c6d80/',
         'md5': 'f79b009c66194acacd40712a6778acfa',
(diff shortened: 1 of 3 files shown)


FelixSchwarz:
just a random reaction from IRC (without any authority on the project): how about splitting this into multiple commits? Each commit should do just one thing.

BagiraHun:
OK, we can split into two commits:
* bugfix: extractors.py and indavideo.py changes
* support emedded videos: generic.py changes

Bugfix includes also site strucure changes and code simplify, but this cannot split into multiple commits.

FelixSchwarz:
That sounds like a good plain - please note that I'm a completely outsider so I don't claim any authority. I'm just a user (with a bit of Python knowledge) who was frustrated about the number of open pull requests.

nyuszika7h:
Is anything going to happen with this?

BagiraHun:
Sadly no and I don't know how can I accelerate the code review process.

youtube_dl/extractor/indavideo.py
```diff
@@ -27,42 +35,79 @@ class IndavideoEmbedIE(InfoExtractor):
...
+        'only_matching': True,
+    }, {
+        'url': 'http://palyazat.indavideo.hu/video/Embertelen_dal_Dodgem_egyuttes',
+        'only_matching': True,
     }]
 
+    @staticmethod
+    def _extract_urls(webpage):
+        return [url for _, url in re.findall(
+            r'<(?:iframe[^>]+src|object[^>]+data)=(["\'])(?P<url>(?:https?:)?//(?:(?:.+?\.)?indavideo\.hu/(?:player/)?video/|assets\.indavideo\.hu/swf/(?:inda)?player\.swf\?.*\b(?:v(?:ID|id))=)[a-zA-Z0-9-_]+)(?:\?|&|\1)',
```
dstftw: Add test for video embedded with `<object data`.

youtube_dl/extractor/indavideo.py
```diff
@@ -9,8 +11,14 @@
 )
 
 
-class IndavideoEmbedIE(InfoExtractor):
```
BagiraHun: I've merged the two way of extractions (download and embedding). I think this general name (without embed) describes better the function. So I didn't chane the name, I've only deleted the Embeded function, merging its functionality to the other one.

dstftw: As said **don't touch** extractor's name. We don't need unnecessary download archive breakages.

youtube_dl/extractor/indavideo.py
```diff
@@ -27,42 +35,79 @@ class IndavideoEmbedIE(InfoExtractor):
...
+
     def _real_extract(self, url):
         video_id = self._match_id(url)
 
         video = self._download_json(
             'http://amfphp.indavideo.hu/SYm0json.php/player.playerHandler.getVideoData/%s' % video_id,
             video_id)['data']
 
-        title = video['title']
+        title = video.get('title')
```
BagiraHun: Absolutly correct. I can add '|| "No title"' to the end of the line instead of exception

dstftw: No, you don't. Title is mandatory meaning you must fail if you can't extract it. Read new extractor tutorial.

youtube_dl/extractor/indavideo.py
BagiraHun: I understand, but what's the point if I want to delete one of the functions because its function merged to the other one. In my opinion this is **not changing of the name**, because before the commit there was 2 extractors: IndavideoIE and IndavideoEmbedIE, with the commit I've deleted the IndavideoEmbedIE

dstftw: This is **changing name** because you've merged `IndavideoIE` into `IndavideoEmbedIE` (or simply dropped it) and then renamed it to `IndavideoIE` as a result you've **broken** all download archives made with previous versions of youtube-dl. Just do what is asked.

BagiraHun: I understand it. Due backward compability I added back the IndavidedoEmbedIE class.

youtube_dl/extractor/indavideo.py
```diff
@@ -27,42 +35,79 @@ class IndavideoEmbedIE(InfoExtractor):
...
+
+        formats = []
+        for video_url in video_urls:
+            _height = self._search_regex(
+                r'\.(\d{3,4})\.mp4(?:\?|$)', video_url, 'height', default=None)
+            _url = video_url
+            if filesh:
+                if _height in filesh:
+                    _url += '&' if '?' in _url else '?'
+                    _url += "token=" + filesh.get(_height)
```
BagiraHun: Thanx for the tip

Issue #12007, opened 2017-02-07
Videos from Pornhub is not possible to download


Modryz:
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2017.02.07**

- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---
```
# youtube-dl -v http://www.pornhub.com/view_video.php?viewkey=ph574f0d                                                                                        6f2d8fa
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'-v', u'http://www.pornhub.com/view_video.php?viewk                                                                                        ey=ph574f0d6f2d8fa']
[debug] Encodings: locale ANSI_X3.4-1968, fs ANSI_X3.4-1968, out ANSI_X3.4-1968,                                                                                         pref ANSI_X3.4-1968
[debug] youtube-dl version 2017.02.07
[debug] Python version 2.7.6 - Linux-3.16.6-042stab120.11-x86_64-with-Ubuntu-14.                                                                                        04-trusty
[debug] exe versions: none
[debug] Proxy map: {}
[PornHub] ph574f0d6f2d8fa: Downloading webpage
ERROR: No video formats found; please report this issue on https://yt-dl.org/bug                                                                                         . Make sure you are using the latest version; see  https://yt-dl.org/update  on                                                                                         how to update. Be sure to call youtube-dl with the --verbose flag and include i                                                                                        ts complete output.
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/youtube_dl/YoutubeDL.py", line 69                                                                                        6, in extract_info
    ie_result = ie.extract(url)
  File "/usr/local/lib/python2.7/dist-packages/youtube_dl/extractor/common.py",                                                                                         line 369, in extract
    return self._real_extract(url)
[...]

PR #12018, opened 2017-02-08
[Pornhub] Fix for Issue #12007

ThomasChr:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Fixes the pronhub extractor. Uses exec() which is not the smartest solution... suggestions are welcomed!
UPDATE: Does not use exec() anymore.

### DIFF
--- a/youtube_dl/extractor/pornhub.py
+++ b/youtube_dl/extractor/pornhub.py
@@ -6,7 +6,6 @@ import os
 import re
 
 from .common import InfoExtractor
-from ..jsinterp import JSInterpreter
 from ..compat import (
     compat_HTTPError,
     compat_urllib_parse_unquote,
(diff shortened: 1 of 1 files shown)


ThomasChr:
Fix for issue #12007 - I'm not comfortable with using exec() but it worked fine...

yan12125:
You should not use exec. Random codes may be executed.

ThomasChr:
@yan12125 I'm aware of that. Any suggestions how to solve the problem we're facing?

Pornhub does first define variables:
`a = 'http://www'`
`b = 'pornhub.com/video.mp4'`

And than the video url gets merged from the variables:
`url = a + b`

So I need to find some kind of way to execute this Javascript-Code - I don't think that writing a parser myself would be the solution. I hoped that youtube-dl had some "execute this javascript code"-Function.

yan12125:
There's a jsinterp module. See youtube.py for an example usage

ThomasChr:
@yan12125 Fixed it, it's kind of simple parsing, so no need to use JSInterpreter.

MikeRich88:
All you need to do is request the video with a cookie "platform=tv" then you get an unobfuscated video URL.

PR #12034, opened 2017-02-09
Changing get_element_by_attribute() and get_element_by_class() to return a list of results. See PR #11956

ThomasChr:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

As proposed by @yan12125 

> This is my idea:

    get_element_by_id is not touched (by HTML standards IDs are unique)
    get_element_by_attribute => get_elements_by_attribute
    get_element_by_class => get_elements_by_class

The two functions get_element_by_attribute and get_element_by_class are now returning a list of results.
The have also been renamed and all extractors changed to use the new functions.

### DIFF
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -337,17 +337,30 @@ def get_element_by_id(id, html):
 
 
 def get_element_by_class(class_name, html):
-    return get_element_by_attribute(
+    """Return the content of the first tag with the specified class in the passed HTML document"""
+    retval = get_elements_by_class(class_name, html)
+    return retval[0] if retval else None
+
+
+def get_element_by_attribute(attribute, value, html, escape_value=True):
+    retval = get_elements_by_attribute(attribute, value, html, escape_value)
+    return retval[0] if retval else None
+
+
+def get_elements_by_class(class_name, html):
+    """Return the content of all tags with the specified class in the passed HTML document as a list"""
+    return get_elements_by_attribute(
         'class', r'[^\'"]*\b%s\b[^\'"]*' % re.escape(class_name),
         html, escape_value=False)
 
 
-def get_element_by_attribute(attribute, value, html, escape_value=True):
+def get_elements_by_attribute(attribute, value, html, escape_value=True):
     """Return the content of the tag with the specified attribute in the passed HTML document"""
 
     value = re.escape(value) if escape_value else value
 
-    m = re.search(r'''(?xs)
+    retlist = []
+    for m in re.finditer(r'''(?xs)
         <([a-zA-Z0-9:._-]+)
          (?:\s+[a-zA-Z0-9:._-]+(?:=[a-zA-Z0-9:._-]*|="[^"]*"|='[^']*'))*?
          \s+%s=['"]?%s['"]?
@@ -355,16 +368,15 @@ def get_element_by_attribute(attribute, value, html, escape_value=True):
         \s*>
         (?P<content>.*?)
         </\1>
-    ''' % (re.escape(attribute), value), html)
+    ''' % (re.escape(attribute), value), html):
+        res = m.group('content')
 
-    if not m:
-        return None
-    res = m.group('content')
+        if res.startswith('"') or res.startswith("'"):
+            res = res[1:-1]
 
-    if res.startswith('"') or res.startswith("'"):
-        res = res[1:-1]
+        retlist.append(unescapeHTML(res))
 
-    return unescapeHTML(res)
+    return retlist
 
 
 class HTMLAttributeParser(compat_HTMLParser):
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -34,6 +34,9 @@ from youtube_dl.utils import (
     find_xpath_attr,
     fix_xml_ampersands,
     get_element_by_class,
+    get_element_by_attribute,
+    get_elements_by_class,
+    get_elements_by_attribute,
     InAdvancePagedList,
     intlist_to_bytes,
     is_html,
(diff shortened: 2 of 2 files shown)


dstftw:
Old functions must be preserved untouched.

ThomasChr:
@dstftw That was exactly what I proposed to @yan12125 in PR #11956 .
And what he didn't want! See the PR #11956 and see the message at top of this PR.

Could you please agree on that before I make any changes to this PR?
Changing all Extractors to use the new functions was a lot of work for me which could have been avoided...

dstftw:
Rename of any util method breaks all 3rdparty code that uses it. This is unwanted and should be avoided.

ThomasChr:
@dstftw Thats my opinion also. So I just add the new functions in this PR!

yan12125:
@dstftw For me only the ```YoutubeDL``` class should have stable APIs. But if you really want to preserve old function names, I suggest to call new functions from old functions, mark old functions as deprecated and use new functions in all extractors. The deprecation can be documented in change logs. After a few months deprecated functions can be removed.

dstftw:
>For me only the ```YoutubeDL``` class should have stable APIs.

I see no difference between `YoutubeDL` and `utils`. Both are encouraged to be used in code. Both should keep the interfaces.

>mark old functions as deprecated and use new functions in all extractors

Why deprecate at all? They should stay at least for convenience. If one only wants to get the first attribute with `get_elements_*` he'll be forced to do `[0]` all the time that is tedious and adds unnecessary clutter.

ThomasChr:
So everyone fine with this version?
I just added the new functions without using them anywhere.

I would like to get this changes to close PR #11956 and Issue #11953

dstftw:
1. Code duplication in single and multiple function versions should be eliminated.
2. Commits should be squashed.

ThomasChr:
@dstftw 
1. If you mind telling me how you would like to achieve that? Calling the old function in a loop won't work without a ugly loop which cuts of parts of the html. I'm open for suggestions.
2. I can squash them, no problem - I only thought you simpley squash them when applying the PR.

yan12125:
> They should stay at least for convenience

I believe codes with get_element_by_class is problematic - it conflicts with how HTML and Javascript's DOM API work. But nothing is impossible, so I have no objections on keeping them.

@ThomasChr Calling new functions from the original ones. And don't forget to update tests for new functions.

ThomasChr:
@yan12125 Yes, calling the new functions from the old ones is the essiest way. Should have seen that.

So I'm:
- Calling the new functions from the old ones
- Write Tests for the new functions
- Squash all my commits to one

Anything else?

dstftw:
@yan12125 how does *give-me-the-first-element-found-with-this-classname* scenario conflicts with DOM?

yan12125:
In Javascript there's getElementsByClassName, getElementById but no getElementByClassName. To get the first element with the given class name, web developers have to use ```getElementsByClassName('foobar')[0]```. If there's really such a use case, ```get_elements_by_class_name('foobar')[0]``` is slightly better than ```get_element_by_class_name('foobar')```, IMO. Basically the concept of "first element with the given class" is not a primitive in DOM manipulations.

dstftw:
>Basically the concept of "first element with the given class" is not a primitive in DOM manipulations.

Well, who cares? youtube-dl is not a web application. It does not even treat HTML as DOM in `get_element_by_attribute` rather it just dumb parse it with regex. So I don't see any point rigorously following concepts that semantically correct in DOM here. We aim to provide brief, concise and convenient set of APIs for data extraction to be used in conjunction youtube-dl, not to strictly follow some 	idiomatics inherent to some field X.

yan12125:
@dstftw Please read my previous responses carefully. My latest attitude is not rejecting preserving get_element_by_class. I just dislike it.

ThomasChr:
With this commit:
- The new functions are included
- The old functions call the new ones
- I have made three more tests
- All tests work fine

Everyone satisfied now?

youtube_dl/utils.py
```diff
@@ -337,34 +337,53 @@ def get_element_by_id(id, html):
...
-    res = m.group('content')
+        if res.startswith('"') or res.startswith("'"):
+            res = res[1:-1]
+
+        retlist.append(unescapeHTML(res))
 
-    if res.startswith('"') or res.startswith("'"):
-        res = res[1:-1]
+    if not retlist:
+        return None
```
dstftw: It should always return a list.

youtube_dl/utils.py
```diff
@@ -337,34 +337,53 @@ def get_element_by_id(id, html):
...
+    else:
+        return None
+
+
+def get_element_by_attribute(attribute, value, html, escape_value=True):
+    retval = get_elements_by_attribute(attribute, value, html, escape_value)
+    if retval:
+        return retval[0]
+    else:
+        return None
```
dstftw: This can be simplified to `return retval[0] if retval else None`.

youtube_dl/utils.py
```diff
@@ -337,34 +337,53 @@ def get_element_by_id(id, html):
...
+
+def get_elements_by_class(class_name, html):
+    return get_elements_by_attribute(
         'class', r'[^\'"]*\b%s\b[^\'"]*' % re.escape(class_name),
         html, escape_value=False)
 
 
-def get_element_by_attribute(attribute, value, html, escape_value=True):
+def get_elements_by_attribute(attribute, value, html, escape_value=True):
     """Return the content of the tag with the specified attribute in the passed HTML document"""
```
dstftw: This doc string does not match the function now.

youtube_dl/utils.py
```diff
@@ -337,34 +337,53 @@ def get_element_by_id(id, html):
...
 def get_element_by_class(class_name, html):
-    return get_element_by_attribute(
+    retval = get_elements_by_class(class_name, html)
+    if retval:
+        return retval[0]
+    else:
+        return None
+
+
+def get_element_by_attribute(attribute, value, html, escape_value=True):
```
dstftw: It worth adding a doc string.

youtube_dl/utils.py
ThomasChr: Wow. Did not know the syntax. Thanks!

youtube_dl/utils.py
ThomasChr: """Return the content of the first tag with the specified ID in the passed HTML document""" - okay?

dstftw: It's a list of elements not the first element.

ThomasChr:
Changed everything and melted into the first commit. That's okay for you?

dstftw:
Check code with flake8.

ThomasChr:
Checked with flake8, found one line with 4 trailing whitespaces :+1:

dstftw:
`.\test\test_utils.py:1137:87: W291 trailing whitespace`

ThomasChr:
Thats so embarrassing. I did not check the utils file :-(
Checked, corrected and recommited.

ThomasChr:
@dstftw Thanks for approving. I'm learning quite a lot and I'm glad to help!

ThomasChr:
Now that it's approved, could anyone remove the "pending fixes" label?

PR #12085, opened 2017-02-11
utils.py: Workaround TypeError with Python 2.7.13 in Windows

### LINKED ISSUE #11540: Downloads fail on Windows with Python 2.7.13

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like that [x])
- Use *Preview* tab to see how your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2016.12.22*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [X ] I've **verified** and **I assure** that I'm running youtube-dl **2016.12.22**

- [X ] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [X ] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

- [X] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add `-v` flag to **your command line** you run youtube-dl with, copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):
```
21:14:47.44 ++c:\s\yt>c:\Python27\Scripts\youtube-dl.exe -v
  File "c:\python27\lib\runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "c:\python27\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "c:\Python27\Scripts\youtube-dl.exe\__main__.py", line 9, in <module>
  File "c:\python27\lib\site-packages\youtube_dl\__init__.py", line 444, in main
    _real_main(argv)
  File "c:\python27\lib\site-packages\youtube_dl\__init__.py", line 56, in _real_main
    setproctitle('youtube-dl')
  File "c:\python27\lib\site-packages\youtube_dl\utils.py", line 1665, in setproctitle
    libc = ctypes.cdll.LoadLibrary('libc.so.6')
  File "c:\python27\lib\ctypes\__init__.py", line 440, in LoadLibrary
    return self._dlltype(name)
  File "c:\python27\lib\ctypes\__init__.py", line 362, in __init__
    self._handle = _dlopen(self._name, mode)
TypeError: LoadLibrary() argument 1 must be string, not unicode

21:15:32.69 ++c:\s\yt>
```
(Note this fails before any [debug] output)
---

### If the purpose of this *issue* is a *site support request* please provide all kinds of example URLs support for which should be included (replace following example URLs by **yours**):
- Single video: https://www.youtube.com/watch?v=BaW_jenozKc
- Single video: https://youtu.be/BaW_jenozKc
- Playlist: https://www.youtube.com/playlist?list=PL4lCao7KL_QFVb7Iudeipvc2BCavECqzc

---

### Description of your *issue*, suggested solution and other information
    
In Python 2.7.13, the Windows version of LoadLibrary requires a string, not a unicode arg. This is due to a change in Python-2.7.12\Modules\_ctypes\callproc.c vs. Python-2.7.13\Modules\_ctypes\callproc.c:
In:
    static PyObject *load_library(PyObject *self, PyObject *args)
there is this change (from 2.7.12 to 2.7.13):
    1280c1284
    <     if (!PyArg_ParseTuple(args, "O|O:LoadLibrary", &nameobj, &ignored))
    ---
    >     if (!PyArg_ParseTuple(args, "S|O:LoadLibrary", &nameobj, &ignored))

I was going to propose just changing:
        libc = ctypes.cdll.LoadLibrary('libc.so.6')
to:
        libc = ctypes.cdll.LoadLibrary(b'libc.so.6')
in utils.py, which does fix the TypeError, but I find that this TypeError does not occur under Python 3.6.0, apparently because callproc.c in 3.6.0 does not have the change shown above. Since this call to LoadLibrary() should fail under Windows anyway, maybe the correct fix is to have setproctitle() in utils.py just return if the LoadLibrary() raises either TypeError or OSError. 

Is is correct under either Python 2 or 3 to pass a unicode string to LoadLibrary() in either Windows or Posix? Should b'libc.so.6' be used in all cases (Windows, Posix/Linux, Python 2 or 3)? I note that in Linux (Posix) environment, the LoadLibrary() call goes to py_dl_open() instead of load_library(), and it also does not have the "S" typecheck in 2.7.13 or in 3.6.0.

wiiaboo:

- [x] Bug fix

---

Fixes #11540 
More exactly it's a workaround, not a fix, but people shouldn't be forced to use a months old version just because of a TypeError in a LoadLibrary that isn't even relevant in Windows.

Also recommend using Python2 for the official builds, since using 3.4.4 (more than a year old) is probably one of the bigger reasons why [this error](https://github.com/rg3/youtube-dl/issues?q=is%3Aissue+CERTIFICATE_VERIFY_FAILED+is%3Aclosed) is so prevalent in Windows lately.

### DIFF
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -1684,6 +1684,11 @@ def setproctitle(title):
         libc = ctypes.cdll.LoadLibrary('libc.so.6')
     except OSError:
         return
+    except TypeError:
+        # LoadLibrary in Windows Python 2.7.13 only expects
+        # a bytestring, but since unicode_literals turns
+        # every string into a unicode string, it fails.
+        return
     title_bytes = title.encode('utf-8')
     buf = ctypes.create_string_buffer(len(title_bytes))
     buf.value = title_bytes


yan12125:
Thanks for opening this. How many Python versions have you tested?

wiiaboo:
Well, the TypeError only triggers with Windows, so I tested with 2.7.12 (clean install), which triggered the `OSError` and carried on fine, and 2.7.13 (clean install), which triggers the `TypeError` and also carried on fine.

yan12125:
It's less likely to break existing codes than the ```str()``` approach, so I guess it's safe to merge it. Could you add some comments for ```TypeError```?

wiiaboo:
Ran `py -2 --version` before to make sure it was using the right versions each time, too; then tested running youtube-dl with `py -2 -m youtube_dl`.

wiiaboo:
Should I open a issue asking for the official builds to switch to using Python 2 instead of 3.4.4?

wiiaboo:
The fix would only need to be more involved were LoadLibrary be needed for a Windows lib, but this is the only use of it in Youtube-DL.

dstftw:
Official builds won't switch to python 2 since this brings encoding issues. Actually switch to python 3 from python 2 has taken place due to this.

wiiaboo:
So the only fix for official builds with SSL errors is to use `--no-check-certificate`? How's https://github.com/rg3/youtube-dl/issues/10014 proceeding?

dstftw:
Most of the SSL errors are due to expired certificates in system store removing which fixes the problem.

wiiaboo:
There's no expired certificates on my system store and I can consistently reproduce the SSL errors with 3.4.4.

yan12125:
> I can consistently reproduce the SSL errors with 3.4.4

There are tons of reasons leading to SSL failures; that's why I propose http://bugs.python.org/issue28182

Issue #12005, opened 2017-02-07
Site support request: www.programme-tv.net


loby66:
[x ] Site support request

`
>youtube-dl.exe -U
youtube-dl is up-to-date (2017.02.07)

>youtube-dl.exe -v http://www.programme-tv.net/videos/extraits/81095-guillaume-canet-evoque-les-rumeurs-d-infidelite-de-marion-cotillard-avec-brad-pitt-dans-vivement-dimanche/
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', 'http://www.programme-tv.net/videos/extraits/81095-guillaume-canet-evoque-les-rumeurs-d-infidelite-de-marion-cotillard-avec-brad-pitt-dans-vivement-dimanche/']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2017.02.07
[debug] Python version 3.4.4 - Windows-10-10.0.14393
[debug] exe versions: ffmpeg 2.8.4, ffprobe N-81045-g450cf40, rtmpdump 2.4
[debug] Proxy map: {}
[generic] 81095-guillaume-canet-evoque-les-rumeurs-d-infidelite-de-marion-cotillard-avec-brad-pitt-dans-vivement-dimanche: Requesting header
WARNING: Falling back on generic information extractor.
[generic] 81095-guillaume-canet-evoque-les-rumeurs-d-infidelite-de-marion-cotillard-avec-brad-pitt-dans-vivement-dimanche: Downloading webpage
[generic] 81095-guillaume-canet-evoque-les-rumeurs-d-infidelite-de-marion-cotillard-avec-brad-pitt-dans-vivement-dimanche: Extracting information
ERROR: Unsupported URL: http://www.programme-tv.net/videos/extraits/81095-guillaume-canet-evoque-les-rumeurs-d-infidelite-de-marion-cotillard-avec-brad-pitt-dans-vivement-dimanche/
Traceback (most recent call last):
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\rg3\tmpfjt892j_\build\youtube_dl\YoutubeDL.py", line 696, in extract_info
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\rg3\tmpfjt892j_\build\youtube_dl\extractor\common.py", line 369, in extract
[...]

Issue #11968, opened 2017-02-04
A model, well-commented extractor


johnhawkinson:
Since I'm not intimately familiar with the youtube-dl codebase, when I make a change to an extractor I often want to review other extractors to see how things have been handled previously. But oftentimes, it's really not clear which of several choices is better, what is favored, and what is more elegant.

If I had time to read through every single extractor I could make those determinations, but with 600+ of them, that's kind of daunting. Oftentimes I'll grep for a function and start looking at how abc.py and abcnews.py do it, since they are first alphabetically. But that's perhaps not a great plan.

It would be nice if there were a few well-written sample or "model" extractors that were highlighted as Best Practices that we could look at and review. And if those were clearly identified somehow, so I know where to look (maybe such model extractors exist already, I'm just not sure what to be looking at?). All things being equal, it'd be great if they were at the head of the alphabet, too :)

Thanks.

Also, the youtube-dl coding style seems to discourage comments. Maybe when you're very familiar with the code and the functions in common.py they don't seem so important, but for someone new to the codebase, they can help a lot. So it would be nice if any such model extractors could also be well-commented, or at least more verbosely commented than is typical in the codebase.

yan12125:
That's a great suggestion. However, in youtube-dl there's no uniform coding convention. Coding styles change fast with time, so new or recent changed extractors looks quite different than earlier extractors. (I would rather say my early contributions are like a shit :D) To make things worse, coding styles between major developers are different as they may be involved in other Python projects. I don't think such a model extractor can really exist. If you really need well-written extractors, consider those who are written mostly by major developers (dstftw, phihag, jaimeMF, remitamine and me).

If you really need concrete examples, you may consider to have a look into litv.py, bilibili.py and niconico.py - those are my recent changes.

About comments:

It's also a great idea. I prefer to document functions where they are defined. At least ```InfoExtractor``` and ```YoutubeDL``` should be documented. The former is the base class for all extractors and the latter is the public API of the ```youtube_dl``` module. I imagine there's something like Java's javadoc, and I've heard pydoc, too, but I'm not sure is it suitable. I may try to add some comments to functions in aforementioned classes and see what happened.

johnhawkinson:
### Model extractors

[...]

PR #12099, opened 2017-02-12
[Bostonglobe] New. Nonstandard version of Brightcove

johnhawkinson:
BrightcoveNewIE.extract_urls(): Handle player data when it is all in
attributes of the <video> tag, and also the id is
data-brightcove-video-id not data-video-id.

Add test to generic extractor, note HEAD gives 404, no checksum computed(?).

---
Can someone explain to me why md5 checksums do not appear to be computed by the test harness in this case? Is this an issue with playlists, and can we document that?  I don't see an open issue, but it seems problematic if playlist tests can't be checksummed.

### DIFF
--- a/youtube_dl/extractor/bostonglobe.py
+++ b/youtube_dl/extractor/bostonglobe.py
@@ -1,0 +1,72 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+
+from ..utils import (
+    extract_attributes,
+)
+
+
+class BostonGlobeIE(InfoExtractor):
+    _VALID_URL = r'(?i)https?://(?:www\.)?bostonglobe\.com/.*/(?P<id>[^/]+)/\w+(?:\.html)?'
+    _TESTS = [
+        {
+            'url': 'http://www.bostonglobe.com/metro/2017/02/11/tree-finally-succumbs-disease-leaving-hole-neighborhood/h1b4lviqzMTIn9sVy8F3gP/story.html',
+            'md5': '0a62181079c85c2d2b618c9a738aedaf',
+            'info_dict': {
+                'title': 'A tree finally succumbs to disease, leaving a hole in a neighborhood',
+                'id': '5320421710001',
+                'ext': 'mp4',
+                'description': 'It arrived as a sapling when the Back Bay was in its infancy, a spindly American elm tamped down into a square of dirt cut into the brick sidewalk of 1880s Marlborough Street, no higher than the first bay window of the new brownstone behind it.',
+                'timestamp': 1486877593,
+                'upload_date': '20170212',
+                'uploader_id': '245991542',
+            },
+        },
+        {
+            # Embedded youtube video; we hand it off to the Generic extractor.
+            'url': 'https://www.bostonglobe.com/lifestyle/names/2017/02/17/does-ben-affleck-play-matt-damon-favorite-version-batman/ruqkc9VxKBYmh5txn1XhSI/story.html',
+            'md5': '582b40327089d5c0c949b3c54b13c24b',
+            'info_dict': {
+                'title': "Who Is Matt Damon's Favorite Batman?",
+                'id': 'ZW1QCnlA6Qc',
+                'ext': 'mp4',
+                'upload_date': '20170217',
+                'description': 'md5:3b3dccb9375867e0b4d527ed87d307cb',
+                'uploader': 'The Late Late Show with James Corden',
+                'uploader_id': 'TheLateLateShow',
+            },
+            'expected_warnings': ['404'],
+        },
+    ]
+
+    def _real_extract(self, url):
+        page_id = self._match_id(url)
+        webpage = self._download_webpage(url, page_id)
+
+        page_title = self._og_search_title(webpage, default=None)
+
+        # <video data-brightcove-video-id="5320421710001" data-account="245991542" data-player="SJWAiyYWg" data-embed="default" class="video-js" controls itemscope itemtype="http://schema.org/VideoObject">
+        entries = []
+        for video in re.findall(r'(?i)(<video[^>]+>)', webpage):
+            attrs = extract_attributes(video)
+
+            video_id = attrs.get('data-brightcove-video-id')
+            account_id = attrs.get('data-account')
+            player_id = attrs.get('data-player')
+            embed = attrs.get('data-embed')
+
+            if video_id and account_id and player_id and embed:
+                entries.append(
+                    'http://players.brightcove.net/%s/%s_%s/index.html?videoId=%s'
+                    % (account_id, player_id, embed, video_id))
+
+        if len(entries) == 0:
+            return self.url_result(url, 'Generic')
+        elif len(entries) == 1:
+            return self.url_result(entries[0], 'BrightcoveNew')
+        else:
+            return self.playlist_from_matches(entries, page_id, page_title, ie='BrightcoveNew')
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -36,34 +36,35 @@ from ..utils import (
     clean_html,
     compiled_regex_type,
     determine_ext,
+    determine_protocol,
     error_to_compat_str,
     ExtractorError,
+    extract_attributes,
     fix_xml_ampersands,
     float_or_none,
     GeoRestrictedError,
     GeoUtils,
     int_or_none,
     js_to_json,
+    mimetype2ext,
+    orderedSet,
+    parse_codecs,
+    parse_duration,
     parse_iso8601,
+    parse_m3u8_attributes,
     RegexNotFoundError,
-    sanitize_filename,
     sanitized_Request,
+    sanitize_filename,
     unescapeHTML,
     unified_strdate,
     unified_timestamp,
+    update_Request,
+    update_url_query,
+    urljoin,
     url_basename,
     xpath_element,
     xpath_text,
     xpath_with_ns,
-    determine_protocol,
-    parse_duration,
-    mimetype2ext,
-    update_Request,
-    update_url_query,
-    parse_m3u8_attributes,
-    extract_attributes,
-    parse_codecs,
-    urljoin,
 )
 
 
@@ -713,6 +714,13 @@ class InfoExtractor(object):
         if video_title is not None:
             video_info['title'] = video_title
         return video_info
+
+    def playlist_from_matches(self, matches, video_id, video_title, getter=None, ie=None):
+        urlrs = orderedSet(
+            self.url_result(self._proto_relative_url(getter(m) if getter else m), ie)
+            for m in matches)
+        return self.playlist_result(
+            urlrs, playlist_id=video_id, playlist_title=video_title)
 
     @staticmethod
     def playlist_result(entries, playlist_id=None, playlist_title=None, playlist_description=None):
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -1824,14 +1824,6 @@ class GenericIE(InfoExtractor):
         video_description = self._og_search_description(webpage, default=None)
         video_thumbnail = self._og_search_thumbnail(webpage, default=None)
 
-        # Helper method
-        def _playlist_from_matches(matches, getter=None, ie=None):
-            urlrs = orderedSet(
-                self.url_result(self._proto_relative_url(getter(m) if getter else m), ie)
-                for m in matches)
-            return self.playlist_result(
-                urlrs, playlist_id=video_id, playlist_title=video_title)
-
         # Look for Brightcove Legacy Studio embeds
         bc_urls = BrightcoveLegacyIE._extract_brightcove_urls(webpage)
         if bc_urls:
@@ -1852,28 +1844,28 @@ class GenericIE(InfoExtractor):
         # Look for Brightcove New Studio embeds
         bc_urls = BrightcoveNewIE._extract_urls(webpage)
         if bc_urls:
-            return _playlist_from_matches(bc_urls, ie='BrightcoveNew')
+            return self.playlist_from_matches(bc_urls, video_id, video_title, ie='BrightcoveNew')
 
         # Look for ThePlatform embeds
         tp_urls = ThePlatformIE._extract_urls(webpage)
         if tp_urls:
-            return _playlist_from_matches(tp_urls, ie='ThePlatform')
+            return self.playlist_from_matches(tp_urls, video_id, video_title, ie='ThePlatform')
 
         # Look for Vessel embeds
         vessel_urls = VesselIE._extract_urls(webpage)
         if vessel_urls:
-            return _playlist_from_matches(vessel_urls, ie=VesselIE.ie_key())
+            return self.playlist_from_matches(vessel_urls, video_id, video_title, ie=VesselIE.ie_key())
 
         # Look for embedded rtl.nl player
         matches = re.findall(
             r'<iframe[^>]+?src="((?:https?:)?//(?:www\.)?rtl\.nl/system/videoplayer/[^"]+(?:video_)?embed[^"]+)"',
             webpage)
         if matches:
-            return _playlist_from_matches(matches, ie='RtlNl')
+            return self.playlist_from_matches(matches, video_id, video_title, ie='RtlNl')
 
         vimeo_urls = VimeoIE._extract_urls(url, webpage)
         if vimeo_urls:
-            return _playlist_from_matches(vimeo_urls, ie=VimeoIE.ie_key())
+            return self.playlist_from_matches(vimeo_urls, video_id, video_title, ie=VimeoIE.ie_key())
 
         vid_me_embed_url = self._search_regex(
             r'src=[\'"](https?://vid\.me/[^\'"]+)[\'"]',
@@ -1895,25 +1887,25 @@ class GenericIE(InfoExtractor):
                 (?:embed|v|p)/.+?)
             \1''', webpage)
         if matches:
-            return _playlist_from_matches(
-                matches, lambda m: unescapeHTML(m[1]))
+            return self.playlist_from_matches(
+                matches, video_id, video_title, lambda m: unescapeHTML(m[1]))
 
         # Look for lazyYT YouTube embed
         matches = re.findall(
             r'class="lazyYT" data-youtube-id="([^"]+)"', webpage)
         if matches:
-            return _playlist_from_matches(matches, lambda m: unescapeHTML(m))
+            return self.playlist_from_matches(matches, video_id, video_title, lambda m: unescapeHTML(m))
 
         # Look for Wordpress "YouTube Video Importer" plugin
         matches = re.findall(r'''(?x)<div[^>]+
             class=(?P<q1>[\'"])[^\'"]*\byvii_single_video_player\b[^\'"]*(?P=q1)[^>]+
             data-video_id=(?P<q2>[\'"])([^\'"]+)(?P=q2)''', webpage)
         if matches:
-            return _playlist_from_matches(matches, lambda m: m[-1])
+            return self.playlist_from_matches(matches, video_id, video_title, lambda m: m[-1])
 
         matches = DailymotionIE._extract_urls(webpage)
         if matches:
-            return _playlist_from_matches(matches)
+            return self.playlist_from_matches(matches, video_id, video_title)
 
         # Look for embedded Dailymotion playlist player (#3822)
         m = re.search(
@@ -1922,8 +1914,8 @@ class GenericIE(InfoExtractor):
             playlists = re.findall(
                 r'list\[\]=/playlist/([^/]+)/', unescapeHTML(m.group('url')))
             if playlists:
-                return _playlist_from_matches(
-                    playlists, lambda p: '//dailymotion.com/playlist/%s' % p)
+                return self.playlist_from_matches(
+                    playlists, video_id, video_title, lambda p: '//dailymotion.com/playlist/%s' % p)
 
         # Look for embedded Wistia player
         match = re.search(
@@ -2030,8 +2022,9 @@ class GenericIE(InfoExtractor):
         if mobj is not None:
             embeds = self._parse_json(mobj.group(1), video_id, fatal=False)
             if embeds:
-                return _playlist_from_matches(
-                    embeds, getter=lambda v: OoyalaIE._url_for_embed_code(smuggle_url(v['provider_video_id'], {'domain': url})), ie='Ooyala')
+                return self.playlist_from_matches(
+                    embeds, video_id, video_title,
+                    getter=lambda v: OoyalaIE._url_for_embed_code(smuggle_url(v['provider_video_id'], {'domain': url})), ie='Ooyala')
 
         # Look for Aparat videos
         mobj = re.search(r'<iframe .*?src="(http://www\.aparat\.com/video/[^"]+)"', webpage)
@@ -2093,13 +2086,13 @@ class GenericIE(InfoExtractor):
         # Look for funnyordie embed
         matches = re.findall(r'<iframe[^>]+?src="(https?://(?:www\.)?funnyordie\.com/embed/[^"]+)"', webpage)
         if matches:
-            return _playlist_from_matches(
-                matches, getter=unescapeHTML, ie='FunnyOrDie')
+            return self.playlist_from_matches(
+                matches, video_id, video_title, getter=unescapeHTML, ie='FunnyOrDie')
 
         # Look for BBC iPlayer embed
         matches = re.findall(r'setPlaylist\("(https?://www\.bbc\.co\.uk/iplayer/[^/]+/[\da-z]{8})"\)', webpage)
         if matches:
-            return _playlist_from_matches(matches, ie='BBCCoUk')
+            return self.playlist_from_matches(matches, video_id, video_title, ie='BBCCoUk')
 
         # Look for embedded RUTV player
         rutv_url = RUTVIE._extract_url(webpage)
@@ -2114,32 +2107,32 @@ class GenericIE(InfoExtractor):
         # Look for embedded SportBox player
         sportbox_urls = SportBoxEmbedIE._extract_urls(webpage)
         if sportbox_urls:
-            return _playlist_from_matches(sportbox_urls, ie='SportBoxEmbed')
+            return self.playlist_from_matches(sportbox_urls, video_id, video_title, ie='SportBoxEmbed')
 
         # Look for embedded XHamster player
         xhamster_urls = XHamsterEmbedIE._extract_urls(webpage)
         if xhamster_urls:
-            return _playlist_from_matches(xhamster_urls, ie='XHamsterEmbed')
+            return self.playlist_from_matches(xhamster_urls, video_id, video_title, ie='XHamsterEmbed')
 
         # Look for embedded TNAFlixNetwork player
         tnaflix_urls = TNAFlixNetworkEmbedIE._extract_urls(webpage)
         if tnaflix_urls:
-            return _playlist_from_matches(tnaflix_urls, ie=TNAFlixNetworkEmbedIE.ie_key())
+            return self.playlist_from_matches(tnaflix_urls, video_id, video_title, ie=TNAFlixNetworkEmbedIE.ie_key())
 
         # Look for embedded PornHub player
         pornhub_urls = PornHubIE._extract_urls(webpage)
         if pornhub_urls:
-            return _playlist_from_matches(pornhub_urls, ie=PornHubIE.ie_key())
+            return self.playlist_from_matches(pornhub_urls, video_id, video_title, ie=PornHubIE.ie_key())
 
         # Look for embedded DrTuber player
         drtuber_urls = DrTuberIE._extract_urls(webpage)
         if drtuber_urls:
-            return _playlist_from_matches(drtuber_urls, ie=DrTuberIE.ie_key())
+            return self.playlist_from_matches(drtuber_urls, video_id, video_title, ie=DrTuberIE.ie_key())
 
         # Look for embedded RedTube player
         redtube_urls = RedTubeIE._extract_urls(webpage)
         if redtube_urls:
-            return _playlist_from_matches(redtube_urls, ie=RedTubeIE.ie_key())
+            return self.playlist_from_matches(redtube_urls, video_id, video_title, ie=RedTubeIE.ie_key())
 
         # Look for embedded Tvigle player
         mobj = re.search(
@@ -2185,12 +2178,12 @@ class GenericIE(InfoExtractor):
         # Look for embedded soundcloud player
         soundcloud_urls = SoundcloudIE._extract_urls(webpage)
         if soundcloud_urls:
-            return _playlist_from_matches(soundcloud_urls, getter=unescapeHTML, ie=SoundcloudIE.ie_key())
+            return self.playlist_from_matches(soundcloud_urls, video_id, video_title, getter=unescapeHTML, ie=SoundcloudIE.ie_key())
 
         # Look for tunein player
         tunein_urls = TuneInBaseIE._extract_urls(webpage)
         if tunein_urls:
-            return _playlist_from_matches(tunein_urls)
+            return self.playlist_from_matches(tunein_urls, video_id, video_title)
 
         # Look for embedded mtvservices player
         mtvservices_url = MTVServicesEmbeddedIE._extract_url(webpage)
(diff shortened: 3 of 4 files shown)


yan12125:
As a side note, this pattern is quite similar to #12005:
```
        <video data-video-id="5311302538001"
               data-account="800000640001"
               data-player="HyenP6LuN"
               data-embed="default"
               class="video-js apply-ratio"
                               autoplay                preload
               controls></video>
```

youtube_dl/extractor/brightcove.py
```diff
@@ -508,7 +508,22 @@ def _extract_urls(webpage):
...
+# 	<video data-brightcove-video-id="5320421710001" data-account="245991542" data-player="SJWAiyYWg" data-embed="default" class="video-js" controls itemscope itemtype="http://schema.org/VideoObject">
+        for video_id, account_id, player_id, embed in re.findall(
+                r'''(?sx)
+                    <video[^>]+
+                        data-brightcove-video-id=["\'](\d+|ref:[^"\']+)["\'].*?
+                        data-account=["\'](\d+)["\'].*?
+                        data-player=["\'](\w+)["\'].*?
+                        data-embed=["\'](\w+)["\'].*?
+                    </video>
+                ''', webpage):
```
yan12125: Use ```extract_attributes``` instead. The order of video_id, account_id, ... may change.

johnhawkinson: Done. As in #11968, of course I found myself struggling to find good examples of how it was best used, and found plenty of not-so-awesome ones...

youtube_dl/extractor/brightcove.py
```diff
@@ -508,7 +509,24 @@ def _extract_urls(webpage):
...
                 % (account_id, player_id, embed, video_id))
 
-        return entries
+        # 	<video data-brightcove-video-id="5320421710001" data-account="245991542" data-player="SJWAiyYWg" data-embed="default" class="video-js" controls itemscope itemtype="http://schema.org/VideoObject">
+        for video in re.findall(r'(?i)(<video[^>]+>)', webpage):
+            attrs = extract_attributes(video)
+                
+            video_id    = attrs.get('data-brightcove-video-id')
+            if video_id is None:
+                video_id = attrs.get('data-video-id')
```
yan12125: ```dict_get``` makes codes even shorter

youtube_dl/extractor/generic.py
```diff
@@ -447,6 +447,28 @@ class GenericIE(InfoExtractor):
...
+                    'description': 'It arrived as a sapling when the Back Bay was in its infancy, a spindly American elm tamped down into a square of dirt cut into the brick sidewalk of 1880s Marlborough Street, no higher than the first bay window of the new brownstone behind it.',
+                    'timestamp': 1486877593,
+                    'upload_date': '20170212',
+                    'uploader_id': '245991542',
+                },
+            }],
+            # HEAD requests produce 404 :(
+            'expected_warnings': ['404'],
+
+        },
```
yan12125: Could you also add the test from #12005?

remitamine:
a similar approache has been used from the beginning when i added the extractor but it was changed in 536f819eda975224666374a9ce83cc3472f5aa5a, i guess because this attributes can be used by other websites for other contexts.
also `data-brightcove-video-id` is not used by the standard embeds https://docs.brightcove.com/en/perform/brightcove-player/guides/in-page-embed-player-implementation.html.

youtube_dl/extractor/common.py
```diff
@@ -2194,7 +2194,11 @@ def geo_verification_headers(self):
         return headers
 
     def _generic_id(self, url):
-        return compat_urllib_parse_unquote(os.path.splitext(url.rstrip('/').split('/')[-1])[0])
+        id = compat_urllib_parse_unquote(os.path.splitext(url.rstrip('/').split('/')[-1])[0])
+        # If our id matches a stopword, use the 2nd-to-last URL path component instead
+        if id in ['story']:
+            id = compat_urllib_parse_unquote(os.path.splitext(url.rstrip('/').split('/')[-2])[0])
```
dstftw: There should not be any ad hoc hacks in generic methods.

johnhawkinson:
Requested changes done. Also an addition to _generic_id() to avoid ending up with 'story' as the ID for all Boston Globe urls. I was surprised I couldn't  find other examples in tests that have this sort of problem.

It does look like @remitamine's original approach was more general than what is there now. I'm not sure what's going on there.

I agree `data-brightcove-video-id` isn't in the docs, but it still seems to me that it's appropriate to look for the more-specific identifier before the generic one. But I'm happy to flip the order of the test if someone has a better argument.

youtube_dl/extractor/brightcove.py
```diff
@@ -508,7 +510,21 @@ def _extract_urls(webpage):
...
+            account_id  = attrs.get('data-account')
+            player_id   = attrs.get('data-player')
+            embed       = attrs.get('data-embed')
+
+            if video_id and account_id and player_id and embed:
+                entries.append(
+                    'http://players.brightcove.net/%s/%s_%s/index.html?videoId=%s'
+                    % (account_id, player_id, embed, video_id))
+
+	return entries
```
dstftw: As already mentioned by @remitamine it's not a generic embed and should not be used in generic extractor.

youtube_dl/extractor/common.py
johnhawkinson: Reverted, but I don't understand why...It doesn't seem correct to:
. Create a custom extractor for bostonglobe.com, given that it just another BrightCove site.
. Extract videos all with the id 'story' when there is a simple fix
I'm not sure what qualifies as an 'ad hoc hack' versus a good general fix here. I would have categorized this one differently than you did. Is there something to be done? Happy to put it in a separate PR, also.

youtube_dl/extractor/brightcove.py
johnhawkinson: Huh? That's not what @remitamine said, and this is in brightcove.py.
What change would you like me to make? I am confused.

dstftw: >also `data-brightcove-video-id` is not used by the standard embeds https://docs.brightcove.com/en/perform/brightcove-player/guides/in-page-embed-player-implementation.html.

You should add separate extractor instead. Or should provide another example URLs across web to prove this pattern is used not only on this site.

youtube_dl/extractor/common.py
dstftw: Why do you disallow `story` to be an id? Because you do not want it to be an id in your particular case that is clearly an ad hoc hack cause other extractors may `_generic_id` and may allow `story` to be an id.

johnhawkinson: It seems like `story` is always a bad choice for an id, just like `video` would be a bad choice, or classic english stopwords like `the`, `a`, or `an`.
Note that the proposed patch doesn't disallow it, per se, it just chooses another portion of the URL path instead.

Honestly, this seems like a pretty bad way to design a website. I can't figure out why anyone would do it that way. Still, some UK newspapers seem to use this pattern too:
http://www.derbytelegraph.co.uk/quiz-test-your-derby-county-knowledge-as-the-rams-host-bristol-city/story-30126709-detail/story.html
http://www.stokesentinel.co.uk/stoke-on-trent-central-candidates-attend-first-public-hustings/story-30115329-detail/story.html

I did check and there are no tests in any extractor that expect an id of 'story'.

Anyhow, whatever, it seems to me it's better with this than without this, but it doesn't affect core functionality and you disagree, so that's that. And it's currently out of the branch.

dstftw: `story` may be a 5-character generated id even though the probability of this outcome is close to zero. Even if we stick with this stopword approach why only `story`? There lots of other words and lots of other languages. It's just not feasible.
In case of generic extractor the id is more like a dummy placeholder. It's not used to identify something and you can't make it meaningful for all the cases while in generic extractor anyway.

youtube_dl/extractor/brightcove.py
johnhawkinson: Well, I definitely don't have any other examples. My best guess from the few google hits for "data-brightcove-video-id" is that this is an older specification that is deprecated but some sites still use it.

A separate extractor? That seems like a lot of work and I don't see what the benefit would be? A lot of code duplication, extra maintenance work, and no actual gain? It would just end up doing the same thing?

I'm also not sure if you mean I should have a `BostonGlobeIE(BrightcoveNewIE)` in `bostonglobe.py` and bypass the generic extractor (I guess this could handle the `video_id` problem?), or a `BrightcoveAltVideoTagIE(BrightcoveNewIE)` in `brightcove.py` that handles this special case of brightcove with the `data-brightcove-video-id` instead of `data-video-id`?

All of these choices seem to introduce a lot more complexity than the current proposal (which implements this difference in one line), so I feel like I am misunderstanding you.

Thanks. Sorry for being obtuse.

youtube_dl/extractor/common.py
johnhawkinson: Well, `story` was the only example I could find. I assume others would add more words to this list as they came up.
I hadn't appreciated that the id did not appear in the output filename, I see your point. (It does appear in the debug logs, e.g. --write-pages and soforth, but not in the final video output). That does make this seem not worth the effort.

youtube_dl/extractor/brightcove.py
yan12125: Besides providing another example, you can also prove that Brightcove's Javascript indeed references data-brightcove-video-id. For example, in https://github.com/rg3/youtube-dl/blob/732d116/youtube_dl/extractor/jwplatform.py#L36, there are lots of deprecated usages.

johnhawkinson: http://www.bostonglobe.com/js/globe-brightcove-player.js

```
// *
// *  Brightcove Player - Populates all brightcove players with video and DFP preroll
// *  Originally written for the BDC brightcove plugin, modified for Globe
// *
...
	/**
	 * Instantiate the player
	 */
	module.createPlayer = function(i) {
		// Let's create a player
		var brightcovePlayer = videojs(this.id);

		// Load the ad ONLY ONCE when this player emits event 'loadedmetadata'
		brightcovePlayer.one( 'loadedmetadata', (module.addAdvertisement).bind(brightcovePlayer) );
		/**
		 * Okay now, let's make a request and give it the loadVideo callback.
		 * We give it the id from the data attribute 'data-brightcove-video-id'
		 */
		var brightcoveVideoRequest = brightcovePlayer.catalog.getVideo( $(this).data('brightcove-video-id'), (module.loadVideo).bind(brightcovePlayer) );
	};
```

yan12125: Thanks. It turns out that data-brightcove-video-id is a bostonglobe-specific usage, so a dedicated extractor is better.

johnhawkinson: Seems a waste, but how should I do it. I asked above:

> I'm also not sure if you mean I should have a `BostonGlobeIE(BrightcoveNewIE)` in `bostonglobe.py` and bypass the generic extractor (I guess this could handle the `video_id` problem?), or a `BrightcoveAltVideoTagIE(BrightcoveNewIE)` in `brightcove.py` that handles this special case of brightcove with the `data-brightcove-video-id` instead of `data-video-id`?

Thanks.

yan12125: You'll need ```BostonGlobeIE``` in ```bostonglobe.py```

johnhawkinson: Done. I have to say, it seems to have both more code duplication (from brightcove.py and generic.py both) as well as more fragile (what happens when there are other kinds of videos on bostonglobe.com that the Generic extractor would have caught, but this one won't)...just shows, once again, I don't understand the youtube-dl philosophy...maybe that could be better documented.

johnhawkinson:
Also, can anyone please help with:

> Can someone explain to me why md5 checksums do not appear to be computed by the test harness in this case? Is this an issue with playlists, and can we document that?  I don't see an open issue, but it seems problematic if playlist tests can't be checksummed.

Thanks.

remitamine:
md5 checksum will be checked if it's available in the test case even in playlists:https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/orf.py#L30.

johnhawkinson:
> md5 checksum will be checked if it's available in the test case even in playlists:

When I added a dummy md5 checksum (either in the playlist entry or in the overall test), no error was reported. Did I do it wrong?

johnhawkinson:
Specifically, with an md5 field inside the playlist info_dict:
```
FAIL: test_BostonGlobe (__main__.TestDownload)
  File "test/test_download.py", line 210, in test_template
    expect_info_dict(self, info_dict, tc.get('info_dict', {}))
  File "/Users/jhawk/src/youtube-dl/test/helper.py", line 177, in expect_info_dict
  File "/Users/jhawk/src/youtube-dl/test/helper.py", line 173, in expect_dict
  File "/Users/jhawk/src/youtube-dl/test/helper.py", line 167, in expect_value
    'Invalid value for field %s, expected %r, got %r' % (field, expected, got))
AssertionError: Invalid value for field md5, expected u'dummy1', got None
```
And an md5 field in the test itself is ignored:
```
    _TEST = {
            'url': 'http://www.bostonglobe.com/metro/2017/02/11/tree-finally-succumbs-disease-leaving-hole-neighborhood/h1b4lviqzMTIn9sVy8F3gP/story.html',
            'md5': 'dummy3',
...
```

remitamine:
i'm not sure why it's not working for you, but i just tested it with an md5 value in list item, it raise an error when the md5 is wrong.

johnhawkinson:
I tried to run the `orf.py` test, but 

```
Skipping ORFTVthek: Blocked outside of Austria / Germany
```

and uncommenting the skip directive I get a 404.

Do you have another example?

remitamine:
```
python test_download.py TestDownload.test_Seeker_1
[Seeker] changes-expected-at-zoos-following-recent-gorilla-lion-shootings: Downloading webpage
[download] Downloading playlist: After Gorilla Killing, Changes Ahead for Zoos
[Seeker] playlist After Gorilla Killing, Changes Ahead for Zoos: Collected 1 video ids (downloading 1 of them)
[revision3:embed] 67558: Downloading JSON metadata
[revision3:embed] 67558: Downloading m3u8 information
[info] Writing video description metadata as JSON to: 67558.info.json
[debug] Invoking downloader on 'http://videos.revision3.com/revision3/web/dnews/2260/dnews--2260--the-pros-cons-of-zoos--hd720p30.h264.mp4'
[download] Destination: 67558.mp4
[download] 100% of 10.00KiB in 00:01
[download] Finished downloading playlist: After Gorilla Killing, Changes Ahead for Zoos
.
Ran 1 test in 12.183s

OK
```
changing the md5:
```
python test_download.py TestDownload.test_Seeker_1
[Seeker] changes-expected-at-zoos-following-recent-gorilla-lion-shootings: Downloading webpage
[download] Downloading playlist: After Gorilla Killing, Changes Ahead for Zoos
[Seeker] playlist After Gorilla Killing, Changes Ahead for Zoos: Collected 1 video ids (downloading 1 of them)
[revision3:embed] 67558: Downloading JSON metadata
[revision3:embed] 67558: Downloading m3u8 information
[info] Writing video description metadata as JSON to: 67558.info.json
[debug] Invoking downloader on 'http://videos.revision3.com/revision3/web/dnews/2260/dnews--2260--the-pros-cons-of-zoos--hd720p30.h264.mp4'
[download] Destination: 67558.mp4
[download] Finished downloading playlist: After Gorilla Killing, Changes Ahead for Zoos
F
FAIL: test_Seeker_1 (__main__.TestDownload)
  File "test_download.py", line 202, in test_template
    self.assertEqual(md5_for_file, tc['md5'])
AssertionError: '83bcd157cab89ad7318dd7b8c9cf1306' != '83bcd57cab89ad7318dd7b8c9cf1306'
- 83bcd157cab89ad7318dd7b8c9cf1306
?      -
+ 83bcd57cab89ad7318dd7b8c9cf1306


Ran 1 test in 7.770s

FAILED (failures=1)
```

johnhawkinson:
Thanks, @remitamine, I feel dumb!
I had tried it in all 3 wrong places: the top-level test, the top-level info_dict, and the playlist info_dict. But the right place is the playlist outside the info_dict, per 349ea41bb8902fb76aca37d4c0dd4cfbe1498717.

Thanks for leading me to the right place.

youtube_dl/extractor/brightcove.py
```diff
@@ -508,6 +510,20 @@ def _extract_urls(webpage):
...
+            video_id    = attrs.get('data-video-id')
+            account_id  = attrs.get('data-account')
+            player_id   = attrs.get('data-player')
+            embed       = attrs.get('data-embed')
+
+            if video_id and account_id and player_id and embed:
+                entries.append(
+                    'http://players.brightcove.net/%s/%s_%s/index.html?videoId=%s'
+                    % (account_id, player_id, embed, video_id))
+
```
yan12125: As remitamine has said, there's a similar detection at https://github.com/rg3/youtube-dl/blob/16e2c8f/youtube_dl/extractor/brightcove.py#L494-L509. It's better to merge them together. Specificially, the detection can be done in 2 steps:

1. Detect the Brightcove loader script. It can be either a <script> tag, as [recommended by Brightcove](https://support.brightcove.com/en/video-cloud/docs/dynamically-assigning-videos-player) or via [script.js](https://github.com/ded/script.js), just like the example in http://www.programme-tv.net/videos/extraits/81095-guillaume-canet-evoque-les-rumeurs-d-infidelite-de-marion-cotillard-avec-brad-pitt-dans-vivement-dimanche/:
```Javascript
$script([
            '//players.brightcove.net/800000640001/HyenP6LuN_default/index.min.js',
            '//players.brightcove.net/800000640001/Hyx5mvd4_default/index.min.js'
        ], 'brightcove')
```
2. Check <video> tags. Just like what Brightcove loader scripts do: (below is taken from http://players.brightcove.net/800000640001/Hyx5mvd4_default/index.min.js)
```Javascript
    i = document.querySelectorAll("video");
    for (k = 0; k < i.length; k++) {
        j = i[k];
        if (j.getAttribute("data-player") === "Hyx5mvd4" && j.getAttribute("data-embed") === "default") {
            if (/(?:^|\s)vjs-tech(?:\s|$)/.test(j.className)) {
                videojs.log("Ignoring already initialized player");
                continue
            }
            h(j)
        }
    }
```

youtube_dl/extractor/bostonglobe.py
```diff
@@ -0,0 +1,71 @@
...
+        webpage = self._download_webpage(url, page_id)
+
+        # cribbed from generic.py
+        page_title = self._og_search_title(
+            webpage, default=None) or self._html_search_regex(
+            r'(?s)<title>(.*?)</title>', webpage, 'video title',
+            default='video')
+
+        # from generic.py
+        def _playlist_from_matches(matches, getter=None, ie=None):
```
yan12125: If a method is used by at least two extractors, it can be moved to ```common.py```

youtube_dl/extractor/bostonglobe.py
```diff
@@ -0,0 +1,71 @@
...
+                for m in matches)
+            return self.playlist_result(
+                urlrs, playlist_id=page_id, playlist_title=page_title)
+
+        # 	<video data-brightcove-video-id="5320421710001" data-account="245991542" data-player="SJWAiyYWg" data-embed="default" class="video-js" controls itemscope itemtype="http://schema.org/VideoObject">
+        entries = []
+        for video in re.findall(r'(?i)(<video[^>]+>)', webpage):
+            attrs = extract_attributes(video)
+
+            video_id    = dict_get(attrs, ['data-brightcove-video-id', 'data-video-id'])
```
yan12125: As this code is moved to a dedicated extractor, only 'data-brightcove-video-id' should be detected.

youtube_dl/extractor/brightcove.py
johnhawkinson: Honestly, I'm pretty confused what you're asking for. I agree they ought to be merged, and I'm still not clear why a separate extractor is necessary and why this can't just go in `brightcove.py` and handle `data-brightcove-video-id`.

But here you seem to be suggesting `brightcove.py` should change its detection heuristic to search for the script tag?  That doesn't work for the `bostonglobe.com` case, as far as I can tell.

And as for the 2nd, checking tags, isn't that what the code is doing?
Are you saying this should get folded back into `brightcove.py`'s `_extract_urls()` method, and then `bostonglobe.py` should import that? I don't see how to do that without adding `data-brightcove-video-id` to `brightcove.py`, which is what you want me not to do?

So, I'm sorry, I don't understand what you are asking for here.

yan12125: I was talking about the video on programme-tv.net. That's unrelated to bostonglobe.com.

> That doesn't work for the bostonglobe.com case

Of course it won't as these lines are for programme-tv.net

> and then bostonglobe.py should import that

Nope. bostonglobe.py requires special handling and is unrelated to _extract_urls()

> As a side note, this pattern is quite similar to #12005

That's just a reminder for myself but not requiring you to implement both patterns in a single pull request. I'm sorry if that's misleading. And apparently it's time to seperate codes for bostonglobe.com and programme-tv.net into two pull requests. Otherwise there might be more confusions.

johnhawkinson: Done, PR #12178

youtube_dl/extractor/bostonglobe.py
johnhawkinson: Done.

youtube_dl/extractor/bostonglobe.py
johnhawkinson: Done.

johnhawkinson:
I believe this is all set now.
I do have a reservation, which I expressed before:

> seems to … [be] more fragile (what happens when there are other kinds of videos on bostonglobe.com that the Generic extractor would have caught, but this one won't).

Indeed, I found some examples that use YouTube like
https://www.bostonglobe.com/lifestyle/names/2017/02/17/does-ben-affleck-play-matt-damon-favorite-version-batman/ruqkc9VxKBYmh5txn1XhSI/story.html
and
https://www.bostonglobe.com/sports/2017/02/18/racecar-that-runs-chocolate-may-not-sweet-idea/rIj2EAeCXZz3VM1w0obAaO/story.html
that the `GenericIE` supports but that support now breaks.  I don't know how to handle this. Extending bostonglobe.py to know explicitly about YouTube seems wrong. What if there are other types? Indeed
https://www.bostonglobe.com/sports/bruins/2017/02/18/ottawa-acquisition-mike-condon-pays-off/76G520N2pqH5bnbHYjRBfL/story.html
seems to be a Twitter embedded video, although this one the Generic extractor doesn't catch (a bug I would assume?)

Should I do what lcp.py does:
https://github.com/rg3/youtube-dl/blob/c58b7ffef43f60fa6a183c849cfdca42e36eae0c/youtube_dl/extractor/lcp.py#L71-L76

It seems a cheap trick, and doesn't seem to be done very often (but maybe it should be?)

I could talk about how I don't understand the code philosophy, but you knew that already :)

johnhawkinson:
It's been a while, I haven't heard anything. Is there anything I can do to speed this up?
Given no feedback, I went ahead and implemented the fallback to the `Generic` extractor that I asked about last month.

youtube_dl/extractor/bostonglobe.py
```diff
@@ -0,0 +1,62 @@
...
+            player_id   = attrs.get('data-player')
+            embed       = attrs.get('data-embed')
+
+            if video_id and account_id and player_id and embed:
+                entries.append(
+                    'http://players.brightcove.net/%s/%s_%s/index.html?videoId=%s'
+                    % (account_id, player_id, embed, video_id))
+
+        if len(entries) == 0:
+            return self.url_result(url, 'Generic')
```
yan12125: Any non-Brightcove example videoson bostonglobe.com? Please add them to _TESTS.

youtube_dl/extractor/bostonglobe.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def _real_extract(self, url):
+        page_id = self._match_id(url)
+        webpage = self._download_webpage(url, page_id)
+
+        # cribbed from generic.py
+        page_title = self._og_search_title(
+            webpage, default=None) or self._html_search_regex(
+            r'(?s)<title>(.*?)</title>', webpage, 'video title',
+            default='video')
```
yan12125: As we're now in a dedicated extractor, just choose a better one from either the value of ```og:title``` or ```<title>``` contents. This can make codes simpler.

youtube_dl/extractor/bostonglobe.py
```diff
@@ -0,0 +1,62 @@
...
+                'id': '5320421710001',
+                'ext': 'mp4',
+                'title': 'A tree finally succumbs to disease, leaving a hole in a neighborhood',
+                'description': 'It arrived as a sapling when the Back Bay was in its infancy, a spindly American elm tamped down into a square of dirt cut into the brick sidewalk of 1880s Marlborough Street, no higher than the first bay window of the new brownstone behind it.',
+                'timestamp': 1486877593,
+                'upload_date': '20170212',
+                'uploader_id': '245991542',
+            },
+        'md5': '0a62181079c85c2d2b618c9a738aedaf',
+        # HEAD requests produce 404 :(
```
yan12125: I guess this can be removed as the new extractor doesn't try HEAD before the actual request?

johnhawkinson:
Thanks, @yan12125. All set on your review
* I'm confused about where the `HEAD` requests are generated from, but you're right that it didn't cause a problem for the first test. But the second test I just added did have that problem (the YouTube embed). Weird.
* `_og_search_title()` seems less fragile so I went with that. Not sure how to choose when they're both valid. I don't see any guidance in the code on priority.
* Youtube example added.

yan12125:
Please fix coding styles.
```
$ flake8 
./youtube_dl/extractor/bostonglobe.py:8:1: F401 '..utils.orderedSet' imported but unused
./youtube_dl/extractor/bostonglobe.py:13:1: E302 expected 2 blank lines, found 1
./youtube_dl/extractor/bostonglobe.py:19:17: F601 dictionary key 'title' repeated with different values
./youtube_dl/extractor/bostonglobe.py:22:17: F601 dictionary key 'title' repeated with different values
./youtube_dl/extractor/bostonglobe.py:28:9: E122 continuation line missing indentation or outdented
./youtube_dl/extractor/bostonglobe.py:44:9: E123 closing bracket does not match indentation of opening bracket's line
./youtube_dl/extractor/bostonglobe.py:51:1: W293 blank line contains whitespace
./youtube_dl/extractor/bostonglobe.py:57:21: E221 multiple spaces before operator
./youtube_dl/extractor/bostonglobe.py:58:23: E221 multiple spaces before operator
./youtube_dl/extractor/bostonglobe.py:59:22: E221 multiple spaces before operator
./youtube_dl/extractor/bostonglobe.py:60:18: E221 multiple spaces before operator
./youtube_dl/extractor/generic.py:2497:20: F821 undefined name '_playlist_from_matches'
```

> I don't see any guidance in the code on priority.

If an approach works and is easy to understand, it's good. No priorities are enforced.

youtube_dl/extractor/bostonglobe.py
```diff
@@ -0,0 +1,72 @@
...
+            'info_dict': {
+                'title': "Who Is Matt Damon's Favorite Batman?",
+                'id': 'ZW1QCnlA6Qc',
+                'ext': 'mp4',
+                'upload_date': '20170217',
+                'description': 'md5:3b3dccb9375867e0b4d527ed87d307cb',
+                'uploader': 'The Late Late Show with James Corden',
+                'uploader_id': 'TheLateLateShow',
+            },
+            'md5': '582b40327089d5c0c949b3c54b13c24b',
```
yan12125: The convention here is put 'md5' before 'info_dict'

youtube_dl/extractor/bostonglobe.py
```diff
@@ -0,0 +1,72 @@
...
+                'title': 'A tree finally succumbs to disease, leaving a hole in a neighborhood',
+                'description': 'It arrived as a sapling when the Back Bay was in its infancy, a spindly American elm tamped down into a square of dirt cut into the brick sidewalk of 1880s Marlborough Street, no higher than the first bay window of the new brownstone behind it.',
+                'timestamp': 1486877593,
+                'upload_date': '20170212',
+                'uploader_id': '245991542',
+            },
+        'md5': '0a62181079c85c2d2b618c9a738aedaf',
+        },
+        {
+            'url': 'https://www.bostonglobe.com/lifestyle/names/2017/02/17/does-ben-affleck-play-matt-damon-favorite-version-batman/ruqkc9VxKBYmh5txn1XhSI/story.html',
```
yan12125: Could you add a comment that it's a YouTube embedded video?

youtube_dl/extractor/common.py
```diff
@@ -43,6 +43,7 @@
     GeoRestrictedError,
     GeoUtils,
     int_or_none,
+    orderedSet,
```
yan12125: Imported items should be sorted alphanumerically. Note that in this project symbols are case-insensitive during sorting.

(I know there are many places not sorted well; it would be great if an automation tool or someone can fix them...)

johnhawkinson:
All set, hopefully.

> If an approach works and is easy to understand, it's good. No priorities are enforced.

Well, it just seems like it's a bad plan to go from a flexible mechanism with a good fallback to a fragile mechanism that's more likely to break.

Maybe not this extractor or this website, but across the project, at some point title extraction is going to fail in some places. Why should the only the sites depending on the Generic extractor get the benefit of this?

Ah, well, philosophy :)

youtube_dl/extractor/common.py
yan12125: Sorry I may be inaccurate. Capitalized symbols are sorted together with non-capitalized symbols. For example, this is correct:
```
     error_to_compat_str,
     ExtractorError,
     fix_xml_ampersands,
     float_or_none,
     GeoRestrictedError,
```

yan12125:
> Why should the only the sites depending on the Generic extractor get the benefit of this?

For metadata, we want accuracy instead of flexible. In Bostonglobe there happens to be not much difference.

youtube_dl/extractor/common.py
johnhawkinson: I actually read you correctly but screwed up the implementation. I used `sort -i` instead of `sort -f` and didn't look sufficiently closely at the results before committing them, which is the real error. Fixed.

youtube_dl/extractor/bostonglobe.py
```diff
@@ -0,0 +1,72 @@
...
+
+from .common import InfoExtractor
+
+from ..utils import (
+    extract_attributes,
+)
+
+
+class BostonGlobeIE(InfoExtractor):
+    _VALID_URL = r'https?(?i)://(?:www\.)?bostonglobe\.com/.*/(?P<id>[^/]+)/\w+(?:\.html)?'
```
yan12125: Flags like ```(?i)``` must be in the start of a regular expression since Python 3.6. See http://bugs.python.org/issue22493
```
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-vF', 'test:bostonglobe']
[debug] youtube-dl version 2017.03.07
[debug] Git HEAD: 0f6b87d06
[debug] Python version 3.6.0 - Linux-4.10.1-1-ARCH-x86_64-with-arch
[debug] exe versions: ffmpeg 3.2.4, ffprobe 3.2.4, rtmpdump 2.4
[debug] Proxy map: {}
  File "<string>", line 23, in <module>
  File "/home/yen/Projects/youtube-dl/youtube_dl/__init__.py", line 459, in main
    _real_main(argv)
  File "/home/yen/Projects/youtube-dl/youtube_dl/__init__.py", line 449, in _real_main
  File "/home/yen/Projects/youtube-dl/youtube_dl/YoutubeDL.py", line 1883, in download
  File "/home/yen/Projects/youtube-dl/youtube_dl/YoutubeDL.py", line 752, in extract_info
    if not ie.suitable(url):
  File "/home/yen/Projects/youtube-dl/youtube_dl/extractor/common.py", line 365, in suitable
    cls._VALID_URL_RE = re.compile(cls._VALID_URL)
  File "/usr/lib/python3.6/re.py", line 233, in compile
    return _compile(pattern, flags)
  File "/usr/lib/python3.6/re.py", line 301, in _compile
    p = sre_compile.compile(pattern, flags)
  File "/usr/lib/python3.6/sre_compile.py", line 562, in compile
    p = sre_parse.parse(p, flags)
  File "/usr/lib/python3.6/sre_parse.py", line 856, in parse
    p = _parse_sub(source, pattern, flags & SRE_FLAG_VERBOSE, False)
  File "/usr/lib/python3.6/sre_parse.py", line 415, in _parse_sub
    itemsappend(_parse(source, state, verbose))
  File "/usr/lib/python3.6/sre_parse.py", line 743, in _parse
    DeprecationWarning, stacklevel=7
DeprecationWarning: Flags not at the start of the expression https?(?i)://(?:www\ (truncated)
```

johnhawkinson: Done. Also with a re-sorting of `common.py`'s import clause. Although I am confused:

> Flags like (?i) must be in the start of a regular expression since Python 3.6. See http://bugs.python.org/issue22493

Why does this not show up in the Travis output for Python 3.6?: https://travis-ci.org/rg3/youtube-dl/jobs/208950149
It sure would have been helpful to know.

Thanks again.

johnhawkinson:
> Last final issue! Could you squash commits?

Was I supposed to enter a comment after squashing? **Squashed.**

yan12125:
Merged. Thanks for the patience and the work!

PR #12158, opened 2017-02-16
Jkanime extractor

mcerdeira:
Jkanime.net added


dstftw:
youtube-dl does not support sites dedicated to [copyright infringement](https://github.com/rg3/youtube-dl#can-you-add-support-for-this-anime-video-site-or-site-which-shows-current-movies-for-free).

Issue #9163, opened 2016-04-11
Brightcove player not being detected


SeanPM5:
- [X] I've **verified** and **I assure** that I'm running youtube-dl **2016.04.06**
- [X] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [X] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
### What is the purpose of your _issue_?
- [X] Bug report (encountered problems with youtube-dl)
### Description of your _issue_, suggested solution and other information

Brightcove Player is not being detected on people.com (a major website, ranked 620 on alexa) and ew.com. I'm not sure if this falls under a site support request or bug, but I believe it's a bug because of what I mention at the bottom (and the fact that this player is detected fine on most other sites).

Example page(s):
http://www.people.com/people/videos/0,,20999362,00.html
http://www.people.com/people/videos/0,,20991602,00.html
http://www.people.com/article/greys-anatomy-exclusive-is-meredith-moving-on-mcdreamy
Many more video links: http://www.people.com/people/videos/
http://www.ew.com/article/2016/02/02/supernatural-sneak-peek-winchesters-jody-mills
http://www.ew.com/article/2016/03/11/bates-motel-sneak-peek-norman-threatens-norma

youtube-dl output:

```
iMac:~ Sean$ youtube-dl -v http://www.people.com/people/videos/0,,20991602,00.html
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://www.people.com/people/videos/0,,20991602,00.html']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.04.06
[debug] Python version 2.7.11 - Darwin-15.3.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg 3.0.1, ffprobe 3.0.1, rtmpdump 2.4
[debug] Proxy map: {}
[generic] 0,,20991602,00: Requesting header
WARNING: Falling back on generic information extractor.
[generic] 0,,20991602,00: Downloading webpage
[generic] 0,,20991602,00: Extracting information
ERROR: Unsupported URL: http://www.people.com/people/videos/0,,20991602,00.html
Traceback (most recent call last):
[...]

PR #12178, opened 2017-02-19
[Brightcove] Support <video> URLs, e.g. programme-tv

### LINKED ISSUE #12005: Site support request: www.programme-tv.net

johnhawkinson:
Fixes #12005 
Abstracted out from PR #12099


FelixSchwarz:
just a random reaction from IRC (I tend to agree without any authority on the project): could you squash these commits into logical chunks?

johnhawkinson:
Thanks, @FelixSchwarz, what IRC channel should I be following?

They should just squash down to one logical chunk -- there's the code and the test. Am I missing something? Would squashing them into one somehow aid in review?

I'm happy to squash them if that would assist in getting them committed, although I'm loathe to lose history while I'm still uncertain if they are ready to commit? Thanks.

FelixSchwarz:
> Thanks, @FelixSchwarz, what IRC channel should I be following?

#youtube-dl on freenode

> They should just squash down to one logical chunk -- there's the code and the test. Am I missing
> something? Would squashing them into one somehow aid in review?

Honestly - I don't know. I'm just a random youtube-dl user who knows a bit of Python and I was frustrated that there are so many open pull requests and some - like yours - did not receive any comments in several weeks.

So I mentioned this on IRC and another user suggested squashing. From a personal coding style I tend to agree. The resulting changes are quite small so personally I think the history for these few lines is not really important. If some of the lines are a result of many hours/subtle gotchas I recommend adding some code comments.

One thing which should definitively go is `# 	<video data-brightcove-video-id="5320421710001"` because that should be covered by tests.

johnhawkinson:
> So I mentioned this on IRC and another user suggested squashing. 

Unless that user was someone with commit access to this repo, I would be very reluctant. The youtube-dl maintainers often request squashing in a pull request, but only after they have approved the implementation in-concept. Since there hasn't been any such approval here, I would be reluctant to squash without direction, because it rewrites history and that is bad.

If you want to view this PR in a "squashed view", the  "Filed changed" button will do it for you: https://github.com/rg3/youtube-dl/pull/12178/files

> From a personal coding style I tend to agree.

I can't tell which position you agree with, but probably it's not mine :)

> If some of the lines are a result of many hours/subtle gotchas I recommend adding some code comments.

I believe the youtube-dl maintainers disagree. My coding style is to liberally write comments, and I am regularly asked to remove them from youtube-dl PRs.

> One thing which should definitively go is # <video data-brightcove-video-id="5320421710001" because that should be covered by tests.

No, the purpose of that comment is not related to the tests.
That comment is an example of the syntax that is then parsed by the subsequent block of code. The point of the comment is so that before you read the code, you can understand what it is trying to do, and what it is trying to parse. Perhaps it is unnecessary (I suspect maintainers may say so? But since I disagree I will wait for them to do so), but it is certainly not covered by tests.

yan12125:
See my previous idea: https://github.com/rg3/youtube-dl/pull/12099#discussion_r101892750. I'm not sure whether it's doable or not to merge the new detection approach and the existing one implemented in https://github.com/rg3/youtube-dl/blob/16e2c8f/youtube_dl/extractor/brightcove.py#L494-L509. It not please add a TODO comment and people can revisit it if the Brightcove API is changed in the future.

johnhawkinson:
> See my previous idea: https://github.com/rg3/youtube-dl/pull/12099#discussion_r101892750.

Oh hey! When you mentioned it there I didn't understand what you mean, but now I see it. Yes, that makes sense, *UNFORTUNATELY*:

> I'm not sure whether it's doable or not to merge the new detection approach and the existing one implemented in https://github.com/rg3/youtube-dl/blob/16e2c8f/youtube_dl/extractor/brightcove.py#L494-L509.

As far as I can tell, this code is not used by any tests.
I ran all the `brightcove.py` tests, as well as all the the Generic tests, with something like:

```
$ for i in `seq 1 100`; do echo Generic_$i; python test/test_download.py TestDownload.test_Generic_$i; done
```

I added a test print statement at the beginning of `_extract_urls()` (line 488), and also at the end of the section you highlighted (line 507). The former tripped 62 times, the latter none.

I'll try to look in the wider Internet, but that's sort of annoying.

johnhawkinson:
OK, @yan12125, how does that look? I did find some examples of the [2] -style Brightcoves. I added a test for one in Generic, and my other example was `http://www.darkreading.com/so-you-think-you-know-risk-management/v/d-id/1316742` (I assume it doesn't make sense to add two tests for the same thing? Esp. in Generic which has many many tests)

johnhawkinson:
Preemptively address `flake8` report; also Travis caught an error.
Although as I groused in 63108b8 it seems a shame to move from
```
            if not account_id: account_id  = attrs.get('data-account')
            if not player_id:  player_id   = attrs.get('data-player')
            if not embed:      embed       = attrs.get('data-embed')
```
to
```

            if not account_id:
                account_id = attrs.get('data-account')
            if not player_id:
                player_id = attrs.get('data-player')
            if not embed:
                embed = attrs.get('data-embed')
```

Am I crazy for finding the first one more readable by a lot? Oh well...

youtube_dl/extractor/brightcove.py
```diff
@@ -489,34 +491,56 @@ def _extract_url(webpage):
...
+            if not account_id:
+                account_id = attrs.get('data-account')
+            if not player_id:
+                player_id = attrs.get('data-player')
+            if not embed:
+                embed = attrs.get('data-embed')
+
+            # According to [5] data-video-id may be prefixed with 'ref:'
+            if video_id:
+                video_id = video_id.rpartition('ref:')[2]
```
dstftw: `ref:` should not be removed from video id.

youtube_dl/extractor/generic.py
```diff
@@ -1852,7 +1887,7 @@ def _playlist_from_matches(matches, getter=None, ie=None):
         # Look for Brightcove Legacy Studio embeds
         bc_urls = BrightcoveLegacyIE._extract_brightcove_urls(webpage)
         if bc_urls:
-            self.to_screen('Brightcove video detected.')
+            self.to_screen('Brightcove (legacy) video detected.')
```
dstftw: To be removed.

youtube_dl/extractor/brightcove.py
```diff
@@ -489,34 +491,56 @@ def _extract_url(webpage):
     @staticmethod
     def _extract_urls(webpage):
         # Reference:
-        # 1. http://docs.brightcove.com/en/video-cloud/brightcove-player/guides/publish-video.html#setvideoiniframe
+        # 1. http://docs.brightcove.com/en/video-cloud/brightcove-player/guides/publish-video.html#tag
         # 2. http://docs.brightcove.com/en/video-cloud/brightcove-player/guides/publish-video.html#setvideousingjavascript
-        # 3. http://docs.brightcove.com/en/video-cloud/brightcove-player/guides/embed-in-page.html
-        # 4. https://support.brightcove.com/en/video-cloud/docs/dynamically-assigning-videos-player
+        # 3. http://docs.brightcove.com/en/video-cloud/brightcove-player/guides/publish-video.html#setvideoiniframe
```
dstftw: What's the point of this reordering?

youtube_dl/extractor/brightcove.py
```diff
@@ -489,34 +491,56 @@ def _extract_url(webpage):
...
+        # 4. http://docs.brightcove.com/en/video-cloud/brightcove-player/guides/in-page-embed-player-implementation.html
+        # 5. https://support.brightcove.com/en/video-cloud/docs/dynamically-assigning-videos-player
+
+        # [1] looks like:
+        # <video data-video-id="5320421710001" data-account="245991542" data-player="SJWAiyYWg" data-embed="default" class="video-js" controls itemscope itemtype="http://schema.org/VideoObject">
```
dstftw: Why is this located here and not where it's actually extracted?

youtube_dl/extractor/generic.py
```diff
@@ -1869,6 +1904,7 @@ def _playlist_from_matches(matches, getter=None, ie=None):
         # Look for Brightcove New Studio embeds
         bc_urls = BrightcoveNewIE._extract_urls(webpage)
         if bc_urls:
+            self.to_screen('Brightcove (new) video detected.')
```
dstftw: To be removed.

youtube_dl/extractor/brightcove.py
johnhawkinson: As explained in 15859c6845b158d80e6e3db94198d57602ef333c:
>  [Brightcove:new] reorder doc references in order of HTML page
> Otherwise it is very confusing to read them out of order.

That is, the links mostly to #portions of the same webpage, but it's confusing to try to read them in a different order than the webpage is in.

youtube_dl/extractor/brightcove.py
johnhawkinson: Oversight. Will move.

youtube_dl/extractor/brightcove.py
johnhawkinson: This was pre-existing, but ref: was removed by regexp rather than explicitly. See 49571c1c2fc0872d8e8cf341cdeacd57d8885236:

```diff
-                # According to [4] data-video-id may be prefixed with ref:
-                r'''(?sx)
-                    <video[^>]+
-                        data-video-id=["\'](\d+|ref:[^"\']+)["\'][^>]*>.*?
```

Are you confident it won't break things to retain ref:?

youtube_dl/extractor/generic.py
johnhawkinson: OK. The intention was to be consistent with the other case…

youtube_dl/extractor/brightcove.py
dstftw: It's more confusing to see extraction to happen in different order.

youtube_dl/extractor/brightcove.py
dstftw: It's not removed: `(\d+|ref:[^"\']+)` captures number or string starting with `ref:`.

youtube_dl/extractor/generic.py
dstftw: The other case was some legacy code. I don't see much sense in such messages since it's clear what extractor is delegated to since all messages from the final extractor are prefixed with `IE_NAME`.

youtube_dl/extractor/brightcove.py
johnhawkinson: Indeed, I misread the regexp. Thanks.

youtube_dl/extractor/brightcove.py
johnhawkinson: It was never in order. Before 15859c6 it went [1], [2], [5], [3], [4].
Now it is [3], [1], [2], [4], [5].

So if anything it's more in order now?

dstftw: Before these changes there were 4 references and extraction followed them strictly in order of reference declaration.

johnhawkinson: I guess so…the code's not now so linear, since we've merged cases as Yen requested, so it doesn't quite go back that way.

Do you want me to put them back in the older order anyhow? Happy to, but it does seem more readable now than it was.

dstftw: At least from iframe extraction after video tag extraction.

johnhawkinson: Reordered.

youtube_dl/extractor/brightcove.py
```diff
@@ -501,22 +504,39 @@ def _extract_urls(webpage):
...
+
+        for video, script_tag, account_id, player_id, embed in re.findall(
+            r'''(?isx)
+                (<video[^>]+>)
+                (?:.*?
+                  (<script[^>]+
+                    src=["\'](?:https?:)?//players\.brightcove\.net/
+                    (\d+)/([^/]+)_([^/]+)/index(?:\.min)?\.js
+                  )
+                )?
```
dstftw: Another problem when match happens by video tag only is that there is actually no indication whether it's a brightcove embed or not at all. Only common attributes `data-video-id`, `data-account`, `data-player` and `data-embed` that technically may be used by others. Previously such indication was obtained by matching script tag with brightcove JS. There are some cases when script tag is located before video tag that were not detected previously. Do you have example URLs with video tags only and without brightcove indication? If not there should be added an additional check for presence of brightcove JS script on the page.

johnhawkinson: > Do you have example URLs with video tags only and without brightcove indication? 

This is a good point. I have two examples that have what I would call bad "indications" and I'm not sure how they qualify. From #9163 there's  http://ew.com/tv/2017/03/08/guardians-galaxy-cartoon-exclusive-clip/ which looks like this (adding line breaks):

```html
<!-- Start of Brightcove Player -->
<div class="video__wrapper">
<video
  data-account="219646971" data-player="default"
  data-video-id="5352121610001" data-embed="default" class="video-js"
  data-autoplay="false" data-viewplay="true" data-stickyplay="true"
  controls playsinline style="width: 100%; height: 100%;">
</video>
</div>
<!-- End of Brightcove Player -->
```

so the only indication is the comments surrounding it.

And also #12005 gives us http://www.programme-tv.net/videos/extraits/81095-guillaume-canet-evoque-les-rumeurs-d-infidelite-de-marion-cotillard-avec-brad-pitt-dans-vivement-dimanche/ which has lines 41-46:

```html
    	    <script>
		$script([
			'//players.brightcove.net/800000640001/HyenP6LuN_default/index.min.js',
			'//players.brightcove.net/800000640001/Hyx5mvd4_default/index.min.js'
		], 'brightcove')
	</script>
```

and then the video itself around line 467:

```html
<div class="video_partenaire id_brightcove_5311302538001 d-b p-r maxw-100">
    <div class="ratio-16-9">
        <video data-video-id="5311302538001"
               data-account="800000640001"
               data-player="HyenP6LuN"
               data-embed="default"
               class="video-js apply-ratio"
               autoplay preload controls>
	</video>
    </div>
</div>
```
So there's the `<script>` that doesn't fit the standard pattern (no `src=` attribute), although it's close, modulo the JQuery or whatever `$script` is. And the `id_brightcove_5311302538001` in the `class` attribute of the grandparent `<div>` . Sketchy.

And also there's the `[Bostonglobe]` case from #12099 which you don't consider appropriate for this extractor but we should at least think about, which has brightcove refences but no script reference (the script references are buried in javascript loaded by javascript loaded by javascript):

```html
<div class="videoplayer" id="video">
  <video data-brightcove-video-id="5320421710001" data-account="245991542" data-player="SJWAiyYWg" data-embed="default" class="video-js" controls itemscope itemtype="http://schema.org/VideoObject">
    <meta itemprop="name" content="A tree finally succumbs to disease, leaving a hole in a neighborhood" />
    <meta itemprop="thumbnailURL" content="http://www.bostonglobe.com/rf/image_130x80/Boston/2011-2020/2017/02/12/BostonGlobe.com/Metro/Videos/Sources/5320421710001.jpg" />
    <meta itemprop="contentURL" content="https://secure.brightcove.com/services/mobile/streaming/index/master.m3u8?videoId=5320421710001&amp;pubId=245991542&amp;secure=true" />
    <meta itemprop="width" content="867" />
    <meta itemprop="height" content="486" />
  </video>
</div>
```
That is identifyable both by the `data-brightcove-video-id` attribute and also the presence of the `<meta itemprop="contentURL">` that with a Brightcove URL (Huh, hadn't noticed that before).

So, do you consider any of the above indications sufficient to check for?

Maybe the right answer is to move the search for `<video>`  tags into `generic.py`, and then call a series of methods from extractors (like `._extract_urls()` but different) to determine if they want to try to retrieve (perhaps the Brightcove methods would return `Yes` in cases where the standard `<script ... players.brightcove.net>` was present, and `Maybe` in the other cases. Then `GenericIE` could call the `Yes` extractor(s) if there is one (any), or iterate over the `Maybe`s if not.

This also gets tricky if we allow for the possibility of multiple videos on a single page, esp. from different embedding sites.

In practice, does youtube-dl support any other extractors for `<video>` tags with this kind of ambiguity, and are any of them likely embedable in different domains like Brightcove is?

`CrackedIE` has a straight-up `<video src="/path/to/file.mp4">`, as does `VimeoIE`; `LifeNewsIE` has some similar support though the tests don't include those cases, `PornHubIE`, and `XHamsterIE` too. `TwitterIE` deals with `<video class="animated-gif">`, although I didn't find a test that used it.

`LibraryOfCongressIE` has a `data-download` (though the extractor seems not to use it, relying on a uuid?). `SVTPlayIE` uses `data-video-id`. 

`VideoMegaIE` uses it to get a thumbnail, only.

Oh, and I guess `_parse_html5_media_entries()` also deals with `<video>` tags, but I grepped for `'<video'` and that missed `r'(?s)(<(?P<tag>video|audio)(`,. I'm not quite sure of the applicability of that...

Oh, also `RTSIE`, maybe?

dstftw: >So, do you consider any of the above indications sufficient to check for?

No, I don't think any of non standard embed code should be considered for generic embeds.

>Maybe the right answer is to move the search for `<video>`  tags into `generic.py`, and then call a series of methods from extractors (like `._extract_urls()` but different) to determine if they want to try to retrieve (perhaps the Brightcove methods would return `Yes` in cases where the standard `<script ... players.brightcove.net>` was present, and `Maybe` in the other cases. Then `GenericIE` could call the `Yes` extractor(s) if there is one (any), or iterate over the `Maybe`s if not.

Don't know if I get this right but one possible solution is to separate brightcove video tag only extraction (without brightcove indication) to another *ambiguous* method in `BrightcoveNewIE` and call it very last in generic extractor as a last resort. This will allow not to accidentally shadow another extractors' embeds but still detect such video only tag brigthcove embeds. Though this still may produce false positives. Another solution is to prohibit such generic embeds at all and add a separate extractor for each such case.

dstftw: Looks like such embeds are quite possible:
https://publicwww.com/websites/%22%3Cvideo%22+%22data-video-id%22+%22data-account%22+%22data-player%22+%22data-embed%22+-brightcove/

johnhawkinson: > one possible solution is to separate brightcove video tag only extraction (without brightcove indication) to another ambiguous method in BrightcoveNewIE and call it very last in generic extractor as a last resort. 

This is similar to what I was suggesting. Your suggestion is simpler but less general.
I guess the question is really what we can expect to see in the future from other embedded players that use `<video>` tags like this. Should we plan ahead and produce a scheme for them that will work reasonably (my suggestion), or should we only handle Brightcove today and worry about generalizing it later (yours).

I'm not sure what you mean by _ambiguous_, though.

One consequence of this separation is that it basically undoes the merging that @Yan12125 had suggested I implement in https://github.com/rg3/youtube-dl/pull/12099#discussion_r101892750.
It may be that we are letting "best" be the enemy of "good enough."

Ah well…

> Another solution is to prohibit such generic embeds at all and add a separate extractor for each such case.

This doesn't seem good -- the examples we see here are all about Brightcove, and trying to add them site by site is basically maintaining a list of websites that do this and that list will always be incomplete and annoying to maintain.

PR #12222, opened 2017-02-22
[srgssr] Migrate to integrationlayer version 2.0 API

goggle:
## Before submitting a *pull request* make sure you have:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description

The SRGSSR Play websites now often uses the integrationlayer version 2.0
API instead of version 1.0. I have modified the SRGSSR information
extractor to use this new integrationlayer instead of the old one. All
the old media supports this new version too, so there is no need to
stick with the old one. It's possible, that the support for the old
integrationlayer will be dropped, so this switch has to made anyway.

Here is a list of the changes:
 - Use integrationlayer version 2.0 API instead of version 1.0.
 - Assure consistant media IDs. In the old version of the information
   extractor, youtube-dl extracts the same video for the urls
   "http://www.srf.ch/play/tv/schweiz-aktuell/video/schweiz-aktuell-vom-22-02-2017?id=d0206674-6125-49ef-b85d-3cf36d24d582"
   and
   "http://www.srf.ch/play/tv/schweiz-aktuell/video/walliser-baubaubranche-wehrt-sich?id=967590f0-f812-4941-8f6a-06a2db7bd083",
   but uses different media IDs. Now it still extracts the same videos
   (since there is no support to cut videos into parts in youtube-dl,
   right?), but it uses the same media IDs. So we always have consistant
   media IDs for the media.
 - Add extraction of media duration.
 - Add extraction of video subtitles.
 - Use multiline regular expressions for `_VALID_URL` for better
   readability.
 - Indicate direct podcast downloads in `format_id`.
 - Update tests.

### DIFF
--- a/youtube_dl/extractor/srgssr.py
+++ b/youtube_dl/extractor/srgssr.py
@@ -124,7 +124,6 @@ class SRGSSRIE(InfoExtractor):
 
         return subtitles
 
-
     def _real_extract(self, url):
         bu, media_type, url_id = re.match(self._VALID_URL, url).groups()
 
@@ -243,18 +242,18 @@ class SRGSSRPlayIE(InfoExtractor):
             'ext': 'mp4',
             'title': 'Schwander, Rot-Grüne Stadtpolitik, Min Li Marti, Tamilen-Kirche',
             'description': 'Verbissener Kampf / Vertreibung der Büezer / Theke: Min Li Marti / Geldsegen für den Pastor',
-            'duration':  2630.0,
+            'duration': 2630.0,
             'upload_date': '20170208',
             'timestamp': 1486583589,
             'thumbnail': r're:^https?://.*1486587225\.png$',
             'subtitles': {
                 'de': [{
-                            'ext': 'ttml',
-                            'url': 're:^https://.*\.ttml$',
-                        }, {
-                            'ext': 'vtt',
-                            'url': 're:^https://.*\.vtt$',
-                        }]
+                    'ext': 'ttml',
+                    'url': 're:^https://.*\.ttml$',
+                }, {
+                    'ext': 'vtt',
+                    'url': 're:^https://.*\.vtt$',
+                }]
             },
         },
     }, {
--- a/youtube_dl/extractor/srgssr.py
+++ b/youtube_dl/extractor/srgssr.py
@@ -7,13 +7,36 @@ from .common import InfoExtractor
 from ..compat import compat_urllib_parse_urlparse
 from ..utils import (
     ExtractorError,
+    float_or_none,
     parse_iso8601,
+    str_or_none,
     qualities,
 )
 
 
 class SRGSSRIE(InfoExtractor):
-    _VALID_URL = r'(?:https?://tp\.srgssr\.ch/p(?:/[^/]+)+\?urn=urn|srgssr):(?P<bu>srf|rts|rsi|rtr|swi):(?:[^:]+:)?(?P<type>video|audio):(?P<id>[0-9a-f\-]{36}|\d+)'
+    _VALID_URL = r'''(?x)
+                    (?:
+                        https?://tp\.srgssr\.ch/p(?:/[^/]+)+\?urn=urn|
+                        srgssr
+                    ):
+                    (?P<bu>
+                        srf|
+                        rts|
+                        rsi|
+                        rtr|
+                        swi
+                    ):(?:[^:]+:)?
+                    (?P<type>
+                        video|
+                        audio
+                    ):
+                    (?P<id>
+                        [0-9a-f\-]{36}|
+                        \d+
+                    )
+                    '''
+
     _GEO_BYPASS = False
     _GEO_COUNTRIES = ['CH']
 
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/srgssr.py
```diff
@@ -36,133 +59,296 @@ def _get_tokenized_src(self, url, video_id, format_id):
...
+                            \d+
+                        )
+                    '''
     _TESTS = [{
+        # ID in url not the same as media ID, no Save button, no description
         'url': 'http://www.srf.ch/play/tv/10vor10/video/snowden-beantragt-asyl-in-russland?id=28e1a57d-5b76-4399-8ab3-9097f071e6c5',
         'md5': 'da6b5b3ac9fa4761a942331cef20fcb3',
         'info_dict': {
-            'id': '28e1a57d-5b76-4399-8ab3-9097f071e6c5',
+            'id': '0d181f8f-07ae-46b2-8d3b-7619c0efb0e3',
```
dstftw: Ids must stay intact.

youtube_dl/extractor/srgssr.py
```diff
@@ -36,133 +59,296 @@ def _get_tokenized_src(self, url, video_id, format_id):
...
             'ext': 'mp4',
+            'title': '10vor10 vom 01.07.2013',
```
dstftw: Title is invalid.

youtube_dl/extractor/srgssr.py
goggle: I can revert this, but then the two urls
 - http://www.srf.ch/play/tv/10vor10/video/snowden-beantragt-asyl-in-russland?id=28e1a57d-5b76-4399-8ab3-9097f071e6c5
 - http://www.srf.ch/play/tv/10vor10/video/newsflash?id=493b764e-355b-440a-9257-23260a48a217
will extract exactly the same video, but with different video IDs. Would that be expected behavior?

dstftw: That's completely different videos.

youtube_dl/extractor/srgssr.py
goggle: Why is the title invalid? youtube-dl extracts the whole episode of the show "10vor10" for the given url. The previously mentioned title "Snowden beantragt Asyl in Russland" will be only a small part of the video.

It would be nice to extract these segments from the video, but that would only be possible by postprocessing the downloaded video by cutting it from a given starting point to a given ending point. Does youtube-dl support that?

youtube_dl/extractor/srgssr.py
goggle: Yes, but youtube-dl extracts the whole episode of these videos (even before my commit), and not the single segments, because there is no explicit download source for the single segments...

youtube_dl/extractor/srgssr.py
dstftw: Cause http://www.srf.ch/play/tv/10vor10/video/snowden-beantragt-asyl-in-russland?id=28e1a57d-5b76-4399-8ab3-9097f071e6c5 is a link to a segment not the whole episode.

goggle: That's true, but why does youtube-dl download the whole episode of the show with the title and ID of the segment, and not just the segment?

Is it technically not easily possible to extract segments from sources which refer to a whole episode?

youtube_dl/extractor/srgssr.py
goggle: Using youtube-dl with the above mentioned urls leads to exactly the same video file:
```
~/test $ youtube-dl --version
2017.02.22
~/test $ youtube-dl "http://www.srf.ch/play/tv/10vor10/video/snowden-beantragt-asyl-in-russland?id=28e1a57d-5b76-4399-8ab3-9097f071e6c5" -q
~/test $ youtube-dl "http://www.srf.ch/play/tv/10vor10/video/newsflash?id=493b764e-355b-440a-9257-23260a48a217" -q
~/test $ ls -l *.mp4
-rw-r--r-- 1 alex wheel 369252868 23. Feb 15:10  Newsflash-493b764e-355b-440a-9257-23260a48a217.mp4
-rw-r--r-- 1 alex wheel 369252868 23. Feb 15:09 'Snowden beantragt Asyl in Russland-28e1a57d-5b76-4399-8ab3-9097f071e6c5.mp4'
~/test $ diff Newsflash-493b764e-355b-440a-9257-23260a48a217.mp4 Snowden\ beantragt\ Asyl\ in\ Russland-28e1a57d-5b76-4399-8ab3-9097f071e6c5.mp4 
```

youtube_dl/extractor/srgssr.py
remitamine: i think that can be fixed by followning what's been done in the website player.
for http://www.srf.ch/play/tv/10vor10/video/snowden-beantragt-asyl-in-russland?id=28e1a57d-5b76-4399-8ab3-9097f071e6c5 the player will request the m3u8 url with `start=532.255&end=646.082` appended to the url while in http://www.srf.ch/play/tv/10vor10/video/newsflash?id=493b764e-355b-440a-9257-23260a48a217 it uses `start=1183.96&end=1228.52` so the manifest urls and duration are the ones that should be changed not the other video information(id, title...).

goggle: Oh, that's cool. Can this be done for both, the m3u8- and f4m-url in the same way? I will try to implement this. 

For direct links to video files that's not possible, right?

goggle:
I have fixed the problem of "different IDs for the same content".
List of changes:
 - Revert to the old behavior of extracting media IDs.
 - Support segmented videos (extract only the relevant parts of the
   whole episode).
 - Reduce verbosity of `_VALID_URL.`

PR #12240, opened 2017-02-23
Addition to README.md

SANGERA2:
Added advice to line 135 regarding not using min-filesize to try and download longer videos only based on their file-size. It will select just the long videos to download and will download the video track, then it will not download the audio track as it's too small and you end up with an un-merged file. Using duration in match-filter works fine however.

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Small addition to README.md. Added advice to line 135 regarding not using min-filesize to try and download longer videos only based on their file-size. It will select just the long videos to download and will download the video track, then it will not download the audio track as it's too small and you end up with an un-merged file. Using duration in match-filter works fine however.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,11 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
                                      caseless sub-string)
     --max-downloads NUMBER           Abort after downloading NUMBER files
     --min-filesize SIZE              Do not download any videos smaller than
-                                     SIZE (e.g. 50k or 44.6m)
+                                     SIZE (e.g. 50k or 44.6m). Use match-filter
+                                     with the duration key instead of this to
+                                     filter for long videos only as this will
+                                     often download the video track and skip
+                                     the audio if they are downloaded separately.
     --max-filesize SIZE              Do not download any videos larger than SIZE
                                      (e.g. 50k or 44.6m)
     --date DATE                      Download only videos uploaded in this date


README.md
```diff
@@ -132,7 +132,11 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
                                      caseless sub-string)
     --max-downloads NUMBER           Abort after downloading NUMBER files
     --min-filesize SIZE              Do not download any videos smaller than
-                                     SIZE (e.g. 50k or 44.6m)
+                                     SIZE (e.g. 50k or 44.6m). Use match-filter
+                                     with the duration key instead of this to
+                                     filter for long videos only as this will
+                                     often download the video track and skip
+                                     the audio if they are downloaded separately.
```
dstftw: 1. This file is autogenerated and should not be touched by hand.
2. `--min-filesize` works differently. `--match-filter` is based on filesize metadata while `--min-filesize` uses the actual filesize obtained via `Content-Length` HTTP header.

SANGERA2: Can you tell me which file I should edit in order for this to be regenerated correctly? I've just made a change to options.py to reflect the same thing.

2 & 3 are very good points. But I'm sure there are other people who are as silly/lazy as me who try to use the simple option of filtering by file-size instead of duration within match-filter! This added text is here to warn them off that before they waste an hour of their lives trying to figure out why they are only getting the video track! :)

Thanks.

dstftw: This is the place for commands' descriptions not suggestions or write-ups. One who use this should be aware of what he is doing.

Issue #10993, opened 2016-10-21
Use /tmp directory for frag and .vtt files


asgh:
I sometimes see leftover `--Frag17` and `-.ru.vtt` files in my home directory. I assume it's from me interrupting youtube-dl during a download.

Can you use the /tmp (`TMPDIR`) instead for those files?

jesus2099:
Or a parameter.
I would like to use my `%TEMP%` folder for those fragements, as it’s on an SSD, faster drive.

yan12125:
Related: #11054

Stonedestroyer:
The way I want this to work is that you can specify the intermediate folder, what are the plans on this? As it would be good if it also could move after task done or after each video and you could decide.

lvmm:
I would also like to see a configurable temporary file folder, audio+video merging would work so much faster if source files were on a separate filesystem.

nyuszika7h:
I would like to see this as well, it would be useful to download videos to a folder monitored by Serviio without manually moving them. youtube-dl sometimes fails to delete the temporary files after merging because Serviio opens them for a few seconds.

sebastianhelbig:
Using the current working directory for temporary files is never a good idea. youtube-dl should at least use the temporary folder defined by the system.

PR #12274, opened 2017-02-26
Add post processor to move after completion

balasankarc:

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

This PR provides a post-processor that will easily move the completed files to a specific destination. Consider this use case
1. User uses a Raspberry Pi with limited storage as a head-less video downloader
2. User periodically backup Downloads folder to a NAS
3. Only completed videos should go into Downloads folder, else chances are there the backup script to be ran when downloading is going on, thus copying partial file and removing it from source

This PR, essentially solves point 3 of the above. The files are moved to the specified folder **only after its completion**. 

I understand that there may be already ways to do this, using a combination of options. But, this seems to be a general use case. Also, this PR is created for scratching my personal itch, and I, being a FOSS enthusiast, believe in contributing back to the source. So, reviews/criticisms/rejections welcome. :)

### DIFF
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -290,6 +290,11 @@ def _real_main(argv=None):
             'key': 'ExecAfterDownload',
             'exec_cmd': opts.exec_cmd,
         })
+    if opts.move:
+        postprocessors.append({
+            'key': 'Move',
+            'destination': opts.move,
+        })
     external_downloader_args = None
     if opts.external_downloader_args:
         external_downloader_args = compat_shlex_split(opts.external_downloader_args)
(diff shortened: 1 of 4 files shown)


yan12125:
I guess ```--exec 'mv {} /destination/path/'``` already works?

balasankarc:
@yan12125 Of course it does. But, as I said, this seemed like a general use case and good enough to be added into the code base as a dedicated option. :slightly_smiling_face: 

PS: Users need not know `mv` command and it's counterparts in other OSs.

yan12125:
In general it's better not to add more options if existing options suffice. There are already too many options.

And yes, indeed there are cases that intermediate files and the final result should be placed in different folders. A more generic approach can be found at #10993. Anyway thanks for the contribution!

balasankarc:
@yan12125 Sure. Makes sense. :slightly_smiling_face:

PR #12304, opened 2017-02-28
[vier] Add support for vijf.be

vierbergenlars:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

vier.be and vijf.be run the same CMS and are property of the same company (SBS Belgium),
so the same extractor can be used for both of them.

Is this way of adding support for another (but closely related) site to an existing extractor acceptable, or is it preferred to create a separate extractor for this site? In that case it would be a lot of unnecessary code duplication.

### DIFF
--- a/youtube_dl/extractor/vier.py
+++ b/youtube_dl/extractor/vier.py
@@ -9,7 +9,7 @@ from .common import InfoExtractor
 
 class VierIE(InfoExtractor):
     IE_NAME = 'vier'
-    _VALID_URL = r'https?://(?:www\.)?vier\.be/(?:[^/]+/videos/(?P<display_id>[^/]+)(?:/(?P<id>\d+))?|video/v3/embed/(?P<embed_id>\d+))'
+    _VALID_URL = r'https?://(?:www\.)?(?P<site>vier|vijf)\.be/(?:[^/]+/videos/(?P<display_id>[^/]+)(?:/(?P<id>\d+))?|video/v3/embed/(?P<embed_id>\d+))'
     _TESTS = [{
         'url': 'http://www.vier.be/planb/videos/het-wordt-warm-de-moestuin/16129',
         'info_dict': {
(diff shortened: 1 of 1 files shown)


docs/supportedsites.md
```diff
@@ -877,8 +877,8 @@
  - **vidme:user**
  - **vidme:user:likes**
  - **Vidzi**
- - **vier**
- - **vier:videos**
+ - **vier**: vier.be and vijf.be
+ - **vier:videos**: vier.be and vijf.be all videos of a program
```
dstftw: This file is auto generated.

youtube_dl/extractor/vier.py
```diff
@@ -105,13 +126,13 @@ def _real_extract(self, url):
         entries = []
         for current_page_id in itertools.count(start_page):
             current_page = self._download_webpage(
-                'http://www.vier.be/%s/videos?page=%d' % (program, current_page_id),
+                'http://www.%s.be/%s/videos?page=%d' % (site, program, current_page_id),
                 program,
                 'Downloading page %d' % (current_page_id + 1))
             page_entries = [
-                self.url_result('http://www.vier.be' + video_url, 'Vier')
+                self.url_result('http://www.'+site+'.be' + video_url, site.title())
```
dstftw: Wrong. Extractor key is `Vier`.

dstftw:
Check code with flake8.

youtube_dl/extractor/vier.py
vierbergenlars: Oops, I didn't get that, and the tests did not catch that. I guess Idid a too literal search and replace operation.

youtube_dl/extractor/vier.py
```diff
@@ -105,13 +126,13 @@ def _real_extract(self, url):
...
             current_page = self._download_webpage(
-                'http://www.vier.be/%s/videos?page=%d' % (program, current_page_id),
+                'http://www.%s.be/%s/videos?page=%d' % (site, program, current_page_id),
                 program,
                 'Downloading page %d' % (current_page_id + 1))
             page_entries = [
-                self.url_result('http://www.vier.be' + video_url, 'Vier')
+                self.url_result('http://www.' + site + '.be' + video_url, 'Vier')
                 for video_url in re.findall(
-                    r'<h3><a href="(/[^/]+/videos/[^/]+(?:/\d+)?)">', current_page)]
```
dstftw: `<h3>` is intentional.

vierbergenlars: You are right, it does seem to match the videos twice. I did not catch that issue.

vijf.be seems to have a `<h2>` tag in front of the video URL though. Is it warranted to change the regex to `<h[23]>`, or is this the point where I should build a separate extractor?

dstftw: Since video ids seems unique across both sites I see no point separating extractors.

youtube_dl/extractor/vier.py
```diff
@@ -105,13 +126,13 @@ def _real_extract(self, url):
...
+                    r'<h[2|3]><a href="(/[^/]+/videos/[^/]+(?:/\d+)?)">', current_page)]
```
dstftw: It should not match `h|`.

vierbergenlars: Definitely not. I confused some regex syntax here. (`[23]` and `(2|3)`)

PR #12307, opened 2017-02-28
Fix "'str' object does not support item assignment"

rndusr:
_find_jwplayer_data() assumes setup() is called with JSON, but it (sometimes?)
gets a variable that contains that expected JSON object.

This commit detects that by check for the first character being "{" and, if not,
resolving the variable to the JSON string.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -2161,18 +2161,24 @@ class InfoExtractor(object):
                     })
         return formats
 
-    @staticmethod
-    def _find_jwplayer_data(webpage):
+    def _find_jwplayer_data(self, webpage, video_id=None, transform_source=js_to_json):
         mobj = re.search(
             r'jwplayer\((?P<quote>[\'"])[^\'" ]+(?P=quote)\)\.setup\s*\((?P<options>[^)]+)\)',
             webpage)
         if mobj:
-            return mobj.group('options')
+            try:
+                jwplayer_data = self._parse_json(mobj.group('options'),
+                                                 video_id=video_id,
+                                                 transform_source=transform_source)
+            except ExtractorError:
+                pass
+            else:
+                if isinstance(jwplayer_data, dict):
+                    return jwplayer_data
 
     def _extract_jwplayer_data(self, webpage, video_id, *args, **kwargs):
-        jwplayer_data = self._parse_json(
-            self._find_jwplayer_data(webpage), video_id,
-            transform_source=js_to_json)
+        jwplayer_data = self._find_jwplayer_data(
+            webpage, video_id, transform_source=js_to_json)
         return self._parse_jwplayer_data(
             jwplayer_data, video_id, *args, **kwargs)
 
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -972,6 +972,20 @@ class GenericIE(InfoExtractor):
                 'thumbnail': r're:^https?://.*\.jpg$',
             },
         },
+        {
+            # JWPlayer config passed as variable
+            'url': 'http://www.txxx.com/videos/3326530/ariele/',
+            'info_dict': {
+                'id': '3326530_hq',
+                'ext': 'mp4',
+                'title': 'ARIELE | Tube Cup',
+                'uploader': 'www.txxx.com',
+                'age_limit': 18,
+            },
+            'params': {
+                'skip_download': True,
+            }
+        },
         # rtl.nl embed
         {
             'url': 'http://www.rtlnieuws.nl/nieuws/buitenland/aanslagen-kopenhagen',
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -2518,14 +2518,10 @@ class GenericIE(InfoExtractor):
                 self._sort_formats(entry['formats'])
             return self.playlist_result(entries)
 
-        jwplayer_data_str = self._find_jwplayer_data(webpage)
-        if jwplayer_data_str:
-            try:
-                jwplayer_data = self._parse_json(
-                    jwplayer_data_str, video_id, transform_source=js_to_json)
-                return self._parse_jwplayer_data(jwplayer_data, video_id)
-            except ExtractorError:
-                pass
+        jwplayer_data = self._find_jwplayer_data(
+            webpage, video_id, transform_source=js_to_json)
+        if jwplayer_data:
+            return self._parse_jwplayer_data(jwplayer_data, video_id)
 
         def check_video(vurl):
             if YoutubeIE.suitable(vurl):
--- a/youtube_dl/extractor/tvnoe.py
+++ b/youtube_dl/extractor/tvnoe.py
@@ -31,9 +31,8 @@ class TVNoeIE(InfoExtractor):
             r'<iframe[^>]+src="([^"]+)"', webpage, 'iframe URL')
 
         ifs_page = self._download_webpage(iframe_url, video_id)
-        jwplayer_data = self._parse_json(
-            self._find_jwplayer_data(ifs_page),
-            video_id, transform_source=js_to_json)
+        jwplayer_data = self._find_jwplayer_data(
+            ifs_page, video_id, transform_source=js_to_json)
         info_dict = self._parse_jwplayer_data(
             jwplayer_data, video_id, require_title=False, base_url=iframe_url)
 


yan12125:
Any example?

rndusr:
[NSFW] http://www.txxx.com/videos/3326530/ariele/

yan12125:
OK. Please add a test to generic.py

rndusr:
Well, I tried...

But I have no idea how that _TESTS format works and how to run the tests defined
by it. Running all tests takes forever and many of them fail.

yan12125:
> But I have no idea how that _TESTS format works and how to run the tests defined by it.

FYI: I have a helper script that determines the test ID at https://github.com/yan12125/youtube-dl/blob/helper-dev-scripts/devscripts/contrib/find_url.py
```
$ python ./devscripts/contrib/find_url.py http://www.txxx.com/videos/3326530/ariele/ 
test_Generic_104

$ python test/test_download.py TestDownload.test_Generic_104 
[generic] ariele: Requesting header
[generic] ariele: Downloading webpage
[generic] ariele: Extracting information
[info] Writing video description metadata as JSON to: test_Generic_104_3326530_hq.info.json
.
Ran 1 test in 1.780s

OK
```

rndusr:
Thanks. Hopefully I can remember that until next time.

The test seems to pass now. Anything else?

youtube_dl/extractor/common.py
```diff
@@ -2167,7 +2167,14 @@ def _find_jwplayer_data(webpage):
             r'jwplayer\((?P<quote>[\'"])[^\'" ]+(?P=quote)\)\.setup\s*\((?P<options>[^)]+)\)',
             webpage)
         if mobj:
-            return mobj.group('options')
+            options = mobj.group('options').strip()
+            # We're looking for JSON
+            if options[0] != '{':
+                # setup() got a variable as an argument - resolve it
+                regex = r'%s\s*=\s*(?P<options>\{[^\}]\})' % options
```
yan12125: 1. ```options``` should be escaped - ```re.escape```
2. Nested variables won't be detected. For example: (from https://developer.jwplayer.com/jw-player/docs/developer-guide/customization/configuration-reference/)
```
var config = {
    "playlist": [{
        "file": "/assets/sintel.mp4",
        "image": "/assets/sintel.jpg",
        "title": "Sintel Trailer",
        "mediaid": "ddra573"
    },{
        "file": "/assets/bigbuckbunny.mp4",
        "image": "/assets/bigbuckbunny.jpg",
        "title": "Big Buck Bunny Trailer",
        "mediaid": "ddrx3v2"
    }]
};
```

youtube_dl/extractor/common.py
```diff
@@ -2167,7 +2167,14 @@ def _find_jwplayer_data(webpage):
...
+                mobj = re.search(regex, webpage)
+                if mobj:
```
yan12125: If there's no match here, options should still be returned as it might be really a valid JSON. For example:
```
jwplayer('foobar').setup(
    /* some comments */
    {
        'file': 'foobar.mp4'
    }
);
```

youtube_dl/extractor/common.py
```diff
@@ -2167,7 +2167,14 @@ def _find_jwplayer_data(webpage):
             r'jwplayer\((?P<quote>[\'"])[^\'" ]+(?P=quote)\)\.setup\s*\((?P<options>[^)]+)\)',
             webpage)
         if mobj:
-            return mobj.group('options')
+            options = mobj.group('options').strip()
+            # We're looking for JSON
+            if options[0] != '{':
```
yan12125: If ```options[0]``` _is_ ```{```, options should be returned.

rndusr:
Thank you for your comments. I think I fixed two of the things you mentioned,
but the nested variables is tricky, especially since I don't have a test URL for
it. Do you have one? Is this a blocking issue?

yan12125:
There's one in generic.py marked as "Complex jwplayer" (test_Generic_62)

rndusr:
But that already works.

```
% python3 test/test_download.py TestDownload.test_Generic_62
[generic] videos: Requesting header
[generic] videos: Downloading webpage
[generic] videos: Extracting information
[info] Writing video description metadata as JSON to: test_Generic_62_videos.info.json
[debug] Invoking downloader on 'http://cdn.dbolical.com/cache/videos/games/1/50/49678/encode720p_mp4/king-machine-trailer.mp4'
[download] Destination: test_Generic_62_videos.mp4
.
Ran 1 test in 1.982s

OK
```

yan12125:
I've changed my last comment. Email clients won't receive the update, could you check it out on Github?

rndusr:
Thanks for the headsup.

So you won't merge this for now? Why? It fixes a known bug. If another URL
doesn't work, that's different bug that should be taken care of when it happens,
not weeks/months/years before.

I'm also pretty sure that the perfect regex does not exist. There will always be
some way to break it. You can't parse JS with a regex.

youtube_dl/extractor/common.py
```diff
@@ -2167,7 +2167,16 @@ def _find_jwplayer_data(webpage):
...
+                # setup() got a variable as an argument - resolve it
+                regex = r'%s\s*=\s*(?P<options>\{[^\}]\})' % re.escape(options)
+                mobj = re.search(regex, webpage)
```
yan12125: Sorry for previous noises. I just actually checked the HTML source of your example and found that this PR doesn't work as expected. ("resolving the variable to the JSON string") Try to print the value of ```mobj``` after this line.

rndusr:
Damn. Sorry for the crappy pull request.

It looks like if _find_jwplayer_data() returns None, it falls back to looking
for "gorilla-vid style embedding" (line 2531 in generic.py), which finds the
correct video URL in my example. So all my PR does is returning None instead of
"jwsettings" (the variable name), which works but not as intended.

The only practical solution I see is to check whether `options` contains proper
JSON. Something like this:

    if isinstance(json.loads(js_to_json(options)), dict):
        return options

Do you have a better idea?

yan12125:
Checking whether it's valid JSON or not sounds good. FYI: ```_parse_json()``` can make it simpler

youtube_dl/extractor/common.py
```diff
@@ -2161,13 +2161,23 @@ def _extract_wowza_formats(self, url, video_id, m3u8_entry_protocol='m3u8_native
...
         mobj = re.search(
             r'jwplayer\((?P<quote>[\'"])[^\'" ]+(?P=quote)\)\.setup\s*\((?P<options>[^)]+)\)',
             webpage)
         if mobj:
-            return mobj.group('options')
+            options = mobj.group('options').strip()
+
+            def is_json(string):
+                try:
+                    dct = self._parse_json(options, video_id=None, transform_source=js_to_json)
```
yan12125: Minor issue: there's no need to assign a variable if it's not used.

youtube_dl/extractor/generic.py
```diff
@@ -1507,6 +1507,20 @@ class GenericIE(InfoExtractor):
...
+                'id': '3326530_hq',
+                'ext': 'mp4',
+                'title': 'ARIELE | Tube Cup',
+                'uploader': 'www.txxx.com',
+                'age_limit': 18,
+            },
+            'params': {
+                'skip_download': True,
+            }
+        },
```
yan12125: And it's better to place jwplayer-related tests together

rndusr:
_parse_json() returns the same string in quotes if it's not a dictionary. But we
want a dictionary, which is checked if no exception is raised by _parse_json().

I guess I should've named the function "is_json_dict" instead.

yan12125:
After thinking again I guess we can go even further - making ```_find_jwplayer_data``` return dictionaries. In the latest patch both ```_find_jwplayer_data``` and ```_extract_jwplayer_data``` call ```_parse_json```, which is redundant.

rndusr:
But whatever _find_jwplayer_data returns is passed to _parse_json, which expects
a string. _parse_json is used in other methods, so I'd have to make changes
there or in _parse_json itself. Which I could do, but many of the tests fail for
me, and they take hours to process, so I don't have any straightforward way to
see if and how I break anything.

I agree, parsing the same string twice is ugly. But the limiting factor will
always be the download rate. An extra call to _parse_json won't be noticable.

yan12125:
> I don't have any straightforward way to see if and how I break anything.

Currently only a few extractors call _find_jwplayer_data or _extract_jwplayer_data. Checking relevant tests is sufficient.

> An extra call to _parse_json won't be noticable.

The point is maintainability rather than performance. Duplicated _parse_json calls are somewhat confusing.

rndusr:
> Duplicated _parse_json calls are somewhat confusing.

I think the function name "is_json_dict" makes it pretty clear why _parse_json
is called there. If there is any doubt, reading the six lines of code shouldn't
take too long.

But yes, it's still ugly. I'll try to make it nicer.

BTW: I would rebase the PR branch when you're happy with it because there are
quite a few useless commits in there by now. Unless that makes things more
awkward for you?

yan12125:
For now it's up to you to do rebasing or not. The only mandatory timing is before landing the patch to master.

rndusr:
Please review my latest attempt to fixing this.

yan12125:
Hello, could you fix conflicts with the current git master?

rndusr:
Sorry, my bad.

yan12125:
Thanks for your work and patience!

rndusr:
> Thanks for your work and patience!

Likewise! :)

PR #12328, opened 2017-03-02
[3plus] Add new information extractor

goggle:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

### Description
[3 Plus](http://www.3plus.tv/) is a Swiss private TV station. They publish their own TV production on their website. It would be nice to have youtube-dl support for it. Here is a suggestion for a information extractor.

### DIFF
--- a/youtube_dl/extractor/threeplus.py
+++ b/youtube_dl/extractor/threeplus.py
@@ -5,54 +5,25 @@ import re
 
 from .common import InfoExtractor
 from .threeqsdn import ThreeQSDNIE
-from ..utils import get_element_by_class
+from ..utils import (
+    get_element_by_class,
+    smuggle_url,
+)
 
 
-class ThreePlusBaseIE(InfoExtractor):
-    _HOST_URL = 'http://playout.3qsdn.com/'
-
-    def _get_title_and_description(self, video_id):
-        webpage = self._download_webpage(
-            self._HOST_URL + video_id, video_id)
-        title = self._og_search_title(webpage)
-        description = self._og_search_description(webpage)
-        return title, description
-
-    def _get_real_video_id(self, video_id):
-        video_info = self._download_webpage(self._HOST_URL + video_id + '?js=true', video_id)
-        video_id = self._search_regex(
-            r'sdnPlayoutId\s*:\s*\'(.+?)\'', video_info, 'Real video id', default=video_id)
-        video_id = video_id.replace('\\x2D', '-')
-        return video_id
-
-    def _extract_from_id(self, video_id):
-        video_id = self._get_real_video_id(video_id)
-        title, description = self._get_title_and_description(video_id)
-        return {
-            '_type': 'url_transparent',
-            'ie_key': ThreeQSDNIE.ie_key(),
-            'url': self._HOST_URL + video_id,
-            'id': video_id,
-            'title': title,
-            'description': description,
-        }
-
-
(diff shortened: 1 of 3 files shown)


dstftw:
3qsdn is already supported.

youtube_dl/extractor/threeplus.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def _get_title_and_description(self, video_id):
+        webpage = self._download_webpage(
+            self._HOST_URL + video_id, video_id)
+        title = self._og_search_title(webpage)
+        description = self._og_search_description(webpage)
+        return title, description
+
+    def _get_real_video_id(self, video_id):
+        video_info = self._download_webpage(self._HOST_URL + video_id + '?js=true', video_id)
```
dstftw: This should be handled in 3qsdn extractor.

goggle: I assumed that but was afraid, that I would break other stuff depending on 3qsdn. [Here](http://playout.3qsdn.com/b964ee16-ed42-11e6-a78b-0cc47a188158?js=true) is an example script. My code searches for the first appearing field key `sdnPlayoutId` and extracts its value. Should I do exactly the same, but in the 3qsdn extractor?

dstftw: You should add support for this playlist-alike 3qsdn URLs in any non-breaking way.

youtube_dl/extractor/threeqsdn.py
```diff
@@ -69,11 +79,20 @@ def _extract_url(webpage):
...
     def _real_extract(self, url):
         video_id = self._match_id(url)
+        url, smuggled_data = unsmuggle_url(url, {})
 
         js = self._download_webpage(
-            'http://playout.3qsdn.com/%s' % video_id, video_id,
+            self._API_URL + video_id, video_id,
             query={'js': 'true'})
 
+        playout_ids = [m.group('id').replace('\\x2D', '-') for m in re.finditer(
```
dstftw: Playlist extraction should only take place after no video formats found.

youtube_dl/extractor/threeqsdn.py
```diff
@@ -132,11 +151,16 @@ def extract_formats(item_url, item={}):
 
         self._sort_formats(formats)
 
-        title = self._live_title(video_id) if live else video_id
+        webpage = self._download_webpage(self._API_URL + video_id, video_id)
```
dstftw: This will break the extraction if webpage download fails. Also, `title` and `description` from the webpage does not bring any senseless names so there is no point extracting them. Same title may be extracted from JS code.

goggle:
Hi dstftw,
I've seen that you have "self-assigned" this pull request. Does this mean that you will take care of this, or should I try to improve the mentioned points anyway? Thanks.

dstftw:
This means I'll take care of merging this PR.

heubergen:
I know it's old but because I'm currently using the tool to download something from the site I wanted to ask for an update on this? The download seems to work with the generic extractor but there is probably still room for improvement.

goggle:
@heubergen This is a very old pull request. I have just had a look at 3plus.tv and it looks completely different than it used to look, so I assume that these changes won't work anymore anyway. If the download of videos works with the generic extractor, that's great. Do you miss anything? If so, you can open a feature request to support [3plus.tv] natively.

PR #12361, opened 2017-03-05
[cbs.com:playlist] Extract playlists of CBS shows.

johnhawkinson:
- [x] New feature

Support extracting playlists of full episode as well as episode clips from `cbs.com`.

PR #12362, opened 2017-03-05
[external:ffmpeg] In test harness, Limit to 10k download size

johnhawkinson:
Otherwise, if you screw up a playlist test by including a playlist
dictionary key, you'll be there for eons while it downloads all the
files before erroring out.

Other downloaders check for running under the test harness and enforce a 10k limit.

Found while working on  #12361, where this test:

```
    _TEST = {
        'url': 'http://www.cbs.com/shows/the-late-show-with-stephen-colbert',
        'info_dict': {
            'id': 61456254,
            'title': 'The Late Show with Stephen Colbert',
        },
        'playlist_mincount': 14,
        # If uncommented, the test harness tries to download all 30 playlist entries.
        # Even limited to 10KB each, this can take 15 minutes. Not reasonable.
        'playlist': [{
            'info_dict': {
                'id': 'xxx',
                'ext': 'xxx.mp4',
                },
        }],
    }
```

would download hours and hours of video before erroring out (or maybe succeeding).

### DIFF
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -270,6 +270,10 @@ class FFmpegFD(ExternalFD):
                 args += ['-rtmp_live', 'live']
 
         args += ['-i', url, '-c', 'copy']
+
+        if self.params.get('test', False):
+            args += ['-fs', compat_str(self._TEST_FILE_SIZE)] # -fs limit_size (output), expressed in bytes
+
         if protocol in ('m3u8', 'm3u8_native'):
             if self.params.get('hls_use_mpegts', False) or tmpfilename == '-':
                 args += ['-f', 'mpegts']


youtube_dl/downloader/external.py
```diff
@@ -191,6 +191,9 @@ def available(cls):
 
     def _call_downloader(self, tmpfilename, info_dict):
         url = info_dict['url']
+
+        is_test = self.params.get('test', False)
```
dstftw: This is only used once.

youtube_dl/downloader/external.py
```diff
@@ -270,6 +273,11 @@ def _call_downloader(self, tmpfilename, info_dict):
                 args += ['-rtmp_live', 'live']
 
         args += ['-i', url, '-c', 'copy']
+
+        if is_test:
+            self.to_screen("Limiting ffmpeg outout to 10k")
```
dstftw: No need in this message.

youtube_dl/downloader/external.py
```diff
@@ -270,6 +273,11 @@ def _call_downloader(self, tmpfilename, info_dict):
...
+            args += ['-fs', '10000'] # -fs limit_size (output), expressed in bytes
```
dstftw: Use `FileDownloader._TEST_FILE_SIZE` instead.

youtube_dl/downloader/external.py
johnhawkinson: hls.py only used it once as well, so I concluded this was preferred style. Sorry.

youtube_dl/downloader/external.py
```diff
@@ -270,6 +271,10 @@ def _call_downloader(self, tmpfilename, info_dict):
                 args += ['-rtmp_live', 'live']
 
         args += ['-i', url, '-c', 'copy']
+
+        if self.params.get('test', False):
+            args += ['-fs', str(self._TEST_FILE_SIZE)] # -fs limit_size (output), expressed in bytes
```
dstftw: Use `compat_str` instead. Also check code with flake8 and squash commits.

youtube_dl/downloader/external.py
```diff
@@ -191,6 +191,7 @@ def available(cls):
 
     def _call_downloader(self, tmpfilename, info_dict):
         url = info_dict['url']
+
```
dstftw: Remove this line.

youtube_dl/downloader/external.py
johnhawkinson: `compat_str` done -- note that `http.py`  uses just plain `str()` as well. Does this even matter for numbers? 
Stand by on squash.

dstftw: `compat_str` most likely appeared after code in `http.py` was written. In general this matters for passing arguments to subprocess. Anyway this is inconsistent when some of arguments are strings and another are bytestrings.

PR #12399, opened 2017-03-08
[telequebec] description array may be empty

scarpentier:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Some telequebec videos do not contain any description metadata. In this case, the downloaded json reports an empty array which crashes the extractor on line 36.

Behavior can be reproduced with this video: http://zonevideo.telequebec.tv/media/30261

### DIFF
--- a/youtube_dl/extractor/telequebec.py
+++ b/youtube_dl/extractor/telequebec.py
@@ -33,7 +33,7 @@ class TeleQuebecIE(InfoExtractor):
             'id': media_id,
             'url': smuggle_url('limelight:media:' + media_data['streamInfo']['sourceId'], {'geo_countries': ['CA']}),
             'title': media_data['title'],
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/telequebec.py
```diff
@@ -33,7 +33,7 @@ def _real_extract(self, url):
             'id': media_id,
             'url': smuggle_url('limelight:media:' + media_data['streamInfo']['sourceId'], {'geo_countries': ['CA']}),
             'title': media_data['title'],
-            'description': media_data.get('descriptions', [{'text': None}])[0].get('text'),
+            'description': media_data.get('descriptions', [{'text': None}])[0].get('text') if media_data.get('descriptions') else '',
```
dstftw: This will fail is `media_data.get('descriptions')` is not a `list`.

PR #12406, opened 2017-03-08
Windows setup via chocolatey repository

majkinetor:
Added note in README.md about installation using Chocolatey repository (similar as Homebrew on OS X).

I am one of the package maintainers and we keep the package up to date automatically as can be seen in package versions on the Chocolatey gallery

https://chocolatey.org/packages/youtube-dl

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,12 @@ If you do not have curl, you can alternatively use a recent wget:
     sudo wget https://yt-dl.org/downloads/latest/youtube-dl -O /usr/local/bin/youtube-dl
     sudo chmod a+rx /usr/local/bin/youtube-dl
 
-Windows users can [download an .exe file](https://yt-dl.org/latest/youtube-dl.exe) and place it in any location on their [PATH](http://en.wikipedia.org/wiki/PATH_%28variable%29) except for `%SYSTEMROOT%\System32` (e.g. **do not** put in `C:\Windows\System32`).
+On Windows, you can also install youtube-dl via [Chocolatey](https://chocolatey.org/packages/youtube-dl) repository:
+
+    choco install youtube-dl
+
+or [download an .exe file](https://yt-dl.org/latest/youtube-dl.exe) and place it in any location on [PATH](http://en.wikipedia.org/wiki/PATH_%28variable%29) except for `%SYSTEMROOT%\System32` (e.g. **do not** put in `C:\Windows\System32`).
+
 
 You can also use pip:
 


dstftw:
Using official binary on Window is preferred.

majkinetor:
It is official binary, as always on Chocolatey. 

You can take a look at the source code of the package automatic updater:

https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/youtube-dl/update.ps1

The lines of interest are:

```
$domain   = 'https://github.com'
$releases = "$domain/rg3/youtube-dl/releases/latest"
```

On our [update report](https://gist.github.com/gep13/bd2eaa76f2a9ab739ca0544c502dca6e) you can see the [acutall URL](https://gist.github.com/gep13/bd2eaa76f2a9ab739ca0544c502dca6e#youtube-dl) that is used to embed the binary into the package:

```
youtube-dl - checking updates using au version 2017.3.8.195050
URL check
  https://github.com/rg3/youtube-dl/releases/download/2017.03.07/youtube-dl.exe
nuspec version: 2017.03.07
remote version: 2017.03.07
No new version found
```

majkinetor:
In the package creation process, there is nothing else done with the exe file. Package provides metadata about the tool only and its standard [nuget](https://docs.microsoft.com/en-us/nuget/) package used on Windows platform.

dstftw:
I mean I don't see any point mentioning just another 3rdparty repo especially before the official exe.

majkinetor:
Whatever you choose really is fine by me. And like I said, it is official exe, just packed in convenient way that can be used to quickly upgrade it (along with other tools on Windows), pinpoint particular version etc and be used with other popular repository tools such as Nexus or Artifactory in a local offline proxy.

There is no difference between choco and homebrew tho and it is mentioned, that is why I thought it is a good idea.

majkinetor:
One other important thing, it automatically installs dependency, vcredist 2010, so it will reduce issues for Windows users.

yan12125:
I think it‘s not bad to mention another installation method if:

1. It's always kept up to date
2. The update process contains client-side GPG key verification. I didn't see GPG related stuffs in your links. Could you give some more info?
3. It's mentioned as an alternative after the official binary. You may argue that it makes users' life easier, but we don't want to put things that we have no control first.

majkinetor:
@yan12125 

1. Yes, it is always kept up to date up to few hours. Our [bot](https://github.com/chocolatey/chocolatey-coreteampackages/commits/master?author=choco-bot) runs minimum every 6 hours and regularly more then that so the update happens with maximum few hours latency then released here. As you can see on [repology](https://repology.org/metapackage/youtube-dl/versions) there are only 4 repositories that keep youtube-dl up to date and choco is one of them. 
2. The client side does checksum checking and GPG code sign is planned in the future (@ferventcoder, can tell more about it)
3. I simply followed the method you used for homebrew. I changed the README to put official windows install before chocolaty.

majkinetor:
It looks now like this (commit wasn't shown here because PR is closed):

---
Windows users can [download an .exe file](https://yt-dl.org/latest/youtube-dl.exe) and place it in any location on their [PATH](http://en.wikipedia.org/wiki/PATH_%28variable%29) except for `%SYSTEMROOT%\System32` (e.g. **do not** put in `C:\Windows\System32`).
Alternatevelly, you can also install and upgrade youtube-dl via [Chocolatey](https://chocolatey.org/packages/youtube-dl) repository:

    choco install youtube-dl

yan12125:
Thanks for the answers. Do you have a link to an issue/a ticket for GPG stuffs in your packaging system? When it's done, we can revisit this change.

majkinetor:
https://github.com/chocolatey/choco/issues/504

Issue #12410, opened 2017-03-09
suffolk.edu/sjc support broke on Feb.16


johnhawkinson:
…even though there's a test for it! Not quite sure I understand, although the test is marked `'skip_download': True` so that might be related, except the failure is pre-download and I don't fully understand.

I left https://github.com/rg3/youtube-dl/commit/a4a554a79354981fcab55de8eaab7b95a40bbb48#commitcomment-21221835 which was probably not useful and I should have just filed an issue.

Support for this was added in response to #11993 on Feb. 6, but it was broken ten days later in a4a554a79354981fcab55de8eaab7b95a40bbb48:

```
pb3:youtube-dl jhawk$ git checkout b898f0a173fa040ddf95dbd97650cec07a8f19f5
Previous HEAD position was a50862b73... [downloader/external] Add missing import and PEP8
HEAD is now at b898f0a17... [elpais] Fix typo and improve extraction (closes #12139)
pb3:youtube-dl jhawk$ python test/test_download.py TestDownload.test_Generic_61
[generic] sjc: Requesting header
[generic] sjc: Downloading webpage
[generic] sjc: Extracting information
[info] Writing video description metadata as JSON to: sjclive.info.json
.
----------------------------------------------------------------------
Ran 1 test in 0.810s

OK
pb3:youtube-dl jhawk$ git checkout a4a554a79354981fcab55de8eaab7b95a40bbb48
Previous HEAD position was b898f0a17... [elpais] Fix typo and improve extraction (closes #12139)
HEAD is now at a4a554a79... [generic] Try parsing JWPlayer embedded videos (closes #12030)
pb3:youtube-dl jhawk$ python test/test_download.py TestDownload.test_Generic_61
[generic] sjc: Requesting header
[generic] sjc: Downloading webpage
[generic] sjc: Extracting information
ERROR: An extractor error has occurred. (caused by KeyError(u'title',)); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/Users/jhawk/src/youtube-dl/youtube_dl/extractor/common.py", line 370, in extract
    return self._real_extract(url)
  File "/Users/jhawk/src/youtube-dl/youtube_dl/extractor/generic.py", line 2507, in _real_extract
    return self._parse_jwplayer_data(jwplayer_data, video_id)
[...]

PR #12414, opened 2017-03-10
[extractor/common] Not all video_data have "title"

mozbugbox:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -2218,7 +2218,7 @@ class InfoExtractor(object):
 
             entries.append({
                 'id': this_video_id,
-                'title': video_data['title'] if require_title else video_data.get('title'),
+                'title': video_data.get('title', '') if require_title else '',
                 'description': video_data.get('description'),
                 'thumbnail': self._proto_relative_url(video_data.get('image')),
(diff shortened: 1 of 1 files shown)


dstftw:
This is intentional. `require_title` must be passed if there is no title.

mozbugbox:
* The logic of that line was wrong to begin with. It passed title no matter if the `require_title` was true or false.
* this line of code will raise exception when "title" field not in the video_data as can be seen in issue #12410 .
* The fix passed back title if `require_title` was True. If not "title" field in `video_data`, an empty string was passed.

mozbugbox:
Actually, the ambiguity was when `require_title == False`, should something passed back for `title`? If a `title` should be passed, as the current code stands, then that line should be changed to 
 ```
               'title': video_data.get('title', ''),
```
i.e. always try to pass the "title" field back.

dstftw:
1. Current logic is perfectly fine.
2. `require_title=False` should be passed if title may not be present. I've already pointed this out.
3. This is wrong. `require_title` only determines whether extraction should fail if title is not available or not.

PR #12428, opened 2017-03-11
[crunchyroll] Extract season name as well as the series

Starsam80:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

In addition to extracting the `series`, we should also extract the `season` name as well. See the test I added for an example.

### DIFF
--- a/youtube_dl/extractor/crunchyroll.py
+++ b/youtube_dl/extractor/crunchyroll.py
@@ -177,6 +177,7 @@ class CrunchyrollIE(CrunchyrollBaseIE):
             'uploader': 'Kadokawa Pictures Inc.',
             'upload_date': '20170118',
             'series': "KONOSUBA -God's blessing on this wonderful world!",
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/crunchyroll.py
```diff
@@ -492,6 +510,7 @@ def _real_extract(self, url):
         series = self._html_search_regex(
             r'id=["\']showmedia_about_episode_num[^>]+>\s*<a[^>]+>([^<]+)',
             webpage, 'series', default=xpath_text(metadata, 'series_title'))
```
dstftw: This fallback should be removed now.

Starsam80: So `default=None`?

dstftw: `fatal=False`.

Starsam80:
Hopefully this is what you wanted

PR #12493, opened 2017-03-19
[Gaskrank] fix for broken site.

motophil:

- [x] Bug fix

---

Fix for broken site gaskrank.tv

### DIFF
--- a/youtube_dl/extractor/gaskrank.py
+++ b/youtube_dl/extractor/gaskrank.py
@@ -52,9 +52,8 @@ class GaskrankIE(InfoExtractor):
         display_id = self._match_id(url)
         webpage = self._download_webpage(url, display_id)
         categories = [re.match(self._VALID_URL, url).group('categories')]
-        title = self._search_regex(
-            r'<meta[^>]+itemprop\s*=\s*"name"[^>]+content\s*=\s*"([^"]+)"',
-            webpage, 'title')
+        title = self._og_search_title(webpage, default=None) or self._html_search_meta(
+            'title', webpage, fatal=True)
 
         mobj = re.search(
             r'Video von:\s*(?P<uploader_id>[^|]*?)\s*\|\s*vom:\s*(?P<upload_date>[0-9][0-9]\.[0-9][0-9]\.[0-9][0-9][0-9][0-9])',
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/gaskrank.py
```diff
@@ -89,23 +85,22 @@ def fix_json(code):
...
-            display_id, transform_source=fix_json, fatal=False)
-
         video_id = self._search_regex(
             r'https?://movies\.gaskrank\.tv/([^-]*?)(-[^\.]*)?\.mp4',
-            playlist.get('0').get('src'), 'video id')
+            webpage, 'video id')
+
+        playlist = re.findall(
+            r'<source([^>]+)',
+            webpage)
```
dstftw: Use `self._parse_html5_media_entries` instead.

motophil: Fixed.

youtube_dl/extractor/gaskrank.py
```diff
@@ -89,35 +82,25 @@ def fix_json(code):
...
-            'categories': categories,
-            'display_id': display_id,
-            'uploader_id': uploader_id,
-            'upload_date': upload_date,
-            'uploader_url': uploader_url,
-            'tags': tags,
-            'view_count': view_count,
-            'average_rating': average_rating,
-        }
+        return self.playlist_result(entries)
```
dstftw: There is always only one video on the page thus you should take first entry and return as regular info dict.

youtube_dl/extractor/gaskrank.py
```diff
@@ -49,19 +48,13 @@ class GaskrankIE(InfoExtractor):
...
-        def fix_json(code):
-            """Removes trailing comma in json: {{},} --> {{}}"""
-            return re.sub(r',\s*}', r'}', js_to_json(code))
 
         display_id = self._match_id(url)
         webpage = self._download_webpage(url, display_id)
         categories = [re.match(self._VALID_URL, url).group('categories')]
         title = self._search_regex(
-            r'movieName\s*:\s*\'([^\']*)\'',
+            r'<meta[^>]+itemprop\s*=\s*"name"[^>]+content\s*=\s*"([^"]+)"',
```
dstftw: Use `self._html_search_meta` or `self._og_search_title` instead.

motophil: Fixed.

youtube_dl/extractor/gaskrank.py
motophil: Fixed.

Issue #12501, opened 2017-03-19
[generic] fragment indicators confuse redirect logic


johnhawkinson:
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2017.03.20**

Presence of a fragment indicator (`#`) at the end of a URL causes youtube-dl to think it is a different URL, and the redirect logic in `extractor/generic.py` gets confused:

```
pb3:extractor jhawk$ youtube-dl -vs --write-pages 'http://www8.hp.com/us/en/solutions/security/thewolf.html?jumpid=va_87trme41uf#'
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'-vs', u'--write-pages', u'http://www8.hp.com/us/en/solutions/security/thewolf.html?jumpid=va_87trme41uf#']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2017.03.20
[debug] Python version 2.7.10 - Darwin-14.5.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg git-2017-02-28-7f62368, ffprobe git-2017-02-28-7f62368, rtmpdump 2.4
[debug] Proxy map: {}
[generic] thewolf: Requesting header
[redirect] Following redirect to http://www8.hp.com/us/en/solutions/security/thewolf.html?jumpid=va_87trme41uf
[generic] thewolf: Requesting header
WARNING: Falling back on generic information extractor.
[generic] thewolf: Downloading webpage
WARNING: URL could be a direct video link, returning it as such.
```

That'd be this check at 1695, I guess:
```
(Pdb) l
1691 	
1692 	        if head_response is not False:
1693 	            # Check for redirect
1694 	            new_url = head_response.geturl()
1695 	            if url != new_url:
1696 ->	                self.report_following_redirect(new_url)
1697 	                if force_videoid:
1698 	                    new_url = smuggle_url(
1699 	                        new_url, {'force_videoid': force_videoid})
1700 	                return self.url_result(new_url)
1701 	
(Pdb) p url
u'http://www8.hp.com/us/en/solutions/security/thewolf.html?jumpid=va_87trme41uf#'
(Pdb) p new_url
u'http://www8.hp.com/us/en/solutions/security/thewolf.html?jumpid=va_87trme41uf'
(Pdb) 
```

And just for the record, no `Location:` header here:
```
[...]

PR #12502, opened 2017-03-20
[generic] misc. fixes (lazyYT test, first_bytes whitespace)

johnhawkinson:
I happened to have b68a812ea839e44148516a34a15193189e58ba77 open recently, and wound up at the English version, i.e. http://www8.hp.com/us/en/solutions/security/thewolf.html?jumpid=va_87trme41uf# via http://www.hp.com, and noted it doesn't work with Youtube-DL.

Unlike the Chinese version, nothing to do with Brightcove.
#12501 is one problem, filed separately (not fixed here).

Here are two fixes for other issues I found, first 00bc75c, that this webpage starts with 512 bytes of whitespace so it fails the `is_html(first_bytes)` check:
```
(Pdb) p first_bytes
'\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n                                            \r\n                                            \r\n                                            \r\n                                            \r\n                                                          \r\n                                            \r\n                                            \r\n\r\n\r\n'
(Pdb) 
```

and also  	6206194, that it uses a pattern similar to LazyYT, whose test is broken because the domain is gone:

```
pb3:extractor jhawk$ host -t a  discourse.ubuntu.com
discourse.ubuntu.com is an alias for ubuntu.bydiscourse.com.
pb3:extractor jhawk$ host -t a  ubuntu.bydiscourse.com.
ubuntu.bydiscourse.com has no A record
```

Of course, after that, Youtube-DL still doesn't work. I'm not sure how much we care, since it's a YouTube embed, so presumably users will know they can click on the YouTube button and find the YouTube URL. But here's the mechanism:

```html
    <div  class="youtube-popup no-default-transition" 
          data-playlist="PLoMwRIIUViGQ48XKzXNxTjvICAiZTLv8I" 
          data-full-video-id="U3QXMMV-Srs">
      <div id="play-list"></div>
      <iframe id="full-video" frameborder="0" allowfullscreen="1" title="YouTube video player" width="640" height="360"></iframe>
      <div class="close-popup"></div>
    </div>
```

which is quite similar to the LazyYT pattern:
```html
<div class="lazyYT"
  data-youtube-id="y9r6qp8UBQc"
  data-ratio="16:9">
    What is Skiplagged?
</div>
```

but it doesn't seem to be used on other sites that I could find, so perhaps it doesn't meet the standards for inclusion. Not sure.

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -902,12 +902,13 @@ class GenericIE(InfoExtractor):
         },
         # LazyYT
         {
-            'url': 'http://discourse.ubuntu.com/t/unity-8-desktop-mode-windows-on-mir/1986',
+            'url': 'https://skiplagged.com/',
             'info_dict': {
-                'id': '1986',
-                'title': 'Unity 8 desktop-mode windows on Mir! - Ubuntu Discourse',
+                'id': 'skiplagged',
+                'title': 'Skiplagged: The smart way to find cheap flights',
             },
-            'playlist_mincount': 2,
+            'playlist_mincount': 1,
+            'add_ie': ['Youtube'],
         },
         # Cinchcast embed
         {
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -1759,7 +1759,7 @@ class GenericIE(InfoExtractor):
             self._sort_formats(info_dict['formats'])
             return info_dict
 
-        if re.match(r'^\s+$', first_bytes):
+        if re.match(r'^\s+$', first_bytes.decode('utf-8', 'replace')):
             self._downloader.report_warning(
                 'First block is just whitespace? Continuing...')
         elif not is_html(first_bytes):
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -1759,9 +1759,12 @@ class GenericIE(InfoExtractor):
             self._sort_formats(info_dict['formats'])
             return info_dict
 
-        # Maybe it's a direct link to a video?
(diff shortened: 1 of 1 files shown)


johnhawkinson:
Whoops, 	00bc75c broke Python 3 because utf-8 decode issues. Fixed in 
	a5d5a2c.

johnhawkinson:
Err, so… 	6206194c5ace95e5a825b4a58a395030804e7c20 was committed as e8e4cc5a6a3ad8bf94d9ff9e5bb2d72712e14c34 recently, but that leaves the `first_bytes` whitespace fix. Not sure they (00bc75ca0115fa57ffc700357ba6ef86f3355bb9, a5d5a2c068b00a8118fa9e3c32a9d93f316b2edd) were deliberately omitted or inadvertently?

Thanks.

johnhawkinson:
@dstftw, did you mean to commit half of this only?

dstftw:
Yes.

PR #12514, opened 2017-03-20
[picarto] Add new extractor, Picarto.TV

TheAMM:
---

- [ ] Improvement
- [ ] Bug fix
- [x] New extractor
- [ ] New feature

---

This PR adds an extractor for Picarto.TV, a live-stream (only) platform for artists.
The site supports HLS streams as well, but MP4 was simple enough, for now.

### DIFF
--- a/youtube_dl/extractor/picarto.py
+++ b/youtube_dl/extractor/picarto.py
@@ -1,0 +1,80 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+import time
+
+from .common import InfoExtractor
+from ..utils import ExtractorError
+from ..compat import compat_str
+
+
+class PicartoIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?picarto\.tv/(?P<id>[a-zA-Z0-9]+)'
+    _TEST = {
+        'url': 'https://picarto.tv/setz',
+        'info_dict': {
+            'id': 'Setz',
+            'ext': 'mp4',
+            'title': 're:^Setz [0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}$',  # match self._live_title
+            'timestamp': int,
+            'is_live': True
+        },
+        'params': {
+            # Livestream
+            'skip_download': True
+        }
+    }
+    _PLACESTREAM_REGEX = r'''(?x)
+                             placeStream\(
+                                "(?P<channel>[a-zA-Z0-9]+)",
+                                (?P<player_id>\d+),
+                                (?P<product>\d+),
+                                (?P<offline_image>\d+),
+                                (?P<online>\d+),
+                                "(?P<token>.+?)",
+                                "(?P<tech>.+?)",
+                                (?P<viewer>\d+)\ *
+                            \);
+                        '''.replace(r',', r',\ *')
+
+    def _real_extract(self, url):
+        url_channel_id = self._match_id(url)
+        stream_page = self._download_webpage('https://picarto.tv/' + url_channel_id, url_channel_id, note="Downloading channel page")
+
+        # Handle nonexistent channels
+        if 'This channel does not exist.' in stream_page:
+            raise ExtractorError("Channel does not exist", expected=True)
+
+        # Grab all relevant stream info
+        placestream_m = re.search(self._PLACESTREAM_REGEX, stream_page)
+
+        if not placestream_m:
+            raise ExtractorError("Unable to fetch channel info")
+        elif int(placestream_m.group('online')) == 0:
+            raise ExtractorError("Stream is offline", expected=True)
+
+        channel_id = placestream_m.group('channel')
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/picarto.py
```diff
@@ -0,0 +1,80 @@
...
+        'params': {
+            # Livestream
+            'skip_download': True
+        }
+    }
+    _PLACESTREAM_REGEX = r'''(?x)
+                             placeStream\(
+                                "(?P<channel>[a-zA-Z0-9]+)",
+                                (?P<player_id>\d+),
+                                (?P<product>\d+),
```
dstftw: Do not capture groups you don't use.

youtube_dl/extractor/picarto.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class PicartoIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?picarto\.tv/(?P<id>[a-zA-Z0-9]+)'
+    _TEST = {
+        'url': 'https://picarto.tv/setz',
+        'info_dict': {
+            'id': 'Setz',
+            'ext': 'mp4',
+            'title': 're:^Setz [0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}$',  # match self._live_title
```
dstftw: What's the point of `# match self._live_title` here? Remove.

youtube_dl/extractor/picarto.py
```diff
@@ -0,0 +1,80 @@
...
+                                "(?P<channel>[a-zA-Z0-9]+)",
+                                (?P<player_id>\d+),
+                                (?P<product>\d+),
+                                (?P<offline_image>\d+),
+                                (?P<online>\d+),
+                                "(?P<token>.+?)",
+                                "(?P<tech>.+?)",
+                                (?P<viewer>\d+)\ *
+                            \);
+                        '''.replace(r',', r',\ *')
```
dstftw: What's the point of this? Include into regex.

youtube_dl/extractor/picarto.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def _real_extract(self, url):
+        url_channel_id = self._match_id(url)
+        stream_page = self._download_webpage('https://picarto.tv/' + url_channel_id, url_channel_id, note="Downloading channel page")
+
+        # Handle nonexistent channels
```
dstftw: Pointless comment should be removed.

youtube_dl/extractor/picarto.py
```diff
@@ -0,0 +1,80 @@
...
+        if 'This channel does not exist.' in stream_page:
+            raise ExtractorError("Channel does not exist", expected=True)
```
dstftw: Use single quotes consistently.

youtube_dl/extractor/picarto.py
```diff
@@ -0,0 +1,80 @@
...
+            raise ExtractorError("Unable to fetch channel info")
+        elif int(placestream_m.group('online')) == 0:
+            raise ExtractorError("Stream is offline", expected=True)
+
+        channel_id = placestream_m.group('channel')
+        player_id = placestream_m.group('player_id')
+        token = placestream_m.group('token')
+
+        # Ask for stream host
+        post_body = ('loadbalancinginfo=' + channel_id).encode('utf-8')
```
dstftw: `utils.urlencode_postdata`.

youtube_dl/extractor/picarto.py
```diff
@@ -0,0 +1,80 @@
...
+        load_balancing_info = self._download_webpage('https://picarto.tv/process/channel', channel_id, data=post_body, note="Fetching load balancer info")
+
+        if not load_balancing_info or load_balancing_info in ('FULL', 'failedGetIP'):
+            raise ExtractorError("Unable to get stream")
+
+        timestamp = time.time()
+
+        video_url = "https://%s-%s/mp4/%s.mp4?token=%s&ts=%s" % (player_id, load_balancing_info,
```
dstftw: m3u8 is also available.

youtube_dl/extractor/picarto.py
TheAMM: To clarify we're trying to match against the output of the `_live_title` function, and not something the site would ever give us. As it's part of the base class and out of my hands, I decided to comment upon it would the `_live_title` ever change.

youtube_dl/extractor/picarto.py
TheAMM: I was being verbose as the placeStream may change in the future, and some of them can be used to fetch a thumbnail or viewer counts, which I didn't consider crucial for first implementation.

youtube_dl/extractor/picarto.py
dstftw: It's obvious from `'is_live': True`.

PR #12558, opened 2017-03-25
[WSJ] Extract videos embedded in articles, too

johnhawkinson:
- [x] Improvement

### DIFF
--- a/youtube_dl/extractor/wsj.py
+++ b/youtube_dl/extractor/wsj.py
@@ -17,6 +17,7 @@ class WSJIE(InfoExtractor):
         )
         (?P<id>[a-zA-Z0-9-]+)'''
     IE_DESC = 'Wall Street Journal'
+    TEMPLATE_URL = 'https://wsj.com/video/./%s'
     _TESTS = [{
         'url': 'http://video-api.wsj.com/api-video/player/iframe.html?guid=1BD01A4C-BFE8-40A5-A42F-8A8AF9898B1A',
         'md5': 'e230a5bb249075e40793b655a54a02e4',
@@ -109,4 +110,4 @@ class WSJArticleIE(WSJIE):
         webpage = self._download_webpage(url, article_id)
         video_id = self._search_regex(r'data-src=["\']([A-Z0-9\-]+)',
                                       webpage, 'video id')
-        return self._extract_video(video_id)
+        return self.url_result(self.TEMPLATE_URL % video_id, WSJIE.ie_key())
--- a/youtube_dl/extractor/wsj.py
+++ b/youtube_dl/extractor/wsj.py
@@ -37,7 +37,9 @@ class WSJIE(InfoExtractor):
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
+        return self._extract_video(video_id)
 
+    def _extract_video(self, video_id):
         api_url = (
             'http://video-api.wsj.com/api-video/find_all_videos.asp?'
             'type=guid&count=1&query=%s&fields=type,hls,videoMP4List,'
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/wsj.py
```diff
@@ -87,3 +89,24 @@ def _real_extract(self, url):
...
+            'uploader_id': 'ralcaraz',
+            'title': 'Bao Bao the Panda Leaves for China',
+        }
+    }]
+
+    def _real_extract(self, url):
+        article_id = self._match_id(url)
+        webpage = self._download_webpage(url, article_id)
+        video_id = self._html_search_regex(r'data-src=["\']([A-Z0-9\-]+)',
+                                           webpage, 'video id')
```
dstftw: `_search_regex` is enough here.

youtube_dl/extractor/wsj.py
```diff
@@ -87,3 +89,24 @@ def _real_extract(self, url):
...
+        return self._extract_video(video_id)
```
dstftw: Instead if this extraction should be delegated to `WSJIE` via `url_result`.

youtube_dl/extractor/wsj.py
```diff
@@ -17,6 +17,7 @@ class WSJIE(InfoExtractor):
         )
         (?P<id>[a-zA-Z0-9-]+)'''
     IE_DESC = 'Wall Street Journal'
+    TEMPLATE_URL = 'https://wsj.com/video/./%s'
```
dstftw: What's the point extracting it here? Also use http://video-api.wsj.com/api-video/player/iframe.html?guid=... instead or add a `wsj:` shortcut.

youtube_dl/extractor/wsj.py
```diff
@@ -87,3 +88,24 @@ def _real_extract(self, url):
             'title': title,
             'categories': info.get('keywords'),
         }
+
+
+class WSJArticleIE(WSJIE):
```
dstftw: There is no point inheriting from `WSJIE`.

youtube_dl/extractor/wsj.py
```diff
@@ -87,3 +88,24 @@ def _real_extract(self, url):
...
+            'title': 'Bao Bao the Panda Leaves for China',
+        }
+    }]
+
+    def _real_extract(self, url):
+        article_id = self._match_id(url)
+        webpage = self._download_webpage(url, article_id)
+        video_id = self._search_regex(r'data-src=["\']([A-Z0-9\-]+)',
+                                      webpage, 'video id')
+        return self.url_result(self.TEMPLATE_URL % video_id, WSJIE.ie_key())
```
dstftw: Also pass `video_id` since it's known beforehand.

johnhawkinson:
Sorry, I was doing some research for this comment when you typed yours, so this is kind of duplicative.

> What's the point extracting it here?

Sorry, I don't understand the question. But maybe it's irrelevent with a `wsj:` shortcut.

---

> Instead of this extraction should be delegated to WSJIE via url_result.

Err...I'm definitely not sure that I did this right.
It kind of feels like introducing a `wsj:` pseudoURL  is overkill here. But maybe I should have done that instead? **Clearly.**

I added a template but unfortunately:

```
    _VALID_URL = r'''(?x)https?://
        (?:
            video-api\.wsj\.com/api-video/player/iframe\.html\?guid=|
            (?:www\.)?wsj\.com/video/[^/]+/
        )
        (?P<id>[a-zA-Z0-9-]+)'''
    TEMPLATE_URL = 'https://wsj.com/video/./%s'
```

It really doesn't seem like it's appropriate to put anything at all in first field, so I went with `/./`; and constructing the iframe URL sees kind of wrong, too.

Incidently, I also wasn't really sure how to name this template function, there seems to be a lot of variety:

```
pb3:extractor jhawk$ gsed -n -e 's/^\s*//' -e '/TEMPLATE[A-Z_]* *=/s/=.*//p' *py| sort | uniq -c | sort -nr
  14 BRIGHTCOVE_URL_TEMPLATE 
  12 _URL_TEMPLATE 
   6 _API_URL_TEMPLATE 
   5 _TEMPLATE_URL 
   5 _PAGE_TEMPLATE 
   2 _VIDEO_PAGE_TEMPLATE 
   2 _VIDEO_ID_TEMPLATE 
   2 _VALID_URL_TEMPLATE 
   2 _MOBILE_TEMPLATE 
   1 _VIDEO_INFO_TEMPLATE 
   1 _URL_VIDEO_TEMPLATE 
   1 _THUMBNAIL_URL_TEMPLATE 
   1 _SUB_LANG_TEMPLATE 
   1 _SONG_URL_TEMPLATE 
   1 _SONG_FILE_URL_TEMPLATE 
   1 _SERVICE_PROVIDER_TEMPLATE 
   1 _SECURITY_URL_TEMPLATE 
   1 _PLAYER_URL_TEMPLATE 
   1 _PAGE_DATA_TEMPLATE 
   1 _MOVIE_TEMPLATE 
   1 _EPISODE_URL_TEMPLATE 
   1 _AUTHORIZATION_URL_TEMPLATE 
   1 _API_QUERY_TEMPLATE 
   1 _API_BASE_TEMPLATE 
   1 _AGECHECK_TEMPLATE 
   1 TP_RELEASE_URL_TEMPLATE 
   1 TEMPLATE_URL 
   1 SOAP_TEMPLATE 
   1 PROGRESSIVE_URL_TEMPLATE 
   1 PG_URL_TEMPLATE 
   1 META_DATA_URL_TEMPLATE 
   1 ADAPTIVE_URL_TEMPLATE 
pb3:extractor jhawk$ 
```

I guess the one takeaway is I should use a leading underscore. I really don't understand why youtube-dl uses them so much, they seem to add a lot of confusion and (not here) make docstrings not work.

If only there were a "A model, well-commented extractor" to look at :) 
#11968

Issue #5778, opened 2015-05-22
Only the first 1050 videos are grabbed from a YouTube channel


ivan:
https://www.youtube.com/user/drighk/videos has about 8,470 videos but `youtube-dl https://www.youtube.com/user/drighk/videos` only grabs the first 1050:

```
# PYTHONPATH=. ./bin/youtube-dl https://www.youtube.com/user/drighk/videos
[youtube:user] drighk: Downloading page #1
[download] Downloading playlist: drighk
[youtube:user] drighk: Downloading page #2
[youtube:user] drighk: Downloading page #3
[youtube:user] drighk: Downloading page #4
[youtube:user] drighk: Downloading page #5
[youtube:user] drighk: Downloading page #6
[youtube:user] drighk: Downloading page #7
[youtube:user] drighk: Downloading page #8
[youtube:user] drighk: Downloading page #9
[youtube:user] drighk: Downloading page #10
[youtube:user] drighk: Downloading page #11
[youtube:user] drighk: Downloading page #12
[youtube:user] drighk: Downloading page #13
[youtube:user] drighk: Downloading page #14
[youtube:user] drighk: Downloading page #15
[youtube:user] drighk: Downloading page #16
[youtube:user] drighk: Downloading page #17
[youtube:user] drighk: Downloading page #18
[youtube:user] drighk: Downloading page #19
[youtube:user] drighk: Downloading page #20
[youtube:user] drighk: Downloading page #21
[youtube:user] drighk: Downloading page #22
[youtube:user] drighk: Downloading page #23
[youtube:user] drighk: Downloading page #24
[youtube:user] drighk: Downloading page #25
[youtube:user] drighk: Downloading page #26
[youtube:user] drighk: Downloading page #27
[youtube:user] drighk: Downloading page #28
[youtube:user] drighk: Downloading page #29
[youtube:user] drighk: Downloading page #30
[youtube:user] drighk: Downloading page #31
[youtube:user] drighk: Downloading page #32
[youtube:user] drighk: Downloading page #33
[youtube:user] drighk: Downloading page #34
[youtube:user] drighk: Downloading page #35
[youtube:user] playlist drighk: Downloading 1050 videos
[download] Downloading video 1 of 1050
[youtube] pY2zXx4MZBI: Downloading webpage
[youtube] pY2zXx4MZBI: Extracting video information
[youtube] pY2zXx4MZBI: Downloading DASH manifest
```

[...]

PR #12559, opened 2017-03-26
dramatically upgrade function of --dateafter option with regards to c…

BenSapiens:
…ertain (bulk video) IEs known to return date-sorted playlists

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

The code contained within this pull request, as stated in the commit message, dramatically upgrades the function of --dateafter with regards to playlists known to be date-ordered. For instance, we know that the values extracted by the YoutubePlaylistIE info extractor will be date-ordered if they originate with the YoutubeUserIE or YoutubeChannelIE info extractors, as both the YoutubeUserIE and YoutubeChannelIE info extractors, were they still used—not patched over by the solution to issue #5778—would return results in chronological order from most recent to least. By automatically not returning any further results after encountering a video date out of the date range object produced from date-restricting arguments passed by the user, we save youtube-dl from unnecessarily downloading the webpage of literally every single video beyond the dates that the user actually wants, which can number in the dozens or the hundreds.

TL;DR — I wrote this code because I wanted my "automatically get such-and-such users' new videos" script to download a total of 10 webpages instead of 1,000, and take 3 minutes to run rather than an hour.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -758,6 +758,15 @@ class YoutubeDL(object):
                                     'and will probably not work.')
 
             try:
+                if self.original_ie:
+                    pass # self.original_ie already exists
+            except AttributeError:
+                try:
+                    self.original_ie = re.search("\.([^\.]+)'>", str(type(ie))).group(1)
+                except:
+                    self.original_ie = None
+
+            try:
                 ie_result = ie.extract(url)
                 if ie_result is None:  # Finished already (backwards compatibility; listformats and friends should be moved here)
                     break
@@ -922,6 +931,8 @@ class YoutubeDL(object):
 
             x_forwarded_for = ie_result.get('__x_forwarded_for_ip')
 
+            reliably_date_ordered_IEs = ('YoutubeChannelIE, YoutubeUserIE')
+
             for i, entry in enumerate(entries, 1):
                 self.to_screen('[download] Downloading video %s of %s' % (i, n_entries))
                 # This __x_forwarded_for_ip thing is a bit ugly but requires
@@ -948,7 +959,20 @@ class YoutubeDL(object):
                 entry_result = self.process_ie_result(entry,
                                                       download=download,
                                                       extra_info=extra)
-                playlist_results.append(entry_result)
+
+                try:
+                    entry_result_date = entry_result['upload_date']
+                    entry_result_date_year = int(entry_result_date[0:4])
+                    entry_result_date_month = int(entry_result_date[4:6])
+                    entry_result_date_day = int(entry_result_date[6:8])
+                    entry_result_date = datetime.date(year=entry_result_date_year, month=entry_result_date_month, day=entry_result_date_day)
+                    if self.original_ie in reliably_date_ordered_IEs and entry_result_date not in self.params['daterange']:
+                        break
+                    else:
+                        playlist_results.append(entry_result)
+                except: # I don't really know what to expect, so in case of error just fall back to the previous, default, behavior
+                    playlist_results.append(entry_result)
+
             ie_result['entries'] = playlist_results
             self.to_screen('[download] Finished downloading playlist: %s' % playlist)
             return ie_result


youtube_dl/YoutubeDL.py
```diff
@@ -758,6 +758,15 @@ def extract_info(self, url, download=True, ie_key=None, extra_info={},
                                     'and will probably not work.')
 
             try:
+                if self.original_ie:
+                    pass # self.original_ie already exists
+            except AttributeError:
+                try:
+                    self.original_ie = re.search("\.([^\.]+)'>", str(type(ie))).group(1)
```
dstftw: This is a bad idea since it breaks reusing of `YoutubeDL` instances. Even worse it stores string instead of just reference.

youtube_dl/YoutubeDL.py
```diff
@@ -758,6 +758,15 @@ def extract_info(self, url, download=True, ie_key=None, extra_info={},
...
+                except:
```
dstftw: Never ever use bare except.

youtube_dl/YoutubeDL.py
```diff
@@ -922,6 +931,8 @@ def iter_playlistitems(format):
 
             x_forwarded_for = ie_result.get('__x_forwarded_for_ip')
 
+            reliably_date_ordered_IEs = ('YoutubeChannelIE, YoutubeUserIE')
```
dstftw: No, no hardcodes here. You are implementing a generic mechanism thus there should be a way to control it on per extractor basis. Or it would be easier to just introduce an option describing playlist order instead of this hack and require user to specify it if he wants such optimization.

youtube_dl/YoutubeDL.py
```diff
@@ -948,7 +959,21 @@ def iter_playlistitems(format):
                 entry_result = self.process_ie_result(entry,
                                                       download=download,
                                                       extra_info=extra)
-                playlist_results.append(entry_result)
+
+                try:
+                    entry_result_date = entry_result['upload_date']
```
dstftw: No `upload_date` is guaranteed to be present.

youtube_dl/YoutubeDL.py
```diff
@@ -948,7 +959,21 @@ def iter_playlistitems(format):
...
+                    entry_result_date_year = int(entry_result_date[0:4])
+                    entry_result_date_month = int(entry_result_date[4:6])
+                    entry_result_date_day = int(entry_result_date[6:8])
+                    entry_result_date = datetime.date(year=entry_result_date_year, month=entry_result_date_month, day=entry_result_date_day)
+                    # if the entries originate with an info extractor known to return date-sorted results, simply break after we meet the first date out of range
+                    if self.original_ie in reliably_date_ordered_IEs and entry_result_date not in self.params['daterange']:
```
dstftw: This won't work at all if starting playlist entry is not in `daterange`.

BenSapiens:
Sorry for the flurry of minor changes. Hopefully the end result is to your satisfaction.

youtube_dl/options.py
```diff
@@ -298,6 +298,10 @@ def _scrub_eq(o):
         metavar='DATE', dest='dateafter', default=None,
         help='Download only videos uploaded on or after this date (i.e. inclusive)')
     selection.add_option(
+        '--date-ordered-playlist',
```
dstftw: Option should be called something like `--date-playlist-order` and should accept `asc`, `desc` or `none`. Code should process playlist according to it.

youtube_dl/YoutubeDL.py
```diff
@@ -948,7 +950,14 @@ def iter_playlistitems(format):
                 entry_result = self.process_ie_result(entry,
                                                       download=download,
                                                       extra_info=extra)
+
+                entry_result_uploaddate = entry_result.get('upload_date')
+                if entry_result_uploaddate:
+                    if self.params.get('date_ordered_playlist') and entry_result_uploaddate not in self.params.get('daterange'):
```
dstftw: I've already pointed out this does not work.

BenSapiens:
Do you want something else done?

youtube_dl/options.py
```diff
@@ -298,6 +298,15 @@ def _scrub_eq(o):
         metavar='DATE', dest='dateafter', default=None,
         help='Download only videos uploaded on or after this date (i.e. inclusive)')
     selection.add_option(
+        '--date-playlist-order',
+        metavar='ORDER', dest='date_playlist_order', default='none',
+        type='choice', choices=['asc', 'desc', 'none'],
+        help='Specify whether a playlist is known to be listed in chronological order. "Asc", "desc" or "none". Descending order is most recent to least, and is useful for batch-downloading the new videos from a Youtube channel. Default is "none", and is like not using this option at all. Causes youtube-dl to automatically stop after encountering the first video date outside of the date range configured by --date, --datebefore, and/or --dateafter.')
+    selection.add_option(
+        '--start-from-earliest',
```
dstftw: There is already `--playlist-reverse`.

youtube_dl/options.py
```diff
@@ -298,6 +298,15 @@ def _scrub_eq(o):
         metavar='DATE', dest='dateafter', default=None,
         help='Download only videos uploaded on or after this date (i.e. inclusive)')
     selection.add_option(
+        '--date-playlist-order',
+        metavar='ORDER', dest='date_playlist_order', default='none',
+        type='choice', choices=['asc', 'desc', 'none'],
+        help='Specify whether a playlist is known to be listed in chronological order. "Asc", "desc" or "none". Descending order is most recent to least, and is useful for batch-downloading the new videos from a Youtube channel. Default is "none", and is like not using this option at all. Causes youtube-dl to automatically stop after encountering the first video date outside of the date range configured by --date, --datebefore, and/or --dateafter.')
```
dstftw: Should be cut down to coupe of sentences. Possible arguments' values should be written all lowercase without quotes. See `--fixup` for an example.

BenSapiens:
Done and done, and also explicitly defined playlist_random and playlist_reverse, since every other option does it.

youtube_dl/YoutubeDL.py
```diff
@@ -948,7 +953,14 @@ def iter_playlistitems(format):
                 entry_result = self.process_ie_result(entry,
                                                       download=download,
                                                       extra_info=extra)
+
+                entry_result_uploaddate = entry_result.get('upload_date')
```
dstftw: Breaks on `-F`.

youtube_dl/YoutubeDL.py
```diff
@@ -948,7 +953,14 @@ def iter_playlistitems(format):
...
+                if entry_result_uploaddate:
+                    if self.params.get('date_playlist_order') in ('desc', 'asc') and entry_result_uploaddate not in self.params.get('daterange'):
```
dstftw: Did you even bother to run your code? `self.params.get('date_playlist_order')` will always be `None` thus `break` will never trigger since [you did not forward it to `params`](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/__init__.py#L310-L429). Even if you did, it **won't work** (repeating this 3rd time already) if `daterange` is an **inner interval** inside `daterange`. In such case if the first/last video (depending on the order) does not belong to `daterange` extraction will stop on it and won't process further items that may belong to `daterange`.

youtube_dl/options.py
```diff
@@ -482,11 +489,11 @@ def _scrub_eq(o):
         help=optparse.SUPPRESS_HELP)
     downloader.add_option(
         '--playlist-reverse',
-        action='store_true',
+        dest='playlist_reverse', action='store_true',
```
BenSapiens: Okay.

youtube_dl/YoutubeDL.py
BenSapiens: Nice. Fixed.

youtube_dl/YoutubeDL.py
BenSapiens: > Did you even bother to run your code?

I've been running youtube-dl as described here: https://github.com/rg3/youtube-dl#embedding-youtube-dl, and completely forgot that I was manually specifying `ydl_opts`.

> In such case if the first/last video (depending on the order) does not belong to daterange extraction will stop on it and won't process further items that may belong to daterange.

That's literally the entire point.

dstftw: No, it's not the point. It **must not** skip items than belong to the range specified.

BenSapiens:
Alright...

BenSapiens:
scratch that

BenSapiens:
New commit is what preceding commit should have been.

Also, I wonder if, when encountering a "private video"  in a playlist, might it not be better to continue downloading the rest of the playlist?

Finally, I'm done. It's there; it works; hopefully all those people who asked you for this feature over the years will benefit.

In parting I would leave you with these words: "Any customer can have a car painted any color that he wants so long as it is black." — Henry Ford

BenSapiens:
Alright. I'm back. What do you want to get this merged? I want the many other users of youtube-dl to benefit.

PR #12603, opened 2017-03-30
[openload] switch to pairing method

### LINKED ISSUE #10408: openload.co extractor not working
youtube-dl --get-url --verbose  https://openload.co/embed/kUEfGclsU9o/
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--get-url', u'--verbose', u'https://openload.co/embed/kUEfGclsU9o/']
[debug] youtube-dl version 2016.08.22
[debug] Python version 2.6.6 - Linux-2.6.32-642.1.1.el6.x86_64-x86_64-with-centos-6.8-Final
[debug] exe versions: ffmpeg 0.6.5, ffprobe 0.6.5
[debug] Proxy map: {}
ERROR: Unable to extract link image; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 691, in extract_info
    ie_result = ie.extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 347, in extract
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/openload.py", line 62, in _real_extract
    r']+id="linkimg"[^>]+src="([^"]+)"', webpage, 'link image')
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 650, in _search_regex
RegexNotFoundError: Unable to extract link image; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.

Tithen-Firion:

---

- [ ] At least skimmed through [adding new extractor tutorial](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/rg3/youtube-dl#youtube-dl-coding-conventions) sections

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Removed ID decoding and added pairing method:
Extractor checks if your IP address is paired with Openload. If it is, you get nice download. If it's not, `ExtractorError` with url to pairing page is raised.

If you use `youtube-dl` on some server and can't pair IP of that machine directly, just set up a proxy on that server and pair on any device using that proxy. (tested, works)

### DIFF
--- a/youtube_dl/extractor/openload.py
+++ b/youtube_dl/extractor/openload.py
@@ -58,6 +58,12 @@ class OpenloadIE(InfoExtractor):
         'only_matching': True,
     }]
 
+    _API_URL = 'https://api.openload.co/1'
+    _PAIR_INFO_URL = _API_URL + '/streaming/info'
+    _GET_VIDEO_URL = _API_URL + '/streaming/get?file={0}'
+
+    _PAIR_NEEDED = 'Open this url: {0}, solve captcha, click "Pair" button and try again'
+
     @staticmethod
     def _extract_urls(webpage):
         return re.findall(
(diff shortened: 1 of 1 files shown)


Tithen-Firion:
I wonder how often do they plan to change that decoding method. Should I update it or remove decoding at all and leave just the pairing method?

Tithen-Firion:
I guess that's the only way we can get it to work without any dependencies. It will close #10408 for good.

yan12125:
I wouldn't say this is a long term fix. A long term fix shouldn't require extra steps, thus it needs a full-functional Javascript interpreter. But indeed it's better to have a partially working extractor...

Please mark all tests as skipped. There are definitely no pairing on Travis CI servers,

ozcs:
so is the pairing issue fixed with this new update?

Tithen-Firion:
@yan12125 looking at [deobfuscated code](https://gist.github.com/Tithen-Firion/7ecd8cf00cfcde1955e0be7027d4bf23) I can tell you it will be hard to make JS interpreter that passes all those checks. Even if it does they will add new one which fails. Keep in mind that those checks are obfuscated on their webpage so removing them from the code would be hard.

[Currently working solution](https://github.com/Tithen-Firion/youtube-dl/issues/1#issuecomment-292841180) that can't be merged to master requires you to install `PhantomJS` because even `jsdom` for `Node.js` failed.

Changed all tests to `only_matching`.

@ozcs What `pairing issue`? This commit **introduces** pairing. You open http://openload.co/pair, solve captcha, press `pair` and can download using `youtube-dl` for 4 hours or 5 files, whichever comes first.

ozcs:
why watching through browser does not require pairing but using the script does?

dstftw:
Because
>watching through browser

ozcs:
how can we emulate a browser?

Hrxn:
> why watching through browser does not require pairing but using the script does?

I don't think that is actually necessary. The server can't "see" the other side of a HTTP connection.

ozcs:
so why it makes different ? with browser , no pairing, without browser pairing

Tithen-Firion:
> how can we emulate a browser?

In pure Python? We can't. At least there is no solution that would work. 

@ozcs @Hrxn Openload decodes video URL in your browser. Browsers use JavaScript. `youtube-dl` is written in Python. See the problem? We can't run decoding code in `youtube-dl`. We can rewrite it in Python but Openload changes the code very often. In fact they probably follow development of Openload extractor for `youtube-dl` and change the code whenever we get the working version.

Of course you can try my [experimental module](https://github.com/Tithen-Firion/youtube-dl/issues/1#issue-217964016) that can't be merged because it uses 3rd party software.

Hrxn:
Yes, I'm aware. What I meant was that you can transparently emulate a browser on the client side, and the server cannot "see" this. But for the given case here, we would need full JavaScript support, that is right.

yan12125:
> Currently working solution that can't be merged to master requires you to install PhantomJS because even jsdom for Node.js failed.

Actually I feel PhantomJS a good choice. I reject js2py because it uses ```eval```, which is a good entry point for remote code execution. As PhantomJS is a real browser, it should be as safe as a browser. I'm not against non-Python solutions. As long as it runs on major platforms (Windows, Linux, Mac) and easy to install, it's a good choice. @Tithen-Firion Your PhantomJS-based extractor looks fine - just with some minor issues. Would you like to open a new pull request?

> I can tell you it will be hard to make JS interpreter that passes all those checks. Even if it does they will add new one which fails. Keep in mind that those checks are obfuscated on their webpage so removing them from the code would be hard.

I'm not worried :) The goal of the new Javascript interpreter is to provide as much functionality as real browser engines, like Gecko's SpiderMonkey, Blink's V8, Webkit's JavascriptCore or Edge's ChakraCore. Our implementation can be simpler as we don't need JIT. ES2015 and WebAssembly can be skipped as well (I can't imagine serious websites targetting those).

Tithen-Firion:
@yan12125 they broke PhantomJS already. It adds `window._phantom` and `window.callPhantom` - Openload checks for those variables. They'll do the same if we use [WebDriver](https://pypi.python.org/pypi/selenium). The code is obfuscated so it's hard to remove those checks. As they follow development of Openload extractor and react to new versions accordingly, any public code will be useless after a while. Sure, I can play cat-and-mouse game with them and see how far they will go but I don't think that counts as a long term solution.

yan12125:
Is it possible to hide Phantom-specific variables?

Tithen-Firion:
Uh, I've tried many complicated things, some of them worked but I already knew how Openload can patch them. And then I tried the simplest thing.
```javascript
delete window._phantom;
delete window.callPhantom;
```
and it worked. :| 

Can you merge this one for now? I want to have pairing method as a fallback if PhantomJS fails again. And in case it is not installed.

yan12125:
> and it worked. :|

That's great! Thanks for trying PhantomJS out. That's an important step on extending youtube-dl support for more sites.

> Can you merge this one for now? I want to have pairing method as a fallback if PhantomJS fails again. And in case it is not installed.

The idea of youtube-dl is making the whole process automatic and getting rid of human intervention. The pairing approach is against that idea. Also, I don't think there's an easy way to block a real browser like PhantomJS unless introducing CAPTCHA for _all_ videos, even in ordinary browsers, so a fallback is not necessary.

PR #12643, opened 2017-04-04
[Streamango] Add new extractor

mphe:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Add extractor for http://streamango.com/

### DIFF
--- a/youtube_dl/extractor/streamango.py
+++ b/youtube_dl/extractor/streamango.py
@@ -13,7 +13,7 @@ class StreamangoIE(InfoExtractor):
             'id': 'clapasobsptpkdfe',
             'ext': 'mp4',
             'title': '20170315_150006.mp4',
-            'url': r're:https://streamango\.com/v/d/clapasobsptpkdfe~[0-9]{10}~185\.61\.0\.0~.{8}/720',
+            'url': r're:https://streamango\.com/v/d/clapasobsptpkdfe~[0-9]{10}~(?:[0-9]+\.){3}[0-9]+~.{8}/720',
         }
     }, {
         'url': 'https://streamango.com/embed/clapasobsptpkdfe/20170315_150006_mp4',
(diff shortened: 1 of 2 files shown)


dstftw:
All formats should be extracted.

mphe:
Done, added dash support.
But there's a problem. For some reason the download gets stuck when using dash. Sometimes, after a while, it continues just to get stuck again immediately. I have no idea what's causing this, maybe server side problems.
Is there a way to set the normal mp4 download as default?

dstftw:
>For some reason the download gets stuck when using dash

Works fine for me.

youtube_dl/extractor/streamango.py
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }, {
+        'url': 'https://streamango.com/embed/clapasobsptpkdfe/20170315_150006_mp4',
+        'only_matching': True,
+    }]
+
+    def _real_extract(self, url):
+        def extract_url(urltype):
+            return self._search_regex(
+                r'type\s*:\s*["\']{}["\']\s*,\s*src\s*:\s*["\'](?P<url>.+?)["\'].*'.format(urltype),
```
dstftw: `{}` does not work in python 2.6.

youtube_dl/extractor/streamango.py
```diff
@@ -0,0 +1,54 @@
...
+                webpage, 'video URL', group='url')
+
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._og_search_title(webpage)
+        url = 'https:' + extract_url('video/mp4')
+        dashurl = extract_url(r'application/dash\+xml')
```
dstftw: Extraction should not break if one of the formats is missing.

youtube_dl/extractor/streamango.py
```diff
@@ -0,0 +1,54 @@
...
+
+        formats = [{
+            'url': url,
+            'ext': 'mp4',
+            'width': 1280,
+            'height': 720,
```
dstftw: This is not true in general.

youtube_dl/extractor/streamango.py
```diff
@@ -0,0 +1,54 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class StreamangoIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?streamango\.com/(?:f|embed)/(?P<id>.+?)/(?:.+)'
```
dstftw: `https://streamango.com/f/clapasobsptpkdfe` is not matched by this regex.

dstftw:

PR #12649, opened 2017-04-05
Add a list of other FLOSS video downloaders

pabs3:
This can be useful for sites not supported by youtube-dl.

This can be useful when one site breaks and other downloaders
support it but use a different method that still works.


dstftw:
Out of scope.

PR #12679, opened 2017-04-08
[kaltura] Support iframe embeds, with test

johnhawkinson:
Note that these need to back to through the Generic extractor because
the iframe URLs may be redirects that cannot be parsed by `KalturaIE`
without being followed (as in this test), and Generic checks for such redirects and
follows them. Hence dropping the IE from `url_result()`.

---

- [x] Improvement

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -1080,6 +1080,21 @@ class GenericIE(InfoExtractor):
             },
             'add_ie': ['Kaltura'],
         },
+        {
+            # Kaltura iframe embed
+            'url': 'http://www.gsd.harvard.edu/event/i-m-pei-a-centennial-celebration/',
+            'md5': 'ae5ace8eb09dc1a35d03b579a9c2cc44',
+            'info_dict': {
+                'id': '0_f2cfbpwy',
+                'ext': 'mp4',
+                'title': 'I. M. Pei: A Centennial Celebration',
+                'description': 'md5:1db8f40c69edc46ca180ba30c567f37c',
+                'upload_date': '20170403',
+                'uploader_id': 'batchUser',
+                'timestamp': 1491232186,
+            },
+            'add_ie': ['Kaltura'],
+        },
         # Eagle.Platform embed (generic URL)
         {
             'url': 'http://lenta.ru/news/2015/03/06/navalny/',
@@ -2290,7 +2305,7 @@ class GenericIE(InfoExtractor):
         # Look for Kaltura embeds
         kaltura_url = KalturaIE._extract_url(webpage)
         if kaltura_url:
-            return self.url_result(smuggle_url(kaltura_url, {'source_url': url}), KalturaIE.ie_key())
+            return self.url_result(smuggle_url(kaltura_url, {'source_url': url}))
 
         # Look for Eagle.Platform embeds
         eagleplatform_url = EaglePlatformIE._extract_url(webpage)
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -2305,7 +2305,7 @@ class GenericIE(InfoExtractor):
         # Look for Kaltura embeds
         kaltura_url = KalturaIE._extract_url(webpage)
         if kaltura_url:
-            return self.url_result(smuggle_url(kaltura_url, {'source_url': url}))
+            return self.url_result(smuggle_url(kaltura_url, {'source_url': url}), KalturaIE.ie_key())
 
         # Look for Eagle.Platform embeds
         eagleplatform_url = EaglePlatformIE._extract_url(webpage)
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/kaltura.py
```diff
@@ -139,6 +140,13 @@ def _extract_url(webpage):
                 url = smuggle_url(url, {'service_url': service_url.group(1)})
             return url
 
+        # Check for an iframe, which may require redirection.
+        mobj = re.search(
+            r"<iframe[^>]+src=['\"](?P<url>(https?:)?//www\.kaltura\.com/[^'\"]+)['\"]",
```
dstftw: You should extract `partner_id` and `entry_id` and return `kaltura:...` shortcut.

youtube_dl/extractor/generic.py
```diff
@@ -2290,7 +2305,7 @@ def _real_extract(self, url):
         # Look for Kaltura embeds
         kaltura_url = KalturaIE._extract_url(webpage)
         if kaltura_url:
-            return self.url_result(smuggle_url(kaltura_url, {'source_url': url}), KalturaIE.ie_key())
+            return self.url_result(smuggle_url(kaltura_url, {'source_url': url}))
```
dstftw: No.

youtube_dl/extractor/kaltura.py
johnhawkinson: OK.

It seemed more robust to follow the redirect and use the existing parser. For instance, this test, `http://www.gsd.harvard.edu/event/i-m-pei-a-centennial-celebration/` contains (linebreaks in URLs by me):

```
<iframe
  src="//www.kaltura.com/p/1487951/sp/148795100/embedIframeJs/uiconf_id/26739141
  /partner_id/1487951?iframeembed=true&playerId=kaltura-video-embed-&
  entry_id=0_f2cfbpwy"
  width="1280" height="720" allowfullscreen webkitallowfullscreen
  mozAllowFullScreen frameborder="0">
```

This redirects to:

```
Location: http://cdnapi.kaltura.com/html5/html5lib/v2.46/mwEmbedFrame.php/
  p/1487951/uiconf_id/26739141/entry_id/0_f2cfbpwy?wid=_1487951&
  iframeembed=true&playerId=kaltura-video-embed-1&entry_id=0_f2cfbpwy
```

But I can certainly pull out the `&entry_id=0_f2cfbpwy` from the iframe src.
It just seemed more fragile.

dstftw: No, your approach is more fragile. You match just everything on kaltura.com domain and it may redirect to anywhere and do anything. This iframe embed is unambiguous, exact and explicit [official embed code](https://knowledge.kaltura.com/embedding-kaltura-media-players-your-site) that has well defined semantics and won't change in future.

youtube_dl/extractor/kaltura.py
```diff
@@ -127,7 +128,17 @@ def _extract_url(webpage):
                         (?P<q2>["\'])entry_?[Ii]d(?P=q2)
                     )\s*:\s*
                     (?P<q3>["\'])(?P<id>(?:(?!(?P=q3)).)+)(?P=q3)
-                ''', webpage))
+                ''', webpage) or
+            re.search(
+                # <iframe src="http://www.kaltura.com/p/{PARTNER_ID}/sp/{PARTNER_ID}00/embedIframeJs/uiconf_id/{UICONF_ID}/partner_id/{PARTNER_ID}?iframeembed=true&playerId={UNIQUE_OBJ_ID}&entry_id={ENTRY_ID}" width="400" height="330" allowfullscreen webkitallowfullscreen mozAllowFullScreen frameborder="0"></iframe>
+                r'''(?xs)
+                    (?P<q1>["\'])
```
dstftw: This should include `iframe` part.

youtube_dl/extractor/kaltura.py
```diff
@@ -127,7 +128,17 @@ def _extract_url(webpage):
                         (?P<q2>["\'])entry_?[Ii]d(?P=q2)
                     )\s*:\s*
                     (?P<q3>["\'])(?P<id>(?:(?!(?P=q3)).)+)(?P=q3)
-                ''', webpage))
+                ''', webpage) or
+            re.search(
+                # <iframe src="http://www.kaltura.com/p/{PARTNER_ID}/sp/{PARTNER_ID}00/embedIframeJs/uiconf_id/{UICONF_ID}/partner_id/{PARTNER_ID}?iframeembed=true&playerId={UNIQUE_OBJ_ID}&entry_id={ENTRY_ID}" width="400" height="330" allowfullscreen webkitallowfullscreen mozAllowFullScreen frameborder="0"></iframe>
```
dstftw: Just leave a link to kaltura embedding page.

johnhawkinson:
**Done.**

> This should include iframe part.

I'd originally considered that but the 2nd `re.search()` didn't anchor the regexp in any way like that, so I figured that was the better example. Apparently not :)

dstftw:
Thanks.
As for second regex, many examples have been seen that embed it without `<script src=`.

Issue #12692, opened 2017-04-09
As a last resort, why not have GenericIE check every iframe? (vanityfair.com)


johnhawkinson:
http://www.vanityfair.com/hollywood/2017/04/louis-ck-snl-monologue-white-privilege has an iframe that goes directly to an mp4:

```html
<div
  class="embed iframe-embed-component "
  data-type="iframe"
  data-url="https://cdn-e2.streamable.com/video/mp4/wfatk.mp4?token=1492919903_c3707bc86745b37c149c2be0da56b98a275b2b74"
  >
    <iframe height="360" width="640"
      src="https://cdn-e2.streamable.com/video/mp4/wfatk.mp4?token=1492919903_c3707bc86745b37c149c2be0da56b98a275b2b74"
    >
    </iframe>
</div>
```

and `StreamableIE` doesn't know anything about this sort of URL
    `    _VALID_URL = r'https?://streamable\.com/(?:e/)?(?P<id>\w+)'`
(and there's no discussion of it at https://streamable.com/documentation)

So the generic extractor just fails. But of course the raw .mp4 in the `<iframe src>` retrieves just fine (with Youtube-Dl or anything else):

```
curl -I 'https://cdn-e2.streamable.com/video/mp4/wfatk.mp4?token=1492919903_c3707bc86745b37c149c2be0da56b98a275b2b74'
HTTP/1.1 200 OK
x-amz-id-2: deVubJEfE2aanNg7M+Ix3q5KMWx/yTUGX2Hi70bvprc5DRwmMDL+MHlvtQlDLFx58z0KLiStRrE=
x-amz-request-id: 7DD9384BF10DF36C
Last-Modified: Sun, 09 Apr 2017 03:58:16 GMT
ETag: "1f5e254725c8eb94c09a9e3abbf515e6-31"
Cache-Control: max-age=315360000
Content-Type: video/mp4
Server: AmazonS3
Via: 1.1 varnish
Fastly-Debug-Digest: 7877314f0efbb23d37cdde0e9077f4f5133c9d5549b37d1e9bf1f4c07fdc4adc
Access-Control-Allow-Origin: *
Content-Length: 157744192
Accept-Ranges: bytes
Date: Sun, 09 Apr 2017 05:13:50 GMT
Via: 1.1 varnish
Connection: keep-alive
X-Served-By: cache-iad2138-IAD, cache-bos8234-BOS
X-Cache: HIT, HIT
X-Cache-Hits: 0, 0
[...]

PR #12697, opened 2017-04-10
[Generic] Check iframes for known file extensions

johnhawkinson:
As discussed in  #12692, this is the easy option, 1.:

> 1. Check file extension of iframes for ../utils/KNOWN_EXTENSIONS. (This could happen fairly early in GenericIE)

Since Sergey said

> By default only 1 can be applied.

_Unfortunately_, between then and now(*), the test case I had updated to point to a Youtube video, and I don't have a test case for the direct iframe link. I searched https://publicwww.com/ and didn't come up with anything. I guess they're not common.

So I don't know if this is worth committing without a test. It works against my local webserver with the iframe from yesterday inserted:

```
<iframe height="360" width="640"
      src="https://cdn-e2.streamable.com/video/mp4/wfatk.mp4?token=1492919903_c3707bc86745b37c149c2be0da56b98a275b2b74"
    >
    </iframe>
```
Or I could add a test to `test_InfoExtractor.py`?

---
(*) I suppose maybe it's fortunate that it happened so quickly, rather than happening next week after this was committed and breaking the test case while no one was watching.

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -21,6 +21,7 @@ from ..utils import (
     HEADRequest,
     is_html,
     js_to_json,
+    KNOWN_EXTENSIONS,
     orderedSet,
     sanitized_Request,
     smuggle_url,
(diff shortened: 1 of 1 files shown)


johnhawkinson:
fyi, VF has another of these tonight

http://www.vanityfair.com/hollywood/2017/04/snl-melissa-mccarthy-sean-spicer-easter-bunny-hitler

which has

```html
<div
  class="embed iframe-embed-component "
  data-type="iframe"
  data-url="https://cdn-e2.streamable.com/video/mp4/wb4dr.mp4?token=1493525991_6e4d1806902118fba090b4e528109b806e5d8c4d"
>
  <iframe
    height="360" width="640"
    src="https://cdn-e2.streamable.com/video/mp4/wb4dr.mp4?token=1493525991_6e4d1806902118fba090b4e528109b806e5d8c4d"
  >
  </iframe>
</div>
```

And this code properly works against it.

I imagine they'll replace it with a Youtube link when SNL posts their youtube stuff, though, like last week…

PR #12698, opened 2017-04-10
Various CondeNast fixes

johnhawkinson:
After having my test case for #12697 disappear on me, naturally I looked around http://www.vanityfair.com for another example. I did not find one, but I did find other stuff that doesn't work. (...)

So, this PR fixes 2 existing test cases in `CondeNastIE`, fixes some test metadata in `GenericIE`, and adds support for embeds of the form:

```
<div
  class="embed block cneembed"
  data-url="//player.cnevids.com/embedjs/58e54e8dfd2e615252000010/51cc9f86bb8f554e4e000002.js?autoplay=1&amp;muted=1"
>
```

In `GenericIE`, which is just a trivial change to the regexp that already handles iframes and `<script>` embeds there.

Note that in refactoring e70b856e6e11b724f7dfbaca52fadf360b26fed4 it seems there's a case not covered by a test that was added by @remitamine last year. It's marked `# xxx`, but maybe I should remove that comment.

### DIFF
--- a/youtube_dl/extractor/condenast.py
+++ b/youtube_dl/extractor/condenast.py
@@ -119,7 +119,7 @@ class CondeNastIE(InfoExtractor):
     def _extract_video_params(self, webpage):
         query = {}
         params = self._search_regex(
-            r'(?s)var params = {(.+?)}[;,]', webpage, 'player params', default=None)
+            r'(?s)\bparams = {(.+?)}[;,]', webpage, 'player params', default=None)
         if params:
             query.update({
                 'videoId': self._search_regex(r'videoId: [\'"](.+?)[\'"]', params, 'video id'),
--- a/youtube_dl/extractor/condenast.py
+++ b/youtube_dl/extractor/condenast.py
@@ -62,18 +62,20 @@ class CondeNastIE(InfoExtractor):
     EMBED_URL = r'(?:https?:)?//player(?:-backend)?\.(?:%s)\.com/(?:embed(?:js)?|(?:script|inline)/video)/.+?' % '|'.join(_SITES.keys())
 
     _TESTS = [{
+        # Now an iframe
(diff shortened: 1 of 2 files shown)


johnhawkinson:
Just rebased this branch in light of 6c52477f595176d16f063407bcf44439b9c26022.
@remitamine do you want to take a look?

youtube_dl/extractor/condenast.py
```diff
@@ -62,18 +62,20 @@ class CondeNastIE(InfoExtractor):
...
             'ext': 'mp4',
             'title': '3D Printed Speakers Lit With LED',
-            'description': 'Check out these beautiful 3D printed LED speakers.  You can\'t actually buy them, but LumiGeek is working on a board that will let you make you\'re own.',
+            # 'description': 'Check out these beautiful 3D printed LED speakers.  You can\'t actually buy them, but LumiGeek is working on a board that will let you make you\'re own.',
             'uploader': 'wired',
             'upload_date': '20130314',
             'timestamp': 1363219200,
         }
     }, {
+        # older JS embed: var params = { \n videoId:
```
remitamine: now, the `VideoId` is not included in the params, instead there is:
```
_cne.watchPage.embedPlayer(params, 'http://player-backend.cnevids.com/inline/video/58d1865bfd2e6126e2000015.js?');
```

youtube_dl/extractor/condenast.py
```diff
@@ -119,22 +121,37 @@ def _extract_series(self, url, webpage):
...
             query.update({
                 'videoId': params['data-video'],
                 'playerId': params['data-player'],
                 'target': params['id'],
             })
+            return query
+
+        iframe = self._search_regex(
+            r'<iframe[^>]+src="([^"]+)"', webpage, 'iframe player params')
+        query.update({
```
remitamine: i think it would be better to add support for this embed urls and return a url result instead of inline handling of the embed url.

PR #12725, opened 2017-04-12
[radiocloud.jp] Add new extractor

EliteTK:
### I have

### One of the following applies

### The purpose of this pull request
- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

### Description

This is a new extractor for https://radiocloud.jp/.

This is a audio-only website, if that is not within the scope of this project, my apologies, please close the pull request.

Additionally I can also confirm this is not a website which is based around copyright infringement, from the little Japanese that I know it appears to be the official website for some Japanese radio station.

The website restricts access to older recordings to users without an account, however this restriction is not enforced by anything more than a semi-transparent div preventing clicks, therefore authentication was not implemented in the extractor as it is unnecessary.

The website doesn't have any obvious way of getting the URL for a specific recording, to get one you need to view the RSS feed for a radio program and then use one of the URLs provided there.

The extractor currently makes no attempt to create a playlist of all recordings found on the archive page of a program as there could be quite a few recordings and I personally don't see it as a useful feature, but it could be done.

The extractor also makes no attempt to guess which recording you wish to download if the URL provided does not refer to a specific content_id.

A note about some regex:

``` python
        file_url = self._search_regex(r'var\s*source\s*=\s*"(.+?)"', webpage, 'url')
```

I think this regex is a bit iffy, but I couldn't think of a better way.

Thank you for your time.

Edit: removed note about .strip() - I was mistaken.

### DIFF
--- a/youtube_dl/extractor/radiocloud.py
+++ b/youtube_dl/extractor/radiocloud.py
@@ -1,0 +1,70 @@
+# coding: utf-8
+from __future__ import unicode_literals
+import re
+
+from .common import InfoExtractor
+
+from ..utils import (
+    ExtractorError,
+    extract_attributes,
+    get_elements_by_class,
+)
+
+
+class RadioCloudIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?radiocloud\.jp/archive/.*?content_id=(?P<id>[0-9]+)'
+    _TEST = {
+        'url': 'https://radiocloud.jp/archive/so?content_id=13165',
+        'md5': '6f97afc008931d72c21db1331dbdce94',
+        'info_dict': {
+            'id': '13165',
+            'ext': 'm4a',
+            'title': '毒蝮三太夫のミュージックプレゼント：千葉県市川市鬼高『お酒の専門店「酒壱番」』編',
+            'description': '1/16(月)放送(11:20～11:40頃)から。訪問先は、千葉県市川市鬼高3-15−5『お酒の専門店「酒壱番」』さん。47年続く国民的中継番組のトーク部分のみをお楽しみください。',
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        def get_input_by_name(html, name):
+            for e in re.finditer(r'<input.*?>', html):
+                a = extract_attributes(e.group())
+                if a.get('name') == name:
+                    return a.get('value')
+            return None
+
+        element = None
+        for e in get_elements_by_class('contents_box', webpage):
+            if get_input_by_name(e, 'content_id') == video_id:
+                element = e
+                break
+
+        if not element:
+            raise ExtractorError('Could not find details of id %s' % video_id)
+
+        file_url = get_input_by_name(element, 'file_url')
+        if not file_url:
+            raise ExtractorError('Could not find player URL')
+        file_url = 'https:' + file_url
+
+        title = get_input_by_name(element, 'title')
+        if not title:
+            title = self._html_search_regex(r'<span>(.+?)</span>', element, 'title')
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/radiocloud.py
```diff
@@ -0,0 +1,70 @@
...
+            return None
+
+        element = None
+        for e in get_elements_by_class("contents_box", webpage):
+            if get_input_by_name(e, 'content_id') == video_id:
+                element = e
+                break
+
+        if not element:
+            raise ExtractorError('Could not find details of id {}'.format(video_id))
```
dstftw: `{}` won't on python 2.6.

youtube_dl/extractor/radiocloud.py
```diff
@@ -0,0 +1,70 @@
...
+
+        def get_input_by_name(html, name):
+            for e in re.finditer(r'<input.*?>', html):
+                a = extract_attributes(e.group())
+                if a.get('name') == name:
+                    return a.get('value')
+            return None
+
+        element = None
+        for e in get_elements_by_class("contents_box", webpage):
```
dstftw: Single quotes.

youtube_dl/extractor/radiocloud.py
```diff
@@ -0,0 +1,70 @@
...
+            title = self._html_search_regex(r'<span>(.+?)</span>', element, 'title')
+        description = self._html_search_regex(r'<div>(.+?)</div>', element,
+                                              'description', fatal=False)
+
+        webpage = self._download_webpage(file_url, video_id,
+                                         headers={'Referer': url},
+                                         note='Downloading player',
+                                         errnote='Unable to download player')
+
+        file_url = self._search_regex(r'var\s*source\s*=\s*"(.+?)"', webpage, 'url')
```
dstftw: Should be `var\s+`.

youtube_dl/extractor/radiocloud.py
```diff
@@ -0,0 +1,70 @@
...
+    _VALID_URL = r'https?://(?:www\.)?radiocloud\.jp/archive/.*?content_id=(?P<id>[0-9]+)'
+    _TEST = {
+        'url': 'https://radiocloud.jp/archive/dc?content_id=15326',
+        'md5': 'd7bcf8b2877e694543a1f462b458a665',
+        'info_dict': {
+            'id': '15326',
+            'ext': 'm4a',
+            'title': '「オープニング」 ',
+            'description': '「“フィギュアスケートは人生”引退会見で浅田真央さんは何を語ったのか？」スポーツライター・青嶋ひろのさんが解説！',
+        }
```
dstftw: Test fails.

EliteTK: Interesting, it seems like for people without an account the page loads content a week or two into the past. For people who are logged in with an account the page loads content a month into the past.

I'll try to work out if I can find a way to load user restricted content without authentication. If that fails, I'll implement authentication (but I'll make it into its own commit instead of amending the current  commit like for the other fixes).

I'm not sure if there is any access to content older than a month. I'll fetch the extracted URL for a recording about to expire and see if I can still access it after it expires.

If I can't find a way to download content which the website does not give direct access to even for logged in users, what should I do for tests which have an "expiration date" so to speak?

EliteTK: @dstftw any ideas about this?

EliteTK:
@dstftw Sorry for the long wait. I've been a bit busy and was for a while happy with my own solution.
 
I've made the requested changes.

Upon further inspection there is an "end_date" or expiry date on each upload. I did a search across all uploads on the website (with a separate script) and found an upload which expires in 2038. When 2038 is approaching, feel free to ask for a new test :P .

So far I've made do with my own scripts but it would be really neat to get this in youtube_dl so I can use this feature directly with mpv.

PR #12539, opened 2017-03-24
[vier] Extract more info

mrBliss:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [X] At least skimmed through [adding new extractor tutorial](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/rg3/youtube-dl#youtube-dl-coding-conventions) sections
- [X] [Searched](https://github.com/rg3/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [X] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [X] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

Extract the `episode_number` and `upload_date`. Also extract the real
`description`.

The main goal of this PR is to make it easier to automate downloading whole seasons (hence the `episode_number` field).

PR #12801, opened 2017-04-20
[vier] Support authentication

mrBliss:
Full episodes now require authentication.

---

- [ ] [Searched](https://github.com/rg3/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests

- [ ] Bug fix
- [X] Improvement
- [ ] New extractor
- [ ] New feature

---

Since today or yesterday, vier.be requires users to log in before they can watch complete episodes. Short clips can still be watched without authentication.

This PR adds support for authentication.

### DIFF
--- a/youtube_dl/extractor/vier.py
+++ b/youtube_dl/extractor/vier.py
@@ -122,6 +122,9 @@ class VierIE(InfoExtractor):
                 'http://www.%s.be/video/v3/embed/%s' % (site, video_id),
                 display_id)
 
+        video_id = self._search_regex(
+            [r'data-nid="(\d+)"', r'"nid"\s*:\s*"(\d+)"'],
+            webpage, 'video id')
         application = self._search_regex(
             [r'data-application="([^"]+)"', r'"application"\s*:\s*"([^"]+)"'],
             webpage, 'application', default=site + '_vod')
--- a/youtube_dl/extractor/vier.py
+++ b/youtube_dl/extractor/vier.py
@@ -98,7 +98,7 @@ class VierIE(InfoExtractor):
 
         login_error = self._html_search_regex(
             r'(?s)<div class="messages error">\s*<div>\s*<h2.+?</h2>(.+?)<',
-            login_page, 'login error', default=None, fatal=False)
+            login_page, 'login error', default=None)
         if login_error:
             self.report_warning('Unable to log in: %s' % login_error)
         else:
@@ -117,7 +117,8 @@ class VierIE(InfoExtractor):
         webpage = self._download_webpage(url, display_id)
 
         if r'id="user-login"' in webpage:
-            self.report_warning('Log in to extract metadata', video_id=video_id)
+            self.report_warning(
+                'Log in to extract metadata', video_id=display_id)
             webpage = self._download_webpage(
                 'http://www.%s.be/video/v3/embed/%s' % (site, video_id),
                 display_id)
--- a/youtube_dl/extractor/vier.py
+++ b/youtube_dl/extractor/vier.py
@@ -96,11 +96,13 @@ class VierIE(InfoExtractor):
             }),
             headers={'Content-Type': 'application/x-www-form-urlencoded'})
 
-        if 'Gebruikersnaam of wachtwoord is onbekend' in login_page:
-            raise ExtractorError(
-                'Unable to login: incorrect credentials', expected=True)
-
-        self._logged_in = True
+        login_error = self._html_search_regex(
+            r'(?s)<div class="messages error">\s*<div>\s*<h2.+?</h2>(.+?)<',
+            login_page, 'login error', default=None, fatal=False)
+        if login_error:
+            self.report_warning('Unable to log in: %s' % login_error)
+        else:
+            self._logged_in = True
 
     def _real_extract(self, url):
         mobj = re.match(self._VALID_URL, url)
--- a/youtube_dl/extractor/vier.py
+++ b/youtube_dl/extractor/vier.py
@@ -6,6 +6,7 @@ import itertools
 
 from .common import InfoExtractor
 from ..utils import (
+    ExtractorError,
     urlencode_postdata,
 )
 
@@ -85,7 +86,7 @@ class VierIE(InfoExtractor):
         if username is None or password is None:
             return
 
-        self._request_webpage(
+        login_page = self._download_webpage(
             'http://www.%s.be/user/login' % site,
             None, note='Logging in', errnote='Unable to log in',
             data=urlencode_postdata({
@@ -94,6 +95,11 @@ class VierIE(InfoExtractor):
                 'pass': password,
             }),
             headers={'Content-Type': 'application/x-www-form-urlencoded'})
+
+        if 'Gebruikersnaam of wachtwoord is onbekend' in login_page:
+            raise ExtractorError(
+                'Unable to login: incorrect credentials', expected=True)
+
         self._logged_in = True
 
     def _real_extract(self, url):
--- a/youtube_dl/extractor/vier.py
+++ b/youtube_dl/extractor/vier.py
@@ -5,12 +5,16 @@ import re
 import itertools
 
 from .common import InfoExtractor
+from ..utils import (
+    urlencode_postdata,
+)
 
 
 class VierIE(InfoExtractor):
     IE_NAME = 'vier'
     IE_DESC = 'vier.be and vijf.be'
     _VALID_URL = r'https?://(?:www\.)?(?P<site>vier|vijf)\.be/(?:[^/]+/videos/(?P<display_id>[^/]+)(?:/(?P<id>\d+))?|video/v3/embed/(?P<embed_id>\d+))'
+    _NETRC_MACHINE = 'vier'
     _TESTS = [{
         'url': 'http://www.vier.be/planb/videos/het-wordt-warm-de-moestuin/16129',
         'info_dict': {
@@ -44,6 +48,24 @@ class VierIE(InfoExtractor):
         'url': 'http://www.vier.be/video/v3/embed/16129',
         'only_matching': True,
     }]
+
+    def _real_initialize(self):
+        self._login()
+
+    def _login(self):
+        username, password = self._get_login_info()
+        if username is None:
+            return
+
+        self._request_webpage(
+            'http://www.vier.be/user/login',
+            None, note='Logging in', errnote='Could not log in',
+            data=urlencode_postdata({
+                'form_id': 'user_login',
+                'name': username,
+                'pass': password,
+            }),
+            headers={'Content-Type': 'application/x-www-form-urlencoded'})
 
     def _real_extract(self, url):
         mobj = re.match(self._VALID_URL, url)
--- a/youtube_dl/extractor/vier.py
+++ b/youtube_dl/extractor/vier.py
@@ -49,16 +49,13 @@ class VierIE(InfoExtractor):
         'only_matching': True,
     }]
 
-    def _real_initialize(self):
-        self._login()
-
-    def _login(self):
+    def _login(self, site):
         username, password = self._get_login_info()
-        if username is None:
-            return
+        if username is None or password is None:
+            self.raise_login_required()
 
         self._request_webpage(
-            'http://www.vier.be/user/login',
+            'http://www.%s.be/user/login' % site,
             None, note='Logging in', errnote='Could not log in',
             data=urlencode_postdata({
                 'form_id': 'user_login',
@@ -74,6 +71,10 @@ class VierIE(InfoExtractor):
         site = mobj.group('site')
 
         webpage = self._download_webpage(url, display_id)
+
+        if re.search(r'id="user-login"', webpage):
+            self._login(site)
+            webpage = self._download_webpage(url, display_id)
 
         video_id = self._search_regex(
             [r'data-nid="(\d+)"', r'"nid"\s*:\s*"(\d+)"'],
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/vier.py
```diff
@@ -45,6 +49,24 @@ class VierIE(InfoExtractor):
...
+    def _real_initialize(self):
+        self._login()
+
+    def _login(self):
+        username, password = self._get_login_info()
+        if username is None:
+            return
+
+        self._request_webpage(
+            'http://www.vier.be/user/login',
```
mrBliss: @dstftw: Because `_login` doesn't have access to the URL, we don't know here whether we're dealing with vier.be or vijf.be, so I *temporarily* hardcoded vier.be here.

How do other extractors deal with this? I couldn't find anything by grepping for `<site>` + `login`.

My current plan: add `site` as parameter to the `_login` method (rename `_login` to something else?) and call it in `_real_extract` after we have extracted the `site`. Furthermore, set and check `self._logged_in` so authentication happens at most once.

youtube_dl/extractor/vier.py
```diff
@@ -45,6 +49,24 @@ class VierIE(InfoExtractor):
         'only_matching': True,
     }]
 
+    def _real_initialize(self):
+        self._login()
```
mrBliss: @dstftw: There is no reliable way to determine whether a video requires authentication based on the URL. Only after the webpage is downloaded can we check this. How do I best deal with this?

My current plan: search for a login form in the webpage, if there's one, log in and retry.

midas02:
@mrBliss Have you checked my comment at https://github.com/rg3/youtube-dl/issues/7168? Apparently there is a way to avoid authentication, but last time I checked, I couldn't find videos requiring authentication to test if that claim is still valid.

mrBliss:
@midas02 Thanks for the hint! I have pushed a new version that uses the trick mentioned in the comment.

@dstftw You can ignore the questions I asked before. If you want, I'll rebase away the three intermediary commits.

midas02:
Do you have an example of a video requiring authentication to test the mechanism? A couple of weeks ago, I couldn't find any.
Ps: be aware that 90% of Vier videos are geo-restricted. I can only access them through a VPN.

mrBliss:
Since today (maybe yesterday already), all videos on http://www.vier.be/volledige-afleveringen require authentication, at least for me. Can you check whether you need authentication too?

midas02:
Ahaa... ok, they changed everything it seems. By the way, these are not geo-restricted: Baksteen in de Maag, Bartel, De Sollicitatie, Rijden doe je Zo, and a few more.

youtube_dl/extractor/vier.py
```diff
@@ -49,13 +49,12 @@ def _real_extract(self, url):
...
+        video_id = mobj.group('id') or embed_id
         site = mobj.group('site')
 
-        webpage = self._download_webpage(url, display_id)
-
-        video_id = self._search_regex(
-            [r'data-nid="(\d+)"', r'"nid"\s*:\s*"(\d+)"'],
-            webpage, 'video id')
+        webpage = self._download_webpage(
+            "http://www.%s.be/video/v3/embed/%s" % (site, video_id),
```
dstftw: Use single quotes. This change breaks all tests. No purpose of this change described, no tests. Change does not match PRs title.

mrBliss:
Problem: bypassing the authentication by using the `http://www.vier.be/video/v3/embed/<video_id>`[trick](https://bitbucket.org/basrieter/xbmc-online-tv/issues/809) works fine for downloading the video, but we lose access to all the metadata (title, description, etc.). That's the reason why the tests are now broken.

I fear that I might have to switch back to actually performing authentication instead of bypassing it.

mrBliss:
@dstftw can you look at this, if you have some time to spare?

youtube_dl/extractor/vier.py
```diff
@@ -53,6 +85,10 @@ def _real_extract(self, url):
 
         webpage = self._download_webpage(url, display_id)
 
+        if re.search(r'id="user-login"', webpage):
```
dstftw: `if 'id="user-login"' in webpage:`.

youtube_dl/extractor/vier.py
```diff
@@ -38,13 +42,41 @@ class VierIE(InfoExtractor):
...
         'only_matching': True,
     }, {
         'url': 'http://www.vier.be/video/v3/embed/16129',
         'only_matching': True,
     }]
 
+    def _login(self, site):
+        username, password = self._get_login_info()
+        if username is None or password is None:
+            self.raise_login_required()
```
dstftw: There could be a fallback to video/embed URL and extraction without metadata with a warning to use a credentials for correct metadata printed.

youtube_dl/extractor/vier.py
```diff
@@ -38,13 +42,41 @@ class VierIE(InfoExtractor):
...
+
+        self._request_webpage(
+            'http://www.%s.be/user/login' % site,
+            None, note='Logging in', errnote='Could not log in',
```
dstftw: Standard message is `Unable to log in`.

youtube_dl/extractor/vier.py
```diff
@@ -53,6 +85,10 @@ def _real_extract(self, url):
 
         webpage = self._download_webpage(url, display_id)
```
dstftw: If makes sense to login in the first place when credentials are provided.

mrBliss:
@dstftw Thanks for the feedback, I have addressed your comments.

youtube_dl/extractor/vier.py
```diff
@@ -24,38 +28,93 @@ class VierIE(InfoExtractor):
...
+    }, {
         'url': 'http://www.vier.be/planb/videos/mieren-herders-van-de-bladluizen',
         'only_matching': True,
     }, {
         'url': 'http://www.vier.be/video/v3/embed/16129',
         'only_matching': True,
     }]
 
+    def _real_initialize(self):
+        self._logged_in = False
```
dstftw: There are two unrelated flags `_logged_in` and `logged_in`.

youtube_dl/extractor/vier.py
```diff
@@ -24,38 +28,93 @@ class VierIE(InfoExtractor):
...
+        self._request_webpage(
+            'http://www.%s.be/user/login' % site,
+            None, note='Logging in', errnote='Unable to log in',
+            data=urlencode_postdata({
+                'form_id': 'user_login',
+                'name': username,
+                'pass': password,
+            }),
+            headers={'Content-Type': 'application/x-www-form-urlencoded'})
+        self.logged_in = True
```
dstftw: This is not necessarily true. Login errors should be detected and output.

youtube_dl/extractor/vier.py
```diff
@@ -24,38 +28,93 @@ class VierIE(InfoExtractor):
...
         'only_matching': True,
     }]
 
+    def _real_initialize(self):
+        self._logged_in = False
+
+    def _login(self, site):
+        username, password = self._get_login_info()
+        if username is None or password is None:
+            self.logged_in = False
```
dstftw: This is not true either. Login may be achieved via authorized cookies.

youtube_dl/extractor/vier.py
mrBliss: Oops

youtube_dl/extractor/vier.py
mrBliss: Indeed, I will add a check.

youtube_dl/extractor/vier.py
mrBliss: Yes, but it doesn't matter. I only use `self._logged_in` to determine whether we should log in (in `self._real_extract`, because we need to extract the `site` from the URL in order to log in).

I do not use `self._logged_in` to determine whether we should fall back to the embed URL trick. The fallback happens when a login form was detected in the page, which won't happen if the user either passed credentials, or passed authorisation cookies.

Also keep in mind that when no credentials are passed, `self._login` just does nothing.

youtube_dl/extractor/vier.py
```diff
@@ -24,38 +29,97 @@ class VierIE(InfoExtractor):
...
+            None, note='Logging in', errnote='Unable to log in',
+            data=urlencode_postdata({
+                'form_id': 'user_login',
+                'name': username,
+                'pass': password,
+            }),
+            headers={'Content-Type': 'application/x-www-form-urlencoded'})
+
+        if 'Gebruikersnaam of wachtwoord is onbekend' in login_page:
+            raise ExtractorError(
```
dstftw: No exception here just warning.

youtube_dl/extractor/vier.py
```diff
@@ -24,38 +29,97 @@ class VierIE(InfoExtractor):
...
+                'Unable to login: incorrect credentials', expected=True)
```
dstftw: You should **capture**  error message and **output** it.

mrBliss:
I've pushed another commit addressing your latest remarks.

youtube_dl/extractor/vier.py
```diff
@@ -24,38 +29,99 @@ class VierIE(InfoExtractor):
...
         site = mobj.group('site')
 
+        if not self._logged_in:
+            self._login(site)
+
         webpage = self._download_webpage(url, display_id)
 
-        video_id = self._search_regex(
-            [r'data-nid="(\d+)"', r'"nid"\s*:\s*"(\d+)"'],
-            webpage, 'video id')
```
dstftw: This should not be removed.

mrBliss: Why not? The `video_id` is now extracted from the URL by [L111](https://github.com/rg3/youtube-dl/pull/12801/files#diff-33948a4255404024a2f2ef084e9c2b03R111), the webpage is not needed for this.

I have verified for all tests videos that the `video_id` extracted from the URL is the same as the one extracted from the webpage. The only exception is http://www.vier.be/planb/videos/mieren-herders-van-de-bladluizen because it is no longer available. I checked the website and I couldn't find any video URL that didn't contain the `video_id` in the URL. I guess I should just replace the outdated test URL with a recent one.

dstftw: Not all URLs contain video id.

mrBliss: Alright, after searching some more, I did find some videos not containing the `video_id` in the URL. So I have resurrected the snippet. (BTW: it would be better if you just directly linked to a counterexample instead of making a blanket statement.)

We rely on the `video_id` from the URL for the fallback, so when it is missing, the fallback will fail. However, I believe only videos not requiring authentication and thus not requiring the fallback can have a URL not containing the `video_id`. So do you agree that we need no additional check/warning for this?

youtube_dl/extractor/vier.py
```diff
@@ -24,35 +29,99 @@ class VierIE(InfoExtractor):
...
+            data=urlencode_postdata({
+                'form_id': 'user_login',
+                'name': username,
+                'pass': password,
+            }),
+            headers={'Content-Type': 'application/x-www-form-urlencoded'})
+
+        login_error = self._html_search_regex(
+            r'(?s)<div class="messages error">\s*<div>\s*<h2.+?</h2>(.+?)<',
+            login_page, 'login error', default=None, fatal=False)
```
dstftw: `default` implies non `fatal`.

youtube_dl/extractor/vier.py
```diff
@@ -24,35 +29,99 @@ class VierIE(InfoExtractor):
...
+        video_id = mobj.group('id') or embed_id
         site = mobj.group('site')
 
+        if not self._logged_in:
+            self._login(site)
+
         webpage = self._download_webpage(url, display_id)
 
+        if r'id="user-login"' in webpage:
+            self.report_warning('Log in to extract metadata', video_id=video_id)
```
dstftw: Should be `display_id`.

youtube_dl/extractor/vier.py
```diff
@@ -24,35 +29,99 @@ class VierIE(InfoExtractor):
...
+            webpage = self._download_webpage(
+                'http://www.%s.be/video/v3/embed/%s' % (site, video_id),
+                display_id)
+
         video_id = self._search_regex(
             [r'data-nid="(\d+)"', r'"nid"\s*:\s*"(\d+)"'],
             webpage, 'video id')
```
dstftw: There could be a fallback with id extracted from URL.

mrBliss: What do you mean? That line is exactly the fallback with the id extracted from the URL.

dstftw: No, it's not. If video_id extraction from page fails whole extraction fails.

mrBliss: I have not been able to find a video for which the `video_id` extraction from the page fails. All pages seem to contain the `video_id`, even pages of which the URL doesn't contain the `video_id`.

Could you please link to a video for which this happens?  It's hard to fix something when I have to keep guessing the scenario you are talking about.

dstftw: What's not clear? `video_id` extraction from `webpage` does not participate in formats extraction thus should not be fatal. It does not matter whether you can find an example URL or not for which `video_id` extraction fails. It may not fail now but may start failing in future (e.g. due to regexes change) breaking the whole extraction.
I don't have such example URL. Change regexes to invalid ones to test such scenario.

mrBliss: Thank you for the explanation, now it is clear to me what you meant. I have pushed a commit that addresses this.

P.S. I now realise that my confusion stems from the fact that I thought your first comment was about lines 121-123 instead of 125-127 (I was only looking at the green parts of the diff).

mrBliss:
@dstftw Thanks again for you patience with me.

I have rebased and updated #12539. Whenever you have some more time, feel free to take a look at it.

Issue #12832, opened 2017-04-25
[BBC Radio 4] "best" downloads "worst"?


optimiz:
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2017.04.17**
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)

---
Hello, first of all, thanks for this very useful program.  When attempting to download the "best" version of a BBC Radio 4 show, youtube-dl ends up downloading the "worst" version instead.  Using the following example, which has a 320k stream, specify "-f best", youtube-dl passes the parsed m3u8 to ffmpeg for download, but the only stream downloaded is the 48k stream, the other three streams (including the requested best 320k one) are dropped with a "_No longer receiving playlist #_" notice.  (**It is _not_ geoblocked.**)

`$ youtube-dl -f best http://www.bbc.co.uk/programmes/b08mb1gb`
>[bbc.co.uk] b08mb1gb: Downloading video page
[bbc.co.uk] b08mb1fh: Downloading media selection XML
[bbc.co.uk] b08mb1fh: Downloading m3u8 information
[bbc.co.uk] b08mb1fh: Downloading m3u8 information
[bbc.co.uk] b08mb1fh: Downloading m3u8 information
[bbc.co.uk] b08mb1fh: Downloading m3u8 information
[bbc.co.uk] b08mb1fh: Downloading m3u8 information
[bbc.co.uk] b08mb1fh: Downloading m3u8 information
[bbc.co.uk] b08mb1fh: Downloading m3u8 information
[bbc.co.uk] b08mb1fh: Downloading m3u8 information
[download] Destination: b08mb1fh.mp4
ffmpeg version 2.8.10 Copyright (c) 2000-2016 the FFmpeg developers
  built with gcc 5.3.1 (GCC) 20160406 (Red Hat 5.3.1-6)
[...]

PR #12839, opened 2017-04-25
Don't list master m3u8 playlists in format list

dstftw:
- [x] Improvement

---

None of the manifests (HDS, DASH, ISM) are listed in `-F` apart from HLS that is inconsistent. I propose not to show HLS manifest in format list either for symmetry. This will allow to make `worst*` format selectors to actually work and fix situations like in #12832.

This may probably break some extractor logic based on m3u8 formats count.

### DIFF
--- a/test/test_InfoExtractor.py
+++ b/test/test_InfoExtractor.py
@@ -184,14 +184,6 @@ jwplayer("mediaplayer").setup({"abouttext":"Visit Indie DB","aboutlink":"http:\/
                 'pluzz_francetv_11507',
                 'http://replayftv-vh.akamaihd.net/i/streaming-adaptatif_france-dom-tom/2017/S16/J2/156589847-58f59130c1f52-,standard1,standard2,standard3,standard4,standard5,.mp4.csmil/master.m3u8?caption=2017%2F16%2F156589847-1492488987.m3u8%3Afra%3AFrancais&audiotrack=0%3Afra%3AFrancais',
                 [{
-                    'url': 'http://replayftv-vh.akamaihd.net/i/streaming-adaptatif_france-dom-tom/2017/S16/J2/156589847-58f59130c1f52-,standard1,standard2,standard3,standard4,standard5,.mp4.csmil/master.m3u8?caption=2017%2F16%2F156589847-1492488987.m3u8%3Afra%3AFrancais&audiotrack=0%3Afra%3AFrancais',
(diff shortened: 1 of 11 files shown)


yan12125:
Some 3rd party applications may want to access the URL to the master playlist (for implementing custom format selection, etc.) Removing the meta format will break them. Maybe adding a new field in HLS formats for the relevant master playlist?

By the way, the call to ```_m3u8_meta_format``` in anvato.py should be removed, too.

dstftw:
There is already such field available `manifest_url`.

yan12125:
For HLS formats, 'manifest_url' has the same content as 'url'. They're both media playlists: https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L1441. Here's an example:
```
$ curl $(youtube-dl -j test:twitter | jq -r '.formats[-1].manifest_url')
#EXTM3U
#EXT-X-VERSION:6
#EXT-X-MEDIA-SEQUENCE:0
#EXT-X-TARGETDURATION:4
#EXT-X-ALLOW-CACHE:YES
#EXTINF:3.000,
/ext_tw_video/643211870443208704/pu/vid/0/3000/480x480/drJuCDtPWmhsWQgu.ts
#EXTINF:3.000,
/ext_tw_video/643211870443208704/pu/vid/3000/6000/480x480/T3-KFiRCoKI-GxEh.ts
#EXTINF:3.000,
/ext_tw_video/643211870443208704/pu/vid/6000/9000/480x480/bQqxYujkT6S_lltC.ts
#EXTINF:3.922,
/ext_tw_video/643211870443208704/pu/vid/9000/12922/480x480/3s5y6cPmIeFJMrx_.ts
#EXT-X-ENDLIST
```

dstftw:
It's a bug.

dstftw:
Alternatively, on opposite may be done: include all manifests to format list but don't consider in format selection

yan12125:
I feel changing the content of ```manifest_url``` in HLS formats is a better solution than including 'hidden' formats. I guess hidden formats will bring headaches to both maintainers as well as users.

yan12125:
Thanks. Could you remove the meta format from anvato.py too?

dstftw:
Cherry-picked.

yan12125:
Thanks! I guess ```InfoExtractor._m3u8_meta_format``` can be removed as well as there are no clients now?

PR #12937, opened 2017-04-30
[Rmc decouverte] Fix extractor

midas02:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Extractor was no longer working. Fixed now.

### DIFF
--- a/youtube_dl/extractor/rmcdecouverte.py
+++ b/youtube_dl/extractor/rmcdecouverte.py
@@ -15,13 +15,13 @@ class RMCDecouverteIE(InfoExtractor):
     _TEST = {
         'url': 'http://rmcdecouverte.bfmtv.com/mediaplayer-replay/?id=116548',
         'info_dict': {
-            'id': '5111223049001',
+            'id': '5411254766001',
             'ext': 'mp4',
-            'title': ': LES HEROS DU 88e ETAGE',
-            'description': 'Découvrez comment la bravoure de deux hommes dans la Tour Nord du World Trade Center a sauvé  la vie d\'innombrables personnes le 11 septembre 2001.',
+            'title': '39/45:LE RESEAU DES FAUX BILLETS',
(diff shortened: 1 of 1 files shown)


midas02:
Some comments.

1. The old code seemed to be overly complicated, but I've left it in place since I don't know if it still serves a purpose. The new code has been added with an If/else statement.
2. The code will fail if `brightcove_id` returns none. Is there a way of catching this?
3. The `data-video-id` key is referenced between ```<div class="brightcove_container">... </div>```. Should its detection be made conditional?

dstftw:
2. Don't pass `default` and it won't be `None`.
3. Conditional to what?

midas02:
2. That's not what I meant. Is there a way of returning an error message if the video_id is not found. Currently it just seems as if the programme has crashed.
3. Conditional to being caught in <div class="brightcove_container">.

dstftw:
>a way of returning an error message if the video_id is not found

This is exactly what will happen if no `default` is passed.

>Conditional to being caught in `<div class="brightcove_container">`.

Explain **distinctly** what are you trying to do.

midas02:
Well, a test to extract `data-video-id` only if it is being referenced in a class `"brightcove_container"`. I've tried to implement it, but in my opinion it will make the regex needlessly complicated and less robust.

youtube_dl/extractor/rmcdecouverte.py
```diff
@@ -35,5 +35,8 @@ def _real_extract(self, url):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
         brightcove_legacy_url = BrightcoveLegacyIE._extract_brightcove_url(webpage)
-        brightcove_id = compat_parse_qs(compat_urlparse.urlparse(brightcove_legacy_url).query)['@videoPlayer'][0]
+        if brightcove_legacy_url:
+            brightcove_id = compat_parse_qs(compat_urlparse.urlparse(brightcove_legacy_url).query)['@videoPlayer'][0]
+        else:
+            brightcove_id = self._html_search_regex(r'data-video-id="(.*?)"', webpage, 'error message')
```
dstftw: There is no point extracting empty video id. What the `error message`?

Issue #9676, opened 2016-06-02
output template support for %(channel_id)s


glenn-slayden:
## Please follow the guide below
- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your _issue_ (like that [x])
- Use _Preview_ tab to see how your issue will actually look like

---
### Make sure you are using the _latest_ version: run `youtube-dl --version` and ensure your version is _2016.06.02_. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.06.02**
### Before submitting an _issue_ make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
### What is the purpose of your _issue_?
- [ ] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [x] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

The output template should have a way to reference the **YouTube channel id**, i.e.

`%(channel_id)s`

Currently,  `%(uploader_id)s` falls back to rendering the channel id if the uploader doesn't have a name, but there's no way to force this behavior if the uploader _does_ have a name.

_Note:_ If this feature is already available somewhere in youtube-dl, then I could not find it in the documentation, so documentation should be added or improved.

Synetech:
You can use `%(uploader)s` to force it to use the uploader’s name rather than their ID (it’s the second item listed in the `--help`).

Synetech:
You can use `%(uploader)s` to force it to use the uploader’s name rather than their ID (it’s the second item listed in the `--help`).

Synetech:
You can use `%(uploader)s` to force it to use the uploader’s name rather than their ID (it’s the second item listed in the `--help`).

Synetech:
You can use `%(uploader)s` to force it to use the uploader’s name rather than their ID (it’s the second item listed in the `--help`).

Synetech:
[...]

PR #12939, opened 2017-04-30
add code from @glenn-slayden

Minimally:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Resolve issue #9676
output template support for %(channel_id)s

Implement the code provided by @glenn-slayden to specify %(channel_id)s in youtube template output. This is a relatively small Improvement that allows youtube-dl to work with channel ids that are the default way of identifying channels in Data API v3. See: https://developers.google.com/youtube/v3/guides/working_with_channel_ids

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -1455,6 +1455,14 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
         # Start extracting information
         self.report_information_extraction(video_id)
 
+        # channel_id
+        mobj = re.search(r'itemprop="channelId" content="(UC[-_A-Za-z0-9]{21}[AQgw])"', video_webpage)
+        if mobj is not None:
+            video_channel_id = mobj.group(1)
+        else:
+            self._downloader.report_warning('unable to extract channelId')
+            video_channel_id = None
+
         # uploader
         if 'author' not in video_info:
             raise ExtractorError('Unable to extract uploader name')
@@ -1773,6 +1781,7 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
 
         return {
             'id': video_id,
+            'channel_id': video_channel_id,
             'uploader': video_uploader,
             'uploader_id': video_uploader_id,
             'uploader_url': video_uploader_url,


dstftw:
1. Write meaningful commit names.
2. First of all this meta field should be [introduced](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L89-L285). We only add generic meta fields that can be applied to any extractor thus you'll have to demonstrate how it can be used for another extractors.

glenn-slayden:
>@dstftw commented on Apr 30:
 First of all this meta field should be _introduced_.

The reason a new meta field was used here is because the youtube extractor currently has inconsistent/unpredictable behavior in its implementation of `%(uploader_id)s` . The problem is explained in detail [here](https://github.com/rg3/youtube-dl/issues/9676). In short, currently `%(uploader_id)s` unpredictably returns either an identifier of the form

    UC[-_A-Za-z0-9]{21}[AQgw]

...or a human-readable name. For a so-called "id" meta field, the latter behvaior seems patently incorrect, but I imagined that fixing it so that `%(uploader_id)s` always returns the **channel id** would break existing scenarios that prefer a human-readable name when available, and I don't know the `youtube-dl` policy on breaking changes.

The true root of the problem is found in the **meta** definitions you linked to:

    uploader_id:    Nickname or id of the video uploader

At a basic level, why would any kind of "id" ever be conflated a "name"?  As a technical specification, "nickname **or** id" is hardly useful, if both can ever distinctly apply. It's lossy at worst, and inconsistent at best. It seems to me that `uploader_id` and `uploader_nickname` should be seprate meta fields _from the very start_, and if one or the other is not available, ***only then*** can you have them gracefully fail-over.

@dstftw, please advise on how to proceed with resolving this issue: Should I change the behavior of the existing meta field `%(uploader_id)s`, fixing it so that it, in fact, always returns a (semantic) "id" like it should? If so, which existing meta field should the human-readable "uploader nickname" be (consistently) mapped to? The fail-over cases follow naturally from these choices. Many thanks.

dstftw:
>At a basic level, why would any kind of "id" ever be conflated a "name"?  As a technical specification, "nickname **or** id" is hardly useful, if both can ever distinctly apply.

It's **not name**. It's **nickname** that is usually used to identify user on some service. It's a **generic field** that is used across **many extractors**, some may use id for identification, some may use nickname or login. Again it's **not a technical specification** and it can't be.

>please advise on how to proceed with resolving this issue

Start with reading my previous post.

glenn-slayden:
Obviously, I did, and it seems like you did not read mine. Except for dubiously asserting that **youtube-dl** "can't" exhibit consistent behavior, you didn't address the information-loss/conflation issue that I raised in some detail. I also proposed a specific alternative in the last paragraph, seeking your guidance about policy preference. It's totally fine if you don't care; it would be less brusque to simply say so.

dstftw:
No, you didn't read my post.

I'm well aware of the problem you've described, no need to repeat it over and over again. Though it's not even a problem per se. Both, username and channel id, clearly identify the uploader from different sides thus it's perfectly fine. `uploader_id` gives **no guarantees on predictability** it provides a **unique identifier** for uploader in arbitrary form.

Now let's go over my post that you didn't manage to do yourself:

>First of all this meta field should be [**introduced**](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/common.py#L89-L285)

`channel_id` is not introduced.

>We only add generic meta fields that can be applied to any extractor thus you'll have to **demonstrate** how it can be used **for another extractors**.

No examples apart from Youtube are demonstrated.

glenn-slayden:
We are talking past each other. My proposal was simply to make the format emitted by `uploader_id` **consistent** for the youtube extractor. There are [use cases](https://github.com/rg3/youtube-dl/issues/9676) for having a predictable, machine-readable format, which are ruled out by insisting on the opaqueness of this meta field. That's prefectly fine; obviously it's your project.

As noted, the _only_ reason my fix defined a new meta field was to _avoid breaking existing functionality_ for the youtube extractor. I never claimed it made sense globally; relevance to other extractors is a _non-sequitur_.

Now although you didn't specifically say so, it logically follows from the opaqueness policy you just stated that you would not object to fixing this problem by **_always returning the (e.g.) UCLqxVugv74EIW3VWh2NOa3Q format_** for the `%(uloader_id)s` meta field.

I'm not trying to be snarky here, but this would be a much simpler fix which satisfies your philosophy, introduces no new meta fields and would completely resolve my issue. It would also irreparably break the existing persistent directory structures for thousands of users. Would you in fact approve such a change?

glenn-slayden:
Thinking further about the fix I just outlined, it might be possible to mitigate the effects on existing directory structures by alerting users to change their templates. The trouble is, I seem to recall that `%(uploader)`  isn't necessarily the same as the "nickname" which (currently) is sometimes returned in `%(uploader_id)` (?). If these can be different, then my proposed change would render the "nickname" inaccessible, right?

dstftw:
>Now although you didn't specifically say so, it logically follows from the opaqueness policy you just stated that you would not object to fixing this problem by **_always returning the (e.g.) UCLqxVugv74EIW3VWh2NOa3Q format_** for the `%(uploader_id)s` meta field.

Absolutely not, it does not follows and I didn't state anything like that. `uploader_id`'s behavior should be preserved and any change should be backward compatible whenever possible.

>I seem to recall that `%(uploader)`  isn't necessarily the same as the "nickname" which (currently) is sometimes returned in `%(uploader_id)`

Yes, these are [different](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/youtube.py#L471-L472) meta fields.

>my proposed change would render the "nickname" inaccessible, right?

Yes.

glenn-slayden:
>...I didn't state anything like that

And I didn't say you did. My point there was that token _opaqueness_ policy interacts with token _persistence_ policy. If users expect  `%(uploader_id)s` values to remain valid beyond a single session then "opaqueness" may be violated in some sense to facilitate that local strorage. An example would be tokens used in directory/file paths not containing certain characters, etc. 

> any change should be backward compatible whenever possible.

Right, which is why a different meta field was used, thus leaving the inconsistent behavior of `%(uploader_id)s` as-is.

Thanks.

PR #9409, opened 2016-05-05
introduce chapters field

remitamine:

PR #12949, opened 2017-05-01
introduce chapters field(2)

remitamine:
- [x] New feature
---
continuation of #9409, it should work now in both unix and windows systems.

### DIFF
--- a/youtube_dl/extractor/pbs.py
+++ b/youtube_dl/extractor/pbs.py
@@ -8,6 +8,7 @@ from ..utils import (
     ExtractorError,
     determine_ext,
     int_or_none,
+    float_or_none,
     js_to_json,
     strip_jsonp,
     strip_or_none,
@@ -464,6 +465,7 @@ class PBSIE(InfoExtractor):
                     redirects.append(redirect)
                     redirect_urls.add(redirect_url)
 
+        chapters = []
         # Player pages may also serve different qualities
         for page in ('widget/partnerplayer', 'portalplayer'):
             player = self._download_webpage(
@@ -479,6 +481,20 @@ class PBSIE(InfoExtractor):
                     extract_redirect_urls(video_info)
                     if not info:
                         info = video_info
+                if not chapters:
+                    for chapter_data in re.findall(r'(?s)chapters\.push\(({.*?})\)', player):
+                        chapter = self._parse_json(chapter_data, video_id, js_to_json, fatal=False)
+                        if not chapter:
+                            continue
+                        start_time = float_or_none(chapter.get('start_time'), 1000)
+                        duration = float_or_none(chapter.get('duration'), 1000)
+                        if start_time is None or duration is None:
+                            continue
+                        chapters.append({
+                            'start_time': start_time,
+                            'end_time': start_time + duration,
+                            'title': chapter.get('title'),
+                        })
 
         formats = []
         http_url = None
@@ -588,4 +604,5 @@ class PBSIE(InfoExtractor):
             'upload_date': upload_date,
             'formats': formats,
             'subtitles': subtitles,
+            'chapters': chapters,
         }
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -4,6 +4,7 @@ import io
 import os
 import subprocess
 import time
+import re
 
 
 from .common import AudioConversionError, PostProcessor
@@ -22,6 +23,7 @@ from ..utils import (
     subtitles_filename,
     dfxp2srt,
     ISO639Utils,
+    replace_extension,
 )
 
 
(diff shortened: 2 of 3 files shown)


youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -429,20 +431,43 @@ def add(meta_list, info_list=None):
 
         filename = info['filepath']
         temp_filename = prepend_extension(filename, 'temp')
+        in_filenames = [filename]
+        tmp_filenames = []
+
+        options = ['-map_metadata', '1']
```
dstftw: This must be applied only in case of chapters.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -429,20 +431,43 @@ def add(meta_list, info_list=None):
...
+                for chapter in chapters:
+                    metadata_file_content += '[CHAPTER]\nTIMEBASE=1/1000\n'
+                    metadata_file_content += 'START=%d\n' % (chapter['start_time'] * 1000)
+                    metadata_file_content += 'END=%d\n' % (chapter['end_time'] * 1000)
+                    chapter_title = chapter.get('title')
+                    if chapter_title:
+                        metadata_file_content += 'title=%s\n' % ffmpeg_escape(chapter_title)
+                f.write(metadata_file_content)
+                in_filenames.append(metadata_filename)
+                tmp_filenames.append(metadata_filename)
```
dstftw: What's the point of this? This is an implementation detail and it should not leave any trace if `-k` is used.

remitamine: i remember that @yan12125 requested before to keep the `metadata` file.

youtube_dl/postprocessor/ffmpeg.py
remitamine: it might be a good idea to keep the original metadata even when there is no `chapters`.

dstftw: You now have a crash in all cases when no chapters are available cause you reference to a metadata files that does not exist. What original metadata are you talking about?

youtube_dl/postprocessor/ffmpeg.py
yan12125: Yes - keeping intermediate files is useful for debugging. People can run the same ffmpeg command again and see what's going on.

dstftw: `-k` is supposed to work for explicit files not for debugging purposes.

youtube_dl/postprocessor/ffmpeg.py
remitamine: ahhh, as i created the patch a long time, i forget that the `1` is a index of input file and it doesn't map metadata from the original video file, will fix it.

youtube_dl/postprocessor/ffmpeg.py
yan12125: OK. I'll open an issue for discussing this. For now you can remove this line.

remitamine:
anything else?

youtube_dl/extractor/pbs.py
```diff
@@ -479,6 +481,21 @@ def extract_redirect_urls(info):
                     extract_redirect_urls(video_info)
                     if not info:
                         info = video_info
+                if not chapters:
+                    chapters_data = re.findall(r'(?s)chapters\.push\(({.*?})\)', player) or []
```
dstftw: `or []` is superfluous here.

remitamine: removed.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -429,17 +431,40 @@ def add(meta_list, info_list=None):
...
+                    if chapter_title:
+                        metadata_file_content += 'title=%s\n' % ffmpeg_escape(chapter_title)
+                f.write(metadata_file_content)
+                in_filenames.append(metadata_filename)
+                options.extend(['-map_metadata', '1'])
+
         self._downloader.to_screen('[ffmpeg] Adding metadata to \'%s\'' % filename)
-        self.run_ffmpeg(filename, temp_filename, options)
+        self.run_ffmpeg_multiple_files(in_filenames, temp_filename, options)
+        if chapters and os.path.isfile(metadata_filename):
```
dstftw: Basically `os.path.isfile(metadata_filename)` is superfluous here since we control the file lifetime on our own and since we don't delete the file it should exist. This check may only fail if someone touched our file that is unexpected scenario that normally should not happen. In such cases we should stop right at failed `os.remove` rather than skipping such unexpected outcome with this check. If someone touches our files then it's definitely wrong and we should not continue.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -429,17 +431,40 @@ def add(meta_list, info_list=None):
...
-            options = ['-vn', '-acodec', 'copy']
+            options.extend(['-vn', '-acodec', 'copy'])
         else:
-            options = ['-c', 'copy']
+            options.extend(['-c', 'copy'])
 
         for (name, value) in metadata.items():
             options.extend(['-metadata', '%s=%s' % (name, value)])
 
+        metadata_filename = encodeFilename(replace_extension(filename, 'meta'))
```
dstftw: This can be moved inside `if chapters:` condition.

youtube_dl/postprocessor/ffmpeg.py
remitamine: i added the check just in case that file couldn't be created for some reason, if there is no possibility of such case then there is no need for this case.

youtube_dl/postprocessor/ffmpeg.py
remitamine: related to `os.path.isfile` check.

youtube_dl/postprocessor/ffmpeg.py
dstftw: Exception is raised if the file can't be created or opened. You'll never reach here in this case.

remitamine: it was a misunderstanding of `with` statement, `os.path.isfile` removed.

remitamine:
BTW, thanks for help by testing and reviewing.

Hrxn:
Can this affect video quality? I mean, it shouldn't, but you never know..

PR #12978, opened 2017-05-03
[Esquire] Add new extractor

robcortez:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Adding support for tv.esquire.com with a new extractor.

### DIFF
--- a/youtube_dl/extractor/esquire.py
+++ b/youtube_dl/extractor/esquire.py
@@ -1,0 +1,98 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .adobepass import AdobePassIE
+from ..utils import (
+    extract_attributes,
+    smuggle_url,
+    update_url_query,
+)
+
+
+class EsquireIE(AdobePassIE):
+    _VALID_URL = r'https?://tv\.esquire\.com/now/(?:[^/]+)/?(?:full-episode)?/(?P<title>[^/?#]+)/(?P<id>[^/?#]+)'
+    _TESTS = [{
+        'url': 'http://tv.esquire.com/now/team-ninja-warrior/full-episode/finals-week-1/631550531649',
+        'md5': '436ee8095d7179704cf2738994d36a20',
+        'info_dict': {
+            'id': '631550531649',
+            'ext': 'mp4',
+            'title': 'Finals Week 1',
+            'description': 'The first finals episode features the winners of 3 qualifying episodes.',
+            'upload_date': '20160301',
+            'timestamp': 1456808400,
+            'uploader': 'NBCU-MPAT',
+        }
+    }, {
+        'url': 'http://tv.esquire.com/now/friday-night-tykes/full-episode/if-you-wanna-show…/903098435679',
+        'md5': 'f1bf3f934ad55424d8c9333a4ab5d3aa',
+        'info_dict': {
+            'id': '903098435679',
+            'ext': 'mp4',
+            'title': 'If You Wanna Show\u2026',
+            'description': 'On the season Finale of Friday Night Tykes, two champions will be crowned.',
+            'upload_date': '20170321',
+            'timestamp': 1490068800,
+            'uploader': 'NBCU-MPAT',
+        }
+    }, {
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/esquire.py
```diff
@@ -0,0 +1,98 @@
...
+            'id': '631550531649',
+            'ext': 'mp4',
+            'title': 'Finals Week 1',
+            'description': 'The first finals episode features the winners of 3 qualifying episodes.',
+            'upload_date': '20160301',
+            'timestamp': 1456808400,
+            'uploader': 'NBCU-MPAT',
+        }
+    }, {
+        'url': 'http://tv.esquire.com/now/friday-night-tykes/full-episode/if-you-wanna-show…/903098435679',
```
remitamine: if the test case has the same flow as the previous one then there is no need for another test.

youtube_dl/extractor/esquire.py
```diff
@@ -0,0 +1,98 @@
...
+                adobe_pass.get('adobePassResourceId', 'esquire'),
+                title, video_id, player_params.get('data-episode-rating', 'TV-14'))
+            query['auth'] = self._extract_mvpd_auth(
+                url, video_id, adobe_pass.get('adobePassRequestorId', 'esquire'), resource)
+
+        info = self._search_json_ld(webpage, video_id, default={})
+        info.update({
+            '_type': 'url_transparent',
+            'url': smuggle_url(update_url_query(
+                'http://link.theplatform.com/s/%s/%s' % (account_pid, path),
```
remitamine: can be extracted directly from `data-release-url` in the webpage.

youtube_dl/extractor/esquire.py
```diff
@@ -0,0 +1,98 @@
...
+    extract_attributes,
+    smuggle_url,
+    update_url_query,
+)
+
+
+class EsquireIE(AdobePassIE):
+    _VALID_URL = r'https?://tv\.esquire\.com/now/(?:[^/]+)/?(?:full-episode)?/(?P<title>[^/?#]+)/(?P<id>[^/?#]+)'
+    _TESTS = [{
+        'url': 'http://tv.esquire.com/now/team-ninja-warrior/full-episode/finals-week-1/631550531649',
```
remitamine: as the test require authentication to pass, it should be skipped(otherwise it will fail in Travis CI).

youtube_dl/extractor/esquire.py
```diff
@@ -0,0 +1,98 @@
...
+                drupal_settings = self._parse_json(drupal_settings, video_id, fatal=False)
+                if drupal_settings:
+                    adobe_pass = drupal_settings.get('adobePass', {})
+            resource = self._get_mvpd_resource(
+                adobe_pass.get('adobePassResourceId', 'esquire'),
+                title, video_id, player_params.get('data-episode-rating', 'TV-14'))
+            query['auth'] = self._extract_mvpd_auth(
+                url, video_id, adobe_pass.get('adobePassRequestorId', 'esquire'), resource)
+
+        info = self._search_json_ld(webpage, video_id, default={})
```
remitamine: if the site does not offer `JSON-LD` data then there is no need to extract them.

youtube_dl/extractor/esquire.py
```diff
@@ -0,0 +1,98 @@
...
+                webpage, 'drupal settings', fatal=False)
+            if drupal_settings:
+                drupal_settings = self._parse_json(drupal_settings, video_id, fatal=False)
+                if drupal_settings:
+                    adobe_pass = drupal_settings.get('adobePass', {})
+            resource = self._get_mvpd_resource(
+                adobe_pass.get('adobePassResourceId', 'esquire'),
+                title, video_id, player_params.get('data-episode-rating', 'TV-14'))
+            query['auth'] = self._extract_mvpd_auth(
+                url, video_id, adobe_pass.get('adobePassRequestorId', 'esquire'), resource)
```
remitamine: `esquire` is not a valid adobe pass requestor id, the fallback value should be `style`(the value that is actually used now), the same apply to adobe pass resource id.

PR #13016, opened 2017-05-07
[DRTV] Fix issue where sign language version of video would be downloaded

RasmusRendal:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Fixing issue [13013](https://github.com/rg3/youtube-dl/issues/13013) where the sign language version of a video was downloaded, instead of the actual video.

### DIFF
--- a/youtube_dl/extractor/drtv.py
+++ b/youtube_dl/extractor/drtv.py
@@ -45,6 +45,18 @@ class DRTVIE(InfoExtractor):
             'upload_date': '20160902',
             'duration': 131.4,
         },
+    }, {
+        'url': 'https://www.dr.dk/tv/se/historien-om-danmark/-/historien-om-danmark-stenalder',
+        'md5': '7c8ca12e6c3d3e3edd59ba5a9b7ca10a',
+        'info_dict': {
+            'id': 'historien-om-danmark-stenalder',
+            'ext': 'mp4',
+            'title': 'Historien om Danmark: Stenalder (1)',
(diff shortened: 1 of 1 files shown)


dstftw:
1. Do not skip formats.
2. Using single quotes.
3. Squash commits.

RasmusRendal:
Should I just use the format flag for differing between the sign language version and the regular one? I was of the impression that the formats were for video quality.

dstftw:
What you should do is to assign lower `preference` for sign language versions.

RasmusRendal:
Alright, I did all the things now. Does this look good?

youtube_dl/extractor/drtv.py
```diff
@@ -85,7 +97,11 @@ def _real_extract(self, url):
             kind = asset.get('Kind')
             if kind == 'Image':
                 thumbnail = asset.get('Uri')
-            elif kind in ('VideoResource', 'AudioResource'):
+            preference = 0
```
dstftw: What's the point extracting it here?

youtube_dl/extractor/drtv.py
```diff
@@ -85,7 +97,11 @@ def _real_extract(self, url):
...
+
+            sign_language = asset.get('Target') == 'SignLanguage'
```
dstftw: What's the point extracting it as separate variable?

youtube_dl/extractor/drtv.py
```diff
@@ -85,7 +97,11 @@ def _real_extract(self, url):
...
+
+            if kind in ('VideoResource', 'AudioResource'):
```
dstftw: What's the point making it `if` when it should be `elif`?

youtube_dl/extractor/drtv.py
RasmusRendal: spoken_language worked like that, so I just assumed that was standard.

dstftw: Ok, it's fine to separate it in order not to calculate it every time in a loop over links, but it should be located near `spoken_subtitles`.

PR #13024, opened 2017-05-08
NonkTube Add new extractor

sarbajitsaha:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Added extractor for videos from https://www.nonktube.com
Example of video pages supported - https://www.nonktube.com/video/117836/asian-cute-girl-escaped-from-school-with-her-new-bf

Flake8 compliant 

In reference to issue https://github.com/rg3/youtube-dl/issues/8647

### DIFF
--- a/youtube_dl/extractor/nonktube.py
+++ b/youtube_dl/extractor/nonktube.py
@@ -2,12 +2,11 @@
 from __future__ import unicode_literals
 
 from .common import InfoExtractor
-
 import re
 
 
 class NonkTubeIE(InfoExtractor):
-    _VALID_URL = r'https?:\/\/(?:www\.)?nonktube\.com\/video\/(?P<id>[0-9]+)\/[0-9a-zA-Z\-]*'
+    _VALID_URL = r'https?://(?:www\.)?nonktube\.com\/video/(?P<id>[0-9]+)/[0-9a-zA-Z\-]*'
     _TESTS = [{
         'url': 'https://www.nonktube.com/video/118636/sensual-wife-uncensored-fucked-in-hairy-pussy-and-facialized',
         'md5': 'ccd2df2aef9f44e51773c8ea59de6fd1',
@@ -16,6 +15,8 @@ class NonkTubeIE(InfoExtractor):
             'ext': 'mp4',
             'title': 'Sensual Wife Uncensored Fucked In Hairy Pussy And Facialized',
             'thumbnail': 'https://www.nonktube.com/media/videos/tmb_2/118636/default.jpg',
+            'uploader': 'https://www.nonktube.com/user/cumdrinkingwife',
+            'age_limit': 18
         }
     }, {
         'url': 'https://www.nonktube.com/video/118698/2-japanese-girls-sharing-one-dick-uncesnored',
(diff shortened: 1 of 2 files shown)


dstftw:
Bother to read

and fix all issues.

sarbajitsaha:
I was wondering if anything else is needed. I went through those two URLs.

I have added two tests, the code is Flake8 compliant and seems to work with both Python2 and 3.

I am returning id, title(two options to extract this), uploader(None if unable to extract otherwise the link to the user profile), thumbnail link, video URL(only format is mp4) and age limit is 18 as its an adult site. Do I need to add anything more?

youtube_dl/extractor/nonktube.py
```diff
@@ -0,0 +1,68 @@
...
+            'id': '118698',
+            'ext': 'mp4',
+            'title': '2 Japanese Girls Sharing One Dick Uncesnored',
+            'thumbnail': 'https://www.nonktube.com/media/videos/tmb_2/118698/default.jpg',
+            'uploader': 'https://www.nonktube.com/user/cumdrinkingwife',
+            'age_limit': 18
+        }
+    }]
+    DOWNLOAD_URL_TEMPLATE = 'https://cdn.nonktube.com/%s.mp4'
+    UPLOADER_URL_TEMPLATE = 'https://www.nonktube.com/user/%s'
```
dstftw: No need in these variables.

youtube_dl/extractor/nonktube.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, video_id)
+
+        video_url = self.DOWNLOAD_URL_TEMPLATE % str(video_id)
```
dstftw: `video_id` is already a string.

youtube_dl/extractor/nonktube.py
```diff
@@ -0,0 +1,68 @@
...
+
+        title = self._html_search_regex(
+            [r'<title>(?P<title>.+?) - NonkTube.com</title>',
+             r'<h1 class=.*>(?P<title>.+?)</h1>'], webpage, 'title', group='title')
```
dstftw: No need in explicit group name.

youtube_dl/extractor/nonktube.py
```diff
@@ -0,0 +1,68 @@
...
+        'url': 'https://www.nonktube.com/video/118698/2-japanese-girls-sharing-one-dick-uncesnored',
+        'md5': 'bb4bdd0a61f591e2f5f7429dc860327e',
+        'info_dict': {
+            'id': '118698',
+            'ext': 'mp4',
+            'title': '2 Japanese Girls Sharing One Dick Uncesnored',
+            'thumbnail': 'https://www.nonktube.com/media/videos/tmb_2/118698/default.jpg',
+            'uploader': 'https://www.nonktube.com/user/cumdrinkingwife',
+            'age_limit': 18
+        }
```
dstftw: Remove all duplicated tests.

youtube_dl/extractor/nonktube.py
```diff
@@ -0,0 +1,68 @@
...
+        thumbnail = self._html_search_regex(
+            [r'<img src="(?P<thumbnail>.+?)" title="' + re.escape(title) + r'" alt="' + re.escape(title) + r'" width="[0-9]*" height="[0-9]*" />'],
+            webpage, 'thumbnail', fatal=False, group='thumbnail')
+
+        # only one link of format /user/ is present in every video which gives the uploader username
+        uploader = self._html_search_regex(
+            [r'<a href="/user/(?P<uploader_id>[0-9A-Za-z]+?)">'],
+            webpage, 'uploader', fatal=False, group='uploader_id')
+        if uploader is not None:
+            uploader = self.UPLOADER_URL_TEMPLATE % uploader
```
dstftw: `uploader_url`.

youtube_dl/extractor/nonktube.py
```diff
@@ -0,0 +1,68 @@
...
+        'info_dict': {
+            'id': '118698',
+            'ext': 'mp4',
+            'title': '2 Japanese Girls Sharing One Dick Uncesnored',
+            'thumbnail': 'https://www.nonktube.com/media/videos/tmb_2/118698/default.jpg',
+            'uploader': 'https://www.nonktube.com/user/cumdrinkingwife',
+            'age_limit': 18
+        }
+    }]
+    DOWNLOAD_URL_TEMPLATE = 'https://cdn.nonktube.com/%s.mp4'
```
dstftw: No hardcodes. Extraction should be implemented in terms of `NuevoBaseIE` instead.

sarbajitsaha: Does this class return a list of formats with urls in which the video can be downloaded? It seems that for this website, all the videos are in mp4 format and I can't even control the video resolution, so every video will be of that URL format only. So can I use this in that case?

```python
 video_url = 'https://cdn.nonktube.com/%s.mp4' % video_id
```

dstftw: You should not touch extraction at all. It's already implemented in `NuevoBaseIE`.

sarbajitsaha: Ok. I checked out the extractor for lovehomeporn which seems to be similar

```python
 info = self._extract_nuevo(
            'http://lovehomeporn.com/media/nuevo/config.php?key=%s' % video_id,
            video_id)
```

I found a similar url in the website - https://www.nonktube.com/media/nuevo/config.php?key=118636. However, unlike the former which showed the xml when I loaded the link in the browser, here it shows **"restricted area"** I tried passing the original video URL to extract_nuevo function which also didn't work. In both cases, it raised error.

dstftw: See how browser calls it.

sarbajitsaha: As soon as I press the play button, in the network tab(Chrome), I get this request 

Request URL:https://cdn.nonktube.com/117475.mp4?start=0

which seems to be the original video file link. That's why I structured it that way earlier. Can't seem to get any other format than mp4, nor can I change the resolution. I

PR #13032, opened 2017-05-09
[Beam:Vod] Add new extractor

mikf:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This is a new extractor for VODs from the https://beam.pro/ streaming service.

### DIFF
--- a/youtube_dl/extractor/beampro.py
+++ b/youtube_dl/extractor/beampro.py
@@ -6,16 +6,30 @@ from ..utils import (
     ExtractorError,
     clean_html,
     compat_str,
+    float_or_none,
     int_or_none,
     parse_iso8601,
     try_get,
+    urljoin,
 )
 
 
-class BeamProLiveIE(InfoExtractor):
+class BeamProBaseIE(InfoExtractor):
+    _RATINGS = {'family': 0, 'teen': 13, '18+': 18}
+
+    def _extract_channel_info(self, chan):
+        user_id = chan.get('userId') or try_get(chan, lambda x: x['user']['id'])
+        return {
+            'uploader': chan.get('token') or try_get(
+                chan, lambda x: x['user']['username'], compat_str),
+            'uploader_id': compat_str(user_id) if user_id else None,
+            'age_limit': self._RATINGS.get(chan.get('audience')),
+        }
+
+
+class BeamProLiveIE(BeamProBaseIE):
     IE_NAME = 'Beam:live'
     _VALID_URL = r'https?://(?:\w+\.)?beam\.pro/(?P<id>[^/?#&]+)'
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/beampro.py
```diff
@@ -71,3 +73,93 @@ def _real_extract(self, url):
...
+            'duration': float_or_none(vod_info.get('duration')),
+            'thumbnail': thumbnail_url,
+            'timestamp': parse_iso8601(vod_info.get('createdAt')),
+            'uploader': chan.get('token') or try_get(
+                chan, lambda x: x['user']['username'], compat_str),
+            'uploader_id': compat_str(user_id) if user_id else None,
+            'age_limit': BeamProLiveIE._RATINGS.get(chan.get('audience')),
+            'view_count': int_or_none(vod_info.get('viewsTotal')),
+            'formats': formats,
+        }
```
dstftw: 151-165 code duplication.

youtube_dl/extractor/beampro.py
```diff
@@ -6,15 +6,17 @@
...
     parse_iso8601,
     try_get,
+    urljoin,
 )
 
 
 class BeamProLiveIE(InfoExtractor):
     IE_NAME = 'Beam:live'
-    _VALID_URL = r'https?://(?:\w+\.)?beam\.pro/(?P<id>[^/?#&]+)'
+    _VALID_URL = r'https?://(?:\w+\.)?beam\.pro/(?P<id>[^/?#&]+)$'
```
dstftw: Should match `http://www.beam.pro/niterhayven/` and `http://www.beam.pro/niterhayven?abc=123`.

youtube_dl/extractor/beampro.py
mikf: I'm assuming you mean lines 151-165 contain duplicate code from the BeamProLiveIE class.
Is it OK if I just create a common base class and put all the duplicate code in there?

youtube_dl/extractor/beampro.py
mikf: The issue here is that `http://www.beam.pro/niterhayven?vod=123`, which is a VOD URL, would also get matched by this. I tried to fix this in ccc035c1fb2b6080b4dab9e2cb0b8273f734c1cc by reordering the import of those extractors, but I don't know if this acceptable.

dstftw: Override `suitable`.

mikf:
The requested changes should have all been addressed.
Are there any other issues left or is there anything else that I should do?

On another note: Beam VODs are only available for 2 weeks after creation, so any test is going to fail after that time. Is there a preferred way of handling cases like that?

PR #13053, opened 2017-05-11
[Aljazeera] improve _VALID_URL

steebchen:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Add a new type of link to _VALID_URL with a test. I was not sure whether all those info_dict fields are needed for the second test, but just added them.

### DIFF
--- a/youtube_dl/extractor/bandcamp.py
+++ b/youtube_dl/extractor/bandcamp.py
@@ -47,6 +47,7 @@ class BandcampIE(InfoExtractor):
         mobj = re.match(self._VALID_URL, url)
         title = mobj.group('title')
         webpage = self._download_webpage(url, title)
+        thumbnail = self._html_search_meta('og:image', webpage, default=None)
         m_download = re.search(r'freeDownloadPage: "(.*?)"', webpage)
         if not m_download:
             m_trackinfo = re.search(r'trackinfo: (.+),\s*?\n', webpage)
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/aljazeera.py
```diff
@@ -4,9 +4,9 @@
 
 
 class AlJazeeraIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?aljazeera\.com/programmes/.*?/(?P<id>[^/]+)\.html'
+    _VALID_URL = r'https?://(?:www\.)?aljazeera\.com/(programmes|video)/.*?/(?P<id>[^/]+)\.html'
```
dstftw: Don't capture.

youtube_dl/extractor/aljazeera.py
```diff
@@ -19,7 +19,20 @@ class AlJazeeraIE(InfoExtractor):
...
+            'ext': 'mp4',
+            'title': 'Sierra Leone: 709-carat diamond to be auctioned off',
+            'description': 'A 709-carat diamond is being auctioned in Sierra Leone\'s capital Freetown.',
+            'uploader_id': '665003303001',
+            'timestamp': 1494496255,
+            'upload_date': '20170511',
+        },
+        'add_ie': ['BrightcoveNew'],
+        'skip': 'Not accessible from Travis CI server',
+    }]
```
dstftw: `only_matching`.

dstftw:
Squash commits.

steebchen:
Fixed the changes & squashed it – however, I am still trying to squash this damn b89b54e commit.. Apart from this, did I do everything correctly? And, just out of curiosity, why can't you just squash all those commits when merging the PR?

Issue #8944, opened 2016-03-23
ExtractorError(u'Failed to download MPD manifest: HTTP Error 403: Forbidden (caused by HTTPError());


abderrahmane-tj:
Love your work <3

error suggested i report this, i am not sure if this is the right place to do so but here we go :

Running Windows 8.1 64
using cmder

```
λ youtube-dl.exe -F https://www.youtube.com/watch?v=ZxRMAW4C9iw --verbose
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-F', u'https://www.youtube.com/watch?v=ZxRMAW4C9iw', u'--verbose']
[debug] Encodings: locale cp1252, fs mbcs, out cp437, pref cp1252
[debug] youtube-dl version 2016.03.18
[debug] Python version 2.7.10 - Windows-8-6.2.9200
[debug] exe versions: none
[debug] Proxy map: {}
[youtube] ZxRMAW4C9iw: Downloading webpage
[youtube] ZxRMAW4C9iw: Downloading video info webpage
[youtube] ZxRMAW4C9iw: Extracting video information
[youtube] {22} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {43} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {18} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {5} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {36} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {17} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {137} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {299} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {136} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {298} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {135} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {134} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {133} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {160} signature length 40.43, html5 player en_US-vflpmpoFG
[youtube] {140} signature length 40.43, html5 player en_US-vflpmpoFG
[...]

PR #13156, opened 2017-05-19
[youtube] Fixes issue with not being able to decode signatures on DASH MPD URLs (#8944)

Protuhj:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Fixes Issue #8944 - ExtractorError(u'Failed to download MPD manifest: HTTP Error 403: Forbidden (caused by HTTPError());  

The get_video_info API endpoint requires the 'sts' parameter be included, otherwise the decoding of the signature will fail.

Adding this parameter doesn't affect other videos without encoded signatures. 

If the sts regex doesn't match, then the code will fail like it currently does.

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -1435,9 +1435,11 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
         else:
             age_gate = False
             video_info = None
+            sts = ''
             # Try looking directly into the video webpage
             ytplayer_config = self._get_ytplayer_config(video_id, video_webpage)
             if ytplayer_config:
+                sts = ytplayer_config['sts']
                 args = ytplayer_config['args']
                 if args.get('url_encoded_fmt_stream_map'):
                     # Convert to the same format returned by compat_parse_qs
@@ -1460,7 +1462,6 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
                 # video with such 'manifest behavior' see https://github.com/rg3/youtube-dl/issues/6093)
 
                 # up-to-date sts value is required to properly decode the signature
-                sts = self._search_regex(r'"sts"\s*:\s*(\d+)', video_webpage, 'sts', default='')
                 self.report_video_info_webpage_download(video_id)
                 for el_type in ['&el=info', '&el=embedded', '&el=detailpage', '&el=vevo', '']:
                     video_info_url = (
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/youtube.py
```diff
@@ -1458,11 +1458,14 @@ def add_dash_mpd(video_info):
                 # manifest pointed by get_video_info's dashmpd).
                 # The general idea is to take a union of itags of both DASH manifests (for example
                 # video with such 'manifest behavior' see https://github.com/rg3/youtube-dl/issues/6093)
+
+                # up-to-date sts value is required to properly decode the signature
+                sts = self._search_regex(r'"sts"\s*:\s*(\d+)', video_webpage, 'sts', default='')
```
dstftw: Should be extracted from `ytplayer_config`.

Protuhj: I see that now, I'll modify it when I'm at home later.

Protuhj: Updated the code, see if that is good enough for ya,

youtube_dl/extractor/youtube.py
```diff
@@ -1435,9 +1435,11 @@ def add_dash_mpd(video_info):
         else:
             age_gate = False
             video_info = None
+            sts = ''
             # Try looking directly into the video webpage
             ytplayer_config = self._get_ytplayer_config(video_id, video_webpage)
             if ytplayer_config:
+                sts = ytplayer_config['sts']
```
dstftw: It should not break extraction if `sts` key is missing.

Protuhj: I'm not a python dev, so I assume using the 'get' method of the ytplayer_config dictionary is correct.

youtube_dl/extractor/youtube.py
```diff
@@ -1458,11 +1460,13 @@ def add_dash_mpd(video_info):
                 # manifest pointed by get_video_info's dashmpd).
                 # The general idea is to take a union of itags of both DASH manifests (for example
                 # video with such 'manifest behavior' see https://github.com/rg3/youtube-dl/issues/6093)
+
+                # up-to-date sts value is required to properly decode the signature
```
dstftw: This comment is pointless at this place.

dstftw:

Protuhj:
No thanks, just sharing some useful knowledge; thanks for asking though!

PR #13174, opened 2017-05-20
[Stream.cz] add subtitle support

oskar456:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This adds subtitle support for Stream.cz. Also tests are fixed (they probably somehow changed the video files).

### DIFF
--- a/youtube_dl/extractor/streamcz.py
+++ b/youtube_dl/extractor/streamcz.py
@@ -85,6 +85,14 @@ class StreamCZIE(InfoExtractor):
         else:
             title = data['name']
 
+        subtitles = {}
+        srt_url = data.get('subtitles_srt')
+        if srt_url:
+            subtitles['cs'] = [{
+                'ext': 'srt',
+                'url': srt_url,
+            }]
+
         return {
             'id': video_id,
             'title': title,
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/streamcz.py
```diff
@@ -93,4 +93,15 @@ def _real_extract(self, url):
             'description': data.get('web_site_text'),
             'duration': int_or_none(data.get('duration')),
             'view_count': int_or_none(data.get('views')),
+            'subtitles': self.extract_subtitles(data.get('subtitles_srt')),
         }
+
+    @staticmethod
+    def _get_subtitles(subs):
```
dstftw: What's the point of this?

oskar456: It's the method called by `self.extract_subtitles()` common method when subtitles are actually requested by user.

In this case, it's probably unnecessary.

dstftw: This methods is only used when subtitles' extraction depends on additional expensive work, i.e. network I/O. In this case subtitles are already extracted along with formats and using this method does not make any sense.

oskar456: Thanks for guidance. I've reworked it so it does the subtitle extracting along the main `_real_extract` method.

PR #13178, opened 2017-05-21
[wdr] add support for wdrmaus.de content pages as playlists

rohieb:
The wdrmaus.de content pages directly contain multiple video players without linking to a separate page for each video, like the WDR Mediathek pages do. Therefore, we cannot delegate loading the info_dict from a "video URL" (because it does not exist), but instead we have to grab each video separately from their video player <divs> and load the info_dict from the respective JSONP.

On the other hand, Mediathek pages only contain links to the separate video pages, and no JSONP URL, so we still need to support the old way of loading each video page separately when playing the playlist.

This pull-request implements both ways of loading playlists, and also fixes some outdated tests which fail due to HTTP 404 errors.

### DIFF
--- a/youtube_dl/extractor/wdr.py
+++ b/youtube_dl/extractor/wdr.py
@@ -105,9 +105,9 @@ class WDRBaseIE(InfoExtractor):
 
 
 class WDRIE(WDRBaseIE):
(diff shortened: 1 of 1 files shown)


rohieb:
Overlooked the case where the page is not a playlist, but a single video. Force-pushed.

rohieb:
Tests keep failing in other extractors… Any other feedback on that? :)

PR #13292, opened 2017-06-05
[streamango] fix embed links by making title optional

steebchen:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Use empty title as default to fix embed links where no og:title is set. Tried fatal=False and default=None but these two throwed another error.

### DIFF
--- a/youtube_dl/extractor/streamango.py
+++ b/youtube_dl/extractor/streamango.py
@@ -22,6 +22,13 @@ class StreamangoIE(InfoExtractor):
             'title': '20170315_150006.mp4',
         }
     }, {
+        'url': 'https://streamango.com/embed/foqebrpftarclpob/asdf_asd_2_mp4',
+        'info_dict': {
+            'id': 'foqebrpftarclpob',
+            'ext': 'mp4',
+            'title': 'foqebrpftarclpob',
+        }
+    }, {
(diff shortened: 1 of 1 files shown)


dstftw:
1. Add a test.
2. Use video_id.

steebchen:
1. A test with only `info_dict.title: ''` or do I need to add an md5 and some other stuff?
2. Use video_id where exactly?

dstftw:
1. A test should test what you are trying to fix.
2. Where you use `''`, where else?

steebchen:
1. I know that. I was asking whether I can only use info_dict with title and without md5 and other stuff.
2. Ah, of course. Brain had stopped working.

youtube_dl/extractor/streamango.py
```diff
@@ -22,6 +22,13 @@ class StreamangoIE(InfoExtractor):
             'title': '20170315_150006.mp4',
         }
     }, {
+        'url': 'https://streamango.com/embed/ptcbfodbpadmnarm/Koty_i_zaloty_0001',
```
dstftw: This video has `og:title`.

steebchen: Sorry. Fixed

PR #13315, opened 2017-06-08
[picosong] Add new extractor

magcius:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Added an extractor for a popular music sharing website, picosong. A team of a project I work on often uses this site to share mp3 files and I wanted to automate a large part of the work we currently do manually. YouTube-DL seemed like a good fit for the job.

### DIFF
--- a/youtube_dl/extractor/picosong.py
+++ b/youtube_dl/extractor/picosong.py
@@ -1,0 +1,78 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from ..utils import urljoin, parse_duration, parse_filesize
+
+
+class PicosongIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?picosong\.com/(?P<id>[0-9A-Za-z]{4})'
+    _TESTS = [{
+        'url': 'http://picosong.com/7uer/',
+        'md5': '7f2ee011ba3ca2ac180a35deb10feabf',
+        'info_dict': {
+            'id': '7uer',
+            'ext': 'mp3',
+            'artist': 'Xarlable',
+            'title': 'Rock My Forum',
+            'album': 'Epic',
+            'genre': 'Howto & Style',
+            'duration': 187.0,
+            'filesize_approx': 4700000,
+            'asr': 44100,
+            'abr': 210,
+        },
+    }, {
+        'url': 'http://picosong.com/i3gs/',
+        'md5': 'd4ec5d64d97e28b45d2710c096c29f70',
+        'info_dict': {
+            'id': 'i3gs',
+            'ext': 'mp3',
+            'title': 'Rick.mp3',
+            'duration': 26.0,
+            'filesize_approx': 1020000,
+            'asr': 48000,
+            'abr': 320,
+        },
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        page = self._download_webpage(url, video_id)
+        download_path = self._search_regex(
+            r'value="/static/swf/picoplayer\.swf\?file=([^"]+)"',
+            page, 'url')
+        download_url = urljoin(url, download_path)
+
+        def extract_info_td(key, fatal=True):
+            return self._html_search_regex(
+                r'%s:</td>\s*<td class="value">([^<]*)</td>' % (key, ),
+                page, key, fatal=fatal)
+
+        title = extract_info_td('Title')
+        if not title:
+            title = extract_info_td('Filename')
+
+        def parse_digits(digits, key, fatal=True):
+            if digits is None:
+                return None
+            m = self._search_regex(r'(^[0-9]+)', digits, key, fatal=fatal)
+            if m is None:
+                return None
+            return int(m)
+
+        return {
+            'id': video_id,
+            'title': title,
+            'artist': extract_info_td('Artist', fatal=False),
+            'album': extract_info_td('Album', fatal=False),
+            'genre': extract_info_td('Genre', fatal=False),
+            'duration': parse_duration(extract_info_td('Length', fatal=False)),
+            'filesize_approx': parse_filesize(extract_info_td('File size', fatal=False)),
+            'abr': parse_digits(extract_info_td('Bitrate', fatal=False), 'Bitrate', fatal=False),
+            'asr': parse_digits(extract_info_td('Sample Rate', fatal=False), 'Sample Rate', fatal=False),
+            'url': download_url,
+        }
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -760,6 +760,7 @@ from .periscope import (
 from .philharmoniedeparis import PhilharmonieDeParisIE
 from .phoenix import PhoenixIE
 from .photobucket import PhotobucketIE
+from .picosong import PicosongIE
 from .piksel import PikselIE
 from .pinkbike import PinkbikeIE
 from .pladform import PladformIE


youtube_dl/extractor/picosong.py
```diff
@@ -0,0 +1,47 @@
...
+            'id': '7uer',
+            'ext': 'mp3',
+            'artist': 'Xarlable',
+            'title': 'Rock My Forum',
+            'album': 'Epic',
+            'genre': 'Howto & Style',
+        }
+    }
+
+    def _extract_info_td(self, page, key):
```
dstftw: Change to nested function. Capture `page` from outside code instead of passing as argument.

youtube_dl/extractor/picosong.py
```diff
@@ -0,0 +1,47 @@
...
+        # <td class="key">Title:</td>
+        # <td class="value">Rock My Forum</td>
```
dstftw: Remove pointless comments.

magcius: The goal is to show an example of the input we should be seeing, but OK.

magcius:
Fixed review feedback

youtube_dl/extractor/picosong.py
```diff
@@ -0,0 +1,47 @@
...
+        return self._html_search_regex(
+            r'%s:</td>\s*<td class="value">([^<]+)' % (key.capitalize(), ),
+            page, key)
```
dstftw: This will break extraction if some field is missing. Optional fields should not break extraction.

magcius: If a field is missing on the website, then it will have `<td class="value"></td>`. See e.g. http://picosong.com/i3gs/

I've never seen otherwise.

dstftw: And `_html_search_regex` will fail breaking the extraction completely.

magcius:
Updated for fixes, including test case.

dstftw:
Nothing changed. Extraction is still broken if optional field is missing.

magcius:
What? I rewrote the regex and added a test case including optional fields.

dstftw:
Bother to read [youtube-dl coding conventions](https://github.com/rg3/youtube-dl#youtube-dl-coding-conventions).

magcius:
Are you suggesting that instead of putting an empty string in these fields, they should not exist? I don't see that written in the guide, unless I missed something.

I could pass `fatal=False` to these extraction bits, but I consider it a very bad bug if the placeholder field doesn't exist in the HTML.

Or are you suggesting that `filename` might be missing as well? I haven't seen any evidence of a missing filename field.

dstftw:
Do you even read what is addressed to you? Read coding conventions where **everything is explained**.

magcius:
I thought the document was a bit unclear, but I went and added `fatal` parameter for optional fields.

youtube_dl/extractor/picosong.py
```diff
@@ -0,0 +1,57 @@
...
+        page = self._download_webpage(url, video_id)
+        # <param name="movie" value="/static/swf/picoplayer.swf?file=/cdn/085a4492f79c0c487bb656afa902213e.mp3"/>
+        download_path = self._search_regex(
+            r'value="/static/swf/picoplayer.swf\?file=([^"]+)"',
+            page, 'url')
+        download_url = urljoin(url, download_path)
+
+        title = extract_info_td('title')
+        if not title:
+            title = extract_info_td('filename')
```
dstftw: This is never reachable.

youtube_dl/extractor/picosong.py
```diff
@@ -0,0 +1,57 @@
...
+            'genre': 'Howto & Style',
+        },
+    }, {
+        'url': 'http://picosong.com/i3gs/',
+        'md5': 'd4ec5d64d97e28b45d2710c096c29f70',
+        'info_dict': {
+            'id': 'i3gs',
+            'ext': 'mp3',
+            'title': 'Rick.mp3',
+        },
```
dstftw: Remove all duplicate tests.

youtube_dl/extractor/picosong.py
```diff
@@ -0,0 +1,57 @@
...
+            return self._html_search_regex(
+                r'%s:</td>\s*<td class="value">([^<]*)</td>' % (key.capitalize(), ),
+                page, key, fatal=fatal)
+
+        video_id = self._match_id(url)
+
+        page = self._download_webpage(url, video_id)
+        # <param name="movie" value="/static/swf/picoplayer.swf?file=/cdn/085a4492f79c0c487bb656afa902213e.mp3"/>
+        download_path = self._search_regex(
+            r'value="/static/swf/picoplayer.swf\?file=([^"]+)"',
```
dstftw: Dots should be escaped. Query may contain more arguments that will be incorrectly captured by this regex as path.

youtube_dl/extractor/picosong.py
```diff
@@ -0,0 +1,57 @@
...
+    def _real_extract(self, url):
+        def extract_info_td(key, fatal=True):
+            return self._html_search_regex(
+                r'%s:</td>\s*<td class="value">([^<]*)</td>' % (key.capitalize(), ),
+                page, key, fatal=fatal)
+
+        video_id = self._match_id(url)
+
+        page = self._download_webpage(url, video_id)
+        # <param name="movie" value="/static/swf/picoplayer.swf?file=/cdn/085a4492f79c0c487bb656afa902213e.mp3"/>
```
dstftw: Remove.

youtube_dl/extractor/picosong.py
```diff
@@ -0,0 +1,57 @@
...
+        'md5': 'd4ec5d64d97e28b45d2710c096c29f70',
+        'info_dict': {
+            'id': 'i3gs',
+            'ext': 'mp3',
+            'title': 'Rick.mp3',
+        },
+    }]
+
+    def _real_extract(self, url):
+        def extract_info_td(key, fatal=True):
```
dstftw: Move to the place where it's first used.

youtube_dl/extractor/picosong.py
```diff
@@ -0,0 +1,57 @@
...
+        title = extract_info_td('title')
+        if not title:
+            title = extract_info_td('filename')
+
+        return {
+            'id': video_id,
+            'title': title,
+            'artist': extract_info_td('artist', fatal=False),
+            'album': extract_info_td('album', fatal=False),
+            'genre': extract_info_td('genre', fatal=False),
```
dstftw: There are more metadata to be extracted: `duration`, `abr`, `asr`, `filesize_approx`.

youtube_dl/extractor/picosong.py
magcius: Hm? This tests missing fields.

youtube_dl/extractor/picosong.py
magcius: This is reachable in case we see a blank title, no?

magcius: Confirmed that this is reachable in the case of the second test, Rick.mp3

dstftw: This is not reachable when title extraction fails.

magcius: This handles the case when title extraction succeeds but results in an empty string, like in http://picosong.com/i3gs/

magcius:
Updated for the rest of the feedback.

magcius:
Any updates here? I've fixed the feedback you've mentioned.

youtube_dl/extractor/picosong.py
dstftw: Not fixed.

youtube_dl/extractor/picosong.py
dstftw: Not fixed.

youtube_dl/extractor/picosong.py
dstftw: So what? It's still not reachable when 46 fails. Extracting empty strings does not make any sense.

youtube_dl/extractor/picosong.py
```diff
@@ -0,0 +1,78 @@
...
+
+        return {
+            'id': video_id,
+            'title': title,
+            'artist': extract_info_td('Artist', fatal=False),
+            'album': extract_info_td('Album', fatal=False),
+            'genre': extract_info_td('Genre', fatal=False),
+            'duration': parse_duration(extract_info_td('Length', fatal=False)),
+            'filesize_approx': parse_filesize(extract_info_td('File size', fatal=False)),
+            'abr': parse_digits(extract_info_td('Bitrate', fatal=False), 'Bitrate', fatal=False),
```
dstftw: `utils.int_or_none`.

youtube_dl/extractor/picosong.py
```diff
@@ -0,0 +1,78 @@
...
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        page = self._download_webpage(url, video_id)
+        download_path = self._search_regex(
+            r'value="/static/swf/picoplayer\.swf\?file=([^"]+)"',
+            page, 'url')
+        download_url = urljoin(url, download_path)
+
+        def extract_info_td(key, fatal=True):
```
dstftw: It's more rational to use `False` as default since it's used more often in this code.

youtube_dl/extractor/picosong.py
magcius: The data I have is ```
<td class="key">Title:</td>
<td class="value"></td>```

How should I extract this file, other than falling back to the filename if no title is provided?

dstftw: You are not falling back. If regex does not match HTML you will fail immediately at 46.
```
<td class="key">Title:</td>
<td class="value"></td>
```
may not be present in HTML code and you will never reach your "fallback" in this case.

magcius: This code is reachable. The second testcase you keep telling me to remove reaches this code. I tried removing this code and the testcase failed, so I guarantee you it is reachable.

The regex `Title:</td>\s*<td class="value">([^<]*)</td>` matches the block above, but matches an empty string ("") and returns that -- which doesn't appear to be a fatal error case.

dstftw: Do you even read my post?

magcius: The one where you keep on saying "the code is unreachable" when I'm telling you it is reachable, because I ran it, and the testcase reaches it?

I get the sense we're talking past each other. I admit I might not be understanding you here. Could you try running my extractor on http://picosong.com/i3gs/ and double-checking that that code is reachable? It is reachable for me.

dstftw: I'm **not talking about any particular test case**. I'm talking about a potential situation when this code:
```
<td class="key">Title:</td>
<td class="value"></td>
```
**is not present on the webpage**. 46 will **fail** in this case because it has `fatal=True` and you will not reach a "fallback" in this case.

magcius: Oh, OK. Sorry, I misunderstood you. That's not what the filename fallback is about. There are three possible scenarios here:

1. Title is present:
```
<td class="key">Title:</td>
<td class="value">Rock My Forum</td>
```

This is the case of the first testcase. In this case, we want to use "Rock My Forum" as the title here.

2. The title field is seen but blank:
```
<td class="key">Title:</td>
<td class="value"></td>
```

In this case, I chose to fall back to the filename. This is the case of the second testcase ("Rick.mp3") and is what this code is about.

3. The fields are not on the web page at all. In this case, this likely means that picosong.com is down or broken, so this is a fatal error, which should work as intended.

This is what my code is trying to do. Does this help clear this code up?

dstftw: 3 is easily possible, say `Title` is renamed to `Song`. With proper fallback extractor continues to work. Without it - fails completely. Read coding conventions where the rationale clearly explained.

Issue #12058, opened 2017-02-10
[drtuber] Change in video config location


einstein95:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like that [x])
- Use *Preview* tab to see how your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2017.02.10*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2017.02.10**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

```
$ youtube-dl http://www.drtuber.com/video/3327619/stepbros-decided-to-double-team-amber --verbose
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['http://www.drtuber.com/video/3327619/stepbros-decided-to-double-team-amber', '--verbose']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2017.02.10
[debug] Python version 3.6.0 - Linux-4.8.13-1-ARCH-x86_64-with-arch
[debug] exe versions: ffmpeg N-83248-gb4a13d442a, ffprobe N-83248-gb4a13d442a, rtmpdump 2.4
[debug] Proxy map: {}
[DrTuber] stepbros-decided-to-double-team-amber: Downloading webpage
[...]

PR #13322, opened 2017-06-08
[DrTuber] Fetch video URL from JSON stream (fixes issue #12058)

parmjitv:

---

- [x ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Updated the [DrTuber.com] extractor to fetch the video URL from the appropriate JSON stream.  Much thanks to einstein95 for his guidance.

Cheers,

Parmjit V.

### DIFF
--- a/youtube_dl/extractor/drtuber.py
+++ b/youtube_dl/extractor/drtuber.py
@@ -44,20 +44,21 @@ class DrTuberIE(InfoExtractor):
         webpage = self._download_webpage(
             'http://www.drtuber.com/video/%s' % video_id, display_id)
 
-        video_data = self._download_json('http://www.drtuber.com/player_config_json/', video_id, query={
-            'vid': video_id,
-            'embed': 0,
-            'aid': 0,
-            'domain_id': 0,
-        })
+        video_data = self._download_json(
+            'http://www.drtuber.com/player_config_json/', video_id, query={
+                'vid': video_id,
+                'embed': 0,
+                'aid': 0,
+                'domain_id': 0,
+            })
 
         formats = []
-        for video_url_key, video_url_value in video_data['files'].items():
-            if video_url_value:
+        for format_id, video_url in video_data['files'].items():
+            if video_url:
                 formats.append({
-                    'format_id': video_url_key.upper(),
-                    'quality': 2 if video_url_key == 'hq' else 1,
-                    'url': video_url_value
+                    'format_id': format_id,
+                    'quality': 2 if format_id == 'hq' else 1,
+                    'url': video_url
                 })
         self._sort_formats(formats)
 
--- a/youtube_dl/extractor/drtuber.py
+++ b/youtube_dl/extractor/drtuber.py
@@ -52,13 +52,15 @@ class DrTuberIE(InfoExtractor):
         })
 
         formats = []
-        for video_url_key, video_url_value in video_data['files'].items():
-            if video_url_value:
-                formats.append({
-                    'format_id': video_url_key.upper(),
-                    'quality': 2 if video_url_key == 'hq' else 1,
-                    'url': video_url_value
-                })
+        for video_key, video_value in video_data.items():
+            if video_key == 'files':
+                for video_url_key, video_url_value in video_data[video_key].items():
+                    if video_url_value:
+                        formats.append({
+                            'format_id': video_url_key.upper(),
+                            'quality': 2 if video_url_key == 'hq' else 1,
+                            'url': video_url_value
+                        })
         self._check_formats(formats, video_id)
         self._sort_formats(formats)
 
--- a/youtube_dl/extractor/drtuber.py
+++ b/youtube_dl/extractor/drtuber.py
@@ -6,7 +6,6 @@ from .common import InfoExtractor
 from ..utils import (
     NO_DEFAULT,
     str_to_int,
-    ExtractorError,
 )
 
 
@@ -50,20 +49,16 @@ class DrTuberIE(InfoExtractor):
             'embed': 0,
             'aid': 0,
             'domain_id': 0,
-        }, fatal=False)
+        }, fatal=True)
 
         formats = []
-        if video_data:
-            for video_url_key in video_data.get('files', {}):
-                # high-quality video format is preferred, if available
-                if video_data['files'][video_url_key]:
-                    formats.append({
-                        'format_id': video_url_key.upper(),
-                        'quality': 2 if video_url_key == 'hq' else 1,
-                        'url': video_data['files'][video_url_key]
-                    })
-        if not formats:
-            raise ExtractorError('Video %s is not available' % video_id, expected=True)
+        for video_url_key in video_data.get('files', {}):
+            if video_data['files'][video_url_key]:
+                formats.append({
+                    'format_id': video_url_key.upper(),
+                    'quality': 2 if video_url_key == 'hq' else 1,
+                    'url': video_data['files'][video_url_key]
+                })
         self._check_formats(formats, video_id)
         self._sort_formats(formats)
 
--- a/youtube_dl/extractor/drtuber.py
+++ b/youtube_dl/extractor/drtuber.py
@@ -52,15 +52,13 @@ class DrTuberIE(InfoExtractor):
         })
 
         formats = []
-        for video_key, video_value in video_data.items():
-            if video_key == 'files':
-                for video_url_key, video_url_value in video_data[video_key].items():
-                    if video_url_value:
-                        formats.append({
-                            'format_id': video_url_key.upper(),
-                            'quality': 2 if video_url_key == 'hq' else 1,
-                            'url': video_url_value
-                        })
+        for video_url_key, video_url_value in video_data['files'].items():
+            if video_url_value:
+                formats.append({
+                    'format_id': video_url_key.upper(),
+                    'quality': 2 if video_url_key == 'hq' else 1,
+                    'url': video_url_value
+                })
         self._check_formats(formats, video_id)
         self._sort_formats(formats)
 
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +44,15 @@ def _real_extract(self, url):
         webpage = self._download_webpage(
             'http://www.drtuber.com/video/%s' % video_id, display_id)
 
-        video_url = self._html_search_regex(
-            r'<source src="([^"]+)"', webpage, 'video URL')
+        # the video URL must be extracted from an external JSON stream
```
dstftw: Remove pointless comments.

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +44,15 @@ def _real_extract(self, url):
...
+        video_data = self._download_json(
+            'http://www.drtuber.com/player_config_json/?vid=%s&embed=0&aid=0&domain_id=0'
```
dstftw: Query should be passed as `query`.

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +44,15 @@ def _real_extract(self, url):
...
+            % video_id, video_id)
+        # is a high-quality video available?
+        video_url = video_data['files']['hq']
```
dstftw: Extraction is broken if any these key is missing.

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +44,15 @@ def _real_extract(self, url):
...
+        if video_url is None:
+            # nope - use the standard-quality video instead
+            video_url = video_data['files']['lq']
```
dstftw: Both formats should be extracted.

youtube_dl/extractor/drtuber.py
parmjitv: Correct.  The extractor expects that the ['files', 'hq', 'lq'] keys be present in the JSON data in order to determine the video URL(s).  If this is not the case, something has changed on the server end of things.

How should this scenario be handled?  Should the extractor just raise a new ExtractorError and exit?

youtube_dl/extractor/drtuber.py
parmjitv: To clarify: when both a high-quality and low-quality video file are available for a given video ID, *both* video files should be downloaded by the extractor.  Is this correct?

youtube_dl/extractor/drtuber.py
dstftw: Bother to read

parmjitv:
I have applied the requested fixes - please let me know if this is now to your satisfaction.  Thanks!

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +44,33 @@ def _real_extract(self, url):
...
+                    'url': video_url_lq
+                })
+            # high-quality video format is preferred, if available
+            video_url_hq = video_data.get('files', {}).get('hq')
+            if video_url_hq:
+                formats.append({
+                    'format_id': 'HQ',
+                    'quality': 2,
+                    'url': video_url_hq
+                })
```
dstftw: 56-71 code duplication.

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +44,33 @@ def _real_extract(self, url):
...
+        self._check_formats(formats, video_id)
```
dstftw: Add test for a video with unavailable formats.

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +44,33 @@ def _real_extract(self, url):
...
+        self._sort_formats(formats, 'quality')
```
dstftw: Default sorting is just fine. Remove `field_preference`.

parmjitv:
I have applied the requested fixes, please review at your convenience.  Thanks!

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +45,27 @@ def _real_extract(self, url):
...
+            for video_url_key in video_data.get('files', {}):
+                # high-quality video format is preferred, if available
+                if video_data['files'][video_url_key]:
+                    formats.append({
+                        'format_id': video_url_key.upper(),
+                        'quality': 2 if video_url_key == 'hq' else 1,
+                        'url': video_data['files'][video_url_key]
+                    })
+        if not formats:
+            raise ExtractorError('Video %s is not available' % video_id, expected=True)
```
dstftw: This is already checked in `_sort_formats`.

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +45,27 @@ def _real_extract(self, url):
...
+            'vid': video_id,
+            'embed': 0,
+            'aid': 0,
+            'domain_id': 0,
+        }, fatal=False)
+
+        formats = []
+        if video_data:
+            for video_url_key in video_data.get('files', {}):
+                # high-quality video format is preferred, if available
```
dstftw: Remove pointless comments.

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +45,27 @@ def _real_extract(self, url):
...
             'http://www.drtuber.com/video/%s' % video_id, display_id)
 
-        video_url = self._html_search_regex(
-            r'<source src="([^"]+)"', webpage, 'video URL')
+        video_data = self._download_json('http://www.drtuber.com/player_config_json/', video_id, query={
+            'vid': video_id,
+            'embed': 0,
+            'aid': 0,
+            'domain_id': 0,
+        }, fatal=False)
```
dstftw: `fatal=False` is pointless here since you don't have any fallback.

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +45,27 @@ def _real_extract(self, url):
...
+
+        formats = []
+        if video_data:
```
dstftw: This is pointless, you don't have any fallback.

parmjitv:
I have applied the requested fixes, please review at your convenience. Thanks!

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +44,23 @@ def _real_extract(self, url):
...
             'http://www.drtuber.com/video/%s' % video_id, display_id)
 
-        video_url = self._html_search_regex(
-            r'<source src="([^"]+)"', webpage, 'video URL')
+        video_data = self._download_json('http://www.drtuber.com/player_config_json/', video_id, query={
+            'vid': video_id,
+            'embed': 0,
+            'aid': 0,
+            'domain_id': 0,
+        }, fatal=True)
```
dstftw: `True` is default.

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +44,23 @@ def _real_extract(self, url):
...
+
+        formats = []
+        for video_url_key in video_data.get('files', {}):
```
dstftw: What's the point of `.get('files', {})`?
Use `for k,v in data.items()` instead.

parmjitv:
Would it be logical to use something like [ for k,v in data['files'].items() ] instead?

The video files are referenced as sub-keys in the the JSON data - the above snippet would allow us to get at this data immediately, without having to loop across a bunch of irrelevant values.

parmjitv:
I have applied the requested fixes, please review at your convenience. Thanks!

dstftw:
>Would it be logical to use something like [ for k,v in data['files'].items() ] instead?

This is what I've suggested.

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +44,25 @@ def _real_extract(self, url):
...
+            'vid': video_id,
+            'embed': 0,
+            'aid': 0,
+            'domain_id': 0,
+        })
+
+        formats = []
+        for video_key, video_value in video_data.items():
+            if video_key == 'files':
+                for video_url_key, video_url_value in video_data[video_key].items():
```
dstftw: 55-57 what are you doing? `for video_url_key, video_url_value in video_data['files'].items():`. That's all.

youtube_dl/extractor/drtuber.py
```diff
@@ -44,8 +44,25 @@ def _real_extract(self, url):
...
+                    if video_url_value:
+                        formats.append({
+                            'format_id': video_url_key.upper(),
+                            'quality': 2 if video_url_key == 'hq' else 1,
+                            'url': video_url_value
+                        })
+        self._check_formats(formats, video_id)
```
dstftw: See previous review.

youtube_dl/extractor/drtuber.py
parmjitv: I figured an explicit check for the 'files' subkey would be a cleaner way of getting at the video data.  I will revert the change as you request.

youtube_dl/extractor/drtuber.py
parmjitv: "Add test for a video with unavailable formats."

Can you please clarify?  Are you referring to a test case in which the video URL yields a 404 NotFound or otherwise generic 'not available' error?

parmjitv:
1) The for-in loop has been corrected with the latest commit.

2) Can you please elaborate on the last point of the previous review - adding a test for a video with 'unavailable formats'?  Does this refer to a test case where the video URL gives a 404 NotFound or unavailable error?  Or does this refer to a video that has an unexpected format such as mp4, flv etc.

I have read the documentation and looked at the other extractors, but I do not see any references to a test of this nature.  I would appreciate any clarification.  Thank you!

dstftw:
`_check_formats` is used to check formats availability and filter unavailable ones. It's only used for extractors that may serve unavailable formats. You must provide such example URL as you use it your code.

parmjitv:
Ah, that makes sense!  Thanks for letting me know about the correct use.

AFAIK, this site does not serve any unavailable formats for videos - shall I just remove the call to `_check_formats` for this extractor?

parmjitv:
I have removed the unnecessary call to `_check_formats` - please review at your convenience and let me know if there are any further corrections to be made.  Thanks!

PR #13414, opened 2017-06-17
Support for rai livestreams

rabblac:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This Pull Requests aims to support rai live streams listed [here](http://www.raiplay.it/dirette/).

On youtube-dl seems to work fine, but other players try to get index m3u8 the next fragments, which leads to an access denied error. (if you need them, [mpv logs with the two streams](https://gist.github.com/rabblac/0f6506da4ec37a9ae119d9dee5d5770b))

### DIFF
--- a/youtube_dl/extractor/rai.py
+++ b/youtube_dl/extractor/rai.py
@@ -225,44 +225,12 @@ class RaiPlayLiveIE(RaiBaseIE):
         webpage = self._download_webpage(url, channel)
         re_id = r'<div([^>]*)data-uniquename=(["\'])[\w-]*(?P<id>%s)(\2)([^>]*?)>' % RaiBaseIE._UUID_RE
         video_id = self._html_search_regex(re_id, webpage, 'livestream-id', group='id')
-        url = 'http://www.raiplay.it/dirette/ContentItem-%s.html' % video_id
 
-        media = self._download_json(
-            '%s?json' % url, video_id, 'Downloading video JSON')
-
-        title = media['name']
-        video = media['video']
-
-        relinker_info = self._extract_relinker_info(video['contentUrl'], video_id)
-        self._sort_formats(relinker_info['formats'])
-
-        thumbnails = []
-        if 'images' in media:
-            for _, value in media.get('images').items():
-                if value:
-                    thumbnails.append({
-                        'url': value.replace('[RESOLUTION]', '600x400')
-                    })
-
-        timestamp = unified_timestamp(try_get(
-            media, lambda x: x['availabilities'][0]['start'], compat_str))
-
-        subtitles = self._extract_subtitles(url, video.get('subtitles'))
-
-        info = {
-            'id': video_id,
-            'title': title,
-            'alt_title': media.get('subtitle'),
-            'description': media.get('description'),
-            'uploader': media.get('channel'),
-            'creator': media.get('editor'),
-            'timestamp': timestamp,
-            'thumbnails': thumbnails,
-            'subtitles': subtitles,
+        return {
+            '_type': 'url_transparent',
+            'url': 'http://www.raiplay.it/dirette/ContentItem-%s.html' % video_id,
+            'ie_key': RaiPlayIE.ie_key()
         }
-
-        info.update(relinker_info)
-        return info
 
 
 class RaiIE(RaiBaseIE):
--- a/youtube_dl/extractor/rai.py
+++ b/youtube_dl/extractor/rai.py
@@ -136,7 +136,7 @@ class RaiPlayIE(RaiBaseIE):
             'upload_date': '20161029',
             'series': 'La Casa Bianca',
             'season': '2016',
-        }
+        },
     }, {
         'url': 'http://www.raiplay.it/video/2014/04/Report-del-07042014-cb27157f-9dd0-4aee-b788-b1f67643a391.html',
         'md5': '8970abf8caf8aef4696e7b1f2adfc696',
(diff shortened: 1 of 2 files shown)


dstftw:
Add as separate extractor.

gfabiano:
yeah @rabblac , I was also working on this extractor and I noticed that mpv does not handle it good. I don't understand why because it uses ffmpeg as youtube-dl. The only solution for now i found is to use pipes:
`youtube-dl http://raiplay.it/dirette/rai3 -o - | mpv -`

rabblac:
That's why mpv asks every time for the index m3u8 for new fragments, meanwhile youtube-dl increments by one the fragment.
Anyway, thanks for the fast reply dstftw, I'm working on this.

youtube_dl/extractor/rai.py
```diff
@@ -218,7 +208,60 @@ def _real_extract(self, url):
...
+            'description': media.get('description'),
+            'uploader': media.get('channel'),
+            'creator': media.get('editor'),
+            'timestamp': timestamp,
+            'thumbnails': thumbnails,
+            'subtitles': subtitles,
+        }
 
+        info.update(relinker_info)
         return info
```
gfabiano: It's not better in this way?
```python
class RaiPlayLiveIE(RaiBaseIE):
    _VALID_URL = r'(?P<url>https?://(?:www\.)?raiplay\.it/dirette/(?P<id>\w*))'
    _TEST = {
        'url': 'http://www.raiplay.it/dirette/rai3',
        'only_matching': True,
    }

    def _real_extract(self, url):
        url, channel = mobj.group('url', 'id')

        webpage = self._download_webpage(url, channel)
        re_id = r'<div([^>]*)data-uniquename=(["\'])[\w-]*(?P<id>%s)(\2)([^>]*?)>' % RaiBaseIE._UUID_RE
        video_id = self._html_search_regex(re_id, webpage, 'livestream-id', group='id')

        return {
            '_type': 'url_transparent',
            'url': 'http://www.raiplay.it/dirette/ContentItem-%s.html' % video_id,
            'ie_key': RaiPlayIE.ie_key()
        }
```

rabblac: It works well, so why not? Let's see what @dstftw says.

youtube_dl/extractor/rai.py
```diff
@@ -208,7 +208,60 @@ def _real_extract(self, url):
...
+            'alt_title': media.get('subtitle'),
+            'description': media.get('description'),
+            'uploader': media.get('channel'),
+            'creator': media.get('editor'),
+            'timestamp': timestamp,
+            'thumbnails': thumbnails,
+            'subtitles': subtitles,
+        }
+
+        info.update(relinker_info)
```
dstftw: 230-264 no copy pastes.

dstftw:
As already mentioned by @gfabiano you should **delegate** extraction to `RaiPlayIE` **not copy paste code**.

rabblac:
Sure thing, didn't know that was possible. Commit sent.

youtube_dl/extractor/rai.py
```diff
@@ -208,10 +208,31 @@ def _real_extract(self, url):
...
+class RaiPlayLiveIE(RaiBaseIE):
+    _VALID_URL = r'(?P<url>https?://(?:www\.)?raiplay\.it/dirette/(?P<id>\w*))'
+    _TEST = {
+        'url': 'http://www.raiplay.it/dirette/rai3',
+        'only_matching': True,
+    }
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        url, channel = mobj.group('url', 'id')
```
dstftw: What's the point capturing `url` when you already have it in `url`. Rewrite with `_match_id`.

youtube_dl/extractor/rai.py
```diff
@@ -208,10 +208,31 @@ def _real_extract(self, url):
...
+
+        webpage = self._download_webpage(url, channel)
+        re_id = r'<div([^>]*)data-uniquename=(["\'])[\w-]*(?P<id>%s)(\2)([^>]*?)>' % RaiBaseIE._UUID_RE
+        video_id = self._html_search_regex(re_id, webpage, 'livestream-id', group='id')
+
+        return {
+            '_type': 'url_transparent',
+            'url': 'http://www.raiplay.it/dirette/ContentItem-%s.html' % video_id,
+            'ie_key': RaiPlayIE.ie_key()
```
dstftw: Use `url_result` if you don't adjust any metadata.

youtube_dl/extractor/rai.py
```diff
@@ -156,7 +156,7 @@ class RaiPlayIE(RaiBaseIE):
         },
         'params': {
             'skip_download': True,
-        },
+        }
```
dstftw: Remove all pointless changes.

youtube_dl/extractor/rai.py
```diff
@@ -208,10 +208,27 @@ def _real_extract(self, url):
         }
 
         info.update(relinker_info)
-
         return info
 
 
+class RaiPlayLiveIE(RaiBaseIE):
+    _VALID_URL = r'https?://(?:www\.)?raiplay\.it/dirette/(?P<id>\w*)'
```
dstftw: http://www.raiplay.it/dirette/ should not be matched.

gfabiano: You need only to change `(?P<id>\w*)` with `(?P<id>\w+)`

rabblac: Thanks

PR #13449, opened 2017-06-21
[Panopto] Add Panopto extractors

kmark:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

No test cases are included as I am not aware of any publicly
available Panopto recordings that this extractor will work with.

Supports downloading individual recordings or entire folders
recursively. Folders are seperated with a ' -- ' in the playlist
title.

Cookies are likely required to use this extractor specifically their
.ASPXAUTH cookie which can be obtained from your browser after
logging in.

--write-all-thumbnails can be used to download PowerPoint slides if
they are not included as a video stream.

Suggested output format is 'out/%(playlist)s/%(title)s.%(ext)s'

### DIFF
--- a/youtube_dl/extractor/panopto.py
+++ b/youtube_dl/extractor/panopto.py
@@ -17,6 +17,7 @@ import json
 
 
 class PanoptoBaseIE(InfoExtractor):
+    """The base class with common methods for Panopto extractors."""
 
     @classmethod
     def _match_organization(cls, url):
@@ -28,6 +29,7 @@ class PanoptoBaseIE(InfoExtractor):
 
 
 class PanoptoIE(PanoptoBaseIE):
+    """Extracts a single Panopto video including all available streams."""
 
     _VALID_URL = r'^https?:\/\/(?P<org>[a-z0-9]+)\.hosted\.panopto.com\/Panopto\/Pages\/Viewer\.aspx\?id=(?P<id>[a-f0-9-]+)'
 
@@ -130,6 +132,8 @@ class PanoptoIE(PanoptoBaseIE):
 
 
 class PanoptoFolderIE(PanoptoBaseIE):
+    """Recursively extracts a folder of Panopto videos, digging as far as possible into subfolders."""
+
     _VALID_URL = r'^https?:\/\/(?P<org>[a-z0-9]+)\.hosted\.panopto.com\/Panopto\/Pages\/Sessions\/List\.aspx#folderID=(?:"|%22)(?P<id>[a-f0-9-]+)'
 
     def _real_extract(self, url):
--- a/youtube_dl/extractor/panopto.py
+++ b/youtube_dl/extractor/panopto.py
@@ -196,7 +196,7 @@ class PanoptoIE(PanoptoBaseIE):
 class PanoptoFolderIE(PanoptoBaseIE):
     """Recursively extracts a folder of Panopto videos, digging as far as possible into subfolders."""
 
-    _VALID_URL = r'^https?://(?P<org>[a-z0-9]+)\.hosted\.panopto.com/Panopto/Pages/Sessions/List\.aspx#folderID=(?:"|%22)(?P<id>[a-f0-9-]+)'
+    _VALID_URL = r'^https?://(?P<org>[a-z0-9]+)\.hosted\.panopto.com/Panopto/Pages/Sessions/List\.aspx(?:\?.*)?#(?:.*&)?folderID=(?:"|%22)(?P<id>[a-f0-9-]+)'
     _TESTS = [
         {
             'url': 'https://demo.hosted.panopto.com/Panopto/Pages/Sessions/List.aspx#folderID=%224540f269-8bb1-4352-b5dc-64e5919d1c40%22',
--- a/youtube_dl/extractor/panopto.py
+++ b/youtube_dl/extractor/panopto.py
@@ -33,6 +33,48 @@ class PanoptoIE(PanoptoBaseIE):
     """Extracts a single Panopto video including all available streams."""
 
     _VALID_URL = r'^https?:\/\/(?P<org>[a-z0-9]+)\.hosted\.panopto.com\/Panopto\/Pages\/Viewer\.aspx\?id=(?P<id>[a-f0-9-]+)'
+    _TESTS = [
+        {
+            'url': 'https://demo.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=26b3ae9e-4a48-4dcc-96ba-0befba08a0fb',
+            'md5': '06fb292a3510aa5bc5f0c950fe58c9f7',
+            'info_dict': {
+                'id': '26b3ae9e-4a48-4dcc-96ba-0befba08a0fb',
+                'ext': 'mp4',
+                'title': 'Panopto for Business',
+                'uploader': 'Ari Bixhorn',
+                'upload_date': '20160328',
+                'timestamp': 1459184200.3759995,
+            },
+        },
+        {
+            'url': 'https://demo.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=ed01b077-c9e5-4c7b-b8ff-15fa306d7a59',
+            'info_dict': {
+                'id': 'ed01b077-c9e5-4c7b-b8ff-15fa306d7a59',
+                'title': 'Overcoming Top 4 Challenges of Enterprise Video',
+                'uploader': 'Panopto Support',
+                'timestamp': 1449409251.8579998,
+            },
+            'playlist': [
+                {
+                    'md5': 'e22b5a284789ba2681e4fe215352d816',
+                    'info_dict': {
+                        'id': '15ad06ef-3f7d-4074-aa4a-87c41dd18f9c',
+                        'ext': 'mp4',
+                        'title': 'OBJECT',
+                    },
+                },
+                {
+                    'md5': '4396cbff07e7b883ca522a6783dc6a70',
+                    'info_dict': {
+                        'id': '7668d6b2-dc81-421d-9853-20653689e2e8',
+                        'ext': 'mp4',
+                        'title': 'DV',
+                    },
+                },
+            ],
+            'playlist_count': 2,
+        },
+    ]
 
     @staticmethod
     def _get_contribs_str(contribs):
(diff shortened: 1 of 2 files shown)


dstftw:
1. Read
>[youtube-dl coding conventions](https://github.com/rg3/youtube-dl#youtube-dl-coding-conventions)

and fix all issues.
2. You have to provide account credentials and concrete example URL for testing.

kmark:
I've fixed all the issues noted by QuantifiedCode. I've also added tests for the Panopto extractor that both work on my machine but fail the Travis test. I have no idea why this is. Please let me know if there are any further places where conventions aren't being followed.

As noted in one of my commits:

> Testing may be impossible for the Folder extractor, or I'm just
doing it wrong. With the current test we enter a catch-22 where it
claims we need an 'ext' entry to continue testing but upon adding
that it claims it expected None.

I've still included the faulty PanoptoFolder test for review and reproduction of this issue.

Thanks!

youtube_dl/extractor/panopto.py
```diff
@@ -0,0 +1,296 @@
...
+            cls._VALID_URL_RE = re.compile(cls._VALID_URL)
+        m = cls._VALID_URL_RE.match(url)
+        assert m
+        return compat_str(m.group('org'))
+
+
+class PanoptoIE(PanoptoBaseIE):
+    """Extracts a single Panopto video including all available streams."""
+
+    _VALID_URL = r'^https?:\/\/(?P<org>[a-z0-9]+)\.hosted\.panopto.com\/Panopto\/Pages\/Viewer\.aspx\?id=(?P<id>[a-f0-9-]+)'
```
dstftw: No need to escape `\`.

youtube_dl/extractor/panopto.py
```diff
@@ -0,0 +1,296 @@
...
+    ]
+
+    @staticmethod
+    def _get_contribs_str(contribs):
+        """Returns a comma-delimited string of contributors."""
+        s = ''
+        for c in contribs:
+            display_name = c.get('DisplayName')
+            if display_name is not None:
+                s += '{}, '.format(display_name)
```
dstftw: `{}` does not work in python 2.6.

youtube_dl/extractor/panopto.py
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        if len(streams) == 1:
+            result['formats'] = streams[0]['formats']
+        else:
+            result['_type'] = 'multi_video'
+            result['entries'] = streams
+
+        # We already know Delivery exists since we need it for stream extraction
+        contributors = delivery_info['Delivery'].get('Contributors')
```
dstftw: Read coding conventions and fix all optional fields.

kmark: I'm not sure how this is at odds with the coding conventions. Are you referring to the use of `['Delivery']`? The Delivery key _must_ exist assuming that code is ever executed because the key is needed to retrieve non-optional fields earlier. It would fail at line 120:

```python
for this_stream in delivery_info['Delivery']['Streams']:
```

Because we need `['Delivery']` (and more specifically `['Delivery']['Streams']`) to even find the non-optional information.

kmark:
The Travis errors appear to be from ffmpeg not being available:

```
DownloadError: ERROR: m3u8 download detected but ffmpeg or avconv could not be found. Please install one.
```

Once ffmpeg is installed it fixed the above error but the tests still failed with MD5 mismatches. It looks like the `--test` option isn't working properly on Linux:

```bash
$ python2.7 -m youtube_dl --test https://demo.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=26b3ae9e-4a48-4dcc-96ba-0befba08a0fb
$ wc -c Panopto\ for\ Business-26b3ae9e-4a48-4dcc-96ba-0befba08a0fb.mp4
13663 Panopto for Business-26b3ae9e-4a48-4dcc-96ba-0befba08a0fb.mp4
$ md5sum Panopto\ for\ Business-26b3ae9e-4a48-4dcc-96ba-0befba08a0fb.mp4
$ 76d6bc8e500b1e47b53541514e3d1ea6  Panopto for Business-26b3ae9e-4a48-4dcc-96ba-0befba08a0fb.mp4
```

or macOS:
```bash
$ python2.7 -m youtube_dl --test https://demo.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=26b3ae9e-4a48-4dcc-96ba-0befba08a0fb
$ gwc -c Panopto\ for\ Business-26b3ae9e-4a48-4dcc-96ba-0befba08a0fb.mp4
13679 Panopto for Business-26b3ae9e-4a48-4dcc-96ba-0befba08a0fb.mp4
$ gmd5sum Panopto\ for\ Business-26b3ae9e-4a48-4dcc-96ba-0befba08a0fb.mp4
06fb292a3510aa5bc5f0c950fe58c9f7  Panopto for Business-26b3ae9e-4a48-4dcc-96ba-0befba08a0fb.mp4
```

Different lengths means both platforms will generate different MD5 sums during testing. If I truncate both to 10241 bytes I get the same MD5 but neither platform will test correctly against that MD5.

kmark:
Looks like the issue is with youtube-dl assuming that the `-fs` option of FFmpeg will limit the file size to 10241 bytes in a deterministic fashion. According to [the documentation](http://ffmpeg.org/ffmpeg.html#Main-options) it doesn't guarantee a final file size and even claims the download will be larger by some undefined degree:

> Set the file size limit, expressed in bytes. No further chunk of bytes is written after the limit is exceeded. The size of the output file is slightly more than the requested file size.

So we have two different versions of FFmpeg on two different platforms with FFmpeg's `-fs` option producing two different final file sizes. When I bring the FFmpeg versions closer together on both platforms (latest or near-latest from master) the file sizes are equal again. Unless I'm misunderstanding something about youtube-dl, I would put this in the bug category.

youtube_dl/extractor/panopto.py
kmark: @dstftw Is this still an issue?

youtube_dl/extractor/panopto.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+class PanoptoIE(PanoptoBaseIE):
+    """Extracts a single Panopto video including all available streams."""
+
+    _VALID_URL = r'^https?://(?P<org>[a-z0-9]+)\.hosted\.panopto.com/Panopto/Pages/Viewer\.aspx\?id=(?P<id>[a-f0-9-]+)'
+    _TESTS = [
+        {
+            'url': 'https://demo.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=26b3ae9e-4a48-4dcc-96ba-0befba08a0fb',
+            'md5': 'e8e6ef6b0572dd5985f5f8c3e096f717',
```
rgreenblatt, line 39:
```suggestion
            'md5': '048dab5c2f8ab97f2bd75ab4cf3f463a',
```
Maybe this video was changed, this now seems to be the correct hash.

youtube_dl/extractor/panopto.py
```diff
@@ -0,0 +1,277 @@
...
+            )
+
+        streams = []
+        for this_stream in delivery_info['Delivery']['Streams']:
+            new_stream = {
+                'id': this_stream['PublicID'],
+                'title': this_stream['Tag'],
+                'formats': [],
+            }
+            if 'StreamHttpUrl' in this_stream:
```
rgreenblatt, line 126:
```suggestion
            if ('StreamHttpUrl' in this_stream and this_stream['StreamHttpUrl'] is not None):

```
It looks like this now can be null/none

youtube_dl/extractor/panopto.py
```diff
@@ -0,0 +1,277 @@
...
+                new_stream['formats'].append({
+                    'url': this_stream['StreamHttpUrl'],
+                })
+            if 'StreamUrl' in this_stream:
```
rgreenblatt, line 130:
```suggestion
            if ('StreamUrl' in this_stream and this_stream['StreamUrl'] is not None):
```
Maybe the same change as for StreamHttpUrl should also be done for StreamUrl? I haven't seen any cases where this is None/null, but it seems like a good idea.

rgreenblatt:
I tested this PR and I was able to get things working (just test/test_download.py and an example private video) after a few changes (see review).

youtube_dl/extractor/panopto.py
```diff
@@ -0,0 +1,277 @@
...
+                m3u8_formats = self._extract_m3u8_formats(this_stream['StreamUrl'], video_id, 'mp4')
+                self._sort_formats(m3u8_formats)
+                new_stream['formats'].extend(m3u8_formats)
```
rgreenblatt, line 133:
```suggestion
                get_stream_url_matcher = lambda mid : re.compile(r'^https?://.*?/.*?/.*?' + mid + r'\?InvocationID.*')
                if '_M3U8_MATCHER' not in self.__dict__:
                    self._M3U8_MATCHER = get_stream_url_matcher(r'\.hls/master\.m3u8')
                if '_MP4_MATCHER' not in self.__dict__:
                    self._MP4_MATCHER = get_stream_url_matcher(r'\.mp4')
                if self._M3U8_MATCHER.match(this_stream['StreamUrl']):
                    m3u8_formats = self._extract_m3u8_formats(this_stream['StreamUrl'], video_id, 'mp4')
                    self._sort_formats(m3u8_formats)
                    new_stream['formats'].extend(m3u8_formats)
                elif self._MP4_MATCHER.match(this_stream['StreamUrl']):
                    new_stream['formats'].append({'url': this_stream['StreamUrl']})
                else:
                    raise ExtractorError('Unexpected StreamUrl format')
```

In some cases, it looks like StreamUrl is a single mp4 file instead of m3u8 (when there is only a single video I assume?). I don't know how fragile this regex is, but it passes all of my tests.

jxu:
https://github.com/jstrieb/panopto-download I just found about this that might help. I looked briefly through the PR and saw the links looked like they were hardcoded. They are available through the RSS feed of each folder currently.

rgreenblatt:
> https://github.com/jstrieb/panopto-download I just found about this that might help. I looked briefly through the PR and saw the links looked like they were hardcoded. They are available through the RSS feed of each folder currently.

Interestingly, the links in the RSS feed don't require any authentication, so it should be possible to convert from the folder URL to the RSS feed url and download all the videos without requiring cookies. Unfortunately, I don't think its possible to get the RSS feed url directly from a viewer video url, so cookies are needed when downloading a single video. 

For my personal use cases, using the RSS feed directly is perfect. I set up a daemon which periodically runs youtube-dl on the RSS feed and sends notifications when a new video is downloaded. See [this script](https://github.com/rgreenblatt/dotfiles/blob/master/shell/.local/bin/download_feed_daemon) from my dotfiles. This doesn't require this PR at all. 

For the use case of downloading a single video, I think that using the RSS feed would increase complexity.

For the folder use case, I think that using the RSS should be considerably cleaner, but I don't know how that would work for recursive folder downloads. Also, I haven't tested recursive folder downloads, so it is possible that they are broken right now. I probably won't write this change for a few reasons, but I think this would be reasonably easy to do (aside from subfolders perhaps).

coletdjnz:
Reviving this in yt-dlp: https://github.com/yt-dlp/yt-dlp/pull/2908

lebdron:
@dirkf Hi! Any chance to get this merged? Thanks!

coletdjnz:
iirc there were some bugs in core that had to be fixed for panopto to work completely, so those will need to be backported too

jxu:
at the risk of sounding like a broken record, if you compare youtube-dl commits and bufixes/new features to yt-dlp, youtube-dl as (mostly) controlled by @dirkf will always be behind in a project that has to keep up with constantly changing unofficial APIs

lebdron:
@jxu no worries, I am working on a backport https://github.com/ytdl-org/youtube-dl/pull/31097 so that there will be at least some implementation

jxu:
@lebdron I digress (rant): from what I can tell, unnecessary energy is spent backporting, but that is the reality of having a better maintained fork compete with the formerly undisputed project and fracturing the community. Especially when the project insists on supporting python2 and the 0.1% of people who choose run youtube-dl on an ancient embedded device that somehow doesn't have python3 but those people put the burden on contributors.

dirkf:
@jxu can feel free not to spend any effort back-porting code to yt-dl, and should probably ignore yt-dl altogether.

Meanwhile thx @lebdron: please include "closes # 13449" in your PR text (but without the space after the #).

PR #13489, opened 2017-06-25
[abc.net.au:iview] Support hq streams when available

d2au:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Adds support for new higher quality streams if available.

### DIFF
--- a/youtube_dl/extractor/abc.py
+++ b/youtube_dl/extractor/abc.py
@@ -124,7 +124,12 @@ class ABCIViewIE(InfoExtractor):
         title = video_params.get('title') or video_params['seriesTitle']
         stream = next(s for s in video_params['playlist'] if s.get('type') == 'program')
 
-        formats = self._extract_akamai_formats(stream['hds-unmetered'], video_id)
+        if 'sd' in stream["stream-labels"]:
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/abc.py
```diff
@@ -124,7 +124,12 @@ def _real_extract(self, url):
         title = video_params.get('title') or video_params['seriesTitle']
         stream = next(s for s in video_params['playlist'] if s.get('type') == 'program')
 
-        formats = self._extract_akamai_formats(stream['hds-unmetered'], video_id)
+        if 'sd' in stream["stream-labels"]:
```
dstftw: Breaks extraction if there is no `stream-labels` key.

youtube_dl/extractor/abc.py
```diff
@@ -124,7 +124,12 @@ def _real_extract(self, url):
...
+            hds = stream['streams']['hds']['sd'].replace('metered', 'um')
+        else:
+            hds = stream['hds-unmetered']
```
dstftw: Both should be extracted.

PR #13525, opened 2017-06-29
[cjsw] Add new extractor

christsmith:

---

- [ ] Bug fix
- [ ] Improvement
- [X] New extractor
- [X] New feature

---

Added new extractor for cjsw radio podcasts. Podcasts only seem to last for a few months on cjsw.com, so I'll have to update the tests occasionally.

### DIFF
--- a/youtube_dl/extractor/cjsw.py
+++ b/youtube_dl/extractor/cjsw.py
@@ -1,0 +1,50 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    unescapeHTML,
+)
+
+
+class CJSWIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?cjsw\.com/program/(?P<program_name>[\S]+)/(?P<id>[0-9]+)/?'
+    IE_NAME = 'cjsw'
+    _TEST = {
+        'url': 'http://cjsw.com/program/freshly-squeezed/episode/20170620',
+        'md5': 'cee14d40f1e9433632c56e3d14977120',
+        'info_dict': {
+            'id': '20170620',
+            'ext': 'mp3',
+            'title': 'Freshly Squeezed',
+            'description': 'Sled Island artists featured // Live session with Phi Pho, followed by a live session with Sinzere & The Late Nights! // Stay Fresh Y\'all!!',
+        }
+    }
+
+    def _real_extract(self, url):
+        episode_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, episode_id)
+
+        episode_controls = re.search(r'<div[^>]+class=(["\'])episode-controls\1[^>]*>', webpage)
+        if not episode_controls:
+            raise ExtractorError('No streamable podcast', video_id=episode_id, expected=True)
+
+        title = unescapeHTML(self._search_regex(
+            r'<button[^>]+data-showname=(["\'])(?P<title>.*?)\1[^>]*>', webpage, 'title', group='title'))
+        description = unescapeHTML(self._html_search_regex(
+            r'<p>(?P<description>.*?)</p>', webpage, 'description', fatal=False))
+        formats = [{
+            'url': self._html_search_regex(
+                r'<button[^>]+data-audio-src=(["\'])(?P<audio_url>.*?)\1[^>]*>', webpage, 'audio_url', group='audio_url'),
+            'ext': 'mp3',
+        }]
+        return {
+            'id': episode_id,
+            'title': title,
+            'description': description,
+            'formats': formats,
+        }
--- a/youtube_dl/extractor/cjsw.py
+++ b/youtube_dl/extractor/cjsw.py
@@ -1,17 +1,14 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
-import re
-
 from .common import InfoExtractor
 from ..utils import (
     ExtractorError,
-    unescapeHTML,
 )
 
 
 class CJSWIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?cjsw\.com/program/(?P<program_name>[\S]+)/(?P<id>[0-9]+)/?'
+    _VALID_URL = r'https?://(?:www\.)?cjsw\.com/program/\S+/(?P<id>[0-9]+)'
     IE_NAME = 'cjsw'
     _TEST = {
         'url': 'http://cjsw.com/program/freshly-squeezed/episode/20170620',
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/cjsw.py
```diff
@@ -0,0 +1,50 @@
...
+        episode_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, episode_id)
+
+        episode_controls = re.search(r'<div[^>]+class=(["\'])episode-controls\1[^>]*>', webpage)
+        if not episode_controls:
+            raise ExtractorError('No streamable podcast', video_id=episode_id, expected=True)
+
+        title = unescapeHTML(self._search_regex(
+            r'<button[^>]+data-showname=(["\'])(?P<title>.*?)\1[^>]*>', webpage, 'title', group='title'))
```
dstftw: Empty string capture does not make any sense.

youtube_dl/extractor/cjsw.py
```diff
@@ -0,0 +1,50 @@
...
+        description = unescapeHTML(self._html_search_regex(
+            r'<p>(?P<description>.*?)</p>', webpage, 'description', fatal=False))
```
dstftw: Empty string capture does not make any sense.

youtube_dl/extractor/cjsw.py
```diff
@@ -0,0 +1,50 @@
...
+        formats = [{
+            'url': self._html_search_regex(
+                r'<button[^>]+data-audio-src=(["\'])(?P<audio_url>.*?)\1[^>]*>', webpage, 'audio_url', group='audio_url'),
```
dstftw: Empty string capture does not make any sense.

youtube_dl/extractor/cjsw.py
```diff
@@ -0,0 +1,50 @@
...
+            'ext': 'mp3',
```
dstftw: `'vcodec': 'none'`.

youtube_dl/extractor/cjsw.py
```diff
@@ -0,0 +1,50 @@
...
+            'description': 'Sled Island artists featured // Live session with Phi Pho, followed by a live session with Sinzere & The Late Nights! // Stay Fresh Y\'all!!',
+        }
+    }
+
+    def _real_extract(self, url):
+        episode_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, episode_id)
+
+        episode_controls = re.search(r'<div[^>]+class=(["\'])episode-controls\1[^>]*>', webpage)
```
dstftw: `_search_regex`.

youtube_dl/extractor/cjsw.py
```diff
@@ -0,0 +1,50 @@
...
+
+from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    unescapeHTML,
+)
+
+
+class CJSWIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?cjsw\.com/program/(?P<program_name>[\S]+)/(?P<id>[0-9]+)/?'
```
dstftw: Don't capture groups you don't use.

dstftw: `/?` is pointless at the end.

youtube_dl/extractor/cjsw.py
```diff
@@ -0,0 +1,50 @@
...
+
+        webpage = self._download_webpage(url, episode_id)
+
+        episode_controls = re.search(r'<div[^>]+class=(["\'])episode-controls\1[^>]*>', webpage)
+        if not episode_controls:
+            raise ExtractorError('No streamable podcast', video_id=episode_id, expected=True)
+
+        title = unescapeHTML(self._search_regex(
+            r'<button[^>]+data-showname=(["\'])(?P<title>.*?)\1[^>]*>', webpage, 'title', group='title'))
+        description = unescapeHTML(self._html_search_regex(
```
dstftw: `unescapeHTML` is already done.

christsmith:
@dstftw Thanks for the code review, it was very helpful. I agree with all your comments & made the changes.

youtube_dl/extractor/cjsw.py
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    def _real_extract(self, url):
+        episode_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, episode_id)
+
+        episode_controls = self._search_regex(r'<div[^>]+class=(["\'])episode-controls\1[^>]*>', webpage, 'episode_controls', fatal=False)
+        if not episode_controls:
+            raise ExtractorError('No streamable podcast', video_id=episode_id, expected=True)
```
dstftw: `episode_controls` is not used. Remove 29-31.

youtube_dl/extractor/cjsw.py
```diff
@@ -0,0 +1,48 @@
...
+
+        title = self._html_search_regex(
+            r'<button[^>]+data-showname=(["\'])(?P<title>.+?)\1[^>]*>', webpage, 'title', group='title')
+        description = self._html_search_regex(
+            r'<p>(?P<description>.+?)</p>', webpage, 'description', fatal=False)
+        formats = [{
+            'url': self._html_search_regex(
```
dstftw: `_search_regex` is enough.

youtube_dl/extractor/cjsw.py
```diff
@@ -0,0 +1,48 @@
...
+        episode_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, episode_id)
+
+        episode_controls = self._search_regex(r'<div[^>]+class=(["\'])episode-controls\1[^>]*>', webpage, 'episode_controls', fatal=False)
+        if not episode_controls:
+            raise ExtractorError('No streamable podcast', video_id=episode_id, expected=True)
+
+        title = self._html_search_regex(
+            r'<button[^>]+data-showname=(["\'])(?P<title>.+?)\1[^>]*>', webpage, 'title', group='title')
```
dstftw: `_search_regex` is enough.

dstftw: Will capture incorrect data for `data-showname="" some-attr="blah"`.

youtube_dl/extractor/cjsw.py
```diff
@@ -0,0 +1,48 @@
...
+        description = self._html_search_regex(
+            r'<p>(?P<description>.+?)</p>', webpage, 'description', fatal=False)
+        formats = [{
+            'url': self._html_search_regex(
+                r'<button[^>]+data-audio-src=(["\'])(?P<audio_url>.+?)\1[^>]*>', webpage, 'audio_url', group='audio_url'),
```
dstftw: Will capture incorrect data for `data-audio-src="" some-attr="blah"`.

youtube_dl/extractor/cjsw.py
christsmith: How does this look?

`r'<button[^>]+data-showname=(["\'])(?P<title>(?!\1).+?)\1[^>]*>'`

youtube_dl/extractor/cjsw.py
christsmith: Some cjsw.com podcast pages don't have a stream to download ([this one for example](http://cjsw.com/program/bass-ackwards/episode/20170407/)). These lines are to search for the episode_controls div that is present when the page has a stream to download & to fail in an expected way if it doesn't. When the page doesn't have anything to download, it would be nice to make it clear that this isn't a bug with youtube_dl.

This does however add unnecessary complexity that could break the extractor when it should be able to work. I'm guessing this is why you want it out?

I saw that the [bandcamp extractor](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/bandcamp.py#L62) did this, but in a way that doesn't risk it breaking unnecessarily.

dstftw: >This does however add unnecessary complexity that could break the extractor when it should be able to work. I'm guessing this is why you want it out?

Yes.

PR #13555, opened 2017-07-03
Loosen tfa_results check

thehunmonkgroup:
- [ x] [Searched](https://github.com/rg3/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests

- [ x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)

- [ x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Can also return None as well as False, check for falsey values instead.

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -213,7 +213,7 @@ class YoutubeBaseInfoExtractor(InfoExtractor):
                     self._TFA_URL.format(tl), tfa_req,
                     'Submitting TFA code', 'Unable to submit TFA code')
 
-                if tfa_results is False:
+                if not tfa_results:
                     return False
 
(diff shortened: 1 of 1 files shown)


thehunmonkgroup:
Can someone please explain to me what I need to do to move this pull request forward? The test failures seem to have nothing to do with the minor change I implemented. Does somebody need to kick the test bot or something?

dstftw:
Clear explanation what it's supposed to fix, steps to reproduce.

thehunmonkgroup:
I was trying to get TFA working, and noticed that tfa_results, in my case, was returning as None, not False -- thus the check failed and the code did not properly return when it should be in that case. Loosening the check to a falsey value seems a reasonable solution to the issue, as that will catch both None and False, while passing any non-falsey value.

PR #13565, opened 2017-07-04
Extractor for rutube.ru playlists (issues/13534)

luceatnobis:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

### DIFF
--- a/youtube_dl/extractor/rutube.py
+++ b/youtube_dl/extractor/rutube.py
@@ -7,6 +7,8 @@ import itertools
 from .common import InfoExtractor
 from ..compat import (
     compat_str,
+    compat_parse_qs,
+    compat_urllib_parse_urlparse,
 )
 from ..utils import (
     determine_ext,
@@ -18,7 +20,7 @@ from ..utils import (
 class RutubeIE(InfoExtractor):
     IE_NAME = 'rutube'
     IE_DESC = 'Rutube videos'
-    _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/(?P<id>[\da-z]{32})(/)?$'
+    _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/(?P<id>[\da-z]{32})'
 
     _TESTS = [{
         'url': 'http://rutube.ru/video/3eac3b4561676c17df9132a9a1e62e3e/',
@@ -44,6 +46,10 @@ class RutubeIE(InfoExtractor):
         'url': 'http://rutube.ru/embed/a10e53b86e8f349080f718582ce4c661',
         'only_matching': True,
     }]
+
+    @classmethod
+    def suitable(cls, url):
+        return False if RutubePlaylistIE.suitable(url) else super(RutubeIE, cls).suitable(url)
 
     @staticmethod
     def _extract_urls(webpage):
@@ -206,16 +212,20 @@ class RutubePlaylistIE(InfoExtractor):
         'playlist_count': 25,
     }]
 
-    _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/[\da-z]{32}/\?.+pl_id=(?P<pl_id>\d+).*$'
+    _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/[\da-z]{32}/\?(?:.+)?pl_id=(?P<id>\d+).*$'
     _PAGE_TEMPLATE = 'http://rutube.ru/api/playlist/source/%s/?page=%s'
 
+
+    @classmethod
+    def suitable(cls, url):
+        params = compat_parse_qs(compat_urllib_parse_urlparse(url).query)
+        return params.get('pl_id') and params['pl_id'][0].isdigit()
+
     def _real_extract(self, url):
-        m = re.match(self._VALID_URL, url)
-        playlist_id = m.group('pl_id')
+        playlist_id = self._match_id(url)
         return self._extract_playlist(playlist_id)
 
     def _extract_playlist(self, playlist_id):
-        import pdb
         entries = []
         for pagenum in itertools.count(1):
             page_url = self._PAGE_TEMPLATE % (playlist_id, pagenum)
--- a/youtube_dl/extractor/rutube.py
+++ b/youtube_dl/extractor/rutube.py
@@ -216,8 +216,8 @@ class RutubePlaylistIE(InfoExtractor):
     _PAGE_TEMPLATE = 'http://rutube.ru/api/playlist/source/%s/?page=%s'
 
 
-    @classmethod
-    def suitable(cls, url):
+    @staticmethod
+    def suitable(url):
         params = compat_parse_qs(compat_urllib_parse_urlparse(url).query)
         return params.get('pl_id') and params['pl_id'][0].isdigit()
 
--- a/youtube_dl/extractor/rutube.py
+++ b/youtube_dl/extractor/rutube.py
@@ -203,6 +203,7 @@ class RutubePersonIE(RutubeChannelIE):
 
     _PAGE_TEMPLATE = 'http://rutube.ru/api/video/person/%s/?page=%s&format=json'
 
+
 class RutubePlaylistIE(InfoExtractor):
     IE_NAME = 'rutube:playlist'
     IE_DESC = 'Rutube playlists'
@@ -216,7 +217,6 @@ class RutubePlaylistIE(InfoExtractor):
 
     _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/[\da-z]{32}/\?(?:.+)?pl_id=(?P<id>\d+)'
     _PAGE_TEMPLATE = 'http://rutube.ru/api/playlist/source/%s/?page=%s'
-
 
     @staticmethod
     def suitable(url):
--- a/youtube_dl/extractor/rutube.py
+++ b/youtube_dl/extractor/rutube.py
@@ -253,7 +253,7 @@ class RutubePlaylistIE(InfoExtractor):
                     'description': result.get('description'),
                     'thumbnail': result.get('thumbnail_url'),
                     'duration': int_or_none(result.get('duration')),
-                    'category': try_get(result, lambda x: x['category']['name']),
+                    'category': [try_get(result, lambda x: x['category']['name'])],
                     'age_limit': 18 if result.get('is_adult') else 0,
                     'view_count': int_or_none(result.get('hits')),
                     'is_live': result.get('is_livestream'),
--- a/youtube_dl/extractor/rutube.py
+++ b/youtube_dl/extractor/rutube.py
@@ -212,7 +212,7 @@ class RutubePlaylistIE(InfoExtractor):
         'playlist_count': 25,
     }]
 
-    _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/[\da-z]{32}/\?(?:.+)?pl_id=(?P<id>\d+).*$'
+    _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/[\da-z]{32}/\?(?:.+)?pl_id=(?P<id>\d+)'
     _PAGE_TEMPLATE = 'http://rutube.ru/api/playlist/source/%s/?page=%s'
 
 
--- a/youtube_dl/extractor/rutube.py
+++ b/youtube_dl/extractor/rutube.py
@@ -198,7 +198,7 @@ class RutubePlaylistIE(InfoExtractor):
     IE_NAME = 'rutube:playlist'
     IE_DESC = 'Rutube playlists'
     _TESTS = [{
-        'url': 'https://rutube.ru/video/10b3a03fc01d5bbcc632a2f3514e8aab/?pl_id=4252&pl_type=source'
+        'url': 'https://rutube.ru/video/10b3a03fc01d5bbcc632a2f3514e8aab/?pl_id=4252&pl_type=source',
         'info_dict': {
             'id': '4252',
         },
--- a/youtube_dl/extractor/rutube.py
+++ b/youtube_dl/extractor/rutube.py
@@ -231,10 +231,9 @@ class RutubePlaylistIE(InfoExtractor):
             page_url = self._PAGE_TEMPLATE % (playlist_id, pagenum)
 
             # download_json will sent an accept: application/xml header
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/rutube.py
```diff
@@ -17,7 +17,7 @@
 class RutubeIE(InfoExtractor):
     IE_NAME = 'rutube'
     IE_DESC = 'Rutube videos'
-    _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/(?P<id>[\da-z]{32})'
+    _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/(?P<id>[\da-z]{32})(/)?$'
```
dstftw: No. Override `suitable`.

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +193,59 @@ class RutubePersonIE(RutubeChannelIE):
...
+    IE_DESC = 'Rutube playlists'
+    _TESTS = [{
+        'url': 'https://rutube.ru/video/10b3a03fc01d5bbcc632a2f3514e8aab/?pl_id=4252&pl_type=source',
+        'info_dict': {
+            'id': '4252',
+        },
+        'playlist_count': 25,
+    }]
+
+    _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/(?P<id>[\da-z]{32})/.+pl_id=(?P<pl_id>\d+).*$'
```
dstftw: Don't capture groups you don't use.
This should not match `https://rutube.ru/video/10b3a03fc01d5bbcc632a2f3514e8aab/pl_id=4252`.
`.*$` does not make any sense at the end.

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +193,59 @@ class RutubePersonIE(RutubeChannelIE):
...
+                                      pagenum, headers=headers)
+
+            if not page['results']:
+                break
+
+            results = page['results']
+            for result in results:
+                entry = self.url_result(result['video_url'], 'Rutube')
+                entry['id'] = result['id']
+                entry['uploader'] = result['author']['name']
```
dstftw: Read coding conventions on optional fields.

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +193,59 @@ class RutubePersonIE(RutubeChannelIE):
...
+        },
+        'playlist_count': 25,
+    }]
+
+    _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/(?P<id>[\da-z]{32})/.+pl_id=(?P<pl_id>\d+).*$'
+    _PAGE_TEMPLATE = 'http://rutube.ru/api/playlist/source/%s/?page=%s'
+
+    def _real_extract(self, url):
+        m = re.match(self._VALID_URL, url)
+        playlist_id = m.group('pl_id')
```
dstftw: `_match_id`.

youtube_dl/extractor/rutube.py
luceatnobis: To be completely fair, I should use `suitable` here as well. The general issue is that rutube.ru playlists require two parameters on the web interface for playlists to initialise as intended, `pl_id` and `pl_type`. It may be wise to assume that the parameters can be in arbitrary order, however, I couldn't find out how to do that in pure regex. `.*$` was intended to indicate that `pl_type` may well come after `pl_id`.

I'll try to fix this using overloading `suitable`.

youtube_dl/extractor/rutube.py
luceatnobis: How should that look on nested dictionary enries like `result['author']['name']`? Would `result.get('author', {}).get('name')` be suitable?

youtube_dl/extractor/rutube.py
luceatnobis: `match_id` has `id` hardcoded as the capture group that is returned, however, we need to query the API with `pl_id`. Can you explain how you think `match_id` is applicable?

youtube_dl/extractor/rutube.py
dstftw: `try_get`.

youtube_dl/extractor/rutube.py
luceatnobis: In `RutubeIE` as well as in `RutubePlaylistIE`? Please clarify.

youtube_dl/extractor/rutube.py
dstftw: Capture as `id` obviously.

youtube_dl/extractor/rutube.py
luceatnobis: There are several reasons why this strikes me as a non-optimal solution. We're really only asserting that no previous regexes match our URL; this makes no positive case for our playlist URL, we don't test for its characteristics. Its probably not good for further functionality on that site; we don't know what else youtube-dl may want to support in the future. Adopting that approach may lead someone to add a different extractor in the future and our playlist extractor may match that specification, simply because its not defined properly. Trying to then fix the issue makes us fiddle with parts of the program that aren't concerned with the new extractor.

Besides, that only brings us back to the question of how the regular expression for two parameters in a URL with arbitrary order can be resolved. I don't see your approach solving much of a problem.

dstftw: >There are several reasons why this strikes me as a non-optimal solution.

It's not broken as yours at least.

>we don't test for its characteristics

Write code so that it would test "for its characteristics" obviously.

>Adopting that approach may lead someone to add a different extractor in the future and our playlist extractor may match that specification, simply because its not defined properly

No, this won't happen due to `test_all_urls.py`. Conversely, usually distinguishing extractors with regexes only lead to breaking these tests.

>I don't see your approach solving much of a problem.

Write code that will `parse_qs` the query in `suitable` and check presence of required arguments. Solves it perfectly.

luceatnobis:
@dstftw Can I have some feedback on the recent changes I made?

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +200,64 @@ class RutubePersonIE(RutubeChannelIE):
...
+    }]
+
+    _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/[\da-z]{32}/\?(?:.+)?pl_id=(?P<id>\d+).*$'
+    _PAGE_TEMPLATE = 'http://rutube.ru/api/playlist/source/%s/?page=%s'
+
+
+    @staticmethod
+    def suitable(url):
+        params = compat_parse_qs(compat_urllib_parse_urlparse(url).query)
+        return params.get('pl_id') and params['pl_id'][0].isdigit()
```
dstftw: Incorrect.
1. This matches **any** URL with `pl_id` in query.
2. You must check for both `pl_id` and `pl_type`.

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +200,64 @@ class RutubePersonIE(RutubeChannelIE):
...
+    IE_DESC = 'Rutube playlists'
+    _TESTS = [{
+        'url': 'https://rutube.ru/video/10b3a03fc01d5bbcc632a2f3514e8aab/?pl_id=4252&pl_type=source',
+        'info_dict': {
+            'id': '4252',
+        },
+        'playlist_count': 25,
+    }]
+
+    _VALID_URL = r'https?://rutube\.ru/(?:video|(?:play/)?embed)/[\da-z]{32}/\?(?:.+)?pl_id=(?P<id>\d+).*$'
```
dstftw: I've already pointed out: `.*$` is pointless at the end.

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +200,64 @@ class RutubePersonIE(RutubeChannelIE):
...
+                entry['upload_date'] = unified_strdate(result.get('created_ts'))
+                entry['title'] = result.get('title')
+                entry['description'] = result.get('description')
+                entry['thumbnail'] = result.get('thumbnail_url')
+                entry['duration'] = result.get('duration')
+                entry['category'] = try_get(result, lambda x: x['category']['name'])
+                entry['age_limit'] = 18 if result.get('is_adult') else 0
+                entry['view_count'] = result.get('hits')
+                entry['is_live'] = result.get('is_livestream')
+                entry['webpage_url'] = result.get('video_url')
```
dstftw: Replace 245-257 with `entry.update({ ... })`.

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +200,64 @@ class RutubePersonIE(RutubeChannelIE):
...
+            for result in results:
+                entry = self.url_result(result.get('video_url'), 'Rutube')
+                entry['id'] = result['id']
+                entry['uploader'] = try_get(result, lambda x: x['author']['name'])
+                entry['uploader_id'] = try_get(result, lambda x: x['author']['id'])
+                entry['upload_date'] = unified_strdate(result.get('created_ts'))
+                entry['title'] = result.get('title')
+                entry['description'] = result.get('description')
+                entry['thumbnail'] = result.get('thumbnail_url')
+                entry['duration'] = result.get('duration')
```
dstftw: `int_or_none`.

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +200,64 @@ class RutubePersonIE(RutubeChannelIE):
...
+                entry['category'] = try_get(result, lambda x: x['category']['name'])
```
dstftw: Must be `list`.

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +200,64 @@ class RutubePersonIE(RutubeChannelIE):
...
+                entry['age_limit'] = 18 if result.get('is_adult') else 0
+                entry['view_count'] = result.get('hits')
```
dstftw: `int_or_none`.

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +200,64 @@ class RutubePersonIE(RutubeChannelIE):
...
+
+    def _extract_playlist(self, playlist_id):
+        entries = []
+        for pagenum in itertools.count(1):
+            page_url = self._PAGE_TEMPLATE % (playlist_id, pagenum)
+
+            # download_json will sent an accept: application/xml header
+            headers = {'Accept': 'application/json'}
+            page = self._download_json(page_url, playlist_id,
+                                      "Downloading metadata for page %s" %
```
dstftw: Single quotes.

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +200,64 @@ class RutubePersonIE(RutubeChannelIE):
...
+        playlist_id = self._match_id(url)
+        return self._extract_playlist(playlist_id)
+
+    def _extract_playlist(self, playlist_id):
+        entries = []
+        for pagenum in itertools.count(1):
+            page_url = self._PAGE_TEMPLATE % (playlist_id, pagenum)
+
+            # download_json will sent an accept: application/xml header
+            headers = {'Accept': 'application/json'}
```
dstftw: Move into method call.

dstftw:
Check code with flake8.

youtube_dl/extractor/rutube.py
luceatnobis: Err, do you think they should be around the text in the `_download_json` call or do you think they should not be there in the `headers` definition?

youtube_dl/extractor/rutube.py
```diff
@@ -44,6 +48,11 @@ class RutubeIE(InfoExtractor):
         'only_matching': True,
     }]
 
+    @classmethod
+    def suitable(cls, url):
+        parts = compat_urllib_parse_urlparse(url)
+        return super(RutubeIE, cls).suitable(url) and not parts.query
```
dstftw: This does not match https://rutube.ru/video/10b3a03fc01d5bbcc632a2f3514e8aab/?pl_id=4252.

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +202,66 @@ class RutubePersonIE(RutubeChannelIE):
...
+                entry.update({
+                    'id': result['id'],
+                    'uploader': try_get(result, lambda x: x['author']['name']),
+                    'uploader_id': try_get(result, lambda x: x['author']['id']),
+                    'upload_date': unified_strdate(result.get('created_ts')),
+                    'title': result.get('title'),
+                    'description': result.get('description'),
+                    'thumbnail': result.get('thumbnail_url'),
+                    'duration': int_or_none(result.get('duration')),
+                    'category': [try_get(result, lambda x: x['category']['name'])],
```
dstftw: This will result is `[None]` is no category extracted.

youtube_dl/extractor/rutube.py
```diff
@@ -193,3 +202,66 @@ class RutubePersonIE(RutubeChannelIE):
...
+                                       headers={'Accept': 'application/json'})
+
+            if not page['results']:
+                break
+
+            results = page['results']
+            for result in results:
+                entry = self.url_result(result.get('video_url'), 'Rutube')
+                entry.update({
+                    'id': result['id'],
```
dstftw: This will break extraction if no `id` present.

youtube_dl/extractor/rutube.py
luceatnobis: If you enter the URL, it doesn't display the playlist on the web interface; both `pl_id` and `pl_type=source` must be given. A URL with `pl_id` and without `pl_type=source` may never be given.

dstftw: https://rutube.ru/video/10b3a03fc01d5bbcc632a2f3514e8aab/?pl_id=4252 is **not a playlist** and it must be matched by non-playlist extractor.

luceatnobis:
Hey, can you give me an update on this PR?

dstftw:
```
PS C:\dev\youtube-dl\master> py -3.6 .\youtube_dl\__main__.py 'https://rutube.ru/video/10b3a03fc01d5bbcc632a2f3514e8aab/?pl_type=source' -F
[generic] ?pl_type=source: Requesting header
[generic] ?pl_type=source: Downloading webpage
[generic] ?pl_type=source: Extracting information
WARNING: [generic] ?pl_type=source: Failed to parse JSON Invalid control character at: line 6 column 122 (char 284)
ERROR: Unsupported URL: https://rutube.ru/video/10b3a03fc01d5bbcc632a2f3514e8aab/?pl_type=source
```

PR #13609, opened 2017-07-08
Always truncate files to 10241 bytes with --test

kmark:
This should prevent issues like that in #13449 where the FFmpeg
downloader was not reliably trimming to the correct length.

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Always truncate files to 10241 bytes with --test. This should prevent issues like that in #13449 where the FFmpeg downloader was not reliably trimming to the correct length.

### DIFF
--- a/youtube_dl/downloader/common.py
+++ b/youtube_dl/downloader/common.py
@@ -358,7 +358,14 @@ class FileDownloader(object):
                     else '%.2f' % sleep_interval))
             time.sleep(sleep_interval)
 
-        return self.real_download(filename, info_dict)
+        real_download_result = self.real_download(filename, info_dict)
+
+        if real_download_result and self.params.get('test', False):
+            f = open(filename, 'ab')
+            f.truncate(self._TEST_FILE_SIZE)
+            f.close()
+
+        return real_download_result
 
     def real_download(self, filename, info_dict):
         """Real download process. Redefine in subclasses."""


kmark:
Alternative patches: https://gist.github.com/kmark/e1d479cb36581f8af1bb3ffae0c7f83d

_Allow entire file to match or first 10241 bytes_ should be backwards compatible with existing tests. Since this PR isn't backwards compatible (as far as I know) I think this patch would best be applied instead of this PR if at all.

_Only MD5 the first 10241 bytes_ is more of a compliment to this PR.

youtube_dl/downloader/common.py
```diff
@@ -358,7 +358,14 @@ def download(self, filename, info_dict):
                     else '%.2f' % sleep_interval))
             time.sleep(sleep_interval)
 
-        return self.real_download(filename, info_dict)
+        real_download_result = self.real_download(filename, info_dict)
+
+        if real_download_result and self.params.get('test', False):
+            f = open(filename, 'ab')
```
dstftw: `sanitize_open`.

youtube_dl/downloader/common.py
```diff
@@ -358,7 +358,14 @@ def download(self, filename, info_dict):
...
+            f.truncate(self._TEST_FILE_SIZE)
```
dstftw: This will fill up to `_TEST_FILE_SIZE` is the file is `< _TEST_FILE_SIZE`.

dstftw: This will break all ffmpeg tests with md5 calculated prior to this change.

kmark: Indeed, I believe the _Allow entire file to match or first 10241 bytes_ may be more suitable?

kmark: Good catch, thanks. I must've been reading the wrong docs that claimed it would truncate to the given size or do nothing if less.

kmark:
Should I submit the _Allow entire file to match or first 10241 bytes_ patch as an alternative PR?

PR #13613, opened 2017-07-09
[vlive:playlist] Add new extractor

coreynicholson:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

A couple more examples of vlive playlists:
```
http://www.vlive.tv/video/24371/playlist/24384
http://www.vlive.tv/video/20928/playlist/20933
http://www.vlive.tv/video/16181/playlist/17079
```

The reason why there are two tests is because the webpages for the two are slightly different. The webpage for the video in the first test contains playlist information even if you remove the 'playlist' and playlist id segments of the URL. Whereas the second video doesn't contain playlist information if you remove the 'playlist' and playlist id segments of the URL therefore the playlist extractor would not work.

I've put this extractor before the regular video extractor in the extractors file because the playlist URLs match the regular video pattern. This way they both work.

### DIFF
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -1203,6 +1203,7 @@ from .vk import (
     VKWallPostIE,
 )
 from .vlive import (
+    VLivePlaylistIE,
     VLiveIE,
     VLiveChannelIE
 )
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -1203,9 +1203,9 @@ from .vk import (
     VKWallPostIE,
 )
 from .vlive import (
-    VLivePlaylistIE,
     VLiveIE,
-    VLiveChannelIE
+    VLiveChannelIE,
+    VLivePlaylistIE
 )
 from .vodlocker import VodlockerIE
 from .vodpl import VODPlIE
--- a/youtube_dl/extractor/vlive.py
+++ b/youtube_dl/extractor/vlive.py
@@ -292,16 +292,16 @@ class VLivePlaylistIE(InfoExtractor):
         assert video_id_match
         video_id = compat_str(video_id_match.group('video_id'))
 
-        video_url_format = 'http://www.vlive.tv/video/%s'
+        VIDEO_URL_TEMPLATE = 'http://www.vlive.tv/video/%s'
         if self._downloader.params.get('noplaylist'):
             self.to_screen(
                 'Downloading just video %s because of --no-playlist' % video_id)
             return self.url_result(
-                video_url_format % video_id,
+                VIDEO_URL_TEMPLATE % video_id,
                 ie=VLiveIE.ie_key(), video_id=video_id)
-        else:
-            self.to_screen(
-                'Downloading playlist %s - add --no-playlist to just download video' % playlist_id)
+
+        self.to_screen(
+            'Downloading playlist %s - add --no-playlist to just download video' % playlist_id)
 
         webpage = self._download_webpage(
             'http://www.vlive.tv/video/%s/playlist/%s' % (video_id, playlist_id), video_id)
@@ -319,7 +319,7 @@ class VLivePlaylistIE(InfoExtractor):
             item_id = compat_str(item_id)
             entries.append(
                 self.url_result(
-                    video_url_format % item_id,
+                    VIDEO_URL_TEMPLATE % item_id,
                     ie=VLiveIE.ie_key(), video_id=item_id))
 
         return self.playlist_result(
--- a/youtube_dl/extractor/vlive.py
+++ b/youtube_dl/extractor/vlive.py
@@ -261,3 +261,50 @@ class VLiveChannelIE(InfoExtractor):
 
         return self.playlist_result(
             entries, channel_code, channel_name)
+
+
+class VLivePlaylistIE(InfoExtractor):
+    IE_NAME = 'vlive:playlist'
+    _VALID_URL = r'https?://(?:(?:www|m)\.)?vlive\.tv/video/(?P<video_id>[0-9]+)/playlist/(?P<id>[0-9]+)'
+    _TESTS = [{
+        'url': 'http://www.vlive.tv/video/30824/playlist/30826',
+        'info_dict': {
+            'id': '30826',
+            'title': 'TWICE TV5 - TWICE in SWITZERLAND'
+        },
+        'playlist_mincount': 20
+    }, {
+        'url': 'http://www.vlive.tv/video/22867/playlist/22912',
+        'info_dict': {
+            'id': '22912',
+            'title': 'Valentine Day Message from TWICE'
+        },
+        'playlist_mincount': 9
+    }]
+
+    def _real_extract(self, url):
+        playlist_id = self._match_id(url)
+        video_id = self._search_regex(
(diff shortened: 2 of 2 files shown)


coreynicholson:
Looks like the builds are failing because multiple extractors are not allowed to match the same URL, is there any way around this without making the regular VLive video URL pattern restrictive?

dstftw:
Override `suitable`.

coreynicholson:
Made the change, the builds don't seem to be failing because of this extractor anymore.

youtube_dl/extractor/vlive.py
```diff
@@ -261,3 +265,50 @@ def _real_extract(self, url):
...
+            'id': '22912',
+            'title': 'Valentine Day Message from TWICE'
+        },
+        'playlist_mincount': 9
+    }]
+
+    def _real_extract(self, url):
+        playlist_id = self._match_id(url)
+        video_id = self._search_regex(
+            self._VALID_URL, url, 'video id', group='video_id')
```
dstftw: Use bare `re.match`.

youtube_dl/extractor/vlive.py
```diff
@@ -261,3 +265,50 @@ def _real_extract(self, url):
...
+
+        webpage = self._download_webpage(
+            'http://www.vlive.tv/video/%s/playlist/%s' % (video_id, playlist_id), video_id)
+
+        playlist_name = self._html_search_regex(
+            r'<div[^>]+class="[^"]*multicam_playlist[^>]*>\s*<h3[^>]+>([^<]+)',
+            webpage, 'playlist name', fatal=False)
+
+        item_ids = self._search_regex(
+            r'\bvar\s+playlistVideoSeqs\s*=\s*\[([^\]]+)\]',
```
dstftw: No need to escape `]` is character set.

youtube_dl/extractor/vlive.py
```diff
@@ -261,3 +265,50 @@ def _real_extract(self, url):
...
+            webpage, 'playlist item ids', default='')
```
dstftw: `default` is senseless here.

youtube_dl/extractor/vlive.py
```diff
@@ -261,3 +265,50 @@ def _real_extract(self, url):
...
+
+        entries = []
+        for item_id in re.split(r'\s*,\s*', item_ids):
```
dstftw: `_parse_json` instead.

youtube_dl/extractor/vlive.py
```diff
@@ -261,3 +265,50 @@ def _real_extract(self, url):
...
+
+    def _real_extract(self, url):
+        playlist_id = self._match_id(url)
+        video_id = self._search_regex(
+            self._VALID_URL, url, 'video id', group='video_id')
+
+        webpage = self._download_webpage(
+            'http://www.vlive.tv/video/%s/playlist/%s' % (video_id, playlist_id), video_id)
+
+        playlist_name = self._html_search_regex(
```
dstftw: `--no-playlist` is not respected.

youtube_dl/extractor/extractors.py
```diff
@@ -1203,6 +1203,7 @@
     VKWallPostIE,
 )
 from .vlive import (
+    VLivePlaylistIE,
```
dstftw: Not alphabetic.

coreynicholson:
Thanks for your feedback, I've pushed another commit where I've tried to address your comments. Let me know if there's still something you would like me to improve.

youtube_dl/extractor/vlive.py
```diff
@@ -261,3 +265,62 @@ def _real_extract(self, url):
...
+            r'<div[^>]+class="[^"]*multicam_playlist[^>]*>\s*<h3[^>]+>([^<]+)',
+            webpage, 'playlist name', fatal=False)
+
+        item_ids = self._search_regex(
+            r'\bvar\s+playlistVideoSeqs\s*=\s*\[([^]]+)\]',
+            webpage, 'playlist item ids')
+
+        entries = []
+        for item_id in self._parse_json('[%s]' % item_ids, playlist_id):
```
dstftw: You already have `[]` in regex, capture it and parse there.

coreynicholson: Good point, updated.

youtube_dl/extractor/vlive.py
```diff
@@ -261,3 +265,62 @@ def _real_extract(self, url):
...
+        'playlist_mincount': 9
+    }]
+
+    def _real_extract(self, url):
+        playlist_id = self._match_id(url)
+        video_id_match = re.match(self._VALID_URL, url)
+        assert video_id_match
+        video_id = compat_str(video_id_match.group('video_id'))
+
+        video_url_format = 'http://www.vlive.tv/video/%s'
```
dstftw: Uppercase is used for constants. Also this is template not format.

youtube_dl/extractor/vlive.py
```diff
@@ -261,3 +265,62 @@ def _real_extract(self, url):
...
+        if self._downloader.params.get('noplaylist'):
+            self.to_screen(
+                'Downloading just video %s because of --no-playlist' % video_id)
+            return self.url_result(
+                video_url_format % video_id,
+                ie=VLiveIE.ie_key(), video_id=video_id)
+        else:
```
dstftw: `else` is superfluous.

coreynicholson:
Thanks, I've changed the variable name for the constant to uppercase and removed the unnecessary `else`.

youtube_dl/extractor/vlive.py
```diff
@@ -261,3 +265,62 @@ def _real_extract(self, url):
...
+    _VALID_URL = r'https?://(?:(?:www|m)\.)?vlive\.tv/video/(?P<video_id>[0-9]+)/playlist/(?P<id>[0-9]+)'
+    _TESTS = [{
+        'url': 'http://www.vlive.tv/video/30824/playlist/30826',
+        'info_dict': {
+            'id': '30826',
+            'title': 'TWICE TV5 - TWICE in SWITZERLAND'
+        },
+        'playlist_mincount': 20
+    }, {
+        'url': 'http://www.vlive.tv/video/22867/playlist/22912',
```
dstftw: Make this test `only_matching` and add a comment why it even exists.

coreynicholson: If I make it `only_matching` does that mean it will only test that the URL matches the regex? If so, that defeats the point of that test. The URL in the second test links to a video which appears to be linked to the playlist somewhat differently to the URL in the first test.

From my observations, if someone were to download the webpage using only the video id from the given URL, it would not link to the playlist, whereas it would with the URL in the first test. Therefore I think both tests should perform the full webpage download.

i.e:
`http://www.vlive.tv/video/30824` the HTML for this page does contain the playlist
`http://www.vlive.tv/video/22867` the HTML for this page does not

I can add a code comment to clarify this if my assumption about `only_matching` is correct.

dstftw: Both test the same extraction scenario. The rule is one test per extraction scenario.

coreynicholson: Then I think it'll be better to keep the second test and remove the first one entirely, I'll do that.

coreynicholson:
I've pushed the change to remove a test so there's only one.

PR #13664, opened 2017-07-16
Add a filename key in info_dict

Zil0:

---

- [ ] Bug fix
- [x ] Improvement
- [ ] New extractor
- [ ] New feature

---

This allows to easily retrieve the filenames of the downloaded videos, for example when calling extract_info.
Context :
When embedding youtube-dl, I was faced with the issue of knowing what was actually downloaded in order to run some of my own code on it. Sadly, the method prepare_filename only works as expected  when there is no post-processing involved. The only way I found to circumvent this was to parse the logger messages, which is obviously unsatisfying, convoluted and unmaintainable. I asked on IRC if there was another way but got no answer. Hence, here is a simple fix.


Zil0:
Hi, any chance this gets looked at ? I get that embedding ydl is merely a side effect of it being written in Python, with no functionality specifically targeting external use, but this one feels very basic to me...

dstftw:
That's not how it's work. First, you must propose and document a format that will allow unambiguously identify each file. What you have now is a list of filenames without any indication which one is a final file, a subtitle, a thumbnail and so on.

Zil0:
Thanks a lot for your insight. However I'm not sure you understood where I was coming from. What I have is a list of video and audio filenames only. Moreover, it is easy to know which one is the final file (it's the last one), though I absolutely agree doc is needed. What would be the right place ? An addition in the example of the README maybe ?
Also, I believe the issue I want to address is quite different from what you suggested about subtitles and thumbnails filenames. I can always hack my way around determining those, which is not the case of video and audio. Hence what I propose could be seen as new functionality (vs just a convenience feature).
Allow me to illustrate this. After downloading the video from a python script with extract_info, I get a dict with a ton of infos. Now, I want to know exactly what have been downloaded. Assuming I requested them, I can get the thumbnail by doing info_dict['thumbnails']['filename'] (exactly the functionality I'm proposing for vids, by the way), I can get the description by changing the extension in ydl.prepare_filename(info_dict), same for subtitles and I believe all the other writable infos. But what if the audio and video format were incompatible ? Then I have a .mkv, and prepare_info in wrong. There is no way to know when this occurs, because the relevant code is private (a function inside another). Same when any post-processor is ran : it's impossible to know the extension.

dstftw:
>What I have is a list of video and audio filenames only

Wrong. With `keepvideo` you have a list of all files reported by post processors. This may include subtitles, thumbnails and so on.

>What would be the right place ? An addition in the example of the README maybe ?

Right place is the place where the info dict format is documented.

>what I propose could be seen as new functionality (vs just a convenience feature)

That's why it should be carefully unambiguously designed and documented.

>get the thumbnail by doing info_dict['thumbnails']['filename']

Basically, there is no such functionality, it's not documented and serves only for thumbnail embed post processor. It's just a leftover after its work and may be removed any time.

I'm well aware of the problem with [final media extension](https://github.com/rg3/youtube-dl/pull/5710). No need to to repeat it over and over again.

The point here is that proposed solution creates more mess and ambiguity and needs proper design first.

Zil0:
Thanks for bearing with me.
The folks on IRC not understanding the issue genuinely made me think I was the only one to be bothered. Now, turns out my first commit fixes the issue of final media extension, in a not so complicated way (a list of strings corresponding to each requested formats, which I will document in extractors/common.py). A new option like --get-final-filename could be easily added.
The problem you highlighted arises with my second commit. You're obviously right, the solution then becomes messy with a list of list of filenames.
What you suggested was more or less to replace the "hold every filenames" inner list by a dict which keys would be file types like "video", "audio", "thumbnail" and link to more coherent lists. I found no simple way to achieve this : the easiest one may be to have a data structure holding every possible extension corresponding to each file type, which ends up being dirty and hard to maintain. A satisfying solution involves heavily modifying the code to keep track of each file's type from the moment they are downloaded until they leave post-processors. Maybe I can implement this, but I'd like to read what you think first.
Or, I can drop that second commit which provides a feature nobody requested !

Zil0:
So, would dropping the second commit solve the problem of final media extension in an acceptable way ?
It doesn't seem irrelevant to handle and discuss the problem of transitional filenames saving in a different issue/PR.

PR #13669, opened 2017-07-17
[build] Override timestamps in zip file

bmwiedemann:
to make build reproducible.
See https://reproducible-builds.org/ for why this is good.

Copying files to not interfere with freshness detection.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,8 +46,15 @@ tar: youtube-dl.tar.gz
 pypi-files: youtube-dl.bash-completion README.txt youtube-dl.1 youtube-dl.fish
 
 youtube-dl: youtube_dl/*.py youtube_dl/*/*.py
-	zip --quiet youtube-dl youtube_dl/*.py youtube_dl/*/*.py
-	zip --quiet --junk-paths youtube-dl youtube_dl/__main__.py
+	mkdir -p zip
+	for d in youtube_dl youtube_dl/downloader youtube_dl/extractor youtube_dl/postprocessor ; do \
+	  mkdir -p zip/$$d ;\
+	  cp -a $$d/*.py zip/$$d/ ;\
+	done
+	touch -t 200001010101 zip/youtube_dl/*.py zip/youtube_dl/*/*.py
+	mv zip/youtube_dl/__main__.py zip/
+	cd zip ; zip --quiet ../youtube-dl youtube_dl/*.py youtube_dl/*/*.py __main__.py
+	rm -rf zip
 	echo '#!$(PYTHON)' > youtube-dl
 	cat youtube-dl.zip >> youtube-dl
 	rm youtube-dl.zip


Makefile
```diff
@@ -46,8 +46,12 @@ tar: youtube-dl.tar.gz
 pypi-files: youtube-dl.bash-completion README.txt youtube-dl.1 youtube-dl.fish
 
 youtube-dl: youtube_dl/*.py youtube_dl/*/*.py
-	zip --quiet youtube-dl youtube_dl/*.py youtube_dl/*/*.py
-	zip --quiet --junk-paths youtube-dl youtube_dl/__main__.py
+	mkdir -p zip
+	cp -a --parent youtube_dl/*.py youtube_dl/*/*.py zip/
+	touch -d@946684800 zip/youtube_dl/*.py zip/youtube_dl/*/*.py
```
yan12125: The timestamp shouldn't be hard-coded. ```SOURCE_DATE_EPOCH``` environment variable seems a good choice. https://reproducible-builds.org/docs/timestamps/

bmwiedemann: I was initially setting it to 0 because it (hopefully) does not matter for anything and SOURCE_DATE_EPOCH is not set in all build environments.

yan12125: Build systems aiming on reproducible binaries should always set ```SOURCE_DATE_EPOCH```, and for other build systems, using a fixed timestamp here does not guarantee reproducible build results.

bmwiedemann:
updated to use `$SOURCE_DATE_EPOCH`

yan12125:
Thanks. The only remaining issue is that the build fails without ```$SOURCE_DATE_EPOCH```
```
$ make youtube-dl
mkdir -p zip
cp -a --parent youtube_dl/*.py youtube_dl/*/*.py zip/
[ -n "" ] && touch -d@ zip/youtube_dl/*.py zip/youtube_dl/*/*.py
make: *** [Makefile:51: youtube-dl] Error 1
```
Tested on zsh 5.3.1 and bash 4.4.12

bmwiedemann:
hopefully fixed.

yan12125:
```--parent``` is not available for BSD-flavored ```cp```. For example on macOS Sierra:
```
$ make youtube-dl
mkdir -p zip
cp -a --parent youtube_dl/*.py youtube_dl/*/*.py zip/
cp: illegal option -- -
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvX] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvX] source_file ... target_directory
make: *** [youtube-dl] Error 64
```

**Comment by bmwiedemann** [comment]
updated it to emulate cp --parents behaviour

**Comment by yan12125** [comment]
Thanks for the update. However, things are not completely solved. On macOS touch does not have `-d`:
```
$ SOURCE_DATE_EPOCH=123 make -B youtube-dl
mkdir -p zip
for d in youtube_dl youtube_dl/downloader youtube_dl/extractor youtube_dl/postprocessor ; do \
	  mkdir -p zip/$d ;\
	  cp -a $d/*.py zip/$d/ ;\
	done
[ -z "123" ] || touch -d@123 zip/youtube_dl/*.py zip/youtube_dl/*/*.py
touch: illegal option -- d
usage:
touch [-A [-][[hh]mm]SS] [-acfhm] [-r file] [-t [[CC]YY]MMDDhhmm[.SS]] file ...
make: *** [youtube-dl] Error 1
```
I guess ```-t``` provides a similar functionality.

bmwiedemann:
changed it to use a constant timestamp, because 
a) it should not matter and
b) it is a lot easier this way than supporting the different GNU and BSD date versions

yan12125:
The latest patch builds fine on both macOS and Linux. But, is it a good practice to use fixed timestamps for reproducible builds?

bmwiedemann:
If information does not matter, the preferred way is to not store it at all or to at least have a fixed value.

If you want to have some more meaningful timestamp in there (why?) the way to go would be
https://wiki.debian.org/ReproducibleBuilds/TimestampsProposal#Bash_.2F_POSIX_shell

yan12125:
Got it. Thanks for the info!

PR #13695, opened 2017-07-21
[npo] added npo3 support

dubber0:

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

Some NPO videos are also available at npo3.nl, I've added support for that URL and wrote a test for it. The videos at npo3.nl can be downloaded exactly the same way as on npo.nl, so I only had to add a regex. npo3.nl is an official site of NPO.

### DIFF
--- a/youtube_dl/extractor/npo.py
+++ b/youtube_dl/extractor/npo.py
@@ -28,7 +28,7 @@ class NPOBaseIE(InfoExtractor):
 
 class NPOIE(NPOBaseIE):
     IE_NAME = 'npo'
-    IE_DESC = 'npo.nl, npo3.nl and ntr.nl'
+    IE_DESC = 'npo.nl, ntr.nl, omroepwnl.nl, zapp.nl and npo3.nl'
     _VALID_URL = r'''(?x)
                     (?:
                         npo:|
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/npo.py
```diff
@@ -28,14 +28,15 @@ def _get_token(self, video_id):
...
-    IE_DESC = 'npo.nl and ntr.nl'
+    IE_DESC = 'npo.nl, npo3.nl and ntr.nl'
     _VALID_URL = r'''(?x)
                     (?:
                         npo:|
                         https?://
                             (?:www\.)?
                             (?:
                                 npo\.nl/(?!(?:live|radio)/)(?:[^/]+/){2}|
+                                npo3\.nl/(?!/)(?:[^/]+/){2}|
```
dstftw: `(?!/)` makes no sense.

youtube_dl/extractor/npo.py
```diff
@@ -28,14 +28,15 @@ def _get_token(self, video_id):
...
                                 ntr\.nl/(?:[^/]+/){2,}|
```
dstftw: New domain should be embedded here.

youtube_dl/extractor/npo.py
```diff
@@ -89,6 +90,17 @@ class NPOIE(NPOBaseIE):
...
+        'url': 'https://www.npo3.nl/3onderzoekt/16-09-2015/VPWON_1239870',
+        'md5': '5196e6b895a779445a610f1f4c4145e7',
+        'info_dict': {
+            'id': 'VPWON_1239870',
+            'ext': 'm4v',
+            'title': '3Onderzoekt: Krakers',
+            'description': 'md5:0f79d45e282659ba4be99b125196b6af',
+            'upload_date': '20150916',
+            'duration': 1514,
+        },
```
dstftw: Make `only_matching`.

dubber0:
What do you mean with embedded?

PR #13752, opened 2017-07-27
[Enhancement] Extended Date/Time Output Templating Options

kade-robertson:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

The purpose of the pull request is to add a more flexible output for organizing very specifically by the date and/or time of the upload. The string formatting available is fairly limited, so numerous extensions were added to include the option of using the:

- year
- month (numeric and localized string, long or short form)
- day (numeric and localized string, long or short form)
- hour
- minute
- second
- AM/PM indicator

The latter four do require a timestamp value to be present to provide *useful* data, however they will not  cause errors if only the upload date is present (they will just be their defaults). Two extra output options were added for the full time string in 24 or 12-hour format. In the event that neither are present these values are simply ignored so this should be future-proof as long as `upload_date`'s format doesn't change, and if it does it's a fix on a single line. The names were chosen to try and be as descriptive as possible while keeping short enough to fit reasonably in a line, and to not overlap as much as possible.

The desire for such a change was for easy archiving of many channels across long timespans, which is now possible if, for example, a template like this is used:

    %(uploader)s/%(upload_year)s/%(upload_month)s-%(upload_mntn)s/%(upload_day)s/%(upload_time24)s-%(title)s-%(id)s.%(ext)s


dstftw:
Instead of introducing enormous number of new meta fields `upload_date` and `timestamp` (as well as all other date meta fields) should support `strftime`-like syntax in output template.

kade-robertson:
How would you suggest this be implemented with the current string formatting procedure? Some sort of pre-processing of the template?

dstftw:
There is no way to do this without preprocessing.

kade-robertson:
Hm, I'm not really sure of what an ideal clean solution would look like, what if there was an exclamation delimiter like `upload_date!%A` which indicated the string should be pre-processed to be `strftime` format. I don't really know the best solution for this because I would imagine it should also comply afterwards with the user's formatting, i.e. if they just want the first four letters of the output they should still be able to do `%(upload_date!%H%M%s).4s` right?

It would be relatively simple to use a regex replacement sort of setup but it wouldn't be compatible with the above, something like `r'%\((upload_date|timestamp)!(%.*)+?\)s'`

dstftw:
The general idea is to allow something like `%(field[:<field_specific_formatter>])<standard_formatter>`.

cferrey:
Apologies if I'm violating a rule by posting to this thread, but could someone help me with the syntax for calling this feature?  I've tried every iteration I can think of, but keep getting "NA" or "ERROR: not enough arguments for format string" depending on where I try to put the formatting arguments.

I am hoping to call %(upload_date)s three times for %Y, %M, %D and delimit each output with a period, in order to turn 20190108 into 2019.01.08

Thanks!

dstftw:
No such feature exists.

cferrey:
Got it -- thank you for the info. Are there any plans to implement the above?  If not, is there any other way I can accomplish this with youtube-dl alone, or do I need to rename after the fact?

Issue #13840, opened 2017-08-06
[niconico] "Unable to extract video title" on some videos with numeric-only ID-s


Xender:
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2017.08.06**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

```
$ python -m youtube_dl -F http://www.nicovideo.jp/watch/1311771889 --verbose --write-pages

[debug] System config: []
[debug] User config: ['-o', '%(title)s-%(format)s[v=%(id)s&fmt=%(format_id)s].%(ext)s', '-n']
[debug] Custom config: []
[debug] Command-line args: ['-F', 'http://www.nicovideo.jp/watch/1311771889', '--verbose', '--write-pages']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2017.08.06
[debug] Git HEAD: 903a183b6
[debug] Python version 3.6.2 - Linux-4.12.3-1-ARCH-x86_64-with-arch
[debug] exe versions: ffmpeg 3.3.2, ffprobe 3.3.2, rtmpdump 2.4
[debug] Proxy map: {}
[niconico] Logging in
[niconico] 1311771889: Downloading webpage
[niconico] Saving request to 1311771889_http_-_www.nicovideo.jp_watch_1311771889.dump
[niconico] 1311771889: Downloading video info page
[niconico] Saving request to 1311771889_http_-_ext.nicovideo.jp_api_getthumbinfo_1311771889.dump
[niconico] 1311771889: Downloading flv info
[niconico] Saving request to 1311771889_http_-_flapi.nicovideo.jp_api_getflv_1311771889as3=1.dump
[...]

PR #13843, opened 2017-08-06
[niconico] Update fallback video title regex

Xender:
(Eh, see the diff first, it's shorter than description :P)

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

I've updated the fallback regex for extracting the video title, as mentioned in #13840 (it was not the main point of that issue).

Actually this code path seems to be now like a fallback-of-fallback-of-fallback and may not
be reached by anyone ever again, nevertheless if it has not been deleted entirely, it's probably a good idea to keep it up-to-date.

@yan12125 I'm mentioning you here since you are the author of most recent updates to Niconico extractor.

### DIFF
--- a/youtube_dl/extractor/niconico.py
+++ b/youtube_dl/extractor/niconico.py
@@ -340,7 +340,7 @@ class NiconicoIE(InfoExtractor):
             title = self._og_search_title(webpage, default=None)
         if not title:
             title = self._html_search_regex(
-                r'<span[^>]+class="videoHeaderTitle"[^>]*>([^<]+)</span>',
+                r'<h1[^>]+class="txt-title"[^>]*>([^<]+)</h1>',
                 webpage, 'video title')
 
         watch_api_data_string = self._html_search_regex(


yan12125:
```txt-title``` is used only if the user is not logged in? It can be deleted as NicoNico won't work without login :)

Xender:
@yan12125 Okay, I re-checked things and this PR is basically invalid.
Please don't merge.

Conclusions:
- Nico is more messed up than I thought. It serves **three** different version of "watch" page:
  - One for unauthenticated user has the `<h1 itemprop="name" class="txt-title">[title]</h1>`  tag. As you said, Nico requires login anyway so this case is a dead end.
  - One for logged-in user using HTML5 player doesn't have that tag in server HTML, but instead it has the `<div id="js-initial-watch-data">` which is now parsed.
  - One for logged-in user using the old Flash player still has the `<span class="videoHeaderTitle">` tag.
    - Additionally, it has some JSON-s in contents of some `display:none` div-s like `<div id="watchAPIDataContainer">` (and `wallDataContainer`, `playlistDataContainer`, `configDataContainer` (this one is a `<script>` actually :D)).

Now I think that the page displayed for unauthenticated visits may be a kind of pre-rendered version of HTML5 player page, for the use of web spiders (which would explain presence of both OpenGraph meta tags and visible tags with actual content in there).

PR #13871, opened 2017-08-09
Update README.md

siddht4:
added  Linux file permission changes from sudo aka root to  user

This is just a minor improvement/change  to how youtube-dl is set under linux

1. Download youtube-dl to an root only space. [default]
2. Make the youtube-dl installed in the root space as  executable and writable. [default]
3. change the user permission and user group from root:root  to [user]:[user-group] . [Added]

This change actually allowed me to update youtube-dl without providing sudo or su priveleges everytime.

fixes   https://github.com/rg3/youtube-dl/issues/11046

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,22 @@ If you do not have curl, you can alternatively use a recent wget:
 
     sudo wget https://yt-dl.org/downloads/latest/youtube-dl -O /usr/local/bin/youtube-dl
     sudo chmod a+rx /usr/local/bin/youtube-dl
+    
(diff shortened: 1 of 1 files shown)


dstftw:
You should not change the owner of anything under `/usr/bin/local` while in multi-user environment that is the default.

siddht4:
Okay got your point,it is just for a single user system like a personal system.otherwise fakeroot should be a choice

siddht4:
I am not changing any other file permissions for obvious security reason,this was just for youtube-dl python binary

PR #14007, opened 2017-08-24
[kakao] [All issues fixed] Add new extractor

Namnamseo:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

#### Background story

Two Korean companies, [Daum](https://www.daum.net) and [Kakao](http://www.kakaocorp.com), was merged in October 1st, 2014.
Since then, some services of Daum have been closed or changed.

'tvpot'(TV팟), a video sharing service previously served by Daum, is no longer allowing any other activities than watching a video. Instead, Kakao has launched a (Beta) service named 'Kakao TV'(카카오TV).

Now, users can migrate their videos that were previously uploaded in tvpot to Kakao TV. For example, [this video in tvpot](http://tvpot.daum.net/v/v4eeefwPzwww1OHvUzPGvzw) has its corresponding one in [Kakao TV](http://tv.kakao.com/channel/2671005/cliplink/301965083). This means that the owner of the video agreed on migrating the videos.

The reason why they don't migrate it themselves, is that Kakao accounts are different from Daum account (remember that they were two different companies). I feel like they're not trying to just bomb up the old service; thye're encouraging users to keep their videos alive. It has been announced in Feb, 2017 that tvpot is going to be closed someday, but the exact date is not announced yet -- and 6 months have passed. Well... who knows?

#### About this pull request

I've implemented an extractor for Kakao TV videos. This can download videos of the form `tv.kakao.com/channel/(numbers)/cliplink/(numbers)`. Some URLs have the form `tv.kakao.com/v/s2c98ruQGIkQpupf33RGpRl`, but the server appropriately redirects with 302 to the URL in the former form. I've tested that the generic(redirect) extractor works well.

Currently, the JavaScript is not much obfuscated, and it was quite easy to look through how it works:

1. The HTML5 player requests an *impress* of the video. The *impress* contains various informations about the video, including the uploader(here called *channel*), video duration, thumbnails, and a parameter called `tid`.
2. Then, the player requests an *raw* of the video. The *raw* contains the URL of the video file, and a list of video qualities.

`flake8` passed.

Please let me know of any problems in the PR. Thank you.

### DIFF
--- a/youtube_dl/extractor/kakao.py
+++ b/youtube_dl/extractor/kakao.py
@@ -22,6 +22,7 @@ class KakaoIE(InfoExtractor):
             'title': '乃木坂46 バナナマン 「3期生紹介コーナーが始動！顔高低差GPも！」 『乃木坂工事中』',
             'uploader_id': 2671005,
             'uploader': '그랑그랑이',
+            'upload_date': '20170227'
         }
     }, {
         'url': 'http://tv.kakao.com/channel/2653210/cliplink/300103180',
@@ -33,6 +34,7 @@ class KakaoIE(InfoExtractor):
             'title': '러블리즈 - Destiny (나의 지구) (Lovelyz - Destiny)',
             'uploader_id': 2653210,
             'uploader': '쇼 음악중심',
+            'upload_date': '20170129'
         }
     }]
 
--- a/youtube_dl/extractor/kakao.py
+++ b/youtube_dl/extractor/kakao.py
@@ -1,0 +1,138 @@
+# coding: utf-8
+
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import (
+    int_or_none,
+)
+import datetime
+
+
+class KakaoIE(InfoExtractor):
+    _VALID_URL = r'https?://tv.kakao.com/channel/(?P<channel>\d+)/cliplink/(?P<id>\d+)'
+    IE_NAME = 'kakao.com'
+
+    _TESTS = [{
+        'url': 'http://tv.kakao.com/channel/2671005/cliplink/301965083',
+        'md5': '702b2fbdeb51ad82f5c904e8c0766340',
+        'info_dict': {
+            'id': '301965083',
+            'ext': 'mp4',
+            'title': '乃木坂46 バナナマン 「3期生紹介コーナーが始動！顔高低差GPも！」 『乃木坂工事中』',
+            'uploader_id': 2671005,
+            'uploader': '그랑그랑이',
+        }
+    }, {
+        'url': 'http://tv.kakao.com/channel/2653210/cliplink/300103180',
+        'md5': 'a8917742069a4dd442516b86e7d66529',
+        'info_dict': {
+            'id': '300103180',
+            'ext': 'mp4',
+            'description': '러블리즈 - Destiny (나의 지구) (Lovelyz - Destiny)\r\n\r\n[쇼! 음악중심] 20160611, 507회',
+            'title': '러블리즈 - Destiny (나의 지구) (Lovelyz - Destiny)',
+            'uploader_id': 2653210,
+            'uploader': '쇼 음악중심',
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        # Player URL, to be used in Referer header
+        player_url = 'http://tv.kakao.com/embed/player/cliplink/' + video_id + \
+            '?service=kakao_tv&autoplay=1&profile=HIGH&wmode=transparent'
+        player_header = {'Referer': player_url}
+
+        # Request Impress, which contains video information
+        impress = self._download_json(
+            'http://tv.kakao.com/api/v1/ft/cliplinks/%s/impress' % video_id,
+            video_id, 'Downloading video info',
+            query={
+                'player': 'monet_html5',
+                'referer': url,
+                'uuid': '',
+                'service': 'kakao_tv',
+                'section': '',
+                'dteType': 'PC',
+                'fields': 'clipLink,clip,channel,hasPlusFriend,-service,-tagList'
+            }, headers=player_header)
+
+        clipLink = impress['clipLink']
+
+        # Now we request Raw, which contains infos about video files.
+        tid = impress.get('tid', '')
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/kakao.py
```diff
@@ -0,0 +1,140 @@
...
+                'height': int_or_none(fmt.get('height')),
+                'format_note': fmt.get('label', None),
+                'filesize': int_or_none(fmt.get('filesize'))
+            })
+
+        self._sort_formats(formats)
+
+        clip = clipLink['clip']
+        # Parse thumbnails.
+        top_thumbnail = clip.get('thumbnailUrl', None)
```
dstftw: `None` is default.

youtube_dl/extractor/kakao.py
```diff
@@ -0,0 +1,140 @@
...
+        thumbs = []
+        for thumb in clip.get('clipChapterThumbnailList', []):
+            thumbs.append({
+                'url': thumb['thumbnailUrl'],
+                'id': str(thumb['timeInSec']),
```
dstftw: `compat_str`.

youtube_dl/extractor/kakao.py
```diff
@@ -0,0 +1,140 @@
...
+                'profile': 'HIGH',
+                'dteType': 'PC',
+            }, headers=player_header)
+
+        formats = []
+        for fmt in raw['outputList']:
+            profile_name = fmt['profile']
+            # The following request is called when user changes the video quality.
+            # We simulate it here.
+            fmt_url_json = self._download_json(
```
dstftw: Should not be fatal.

youtube_dl/extractor/kakao.py
```diff
@@ -0,0 +1,140 @@
...
+            thumbs.append({
+                'url': thumb['thumbnailUrl'],
+                'id': str(thumb['timeInSec']),
+                'preference': -1 if thumb['isDefault'] else 0
+            })
+        # Parse upload date.
+        upload_date = None
+        try:
+            upload_date = datetime.datetime.strptime(clipLink['createTime'], '%Y-%m-%d %H:%M:%S')
+            upload_date = upload_date.strftime('%Y%m%d')
```
dstftw: `unified_timestamp`.

youtube_dl/extractor/kakao.py
```diff
@@ -0,0 +1,140 @@
...
+            pass
+
+        return {
+            'id': video_id,
+            'title': clip['title'],
+            'formats': formats,
+            'thumbnail': top_thumbnail,
+            'thumbnails': thumbs,
+            'description': clip.get('description'),
+            'uploader': clipLink['channel'].get('name'),
```
dstftw: Optional fields should not break extraction if missing.

youtube_dl/extractor/kakao.py
```diff
@@ -0,0 +1,140 @@
...
+        tid = impress.get('tid', '')
+        raw = self._download_json(
+            'http://tv.kakao.com/api/v1/ft/cliplinks/%s/raw' % video_id,
+            video_id, 'Downloading video formats info',
+            query={
+                'player': 'monet_html5',
+                'referer': url,
+                'uuid': '',
+                'service': 'kakao_tv',
+                'section': '',
```
dstftw: 54-58, 71-75 code duplication.

dstftw:
Read

and fix all issues.

youtube_dl/extractor/kakao.py
Namnamseo: Though they really are, the parameters were hard-coded in the JavaScript. In other words, they need not be any same, or share some of the parameters.
Would it be better to make an (unexplainable) dict of the unknowingly same parameters and append some new parameters for each query?

youtube_dl/extractor/kakao.py
```diff
@@ -0,0 +1,142 @@
...
+                fmt_url_json = self._download_json(
+                    'http://tv.kakao.com/api/v1/ft/cliplinks/%s/raw/videolocation' % video_id,
+                    video_id, 'Downloading video URL for profile %s' % profile_name,
+                    query={
+                        'service': 'kakao_tv',
+                        'section': '',
+                        'tid': tid,
+                        'profile': profile_name
+                    }, headers=player_header, fatal=False)
+                fmt_url = fmt_url_json['url']
```
dstftw: `fmt_url_json` may be `None`.

youtube_dl/extractor/kakao.py
```diff
@@ -0,0 +1,142 @@
...
+                    'format_note': fmt.get('label'),
+                    'filesize': int_or_none(fmt.get('filesize'))
+                })
+            except KeyError:
+                pass
+
+        self._sort_formats(formats)
+
+        clip = clipLink['clip']
+        # Parse thumbnails.
```
dstftw: Remove all pointless comments.

youtube_dl/extractor/kakao.py
```diff
@@ -0,0 +1,142 @@
...
+                'id': compat_str(thumb.get('timeInSec')),
+                'preference': -1 if thumb.get('isDefault') else 0
+            })
+
+        # Parse upload date.
+        upload_date = unified_timestamp(clipLink.get('createTime'))
+
+        return {
+            'id': video_id,
+            'title': clip['title'],
```
dstftw: This should be extracted the very first.

youtube_dl/extractor/kakao.py
Namnamseo: I think it's covered by `try`-`catch` clause.

dstftw: No. `KeyError` != `TypeError`.

Namnamseo: Oh, I got it. By the way, is it desirable to use the clause in this case?

dstftw: It's fine to use it until it produces longer code. You should also limit `try-catch` section to the code it's actually applies for.

Namnamseo:
All issues fixed.

dstftw:

Namnamseo:
I would be honored to.
Under "Seungwon Shin", please.

PR #14093, opened 2017-09-01
[soundcloud] Fix download URL with private tracks

ishitatsuyuki:

- [x] I am the original author of this code (*modification*) and I am willing to release it under [Unlicense](http://unlicense.org/)

- [x] Bug fix

---

The extractor was missing secret_token when fetching download_url, which made it invalid and yielding 404. This changeset includes some level of refactor to avoid duplication.

The "download" format name was changed into "download_url" because the error message was too confusing when it failed:
```
[soundcloud] 340344461: Checking download video format URL
[soundcloud] 340344461: download video format URL is invalid, skipping
```

A link is also provided for testing: https://soundcloud.com/oriuplift/uponly-238-no-talking-wav/s-AyZUd
I have tested with this track locally.

### DIFF
--- a/youtube_dl/extractor/soundcloud.py
+++ b/youtube_dl/extractor/soundcloud.py
@@ -17,7 +17,8 @@ from ..utils import (
     ExtractorError,
     int_or_none,
     unified_strdate,
-    update_url_query)
+    update_url_query,
+)
 
 
 class SoundcloudIE(InfoExtractor):
@@ -166,7 +167,7 @@ class SoundcloudIE(InfoExtractor):
         if info.get('downloadable', False):
             # We can build a direct link to the song
             format_url = update_url_query(
-                'https://api.soundcloud.com/tracks/{0}/download'.format(track_id), query)
+                'https://api.soundcloud.com/tracks/%s/download' % track_id, query)
             formats.append({
                 'format_id': 'download',
                 'ext': info.get('original_format', 'mp3'),
--- a/youtube_dl/extractor/soundcloud.py
+++ b/youtube_dl/extractor/soundcloud.py
@@ -1,8 +1,8 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
-import re
 import itertools
+import re
 
 from .common import (
     InfoExtractor,
@@ -17,7 +17,7 @@ from ..utils import (
     ExtractorError,
     int_or_none,
     unified_strdate,
-)
+    update_url_query)
 
 
 class SoundcloudIE(InfoExtractor):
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/soundcloud.py
```diff
@@ -160,13 +160,16 @@ def _extract_info_dict(self, info, full_title=None, quiet=False, secret_token=No
...
         if info.get('downloadable', False):
             # We can build a direct link to the song
             format_url = (
-                'https://api.soundcloud.com/tracks/{0}/download?client_id={1}'.format(
-                    track_id, self._CLIENT_ID))
+                'https://api.soundcloud.com/tracks/{0}/download?{1}'.format(
+                    track_id, compat_urllib_parse_urlencode(query)))
             formats.append({
-                'format_id': 'download',
+                'format_id': 'download_url',
```
dstftw: Revert.
>Checking download_**url** video format **URL**

makes even less sense.

youtube_dl/extractor/soundcloud.py
```diff
@@ -160,13 +160,16 @@ def _extract_info_dict(self, info, full_title=None, quiet=False, secret_token=No
...
+        query = {'client_id': self._CLIENT_ID}
+        if secret_token is not None:
+            query['secret_token'] = secret_token
         if info.get('downloadable', False):
             # We can build a direct link to the song
             format_url = (
-                'https://api.soundcloud.com/tracks/{0}/download?client_id={1}'.format(
-                    track_id, self._CLIENT_ID))
+                'https://api.soundcloud.com/tracks/{0}/download?{1}'.format(
+                    track_id, compat_urllib_parse_urlencode(query)))
```
dstftw: `update_url_query`.

youtube_dl/extractor/soundcloud.py
```diff
@@ -216,7 +216,7 @@ def _extract_info_dict(self, info, full_title=None, quiet=False, secret_token=No
             # cannot be always used, sometimes it can give an HTTP 404 error
             formats.append({
                 'format_id': 'fallback',
-                'url': info['stream_url'] + '?client_id=' + self._CLIENT_ID,
+                'url': info['stream_url'] + compat_urllib_parse_urlencode(query),
```
dstftw: `update_url_query`.

youtube_dl/extractor/soundcloud.py
ishitatsuyuki: The code would be longer as it is using another format parameter. It's nothing more than more overhead here.

dstftw: It won't be longer.

ishitatsuyuki: This is just an issue of preference. There are many other code written in this style, and on the other hand update_url_query has many redundant decode and encode.

dstftw: This the matter of **code reusage**, not personal preference. You should use the most generic solution available and not **reinvent the wheel**. Other code is this style was written before `update_url_query` was introduced.

youtube_dl/extractor/soundcloud.py
```diff
@@ -160,11 +160,13 @@ def _extract_info_dict(self, info, full_title=None, quiet=False, secret_token=No
...
+        query = {'client_id': self._CLIENT_ID}
+        if secret_token is not None:
+            query['secret_token'] = secret_token
         if info.get('downloadable', False):
             # We can build a direct link to the song
-            format_url = (
-                'https://api.soundcloud.com/tracks/{0}/download?client_id={1}'.format(
-                    track_id, self._CLIENT_ID))
+            format_url = update_url_query(
+                'https://api.soundcloud.com/tracks/{}/download'.format(track_id), query)
```
dstftw: `{}` won't work in python 2.6.

PR #14231, opened 2017-09-16
[pietsmiet] Add new extractor for pietsmiet.de

mhartkorn:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Adds a new extractor for www.pietsmiet.de, a private website for German YouTubers with over 2 million subscribers (https://www.youtube.com/user/PietSmittie). They release bonus content on their website that is not available on YouTube. Recently they switched their video player to HLS and the generic extractor only covers 720p now, not the best available 1080p format.

This extractor supports all provided HLS formats. It also supports the default format that is served as a single file for users that don't have ffmpeg installed. Formats are named `hls-480p`, `hls-720p`, `hls-1080p`, and `http-720p`, with `hls-1080p` being marked as `best` and `http-720p` being on top.

There seems to be a small bug in youtube-dl because even though the m3u8 parser recognizes that `hls-1080p` has 59.92 fps during parsing, in the end it shows 30.0 fps in the format overview.


youtube_dl/extractor/pietsmiet.py
```diff
@@ -0,0 +1,79 @@
...
+        formats = []
+
+        m3u8_manifest_url = data_video['sources'][0]['file']
+        m3u8_formats = self._extract_m3u8_formats(
+            m3u8_manifest_url, page_id, 'mp4', 'm3u8_native',
+            m3u8_id='hls')
+
+        # Give reproducible names for HLS formats instead of hls-<bitrate>
+        for f in m3u8_formats:
+            f['format_id'] = 'hls-{}p'.format(f['height'])
```
dstftw: There is no point in that.

youtube_dl/extractor/pietsmiet.py
```diff
@@ -0,0 +1,79 @@
...
+            format_height_raw = self._search_regex(
+                '([0-9]+)p', http_video['label'], 'http video height',
+                default=720, fatal=False)
+            format_height = int_or_none(format_height_raw)
+
+            if format_height:
+                format_width = float(format_height) * (16 / 9)
+
+                formats.append({
+                    'url': "https:{}".format(http_video['file']),
```
dstftw: `{}` won't work in python 2.6.

youtube_dl/extractor/pietsmiet.py
```diff
@@ -0,0 +1,79 @@
...
+                    'ext': http_video['type'],
+                    'format_id': 'http-{}'.format(http_video['label']),
```
dstftw: Optional data should not break extraction if missing. Read coding conventions.

mhartkorn:
Thank you for your suggestions, I changed the code accordingly.

Issue #6635, opened 2015-08-22
[req] egghead.io


chovy:
technology list: https://egghead.io/technologies/angular2
series list: https://egghead.io/series/react-flux-architecture
search results list: https://egghead.io/search?q=testing

video page: https://egghead.io/lessons/react-development-environment-setup

nodu:
+1 Especially for the Pro videos, -u and -p don't seem to be working

TRox1972:
The single video page provided seems to be a Wistia-embed, which is supported. Search and playlists for the site does not seem to work as of now, though.

chovy:
Here's an option... https://github.com/SimonSelg/egghead-downloader

chovy:
seems to work for free courses, but how do we login to egghead for pro courses?

rizafahmi:
I cannot get courses one. I tried with https://egghead.io/courses/asynchronous-javascript-with-async-await but got `ERROR: Unable to extract title`. Used version 2017.05.01.

shatyajeet:
The regex for extracting title in version 2017.05.14 is `r'<h1 class="title">([^<]+)</h1>'`. But the latest update to the website mentions the title in a `<span>...</span>`.

alexrussell:
Yes - the egghead extractor needs updating to the new way the site works.

Interestingly, the course pages seem to embed a JSON representation of the lessons (it's actually JSON embedded in a `script` tag with its type set to `application/json` in order, seemingly, to hydrate/prime a React component when the page loads), whose format is below (I have beautified the JSON, removed any array repetition and remove keys that aren't overly relevant):

```
<script type="application/json" class="js-react-on-rails-component">
{
  "component_name": "CourseApp",
  "props": {
    "course": {
      "id": 115,
      "duration": 2073,
      "title": "Maintainable CSS using TypeStyle",
      "slug": "maintainable-css-using-typestyle",
      "http_url": "https://egghead.io/courses/maintainable-css-using-typestyle",
      "url": "https://egghead.io/api/v1/series/maintainable-css-using-typestyle",
      "lessons": [{
        "id": 2050,
        "title": "Add type safety to CSS using TypeStyle",
        "slug": "css-add-type-safety-to-css-using-typestyle",
        "duration": 253,
        "series_row_order": -2097151,
        "http_url": "https://egghead.io/lessons/css-add-type-safety-to-css-using-typestyle",
[...]

PR #14388, opened 2017-10-03
[egghead:lesson] detect dash and m3u8 URLs (#6635)

mk-pmb:

- [x] Bug fix (rather, first steps towards one)
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Egghead.io (#6635) seems to have changed its distribution method to dash and m3u8. My patch checks whether the previous method (wistia) is available. If so, it uses that, no change intended. (Can't test since I can't find a video available via the old method.)
If the old method is not available, it attempts to find dash and m3u8 URLs. This step works; however, the parsing methods for both seem to not work properly for egghead's manifests (see below), so today I can only provide some of the steps towards fixing.

```text
  File "/…/ytdl-rg3/youtube_dl/extractor/common.py", line 1753, in _extract_mpd_formats
    formats_dict=formats_dict, mpd_url=mpd_url)
  File "/…/ytdl-rg3/youtube_dl/extractor/common.py", line 1925, in _parse_mpd_formats
    representation_ms_info['total_number'] = int(math.ceil(float(period_duration) / segment_duration))
TypeError: float() argument must be a string or a number
```

### DIFF
--- a/youtube_dl/extractor/egghead.py
+++ b/youtube_dl/extractor/egghead.py
@@ -9,7 +9,46 @@ from ..utils import (
 )
 
 
-class EggheadCourseIE(InfoExtractor):
+class EggheadShared():
+    def extract_lesson_metadata(self, lesson):
+        info = {
+            'title': lesson.get('title'),
+            'description': lesson.get('summary'),
+            'thumbnail': lesson.get('thumb_nail'),
+            'timestamp': unified_timestamp(lesson.get('published_at')),
+            'duration': int_or_none(lesson.get('duration')),
+            'view_count': int_or_none(lesson.get('plays_count')),
+            'tags': try_get(lesson, lambda x: x['tag_list'], list),
+        }
+
+        def find_id_and_dlurl():
+            vid_id = lesson.get('wistia_id')
+            if vid_id:
+                return {'ie_key': 'Wistia', '_type': 'url_transparent',
+                        'id': vid_id, 'url': 'wistia:' + vid_id}
+
+            self.report_warning('Cannot find an proper ID, will use lesson name URL slug')
+            vid_id = self._html_search_regex(
+                r'^https?://egghead\.io/lessons/([A-Za-z0-9][A-Za-z0-9-]*)$',
+                lesson.get('http_url'),
+                'lesson name URL part as ID of last resort',
+                group=1)
+
+            mu = lesson.get('media_urls')
+            if mu:
+                src = mu.get('dash_url')
+                if src:
+                    return {'id': vid_id, 'formats': self._extract_mpd_formats(src, vid_id)}
+                src = mu.get('hls_url')
+                if src:
+                    return {'id': vid_id, 'formats': self._extract_m3u8_formats(src, vid_id, entry_protocol='m3u8_native', m3u8_id='hls')}
+            raise NotImplementedError('Unable to detect download URL')
+        info.update(find_id_and_dlurl())
+
+        return info
+
+
+class EggheadCourseIE(InfoExtractor, EggheadShared):
     IE_DESC = 'egghead.io course'
     IE_NAME = 'egghead:course'
     _VALID_URL = r'https://egghead\.io/courses/(?P<id>[^/?#&]+)'
@@ -25,22 +64,16 @@ class EggheadCourseIE(InfoExtractor):
 
     def _real_extract(self, url):
         playlist_id = self._match_id(url)
-
         course = self._download_json(
-            'https://egghead.io/api/v1/series/%s' % playlist_id, playlist_id)
-
-        entries = [
-            self.url_result(
-                'wistia:%s' % lesson['wistia_id'], ie='Wistia',
-                video_id=lesson['wistia_id'], video_title=lesson.get('title'))
-            for lesson in course['lessons'] if lesson.get('wistia_id')]
-
+            'https://egghead.io/api/v1/series/' + playlist_id, playlist_id)
+        entries = [self.extract_lesson_metadata(lesson)
+                   for lesson in course['lessons']]
         return self.playlist_result(
             entries, playlist_id, course.get('title'),
             course.get('description'))
 
 
-class EggheadLessonIE(InfoExtractor):
+class EggheadLessonIE(InfoExtractor, EggheadShared):
     IE_DESC = 'egghead.io lesson'
     IE_NAME = 'egghead:lesson'
     _VALID_URL = r'https://egghead\.io/lessons/(?P<id>[^/?#&]+)'
@@ -65,20 +98,6 @@ class EggheadLessonIE(InfoExtractor):
 
     def _real_extract(self, url):
         lesson_id = self._match_id(url)
-
         lesson = self._download_json(
-            'https://egghead.io/api/v1/lessons/%s' % lesson_id, lesson_id)
-
-        return {
-            '_type': 'url_transparent',
-            'ie_key': 'Wistia',
-            'url': 'wistia:%s' % lesson['wistia_id'],
-            'id': lesson['wistia_id'],
-            'title': lesson.get('title'),
-            'description': lesson.get('summary'),
-            'thumbnail': lesson.get('thumb_nail'),
-            'timestamp': unified_timestamp(lesson.get('published_at')),
-            'duration': int_or_none(lesson.get('duration')),
-            'view_count': int_or_none(lesson.get('plays_count')),
-            'tags': try_get(lesson, lambda x: x['tag_list'], list),
-        }
+            'https://egghead.io/api/v1/lessons/' + lesson_id, lesson_id)
+        return self.extract_lesson_metadata(lesson)
--- a/AUTHORS
+++ b/AUTHORS
@@ -231,3 +231,4 @@ John Dong
 Tatsuyuki Ishi
 Daniel Weber
 Kay Bouché
+mk-pmb
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -540,6 +540,7 @@ class TestUtil(unittest.TestCase):
         self.assertEqual(parse_duration('87 Min.'), 5220)
         self.assertEqual(parse_duration('PT1H0.040S'), 3600.04)
         self.assertEqual(parse_duration('PT00H03M30SZ'), 210)
+        self.assertEqual(parse_duration('P0Y0M0DT0H2M4.567S'), 124.567)
 
     def test_fix_xml_ampersands(self):
         self.assertEqual(
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -1830,15 +1830,30 @@ def parse_duration(s):
     s = s.strip()
 
     days, hours, mins, secs, ms = [None] * 5
-    m = re.match(r'(?:(?:(?:(?P<days>[0-9]+):)?(?P<hours>[0-9]+):)?(?P<mins>[0-9]+):)?(?P<secs>[0-9]+)(?P<ms>\.[0-9]+)?Z?$', s)
+    m = re.match(r'''(?x)
+        (?:
+            (?:
+                (?:
+                  (?P<days>[0-9]+):
+                )?
+                (?P<hours>[0-9]+):
+            )?
+            (?P<mins>[0-9]+):
+        )?
+        (?P<secs>[0-9]+)
+        (?P<ms>\.[0-9]+)?
+        Z?$''', s)
     if m:
         days, hours, mins, secs, ms = m.groups()
     else:
         m = re.match(
-            r'''(?ix)(?:P?T)?
+            r'''(?ix)P?T?
+                (?:0Y)?
+                (?:0M)?
                 (?:
                     (?P<days>[0-9]+)\s*d(?:ays?)?\s*
                 )?
+                T?
                 (?:
                     (?P<hours>[0-9]+)\s*h(?:ours?)?\s*
                 )?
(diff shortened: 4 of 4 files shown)


youtube_dl/extractor/egghead.py
```diff
@@ -82,3 +78,42 @@ def _real_extract(self, url):
...
+                return { '_type': 'url_transparent', 'ie_key': 'Wistia',
+                    'id': id, 'url': 'wistia:' + id }
+
+            self.report_warning('Cannot find an proper ID, will use URL slug')
+            # !!!! !!!! !!!! !!!! !!!! !!!! !!!! !!!! !!!! !!!! !!!! !!!!
+            # !!  Don't fall for the "id" field in the JSON. Example:
+            # !!  Lesson "javascript-redux-the-single-immutable-state-tree"
+            # !!  has "id": 696 but https://egghead.io/lessons/696
+            # !!  is another video.
+            # !!!! !!!! !!!! !!!! !!!! !!!! !!!! !!!! !!!! !!!! !!!! !!!!
```
dstftw: Remove.

youtube_dl/extractor/egghead.py
```diff
@@ -82,3 +78,42 @@ def _real_extract(self, url):
...
+            info['id'] = id = last_resort_id()
```
dstftw: Don't shadow built-in names. Don't extract function you only use once.

youtube_dl/extractor/egghead.py
```diff
@@ -82,3 +78,42 @@ def _real_extract(self, url):
...
+
+
+        def last_resort_id():
+            id = lesson.get('http_url')
+            if id:
+                id = id.split('https://egghead.io/lessons/')
+                if (len(id) == 2) and (id[0] == ''):
+                    id[0] = id[1].replace('-', '')
+                    if id[0] and id[0].isalnum():
+                        return id[1]
```
dstftw: `_search_regex`.

dstftw:
1. flake8.
2. Tests.

mk-pmb:
Now I managed to delegate the URLs to the generic downloader, with either `mpeg-dash:` or `hls-m3u8:` prefixed. This currently leads to

`ERROR: Unable to download webpage: <urlopen error unknown url type: mpeg-dash> (caused by URLError(u'unknown url type: mpeg-dash',))`

or its `hls-m3u8` twin, so the egghead-specific problems are solved.
All that's left is for someone to implement extractors for these kinds of URLs, and I hope we'll get them because I'd like to use them directly on the CLI anyway.

dstftw:
No such prefixes exist. `_extract_m3u8_formats`/`_extract_mpd_formats` should be used.

mk-pmb:
> _extract_mpd_formats should be used.

That's what I had in the first version, which in case of [this manifest](http://web.archive.org/web/20171003174623im_/https://d2c5owlt6rorc3.cloudfront.net/react-react-in-7-minutes-ba65497c81/react-react-in-7-minutes-ba65497c81.mpd) leads to the parse error cited at the bottom of the original post of this issue.

youtube_dl/extractor/egghead.py
```diff
@@ -82,3 +79,29 @@ def _real_extract(self, url):
             'view_count': int_or_none(lesson.get('plays_count')),
             'tags': try_get(lesson, lambda x: x['tag_list'], list),
         }
+
+        def find_id_and_dlurl():
+            vid_id = lesson.get('wistia_id')
```
joelhooks: This has been removed entirely, we aren't using Wistia for streaming any longer.

youtube_dl/extractor/egghead.py
```diff
@@ -82,3 +79,29 @@ def _real_extract(self, url):
...
+                        'id': vid_id, 'url': 'wistia:' + vid_id}
+
+            self.report_warning('Cannot find an proper ID, will use lesson name URL slug')
+            vid_id = self._html_search_regex(
+                r'^https?://egghead\.io/lessons/([A-Za-z0-9][A-Za-z0-9-]*)$',
+                lesson.get('http_url'),
+                'lesson name URL part as ID of last resort',
+                group=1)
+
+            mu = lesson.get('media_urls')
```
joelhooks: I'm literally going to keep changing these around to different properties as soon as this PR is merged.

mk-pmb:
Rebased on current master. Someone will still have to either get `_extract_m3u8_formats` and `_extract_mpd_formats` to work, or (my preferred solution) implement the `mpeg-dash:` and `hls-m3u8:` prefixes.

mk-pmb:
I've managed to get lessons to work with `_extract_m3u8_formats`, and updated the course playlist scanner to use the same logic. I still can't download because of [this ffmpeg error](https://gist.github.com/mk-pmb/aad06a34de8a9c06bb5e94de53ca728e) but please try whether it works for you.

joelhooks:
You can avoid this hassle by subscribing as a member. Then the api JSON  has a download link and we are rolling out course zip downloads to members soon as well. 😄

Members also get RSS feeds which function as video podcasts for use with apps. It is actually very nice. 

Or keep chasing me with this thing, if it’s a hobby or whatever.

mk-pmb:
> You can avoid this hassle by subscribing as a member.

I assumed so and had even considered suggesting it for the impatient people in the other thread. :-)
However, the egghead part of the hassle was solved even before I opened this PR, so the remaining hassle is all about ytdl and I still refuse to give up on it when the task seems to be so easy. I'm currently building my own HLS downloader in a separate project to try and understand what could be so hard about it.

remitamine:
@mk-pmb when you find a problem you should look for the cause of the problem and try to fix it.
from error you have posted before:
```text
  File "/…/ytdl-rg3/youtube_dl/extractor/common.py", line 1753, in _extract_mpd_formats
    formats_dict=formats_dict, mpd_url=mpd_url)
  File "/…/ytdl-rg3/youtube_dl/extractor/common.py", line 1925, in _parse_mpd_formats
    representation_ms_info['total_number'] = int(math.ceil(float(period_duration) / segment_duration))
TypeError: float() argument must be a string or a number
```
you can see that error is that `float` function received an argument with an unexpected type which is `period_duration`(it's value in this case is `None`) in `_parse_mpd_formats`.
as it's been said in the error go to line 1925 of `common.py` file, and then you can look back where the value of this variable is calculated, you will find that it's calculated using the value of `duration` attribute of `Period` element and fallback to the value of `mediaPresentationDuration` attribute of the root element(`MPD`).
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<MPD id="4bef176d-4bad-4402-82c6-ae49d9b56bd6" profiles="urn:mpeg:dash:profile:isoff-live:2011,urn:com:dashif:dash264" type="static" mediaPresentationDuration="P0Y0M0DT0H1M10.400S" minBufferTime="P0Y0M0DT0H0M2.000S" bitmovin:version="1.19.0" xmlns="urn:mpeg:dash:schema:mpd:2011" xmlns:bitmovin="http://www.bitmovin.net/mpd/2015" xmlns:ns2="http://www.w3.org/1999/xlink">
    <Period id="f6c05353-0e0f-4c7a-b737-35075cac0d99" start="P0Y0M0DT0H0M0.000S">
        <AdaptationSet segmentAlignment="true" mimeType="video/mp4">
            <Representation id="17a5b43d-90ed-49f6-9d3a-6ec0be53918c" bandwidth="2400000" width="1280" height="720" frameRate="25" codecs="avc1.64001f">
                <SegmentTemplate media="video/720p/seg_$Number$.m4s" initialization="video/720p/init.mp4" duration="100000" startNumber="0" timescale="25000"/>
            </Representation>
            <Representation id="baee55d7-3416-4a4f-95ce-a522bd0d3365" bandwidth="1200000" width="854" height="480" frameRate="25" codecs="avc1.64001e">
                <SegmentTemplate media="video/480p/seg_$Number$.m4s" initialization="video/480p/init.mp4" duration="100000" startNumber="0" timescale="25000"/>
            </Representation>
        </AdaptationSet>
        <AdaptationSet lang="en" segmentAlignment="true" mimeType="audio/mp4">
            <AudioChannelConfiguration schemeIdUri="urn:mpeg:dash:23003:3:audio_channel_configuration:2011" value="2"/>
            <Representation id="1b706409-b185-4490-a24a-0a57370db6fe" bandwidth="128000" audioSamplingRate="48000" codecs="mp4a.40.2">
                <SegmentTemplate media="audio/seg_$Number$.m4s" initialization="audio/init.mp4" duration="192000" startNumber="0" timescale="48000"/>
            </Representation>
        </AdaptationSet>
    </Period>
</MPD>
```
from the manifest you can see that there is only the value of `mediaPresentationDuration`(`P0Y0M0DT0H1M10.400S`) that is passed to `parse_duration` function imported from `utils.py`.
so the problem here is that `parse_duration` fails to convert `P0Y0M0DT0H1M10.400S` into a float.
so what needs to be fixed is `parse_duration` function to recognize this format of durations.
from the dash standard the `mediaPresentationDuration` duration is a [`XML Schema` `duration`](https://www.w3.org/TR/xmlschema-2/#duration) which is an [ISO 8601 duration](https://en.wikipedia.org/wiki/ISO_8601#Durations).
using the information from the last link i made these changes to match the format of duration used by the manifest(it does not count for basic and extended formats).
```python
def parse_duration(s):
    if not isinstance(s, compat_basestring):
        return None

    s = s.strip()

    years, months, weeks, days, hours, mins, secs, ms = [None] * 8
    # ISO 8601 duration
    # https://en.wikipedia.org/wiki/ISO_8601#Durations
    m = re.match(r'''(?x)P(?=[0-9]+(?:[,.][0-9]+)?[YMWD]|T[0-9]+(?:[,.][0-9]+)?[HMS])
                     (?:
                         (?P<years>[0-9]+(?:[,.][0-9]+)?)Y
                     )?
                     (?:
                         (?P<months>[0-9]+(?:[,.][0-9]+)?)M
                     )?
                     (?:
                         (?P<weeks>[0-9]+(?:[,.][0-9]+)?)W
                     )?
                     (?:
                         (?P<days>[0-9]+(?:[,.][0-9]+)?)D
                     )?
                     (?:
                         T
                         (?:
                             (?P<hours>[0-9]+(?:[,.][0-9]+)?)H
                         )?
                         (?:
                             (?P<minutes>[0-9]+(?:[,.][0-9]+)?)M
                         )?
                         (?:
                             (?P<seconds>[0-9]+(?:[,.][0-9]+)?)S
                         )?
                     )?''', s)
    if m:
        years, months, weeks, days, hours, mins, secs = map(lambda f: f.replace(',', '.') if f else None, m.groups())
    else:
        m = re.match(r'(?:(?:(?:(?P<days>[0-9]+):)?(?P<hours>[0-9]+):)?(?P<mins>[0-9]+):)?(?P<secs>[0-9]+)(?P<ms>\.[0-9]+)?Z?$', s)
        if m:
            days, hours, mins, secs, ms = m.groups()
        else:
            m = re.match(
                r'''(?ix)(?:P?T)?
                    (?:
                        (?P<days>[0-9]+)\s*d(?:ays?)?\s*
                    )?
                    (?:
                        (?P<hours>[0-9]+)\s*h(?:ours?)?\s*
                    )?
                    (?:
                        (?P<mins>[0-9]+)\s*m(?:in(?:ute)?s?)?\s*
                    )?
                    (?:
                        (?P<secs>[0-9]+)(?P<ms>\.[0-9]+)?\s*s(?:ec(?:ond)?s?)?\s*
                    )?Z?$''', s)
            if m:
                days, hours, mins, secs, ms = m.groups()
            else:
                m = re.match(r'(?i)(?:(?P<hours>[0-9.]+)\s*(?:hours?)|(?P<mins>[0-9.]+)\s*(?:mins?\.?|minutes?)\s*)Z?$', s)
                if m:
                    hours, mins = m.groups()
                else:
                    return None

    duration = 0
    if secs:
        duration += float(secs)
    if mins:
        duration += float(mins) * 60
    if hours:
        duration += float(hours) * 60 * 60
    if days:
        duration += float(days) * 60 * 60 * 24
    if weeks:
        duration += float(weeks) * 60 * 60 * 24 * 7
    # if months:
    #     duration += float(months) * 60 * 60 * 24 * days_in_month
    # if years:
    #     duration += float(years) * 60 * 60 * 24 * days_in_year
    if ms:
        duration += float(ms)
    return duration
```
as the format was partially recognized with this part:
```python
            m = re.match(
                r'''(?ix)(?:P?T)?
                    (?:
                        (?P<days>[0-9]+)\s*d(?:ays?)?\s*
                    )?
                    (?:
                        (?P<hours>[0-9]+)\s*h(?:ours?)?\s*
                    )?
                    (?:
                        (?P<mins>[0-9]+)\s*m(?:in(?:ute)?s?)?\s*
                    )?
                    (?:
                        (?P<secs>[0-9]+)(?P<ms>\.[0-9]+)?\s*s(?:ec(?:ond)?s?)?\s*
                    )?Z?$''', s)
```
it will better to either unify them or remove the duplicated parts.
of course this change will require to add more tests to `test_utils.py` and to handle .

mk-pmb:
Thanks for diving in this deep. I had tracked down the float issue and tried to fix it, but in a manner that wasn't standards compliant and thus rejected. I didn't want to spend time on learning that standard, instead hoping someone who already knows will fix it, since at that point I wasn't even convinced it's important to know the duration.
Shall I try and merge your code snippets from above into my PR, or merge one of your branches into my PR branch?

mk-pmb:
@joelhooks Now I see how the duration issue unexpectedly is indeed egghead-specific. While I expect ytdl to cope with your custom duration format, it looks broken nonetheless: (spaces added for easier reading)
`P 0Y0M 0D T 0H2M11.600S`
The T should probably be right after the P. Years and months seem inappropriate as precise unit of duration because they vary in length.
Just so you know. As said, if it works in browsers, not your problem whether ytdl clings to specs. :-)

mk-pmb:
I fixed the duration parser so it ignores 0 years and 0 months, and now the download works. VLC can't play the resulting file but Totem can, that's good enough for me.

@remitamine Your duration patch is probably better but I wasn't motivated to compare and merge it. Feel free to PR it independently.

iamdubx:
@joelhooks so you actually telling that with PRO `youtube-dl` is working ok? Then why it's not working for me...
Also, wait for Download option for subscribers.

mk-pmb:
@iamdubx Wanna try the version from [my PR branch](https://github.com/mk-pmb/youtube-dl-rg3/tree/egghead-mediaurls-171002)?

PR #14396, opened 2017-10-03
MPD manifests: improve errmsg for missing duration

mk-pmb:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This patch checks for missing data for the segment duration calculation, and if it's missing, raises a more specific error than just the generic type conversion failure.

Example of original error message from #14388:

```text
  File "/…/ytdl-rg3/youtube_dl/extractor/common.py", line 1753, in _extract_mpd_formats
    formats_dict=formats_dict, mpd_url=mpd_url)
  File "/…/ytdl-rg3/youtube_dl/extractor/common.py", line 1925, in _parse_mpd_formats
    representation_ms_info['total_number'] = int(math.ceil(float(period_duration) / segment_duration))
TypeError: float() argument must be a string or a number
```

NB [line 1923 of the old `youtube_dl/extractor/common.py`](https://github.com/rg3/youtube-dl/blob/bfd484ccff4ea9d16af9515416d4f35d14d5b41d/youtube_dl/extractor/common.py#L1923):

As far as I can tell, the 2nd part of the if condition
`and 'segment_duration'` is useless because it will always
pass, so I made a guess at what could have been meant.
We should try and get a linter program to warn about useless
expressions. If we have one, we should put a better hint in
the coding guidelines.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -1920,7 +1920,7 @@ class InfoExtractor(object):
                             # can't be used at the same time
                             if '%(Number' in media_template and 's' not in representation_ms_info:
                                 segment_duration = None
-                                if 'total_number' not in representation_ms_info and 'segment_duration':
+                                if 'total_number' not in representation_ms_info and 'segment_duration' in representation_ms_info:
                                     segment_duration = float_or_none(representation_ms_info['segment_duration'], representation_ms_info['timescale'])
                                     representation_ms_info['total_number'] = int(math.ceil(float(period_duration) / segment_duration))
                                 representation_ms_info['fragments'] = [{


youtube_dl/extractor/common.py
```diff
@@ -1833,6 +1833,8 @@ def extract_Initialization(source):
         formats = []
         for period in mpd_doc.findall(_add_ns('Period')):
             period_duration = parse_duration(period.get('duration')) or mpd_duration
+            if period_duration is None:
```
dstftw: No. `period` is optional attribute.

youtube_dl/extractor/common.py
```diff
@@ -1920,8 +1922,11 @@ def location_key(location):
                             # can't be used at the same time
                             if '%(Number' in media_template and 's' not in representation_ms_info:
                                 segment_duration = None
-                                if 'total_number' not in representation_ms_info and 'segment_duration':
```
dstftw: Supposed to be `'segment_duration' in representation_ms_info`.

youtube_dl/extractor/common.py
```diff
@@ -1920,8 +1922,11 @@ def location_key(location):
...
-                                    segment_duration = float_or_none(representation_ms_info['segment_duration'], representation_ms_info['timescale'])
+                                if 'total_number' not in representation_ms_info:
+                                    rep_segdura = representation_ms_info.get('segment_duration')
+                                    if rep_segdura is None:
```
dstftw: No.

dstftw:
1. This code is invalid. You shall use `start` attribute of `Period` for calculations when `duration` is missing.
2. Add tests with real world DASH manifest.

mk-pmb:
I've reduced the patch to only fix the obvious missing condition, as I can't currently afford the time to learn the DASH spec.

> when duration is missing.

Then the problem is something else anyway, because all the SegmentTemplates do have a duration attribute in [this manifest](http://web.archive.org/web/20171003174623im_/https://d2c5owlt6rorc3.cloudfront.net/react-react-in-7-minutes-ba65497c81/react-react-in-7-minutes-ba65497c81.mpd).

PR #14400, opened 2017-10-03
[doc] coding conventions: add hint to linter

mk-pmb:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This patch makes it easier for potential new contributors to discover that we actually do have (and should use) a linter.


dstftw:
Current wording is clear enough about what and should do and how.

mk-pmb:
> Current wording is clear enough about what and should do and how.

Your opinion surprises me. It might be clear enough indeed, but not really accessible. Why should you care about UX in this case? Because some amount of your time and your effort was wasted.

How?
In your comment to #14388 you had to tell me "flake8.", meaning that indeed we do have a linter. I assume you also had to perform some kind of check in order to discover that I had not linted.

Why?
Yesterday the checbox in the issue lead me to the coding conventions. I read them, and couldn't find any sign of a linter in that chapter. I couldn't believe a project like this would do without, so I searched the text of the page for the word "lint". No match. I thus assumed we wouldn't use a linter here, and thus submitted my PR without linting.

dstftw:
For this particular code I didn't run flake8. It's pretty obvious even after a brief look that the code does not meet PEP 8.
The same checkbox leads to the tutorial (it goes even before mentioning coding conventions) that describes everything needed step by step. You've searched for "lint" other may have searched for "pyflakes", "checker", "flake" or whatever they used to call it. This does not mean all these keywords should be mentioned. There are also many other specific aspects apart from code style that may also be referred differently and it's pointless to include all possible buzzwords for them. All you have to do is to read everything that is linked in template carefully.

mk-pmb:
I see. Thanks for explaining.

PR #14443, opened 2017-10-07
[HRTi] new urls are wihout #

atopuzov:
* New URLs don't have `#`
Not sure if they kept the old ones, as a safe bet making the part of the regex just optional.

### DIFF
--- a/youtube_dl/extractor/hrti.py
+++ b/youtube_dl/extractor/hrti.py
@@ -104,7 +104,7 @@ class HRTiIE(HRTiBaseIE):
                         (?:
                             hrti:(?P<short_id>[0-9]+)|
                             https?://
-                                hrti\.hrt\.hr/\#/video/show/(?P<id>[0-9]+)/(?P<display_id>[^/]+)?
+                                hrti\.hrt\.hr/(?:\#/)?video/show/(?P<id>[0-9]+)/(?P<display_id>[^/]+)?
                         )
                     '''
     _TESTS = [{
@@ -128,6 +128,9 @@ class HRTiIE(HRTiBaseIE):
         'only_matching': True,
     }, {
         'url': 'hrti:2181385',
+        'only_matching': True,
+    }, {
+        'url': 'https://hrti.hrt.hr/video/show/3873068/cuvar-dvorca-dramska-serija-14',
         'only_matching': True,
     }]
 


dstftw:
Carefully read [new issue template](https://github.com/rg3/youtube-dl/issues/new) and provide all requested information.

youtube_dl/extractor/hrti.py
```diff
@@ -104,7 +104,7 @@ class HRTiIE(HRTiBaseIE):
                         (?:
                             hrti:(?P<short_id>[0-9]+)|
                             https?://
-                                hrti\.hrt\.hr/\#/video/show/(?P<id>[0-9]+)/(?P<display_id>[^/]+)?
+                                hrti\.hrt\.hr/(\#/)?video/show/(?P<id>[0-9]+)/(?P<display_id>[^/]+)?
```
dstftw: Don't capture groups you don't use.

youtube_dl/extractor/hrti.py
```diff
@@ -129,6 +129,22 @@ class HRTiIE(HRTiBaseIE):
     }, {
         'url': 'hrti:2181385',
         'only_matching': True,
+    }, {
+        'url': 'https://hrti.hrt.hr/video/show/3873068/cuvar-dvorca-dramska-serija-14',
```
dstftw: `only_matching`.

atopuzov: Remove the entire `info_dict` and add `'only_matching': True,` or just add `'only_matching': True,` and keep the `info_dict`.

dstftw: `info_dict` is pointless with `only_matching`.

youtube_dl/extractor/hrti.py
atopuzov: The group is there for optionality. I wanted to match only `...hrti.hrt.hr/#/video...` and `...hrti.hrt.hr/video...`
Then `\#?/?` (this catches `...hrti.hrt.hr//video...` too) or don't capture with `/#/` (but I'm not sure if there are still links with it). Suggestions?

dstftw: You don't get it. Group is fine, but it **should not** capture.

atopuzov:
Updated with the requested changes.

hmlinaric:
The extractor not working anymore. HRTi has changed web site. Can anyone help with writing a new extractor?
I have added play.rtl.hr in 6play extractor.

PR #14566, opened 2017-10-23
[twitch] Fix clip title detection.

jdpurcell:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

When downloading clips from Twitch, the clip title used to be part of the filename. That broke within the last month or so, and now the filename only has generic information like "channel name playing some game". This PR restores the ability to detect the clip title.

### DIFF
--- a/youtube_dl/extractor/twitch.py
+++ b/youtube_dl/extractor/twitch.py
@@ -609,7 +609,7 @@ class TwitchClipsIE(InfoExtractor):
                 r'(?s)clipInfo\s*=\s*({.+?});', webpage, 'clip info'),
             video_id, transform_source=js_to_json)
 
-        title = clip.get('title') or self._og_search_title(webpage)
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/twitch.py
```diff
@@ -609,7 +609,7 @@ def _real_extract(self, url):
                 r'(?s)clipInfo\s*=\s*({.+?});', webpage, 'clip info'),
             video_id, transform_source=js_to_json)
 
-        title = clip.get('channel_title') or self._og_search_title(webpage)
+        title = clip.get('title') or self._og_search_title(webpage)
```
dstftw: Keep the old way also.

jdpurcell: @dstftw Sure, done.

Issue #5731, opened 2015-05-17
WDR Elefantenseite not supported


derdennis:
Hi all,

I wanted to download some videos from the german [Elefantenseite](http://www.wdrmaus.de/elefantenseite/), a subsite of the already supported [WDR Maus](http://www.wdrmaus.de/).

There are many single video site like [this one](http://www.wdrmaus.de/elefantenseite/#/anke_singt_auf_der_mauer_auf_der_lauer) and youtube-dl is not able to download the video. Output as follows:

```
$ youtube-dl --verbose http://www.wdrmaus.de/elefantenseite/#/anke_singt_auf_der_mauer_auf_der_lauer
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--verbose', u'http://www.wdrmaus.de/elefantenseite/#/anke_singt_auf_der_mauer_auf_der_lauer']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.05.15
[debug] Python version 2.7.6 - Darwin-14.3.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg 2.6.2, ffprobe 2.6.2, rtmpdump 2.3
[debug] Proxy map: {}
[WDRMaus] elefantenseite: Downloading webpage
ERROR: Unable to extract parameters; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 649, in extract_info
    ie_result = ie.extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 273, in extract
    return self._real_extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/wdr.py", line 240, in _real_extract
    r'<a href="\?startVideo=1&amp;([^"]+)"', webpage, 'parameters')
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 564, in _html_search_regex
    res = self._search_regex(pattern, string, name, default, fatal, flags, group)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 555, in _search_regex
    raise RegexNotFoundError('Unable to extract %s' % _name)
[...]

PR #14598, opened 2017-10-27
[wdr] New extractor "wdr:elefant" for videos of "Sendung mit dem Elefanten"

sleske:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

New extractor "wdr:elefant" (sub-extractor of "wdr") to support videos of the "Sendung mit dem Elefanten" (a children's show) at http://www.wdrmaus.de/elefantenseite/ . See commit message for technical details.

### DIFF
--- a/youtube_dl/extractor/wdr.py
+++ b/youtube_dl/extractor/wdr.py
@@ -253,7 +253,10 @@ class WDRElefantIE(WDRBaseIE):
                 'id': 'mdb-1008774',
                 'ext': 'mp4',
                 'age_limit': None,
-                'upload_date': '20091119'
+                'upload_date': '20091119',
+            },
+            'params': {
+                'skip_download' : True,
             },
         },
         {
@@ -264,6 +267,9 @@ class WDRElefantIE(WDRBaseIE):
                 'ext': 'mp4',
                 'age_limit': None,
                 'upload_date': '20150406'
+            },
+            'params': {
+                'skip_download' : True,
             },
         },
     ]
--- a/youtube_dl/extractor/wdr.py
+++ b/youtube_dl/extractor/wdr.py
@@ -16,7 +16,7 @@ from ..utils import (
 
 
 class WDRBaseIE(InfoExtractor):
-    def _extract_wdr_video(self, webpage, display_id):
+    def _extract_jsonp_url(self, webpage, display_id):
         # for wdr.de the data-extension is in a tag with the class "mediaLink"
         # for wdr.de radio players, in a tag with the class "wdrrPlayerPlayBtn"
         # for wdrmaus, in a tag with the class "videoButton" (previously a link
@@ -35,8 +35,9 @@ class WDRBaseIE(InfoExtractor):
 
         media_link_obj = self._parse_json(json_metadata, display_id,
                                           transform_source=js_to_json)
-        jsonp_url = media_link_obj['mediaObj']['url']
+        return media_link_obj['mediaObj']['url']
 
+    def _extract_wdr_video(self, jsonp_url, display_id):
         metadata = self._download_json(
             jsonp_url, display_id, transform_source=strip_jsonp)
 
@@ -206,7 +207,8 @@ class WDRIE(WDRBaseIE):
         display_id = mobj.group('display_id')
         webpage = self._download_webpage(url, display_id)
 
-        info_dict = self._extract_wdr_video(webpage, display_id)
+        jsonp_url = self._extract_jsonp_url(webpage, display_id)
+        info_dict = self._extract_wdr_video(jsonp_url, display_id)
 
         if not info_dict:
             entries = [
--- a/youtube_dl/extractor/wdr.py
+++ b/youtube_dl/extractor/wdr.py
@@ -241,6 +241,59 @@ class WDRIE(WDRBaseIE):
         return info_dict
 
 
+class WDRElefantIE(WDRBaseIE):
+    _VALID_URL = r'https?://(?:www\.)wdrmaus.de/elefantenseite/#(?P<display_id>.+)'
+    IE_NAME = 'wdr:elefant'
+
+    _TESTS = [
+        {
+            'url': 'http://www.wdrmaus.de/elefantenseite/#lieder_geburtstagslied',
+            'info_dict': {
+                'title': 'Ich bin schon 1-2-3',
+                'id': 'mdb-1008774',
+                'ext': 'mp4',
+                'age_limit': None,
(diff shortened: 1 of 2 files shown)


sleske:
Hi, is there any way I can help to get this merged? Or will it just take some time for someone to review?

sleske:
Note: This PR solves issue #5731.

sleske:
@dstftw : Would you consider reviewing this PR? I'd love for this to get merged - it's for a very popular children's TV show page.

youtube_dl/extractor/wdr.py
```diff
@@ -239,6 +241,59 @@ def _real_extract(self, url):
...
+                'ext': 'mp4',
+                'age_limit': None,
+                'upload_date': '20150406'
+            },
+        },
+    ]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        display_id = mobj.group('display_id')
```
dstftw: `self._match_id`.

youtube_dl/extractor/wdr.py
```diff
@@ -239,6 +241,59 @@ def _real_extract(self, url):
...
+                'Is the fragment part of the URL (after the #) correct?',
+                expected=True)
+        xml_metadata_path = table_of_contents[display_id]['xmlPath']
+        xml_metadata = self._download_xml(
+            'https://www.wdrmaus.de/elefantenseite/' + xml_metadata_path, display_id)
+        zmdb_url_element = xml_metadata.find('./movie/zmdb_url')
+        if zmdb_url_element is None:
+            raise ExtractorError(
+                'The URL looks valid, but no video was found. Note that download only works '
+                'on pages showing a single video, not on video selection pages.',
```
dstftw: `<display_id> is not a video`.

youtube_dl/extractor/wdr.py
```diff
@@ -239,6 +241,59 @@ def _real_extract(self, url):
...
+        },
+        {
+            'url': 'http://www.wdrmaus.de/elefantenseite/#folge_ostern_2015',
+            'info_dict': {
+                'title': 'Folge Oster-Spezial 2015',
+                'id': 'mdb-1088195',
+                'ext': 'mp4',
+                'age_limit': None,
+                'upload_date': '20150406'
+            },
```
dstftw: Remove duplicates.

youtube_dl/extractor/wdr.py
```diff
@@ -239,6 +241,59 @@ def _real_extract(self, url):
...
+        {
+            'url': 'http://www.wdrmaus.de/elefantenseite/#lieder_geburtstagslied',
+            'info_dict': {
+                'title': 'Ich bin schon 1-2-3',
+                'id': 'mdb-1008774',
+                'ext': 'mp4',
+                'age_limit': None,
+                'upload_date': '20091119'
+            },
+        },
```
dstftw: `skip_download`.

youtube_dl/extractor/wdr.py
sleske: Could you clarify this comment? Is that the error message that you prefer?

I made the message a bit more verbose to clarify that this is not a copy&paste error - the URL is valid, but there is no video behind it. But I you want a simpler message for youtube-dl, I'll gladly change it.

youtube_dl/extractor/wdr.py
sleske: What do you mean by "Remove duplicates"? What is duplicate here? Sorry, I can't follow.

youtube_dl/extractor/wdr.py
sleske: Ok, done.

youtube_dl/extractor/wdr.py
sleske: Sorry, I don't understand. Do you mean I should use "self._match_id" instead of "mobj.group" in the last line? How would that work, given that "self._match_id" is for matching the id, and not the display_id?

youtube_dl/extractor/wdr.py
dstftw: Remove all tests that test the same extraction scenario.

youtube_dl/extractor/wdr.py
dstftw: Nothing stops from renaming capture group to `id`.

sleske:
@dstftw : Thanks for the explanations. I'm working on the code...

youtube_dl/extractor/wdr.py
sleske: Ah, ok. I added two tests to increase the chances of having at least one working test, because the site removes most videos after some time (with the time depending on strange rules). However, if you feel one test is enough, I'll drop the other.
Done.

youtube_dl/extractor/wdr.py
sleske: Yes, of course I could do that. But as far as I see, the string I'm retrieving here is not the id. The id is retrieved later, in line 95 ( metadata_tracker_data.get('trackerClipId', display_id) ). So by calling this string "id" instead of "display_id" I would just cause confusion, wouldn't I?
If you insist, I'll gladly change "display_id" to "id" (everywhere in the code, otherwise it would be even more confusing) - I'm just not sure that would be helpful.

sleske:
@dstftw : Thank you for your feeback. I have tried to address it (or asked for more information, since some of your comments were quite terse). About your main comment:

> How extractors should be reorganized:
> * New WDRIE should match and process http://deviceids-medp.wdr.de/ondemand/108/1088195.js.
>  * Old WDRIE should be renamed to something else and should delegate url_transparent to new WDRIE.
>  * WDRElefantIE should delegate to new WDRIE

If I understand you correctly, your idea is that the function _extract_wdr_video should be transformed into a new extractor (WDRIE), which the other extractors delegate to. Is that what you mean?
If yes, then that seems problematic to me: While my new extractor, WDRElefantIE, does nothing more after calling _extract_wdr_video, the function is also used by two other extractors (WDRIE and SportschauIE, in sportschau.py). These other two extractors do additional work after calling _extract_wdr_video , so I don't see how they can delegate to another extractor.

sleske:
Also, on a personal note, this looks like a significant refactoring of existing extractors. While I understand the need to keep the code clean, it would be nice if you could add this PR first, so the new extractor can be used. Then I'll gladly help with work on refactoring the code, as far as I can. Thanks!

sleske:
@dstftw : Thanks for reworking and merging this PR! Now I see how you wanted the reorganization to happen.

PR #14636, opened 2017-10-31
Update douyutv and pandatv API

hcwhan:


dstftw:
Don't mix unrelated changes in single PR.

PR #14642, opened 2017-10-31
[daum] Fix clipID-based downloads

Namnamseo:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Some daum.net URLs have a alphanumeric **video ID** called `vid`, which looks like `v2b9eCxzw1EzE6swwSS4Okz`. An example URL is `http://tvpot.daum.net/v/v2b9eCxzw1EzE6swwSS4Okz`

Some other URLs have a numeric **clip ID** called `clipid`, which looks like `66008403`. An example URL is `http://tvpot.daum.net/v/66008403`

Currently, download for URLs with clip IDs is broken. This commit fixes it, and also updates the expected test data.

It achieves it by redirecting it to the `vid`-based URL. The URL can be found from the canonical `link` tag, which is explained later. I'm not sure if the `vid` is always present in this manner in clip ID-based pages. But 10 pages of 10 I saw had them, and I don't think there could be a reason other pages won't have it.

TODO: Some other features of `daum.net`, such as playlists, are still broken. Those should be fixed later.
In fact, TVPot has disabled almost all of its features, so playlists might not work from now on. Some old API endpoints are still alive and current extractor is using them, like in [daum.py#L85](https://github.com/rg3/youtube-dl/blob/6d0630d8801fd3278a05fa7e55a73bd454403e5a/youtube_dl/extractor/daum.py#L85). 

##### Explanation of the canonical `link` tag
`<link rel="canonical" href="http://example.com">`
This kind of tag, presented in `<head>`, is originally meant for search engines to correctly discern and group URLs. Often is the case where the same page can be accessed via a few different URLs. This tag tells the search engine that it can consider this page **as** the page given in the `href` attribute, making the search results faster and more accurate.

### DIFF
--- a/youtube_dl/extractor/daum.py
+++ b/youtube_dl/extractor/daum.py
@@ -40,11 +40,11 @@ class DaumIE(InfoExtractor):
     }, {
         'url': 'http://m.tvpot.daum.net/v/65139429',
         'info_dict': {
-            'id': '65139429',
+            'id': 'v4e99Kd61HUKxI18xR87xRb',
             'ext': 'mp4',
             'title': '1297회, \'아빠 아들로 태어나길 잘 했어\' 민수, 감동의 눈물[아빠 어디가] 20150118',
-            'description': 'md5:79794514261164ff27e36a21ad229fc5',
-            'upload_date': '20150604',
+            'description': 'md5:4c1f30a96780bb3cb739f8878d623998',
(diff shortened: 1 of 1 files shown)


Namnamseo:
bump.

Namnamseo:
Posting a new request.

PR #14616, opened 2017-10-28
[seznamzpravy] Add new extractor

che0:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [x] At least skimmed through [adding new extractor tutorial](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/rg3/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

### Add new extractor for [Seznam Zprávy](https://www.seznam.cz/zpravy)

This extractor extracts video from Seznam Zprávy, as requested in #14102.

PR #14648, opened 2017-11-01
Ignore missing attributes in MPD manifests.

jaykbull:

- [X] Bug fix

---

Some sites, like thisav, does not include all attributes which youtube-dl expects/requires in their MPD manifests.

I don't know the MPD/DASH spec at all, so I can't tell if this makes the manifest itself non-compliant... That said, it works when played in a browser.

By treating these attributes as non-mandatory in code and simply moving along when we cannot find them, we seem to be able to successfully download videos from such sites.

This closes https://github.com/rg3/youtube-dl/issues/13784.

### DIFF
--- a/test/test_InfoExtractor.py
+++ b/test/test_InfoExtractor.py
@@ -563,6 +563,23 @@ jwplayer("mediaplayer").setup({"abouttext":"Visit Indie DB","aboutlink":"http:\/
                     'height': 1080,
                 }]
             ),
+            (
+                # https://github.com/rg3/youtube-dl/issues/13784
+                'thisav',
+                'http://unknown/manifest.mpd',
+                [{
+                    'url': 'http://unknown/300708_dashinit.mp4',
+                    'manifest_url': 'http://unknown/manifest.mpd',
+                    'ext': 'mp4',
+                    'format_note': 'DASH video',
+                    'protocol': None,
+                    'acodec': 'mp4a.40.2',
+                    'vcodec': 'avc3.64001e',
+                    'tbr': 574.578,
+                    'width': 640,
+                    'height': 426,
+                }]
+            )
         ]
 
         for mpd_file, mpd_url, expected_formats in _TEST_CASES:
--- a/test/testdata/mpd/thisav.mpd
+++ b/test/testdata/mpd/thisav.mpd
@@ -1,0 +1,141 @@
+<?xml version="1.0"?>
+<!-- MPD file Generated with GPAC version 0.6.2-DEV-rev604-g600e98a-master  at 2017-10-24T08:10:10.986Z-->
+<MPD xmlns="urn:mpeg:dash:schema:mpd:2011" minBufferTime="PT1.500S" type="static" mediaPresentationDuration="PT0H10M0.192S" maxSegmentDuration="PT0H0M5.005S" profiles="urn:mpeg:dash:profile:full:2011">
+ <ProgramInformation moreInformationURL="http://gpac.sourceforge.net">
+  <Title>300708</Title>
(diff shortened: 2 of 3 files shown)


jaykbull:
@glogachov: Please feel free to test this branch and see if that works as expected for you.

dstftw:
It's a hack that will only work properly in some particular cases. What should be done is a proper implementation of `mediaRange` support. In any case I won't accept any changes to `_parse_mpd_formats` without [tests](https://github.com/rg3/youtube-dl/blob/master/test/test_InfoExtractor.py#L492).

jaykbull:
> It's a hack that will only work properly in some particular cases.

It's a hack. Granted. But as a hack it's also:

* small
* simple
* ultra-local

Right now, without these patches, this code-section has the following state-space. It either:

1. works or
2. completely errors out and crashes.

This simple hack changes this to:

1. still works exactly like before or
2. may actually give the desired result on formerly unsupported input.

From a behavioural perspective, that's clearly an *improvement* for a relatively low cost.

> What should be done is a proper implementation of `mediaRange` support.

In an ideal world sure. I won't be the one providing support for that though. I don't have *near* enough time to research and implement that.

If know for a fact that you or someone else is working on that, awesome. But in the meantime it's been 3 months since the issue was filed, and from what I can see zero activity towards resolving it.

In the face of that, I'll just assume users prefers the approach used in this PR over nothing.

> In any case I won't accept any changes to `_parse_mpd_formats` without tests.

100% fair and thanks for the pointer. I can totally respect that.

Which leaves me the following question: *If* I add such tests for for the cases in the linked issue (and make them go green), will you be willing to accept this patch? Or should I just abandon any such effort because this approach does not represent a "proper" solution"?

jaykbull:
I've added some tests, and I think I also found a bug in `_parse_mpd_formats` in the process, which I've corrected.

Is this better? Is this mergable?

jaykbull:
@dstftw: I've given good and exhaustive response to your comments, and I've added tests to the areas you asked for in the PR.

Should I consider this PR effectively closed, or is it something more you would like me to amend?

Feedback would really be appreciated at this point.

Hrxn:
Looks like the poor youtube-dl devs/maintainers are dealing with some serious backlog...
Patience for all parties involved is advised 😄

jaykbull:
One month and still no feedback...

I'd love to at least get some clarifications on if this PR has any merit or not. If there's something more I can fix, or if I should just close it.

cc: @Hrxn @dstftw

jaykbull:
After this PR has been lingering unattended for 3 months, with no feedback, I see there now is a conflict introduced by https://github.com/rg3/youtube-dl/pull/14844.

From my glancing, it seems the conflict is about a test handling similar test-cases. Was that PR supposed to fix https://github.com/rg3/youtube-dl/issues/13784? Because from my testing, it still seems broken.

As a side-note:

As an end-user I've always looked at `youtube-dl` as an amazing and impressive project. I'm a big proponent of open-source, and I've sent patches and contributions to Microsoft, Mozilla, Cyanogenmod, Cmake and even Emacs, but I've never seen any project as unresponsive as this.

This complete unwillingness to even *comment* on a proposed patch after months to end is certainly making me a bit jaded about any further contributions. It's not sending a good signal to future contributors.

che0:
#14844 was supposed to implement a new extractor, as described in #14616. As you can see that one took more than three months to merge and was rewritten several times per requests of the maintainer. If you want him to merge and support your code, my advice would be to refer to the feedback above:

> It's a hack that will only work properly in some particular cases. What should be done is a proper implementation of `mediaRange` support.

PR #14744, opened 2017-11-14
Support --ffmpeg-format

johnhawkinson:
(previously titled **Support --output-format (and -O)**)

Allow specifying an output format when using an external downloader, either through `--output-format` or `-O`, in preference to the extractor's default. This is particularly helpful as `-O mkv`, to allow Matroska
output from `ffmpeg` in cases where .mp4 isn't a great answer, such as while trying to play during the download.

---

- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

I frequently run start playing downloads while they are running, and .mp4 written by `ffmpeg` is often problematic. Players (usually VLC) have trouble seeking and even more trouble seeking backwards. Using mkv seems to produce a much better result, presumably because there is no issue with missing `moov` atoms.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -253,7 +253,6 @@ class YoutubeDL(object):
                        Progress hooks are guaranteed to be called at least once
                        (with status "finished") if the download is successful.
     merge_output_format: Extension to use when merging formats.
-    outputformat:      Extension and format to use when using an external downloader
     fixup:             Automatically correct known faults of the file.
                        One of:
                        - "never": do nothing
@@ -299,7 +298,7 @@ class YoutubeDL(object):
     the downloader (see youtube_dl/downloader/common.py):
     nopart, updatetime, buffersize, ratelimit, min_filesize, max_filesize, test,
     noresizebuffer, retries, continuedl, noprogress, consoletitle,
-    xattr_set_filesize, external_downloader_args, hls_use_mpegts.
+    xattr_set_filesize, external_downloader_args, hls_use_mpegts, ffmpeg_format.
 
     The following options are used by the post processors:
     prefer_ffmpeg:     If True, use ffmpeg instead of avconv if both are available,
@@ -636,8 +635,9 @@ class YoutubeDL(object):
                     template_dict['resolution'] = '%sp' % template_dict['height']
                 elif template_dict.get('width'):
                     template_dict['resolution'] = '%dx?' % template_dict['width']
-            if self.params.get('outputformat'):
-                template_dict['ext'] = self.params.get('outputformat')
+            # xxx: this can effect non-ffmpeg downloaders
+            if self.params.get('ffmpeg_format'):
+                template_dict['ext'] = self.params.get('ffmpeg_format')
 
             sanitize = lambda k, v: sanitize_filename(
                 compat_str(v),
@@ -1867,8 +1867,9 @@ class YoutubeDL(object):
                         if filename_real_ext == info_dict['ext']
                         else filename)
                     requested_formats = info_dict['requested_formats']
-                    if self.params.get('outputformat'):
-                        info_dict['ext'] = self.params.get('outputformat')
+                    # xxx: this can effect non-ffmpeg downloaders
+                    if self.params.get('ffmpeg_format'):
+                        info_dict['ext'] = self.params.get('ffmpeg_format')
                     if self.params.get('merge_output_format') is None and not compatible_formats(requested_formats):
                         info_dict['ext'] = 'mkv'
                         self.report_warning(
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -253,6 +253,7 @@ class YoutubeDL(object):
                        Progress hooks are guaranteed to be called at least once
                        (with status "finished") if the download is successful.
     merge_output_format: Extension to use when merging formats.
+    outputformat:      Extension and format to use when using an external downloader
     fixup:             Automatically correct known faults of the file.
                        One of:
                        - "never": do nothing
@@ -635,6 +636,8 @@ class YoutubeDL(object):
                     template_dict['resolution'] = '%sp' % template_dict['height']
                 elif template_dict.get('width'):
                     template_dict['resolution'] = '%dx?' % template_dict['width']
+            if self.params.get('outputformat'):
+                template_dict['ext'] = self.params.get('outputformat')
 
             sanitize = lambda k, v: sanitize_filename(
                 compat_str(v),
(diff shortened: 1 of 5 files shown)


johnhawkinson:
Err, and to clarify, `--output-format mkv` is different from `--merge-output-format mkv` because the latter can't start until the downloads are complete, making it essentially useless for playing-while-downloading (aka playing on-the-fly). And also, of course, running `ffmpeg` just once with the desired container format is faster and cheaper than wirting to an intermediate format and then converting.

dstftw:
1. Option name should explicitly reflect a dependency on ffmpeg in its name.
2. Description should explicitly emphasize it will only work with ffmpeg downloader.
3. `-O` should be removed.
4. It should be ignored completely when ffmpeg is not used as external downloader.

johnhawkinson:
> 4\. It should be ignored completely when ffmpeg is not used as external downloader.

I will work to rework this branch to make that change.

> 2\. Description should explicitly emphasize it will only work with ffmpeg downloader.

I had anticipated that it was possible to choose an output format for other downloaders, either now or in the future. Sounds like that's not going to be the case. (Upon review, it looks like `rtmpdump` can only write .flv files, and `youtube-dl`'s native downloaders are not going to gain support for non-mp4 downloaders in other cases). Of course it would work with `avconv`.

> 1\. Option name should explicitly reflect a dependency on ffmpeg in its name.
> 3\. -O should be removed.

Given the existing similar options, the set of prior art is something like this:

```
    --hls-prefer-ffmpeg              Use ffmpeg instead of the native HLS downloader
    --external-downloader COMMAND    Use the specified external downloader. Currently supports aria2c,avconv,axel,curl,ffmpeg,httpie,wget
    --audio-quality QUALITY          Specify ffmpeg/avconv audio quality, insert a value between 0 (better) and 9 (worse) for VBR or a specific bitrate like 128K (default 5)
    --prefer-ffmpeg                  Prefer ffmpeg over avconv for running the postprocessors
    --ffmpeg-location PATH           Location of the ffmpeg/avconv binary; either the path to the binary or its containing directory.
    --external-downloader COMMAND    Use the specified external downloader. Currently supports aria2c,avconv,axel,curl,ffmpeg,httpie,wget
    --external-downloader-args ARGS  Give these arguments to the external downloader
```
I note that `--audio-quality` likely falls in the same category (ffmpeg/avconv only) yet doesn't have `ffmpeg` in the name.

This leaves us with something like `--ffmpeg-output-format` or maybe `--external-downloader-output-format` (which doesn't meet your standard), both of which are long. I guess we could shorten to `--ffmpeg-format` or `--external-downloader-format` all of which aren't short.

What's the standard for whether an option merits a short alias? Does it have to be, e.g., demonstrated frequent use after it has been in the tree for some period of time?

I tend to want short options for anything I expect would be used interactively, whereas options only used in scripts can be long only. Is that the wrong way to look at it?


Thanks.

---

With respect to

> 4\. It should be ignored completely when ffmpeg is not used as external downloader.

You're right, as currently implemented this does too much. For instance:

```
pb3:youtube_dl jhawk$ PYTHONPATH=~/src/youtube-dl python -m youtube_dl -v --no-part -O mkv 'http://media.w3.org/2010/05/sintel/trailer.mp4'
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'-v', u'--no-part', u'-O', u'mkv', u'http://media.w3.org/2010/05/sintel/trailer.mp4']
[debug] youtube-dl version 2017.11.06
[debug] Git HEAD: a906d47bd
[debug] Python version 2.7.10 - Darwin-14.5.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg git-2017-02-28-7f62368, ffprobe git-2017-02-28-7f62368, rtmpdump 2.4
[debug] Proxy map: {}
[generic] trailer: Requesting header
[debug] Default format spec: bestvideo+bestaudio/best
[debug] Invoking downloader on u'http://media.w3.org/2010/05/sintel/trailer.mp4'
[download] Destination: trailer-trailer.mkv
[download] 100% of 4.17MiB in 00:01
pb3:youtube_dl jhawk$ file trailer-trailer.mkv 
trailer-trailer.mkv: ISO Media, MPEG v4 system, version 1
```

There we download an MP4 file and save it with an MKV extension, but the resulting file is not really an MKV file, it just pretends to be one. And that's not good.

johnhawkinson:
>  4. It should be ignored completely when ffmpeg is not used as external downloader.

So, @dstftw, I'm not sure the right way to get here. Filename selection occurs in `prepare_filename()` which substitutes according to the template. Choice of extension seems to need to happen within (or prior to) this function.

But `prepare_filename()` is called in `YoutubeDL.py`, out of scope from where the choice of downloader occurs, and (I think) before the downloader has been selected.

What is the right way to do as you suggest and ignore this option, given that sequencing?

Or is it sufficient, with the option now named `--ffmpeg_format` that if it is specified when ffmpeg is not the downloader, an extension change could happen?

Guidance appreciated.

johnhawkinson:
Any suggestions or advice, please?

dstftw:
>I had anticipated that it was possible to choose an output format for other downloaders, either now or in the future. Sounds like that's not going to be the case.

Well, downloader is generally something that simply downloads and not downloads and transfers to some other format on the fly. I doubt there are currently supported downloaders capable of this apart from ffmpeg/avconv. Assuming there may appear another one capable you may omit ffmpeg from option's name but there should be something that will emphasize that the process is happening on the fly, as for me `--output-format` is too ambiguous.

>What's the standard for whether an option merits a short alias? Does it have to be, e.g., demonstrated frequent use after it has been in the tree for some period of time?

Option should be frequently usable by wide range of users. This option is going to be used by very few people since it will only work for ffmpeg downloader that is rarely used. There are also some other options proposed for `-O`.

johnhawkinson:
@dstftw, when I asked for help, I was really worried about the implementation problems I ran into when trying to follow your direction, i.e. https://github.com/rg3/youtube-dl/pull/14744#issuecomment-344489844

I appreciate your answering the earlier questions, but they were curiosity -type questions. They weren't preventing me from writing code (and, indeed, a4e49f1 was pushed after I asked them).

So, again, I'm still unclear how to avoid doing anything for non-ffmpeg downloaders when the filename selection happens in generic code:

> Filename selection occurs in prepare_filename() which substitutes according to the template. Choice of extension seems to need to happen within (or prior to) this function.

> But prepare_filename() is called in YoutubeDL.py, out of scope from where the choice of downloader occurs, and (I think) before the downloader has been selected.

> What is the right way to do as you suggest and ignore this option, given that sequencing?

Thanks.

dstftw:
In `prepare_filename` you have an access to `self.params.get('external_downloader')` so you can check for ffmpeg. You can also check whether it is going to be selected as downloader later (based on `get_external_downloader` and `can_download`).

PR #14763, opened 2017-11-16
[openload] Support [www.]openload.link URLs and make flake8 pass

jahudka:

---

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Allows a broader range of Openload links to be understood by the _VALID_URL regex and fixes the only error raised by flake8, which complained about using bare `except` at line 143.

### DIFF
--- a/youtube_dl/extractor/openload.py
+++ b/youtube_dl/extractor/openload.py
@@ -140,7 +140,7 @@ class PhantomJSwrapper(object):
         for name in self._TMP_FILE_NAMES:
             try:
                 os.remove(self._TMP_FILES[name].name)
-            except OSError:
+            except (IOError, OSError):
                 pass
 
     def _save_cookies(self, url):
--- a/youtube_dl/extractor/openload.py
+++ b/youtube_dl/extractor/openload.py
@@ -286,6 +286,9 @@ class OpenloadIE(InfoExtractor):
     }, {
         'url': 'https://oload.tv/embed/KnG-kKZdcfY/',
         'only_matching': True,
+    }, {
+        'url': 'http://www.openload.link/f/KnG-kKZdcfY',
+        'only_matching': True,
     }]
 
     _USER_AGENT = 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/56.0.2924.87 Safari/537.36'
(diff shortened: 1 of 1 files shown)


yan12125:
The commit is authored by @mcjahudka. Why?

jahudka:
@yan12125 @mcjahudka is my old github account but it's tied to my current e-mail address, so commits pushed from my laptop end up being attributed to my old account (because my git installation is configured with my current e-mail address). I've now corrected both my accounts settings and it looks as though GitHub propagated them even here, so the commit now says "committed by jahudka"; but they're both me anyway. Sorry about the confusion..

yan12125:
No problem. Could you add an example link to ```_TESTS``` and marking it with ```only_matching```?

youtube_dl/extractor/openload.py
```diff
@@ -140,7 +140,7 @@ def __del__(self):
         for name in self._TMP_FILE_NAMES:
             try:
                 os.remove(self._TMP_FILES[name].name)
-            except:
+            except (OSError, TypeError):
```
dstftw: What's the point of `TypeError` here?

yan12125:
http://www.openload.link/f/KnG-kKZdcfY is dead from here ("Error 502 Bad gateway"). Does it work for you?

youtube_dl/extractor/openload.py
jahudka: `os.remove()` throws TypeError if the passed file path is `None`; the original implementation with a bare `except` would've caught that, so including it was in the interest of not breaking anything while making the code flake8-compliant

dstftw: `self._TMP_FILES[name].name` can't be `None`.

jahudka:
@yan12125 yes, openload.link as a whole has been having some issues recently, but the same content is usually available from openload.co - you just have to manually change the address. The extractor doesn't care about the original page anyway - it just extracts the ID from the URL and then uses a normalized URL to load just the content (or so I understand from reading the code). So the fact that openload.link is currently unavailable wouldn't prevent the extractor from downloading content hosted there.

youtube_dl/extractor/openload.py
jahudka: How so? Sorry, I'm no Python guru - is that some property of dictionaries (or whichever type `self._TMP_FILES` is)?

jahudka: Ah, I see, `_TMP_FILES` is a dictionary, but it only contains `tempfile.NamedTemporaryFile` instances, so it should be safe to assume that `self._TMP_FILES[name].name` won't ever be none.. Okay, I'll push a fix.

Issue #14472, opened 2017-10-11
Vidlii Support (Trying A Post Again)


dkl3:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like this: `[x]`)
- Use the *Preview* tab to see what your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2017.10.12*. If it's not, read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2017.10.12**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [ ] Bug report (encountered problems with youtube-dl)
- [x] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

~$ youtube-dl http://www.vidlii.com/watch?v=vBo2IcrwOkO -v --write-info-json
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['http://www.vidlii.com/watch?v=vBo2IcrwOkO', '-v', '--write-info-json']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2017.10.12
[debug] Python version 3.5.2 - Linux-4.10.0-35-generic-i686-with-LinuxMint-18.2-sonya
[debug] exe versions: ffmpeg 2.8.11-0ubuntu0.16.04.1, ffprobe 2.8.11-0ubuntu0.16.04.1
[debug] Proxy map: {}
[...]

PR #14779, opened 2017-11-17
[vidlii] Add new extractor

patrickj0607:

- [ ] Bug fix
- [ ] Improvement
- [X] New extractor
- [ ] New feature

---

This pull request implements the extractor for vidlii.com, see issue #14472. Further, we implemented two test cases.


svnpenn:
https://github.com/rg3/youtube-dl/issues/13883

svnpenn:
https://github.com/rg3/youtube-dl/pull/14512

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,152 @@
...
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        # extract basic properties of video
+        title = str_or_none(
+            self._html_search_regex(r'<h1>(.+?)</h1>', webpage,
+                                    'title', default=None)) or str_or_none(
+            self._html_search_regex(r'<title>([^<]+?)</title>', webpage,
+                                    'title', default=None)).replace(
```
dstftw: Breaks on `None`.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,152 @@
...
+            " - VidLii", "") or str_or_none(
+            self._html_search_meta('twitter:title', webpage, 'title',
+                                   default=False)).replace(" - VidLii", "")
```
dstftw: Breaks on `default=False`.
`title` must be fatal.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,152 @@
...
+
+        description = strip_or_none(
+            get_element_by_id('des_text', webpage).strip())
```
dstftw: Stripped twice. Breaks on `None`.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,152 @@
...
+
+        uploader = str_or_none(
+            self._html_search_regex(
+                r'<div[^>]+class="wt_person"[^>]*>(?:[^<]+)<a href="\/user\/[^>]*?>([^<]*?)<',
+                webpage,
+                'uploader', default=None)) or str_or_none(
+            self._html_search_regex(
+                r'<img src="[^>]+?class=["\']avt2\s*["\'][^>]+?alt=["\']([^"\']+?)["\']',
+                webpage, 'uploader', default=None))
```
dstftw: Code duplication. Must be single call to search regex.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,152 @@
...
+            webpage, 'url', default=None)
+
+        # get additional properties
+        uploader_url = "https://www.vidlii.com/user/%s" % uploader
+
+        # returns date as YYYYMMDD
+        upload_date = str_or_none(
+            self._html_search_meta('datePublished', webpage, 'upload_date',
+                                   default=False).replace("-",
+                                                          ""))
```
dstftw: Same.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,152 @@
...
+        view_count = int_or_none(self._html_search_regex(
+            r'Views:[^<]*<strong>([^<]*?)<\/strong>', webpage,
+            'view_count', default=None)) or int_or_none(view_count_fb)
+
+        comment_count = int_or_none(self._html_search_regex(
+            r'Comments:[^<]*<strong>([^<]*?)<\/strong>', webpage,
+            'comment_count', default=None)) or int_or_none(
+            self._html_search_regex(
+                r'<span[^>]+id="cmt_num"[^>]*>([^<]+?)<\/span>', webpage,
+                'comment_count', default=None))
```
dstftw: Code duplication.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,152 @@
...
+
+        average_rating = float_or_none(
+            self._html_search_regex(
+                r'{[\s\r]*\$\("#rateYo"\).rateYo\({[^}]*rating:\s*([^,]*?),[^}.]*}',
+                webpage, 'average_rating', default=None))
+        thumbnail_link = self._html_search_regex(
+            r'videoInfo[\s]*=[\s]*{[^}]*img:[\s]*(?:"|\')([^"]*?)(?:"|\')',
+            webpage, 'thumbnail', default=None)
+        thumbnail = 'https://www.vidlii.com%s' % thumbnail_link
```
dstftw: `urljoin`.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Replaced ```default=None``` by an empty string (```default=""```) such that ```replace``` operation can always be performed .

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Changed ```default=False``` to ```default=""```.  See last suggestion.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Removed ```strip()``` call as it is duplicate.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Removed ```str_or_none(...)```. Also changed on all other ```self._html_search_regex(...)``` calls.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Removed ```str_or_none(...)``` and replaced ```default=False``` by ```default=""``` such that ```replace``` can always take place.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Replaced second call of ```int_or_none``` by moving or-expression into brackets.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Use of ```urljoin``` from utils.py. Same changes made for "uploader" attribute.

youtube_dl/extractor/vidlii.py
dstftw: No. `title` is mandatory. Read coding conventions and fix appropriately.

patrickj0607:
Hi dstftw, thanks for your feedback. We just implemented your suggested improvements. We would appreciate a short feedback! Thank you.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,129 @@
...
+            webpage, 'uploader', default=None)
+
+        url = self._html_search_regex(r'videoInfo[\s]*=[\s]*{[^}]*src:[\s]*(?:"|\')([^"]*?)(?:"|\')', webpage, 'url',
+                                      default=None)
+
+        # get additional properties
+        uploader_url = urljoin("https://www.vidlii.com/user/", uploader)
+
+        # returns date as YYYYMMDD
+        upload_date = self._html_search_meta('datePublished', webpage, 'upload_date', default="").replace("-", "")
```
dstftw: No.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        # extract basic properties of video
+        title = self._html_search_regex(r'<h1>(.+?)</h1>', webpage, 'title', default=None) or self._html_search_regex(
+            r'<title>([^<]+?)</title>', webpage, 'title', default="").replace(
+            " - VidLii", "") or self._html_search_meta('twitter:title', webpage, 'title',
+                                                       default="").replace(" - VidLii", "")
```
dstftw: No. `title` is mandatory. Code duplication. Move all `_html_search_regex` into a single call.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,129 @@
...
+
+        description = strip_or_none(get_element_by_id('des_text', webpage))
+
+        uploader = self._html_search_regex(r'<div[^>]+class="wt_person"[^>]*>(?:[^<]+)<a href="\/user\/[^>]*?>([^<]*?)<',
+                                           webpage, 'uploader', default=None) or self._html_search_regex(
+            r'<img src="[^>]+?class=["\']avt2\s*["\'][^>]+?alt=["\']([^"\']+?)["\']',
+            webpage, 'uploader', default=None)
```
dstftw: Same.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,129 @@
...
+
+        url = self._html_search_regex(r'videoInfo[\s]*=[\s]*{[^}]*src:[\s]*(?:"|\')([^"]*?)(?:"|\')', webpage, 'url',
```
dstftw: Don't shadow outer names. No need for bracket when using single character.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # returns date as YYYYMMDD
+        upload_date = self._html_search_meta('datePublished', webpage, 'upload_date', default="").replace("-", "")
+
+        categories = self._html_search_regex(
+            r'<div>Category:\s*<\/div>[\s\r]*<div>[\s\r]*<a href="\/videos\?c=[^>]*>([^<]*?)<\/a>', webpage, 'categories',
+            default=None)
+        tags = re.findall(r'<a href="/results\?q=[^>]*>[\s]*([^<]*)</a>', webpage) or None
+        duration = int_or_none(self._html_search_meta('video:duration', webpage, 'duration', default=False)) or int_or_none(
+            self._html_search_regex(r'videoInfo[^=]*=[^{]*{[^}]*dur:([^,}]*?),', webpage, 'duration', default=None))
```
dstftw: Code duplication.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # extract basic properties of video
+        title = self._html_search_regex(r'<h1>(.+?)</h1>', webpage, 'title', default=None) or self._html_search_regex(
+            r'<title>([^<]+?)</title>', webpage, 'title', default="").replace(
+            " - VidLii", "") or self._html_search_meta('twitter:title', webpage, 'title',
+                                                       default="").replace(" - VidLii", "")
+
+        description = strip_or_none(get_element_by_id('des_text', webpage))
+
+        uploader = self._html_search_regex(r'<div[^>]+class="wt_person"[^>]*>(?:[^<]+)<a href="\/user\/[^>]*?>([^<]*?)<',
```
dstftw: No escapes for `/`.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,129 @@
...
+                                           webpage, 'uploader', default=None) or self._html_search_regex(
+            r'<img src="[^>]+?class=["\']avt2\s*["\'][^>]+?alt=["\']([^"\']+?)["\']',
```
dstftw: Matching between quotes is incorrect.

youtube_dl/extractor/vidlii.py
dstftw: Only whitespace is allowed between `videoInfo` and `=`.

dstftw: `_parse_json`.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,129 @@
...
+
+        view_count_fb = re.findall(r'<strong>([^<]*?)</strong>', get_element_by_class("w_views", webpage))
+        view_count_fb = view_count_fb[0] if view_count_fb else None
+        view_count = int_or_none(self._html_search_regex(r'Views:[^<]*<strong>([^<]*?)<\/strong>', webpage, 'view_count',
+                                                         default=None)) or int_or_none(view_count_fb)
+
+        comment_count = int_or_none(
+            self._html_search_regex(r'Comments:[^<]*<strong>([^<]*?)<\/strong>', webpage, 'comment_count',
+                                    default=None) or
+            self._html_search_regex(r'<span[^>]+id="cmt_num"[^>]*>([^<]+?)<\/span>', webpage, 'comment_count', default=None))
```
dstftw: Code duplication.

youtube_dl/extractor/vidlii.py
patrickj0607: Please tell us exactly how we should handle it:
a) We combine the regular expressions into one single call, then we cannot make the replacement anymore such that tests would fail if any fallback is used. Because fallbacks contain the title plus the suffix " - VidLii".
b) We change the order such that our first extractor (without ```replace```) now is used as the last one. This means, if the first two extractors cannot find any title, None is returned, indicating that a mandatory field is missing. In this case the ```self._html_search_regex``` calls are not merged into a single one.
c) ..do you have any better suggestion?

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Removed ```/```

youtube_dl/extractor/vidlii.py
patrickj0607: Could you kindly explain why it is incorrect or provide us an example where it does not work. Thanks!

youtube_dl/extractor/vidlii.py
patrickj0607: Cannot use ```_parse_json``` here because the information on the video site is no proper JSON, it is a Javascript dictionary (keys do not have required ```"```). 

FIXED: Code duplication removed, changed RE to only allow whitespace characters

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Merged REs into single expression / function call.

youtube_dl/extractor/vidlii.py
dstftw: `js_to_json`.

youtube_dl/extractor/vidlii.py
dstftw: `'` can be used between `"` quotes and vice versa.

youtube_dl/extractor/vidlii.py
patrickj0607: Could you kindly provide us more information. We checked the following:
* ```uploader_url``` uses ```urljoin``` as suggested by you
* Date format of ```upload_date``` is like others extractors use, too.

Should we maybe use "relaseDate" instead of "datePublished" or what are you referring to?

dstftw: Lack of information is denoted by `None` not an empty string. Also `unified_strdate`.

dstftw: Also do not use double quotes.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Used ```js_to_json``` and ```_parse_json``` to convert ```videoInfo``` into dictionary.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Added another fallback for upload_date and added ```unified_strdate``` for conversion into right date format. Thanks for the hint, we didn't see that there exists a helper funtion therefor.

youtube_dl/extractor/vidlii.py
patrickj0607: Sorry, we don't get the point here. We are now gathering the url from the videoInfo dictionary:
```
videoInfo_dict = self._parse_json(js_to_json(self._html_search_regex(r'var videoInfo\s*=\s*({[^}]*})', webpage, 'videoInfo', fatal=False)), video_id)
url = videoInfo_dict.get("src")
```
What do you mean by "Don't shadow outer names"? Thank you!

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Merged both REs into single function call. Added compatbility for use of quotes within value, as suggested by you.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Merged both REs into single function call. Added compatbility for use of quotes within value, as suggested by you.

youtube_dl/extractor/vidlii.py
patrickj0607: Please tell us exactly how we should handle it:

a) We combine the regular expressions into one single call, then we cannot make the replacement anymore such that tests would fail if any fallback is used. Because fallbacks contain the title plus the suffix " - VidLii". That's the reason why we remove the suffix.

b) We change the order such that our first extractor (without replace) now is used as the last one. This means, if the first two extractors cannot find any title, ```None``` is returned, indicating that a mandatory field is missing. In this case the ```self._html_search_regex``` calls cannot be merged into a single one. See following code:

```
        title = (self._html_search_regex(r'<title>([^<]+?)</title>', webpage, 'title', default="") 
or self._html_search_meta('twitter:title', webpage, 'title', default="")).replace(" - VidLii","") 
or self._html_search_regex(r'<h1>(.+?)</h1>', webpage, 'title', default=None)
```

c) ..do you have any better suggestion?

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Renamed ```url``` to ```url_video```, simplified RE

youtube_dl/extractor/vidlii.py
patrickj0607: We now implemented solution b). If you do not like it, please tell us better suited alternatives. Thank you for your helpful feedback!

patrickj0607:
Hi Sergey, do we have to fix anything else? Thanks for your feedback, we really appreciate it!

patrickj0607:
Hi @dstftw, as we are implementing this feature in a software development course, it would be great seeing it to get accepted till end of the week. Do you have any other suggestion for code improvement?

patrickj0607:
Hi @yan12125, as dsftw is not responding anymore, would you kindly provide me feedback to my changes? Thank you very much!

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        # extract videoInfo variable for further use
+        videoInfo_dict = self._parse_json(js_to_json(self._html_search_regex(r'var videoInfo\s*=\s*({[^}]*})', webpage,
+                                                                             'videoInfo', fatal=False)), video_id)
```
dstftw: Must be fatal.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+            'ext': 'mp4'
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        # extract videoInfo variable for further use
+        videoInfo_dict = self._parse_json(js_to_json(self._html_search_regex(r'var videoInfo\s*=\s*({[^}]*})', webpage,
```
dstftw: Use whitespace characters consistently.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+
+from .common import InfoExtractor
+from ..utils import (
+    int_or_none,
+    get_element_by_id, get_element_by_class, strip_or_none,
+    float_or_none, urljoin, js_to_json, unified_strdate)
+
+
+class VidliiIE(InfoExtractor):
+    _VALID_URL = r'(?:https*?:\/\/)*(?:www\.)*vidlii.com\/watch\?v=(?P<id>[^?\s]{11})'
```
dstftw: `[^?\s]` does not make any sense. You already matched `?` previously. I've already pointed out - no escapes for `/`.
`(?:https*?:\/\/)*` does not make any sense. `s*` does not make any sense. `(?:www\.)*` does not make any sense.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+    _TESTS = [{
+        'url': 'https://www.vidlii.com/watch?v=tJluaH4BJ3v',
+        'md5': '9bf7d1e005dfa909b6efb0a1ff5175e2',
+        'info_dict': {
+            'id': 'tJluaH4BJ3v',
+            'title': 'Vidlii is against me',
+            'description': 'I have HAD it. Vidlii does not like me. I have tried to uplaod videos and submit them to the '
+                           'contest and no ne of my videos show up so maybe it is broken for everyone else but this one was '
+                           'trying to submit it because I wanted to submit to the contest :) Tanks I hope the website is '
+                           'fixed PS: Jan you are cool please add my video',
```
dstftw: `md5:...`.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+            'thumbnail': 'https://www.vidlii.com/usfi/thmp/tJluaH4BJ3v.jpg',
```
dstftw: No exact URLs.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+            'uploader': 'APPle5auc31995',
+            'url': 'https://cdn.vidlii.com/videos/tJluaH4BJ3v.mp4',
```
dstftw: No exact URLs.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+            'tags': ['Vidlii', 'Jan', 'Videogames'],
+            'duration': 212,
+            'view_count': int,
+            'comment_count': int,
+            'average_rating': float,
+            'type': 'video',
+            'ext': 'mp4'
+        }
+    }, {
+        'url': 'https://www.vidlii.com/watch?v=vBo2IcrwOkO',
```
dstftw: Remove all duplicate tests.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+            'uploader_url': uploader_url,
+            'upload_date': upload_date,
+            'categories': categories,
+            'tags': tags,
+            'duration': duration,
+            'view_count': view_count,
+            'comment_count': comment_count,
+            'average_rating': average_rating,
+            'thumbnail': thumbnail,
+            'type': video_type
```
dstftw: No such metafield.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+
+        # extract videoInfo variable for further use
+        videoInfo_dict = self._parse_json(js_to_json(self._html_search_regex(r'var videoInfo\s*=\s*({[^}]*})', webpage,
+                                                                             'videoInfo', fatal=False)), video_id)
+
+        # extract basic properties of video
+        title = (self._html_search_regex(r'<title>([^<]+?)</title>', webpage, 'title', default='') or
+                 self._html_search_meta('twitter:title', webpage, 'title', default='')).replace(" - VidLii",
+                                                                                                "") or self._html_search_regex(
+            r'<h1>(.+?)</h1>', webpage, 'title', default=None)
```
dstftw: I've already pointed out - title **must be fatal**.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+
+        description = strip_or_none(get_element_by_id('des_text', webpage))
+
+        uploader = self._html_search_regex(
+            r'<div[^>]+class="wt_person"[^>]*>(?:[^<]+)<a href="/user/[^>]*?>([^<]*?)<|<img src="[^>]+?class=('
+            r'?:"avt2\s*"|\'avt2\s*\')[^>]+?alt=(?:"([^"]+?)"|\'([^\']+?)\')>', webpage, 'uploader', default=None,
+            fatal=False)
```
dstftw: `default` and `fatal` are not used together.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+
+        video_url = videoInfo_dict.get("src")
```
dstftw: Must be fatal.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        # extract videoInfo variable for further use
+        videoInfo_dict = self._parse_json(js_to_json(self._html_search_regex(r'var videoInfo\s*=\s*({[^}]*})', webpage,
+                                                                             'videoInfo', fatal=False)), video_id)
+
+        # extract basic properties of video
+        title = (self._html_search_regex(r'<title>([^<]+?)</title>', webpage, 'title', default='') or
+                 self._html_search_meta('twitter:title', webpage, 'title', default='')).replace(" - VidLii",
```
dstftw: Single quotes.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+
+        categories = self._html_search_regex(
+            r'<div>Category:\s*<\/div>[\s\r]*<div>[\s\r]*<a href="\/videos\?c=[^>]*>([^<]*?)<\/a>', webpage, 'categories',
+            default=None, fatal=False)
+        tags = re.findall(r'<a href="/results\?q=[^>]*>\s*([^<]*)</a>', webpage) or None
+        duration = int_or_none(
+            self._html_search_meta('video:duration', webpage, 'duration', default=False, fatal=False) or
+            videoInfo_dict.get("dur"))
+
+        view_count_fb = re.findall(r'<strong>([^<]*?)</strong>', get_element_by_class("w_views", webpage))
```
dstftw: Breaks when `get_element_by_class` returns `None`.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+        title = (self._html_search_regex(r'<title>([^<]+?)</title>', webpage, 'title', default='') or
+                 self._html_search_meta('twitter:title', webpage, 'title', default='')).replace(" - VidLii",
+                                                                                                "") or self._html_search_regex(
+            r'<h1>(.+?)</h1>', webpage, 'title', default=None)
+
+        description = strip_or_none(get_element_by_id('des_text', webpage))
+
+        uploader = self._html_search_regex(
+            r'<div[^>]+class="wt_person"[^>]*>(?:[^<]+)<a href="/user/[^>]*?>([^<]*?)<|<img src="[^>]+?class=('
+            r'?:"avt2\s*"|\'avt2\s*\')[^>]+?alt=(?:"([^"]+?)"|\'([^\']+?)\')>', webpage, 'uploader', default=None,
```
dstftw: Pass as list of regexes, don't bulk.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,136 @@
...
+        view_count = int_or_none(self._html_search_regex(r'Views:[^<]*<strong>([^<]*?)<\/strong>', webpage, 'view_count',
+                                                         default=None, fatal=False)) or int_or_none(view_count_fb)
+
+        comment_count = int_or_none(
+            self._html_search_regex(r'Comments:[^<]*<strong>([^<]*?)<\/strong>|<span[^>]+id="cmt_num"[^>]*>(['
+                                    r'^<]+?)<\/span>', webpage, 'comment_count',
+                                    default=None, fatal=False))
+
+        average_rating = float_or_none(
+            self._html_search_regex(r'{[\s\r]*\$\("#rateYo"\).rateYo\({[^}]*rating:\s*([^,]*?),[^}.]*}',
```
dstftw: Should be more relaxed. Any minor change HTML layout of the first part will break it.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Now using ``md5`` from ``test.helper``.

youtube_dl/extractor/vidlii.py
patrickj0607: How should the matching be performed, only by considering the thumbnail's file extension (https://www.vidlii.com/*.jpg)?

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Removed as expected value for ``url`` already provided (duplicate).

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Removed ``url`` from ``info_dict``.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED.

youtube_dl/extractor/vidlii.py
patrickj0607: As far as I understood, setting *fatal=True* raises an Exception if the parameter could not be extracted. 

Above you can see something like ``title = get_from_source1 or get_from_source1 or get_from_source1``. If I set *fatal=True* for ``get_from_source1`` the extraction would stop if the title cannot be extracted, thus fallbacks ``get_from_source2`` and ``get_from_source3`` would never be used. How should this be handled? Should I check for a valid value in vidlii class itself?

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED.

youtube_dl/extractor/vidlii.py
patrickj0607: I do not understand what you mean. As described in class ``InfoExtractor``, the field ``uploader`` is optional, thus ``fatal = True`` is not required. 

If you refer to the ``videoInfo_dict``, the ``_html_search_regex`` for retrieving it already uses ``fatal = True``.

Please describe more precisely what you mean.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED: Removed metafield ``type``.

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED.

youtube_dl/extractor/vidlii.py
dstftw: You must make fatal the last source.

youtube_dl/extractor/vidlii.py
dstftw: Remove **this test completely**.

youtube_dl/extractor/vidlii.py
dstftw: See another extractors.

youtube_dl/extractor/vidlii.py
dstftw: This comment is not about uploader. You can't process without `video_url` so it's senseless to extract `src` with non-failing on missing keys `get`.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,127 @@
...
+        webpage = self._download_webpage(url, video_id)
+
+        # extract videoInfo variable for further use
+        videoInfo_dict = self._parse_json(js_to_json(self._html_search_regex(r'var\s*videoInfo\s*=\s*({[^}]*})', webpage,
+                                                                             'videoInfo', fatal=True)), video_id)
+
+        # extract basic properties of video
+        title = (self._html_search_regex(r'<title>([^<]+?)</title>', webpage, 'title', default='', fatal=True) or
+                 self._html_search_meta('twitter:title', webpage, 'title', default='', fatal=True)).replace(' - VidLii', '') \
+                or self._html_search_regex(r'<h1>(.+?)</h1>', webpage, 'title', default=None, fatal=True)
```
dstftw: `default` and `fatal` are not used together.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,127 @@
...
+            'ext': 'mp4'
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        # extract videoInfo variable for further use
+        videoInfo_dict = self._parse_json(js_to_json(self._html_search_regex(r'var\s*videoInfo\s*=\s*({[^}]*})', webpage,
```
dstftw: `varvideoInfo` is not possible, regex is invalid.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,127 @@
...
+                                                                             'videoInfo', fatal=True)), video_id)
```
dstftw: `fatal=True` is default.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,127 @@
...
+            default=None, fatal=False)
+        tags = re.findall(r'<a href="/results\?q=[^>]*>\s*([^<]*)</a>', webpage) or None
+        duration = int_or_none(
+            self._html_search_meta('video:duration', webpage, 'duration', default=False, fatal=False) or
+            videoInfo_dict.get('dur'))
+
+        view_count_fb = re.findall(r'<strong>([^<]*?)</strong>', get_element_by_class('w_views', webpage) or '')
+        view_count_fb = view_count_fb[0] if view_count_fb else None
+        view_count = int_or_none(self._html_search_regex(r'Views:[^<]*<strong>([^<]*?)</strong>', webpage, 'view_count',
+                                                         default=None, fatal=False)) or int_or_none(view_count_fb)
```
dstftw: View count extraction fits in two simple regexes similarly to comment count.

youtube_dl/extractor/vidlii.py
```diff
@@ -0,0 +1,127 @@
...
+                                   fatal=False) or self._html_search_regex(r'<date>(['r'^<]*?)</date>', webpage,
+                                                                           'upload_date',
+                                                                           default=None, fatal=False))
+
+        categories = self._html_search_regex(
+            r'<div>Category:\s*</div>[\s\r]*<div>[\s\r]*<a href="/videos\?c=[^>]*>([^<]*?)</a>', webpage, 'categories',
+            default=None, fatal=False)
+        tags = re.findall(r'<a href="/results\?q=[^>]*>\s*([^<]*)</a>', webpage) or None
+        duration = int_or_none(
+            self._html_search_meta('video:duration', webpage, 'duration', default=False, fatal=False) or
```
dstftw: `default=None`.

youtube_dl/extractor/vidlii.py
patrickj0607: How should the check be performed - by using ``videoInfo_dict['src']`` which automatically raises an exception if the key is not present?

youtube_dl/extractor/vidlii.py
patrickj0607: FIXED.

youtube_dl/extractor/vidlii.py
dstftw: Yes.

youtube_dl/extractor/vidlii.py
patrickj0607: Why? The regex is working, see https://regexr.com/3ikls

dstftw: Read my message.

PR #14847, opened 2017-11-25
Adjust max filename length to fit Docker NAME_MAX

handeln:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Docker / AUFS has a NAME_MAX of 242.

This adjusts the maximum to fit within that limit (237 + 5), allowing `--write-pages` to succeed.


dstftw:
This does not solve the conceptual problem. `NAME_MAX` may be arbitrary, e.g. it's 143 for encrypted FSes.

handeln:
I'm aware of that, but this tiny patch solves one large use case until a more ambitious solution is committed.

Personally, I think `--write-pages` and all other file-writing operations should respect `--output` filename templates. This would be the cleanest solution.

PR #14909, opened 2017-12-04
Fixed parsing

CB49:
Simple fixed a parsing error.

### DIFF
--- a/ip.php
+++ b/ip.php
@@ -1,3 +1,4 @@
 <?php
 header('Content-Type: text/plain');
 echo $_SERVER['REMOTE_ADDR'];
+?>


dstftw:
Closing tag at the end of the file is optional.

rredford6:
@CB49:

> The closing tag of a PHP block at the end of a file is optional, and in some cases omitting it is helpful when using include or require, so unwanted whitespace will not occur at the end of files, and you will still be able to add headers to the response later. It is also handy if you use output buffering, and would not like to see added unwanted whitespace at the end of the parts generated by the included files.

http://php.net/basic-syntax.instruction-separation

Issue #13089, opened 2017-05-14
Solving the CF issue - once and for all


EnginePod:
As some of you may know, CF has a mode called "I'm under attack" which shows a JS challenge.
A lot of sites are using this to prevent DDoS attacks, but it's accidentally also stopping youtube-dl since there's no function that solves the JS challenge.

It's caused me and other users a lot of annoyances (see: https://github.com/rg3/youtube-dl/issues/11572) since we're having to visit the page manually, copy the cookies from the browser to a text file and specify the location of the cookies and the exact user-agent in youtube-dl.  The cookie stops working if your IP changes (if you have a dynamic IP like mine), if the user agent changes (common when the browser is updated) or when the auth expires.

Now there is a solution written in Python called cfscrape (https://github.com/Anorov/cloudflare-scrape/blob/master/cfscrape/__init__.py), but it seems to be using node.js to solve the CF challenge which I think is overkill since the JS challenge can be parsed and solved manually.

---

So instead of having youtube-dl return a 503 error, it would:

1. Upon a 503 on the first request; check if the page is a Cloudflare challenge
- Can easily be done by checking the source for "challenge-form" && "cloudflare" (and a few other ways)

2. Extract the parameters from the HTML form:

```
  <form id="challenge-form" action="/cdn-cgi/l/chk_jschl" method="get">
    <input type="hidden" name="jschl_vc" value="2zd281b23ba38136ffa257335eccbbeb"/>
    <input type="hidden" name="pass" value="1493771132.19-OPc3Nc0Sh+"/>
    <input type="hidden" id="jschl-answer" name="jschl_answer"/>
  </form>
```
3. Construct the URL
`https://website-name.tld/cdn-cgi/l/chk_jschl?jschl_vc=2zd281b23ba38136ffa257335eccbbeb&pass=1493771132.59-OPc4Nn0Sh+&jschl_answer=%s`

4. Now only the `jschl_answer` is needed, which involves going through some obfuscated code:
```
var s, t, o, p, b, r, e, a, k, i, n, g, f, uMIoRrH = {
  "AL": +((!+[] + !![] + !![] + !![]) + (!+[] + !![] + !![] + !![] + !![] + !![] + !![]))
};
t = document.createElement('div');
t.innerHTML = "<a href='/'>x</a>";
t = t.firstChild.href;
r = t.match(/https?:\/\//)[0];
t = t.substr(r.length);
t = t.substr(0, t.length - 1);
[...]

PR #11730, opened 2017-01-15
Fixes #11572 Use cfscrape to avoid Crunchyroll error 503

kidburglar:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [ ] At least skimmed through [adding new extractor tutorial](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/rg3/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

Crunchyroll use not cloudfront when we login. To avoid a 503 error we need to insert the cookies that coudfront send.

I use cfscrape (https://github.com/Anorov/cloudflare-scrape) to grab the cookie that we need and insert them back. It is important that cfscrape have the same user-agent than youtube-dl, so it would be better not to hardcode it (I didn't find how I must do, so feel free to change it).

Issue #11572, opened 2017-01-01
Crunchyroll Login page throwing 503


Xonshiz:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like that [x])
- Use *Preview* tab to see how your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2016.12.22*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x ] I've **verified** and **I assure** that I'm running youtube-dl **2016.12.22**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add `-v` flag to **your command line** you run youtube-dl with, copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):
```
Youtube-DL>youtube-dl -i "http://www.crunchyroll.com/fairy-musketeers/episode-1-akazukin-finally-comes-510412" -u MyUserName -p MyPassword --verbose
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-i', 'http://www.crunchyroll.com/fairy-musketeers/episode-1-akazukin-finally-comes-510412', '-u', 'PRIVATE', '-p', 'PRIVATE', '--verbose']
[debug] Encodings: locale cp1252, fs mbcs, out cp437, pref cp1252
[debug] youtube-dl version 2016.12.31
[...]

PR #14943, opened 2017-12-10
[jsinterp] Partial unary support, NaN, string coercion rules

Tatsh:

---

- [x [Searched](https://github.com/rg3/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

This is a convenience adjustment and fix for CloudFlare DDoS-protected sites, for issues like #13089 and #11572.

This allows any extractor to answer to a JavaScript-based CloudFlare anti-DDoS challenge, so that users do not have to grab their cookies and use the `--cookies` option. This only needs to be done once in the `_real_extract()` method and since `.save_cookie()` is used, the cookies persist during runtime. This makes the first request take 5 seconds, but every subsequent request does not require the challenge as long as cookies are passed.

This patch, unlike #11730, does not add an extra dependency, but it does require a version `node` to be in `PATH`. We might be able to work around this but I would like feedback on the best way. The obfuscation CloudFlare does is just a series of `!` operators to empty array literals `[]` to create ones and zeros, as part of the challenge answer calculation. This is an example obfuscation:

```js
        var s,t,o,p,b,r,e,a,k,i,n,g,f,
            cOsBZOd={"DrzuOhdgdprf":!+[]+!![]};

        t = getSiteHostname();

        cOsBZOd.DrzuOhdgdprf += +((!+[]+!![]+!![]+[])+(+!![]));
        answer = parseInt(cOsBZOd.DrzuOhdgdprf, 10) + t.length; '; 121'
```

In Python, [parsing such an obfuscation is not be very difficult](https://github.com/oczkers/pycfl/blob/master/pycfl.py) (i.e. `!+[]` can be replaced with `1`), but I was also wondering if the `jsinterp` module can handle this without extra work. The only issue is if this obfuscation should change significantly.

Example use:

```python
from ..utils import CloudFlareSimpleJSChallengeMixin

class MyIE(InfoExtractor, CloudFlareSimpleJSChallengeMixin):
    CLOUDFLARE_HOSTNAME = 'website.name'

    def _real_extract(self, url):
        self._do_cloudflare_challenge(self.CLOUDFLARE_HOSTNAME, url, secure=True)

        # Rest of extractor code here
```

### DIFF
--- a/youtube_dl/extractor/nationalgeographic.py
+++ b/youtube_dl/extractor/nationalgeographic.py
@@ -68,7 +68,7 @@ class NationalGeographicVideoIE(InfoExtractor):
 
 class NationalGeographicIE(ThePlatformIE, AdobePassIE):
     IE_NAME = 'natgeo'
-    _VALID_URL = r'https?://channel\.nationalgeographic\.com/(?:wild/)?[^/]+/(?:videos|episodes)/(?P<id>[^/?]+)'
+    _VALID_URL = r'https?://channel\.nationalgeographic\.com/(?:(?:wild/)?[^/]+/)?(?:videos|episodes)/(?P<id>[^/?]+)'
 
     _TESTS = [
         {
@@ -101,6 +101,10 @@ class NationalGeographicIE(ThePlatformIE, AdobePassIE):
         },
         {
             'url': 'http://channel.nationalgeographic.com/the-story-of-god-with-morgan-freeman/episodes/the-power-of-miracles/',
+            'only_matching': True,
+        },
+        {
+            'url': 'http://channel.nationalgeographic.com/videos/treasures-rediscovered/',
             'only_matching': True,
         }
     ]


youtube_dl/utils.py
```diff
@@ -3843,3 +3844,107 @@ def random_birthday(year_field, month_field, day_field):
...
+
+class CouldNotCalculateAnswer(YoutubeDLError):
+    pass
+
+
+class CookieError(YoutubeDLError):
+    pass
+
+
+class CloudFlareSimpleJSChallengeMixin(object):
```
dstftw: CF detection and solving should happen centrally and automatically in `_webpage_read_content`. Otherwise it does not make any sense.

youtube_dl/utils.py
```diff
@@ -3843,3 +3844,107 @@ def random_birthday(year_field, month_field, day_field):
...
+    CF_CHALLENGE_PATH = '/cdn-cgi/l/chk_jschl'
+
+    def set_downloader(self, downloader):
+        self._downloader = downloader
+        self._add_503_handler()
+
+    def _add_503_handler(self):
+        if self._downloader:
+            class Handle503(object):
```
dstftw: No, catch exception.

dstftw:
>does not add an extra dependency, but it does require a version `node`

There is no difference. In both cases you depend on one to work.

Tatsh:
I am going to rework this to parse the JS as it is not very complicated so there won't be a need for Node to be installed.

Tatsh:
Hi I'd like to do the JS interpreting with the `jsinterp` module but the unary operators needed are missing: `!`, `+` (converts to int), `-` (same as Python). Also one unusual thing is how lists are 'added' in JS because JS does not do what Python does.

[Patch here](https://gist.githubusercontent.com/anonymous/49d3996340c8cace8666dc55f410c823/raw/bb35bdc46db7ac7428c5c47f255de4b3d31540bb/stdin) is an attempt to get the three unary operators working: 

This code can parse a string like `function x() { return !+[]+!![] +  +((!+[]+!![]+!![]+[])+(+!![])); }` without crashing, but it gets the wrong value (value should be `33`). It is mostly because I am having trouble making the unary `+` work when it is immediately after an infix `+`. I tried to alleviate this by changing the matching regex to ignore impossible operations between left side and right.

Any help appreciated.

```patch
diff --git a/youtube_dl/jsinterp.py b/youtube_dl/jsinterp.py
index 7bda59610..61a8a2bee 100644
--- a/youtube_dl/jsinterp.py
+++ b/youtube_dl/jsinterp.py
@@ -9,19 +9,37 @@ from .utils import (
     remove_quotes,
 )
 
+
+def _js_unary_plus(s):
+    if isinstance(s, int):
+        return s
+    if isinstance(s, list):
+        return 1 if len(s) else 0
+    raise ExtractorError('Unsupported expression with unary +: %r' % s)
+
+
+def _js_logical_not(s):
+    if isinstance(s, (bool, int)):
+        return not s
+    if isinstance(s, list):
+        return True if len(s) else False
+    raise ExtractorError('Unsupported expression with logical NOT: %r' % s)
+
+
 _OPERATORS = [
-    ('|', operator.or_),
-    ('^', operator.xor),
-    ('&', operator.and_),
-    ('>>', operator.rshift),
-    ('<<', operator.lshift),
-    ('-', operator.sub),
-    ('+', operator.add),
-    ('%', operator.mod),
-    ('/', operator.truediv),
-    ('*', operator.mul),
+    ('!', _js_logical_not, 1),
+    ('|', operator.or_, 2),
+    ('^', operator.xor, 2),
+    ('&', operator.and_, 2),
+    ('>>', operator.rshift, 2),
+    ('<<', operator.lshift, 2),
+    ('-', operator.sub, 2),
+    ('+', operator.add, 2),
+    ('%', operator.mod, 2),
+    ('/', operator.truediv, 2),
+    ('*', operator.mul, 2),
 ]
-_ASSIGN_OPERATORS = [(op + '=', opfunc) for op, opfunc in _OPERATORS]
+_ASSIGN_OPERATORS = [(op + '=', opfunc) for op, opfunc, nargs in _OPERATORS if nargs == 2]
 _ASSIGN_OPERATORS.append(('=', lambda cur, right: right))
 
 _NAME_RE = r'[a-zA-Z_$][a-zA-Z_$0-9]*'
@@ -181,10 +199,36 @@ class JSInterpreter(object):
 
             return obj[member](argvals)
 
-        for op, opfunc in _OPERATORS:
-            m = re.match(r'(?P<x>.+?)%s(?P<y>.+)' % re.escape(op), expr)
+        for op, opfunc, nargs in _OPERATORS:
+            if nargs == 1:
+                m = re.match(r'%s(?P<y>.+)' %  re.escape(op), expr)
                 if not m:
                     continue
+                y, abort = self.interpret_statement(
+                    m.group('y'), local_vars, allow_recursion - 1)
+                if abort:
+                    raise ExtractorError(
+                        'Premature right-side return of %s in %r' % (op, expr))
+                print(y, op)
+                import pdb; pdb.set_trace()
+                return opfunc(y)
+
+            m = re.match(r'(?P<x>(?:\[\]|[0-9]+|\{\}|%s))(?:\s+)?%s(?:\s+)?(?P<y>.+)' % (_NAME_RE, re.escape(op)), expr)
+            if not m:
+                m = re.match(r'%s(?P<y>.+)' %  re.escape(op), expr)
+                if not m:
+                    continue
+                y, abort = self.interpret_statement(
+                        m.group('y'), local_vars, allow_recursion - 1)
+                if abort:
+                    raise ExtractorError(
+                        'Premature right-side return of %s in %r' % (op, expr))
+                if op == '+':
+                    print(y, '+')
+                    return _js_unary_plus(y)
+                if op == '-':
+                    return -y
+                raise ExtractorError('Invalid unary operator %s in %r' % (op, expr))
             x, abort = self.interpret_statement(
                 m.group('x'), local_vars, allow_recursion - 1)
             if abort:
@@ -195,6 +239,11 @@ class JSInterpreter(object):
             if abort:
                 raise ExtractorError(
                     'Premature right-side return of %s in %r' % (op, expr))
+            if isinstance(x, list):
+                x = 1 if len(x) else 0
+            if isinstance(y, list):
+                y = 1 if len(y) else 0
+            #print(x, y, op)
             return opfunc(x, y)
 
         m = re.match(
```

Tatsh:
@dstftw Any chance you could see what's wrong here? I am almost able to pass every test I wrote but the string coercion rules in JS are not so easy to do (the return value has to be a concatenated string that happens to be a number not just a number). I think most of this is because of the way to handle unary vs non-unary in the `interpret_expression()` method around line 247. And the main error with the CF obfuscation is things like `((!` which the beginning of the parsing code will not understand because of the `!`.

These are the tests that fail:

```python
    def test_precedence(self):
        # ...
        jsi = JSInterpreter('function x() { return !+[] +1; }')  # Opposite order works
        self.assertEqual(jsi.call_function('x'), 2)

    def test_cloudflare_obfuscation(self):
        # ...
        jsi = JSInterpreter('function x() { return !+[] + !![] + !![]; }')
        self.assertEqual(jsi.call_function('x'), 3)  # should be an integer not string

        jsi = JSInterpreter('function x() { return !+[] + !![] + !![] + []; }')
        self.assertEqual(jsi.call_function('x'), '3')  # should be a string not integer

        jsi = JSInterpreter('function x() { return !+[] + !![] + ( +((!+[] + !![] + !![] +[]) + (+!![]))); }')
        self.assertEqual(jsi.call_function('x'), '33')  # Cannot parse '((!' (inner-most ()) properly
```

```
FAIL: test_cloudflare_obfuscation (test.test_jsinterp.TestJSInterpreter)
  File "/home/tatsh/dev/youtube-dl/test/test_jsinterp.py", line 221, in test_cloudflare_obfuscation
    self.assertEqual(jsi.call_function('x'), 3)
AssertionError: '3' != 3

FAIL: test_precedence (test.test_jsinterp.TestJSInterpreter)
  File "/home/tatsh/dev/youtube-dl/test/test_jsinterp.py", line 116, in test_precedence
    self.assertEqual(jsi.call_function('x'), 2)
AssertionError: '2' != 2
```

R030t1:
I hate to be the peanut gallery, but why does youtube-dl do its own JS parsing?

sulyi:
@Tatsh After a quick glance to the code you highlighted I'd like to point out that conditional expression is an assignment expression and not part of the number literal. I was using this [ECMA specification][1]

[1]: http://ecma-international.org/ecma-262/5.1/#sec-7.8.3

sulyi:
I've made a [syntax diagram][1] of the grammar in question. 
As I understand the parser's *NumericLiteral* is sightly different from *StringNumericLiteral* used by **ToNumber** abstract operation. More over the global object has properties (among others): **Infinity** and **NaN**, therefore these should be parsed as *Token* :: *IdentifierName*.
I couldn't find anything about binary number in the specs. About octal, I found that it's possible to extend *NumericLiteral*'s grammar with *OctalIntegerLiteral*, but this is not supported in strict mode.
I'm not yet sure how I'll implement these or if these are any use for you, but I hope this clears up some questions about how the interpreter should work.  

[1]: https://gist.github.com/sulyi/1b4b8a34b3a5eb37a20aa83c4d21ef3a

Issue #13271, opened 2017-06-03
Adding track # metadata to mp3 downloads by using the playlist_index #


playlistindex:
- [x ] I've **verified** and **I assure** that I'm running youtube-dl **2017.05.29**

your *issue*?
- [ ] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [x ] Question
- [ ] Other

---

As the title says, I'd like to download audio from youtube playlists while adding track numbers to each audio file based on its position in said youtube playlist.

I have tried tracking down the correct set of command modifiers and have managed to download perfectly healthy .mp3 files, but i have not found the proper set to add the track # metadata so i don't have to add it manually with audacity.

Thanks in advance.

playlistindex:
is this a duplicate of #7393? yes.  has #7393 been dead for a year and remained unresolved?  also yes

playlistindex:
@dstftw

br3wski3:
Opened pull request #14991 which I believe may support the OP request

Issue #7393, opened 2015-11-07
Feature Request - Add playlist number to metadata


DiamondKing099:
I've been using youtube-dl to download audio playlists. Instead of having to go through and manually add all the track numbers, it would be nice if youtube-dl could automatically add them to the file metadata.

(Not sure if this is the correct place to make a feature request, I'm not sure where else to put it.)

gaming-hacker:
make a bash script, grep/find the data you want from the manifest, pipe it to key  "--add-metadata"

jaimeMF:
Post an example url and explain what command you are using and what do you want it to do.

DiamondKing099:
Here's a URL of a YouTube playlist: https://www.youtube.com/watch?v=cgNXw8Qh67k&list=PLYwGqFKNm0YOgUAf9D-XpRAyDPripaXiB

What I would like is for the first track to have the number 1 set as it's track number with the --add-metadata option, 2 for the second track, 3 for the third, you get the idea.
So when I import them into iTunes or some other media player I won't have to manually add the track number to each one.

mohsend:
I came here for a similar feature request. I'm downloading entire video playlists and I would like to have the option to have the number of the video appended to the beginning of the file name. So I can have them sorted by the sequence in the playlist.
For example:
`youtube-dl https://www.youtube.com/playlist?list=PL8dPuuaLjXtMczXZUmjb3mZSU1Roxnrey`
Gets files like this:
`The Big Bang - Crash Course Big History #1-tq6be-CZJ3w.mp4`

What I have in mind:
`youtube-dl --append-playlist-number https://www.youtube.com/playlist?list=PL8dPuuaLjXtMczXZUmjb3mZSU1Roxnrey`
To save files like this:
`001 - The Big Bang - Crash Course Big History #1-tq6be-CZJ3w.mp4`

dstftw:
@mohsend your request is unrelated and already implemented: [output template, playlist_index](https://github.com/rg3/youtube-dl#output-template).

TRox1972:
As @dsftw says, you can use `playlist_index`. Should this be closed?

yan12125:
@mohsend's question is different from @M-W99's.

TRox1972:
Sorry, didn't notice that. But what @M-W99 is asking for is still possible, if a extractor return a value for `track_number` it will be added as metadata to the file if using `--add-metadata`. But I guess that isn't possible with YouTube, even with the `-x` option.

[...]

PR #14991, opened 2017-12-15
ffmpeg postprocessor use playlist_index for track metadata when track_number does not set

br3wski3:

---

- [ ] Bug fix
- [X] Improvement
- [ ] New extractor
- [ ] New feature

---

When using the --add-metadata option and with regard to the ffmpeg postprocessor use playlist_index for track metadata in lieu of track_number when the latter does not exist.
This is a useful behavior to me of which I think would be of use to others. As well, there are also issues filed in regard to this - #13271 and #7393


dstftw:
`playlist_index` is not related to track numbers in general thus it should not be written as track metadata by default.

johnvondoe:
Maybe an option then?

PR #15033, opened 2017-12-19
[la7] Add embedded urls, update tests

timendum:

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

I've update the la7 extractor with new test cases, removing old invalid cases and accepting new urls.

### DIFF
--- a/youtube_dl/extractor/la7.py
+++ b/youtube_dl/extractor/la7.py
@@ -10,9 +10,10 @@ from ..utils import (
 
 class LA7IE(InfoExtractor):
     IE_NAME = 'la7.it'
-    _VALID_URL = r'''(?x)(https?://)?(?:
-        (?:www\.)?la7\.it/([^/]+)/(?:rivedila7|video)/|
-        tg\.la7\.it/repliche-tgla7\?id=
+    _VALID_URL = r'''(?x)(?:https?://)(?:www\.|tg\.)?la7\.it/(?:
+        [^/]+/(?:rivedila7|video)/|
+        embedded/la7[^/]*(?:&|\?)content=|
+        repliche-tgla7\?id=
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/la7.py
```diff
@@ -10,17 +10,17 @@
...
+        embedded/la7[^/]*(?:&|\?)content=
     )(?P<id>.+)'''
 
     _TESTS = [{
         # 'src' is a plain URL
         'url': 'http://www.la7.it/crozza/video/inccool8-02-10-2015-163722',
         'md5': '8b613ffc0c4bf9b9e377169fc19c214c',
         'info_dict': {
-            'id': 'inccool8-02-10-2015-163722',
+            'id': '0_42j6wd36',
```
yan12125: It's better to keep the original ID for existing patterns, or --download-archive will be broken.

timendum: I did not change the id extraction, it changed in the Kaltura extractor.

yan12125:
Sorry for not checking the ID carefully. You are correct.

For the "invalid" test URL, I found a replacement: http://tg.la7.it/repliche-tgla7?id=231700. Could you add the pattern back and update the test?

timendum:
I can add the `_extract_urls` to check these urls, with embedded page, but the test you provided is no longer valid.

Should I push it anyway?

yan12125:
Well, there's another - http://tg.la7.it/repliche-tgla7?id=233847

Better to add a comment that such URLs are valid for only a short time.

timendum:
Now the extractor is capable of extracting video from these url, also added a new test case, with "match only".

Thanks for the input.

PR #15036, opened 2017-12-20
Allow entire file to match or first 10241 bytes

kmark:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This should prevent issues like that in #13449 where the FFmpeg downloader was not reliably trimming to the correct length. All it does is if the MD5 sum of the entire file does not match, attempt to match the MD5 sum of the first 10241 bytes. If neither match then fail. This will limit the second sum attempt to only initially failing tests.

Unlike #13609 this is fully backwards compatible with tests that already work. Tests that rely on FFmpeg's inconsistent behavior will continue to pass as well even though they probably should not. But fixing that is outside the scope of this PR.

### DIFF
--- a/test/test_download.py
+++ b/test/test_download.py
@@ -56,9 +56,9 @@ class YoutubeDL(youtube_dl.YoutubeDL):
         return super(YoutubeDL, self).process_info(info_dict)
 
 
-def _file_md5(fn):
+def _file_md5(fn, size=-1):
     with open(fn, 'rb') as f:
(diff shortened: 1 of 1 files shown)


kmark:
I should mention that there is a chance that this could result in false positives since it's completely naive as to which extractors actually need the 10241 byte limit.

Another idea is to only check the first 10241 bytes if it's on a whitelist of extractors that rely on FFmpeg's inconsistent truncation. Thoughts there?

Issue #14711, opened 2017-11-09
Unable to download from ABC iView (Australia)


simonoldfield:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like this: `[x]`)
- Use the *Preview* tab to see what your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2017.11.06*. If it's not, read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [ x] I've **verified** and **I assure** that I'm running youtube-dl **2017.11.06**

### Before submitting an *issue* make sure you have:
- [ x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [ x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x ] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add the `-v` flag to **your command line** you run youtube-dl with (`youtube-dl -v <your command line>`), copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):

```
 youtube-dl -c --verbose http://iview.abc.net.au/programs/full-out-2015/ZW0846A001S00
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'-c', u'--verbose', u'http://iview.abc.net.au/programs/full-out-2015/ZW0846A001S00']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2017.11.06
[debug] Python version 2.7.10 - Darwin-17.2.0-x86_64-i386-64bit
[...]

PR #15042, opened 2017-12-21
Fix link generation for ABC iview

ryleyangus:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

The ABC recently changed their streaming platform to block non-official clients. This pull request reflects the changes made to the iOS app to generate valid HLS links. As I do not have the key required for the HDS links, I also had to change the higher quality source URL to the HLS option. This fixes Issue #14711


youtube_dl/extractor/abc.py
```diff
@@ -126,12 +129,19 @@ def _real_extract(self, url):
         title = video_params.get('title') or video_params['seriesTitle']
         stream = next(s for s in video_params['playlist'] if s.get('type') == 'program')
 
-        format_urls = [
-            try_get(stream, lambda x: x['hds-unmetered'], compat_str)]
+        TIME = '%.6f' % time.time()
+        URLPARAM = '/auth/hls/sign?ts={}&hn={}&d=(null)'.format(TIME, video_id)
```
dstftw: No `{}`.

youtube_dl/extractor/abc.py
```diff
@@ -126,12 +129,19 @@ def _real_extract(self, url):
...
+        message = str(URLPARAM).encode('utf-8')
```
dstftw: str is pointless here.

youtube_dl/extractor/abc.py
```diff
@@ -126,12 +129,19 @@ def _real_extract(self, url):
...
+        secret = str('MainContainerViewController').encode('utf-8')
```
dstftw: Same.

youtube_dl/extractor/abc.py
```diff
@@ -126,12 +129,19 @@ def _real_extract(self, url):
...
+        SIG = hmac.new(secret, message, digestmod=hashlib.sha256).hexdigest()
+        abc_signature_url = "http://iview.abc.net.au" + URLPARAM + "&sig=" + SIG
```
dstftw: Single quotes.

youtube_dl/extractor/abc.py
```diff
@@ -126,12 +129,19 @@ def _real_extract(self, url):
...
+        urlsuffix = "?hdnea=" + self._download_webpage(abc_signature_url, None)
```
dstftw: `None` is not a video id.

ryleyangus:
Thanks for the feedback dstftw. I wasn't entirely too sure about the preferred formatted string method. Hopefully it is better now.

zoocey:
There is also https://github.com/rg3/youtube-dl/pull/14782 which fixes the same issue.

PR #15068, opened 2017-12-23
[ccma] Fix pattern to support video collection URLs

vlopezj:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Videos from CCMA which are part of a collection do not include the media type (audio or video) in the title. I fixed the URL pattern to support those videos, added a new test case, and added some heuristics to guess the media type based on the source ('video' for the TV channel TV3, and 'audio' for the radio channel Cat(alunya)Radio).

### DIFF
--- a/youtube_dl/extractor/ccma.py
+++ b/youtube_dl/extractor/ccma.py
@@ -13,7 +13,7 @@ from ..utils import (
 
 
 class CCMAIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?ccma\.cat/(?:[^/]+/)*?(?P<type>video|audio)/(?P<id>\d+)'
+    _VALID_URL = r'https?://(?:www\.)?ccma\.cat/(?P<channel>tv3|catradio|[^/*])(?:[^/]+/)*?(?P<type>audio|video|[^/]*)/(?P<id>\d+)/?$'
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/ccma.py
```diff
@@ -13,7 +13,7 @@
 
 
 class CCMAIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?ccma\.cat/(?:[^/]+/)*?(?P<type>video|audio)/(?P<id>\d+)'
+    _VALID_URL = r'https?://(?:www\.)?ccma\.cat/(?P<channel>tv3|catradio|[^/*])(?:[^/]+/)*?(?P<type>audio|video|[^/]*)/(?P<id>\d+)/?$'
```
dstftw: `(?P<channel>tv3|catradio|[^/*])` no, this imposes more strictness.
`[^/*]` incorrect.
`[^/]*` matching empty string does not make sense.
`/?$` incorrect.

vlopezj:
@dstftw I have added an alternation so that the new regular expression is strictly more general than the old one.
However, I do not know how to get the 'id' from the end of the URL without using `/?$`.

PR #15133, opened 2017-12-31
[Tumblr] Add login support

zackfern:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

I've added login support for Tumblr. Some time in the last year, Tumblr has made it impossible to download explicit content without logging in. When you try to view explicit content without being logged in, you're redirected to a splash screen (https://www.tumblr.com/safe-mode?url=...) which youtube-dl then passes to the generic downloader. This results in HTML pages being downloaded with weird names, such as "safe-mode-168487385934.unknown_video"

The addition of login support allows these URLs to be downloaded without issue. I've tested this locally with both the regular `--username` and `--password` flags and the ~/.netrc file and `-n` flag.

I'm not a Python developer by trade, so please excuse the code if it's a bit off. I believe the code passes flake8 - I ran `flake8 tumblr.py` locally and there wasn't any output.

Please let me know if there's anything else I can do to improve the code. Thanks for your work in maintaining youtube-dl!

### DIFF
--- a/youtube_dl/extractor/tumblr.py
+++ b/youtube_dl/extractor/tumblr.py
@@ -4,11 +4,19 @@ from __future__ import unicode_literals
 import re
 
 from .common import InfoExtractor
-from ..utils import int_or_none
+from ..utils import (
+    ExtractorError,
+    int_or_none,
+    sanitized_Request,
+    urlencode_postdata
+)
 
 
 class TumblrIE(InfoExtractor):
     _VALID_URL = r'https?://(?P<blog_name>[^/?#&]+)\.tumblr\.com/(?:post|video)/(?P<id>[0-9]+)(?:$|[/?#])'
+    _NETRC_MACHINE = 'tumblr'
+    _LOGIN_REQUIRED = False
+    _LOGIN_URL = 'https://www.tumblr.com/login'
     _TESTS = [{
         'url': 'http://tatianamaslanydaily.tumblr.com/post/54196191430/orphan-black-dvd-extra-behind-the-scenes',
         'md5': '479bb068e5b16462f5176a6828829767',
@@ -96,6 +104,31 @@ class TumblrIE(InfoExtractor):
         },
         'add_ie': ['Instagram'],
     }]
+
+    def _real_initialize(self):
+        self._login()
+
+    def _login(self):
+        (username, password) = self._get_login_info()
+        if username is None:
+            return
+        self.report_login()
+        webpage = self._download_webpage(self._LOGIN_URL, None, False)
+        form = self._hidden_inputs(webpage)
+        form.update({
+            'user[email]': username,
+            'user[password]': password
+        })
+        post_data = urlencode_postdata(form)
+        login_request = sanitized_Request(self._LOGIN_URL, post_data)
+        login_request.add_header('Content-Type', 'application/x-www-form-urlencoded')
+        login_request.add_header('Referer', self._LOGIN_URL)
+        login_response = self._download_webpage(login_request, None, False, 'Wrong login info')
+
+        # Check the login response from Tumblr for an error message and fail the extraction if we find one.
+        login_errors = self._search_regex(r'Tumblr\.RegistrationForm\.errors = \[(.*)\]', login_response, 'login errors', False, False)
+        if login_errors:
+            raise ExtractorError("Error logging in: %s" % login_errors)
 
     def _real_extract(self, url):
         m_url = re.match(self._VALID_URL, url)
--- a/youtube_dl/extractor/tumblr.py
+++ b/youtube_dl/extractor/tumblr.py
@@ -15,7 +15,6 @@ from ..utils import (
 class TumblrIE(InfoExtractor):
     _VALID_URL = r'https?://(?P<blog_name>[^/?#&]+)\.tumblr\.com/(?:post|video)/(?P<id>[0-9]+)(?:$|[/?#])'
     _NETRC_MACHINE = 'tumblr'
-    _LOGIN_REQUIRED = False
     _LOGIN_URL = 'https://www.tumblr.com/login'
     _TESTS = [{
         'url': 'http://tatianamaslanydaily.tumblr.com/post/54196191430/orphan-black-dvd-extra-behind-the-scenes',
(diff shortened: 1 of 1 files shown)


Hrxn:
Nice.

Good idea, relevant improvement, Thanks for your PR.

Although, for the sake of completeness, I want to mention that there is another method to use youtube-dl in this case, as a workaround:

Sign in to your Tumblr account, enable the content settings in your account preferences, and then export the cookies, either manually or by using on of the many browser extensions for that.
Then, running youtube-dl together with `--cookies COOKIEFILE` works as well..

youtube_dl/extractor/tumblr.py
```diff
@@ -97,6 +105,31 @@ class TumblrIE(InfoExtractor):
...
+        webpage = self._download_webpage(self._LOGIN_URL, None, False)
+        form = self._hidden_inputs(webpage)
+        form.update({
+            'user[email]': username,
+            'user[password]': password
+        })
+        post_data = urlencode_postdata(form)
+        login_request = sanitized_Request(self._LOGIN_URL, post_data)
+        login_request.add_header('Content-Type', 'application/x-www-form-urlencoded')
+        login_request.add_header('Referer', self._LOGIN_URL)
```
dstftw: Move data and query into `_download_webpage` call.

youtube_dl/extractor/tumblr.py
```diff
@@ -97,6 +105,31 @@ class TumblrIE(InfoExtractor):
...
+        login_response = self._download_webpage(login_request, None, False, 'Wrong login info')
+
+        # Check the login response from Tumblr for an error message and fail the extraction if we find one.
+        login_errors = self._search_regex(r'Tumblr\.RegistrationForm\.errors = \[(.*)\]', login_response, 'login errors', False, False)
```
dstftw: 1. Relax whitespace.
2. default and fatal are not used together.
3. Should not match empty string.

youtube_dl/extractor/tumblr.py
```diff
@@ -97,6 +105,31 @@ class TumblrIE(InfoExtractor):
...
+        if login_errors:
+            raise ExtractorError("Error logging in: %s" % login_errors)
```
dstftw: 1. Single quotes.
2. `expected`.

youtube_dl/extractor/tumblr.py
```diff
@@ -4,11 +4,19 @@
...
+    int_or_none,
+    sanitized_Request,
+    urlencode_postdata
+)
 
 
 class TumblrIE(InfoExtractor):
     _VALID_URL = r'https?://(?P<blog_name>[^/?#&]+)\.tumblr\.com/(?:post|video)/(?P<id>[0-9]+)(?:$|[/?#])'
+    _NETRC_MACHINE = 'tumblr'
+    _LOGIN_REQUIRED = False
```
dstftw: Unused.

youtube_dl/extractor/tumblr.py
zackfern: - Could you please elaborate on what you mean by "relax whitespace?" Are you saying I should modify the RegEx to ignore whitespace?
- Please correct me if I'm wrong, but I don't think default and fatal are being used together here. The default is set to false and the fatal flag (which is true by default) is set to false. If I shouldn't be setting both options what would you suggest I do instead?

dstftw: 1. Regex should match any variation of whitespace around `=`.
2. Once you provide default, fatal is not used thus it's pointless to provide fatal along with default.

Hrxn:
I think this might be related (Improvements to Tumblr extractor): https://github.com/rg3/youtube-dl/issues/14764

Because youtube-dl currently does not get FHD videos, even if available.

I realize that this issue probably looks a lot more verbose and complicated than it actually is, sorry for that 😄 

I searched around a bit, and there are, among others, some userscripts which seem to do the same thing.
For example this one here: https://greasyfork.org/en/scripts/32038-tumblr-hd-video-download-buttons

I think the really relevant part is basically this:

```JavaScript
// Typical Video URL Patterns:
// https://vt.media.tumblr.com/tumblr_ID_NUM.mp4
// https://vtt.tumblr.com/tumblr_ID_NUM.mp4
// https://vt.tumblr.com/tumblr_ID_NUM.mp4

var loc = location.toString();

// ----------------------------------------
// DIRECT MP4 URLS
// ----------------------------------------

function redirectToHD() {
	// Check that the URL is a ~.mp4
	if (!loc.endsWith('.mp4')) return;

	var lowQuality = /[$_]\d*.mp4$/;
	// Do not redirect if already HD
	if (!loc.match(lowQuality)) return;
	// Change to HD
	loc = loc.replace(lowQuality, '.mp4');

	// If the URL is HTTP, change it to HTTPS
	if (!loc.startsWith('https://')) {
		loc = loc.replace(/^http/, 'https');
	}

	// Redirect to the HD video
	location.replace(loc);
}
redirectToHD();

```

zackfern:
Hi @dstftw, I made the changes you requested. Could you please take a look when you have a moment? :) Thanks!

youtube_dl/extractor/tumblr.py
```diff
@@ -97,6 +104,31 @@ class TumblrIE(InfoExtractor):
...
+            return
+        self.report_login()
+        webpage = self._download_webpage(self._LOGIN_URL, None, False)
+        form = self._hidden_inputs(webpage)
+        form.update({
+            'user[email]': username,
+            'user[password]': password
+        })
+        login_response = self._download_webpage(
+            sanitized_Request(self._LOGIN_URL, urlencode_postdata(form), {
```
dstftw: Nothing changed.

youtube_dl/extractor/tumblr.py
```diff
@@ -97,6 +104,31 @@ class TumblrIE(InfoExtractor):
...
+                'Content-Type': 'application/x-www-form-urlencoded',
+                'Referer': self._LOGIN_URL
+            }), None, False, 'Wrong login info')
```
dstftw: `False` is not a valid note.

youtube_dl/extractor/tumblr.py
```diff
@@ -97,6 +104,31 @@ class TumblrIE(InfoExtractor):
...
 
+    def _real_initialize(self):
+        self._login()
+
+    def _login(self):
+        (username, password) = self._get_login_info()
+        if username is None:
+            return
+        self.report_login()
+        webpage = self._download_webpage(self._LOGIN_URL, None, False)
```
dstftw: `False` is not a valid note.

youtube_dl/extractor/tumblr.py
```diff
@@ -97,6 +104,31 @@ class TumblrIE(InfoExtractor):
...
+            'user[password]': password
+        })
+        login_response = self._download_webpage(
+            sanitized_Request(self._LOGIN_URL, urlencode_postdata(form), {
+                'Content-Type': 'application/x-www-form-urlencoded',
+                'Referer': self._LOGIN_URL
+            }), None, False, 'Wrong login info')
+
+        # Check the login response from Tumblr for an error message and fail the extraction if we find one.
+        login_errors = self._search_regex(r'Tumblr\.RegistrationForm\.errors\s*=\s*\[[\"|\'](.+)[\"|\']\]', login_response, 'login errors', False)
```
dstftw: `[\"|\']` what is this.

Issue #3384, opened 2014-07-27
Unsupported URL


lightonflux:
Hi, a friend found a URL that isn't supported by youtube-dl and doesn't work with the generic extractor:

http://www.gamepro.de/playstation/spiele/ps4/alien-isolation/videos/50463,76030.html

PR #15179, opened 2018-01-05
extend gamestar extractor to cover gamepro.de

lorem-ipsum:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

[gamepro.de](gamepro.de) is the console focused variant of gamestar.de. Both use the same page layout and the existing gamestar extractor works fine for gamepro. Issue #3384 requested support for it a while ago.

### DIFF
--- a/youtube_dl/extractor/gamestar.py
+++ b/youtube_dl/extractor/gamestar.py
@@ -9,21 +9,27 @@ from ..utils import (
 
 
 class GameStarIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?gamestar\.de/videos/.*,(?P<id>[0-9]+)\.html'
-    _TEST = {
-        'url': 'http://www.gamestar.de/videos/trailer,3/hobbit-3-die-schlacht-der-fuenf-heere,76110.html',
-        'md5': '96974ecbb7fd8d0d20fca5a00810cea7',
-        'info_dict': {
-            'id': '76110',
-            'ext': 'mp4',
-            'title': 'Hobbit 3: Die Schlacht der Fünf Heere - Teaser-Trailer zum dritten Teil',
-            'description': 'Der Teaser-Trailer zu Hobbit 3: Die Schlacht der Fünf Heere zeigt einige Szenen aus dem dritten Teil der Saga und kündigt den...',
-            'thumbnail': r're:^https?://.*\.jpg$',
-            'timestamp': 1406542020,
-            'upload_date': '20140728',
-            'duration': 17
-        }
-    }
+    _VALID_URL = r'https?://(?:www\.)?game(?:pro|star)\.de/videos/.*,(?P<id>[0-9]+)\.html'
+    _TESTS = [
+        {
+            'url': 'http://www.gamestar.de/videos/trailer,3/hobbit-3-die-schlacht-der-fuenf-heere,76110.html',
+            'md5': 'ee782f1f8050448c95c5cacd63bc851c',
+            'info_dict': {
+                'id': '76110',
+                'ext': 'mp4',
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/gamestar.py
```diff
@@ -9,7 +9,7 @@
 
 
 class GameStarIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?gamestar\.de/videos/.*,(?P<id>[0-9]+)\.html'
+    _VALID_URL = r'https?://(?:www\.)?game(pro|star)\.de/videos/.*,(?P<id>[0-9]+)\.html'
```
dstftw: Don't capture groups you don't use. Add a test.

lorem-ipsum: Did both. Also fixed view_count and comment_count.

youtube_dl/extractor/gamestar.py
```diff
@@ -38,11 +54,12 @@ def _real_extract(self, url):
             webpage, 'JSON-LD', group='json_ld'), video_id)
         info_dict = self._json_ld(json_ld, video_id)
         info_dict['title'] = remove_end(info_dict['title'], ' - GameStar')
+        info_dict['title'] = remove_end(info_dict['title'], ' - GamePro')
 
-        view_count = json_ld.get('interactionCount')
+        view_count = int(json_ld.get('interactionCount'))
```
dstftw: int_or_none.

youtube_dl/extractor/gamestar.py
```diff
@@ -9,21 +9,37 @@
...
+            'info_dict': {
+                'id': '95316',
+                'ext': 'mp4',
+                'title': 'Top 10: Indie-Spiele für Nintendo Switch - Video: Tolle Nindies-Games zum Download',
+                'description': 'Wer behauptet, im letzten Jahr seien nicht ausreichend Spiele für die Switch\xa0herausgekommen, der hat offensichtlich nicht richtig...',
+                'thumbnail': r're:^https?://.*\.jpg$',
+                'timestamp': 1515173400,
+                'upload_date': '20180105',
+                'duration': 615,
+            }
```
dstftw: Make only_matching.

lorem-ipsum: I've removed everything except the URL and added the only_matching tag. Is that as desired?

youtube_dl/extractor/gamestar.py
lorem-ipsum: done

youtube_dl/extractor/gamestar.py
```diff
@@ -38,11 +44,12 @@ def _real_extract(self, url):
             webpage, 'JSON-LD', group='json_ld'), video_id)
         info_dict = self._json_ld(json_ld, video_id)
         info_dict['title'] = remove_end(info_dict['title'], ' - GameStar')
+        info_dict['title'] = remove_end(info_dict['title'], ' - GamePro')
```
dstftw: 46-47 code duplication.

lorem-ipsum: Any preferences on how to solve this? I can squeeze it in one line with two calls to `remove_end`, but in my eyes this make things only worse.

dstftw: The point is eliminating code duplication, not squeezing into one line.

lorem-ipsum: I don't really get what you want. I could remove the code duplication by defining a micro-function calling remove_end, or I could write my own remove_end which removing based on a regex or something like this. In my eyes neither change provides any benefit.
Please give me a direction on how I should reformat the code, I think we have different opinions on coding style.

PR #15233, opened 2018-01-12
Mytaratata

cdevienne:
### Did:

### License

### Purpose
- [x] New extractor

---

## Description

Add a new extractor for the "mytaratata.com" website.

### DIFF
--- a/youtube_dl/extractor/mytaratata.py
+++ b/youtube_dl/extractor/mytaratata.py
@@ -35,7 +35,7 @@ class MyTaratataIE(InfoExtractor):
         formats = []
 
         video_source_re = re.compile(
-            r'<div [^>]*class="jwplayer" [^\>]*data-source="(?P<url>https?://.*/Taratata[^"]+\.mp4)"'
+            r'<div [^>]*class="jwplayer" [^\>]*data-source="(?P<url>[^"]+)"'
         )
 
         # The first videos are the live videos, coming in 2 formats. The next videos are
--- a/youtube_dl/extractor/mytaratata.py
+++ b/youtube_dl/extractor/mytaratata.py
@@ -1,0 +1,68 @@
+# coding: utf-8
+
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+
+
+class MyTaratataIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?mytaratata\.com/taratata/(?P<id>[a-zA-Z0-9_\-/]+)'
+    _TEST = {
+        'url': 'http://mytaratata.com/taratata/519/shaka-ponk-camille-et-julie-berthollet-smells-like-teen-spirit-nirvana',
+        'md5': 'c2876e18716b350c9de69cfda2662919',
+        'info_dict': {
+            'id': '519/shaka-ponk-camille-et-julie-berthollet-smells-like-teen-spirit-nirvana',
+            'ext': 'mp4',
+            'title': 'Taratata - Shaka Ponk / Camille et Julie Berthollet "Smells Like Teen Spirit" (Nirvana)',
+            'uploader': 'Taratata',
+            'description': 'Shaka Ponk / Camille et Julie Berthollet "Smells Like Teen Spirit" (Nirvana)',
+            # 'thumbnail': r're:^https?://.*\.jpg$',
+            # TODO more properties, either as:
+            # * A value
+            # * MD5 checksum; start the string with md5:
+            # * A regular expression; start the string with re:
+            # * Any Python type (for example int or float)
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        # TODO more code goes here, for example ...
+        title = self._html_search_regex(r'<title>(.+?)</title>', webpage, 'title')
+
+        formats = []
+
+        video_source_re = re.compile(
+            r'data-source="(?P<url>http://videos.air-productions.cdn.sfr.net'
+            r'/mytaratata/Taratata[^"]+\.mp4)"'
+        )
+
+        last_vid = None
+        for url in video_source_re.findall(webpage):
+            info_m = re.match(r'.*(?P<vid>[0-9]+)-[a-f0-9]+-(?P<w>[0-9]+)x(?P<h>[0-9]+)\.mp4', url)
+            if info_m is None:
+                continue
+            vid = info_m.group('vid')
+            w = info_m.group('w')
+            h = info_m.group('h')
+            if last_vid is None:
+                last_vid = vid
+            if vid != last_vid:
+                break
+
+            formats.append({'url': url, 'width': int(w), 'height': int(h)})
+
+        formats = list(sorted(formats, key=lambda f: f['width']))
+
+        return {
+            'id': video_id,
+            'title': title,
+            'description': self._og_search_description(webpage),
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/mytaratata.py
```diff
@@ -0,0 +1,72 @@
...
+
+class MyTaratataIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?mytaratata\.com/taratata/(?P<id>[a-zA-Z0-9_\-/]+)'
+    _TEST = {
+        'url': 'http://mytaratata.com/taratata/519/shaka-ponk-camille-et-julie-berthollet-smells-like-teen-spirit-nirvana',
+        'md5': '99657330eb7dec6d63a329d7f26ec93e',
+        'info_dict': {
+            'id': '7174',
+            'ext': 'mp4',
+            'title': u'TARATATA N°519 - Shaka Ponk / Camille et Julie Berthollet "Smells Like Teen Spirit" (Nirvana)',
```
dstftw: No `u`s.

youtube_dl/extractor/mytaratata.py
```diff
@@ -0,0 +1,72 @@
...
+
+        video_source_re = re.compile(
+            r'data-source="(?P<url>http://videos.air-productions.cdn.sfr.net'
+            r'/mytaratata/Taratata[^"]+\.mp4)"'
+        )
+
+        # The first videos are the live videos, coming in 2 formats. The next videos are
+        # bonuses, multi-cams... that we won't download.
+        last_vid = None
+        for url in video_source_re.findall(webpage):
```
dstftw: Do not shadow outer names.

youtube_dl/extractor/mytaratata.py
```diff
@@ -0,0 +1,72 @@
...
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._og_search_title(webpage)
+        description = self._og_search_description(webpage)
+
+        formats = []
+
+        video_source_re = re.compile(
+            r'data-source="(?P<url>http://videos.air-productions.cdn.sfr.net'
```
dstftw: More relaxed regex. Don't match exact URL.

youtube_dl/extractor/mytaratata.py
```diff
@@ -0,0 +1,72 @@
...
+                # We found a new video, not another format of the same. Stops here.
+                break
+
+            formats.append({
+                'url': url,
+                'width': int(w),
+                'height': int(h),
+            })
+
+        formats = list(sorted(formats, key=lambda f: f['width']))
```
dstftw: `_sort_formats`.

youtube_dl/extractor/mytaratata.py
```diff
@@ -0,0 +1,72 @@
...
+
+        return {
+            'id': last_vid,
+            'title': '%s - %s' % (title, description),
```
dstftw: Remove description.

youtube_dl/extractor/mytaratata.py
```diff
@@ -0,0 +1,72 @@
...
+    _TEST = {
+        'url': 'http://mytaratata.com/taratata/519/shaka-ponk-camille-et-julie-berthollet-smells-like-teen-spirit-nirvana',
+        'md5': '99657330eb7dec6d63a329d7f26ec93e',
+        'info_dict': {
+            'id': '7174',
+            'ext': 'mp4',
+            'title': u'TARATATA N°519 - Shaka Ponk / Camille et Julie Berthollet "Smells Like Teen Spirit" (Nirvana)',
+            'uploader': 'Taratata',
+            'description': 'Shaka Ponk / Camille et Julie Berthollet "Smells Like Teen Spirit" (Nirvana)',
+            'thumbnail': 'http://static.mytaratata.com/content/image/5a2562a1a5ee5.jpeg',
```
dstftw: `re:`.

youtube_dl/extractor/mytaratata.py
cdevienne: The description in the page actually contains the video title, which is relative to a program number. And the title contains only the program number, which is not enough to distinguish the various video of the program.

youtube_dl/extractor/mytaratata.py
cdevienne: Do you mean I must use a regexp? May I ask why?

youtube_dl/extractor/mytaratata.py
cdevienne: will do.
Just curious why this rule? I though a more precise url would be safer, ie it will break as soon as something changes, so we are sure not to miss important changes in the website structure.

dstftw: No, completely the opposite. More precise URL imposes more restrictions thus easier to break. If they change storage host or path yours will break while matching by only attribute won't.

youtube_dl/extractor/mytaratata.py
dstftw: Cause it will break on any minor change of storage location.

cdevienne:
I think I made all the required changes.
Thanks for the detailed review.

youtube_dl/extractor/mytaratata.py
```diff
@@ -0,0 +1,76 @@
...
+        # The title contains only the program name and episode number.
+        # Each episode containts many videos
+        title = self._og_search_title(webpage)
+        # The description is the title of the video within the episode.
+        description = self._og_search_description(webpage)
+
+        formats = []
+
+        video_source_re = re.compile(
+            r'<div [^>]*class="jwplayer" [^\>]*data-source="(?P<url>https?://.*/Taratata[^"]+\.mp4)"'
```
dstftw: Just like I expected storage location has changed meanwhile so this code is completely broken. As I've already pointed out: match **only** `data-source` attribute.

youtube_dl/extractor/mytaratata.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+
+        # The first videos are the live videos, coming in 2 formats. The next videos are
+        # bonuses, multi-cams... that we won't download.
+        last_vid = None
+        for video_url in video_source_re.findall(webpage):
+            info_m = re.match(
+                r'.*/(?P<vid>[0-9]+)-[a-f0-9]+-(?P<w>[0-9]+)x(?P<h>[0-9]+)\.mp4',
```
dstftw: This may change as well. Add a fallback that just processes all videos without differentiation.

youtube_dl/extractor/mytaratata.py
```diff
@@ -0,0 +1,76 @@
...
+                'height': int(h),
+            })
+
+        self._sort_formats(formats)
+
+        return {
+            'id': last_vid,
+            'title': '%s - %s' % (title, description),
+            'description': description,
+            'uploader': "Taratata",
```
dstftw: What's the point of this?

cdevienne: uploader? I thought it was needed.
Will remove.

youtube_dl/extractor/mytaratata.py
cdevienne: I can do that, but the resulting behavior will probably not be of any use.
I think I prefer to fix the extractor when it changes.
But if I must, for the extractor to be merged, tell me.

youtube_dl/extractor/mytaratata.py
cdevienne: will do

PR #15276, opened 2018-01-16
[Picarto] Add new extractor for recorded Picarto VODs

caiwan:
Added a new extractor to download recorded streams from Picarto.

### DIFF
--- a/youtube_dl/extractor/picarto.py
+++ b/youtube_dl/extractor/picarto.py
@@ -1,8 +1,9 @@
 # coding: utf-8
+import re
+
 from __future__ import unicode_literals
 
 from .common import InfoExtractor
-
 
 class PicartoVodIE(InfoExtractor):
     _VALID_URL = r'https?://(?:www\.)?picarto\.tv/videopopout/(?P<id>[a-zA-Z_\-.0-9]+).flv'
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/picarto.py
```diff
@@ -0,0 +1,38 @@
...
+        vod_script = self._html_search_regex(vod_regex, webpage, "vod_script")
+        vod_url = self._search_regex('vod: \\"([^\\"]*)\\"', vod_script, "vod_url")
+        vod_thumb = self._search_regex('vodThumb: \\"([^\\"]*)\\"', vod_script, "vod_thumb")
+
+        title = video_id
+        
+        return {
+            'id': video_id,
+            'title': title,
+            'description': "",
```
dstftw: No.

youtube_dl/extractor/picarto.py
```diff
@@ -0,0 +1,38 @@
...
+            'uploader': "",
```
dstftw: No.

youtube_dl/extractor/picarto.py
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        vod_regex = "<script[^<]*>[^<]*riot.mount\(([^<]+)\)[^<]*<\/"
+        vod_script = self._html_search_regex(vod_regex, webpage, "vod_script")
+        vod_url = self._search_regex('vod: \\"([^\\"]*)\\"', vod_script, "vod_url")
+        vod_thumb = self._search_regex('vodThumb: \\"([^\\"]*)\\"', vod_script, "vod_thumb")
```
dstftw: `_parse_json`.

youtube_dl/extractor/picarto.py
```diff
@@ -0,0 +1,38 @@
...
+            'title': 'Carrot_2018.01.11.07.55.12',
+            'thumbnail': r're:^https?://.*\.jpg$'
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        vod_regex = "<script[^<]*>[^<]*riot.mount\(([^<]+)\)[^<]*<\/"
```
dstftw: r prefix, single quotes.

Issue #14789, opened 2017-11-18
[utils] js_to_json can't handle scientific notations (was: [twitch:clips] JSONDecodeError for certain clips)


s104346:
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2017.11.15**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add the `-v` flag to **your command line** you run youtube-dl with (`youtube-dl -v <your command line>`), copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):

```
$ python -m youtube_dl --verbose https://clips.twitch.tv/CarelessZealousKangarooNerfBlueBlaster
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['--verbose', 'https://clips.twitch.tv/CarelessZealousKangarooNerfBlueBlaster']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2017.11.15
[debug] Git HEAD: f610dbb05
[debug] Python version 3.6.3 - Linux-4.12.13-1-ARCH-x86_64-with-arch
[debug] exe versions: ffmpeg 3.4, ffprobe 3.4, rtmpdump 2.4
[debug] Proxy map: {}
[twitch:clips] CarelessZealousKangarooNerfBlueBlaster: Downloading webpage
[...]

PR #15372, opened 2018-01-20
[util] improve scientific notation handling in js_to_json()

mikf:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Fix for #14789 

js_to_json() wrongly recognizes the `e`&`E` of scientific notation literals as start of a new JS identifier and encloses it in quotation marks (`4.2e1 -> 4.2"e1"`).
Using a lookbehind assertion to disallow digits directly in front of identifiers fixes this behavior.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -814,6 +814,9 @@ class TestUtil(unittest.TestCase):
         inp = '''{"duration": "00:01:07"}'''
         self.assertEqual(js_to_json(inp), '''{"duration": "00:01:07"}''')
 
+        inp = '''{segments: [{"offset":-3.885780586188048e-16,"duration":39.75000000000001}]}'''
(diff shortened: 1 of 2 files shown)


dstftw:
This allows not only `e` and `E` but other characters as well. `4.2a1` should result in `4.2"a1"` not `4.2a1`.

mikf:
`4.2a1` and  `4.2"a1"` would be invalid anyway and the current regex is much more readable than `(?:(?<![0-9])[eE]|[a-df-zA-DF-Z_])[.a-zA-Z_0-9]*`, which is why I didn't go with that. But if you want it changed, sure.

dstftw:
The point is that it changes the previous behavior. Also the latter regex is much more explicit about its intentions.

PR #15421, opened 2018-01-25
[openload] New --phantomjs-location option

Rudloff:

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

This patch adds a new --phantomjs-location option that allows users to use a phantomjs binary that is not in PATH.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -2246,7 +2246,7 @@ class YoutubeDL(object):
 
         exe_versions = FFmpegPostProcessor.get_versions(self)
         exe_versions['rtmpdump'] = rtmpdump_version()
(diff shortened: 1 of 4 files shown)


dstftw:
Why not just setup `PATH` appropriately? I don't think it deserves a separate command since it's only used in one single place.

Rudloff:
I'm running youtube-dl inside a PHP app and it feels cleaner to just set this flag instead of having to mess with the subprocess's PATH. (Also using PATH means that the binary must be named `phantomjs`.)
But I totally understand if you don't want to add an option just for this.

dstftw:
Adding an option to some 3rdparty software just because you don't want to bother with standard communication mechanisms in your software is definitely not cleaner. I'm rejecting this for now. Though this still may be accepted later if phantomjs is more widely adopted in youtube-dl.

PR #15455, opened 2018-01-30
[discovery] Extract the video access token when possible

cookieguru:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Discovery reveals the stream URL through a REST call.  The REST API uses OAuth.  For videos that do not require a cable subscription, generating a new token on each download works fine.  When a user is logged in, the necessary token is already specified in the cookies sent with the video page.  This PR checks the cookies sent with the page for the cookie containing the token that has authorization to download authenticated videos.  The cookie could be stored in `eosAn`, `eosAf` or not sent at all.  The cookie is rarely sent on the first request; a second request usually has to be made.  I have been unable to find a workaround for cases where the cookie is not sent on the second request.  The existing code of requesting a new token is preserved in case the `eosA*` cookie could not be found.  This is an implementation of a JS userscript outlined in https://github.com/rg3/youtube-dl/issues/14954#issuecomment-360925988

This is my first foray into Python, so code suggestions are encouraged.

### DIFF
--- a/youtube_dl/extractor/discovery.py
+++ b/youtube_dl/extractor/discovery.py
@@ -4,6 +4,7 @@ import random
 import re
 import string
 
+from ..compat import compat_urllib_parse_unquote
 from .discoverygo import DiscoveryGoBaseIE
 from ..utils import (
     ExtractorError,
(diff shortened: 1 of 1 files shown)


cookieguru:
Looks like @remitamine took care of this in a slightly different way with 3cc0d0b8299308958bfe8b4c42c739505df27f50

Issue #15501, opened 2018-02-05
la7: Unable to extract player data


Vrihub:
### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2018.02.04*. If it's not, read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2018.02.04**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
- [x] Checked that provided video/audio/playlist URLs (if any) are alive and playable in a browser

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

Regex error while downloading from la7.it

```
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', '-F', 'http://www.la7.it/omnibus/rivedila7/omnibus-04-02-2018-233121']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2018.02.04
[debug] Python version 3.5.3 (CPython) - Linux-4.9.0-5-amd64-x86_64-with-debian-9.3
[debug] exe versions: ffmpeg 3.2.10-1, ffprobe 3.2.10-1, rtmpdump 2.4
[debug] Proxy map: {}
[la7.it] omnibus-04-02-2018-233121: Downloading webpage
ERROR: Unable to extract player data; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/youtube_dl/YoutubeDL.py", line 785, in extract_info
    ie_result = ie.extract(url)
[...]

PR #15502, opened 2018-02-05
Fix "Unable to extract player data" regex error

Vrihub:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Fix for #15501

### DIFF
--- a/youtube_dl/extractor/la7.py
+++ b/youtube_dl/extractor/la7.py
@@ -49,7 +49,7 @@ class LA7IE(InfoExtractor):
         webpage = self._download_webpage(url, video_id)
 
         player_data = self._parse_json(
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/la7.py
```diff
@@ -49,7 +49,7 @@ def _real_extract(self, url):
         webpage = self._download_webpage(url, video_id)
 
         player_data = self._parse_json(
-            self._search_regex(r'videoLa7\(({[^;]+})\);', webpage, 'player data'),
+            self._search_regex(r'videoParams = ({[^;]+});', webpage, 'player data'),
```
dstftw: 1. Relax regex.
2. Keep the old one as well.

Issue #13689, opened 2017-07-20
FranceTV live stream error


francetvfan:
youtube-dl -F "https://www.france.tv/france-3/direct.html"
[FranceTV] direct: Downloading webpage
[FranceTV] SIM_France3: Downloading video JSON
[FranceTV] SIM_France3: Downloading f4m manifest token
[FranceTV] SIM_France3: Downloading f4m manifest
WARNING: Unable to download f4m manifest: HTTP Error 404: Not Found
[FranceTV] SIM_France3: Downloading f4m manifest token
[FranceTV] SIM_France3: Downloading f4m manifest
WARNING: Unable to download f4m manifest: HTTP Error 403: Forbidden
[FranceTV] SIM_France3: Downloading m3u8 information
WARNING: Failed to download m3u8 information: HTTP Error 403: Forbidden
[FranceTV] SIM_France3: Downloading m3u8 information
WARNING: Failed to download m3u8 information: HTTP Error 403: Forbidden
[FranceTV] SIM_France3: Downloading m3u8 information
WARNING: Failed to download m3u8 information: HTTP Error 403: Forbidden
[FranceTV] SIM_France3: Downloading m3u8 information
WARNING: Failed to download m3u8 information: HTTP Error 403: Forbidden
ERROR: No video formats found; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.

PR #15536, opened 2018-02-09
[francetv] Also fetch the hmac token for m3u8 URLs

mmuman:
they seem to be required now, else we get 403 errors.

Seems to be a partial(?) fix for #13689.

- [x] Bug fix

### DIFF
--- a/youtube_dl/extractor/francetv.py
+++ b/youtube_dl/extractor/francetv.py
@@ -64,9 +64,13 @@ class FranceTVBaseInfoExtractor(InfoExtractor):
                         f4m_url + '&hdcore=3.7.0&plugin=aasp-3.7.0.39.44',
                         video_id, f4m_id=format_id, fatal=False))
             elif ext == 'm3u8':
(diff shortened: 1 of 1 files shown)


dstftw:
Provide concrete example URL this PR is supposed to fix.

mmuman:
https://www.france.tv/france-5/masha-et-michka/153915-la-fievre-de-la-danse.html
https://www.france.tv/france-2/envoye-special/401463-emission-du-jeudi-8-fevrier-2018.html

Oh, now they seem to be working without the fix.
Maybe it's a temporary measure…

dstftw:
There are requests to hdfauth.francetv.fr for me. Closing since current code works.

mmuman:
Possibly only some of their servers check for it, and it will come and bite us again…

mmuman:
Just tested 8faa338 and it works for me.

Still getting some 403 with the URL from #13689 but we do get enough formats to download something. Maybe some of the URLs we get from it are deprecated.

PR #15584, opened 2018-02-13
Added --metadata option, allows the user to specify key/value pairs for encoding metadata.

vayalil:

---

- [ x] At least skimmed through [adding new extractor tutorial](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/rg3/youtube-dl#youtube-dl-coding-conventions) sections

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ x] New feature

---

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1033,7 +1033,7 @@ class YoutubeDL(object):
             '!=': operator.ne,
         }
         operator_rex = re.compile(r'''(?x)\s*
-            (?P<key>width|height|tbr|abr|vbr|asr|filesize|fps)
+            (?P<key>width|height|tbr|abr|vbr|asr|filesize|filesize_approx|fps)
             \s*(?P<op>%s)(?P<none_inclusive>\s*\?)?\s*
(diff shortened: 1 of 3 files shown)


vayalil:
Added --metadata option, which allows the user to specify key/value pairs for encoding metadata, separated by equal sign. Example: --metadata "album = Movie Title artist = Various artists year = "2010"'. Also tries to capture artist information from the file description, if otherwise not available'

vayalil:
Verified with flake8, no errors.

dstftw:
1. Have you even tried to check your changes? It does not work since it's not wired into the code.
2. It must not do anything apart from taking user provided metadata. No extraction from description, etc.
3. Core tests are broken.
4. Duplicate of #6064.

vayalil:
1. Sorry, it will work if you used --add-metadata option as well. Since this --metadata is just providing additional metadatas, so I thought it would be better in that way. 
2. Same as above, it will extract 'artist's name from the file description. More option should be added if required
4. But it is not still added to the main.

PR #15629, opened 2018-02-17
--metadata option, allows the user to specify key/value pairs for encoding metadata

vayalil:

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

1. Have you even tried to check your changes? It does not work since it's not wired into the code.
2. It must not do anything apart from taking user provided metadata. No extraction from description, etc.
3. Core tests are broken.
4. Duplicate of #6064.

1 & 2. Bug fixed
3. Core test was failed with test_unicode_literals -> Fixed
4. The behaviour of #6064 is different. The #6064 is nearly the same as "--metadata-from-title" option, whilst the proposed modification allows user to override the parsed values by the "--add-metadata" or "--metadata-from-title", with user specified values. The proposed option allows user to specify all key/value pairs in a single --metadata call, i.e. "key1 = value1 key2 = value2", much easier than #6064.
--

### DIFF
--- a/youtube_dl/postprocessor/metadatafromuser.py
+++ b/youtube_dl/postprocessor/metadatafromuser.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+"""
+Created on Fri Feb  9 16:23:16 2018
+
+@author: Niras C. Vayalil
+"""
+
+from __future__ import unicode_literals
+from .common import PostProcessor
+import re
+
+
+class MetadataFromUserPP(PostProcessor):
+    def __init__(self, downloader, metadata):
+        super(MetadataFromUserPP, self).__init__(downloader)
+        self._metadata = metadata
+        
+
+    def run(self, info):
+        lastpos = 0
+        attribute = None
+        # Search for key words
+        for match in re.finditer(r'(\w+) *=', self._metadata):
+            if attribute is not None:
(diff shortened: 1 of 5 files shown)


dstftw:
2 not fixed. Custom meta fields must be added with multiple `--add-custom-metadata` options similar to `--add-header`. Metafields must support output template and multiline.

youtube_dl/postprocessor/metadatafromuser.py
```diff
@@ -0,0 +1,47 @@
+from __future__ import unicode_literals
+
+# Created on Fri Feb  9 16:23:16 2018
```
dstftw: Remove.

youtube_dl/postprocessor/metadatafromuser.py
```diff
@@ -0,0 +1,47 @@
...
+            # No attributes is given user, default assume the text as album name
+            attribute = 'album'
+        # Add the last key
+        value = self._metadata[lastpos:].strip()
+        self._downloader.to_screen(
+            '[fromuser] %s: %s' % (attribute, value if value else 'NA'))
+        info[attribute] = value
+
+        # If nothing specified for artist, extract from comments/description
+        if (('artist' not in info) and ('description' in info)):
```
dstftw: I've already pointed out: I won't accept this.

PR #15822, opened 2018-03-11
[TastyTrade] Fix old test, handle alt urls

thepelkus-too:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

TastyTrade added a new URL format that's common to their daily email newsletter. (It may also be accessible from the site, but I'm not sure.) I've changed the regex to recognize the new URL format. The change goes deeper, however, because these other URLs use a different content template that doesn't have the JSON-LD content that the scraper was originally using. The scraper will now fallback to looking for an escaped JSON string that's passed to some inline Javascript and use that to gather metadata.

Additionally, the old tests for this were failing, so I updated the data for the original test so that it now passes.

### DIFF
--- a/youtube_dl/extractor/tastytrade.py
+++ b/youtube_dl/extractor/tastytrade.py
@@ -2,12 +2,7 @@ from __future__ import unicode_literals
 
 from .common import InfoExtractor
 from .ooyala import OoyalaIE
-from youtube_dl.utils import (
-    ExtractorError,
-)
 
-import json
-import re
 import sys
 
 
@@ -56,22 +51,28 @@ class TastyTradeIE(InfoExtractor):
 
         info = {'id': None, 'title': None, 'description': None}
 
-        try:
-            info = self._search_json_ld(webpage, display_id, fatal=False)
-        except ExtractorError as ex:
-            json_string_match = re.search(
-                r'var episodeData = \$.parseJSON\("(?P<episode_json>.*)"\)', webpage, 0)
+        json_ld_info = self._search_json_ld(
+            webpage, display_id, default=None, fatal=False)
 
-            if (json_string_match):
-                escaped_json_string = json_string_match.group('episode_json')
+        if (json_ld_info):
+            info = json_ld_info
+        else:
+            escaped_json_string = self._search_regex(
+                r'var episodeData = \$.parseJSON\("(?P<episode_json>.*)"\)',
+                webpage,
+                'episode json',
+                fatal=False,
+                group='episode_json'
+            )
 
+            if (escaped_json_string):
                 if sys.version_info[0] >= 3:
                     unescaped_json_string = bytes(
                         escaped_json_string, "utf-8").decode('unicode_escape')
                 else:
                     unescaped_json_string = escaped_json_string.decode(
                         'string_escape')
-                metadata = json.loads(unescaped_json_string)
+                metadata = self._parse_json(unescaped_json_string, ooyala_code)
                 info = {
                     'id': metadata.get('mediaId'),
                     'title': metadata.get('title'),
--- a/youtube_dl/extractor/tastytrade.py
+++ b/youtube_dl/extractor/tastytrade.py
@@ -3,8 +3,6 @@ from __future__ import unicode_literals
 from .common import InfoExtractor
 from .ooyala import OoyalaIE
 
-import sys
-
 
 class TastyTradeIE(InfoExtractor):
     _VALID_URL = r'https?://(?:www\.)?tastytrade\.com/tt/(?:shows|daily_recaps)/[^/]+/episodes/(?P<id>[^/?#&]+)'
@@ -32,8 +30,8 @@ class TastyTradeIE(InfoExtractor):
         'info_dict': {
             'id': 'lud3BtZTE6vnRdolxKRlwNoZQvb3z_LT',
             'ext': 'mp4',
-            'title': 'Soybeans & Corn: It\'s Planting Time',
-            'description': 'md5:a523504b1227de1b81faeba2876a6d23',
+            'title': 'TTL_CTGFE_180309_SEG_EDIT.mp4',
+            'description': 'md5:d41d8cd98f00b204e9800998ecf8427e',
         },
         'params': {
             'skip_download': True,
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/tastytrade.py
```diff
@@ -33,7 +54,30 @@ def _real_extract(self, url):
...
             webpage, 'ooyala code', group='code')
 
-        info = self._search_json_ld(webpage, display_id, fatal=False)
+        info = {'id': None, 'title': None, 'description': None}
+
+        try:
+            info = self._search_json_ld(webpage, display_id, fatal=False)
+        except ExtractorError as ex:
+            json_string_match = re.search(
+                r'var episodeData = \$.parseJSON\("(?P<episode_json>.*)"\)', webpage, 0)
```
dstftw: `self._search_regex`.

youtube_dl/extractor/tastytrade.py
```diff
@@ -33,7 +54,30 @@ def _real_extract(self, url):
...
+            if (json_string_match):
+                escaped_json_string = json_string_match.group('episode_json')
+
+                if sys.version_info[0] >= 3:
+                    unescaped_json_string = bytes(
+                        escaped_json_string, "utf-8").decode('unicode_escape')
+                else:
+                    unescaped_json_string = escaped_json_string.decode(
+                        'string_escape')
+                metadata = json.loads(unescaped_json_string)
```
dstftw: `self._parse_json`.

youtube_dl/extractor/tastytrade.py
```diff
@@ -2,19 +2,28 @@
...
+)
+
+import json
+import re
+import sys
 
 
 class TastyTradeIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?tastytrade\.com/tt/shows/[^/]+/episodes/(?P<id>[^/?#&]+)'
+    _VALID_URL = r'https?://(?:www\.)?tastytrade\.com/tt/(shows|daily_recaps)/[^/]+/episodes/(?P<id>[^/?#&]+)'
```
dstftw: No captures for groups you don't use.

youtube_dl/extractor/tastytrade.py
```diff
@@ -33,7 +54,30 @@ def _real_extract(self, url):
             r'data-media-id=(["\'])(?P<code>(?:(?!\1).)+)\1',
             webpage, 'ooyala code', group='code')
 
-        info = self._search_json_ld(webpage, display_id, fatal=False)
+        info = {'id': None, 'title': None, 'description': None}
+
+        try:
+            info = self._search_json_ld(webpage, display_id, fatal=False)
+        except ExtractorError as ex:
```
dstftw: Will never happen since it does not throw if not fatal.

thepelkus-too:
@dstftw I believe all of the comments have been addressed. Please let me know if you think there are any other changes that are warranted, and thank you for your review!

youtube_dl/extractor/tastytrade.py
```diff
@@ -33,7 +49,36 @@ def _real_extract(self, url):
...
+            if (escaped_json_string):
+                if sys.version_info[0] >= 3:
+                    unescaped_json_string = bytes(
+                        escaped_json_string, "utf-8").decode('unicode_escape')
+                else:
+                    unescaped_json_string = escaped_json_string.decode(
+                        'string_escape')
+                metadata = self._parse_json(unescaped_json_string, ooyala_code)
+                info = {
+                    'id': metadata.get('mediaId'),
```
dstftw: No.

youtube_dl/extractor/tastytrade.py
```diff
@@ -33,7 +49,36 @@ def _real_extract(self, url):
...
+            )
+
+            if (escaped_json_string):
+                if sys.version_info[0] >= 3:
+                    unescaped_json_string = bytes(
+                        escaped_json_string, "utf-8").decode('unicode_escape')
+                else:
+                    unescaped_json_string = escaped_json_string.decode(
+                        'string_escape')
+                metadata = self._parse_json(unescaped_json_string, ooyala_code)
```
dstftw: Must not be fatal.

youtube_dl/extractor/tastytrade.py
```diff
@@ -33,7 +49,36 @@ def _real_extract(self, url):
...
+            escaped_json_string = self._search_regex(
+                r'var episodeData = \$.parseJSON\("(?P<episode_json>.*)"\)',
+                webpage,
+                'episode json',
+                fatal=False,
+                group='episode_json'
+            )
+
+            if (escaped_json_string):
+                if sys.version_info[0] >= 3:
```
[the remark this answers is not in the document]
thepelkus-too: The processing done here was to convert an escaped json string into an unescaped string so that it could be parsed properly. I'm open to suggestions for other ways of achieving the same goal.

youtube_dl/extractor/tastytrade.py
thepelkus-too: I'm not sure exactly what's meant by this comment. Could you provide more context?

youtube_dl/extractor/tastytrade.py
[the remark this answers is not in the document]
dstftw: >I'm open to suggestions for other ways of achieving the same goal.

Did you read my message at all? `escaped_json_string.encode('utf-8').decode('unicode_escape')`.

youtube_dl/extractor/tastytrade.py
dstftw: You are using different ids in url and here. Don't do that.

thepelkus-too:
@dstftw : I've removed the majority of the changes in the hopes of simplifying the PR while still adding the capability of handling the new set of URLs that I'd like this extractor to handle.

youtube_dl/extractor/tastytrade.py
thepelkus-too: Ah, okay. Thanks for the clarification.

youtube_dl/extractor/tastytrade.py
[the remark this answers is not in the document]
thepelkus-too: Ah, okay, I misunderstood. I thought you were saying that that had already been applied somewhere, or that it had been effectively applied already. Now I see that you're saying that it can done to the return from the method.

PR #15840, opened 2018-03-12
FFmpegCutVideo postprocessor, for cutting the video after download

marnunez:

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [X] New feature

---

This is a simple FFmpeg postprocessor for cutting videos at start-time and end-time:

`youtube-dl "URL" --start-time 10 --end-time 12`

There is a similar pull request for this feature that's nearly two years old and hasn't been worked for a long time now. This is understandable a hack: the whole video is downloaded for cutting just a segment. But I believe there is a need for this kind of postprocessing, and I've seen many a one-liner for this in the forums. It updates video duration if available, validates start-time and end-time and cuts only when needed, keeping all original stream codecs (-c copy). I believe this could do for the time being.

### DIFF
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -283,6 +283,13 @@ def _real_main(argv=None):
         postprocessors.append({
             'key': 'FFmpegEmbedSubtitle',
         })
+    if opts.starttime or opts.endtime:
+        d = {'key': 'FFmpegCutVideo'}
+        if opts.starttime:
+            d['startTime'] = int(opts.starttime)
+        if opts.endtime:
+            d['endTime'] = int(opts.endtime)
+        postprocessors.append(d)
     if opts.embedthumbnail:
         already_have_thumbnail = opts.writethumbnail or opts.write_all_thumbnails
         postprocessors.append({
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -215,6 +215,59 @@ class FFmpegPostProcessor(PostProcessor):
         return 'file:' + fn if fn != '-' else fn
 
 
+class FFmpegCutVideoPP(FFmpegPostProcessor):
+    def __init__(self, downloader=None, startTime=0, endTime=None):
+        FFmpegPostProcessor.__init__(self, downloader)
+        self._startTime = startTime
+        self._endTime = endTime
+
+    def toTime(self, seconds):
+        m, s = divmod(seconds, 60)
+        h, m = divmod(m, 60)
+        return "%d:%02d:%02d" % (h, m, s)
+
+    def run(self, information):
+        if not self._startTime and not self._endTime:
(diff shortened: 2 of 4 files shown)


youtube_dl/__init__.py
```diff
@@ -283,6 +283,13 @@ def parse_retries(retries):
         postprocessors.append({
             'key': 'FFmpegEmbedSubtitle',
         })
+    if opts.starttime or opts.endtime:
+        d = {'key': 'FFmpegCutVideo'}
+        if opts.starttime:
+            d['startTime'] = int(opts.starttime)
```
dstftw: Incorrect case.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -215,6 +215,66 @@ def _ffmpeg_filename_argument(self, fn):
...
+            message += 'from %s ' % (start)
+            if duration:
+                duration -= self._startTime
+
+        if self._endTime:
+            end = self.toTime(self._endTime)
+            options.extend(['-to', end])
+            message += 'to %s' % (end)
+            if duration:
+                duration -= self._endTime
```
dstftw: Duration calculation is incorrect.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -215,6 +215,66 @@ def _ffmpeg_filename_argument(self, fn):
...
+
+        if '-to' not in options and '-ss' not in options:
+            self._downloader.to_screen('[ffmpeg] Nothing to cut. Keeping original')
+            return [], information
+
+        path = information['filepath']
+        temp_filename = prepend_extension(path, 'temp')
+        self._downloader.to_screen(message)
+        if duration:
+            information['duration'] = duration
```
dstftw: This has no effect. Postprocessors work on info dict copy.

dstftw:
1. It must support cutting into multiple parts in order to be used for cutting by chapters.
2. It must not cut if start time is greater than duration.
3. It must not cut already downloaded files.

marnunez:
I'm on it

youtube_dl/postprocessor/ffmpeg.py
marnunez: I've seen this used [here](https://github.com/marnunez/youtube-dl/blob/afc763ee2e193b5711e9af53b1e133031611e72d/youtube_dl/postprocessor/ffmpeg.py#L299) and [here](https://github.com/marnunez/youtube-dl/blob/afc763ee2e193b5711e9af53b1e133031611e72d/youtube_dl/postprocessor/ffmpeg.py#L354) with filepath and other keys. Does this work in this cases?

Aniruddh-J:
And maybe force downloader to `ffmpeg` if an external downloader is defined in config?

mohseneslami:
It's not working for oreilly working. please fix it.

oreilly change login url members

PR #15888, opened 2018-03-17
Fix --force-ipv4, --force-ipv6, --source-address

Tatsh:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

`--force-ipv4` and `--force-ipv6` are not reliable options in the current code. This mostly applies to `--force-ip4`. The `socket.create_connection()` function which is called by the HTTP connection class ignores if the port is `0` (second part of tuple). It also checks all values returned from `getaddrinfo()` which means if an IPv6 connection works first, that is the connection type that will be used. There is no way to filter for a particular type of family with this API, so this monkey-patches `_create_connection()` and is based on the latest [`socket.create_connection()`](https://github.com/python/cpython/blob/master/Lib/socket.py#L691) which has not been changed in a very long time.

This applies to Linux and macOS. Have not tried Windows.

`strace -e socket` output on `master` with `--force-ipv4`:

```
 $ strace -e socket youtube-dl --force-ipv4 'https://www.youtube.com/watch?v=QwievZ1Tx-8'
WARNING: parsing .netrc: No authenticators for youtube (None, line None)
[youtube] QwievZ1Tx-8: Downloading webpage
socket(AF_INET, SOCK_DGRAM|SOCK_CLOEXEC|SOCK_NONBLOCK, IPPROTO_IP) = 3
socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE) = 3
socket(AF_INET, SOCK_DGRAM|SOCK_CLOEXEC, IPPROTO_IP) = 3
socket(AF_INET6, SOCK_DGRAM|SOCK_CLOEXEC, IPPROTO_IP) = 3
socket(AF_INET6, SOCK_STREAM, IPPROTO_TCP) = 3
[youtube] QwievZ1Tx-8: Downloading video info webpage
```

We want `AF_INET` when `--force-ipv4` is used.

On this branch:

```
 $ strace -e socket youtube-dl --force-ipv4 'https://www.youtube.com/watch?v=QwievZ1Tx-8'
WARNING: parsing .netrc: No authenticators for youtube (None, line None)
[youtube] QwievZ1Tx-8: Downloading webpage
socket(AF_INET, SOCK_DGRAM|SOCK_CLOEXEC|SOCK_NONBLOCK, IPPROTO_IP) = 3
socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE) = 3
socket(AF_INET, SOCK_DGRAM|SOCK_CLOEXEC, IPPROTO_IP) = 3
socket(AF_INET6, SOCK_DGRAM|SOCK_CLOEXEC, IPPROTO_IP) = 3
socket(AF_INET, SOCK_STREAM, IPPROTO_TCP) = 3
[youtube] QwievZ1Tx-8: Downloading video info webpage
```

The single `socket(AF_INET6)` is (I believe) from calling `getaddrinfo()` (which tries all IPs found via DNS including IPv6). This address gets filtered out in the monkey-patched version of `_create_connection()`.

Using `--force-ipv6`/`--source-address ::`/no `--force-ipv4` leaves original functionality intact.

Issue: [IPv4 not working](https://github.com/rg3/youtube-dl/issues/13422)


youtube_dl/utils.py
```diff
@@ -881,7 +881,42 @@ def _create_http_connection(ydl_handler, http_class, is_https, *args, **kwargs):
         kwargs['strict'] = True
     hc = http_class(*args, **compat_kwargs(kwargs))
     source_address = ydl_handler._params.get('source_address')
+
     if source_address is not None:
+        filter_for = socket.AF_INET
+        if '.' not in source_address:
+            filter_for = socket.AF_INET6
```
dstftw: `filter_for = socket.AF_INET if '.' in source_address else socket.AF_INET6`.

youtube_dl/utils.py
```diff
@@ -881,7 +881,42 @@ def _create_http_connection(ydl_handler, http_class, is_https, *args, **kwargs):
...
+        # This is to workaround _create_connection() from socket where it will try all
+        # address data from getaddrinfo() including IPv6. This filters the result from
+        # getaddrinfo() based on the source_address value.
+        # This is based on the cpython socket.create_connection() function.
+        # https://github.com/python/cpython/blob/master/Lib/socket.py#L691
+        def _create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT, source_address=None):
+            host, port = address
+            err = None
+            addrs = socket.getaddrinfo(host, port, 0, socket.SOCK_STREAM)
+            ipv4_addrs = [addr for addr in addrs if addr[0] == filter_for]
```
dstftw: `ipv4_addrs` is a misleading name since ip version depends on `filter_for` and not necessarily equals v4.

youtube_dl/utils.py
```diff
@@ -881,7 +881,42 @@ def _create_http_connection(ydl_handler, http_class, is_https, *args, **kwargs):
...
+                    err = None  # Explicitly break reference cycle
+                    return sock
+                except socket.error as _:
+                    err = _
+                    if sock is not None:
+                        sock.close()
+            if err is not None:
+                raise err
+            else:
+                raise socket.error("getaddrinfo returns an empty list")
```
dstftw: That's not necessarily correct. It may just have all specified ip version addresses filtered out from non empty getaddrinfo addresses list. So should be reworded.

youtube_dl/utils.py
Tatsh: I made the fixes in [this pull request #17362](https://github.com/rg3/youtube-dl/pull/17362) because I deleted the repository and can't connect these.

PR #15927, opened 2018-03-20
[kaltura] add support for embeds specifying a kaltura session

jspiros:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Some sites that embed Kaltura videos include a `ks` session parameter that is required for downloading. To add support for those sites, I added a new regular expression that matches when such a parameter is provided and uses it appropriately.


youtube_dl/extractor/kaltura.py
```diff
@@ -110,6 +110,17 @@ class KalturaIE(InfoExtractor):
...
+                r"""(?xs)
+                    kWidget\.(?:thumb)?[Ee]mbed\(
+                    \{.*?
+                        (?P<q1>['"])wid(?P=q1)\s*:\s*
+                        (?P<q2>['"])_?(?P<partner_id>(?:(?!(?P=q2)).)+)(?P=q2),.*?
+                        (?P<q3>['"])entry_?[Ii]d(?P=q3)\s*:\s*
+                        (?P<q4>['"])(?P<id>(?:(?!(?P=q4)).)+)(?P=q4),.*?
+                        (?P<q5>['"])ks(?P=q5)\s*:\s*
+                        (?P<q6>['"])(?P<ks>(?:(?!(?P=q6)).)+)(?P=q6)(?:,|\s*\})
+                """, webpage) or
```
dstftw: Modify existing regex instead.

dstftw: Add a test.

youtube_dl/extractor/kaltura.py
```diff
@@ -117,7 +117,9 @@ def _extract_url(webpage):
                         (?P<q1>['"])wid(?P=q1)\s*:\s*
                         (?P<q2>['"])_?(?P<partner_id>(?:(?!(?P=q2)).)+)(?P=q2),.*?
                         (?P<q3>['"])entry_?[Ii]d(?P=q3)\s*:\s*
-                        (?P<q4>['"])(?P<id>(?:(?!(?P=q4)).)+)(?P=q4)(?:,|\s*\})
+                        (?P<q4>['"])(?P<id>(?:(?!(?P=q4)).)+)(?P=q4)(?:,.*?
+                        (?P<q5>['"])ks(?P=q5)\s*:\s*
+                        (?P<q6>['"])(?P<ks>(?:(?!(?P=q6)).)+)(?P=q6))?(?:,|\s*\})
```
dstftw: This breaks all non ks embeds. ks part must be optional.

jspiros: Er, it shouldn't break non-ks embeds, unless I'm rustier at Python regular expressions than I thought? Note that the ks part is wrapped in (?: )?, which should make it optional. I tested it against both types and didn't run into any issues, it would just set q5, q6, and ks to None in the match object for embeds without ks.

dstftw: This is unreadable. Indent optional group properly.

jspiros: Done. The original regular expression and its neighbors aren't particularly "readable" anyway, so what would be considered "proper" was unclear, but I hope you'll be happy with the latest commit.

dstftw: The original regex is clear line by line expression without any optional parts.

jspiros:
I'm not entirely sure how to add a test for these changes. Since I'm modifying an existing regex I first looked to see if there was a test that covered the types of embeds that the regex was previously handling (and still does), to build off of. But, it seems if there is any test of the original regex it's related to one (or more) of the `'only_matching': True` tests, and I'm a little unclear as to how to add or use those tests (I couldn't find any documentation or reference to them). I'm also not sure how to add a test for this functionality that wouldn't depend on a third-party site and hoping that they keep their videos up using exactly the same embeds that need testing...

Anyway, since the only remaining requested change to this PR is to add tests, this PR is currently stalled. For what it's worth, I have manually tested these changes both against the types of embeds previously caught by the regex, to ensure they're still caught with the modified version, as well as against the type of embeds that include the kaltura session (which is what this PR adds support for extracting). Everything works well.

PR #15983, opened 2018-03-25
New extractor for Torrins.com

Aniruddh-J:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This extractor requires you to supply login information.

### DIFF
--- a/youtube_dl/extractor/torrins.py
+++ b/youtube_dl/extractor/torrins.py
@@ -3,9 +3,7 @@ from __future__ import unicode_literals
 import re
 
 from .common import InfoExtractor
-from ..compat import (
-    compat_kwargs,
-)
+
 from ..utils import (
     ExtractorError,
     urlencode_postdata,
@@ -44,11 +42,6 @@ class TorrinsIE(InfoExtractor):
             if error_data:
                 error_str += ' - %s' % error_data['formErrors']
             raise ExtractorError(error_str, expected=True)
-
-    def _download_webpage(self, *args, **kwargs):
-        kwargs.setdefault('headers', {})['User-Agent'] = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_5) AppleWebKit/603.2.4 (KHTML, like Gecko) Version/10.1.1 Safari/603.2.4'
-        return super(TorrinsIE, self)._download_webpage(
-            *args, **compat_kwargs(kwargs))
 
     def _real_initialize(self):
         self._login()


youtube_dl/extractor/torrins.py
```diff
@@ -0,0 +1,136 @@
...
+
+        webpage = self._download_webpage(url, course_id)
+
+        title = self._og_search_title(webpage)
+
+        video_json = self._html_search_regex(r"<div id=\"video\" idata='(.+?)'", webpage, 'video formats')
+
+        video_json = self._parse_json(video_json, course_id)
+
+        title = video_json.get('title')
```
dstftw: Read coding conventions on optional fields.

youtube_dl/extractor/torrins.py
```diff
@@ -0,0 +1,136 @@
...
+                'height': 360,
+                'ext': 'mp4'
+            },
+            {
+                'url': video_json.get('high'),
+                'format_id': 'high',
+                'height': 480,
+                'ext': 'mp4'
+            }
+        ]
```
dstftw: Code duplication. `url` must not be `None`.

Aniruddh-J: Fixed.

youtube_dl/extractor/torrins.py
Aniruddh-J: Fixed.

youtube_dl/extractor/torrins.py
dstftw: Nothing changed.

youtube_dl/extractor/torrins.py
```diff
@@ -0,0 +1,136 @@
...
+        },
+        'skip': 'Requires torrins premium account credentials',
+    }
+
+    def _handle_error(self, response):
+        if not isinstance(response, dict):
+            return
+        error = response.get('error')
+        if error:
+            error_str = 'Torrins returned error #%s: %s' % (error.get('code'), error.get('message'))
```
dstftw: `error.get('code')`, `error.get('message')` may be `None`.

youtube_dl/extractor/torrins.py
```diff
@@ -0,0 +1,136 @@
...
+            error_data = error.get('data')
+            if error_data:
+                error_str += ' - %s' % error_data.get('formErrors')
+            raise ExtractorError(error_str, expected=True)
+
+    def _download_webpage(self, *args, **kwargs):
+        kwargs.setdefault('headers', {})['User-Agent'] = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_5) AppleWebKit/603.2.4 (KHTML, like Gecko) Version/10.1.1 Safari/603.2.4'
```
dstftw: Why is that necessary?

youtube_dl/extractor/torrins.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def _real_initialize(self):
+        self._login()
+
+    def _login(self):
+        (username, password) = self._get_login_info()
+        if username is None:
+            return
+
+        self._download_webpage(
```
dstftw: What's the point of this?

youtube_dl/extractor/torrins.py
```diff
@@ -0,0 +1,136 @@
...
+    ExtractorError,
+    urlencode_postdata,
+)
+
+
+class TorrinsIE(InfoExtractor):
+    IE_NAME = 'torrins'
+    _VALID_URL = r'''(?x)
+                    https?://
+                        www\.torrins\.com/(?:guitar|piano|bass)-lessons/(?:song-lessons|style-genre)/[^/]+/(?P<id>[^/]+)/(?P<lesson_id>[^(\.)]+)\.html
```
dstftw: `[^(\.)]+` invalid.

dstftw: Don't capture groups you don't use.

youtube_dl/extractor/torrins.py
```diff
@@ -0,0 +1,136 @@
...
+                    '''
+    _LOGIN_URL = 'https://www.torrins.com/services/user/sign-in'
+    _ORIGIN_URL = 'https://www.torrins.com'
+    _NETRC_MACHINE = 'torrins'
+
+    _TEST = {
+        'url': 'https://www.torrins.com/guitar-lessons/song-lessons/english-songs/another-brick-in-the-wall/song-demo.html',
+        'info_dict': {
+            'id': 'another-brick-in-the-wall',
+            'lesson_id': 'song-demo',
```
dstftw: No such meta field.

youtube_dl/extractor/torrins.py
Aniruddh-J: Torrins rejects request made from all non-browser user agents.

youtube_dl/extractor/torrins.py
Aniruddh-J: Wow! So many mistakes. This needs to be removed.

youtube_dl/extractor/torrins.py
dstftw: youtube-dl does not use non-browser user agent.

youtube_dl/extractor/torrins.py
Aniruddh-J: Well, `url` cannot be none, I am not doing `'url': video_json.get('low')`.

dstftw: Now you break extraction if any of these keys is missing.

youtube_dl/extractor/torrins.py
Aniruddh-J: In that case, I will clean this up.

youtube_dl/extractor/torrins.py
Aniruddh-J: From documentation:

> **Example**
> Say meta from the previous example has a title and you are about to extract it. Since title is a mandatory meta field you should end up with something like:
> 
> `title = meta['title']`
>
> If title disappears from meta in future due to some changes on the hoster's side the extraction would fail since title is mandatory. That's expected.

Isn't that expected?

dstftw: So what? How title applies here?

Aniruddh-J: I see what you mean here. I will make one more change.

youtube_dl/extractor/torrins.py
```diff
@@ -0,0 +1,134 @@
...
+                'ext': 'mp4'
+            })
+
+        if 'high' in video_json:
+            formats.append({
+                'url': video_json['high'],
+                'format_id': 'high',
+                'height': 480,
+                'ext': 'mp4'
+            })
```
dstftw: >Code duplication

Aniruddh-J:
Do you think anymore changes required in this pull request?

youtube_dl/extractor/torrins.py
```diff
@@ -0,0 +1,112 @@
...
+            'description': 'md5:c0d51f6f21ef4ec65f091055a5eef876',
+            'duration': 579.29,
+        },
+        'skip': 'Requires torrins premium account credentials',
+    }
+
+    def _handle_error(self, response):
+        if not isinstance(response, dict):
+            return
+        error = response['error']
```
dstftw: Breaks if no error key.

youtube_dl/extractor/torrins.py
```diff
@@ -0,0 +1,112 @@
...
+
+        title = video_json.get('title') or self._og_search_title(webpage)
+        video_id = video_json['id']
+
+        formats = []
+
+        for format, height in {'low': 240, 'medium': 360, 'high': 480}.items():
+            if format in video_json:
+                formats.append({
+                    'url': video_json['low'],
```
dstftw: Wrong key. Again breaks if no such key.

youtube_dl/extractor/torrins.py
```diff
@@ -0,0 +1,112 @@
...
+
+        video_json = self._html_search_regex(r"<div id=\"video\" idata='(.+?)'", webpage, 'video formats')
+        video_json = self._parse_json(video_json, course_id)
+
+        title = video_json.get('title') or self._og_search_title(webpage)
+        video_id = video_json['id']
+
+        formats = []
+
+        for format, height in {'low': 240, 'medium': 360, 'high': 480}.items():
```
dstftw: Don't shadow built-in names.

dstftw:
For further review you must provide account credentials.

youtube_dl/extractor/torrins.py
Aniruddh-J: Can't break. I am checking if the key exists at line 100.

Aniruddh-J:
Is there any secure way to send account details?

youtube_dl/extractor/torrins.py
Aniruddh-J: Not sure how can I sort the formats then? The website does not provide any width or height information for the media.

dstftw: How is this even related to my comment on not shadowing `format` built-in?

dstftw:
By mail in my profile.

PR #16030, opened 2018-03-29
fix for null "ie_key" value for soundcloud when dumping as json

DerTechNick:

- [X] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This pull fixes a bug occuring when you try to dump a soundcloud playlist/user. If you use the following command the value "ie_key" to determine the service/website is null:

command: `youtube-dl -j --flat-playlist https://soundcloud.com/jayeem_music/sets`
part of output: 
`{"_type": "url", "url": "https://soundcloud.com/jayeem_music/sets/distant", "ie_key": null, "id": "452707167"}`

### DIFF
--- a/youtube_dl/extractor/soundcloud.py
+++ b/youtube_dl/extractor/soundcloud.py
@@ -400,7 +400,7 @@ class SoundcloudPagedPlaylistBaseIE(SoundcloudPlaylistBaseIE):
             for e in collection:
                 permalink_url, entry_id = resolve_permalink_url((e, e.get('track'), e.get('playlist')))
                 if permalink_url:
-                    entries.append(self.url_result(permalink_url, video_id=entry_id))
+                    entries.append(self.url_result(permalink_url, self.ie_key(), entry_id))
 
             next_href = response.get('next_href')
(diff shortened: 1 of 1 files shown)


dstftw:
Null `ie_key` is perfectly legal and means that IE will be guessed from url.

DerTechNick:
yeah, but why don't just add it as an improvement? So you as a dump consumer can more easily recognize which service you have and you dont have to check the url with a regex. Also, the dumper is very inconsistent. For youtube there are just ids as url while for soundcloud and others there are complete links.

dstftw:
Cause it may be not a soundcloud URL at all.

DerTechNick:
This commit only affects soundcloud resources and no others. 
Look, I am creating a kind like music streaming service and I don't know what the user wants to stream. So i decided to use ytdl because of its wide range of supported sites. When the ie_key is null, I have to guess what I am dealing with and this pull is just the solution for this. I just don't have to check the url which might be none because of the YT id url thing. And I think some clarity would also help others instead of just "null".

PR #16059, opened 2018-04-01
Add svtplay series playlist support

wader:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Add support for SVT play series as playlist

### DIFF
--- a/youtube_dl/extractor/svt.py
+++ b/youtube_dl/extractor/svt.py
@@ -9,6 +9,8 @@ from ..utils import (
     dict_get,
     int_or_none,
     try_get,
+    urljoin,
+    compat_str,
 )
 
 
@@ -189,3 +191,58 @@ class SVTPlayIE(SVTBaseIE):
                     r'\s*\|\s*.+?$', '',
                     info_dict.get('episode') or self._og_search_title(webpage))
             return info_dict
+
+
+class SVTPlaylistIE(InfoExtractor):
+    IE_DESC = 'SVT Play serie'
+    _VALID_URL = r'https?://(?:www\.)?svtplay\.se/(?P<id>[^/?&#]+)'
+    IE_NAME = 'svtplay:serie'
+    _TESTS = [{
+        'url': 'https://www.svtplay.se/rederiet',
+        'info_dict': {
+            'id': 'rederiet',
+            'title': 'Rederiet',
+            'description': 'md5:505d491a58f4fcf6eb418ecab947e69e',
+        },
+        'playlist_mincount': 318,
+    }]
+
+    @classmethod
+    def suitable(cls, url):
+        return False if SVTIE.suitable(url) or SVTPlayIE.suitable(url) else super(SVTPlaylistIE, cls).suitable(url)
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        page = self._download_webpage(
+            url, video_id,
+            note='Downloading serie page',
+            errnote='unable to fetch serie page')
+
+        root_json = self._search_regex(
+            r'root\[\'__svtplay\'\]\s*=(.+);\n',
+            page, 'root')
+        root = self._parse_json(root_json, video_id)
+
+        metadata = root.get('metaData', {})
+        related_videos_accordion = root['relatedVideoContent']['relatedVideosAccordion']
+
+        entries = []
+        for season in related_videos_accordion:
+            videos = season.get('videos')
+            if not isinstance(videos, list):
+                continue
+
+            for video in videos:
+                content_url = video.get('contentUrl')
+                if not isinstance(content_url, compat_str):
+                    continue
+                entries.append(
+                    self.url_result(
+                        urljoin(url, content_url),
+                        ie=SVTPlayIE.ie_key(),
+                        video_title=video.get('title')
+                    ))
+
+        return self.playlist_result(
+            entries, video_id, metadata.get('title'), metadata.get('description'))
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -1031,6 +1031,7 @@ from .sunporno import SunPornoIE
 from .svt import (
     SVTIE,
     SVTPlayIE,
+    SVTPlaylistIE,
 )
 from .swrmediathek import SWRMediathekIE
 from .syfy import SyfyIE


dstftw:
Core tests broken.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,53 @@ def _real_extract(self, url):
                     r'\s*\|\s*.+?$', '',
                     info_dict.get('episode') or self._og_search_title(webpage))
             return info_dict
+
+
+class SVTPlaylistIE(InfoExtractor):
+    IE_DESC = 'SVT Play serie'
+    _VALID_URL = r'https?://(?:www\.)?svtplay\.se/(?P<id>[^/]+)($|\?.*)'
```
dstftw: No. Override `suitable` instead.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,53 @@ def _real_extract(self, url):
...
+    IE_NAME = 'svtplay:serie'
+    _TESTS = [{
+        'url': 'https://www.svtplay.se/rederiet',
+        'info_dict': {
+            'id': 'rederiet',
+            'title': 'Rederiet',
+            'description': 'Slaget om Östersjön är en kamp om miljarder, på liv och död. Under seriens gång för Dahlénrederiet en ständig kamp mot  andra rederier om farlederna mellan Sverige och Finland.  Dramaserien om familjen Dahlén, båten M/S Freja och dess besättning är den längsta tv-serien i svensk historia. Totalt sändes Rederiet i 20 säsonger mellan åren 1992-2002. I rollerna: Johannes Brost (Torbjörn "Joker" Jonasson), Hans V Engström (Uno Kronkvist), Gösta Prüzelius (Reidar Dahlén), Suzanne Reuter (Renate Dahlén), Lena Strömdahl (Yvonne Dahlén), Per Holmberg (Henrik Bjurhed), Kenneth Söderman (Tony Sjögren), Peter Harryson (Pehr Silver), Bert-Åke Varg (Gustav Sjögren), Bert-Åke Varg (Gustav Sjögren), Bengt Bauler (Carl Ericson), Ola Forssmed (Micki Sandell) m.fl.',
```
dstftw: `md5:...`.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,53 @@ def _real_extract(self, url):
...
+        },
+        'playlist_mincount': 318,
+    }]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        id = mobj.group('id')
```
dstftw: `_match_id`. Do not shadow built-in names.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,53 @@ def _real_extract(self, url):
...
+
+        root_json = self._search_regex(
+            'root\\[\'__svtplay\'\\]\\s*=(.*);\\s*root\\[\'__env\'\\]',
+            page, "root", default=None, fatal=True)
+        root = self._parse_json(root_json, id)
+
+        metadata = root["metaData"]
+
+        entries = []
+        for season in root["relatedVideoContent"]["relatedVideosAccordion"]:
```
dstftw: Single quotes.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,53 @@ def _real_extract(self, url):
...
+            for video in season["videos"]:
+                contentUrl = video.get("contentUrl")
+                if contentUrl is None:
```
dstftw: `if not content_url:`.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,53 @@ def _real_extract(self, url):
...
+        id = mobj.group('id')
+
+        page = self._download_webpage(
+            url, None,
+            note='Downloading serie page',
+            errnote='unable to fetch serie page', fatal=True)
+
+        root_json = self._search_regex(
+            'root\\[\'__svtplay\'\\]\\s*=(.*);\\s*root\\[\'__env\'\\]',
+            page, "root", default=None, fatal=True)
```
dstftw: default and fatal are not used together. `True` is default.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,53 @@ def _real_extract(self, url):
...
+    }]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        id = mobj.group('id')
+
+        page = self._download_webpage(
+            url, None,
+            note='Downloading serie page',
+            errnote='unable to fetch serie page', fatal=True)
```
dstftw: `True` is default.

wader:
@dstftw now core tests are green. thanks for the feedback

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,57 @@ def _real_extract(self, url):
...
+        'playlist_mincount': 318,
+    }]
+
+    @classmethod
+    def suitable(cls, url):
+        return False if SVTIE.suitable(url) or SVTPlayIE.suitable(url) else super(SVTPlaylistIE, cls).suitable(url)
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        _match_id = mobj.group('id')
```
dstftw: `self._match_id`.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,57 @@ def _real_extract(self, url):
...
+
+        page = self._download_webpage(
+            url, None,
```
dstftw: `None` is not an id.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,57 @@ def _real_extract(self, url):
...
+            note='Downloading serie page',
+            errnote='unable to fetch serie page')
+
+        root_json = self._search_regex(
+            'root\\[\'__svtplay\'\\]\\s*=(.*);\\s*root\\[\'__env\'\\]',
+            page, "root")
+        root = self._parse_json(root_json, _match_id)
+
+        metadata = root["metaData"]
```
dstftw: Quotes.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,57 @@ def _real_extract(self, url):
...
+
+        entries = []
+        for season in root['relatedVideoContent']['relatedVideosAccordion']:
+            for video in season['videos']:
+                contentUrl = video.get('contentUrl')
+                if not contentUrl:
```
dstftw: No camelCase.

wader:
@dstftw Sorry for all the small messups. I've fixed some more stuff also

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,56 @@ def _real_extract(self, url):
...
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        page = self._download_webpage(
+            url, video_id,
+            note='Downloading serie page',
+            errnote='unable to fetch serie page')
+
+        root_json = self._search_regex(
+            'root\\[\'__svtplay\'\\]\\s*=(.*);\\s*root\\[\'__env\'\\]',
```
dstftw: Use raw strings.

dstftw: Capturing empty string is senseless. `\n` instead of `root[__env]`.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,56 @@ def _real_extract(self, url):
...
+                        urljoin(url, content_url),
+                        ie=SVTPlayIE.ie_key(),
+                        video_title=video.get('title')
+                    ))
+
+        return self.playlist_result(
+            entries,
+            video_id,
+            metadata.get('title'),
+            metadata.get('description'))
```
dstftw: Eliminate excessive verbosity.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,56 @@ def _real_extract(self, url):
...
+            url, video_id,
+            note='Downloading serie page',
+            errnote='unable to fetch serie page')
+
+        root_json = self._search_regex(
+            'root\\[\'__svtplay\'\\]\\s*=(.*);\\s*root\\[\'__env\'\\]',
+            page, 'root')
+        root = self._parse_json(root_json, video_id)
+
+        metadata = root['metaData']
```
dstftw: Breaks extraction if missing.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,56 @@ def _real_extract(self, url):
                     r'\s*\|\s*.+?$', '',
                     info_dict.get('episode') or self._og_search_title(webpage))
             return info_dict
+
+
+class SVTPlaylistIE(InfoExtractor):
+    IE_DESC = 'SVT Play serie'
+    _VALID_URL = r'https?://(?:www\.)?svtplay\.se/(?P<id>[^/]+)'
```
dstftw: Captures `?#&` ending as id.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,56 @@ def _real_extract(self, url):
...
+        root_json = self._search_regex(
+            'root\\[\'__svtplay\'\\]\\s*=(.*);\\s*root\\[\'__env\'\\]',
+            page, 'root')
+        root = self._parse_json(root_json, video_id)
+
+        metadata = root['metaData']
+
+        entries = []
+        for season in root['relatedVideoContent']['relatedVideosAccordion']:
+            for video in season['videos']:
```
dstftw: Breaks if no videos in season.

wader:
Ready for merge?

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,67 @@ def _real_extract(self, url):
...
+            errnote='unable to fetch serie page')
+
+        root_json = self._search_regex(
+            r'root\[\'__svtplay\'\]\s*=(.+);\n',
+            page, 'root')
+        root = self._parse_json(root_json, video_id)
+
+        metadata = root.get('metaData')
+        if not metadata:
+            raise ExtractorError('root.metaData missing')
```
dstftw: Remove.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,67 @@ def _real_extract(self, url):
...
+
+        related_video_content = root.get('relatedVideoContent')
+        if not related_video_content:
+            raise ExtractorError('root.relatedVideoContent missing')
```
dstftw: Remove.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,67 @@ def _real_extract(self, url):
...
+
+        related_videos_accordion = related_video_content.get('relatedVideosAccordion')
+        if not related_videos_accordion:
+            raise ExtractorError('root.relatedVideosAccordion missing')
```
dstftw: Remove.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,67 @@ def _real_extract(self, url):
...
+
+        entries = []
+        for season in related_videos_accordion:
+            videos = season.get('videos')
+            if not videos:
```
dstftw: Should check for a list.

youtube_dl/extractor/svt.py
```diff
@@ -189,3 +190,67 @@ def _real_extract(self, url):
...
+                continue
+
+            for video in videos:
+                content_url = video.get('contentUrl')
+                if not content_url:
```
dstftw: Check for `compat_str`.

wader:
@dstftw  Thanks, will fix. Is there any extractor that follows all guide and style -lines? I’ve looked at other extractors for how to handle errors etc but they might be outdated?

dstftw:
Maybe. Try looking at some newly added extractors. The main point of all these checks is to foresee and minimize potential breakages due to unexpected data.

wader:
@dstftw Fixed. Not sure i follow if the idea is to return empty results on unexpected data or to crash with as useful error as possible? for expected errors like auth error etc `ExtractorError` should be used? Looking at some of the new extractors they seem to poke around in JSON without much errors checks, like https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/tennistv.py#L94

Let me know if i should add more check

dstftw:
The idea is not to crash on any missing optional data and crash on missing mandatory data. What do you mean by useful error? Crashing with stacktrace on `[]` operator accessing mandatory key in dict is much more useful than throwing something like `root.relatedVideosAccordion missing`. This information is for developer in the first place not end user. Both means nothing for end user but the former is much more beneficial for developer since it shows exact place of breakage.
tennistv is really bad in terms of robustness and will break on any minor change/unexpected data.

wader:
@dstftw Yes i agree. Handle optional and crash on unexpected with pyhon stacktrace. So i should change `metadata = root.get('metaData')` to `metadata = ['metaData']` and same for `season.get('videos')` and  `video.get('contentUrl')`?  They are not optional and i'm kind of certain they should always exist unless there has been some major change to the site.

dstftw:
No. You use `metaData` only for playlist title and description both of which are considered optional. `videos` and `contentUrl` should not be changed to mandatory as well since you process list of such data: some of which may be present, some - not. Even single one extracted `contentUrl` is better than a crash without anything.

wader:
@dstftw Ok thanks. I've updated the PR now. Only change was `metadata = root.get('metaData', {})` to make `{}` the default value to make `metadata.get('title')` etc don't crash later on.

wader:
@dstftw Thanks for the patience. Next PR will go smoother i think 👍

Issue #16134, opened 2018-04-09
not correct extract video info from www.keezmovies.com


wangxiaogangan:
- [x] Bug report (encountered problems with youtube-dl)

When i use keezmovies.py to download video info from 

**https://www.keezmovies.com/video/Cute-little-azn-chick-strips-for-the-camera-18092651**

the WebPage not find flashvars while a few page contian this item, but most of pages do not have this “flashvars”， the error msg is below：

[KeezMovies] Cute-little-azn-chick-strips-for-the-camera: Downloading webpage
ERROR: No video formats found; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "D:\PROJ\trunk\Url_Extractor\youtube-dl\youtube_dl\YoutubeDL.py", line 785, in extract_info
    ie_result = ie.extract(url)

parmjitv:
Hi @wangxiaogangan,

A lot of pages on the KeezMovies site are now using embedded URLs for the videos, rather than flash vars.  A new PR will submitted shortly to accommodate this in the extractor.

Cheers,

Parmjit V.

PR #16154, opened 2018-04-11
[KeezMovies] Added fallback to generic extractor (fixes issue #16134)

parmjitv:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Updated the extractor for the KeezMovies.com website to use the generic extractor as a fallback for video formats - issue #16134.  Also replaced the dead 404 test.

Cheers,

Parmjit V.


youtube_dl/extractor/keezmovies.py
```diff
@@ -55,7 +55,7 @@ def _extract_info(self, url):
         encrypted = False
 
         def extract_format(format_url, height=None):
-            if not isinstance(format_url, compat_str) or not format_url.startswith('http'):
+            if not isinstance(format_url, compat_str):
```
dstftw: No.

youtube_dl/extractor/keezmovies.py
```diff
@@ -105,7 +105,8 @@ def extract_format(format_url, height=None):
                 raise ExtractorError(
                     'Video %s is no longer available' % video_id, expected=True)
 
-        self._sort_formats(formats)
+        if len(formats) > 0:
+            self._sort_formats(formats)
```
dstftw: No.

youtube_dl/extractor/keezmovies.py
parmjitv: @dstftw - Can you please elaborate?  I believe the provided `format_url` value no longer contains the http/https prefix in the page markup.  Thank you!

youtube_dl/extractor/keezmovies.py
parmjitv: @dstftw - Please elaborate?  This will allow the `extract_format` call to complete cleanly in the event that no formats can be detected, so that the generic extractor can be picked up later.  Thanks!

dstftw: This method is used in another extractors. **Do not change the default behavior**.

parmjitv: Ah, you are correct!  My apologies, I completely neglected to check that.  I will make the necessary corrections.

parmjitv:
@dstftw - I have refactored the code to preserve the default behaviour and also accommodate the recent changes to the KeezMovies website, please review at your convenience.  Incidentally, setting the new `require_http_prefix` flag also fixes the extractors for the ExtremeTube and MofoSex sites - these are currently broken with the same 'video formats not found' error.  Thank you!

youtube_dl/extractor/keezmovies.py
```diff
@@ -20,32 +20,38 @@
...
                       if 'display_id' in mobj.groupdict()
                       else None) or mobj.group('id')
+        return video_id, display_id
 
-        webpage = self._download_webpage(
+    def _download_webpage_age_verified(self, url, display_id):
+        return self._download_webpage(
             url, display_id, headers={'Cookie': 'age_verified=1'})
 
+    def _extract_info(self, url, require_http_prefix=True):
```
dstftw: Remove new parameter. Check url to also start with `//` instead.

youtube_dl/extractor/keezmovies.py
```diff
@@ -122,7 +132,13 @@ def extract_format(format_url, height=None):
...
 
     def _real_extract(self, url):
-        webpage, info = self._extract_info(url)
+        video_id, display_id = self._get_ids(url)
+        webpage = self._download_webpage_age_verified(url, display_id)
+        embed_url = self._search_regex(
+            r'<iframe\s+id="embedPlayer"\s+src="//(.+?)"', webpage, 'embed url', default=None)
+        if embed_url is not None:
+            return self.url_result(embed_url, 'Generic')
+        webpage, info = self._extract_info(url, require_http_prefix=False)
```
dstftw: Do not change the order of extraction.

youtube_dl/extractor/keezmovies.py
```diff
@@ -122,7 +132,13 @@ def extract_format(format_url, height=None):
         }
 
     def _real_extract(self, url):
-        webpage, info = self._extract_info(url)
+        video_id, display_id = self._get_ids(url)
+        webpage = self._download_webpage_age_verified(url, display_id)
+        embed_url = self._search_regex(
+            r'<iframe\s+id="embedPlayer"\s+src="//(.+?)"', webpage, 'embed url', default=None)
```
dstftw: Delegate to generic always when original path found nothing.

youtube_dl/extractor/keezmovies.py
parmjitv: @dstftw - OK!  As we are keeping the same order of extraction, I believe we will also need to check if the `formats` collection contains any entries prior to calling `_sort_formats` - to accommodate those pages that contain only embedded urls intended for the generic extractor.  Is this reasoning correct?  Much appreciated!

parmjitv: @dstftw - Where should the `fatal` flag be inserted?  I don't believe the `_sort_formats` call accepts this flag.  Thanks!

parmjitv: @dstftw - Can you please clarify your comment?  Should a new `fatal` parameter also be passed to the `_sort_formats` call?  I suppose we can update the signature as `def _sort_formats(self, formats, field_preference=None, fatal=True)` to preserve the default behaviour in the other extractors.  Do you agree?

dstftw: `fatal` must be added to `_extract_info`. No changes to core code.

parmjitv: @dstftw - Thanks for the clarification!

parmjitv:
@dstftw - I have revised the code so that `_extract_info` will now accept a `fatal` parameter.  Please let me know if any further changes need to be made.  Thank you!

youtube_dl/extractor/keezmovies.py
```diff
@@ -105,7 +105,13 @@ def extract_format(format_url, height=None):
...
                     'Video %s is no longer available' % video_id, expected=True)
 
-        self._sort_formats(formats)
+        try:
+            self._sort_formats(formats)
+        except ExtractorError as e:
+            if fatal:
+                raise ExtractorError(e, expected=True)
+            else:
+                self._downloader.report_warning(e)
```
dstftw: No need for this noise.

youtube_dl/extractor/keezmovies.py
```diff
@@ -105,7 +105,13 @@ def extract_format(format_url, height=None):
                 raise ExtractorError(
                     'Video %s is no longer available' % video_id, expected=True)
 
-        self._sort_formats(formats)
+        try:
+            self._sort_formats(formats)
+        except ExtractorError as e:
+            if fatal:
+                raise ExtractorError(e, expected=True)
```
dstftw: Just rethrow active exception.

youtube_dl/extractor/keezmovies.py
```diff
@@ -122,7 +128,11 @@ def extract_format(format_url, height=None):
         }
 
     def _real_extract(self, url):
-        webpage, info = self._extract_info(url)
+        webpage, info = self._extract_info(url, fatal=False)
+        if not info['formats']:
+            embed_url = self._search_regex(
+                r'<iframe\s+id="embedPlayer"\s+src="//(.+?)"', webpage, 'embed url', default=None)
+            return self.url_result(embed_url or url, 'Generic')
```
dstftw: No. As already said delegate `url` to generic always when main path fails.

parmjitv:
@dstftw - These corrections have been made, thanks for the prompt review.  Is there anything else that needs attention?

PR #16174, opened 2018-04-13
Indavideo token

dnet:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

It seems that Indavideo has added an extra token to their JSON metadata, and if the token is not used as a URL parameter when requesting the actual video file, the request is denied with a 403 Forbidden response. This pull request extracts the token and appends it to the video URL transparently, using the standard Python 2/3 URL parsing tools to be as future-proof as possible.

### DIFF
--- a/youtube_dl/extractor/indavideo.py
+++ b/youtube_dl/extractor/indavideo.py
@@ -61,7 +61,9 @@ class IndavideoEmbedIE(InfoExtractor):
                 video_urls.append(flv_url)
 
         filesh = video.get('filesh')
-        formats = [self.video_url_to_format(video_url, filesh) for video_url in video_urls]
+        formats = [
+            self.video_url_to_format(video_url, filesh)
+            for video_url in video_urls]
         self._sort_formats(formats)
 
         timestamp = video.get('date')
--- a/youtube_dl/extractor/indavideo.py
+++ b/youtube_dl/extractor/indavideo.py
@@ -10,6 +10,7 @@ except ImportError:
     from urllib import urlencode
 
 from .common import InfoExtractor
+from ..compat import compat_str
 from ..utils import (
     int_or_none,
     parse_age_limit,
@@ -99,7 +100,7 @@ class IndavideoEmbedIE(InfoExtractor):
         height = int_or_none(self._search_regex(
             r'\.(\d{3,4})\.mp4(?:\?|$)', video_url, 'height', default=None))
         if height and filesh:
-            token = filesh.get(str(height))
+            token = filesh.get(compat_str(height))
             if token is not None:
                 us = urlsplit(video_url)
                 query = urlencode(parse_qsl(us.query) + [('token', token)])
--- a/youtube_dl/extractor/indavideo.py
+++ b/youtube_dl/extractor/indavideo.py
@@ -1,7 +1,13 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/indavideo.py
```diff
@@ -1,6 +1,14 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
+try:
+    # Python 3
+    from urllib.parse import urlsplit, urlunsplit, urlencode, parse_qsl
+except ImportError:
+    # Python 2
+    from urlparse import urlsplit, urlunsplit, parse_qsl
+    from urllib import urlencode
```
dstftw: `compat_*` functions.

youtube_dl/extractor/indavideo.py
```diff
@@ -90,6 +95,20 @@ def _real_extract(self, url):
             'formats': formats,
         }
 
+    def video_url_to_format(self, video_url, filesh):
+        height = int_or_none(self._search_regex(
+            r'\.(\d{3,4})\.mp4(?:\?|$)', video_url, 'height', default=None))
+        if height is not None and filesh is not None:
+            token = filesh.get(str(height))
```
dstftw: `compat_str`.

youtube_dl/extractor/indavideo.py
```diff
@@ -90,6 +95,20 @@ def _real_extract(self, url):
...
+            if token is not None:
+                us = urlsplit(video_url)
+                query = urlencode(parse_qsl(us.query) + [('token', token)])
+                video_url = urlunsplit((us.scheme, us.netloc, us.path, query, us.fragment))
```
dstftw: `update_url_query`.

youtube_dl/extractor/indavideo.py
```diff
@@ -90,6 +95,20 @@ def _real_extract(self, url):
             'formats': formats,
         }
 
+    def video_url_to_format(self, video_url, filesh):
+        height = int_or_none(self._search_regex(
+            r'\.(\d{3,4})\.mp4(?:\?|$)', video_url, 'height', default=None))
+        if height is not None and filesh is not None:
```
dstftw: `if height and filesh:`

dnet:
@dstftw wrote

> Does not work:

You invoked a URL with Indavideo directly, I submitted patches regarding embed URLs. So my PR didn't break existing working code, but rather fixed another part. Here's an example that doesn't work with the current `HEAD` but it works with my patch applied: https://embed.indavideo.hu/player/video/c8f72ea6ef

youtube_dl/extractor/indavideo.py
dnet: ACK acb14703b

youtube_dl/extractor/indavideo.py
dnet: ACK 425f79940

youtube_dl/extractor/indavideo.py
dnet: not necessary thanks to your helpful suggestion of `update_url_query` below

youtube_dl/extractor/indavideo.py
dnet: ACK f1a6deb23

dnet:
Ping?

dnet:
Is there a problem with my PR?

PR #16261, opened 2018-04-23
Add author, date and skill level variables to Lynda.com extractor (Update: skill level removed)

rootyElf:
Added author, date and skill level variables which can be used for the output template.


dstftw:
1. Read coding conventions on optional metadata and fix code appropriately.
2. You are not allowed to use just arbitrary meta field names. List of supported meta fields is well defined.

rootyElf:
I just updated the code in a way where it should provide better resistance to API changes and I removed the skill_level template.

Q: How would I go about adding a very site specific template variable? Would I abuse an existing variable which has not been used or is there just no way without defining a new variable name in the docs?

PR #16266, opened 2018-04-24
Trivial patch for SAMPLE-AES support

rincebrain:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Trivial patch to support SAMPLE-AES in video streams, which appears to just be AES-128 by another name.

Worked on the stream I initially ran into SAMPLE-AES with - without this patch, I get a lot of "SAMPLE-AES encryption is not supported yet"; with this patch, I get an intact media file with audio and video.
It appears to also work on the URL linked in #15720.

### DIFF
--- a/youtube_dl/downloader/hls.py
+++ b/youtube_dl/downloader/hls.py
@@ -30,7 +30,7 @@ class HlsFD(FragmentFD):
     @staticmethod
     def can_download(manifest, info_dict):
         UNSUPPORTED_FEATURES = (
-            r'#EXT-X-KEY:METHOD=(?!NONE|AES-128)',  # encrypted streams [1]
+            r'#EXT-X-KEY:METHOD=(?!NONE|AES-128|SAMPLE-AES)',  # encrypted streams [1]
             # r'#EXT-X-BYTERANGE',  # playlists composed of byte ranges of media files [2]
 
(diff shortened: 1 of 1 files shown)


dstftw:
SAMPLE-AES != AES-128 in the first place. Second, this PR effectively solves nothing. Native HLS downloader **does not support** SAMPLE-AES and will download **encrypted** media that **won't play** in media players.

rincebrain:
Really? Because the media I mentioned this "working" on certainly plays in media players after this.

I could certainly buy this not being the correct way to implement it, but it does produce working media.

dstftw:
Try it yourself with URL from the issue you linked.

rincebrain:
Huh, it worked and produced a playable file on the URL I tested that wasn't from that, but the one from that produces, as you say, a nonsense but structurally playable MP4 file.

I wonder if it's just that [the URL I was trying](https://www.cbs.com/shows/the-late-show-with-stephen-colbert/video/4QftZZnxECels7uB29PX_TAWeuhRPVhz/the-late-show-4-23-18-jeffrey-wright-ali-wentworth-brothers-osborne-/) advertises the option but has unencrypted options in the playlist, and it was bailing out early, or what. Because that certainly produces a watchable stream.

Issue #14582, opened 2017-10-25
[niconico] Downloading long videos requires heartbeat signals


Ristellise:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like this: `[x]`)
- Use the *Preview* tab to see what your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2017.10.20*. If it's not, read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2017.10.20**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add the `-v` flag to **your command line** you run youtube-dl with (`youtube-dl -v <your command line>`), copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):

```
Python Console:
youtube_dl.YoutubeDL({'username': "username",'password': 'password','verbose':True}).extract_info(url="http://www.nicovideo.jp/watch/sm29409391")

[debug] Encodings: locale cp1252, fs utf-8, out UTF-8, pref cp1252
[debug] youtube-dl version 2017.10.20
[debug] Python version 3.6.2 - Windows-10-10.0.15063-SP0
[debug] exe versions: ffmpeg 3.3.3, ffprobe 3.3.3
[debug] Proxy map: {}
...
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "C:\Users\joshw\AppData\Local\Programs\Python\Python36-32\lib\site-packages\youtube_dl\YoutubeDL.py", line 795, in extract_info
    return self.process_ie_result(ie_result, download, extra_info)
[...]

PR #16495, opened 2018-05-19
[niconico] use HeartBeat

nekojun:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

I made a reference to the code that commented on a bugfix of #14582 issue.
Since niconico shuts down the connection unless HeartBeat is transmitted while acquiring a movie, HeartBeat is sent every 25 seconds.
Communication not requiring HeartBeat is communicated by default.

Use google translation

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -96,6 +96,7 @@ from .extractor import get_info_extractor, gen_extractor_classes, _LAZY_LOADER
 from .extractor.openload import PhantomJSwrapper
 from .downloader import get_suitable_downloader
 from .downloader.rtmp import rtmpdump_version
+from .downloader.httpfd import HttpHB
 from .postprocessor import (
     FFmpegFixupM3u8PP,
     FFmpegFixupM4aPP,
@@ -1839,7 +1840,12 @@ class YoutubeDL(object):
         if not self.params.get('skip_download', False):
             try:
                 def dl(name, info):
-                    fd = get_suitable_downloader(info, self.params)(self, self.params)
+                    if 'heartbeat_url' in info:
+                        print("use HTTPHB")
+                        fd = HttpHB(self, self.params)
+                    else:
+                        fd = get_suitable_downloader(info, self.params)(self, self.params)
+
                     for ph in self._progress_hooks:
                         fd.add_progress_hook(ph)
                     if self.params.get('verbose'):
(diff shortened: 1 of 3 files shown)


youtube_dl/downloader/httpfd.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import unicode_literals
+
+from .http import HttpFD
+
+try:
+    import urllib.request as compat_urllib_request
+except ImportError:  # Python 2
+    import urllib2 as compat_urllib_request
```
yan12125: Any reason not reusing symbols from compat.py?

youtube_dl/downloader/httpfd.py
```diff
@@ -0,0 +1,32 @@
...
+import threading
+
+
+class HttpHB(HttpFD):
```
yan12125: The class name should ends with FD

youtube_dl/YoutubeDL.py
```diff
@@ -1839,7 +1840,12 @@ def ensure_dir_exists(path):
         if not self.params.get('skip_download', False):
             try:
                 def dl(name, info):
-                    fd = get_suitable_downloader(info, self.params)(self, self.params)
+                    if 'heartbeat_url' in info:
+                        print("use HTTPHB")
```
yan12125: Don't use print() directly as -j will be broken. Use to_screen or related functions instead.

youtube_dl/YoutubeDL.py
```diff
@@ -1839,7 +1840,12 @@ def ensure_dir_exists(path):
         if not self.params.get('skip_download', False):
             try:
                 def dl(name, info):
-                    fd = get_suitable_downloader(info, self.params)(self, self.params)
+                    if 'heartbeat_url' in info:
```
yan12125: Such logic should be implemented in get_suitable_downloader()

youtube_dl/downloader/httpfd.py
yan12125: Thining again and I believe there's no need for a separate class. The heart beat function can be implemented in the FileDownloader class. (not in HttpFD as other stream types, including HLS and HDS, may also need heart beats)

youtube_dl/downloader/httpfd.py
```diff
@@ -0,0 +1,32 @@
...
+                try:
+                    data = info_dict['heartbeat_data'].encode("utf-8")
+                    compat_urllib_request.urlopen(url=info_dict['heartbeat_url'], data=data)
+                    print('heart beat!')
+                except Exception as ex:
+                    print('heart beat fail: ' + ex.message)
+                    pass
+
+                if not result:
+                    timer = threading.Timer(25, heart_beat)
```
yan12125: The number 25 should be configurable

PR #16507, opened 2018-05-21
handle wrong formatted date argument

Arlon1:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

The provided date argument (--date, --datebefore or --dateafter) is parsed by
datetimp.strptime(). This function throws as ValueError exception at failure.

This patch implements exception handling by catching the ValueError exception,
printing an error message and exiting with a non-zero exit code.

### DIFF
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -1270,7 +1270,10 @@ def date_from_str(date_str):
         unit += 's'
         delta = datetime.timedelta(**{unit: time})
         return today + delta
-    return datetime.datetime.strptime(date_str, '%Y%m%d').date()
+    try:
+        return datetime.datetime.strptime(date_str, '%Y%m%d').date()
+    except ValueError:
+        sys.exit('ERROR: cannot parse datestring ' + date_str)
 
 
 def hyphenate_date(date_str):


youtube_dl/utils.py
```diff
@@ -1270,7 +1270,10 @@ def date_from_str(date_str):
         unit += 's'
         delta = datetime.timedelta(**{unit: time})
         return today + delta
-    return datetime.datetime.strptime(date_str, '%Y%m%d').date()
+    try:
+        return datetime.datetime.strptime(date_str, '%Y%m%d').date()
+    except ValueError:
+        sys.exit('ERROR: cannot parse datestring ' + date_str)
```
dstftw: Never ever do that. You must handle it at the place of actual call the method.

Arlon1: Could you please briefly explain, why this exception should be handled in the surrounding of every function call of date_from_string() instead of inside the function?
Catching the exception inside the function should produce less code.

dstftw: Cause it's an **utility function** and it's used in other places you've broken with this change.

Issue #16502, opened 2018-05-20
unable to extract OpenGraph title from CBC.CA


johnnytornado3:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like this: `[x]`)
- Use the *Preview* tab to see what your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2018.05.18*. If it's not, read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x ] I've **verified** and **I assure** that I'm running youtube-dl **2018.05.18**

### Before submitting an *issue* make sure you have:
- [x ] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
- [x] Checked that provided video/audio/playlist URLs (if any) are alive and playable in a browser

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---
C:\Documents and Settings\User\Start Menu\Video>youtube-dl -v http://www.cbc.ca/
natureofthings/episodes/the-kingdom-how-fungi-made-our-world
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', 'http://www.cbc.ca/natureofthings/episodes/the
-kingdom-how-fungi-made-our-world']
[debug] Encodings: locale cp1252, fs mbcs, out cp437, pref cp1252
[debug] youtube-dl version 2018.05.18
[debug] Python version 3.4.4 (CPython) - Windows-XP-5.1.2600-SP3
[debug] exe versions: ffmpeg N-77883-gd7c75a5, ffprobe N-77883-gd7c75a5, phantom
js 1.9.7
[debug] Proxy map: {}
[...]

PR #16527, opened 2018-05-22
[CBC] Fix title extraction (fixes issue #16502)

parmjitv:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Updated the extractor for the cbc.ca website to properly fetch the video title - issue #16502.

From the page source, the order of extraction should be: 1) OpenGraph title if available, 2) Twitter title if available, 3) Document title otherwise.

Cheers,

Parmjit V.

### DIFF
--- a/youtube_dl/extractor/cbc.py
+++ b/youtube_dl/extractor/cbc.py
@@ -20,6 +20,7 @@ from ..utils import (
     parse_duration,
     parse_iso8601,
     parse_age_limit,
+    strip_or_none,
     int_or_none,
     ExtractorError,
 )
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/cbc.py
```diff
@@ -129,15 +129,17 @@ def _extract_player_init(self, player_init, display_id):
     def _real_extract(self, url):
         display_id = self._match_id(url)
         webpage = self._download_webpage(url, display_id)
+        title = self._og_search_title(webpage, default=None) or self._html_search_meta(
+            'twitter:title', webpage, 'title', default=None) or self._html_search_regex(
+                r'<title>(.*?)</title>', webpage, 'title', fatal=True)
```
dstftw: Playlist title must not be fatal. Do not capture empty strings.

parmjitv:
@dstftw - Thanks for the prompt review.  The regex for the document title is now set as non-fatal.

I would expect the document title will be set in the page source to some non-trivial string, so the fatal flag here is somewhat superfluous.  Does this address your concern about capturing an empty string?

dstftw:
Nothing changed, how is this even related to capturing empty strings?

parmjitv:
@dstftw - Ah, of course my regex was off for empty strings.  Thanks for fixing that!

PR #16552, opened 2018-05-26
Fix incorrect line-break in README

jose1711:


dstftw:
This file is auto generated.

PR #16568, opened 2018-05-28
[mediaset] Update extractor for 2018

timendum:

- [x] Bug fix

---

Mediaset has updated the website and newer videos now have a different id on the CDN, this change update the code to reflect the changes.

### DIFF
--- a/youtube_dl/extractor/mediaset.py
+++ b/youtube_dl/extractor/mediaset.py
@@ -10,6 +10,7 @@ from ..utils import (
     parse_duration,
     try_get,
     unified_strdate,
+    ExtractorError
 )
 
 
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/mediaset.py
```diff
@@ -70,10 +86,21 @@ def _extract_urls(webpage):
...
         video_id = self._match_id(url)
 
+        media_info = self._download_json(
+            'https://www.video.mediaset.it/html/metainfo.sjson',
+            video_id, 'Downloading media info', query={
+                'id': video_id
+            })
+
+        if media_info.get('video') and media_info['video'].get('guid'):
+            media_id = media_info['video']['guid']
```
dstftw: `try_get`.

timendum:
I've reviewed the code, it's cleaner now. Thanks for the comment.

youtube_dl/extractor/mediaset.py
```diff
@@ -70,10 +86,18 @@ def _extract_urls(webpage):
     def _real_extract(self, url):
         video_id = self._match_id(url)
 
+        media_info = self._download_json(
+            'https://www.video.mediaset.it/html/metainfo.sjson',
+            video_id, 'Downloading media info', query={
+                'id': video_id
+            })
```
dstftw: metainfo is downloaded twice.

timendum:
I've removed the second download of the same info and added a try/except block to go on in case of error in downloading ISM manifest.

PR #16571, opened 2018-05-29
fix args order in FFmpegFD

noureddin:

- [x] Bug fix

---

In `youtube-dl --external-downloader ffmpeg --external-downloader-args '-ss 1 -t 2'`, youtube-dl will give all the external downloader args to ffmpeg before the input url; but some of them need to be given after it.

For example, ffmpeg will refuse `-to` in the args without this patch. And it won't stop downloading after downloading the specified duration if given `-ss START_TIME -t DURATION` without this patch. Both happen because `-t` and `-to` (and `-fs`) need to come after the input (i.e., before the output).

This patch moves `-t`, `-to`, and `-fs` (if given) to be after the input.

### DIFF
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -233,7 +233,24 @@ class FFmpegFD(ExternalFD):
             # http://trac.ffmpeg.org/ticket/6125#comment:10
             args += ['-seekable', '1' if seekable else '0']
 
-        args += self._configuration_args()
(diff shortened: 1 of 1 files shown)


cvzi:
I tested it with current ffmpeg version 4.3.2-2021-02-27 and found:

*   `-to` should actually be before the input for the download to stop
*   `-t` can be before or after the input, seems to make no difference

PR #16598, opened 2018-06-01
[DW] Update extractor with /playersources API

effleurager:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Makes use of the JSON API http://www.dw.com/playersources/ to find all formats of media available to the website player, instead of only the default version listed as a hidden input on the page.

### DIFF
--- a/youtube_dl/extractor/dw.py
+++ b/youtube_dl/extractor/dw.py
@@ -5,6 +5,7 @@ from .common import InfoExtractor
 from ..utils import (
     int_or_none,
     unified_strdate,
+    ExtractorError
 )
 from ..compat import compat_urlparse
 
@@ -12,6 +13,7 @@ from ..compat import compat_urlparse
 class DWIE(InfoExtractor):
     IE_NAME = 'dw'
     _VALID_URL = r'https?://(?:www\.)?dw\.com/(?:[^/]+/)+(?:av|e)-(?P<id>\d+)'
+    _SOURCES_URL = 'http://www.dw.com/playersources/%s-%s'
     _TESTS = [{
         # video
         'url': 'http://www.dw.com/en/intelligent-light/av-19112290',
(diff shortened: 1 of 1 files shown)


effleurager:
That's fixed now.

youtube_dl/extractor/dw.py
```diff
@@ -55,13 +68,30 @@ def _real_extract(self, url):
...
+                        'format_id': medium.get('label')
+                    })
+                self._sort_formats(formats, 'label')
+        elif hidden_inputs.get('player_type') == 'audio':
+            media = self._download_json(self._SOURCES_URL % ('a', media_id), media_id)
+            formats = []
+            for medium in media:
+                formats.append({
+                    'url': medium.get('file'),
+                })
```
dstftw: Code duplication 80-86, 89-94.

youtube_dl/extractor/dw.py
```diff
@@ -55,13 +68,30 @@ def _real_extract(self, url):
...
+                    transform_source=lambda s: s.replace(
+                        'rtmp://tv-od.dw.de/flash/',
+                        'http://tv-download.dw.de/dwtv_video/flv/'))
+                self._sort_formats(formats)
+            else:
+                media = self._download_json(self._SOURCES_URL % ('v', media_id), media_id)
+                formats = []
+                for medium in media:
+                    formats.append({
+                        'url': medium.get('file'),
```
dstftw: Must not be `None`.

youtube_dl/extractor/dw.py
```diff
@@ -55,13 +68,30 @@ def _real_extract(self, url):
...
+                        'format_id': medium.get('label')
+                    })
+                self._sort_formats(formats, 'label')
```
dstftw: `field_preference` must be a list or a tuple.

youtube_dl/extractor/dw.py
effleurager: > Must not be None.

If ```medium.get('file')``` returns ```None```, it would mean the underlying API has changed. From what I've understood, ```media``` cannot be ```None```. What exactly were you referring to?

dstftw: Whether it has changed or not does not mean there should be a format with invalid URL.

effleurager: Is ```ERROR: No video formats found; please report this issue...``` the preferred response? The format can be changed to be conditionally added in that case.

effleurager: Looking at the [API response](https://www.dw.com/playersources/v-44026376) using the added test case: 
```json
[ {
  "file" : "https://tvdownloaddw-a.akamaihd.net/dwtv_video/flv/vdt/2018/beng180531_004_qwibanweb_01g_sd_avc.mp4",
  "label" : "720"
}, {
  "file" : "https://tvdownloaddw-a.akamaihd.net/dwtv_video/flv/vdt/2018/beng180531_004_qwibanweb_01g_sd_sor.mp4",
  "label" : "288"
} ]
```
Every format should have ```file```. In the event of it not being present, it will fallback to the URL loaded in the page (which is still the current fallback).

effleurager:
The requested changes have been outdated. @dstftw

Issue #15373, opened 2018-01-20
PBS - Some Videos Error Out - AttributeError: 'NoneType' object has no attribute 'group'


SPDurkee:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like this: `[x]`)
- Use the *Preview* tab to see what your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2018.01.18*. If it's not, read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2018.01.18**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add the `-v` flag to **your command line** you run youtube-dl with (`youtube-dl -v <your command line>`), copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):

```
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', 'http://www.pbs.org/wgbh/roadshow/watch/episode/2105-indianapolis-hour-2/']
[debug] Encodings: locale cp1252, fs mbcs, out cp437, pref cp1252
[debug] youtube-dl version 2018.01.18
[debug] Python version 3.4.4 (CPython) - Windows-10-10.0.16299
[...]

PR #16623, opened 2018-06-03
[PBS] Fix AttributeError: 'NoneType'

Urgau:

- [X] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

This is a fix for #15373

### DIFF
--- a/youtube_dl/extractor/pbs.py
+++ b/youtube_dl/extractor/pbs.py
@@ -374,6 +374,7 @@ class PBSIE(InfoExtractor):
             },
             'expected_warnings': ['HTTP Error 403: Forbidden'],
         },
+        {
             'url': 'http://www.pbs.org/wgbh/roadshow/watch/episode/2105-indianapolis-hour-2/',
             'info_dict': {
                 'id': '2365936247',
--- a/youtube_dl/extractor/pbs.py
+++ b/youtube_dl/extractor/pbs.py
@@ -361,6 +361,21 @@ class PBSIE(InfoExtractor):
             },
         },
         {
+            'url': 'https://www.pbs.org/wgbh/masterpiece/episodes/victoria-s2-e1/',
+            'info_dict': {
+                'id': '3007193718',
+                'ext': 'mp4',
+                'title': 'Victoria - A Soldier\'s Daughter / The Green-Eyed Monster',
+                'description': 'md5:37efbac85e0c09b009586523ec143652',
+                'duration': 6292,
+                'thumbnail': r're:^https?://.*\.jpg$',
+            },
+            'params': {
+                'skip_download': True,
+            },
+            'expected_warnings': ['HTTP Error 403: Forbidden'],
+        },
+        {
             'url': 'http://player.pbs.org/widget/partnerplayer/2365297708/?start=0&end=0&chapterbar=false&endscreen=false&topbar=true',
             'only_matching': True,
         },
--- a/youtube_dl/extractor/pbs.py
+++ b/youtube_dl/extractor/pbs.py
@@ -467,7 +467,7 @@ class PBSIE(InfoExtractor):
                 url, display_id, note='Downloading player page',
                 errnote='Could not download player page')
             video_id = self._search_regex(
-                [r'<div\s+id="video_([0-9]+)"', r'"id":[\s]*"([0-9]+)"'], player_page, 'video ID')
+                [r'<div\s+id="video_([0-9]+)"', r'"id":\s*"([0-9]+)"'], player_page, 'video ID')
         else:
             video_id = mobj.group('id')
             display_id = video_id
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/pbs.py
```diff
@@ -465,8 +467,13 @@ def _extract_webpage(self, url):
             player_page = self._download_webpage(
                 url, display_id, note='Downloading player page',
                 errnote='Could not download player page')
-            video_id = self._search_regex(
-                r'<div\s+id="video_([0-9]+)"', player_page, 'video ID')
+            
+            try:
+                video_id = self._search_regex(
+                    r'<div\s+id="video_([0-9]+)"', player_page, 'video ID')
+            except:
```
dstftw: Never use bare except. Use list of regexes in `_search_regex` instead.

youtube_dl/extractor/pbs.py
```diff
@@ -455,7 +455,9 @@ def _extract_webpage(self, url):
 
             if not url:
                 url = self._og_search_url(webpage)
-
+            
+            if url.strip().startswith("//"):
```
dstftw: Single quotes.

youtube_dl/extractor/pbs.py
```diff
@@ -455,7 +455,9 @@ def _extract_webpage(self, url):
...
+                url = "https:" + url.strip()
```
dstftw: `_proto_relative_url`.

youtube_dl/extractor/pbs.py
```diff
@@ -466,7 +467,7 @@ def _extract_webpage(self, url):
                 url, display_id, note='Downloading player page',
                 errnote='Could not download player page')
             video_id = self._search_regex(
-                r'<div\s+id="video_([0-9]+)"', player_page, 'video ID')
+                [r'<div\s+id="video_([0-9]+)"', r'"id":[\s]*"([0-9]+)"'], player_page, 'video ID')
```
dstftw: `[]` is pointless. Must be more relaxed.

Urgau: What ? I need to use [] for making a list like you want.

dstftw: I'm not about list.

Urgau: So execuse me but I don't understand your comment.
What do I have to change?

Urgau: Sorry I thought you were talking about the `[]` around the 2 regex.

Urgau:
Is everythings good for merge ?

Urgau:
Is there anything else to do before the merge ?

dstftw:
Add a test.

youtube_dl/extractor/pbs.py
```diff
@@ -466,7 +467,7 @@ def _extract_webpage(self, url):
                 url, display_id, note='Downloading player page',
                 errnote='Could not download player page')
             video_id = self._search_regex(
-                r'<div\s+id="video_([0-9]+)"', player_page, 'video ID')
+                [r'<div\s+id="video_([0-9]+)"', r'"id":\s*"([0-9]+)"'], player_page, 'video ID')
```
dstftw: I've already pointed out: relax regex. It must match arbitrary quotes and whitespace.

Urgau: I'm sorry but I can't change the new regex otherwise it will point to other things that are very similar but incorrect.

dstftw: Actually this approach is redundant since id can be extracted right from `webpage`.

Urgau: In rare cases it cannot be extracted directly from the webpage (like embed video)

dstftw: That's not an excuse for not adding another id patern for extraction from `webpage` that works fine with your test. Also add a test with "rare" case.

Urgau: Yes, it seems that it works due to this commit (https://github.com/rg3/youtube-dl/commit/d7be7053082055a001b788453d66131a62692b55) which was published after my pull request but merge before me. However my test is valid without the previous commit.

I also maintain that my id extraction can serve but if you insist I can also remove it.

Urgau:
This pull request is also a fix for #16684

youtube_dl/extractor/pbs.py
dstftw: Again: https://www.pbs.org/wgbh/masterpiece/episodes/victoria-s2-e1/ can be processed **without** downloading `player_page` by looking for media id in `webpage`.
This regex is invalid because it matches several ids, it the order changes you'll end up downloading different video.

PR #12516, opened 2017-03-21
Raise exception if jwplayer doesn't have "sources"

moyamo:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [x] At least skimmed through [adding new extractor tutorial](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/rg3/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

YoutubeDL uses a regexp in common.py::_find_jwplayer_data to find the jwplayer options. However the options are found in a javascript function. For example the regexp might match this

    jwplayer('some_string').setup({
        /** Other attributes */
        sources: {
             file: "<url of video>",
             label: "<title of video>",
             type: "mp4"
        }
    });

Since this a valid javascript function, some websites write the options as

    var src = {
        file: "<url of video>",
        label: "<title of video>",
        type: "mp4"
    }
    jwplayer('some_string').setup({
        /** Other attributes */
        sources: src
    });

In this case YoutubeDL won't be able to retrieve sources.file, since the regexp only matches the `.setup(...)` and ignores the `var src = ...` assignment.

This commit makes YoutubeDL raise an ExtractorError in the above case. YoutubeDL will then try alternative methods to retrieve the URL of the video.

This bug is similar to https://github.com/rg3/youtube-dl/pull/12307 . In #12307 The entire `.setup(...)` dict has been extracted into a variable in this bug only the `sources` field has been extracted.

PR #16735, opened 2018-06-14
Continue trying after _parse_jwplayer_data fails

moyamo:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This pull request supersedes #12516. Essentially the problem is that `_parse_jwplayer_data` fails when the "JSON" structure it extracts contains variable references. However it seems an extractor further down in generic.py still works.

This pull requests makes it so that if  `_parse_jwplayer_data` fails instead of giving up youtube-dl continues trying different extractors.

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -2069,6 +2069,20 @@ class GenericIE(InfoExtractor):
             },
             'skip': 'TODO: fix nested playlists processing in tests',
         },
+        {
+            # Test for fallback after _parse_jwplayer_data fails
+            # See https://github.com/rg3/youtube-dl/pull/16735
+            'url': 'http://www.anime1.com/watch/nichijou/episode-1',
+            'info_dict': {
+                'id': '[HorribleSubs] Nichijou - 01 [480p]_a1_BRK',
+                'title': 'Nichijou - Episode 1',
+                'ext': 'mp4',
+                'uploader': 'www.anime1.com',
+            },
+            'params': {
+                'skip_download': True,
+            }
+        },
         # {
         #     # TODO: find another test
         #     # http://schema.org/VideoObject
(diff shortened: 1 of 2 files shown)


youtube_dl/utils.py
```diff
@@ -748,6 +748,7 @@ def __init__(self, msg, tb=None, expected=False, cause=None, video_id=None):
             msg = video_id + ': ' + msg
         if cause:
             msg += ' (caused by %r)' % cause
+        self.msg_without_bug_report = msg
```
dstftw: No.

youtube_dl/extractor/generic.py
```diff
@@ -3119,9 +3119,13 @@ def _real_extract(self, url):
...
             webpage, video_id, transform_source=js_to_json)
         if jwplayer_data:
-            info = self._parse_jwplayer_data(
-                jwplayer_data, video_id, require_title=False, base_url=url)
-            return merge_dicts(info, info_dict)
+            try:
+                info = self._parse_jwplayer_data(
+                    jwplayer_data, video_id, require_title=False, base_url=url)
+                return merge_dicts(info, info_dict)
+            except ExtractorError, e:
```
dstftw: Invalid syntax.

youtube_dl/extractor/generic.py
```diff
@@ -3119,9 +3119,13 @@ def _real_extract(self, url):
...
+                self.to_screen(e.msg_without_bug_report)
```
dstftw: No.

youtube_dl/extractor/generic.py
```diff
@@ -3119,9 +3119,13 @@ def _real_extract(self, url):
...
+                self.to_screen("Trying different extractor")
```
dstftw: No.

moyamo: I'm not sure what's wrong with this message. Should I just remove them.

moyamo: Feedback would be appreciated.

dstftw: There should not be any messages.

PR #16939, opened 2018-07-09
[ADN] Pass the key using --video-password

Kcchouette:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

It's to pass the daily changing key using --video-password, instead of self-rebuilding the software.

Thank you


Kcchouette:
@remitamine @dstftw Can you merge it for next version please?

Kcchouette:
ping @remitamine @dstftw

remitamine:
i simply don't think that using an option for something different(subtitle decryption key) than what is actually meant by the option(password protected videos) is not a good idea, it just adds more confusion.

another way, that might improve the situation is use the API as an alternative way to get the subtitles, but it looks like that the use of the API requires a premium account even for videos that are freely available in the desktop interface.

PR #17059, opened 2018-07-25
Update to Firefox plugin

funky-monkey:
The mentioned Firefox plugin has not been updated in a while - added a new plugin which has a valid export function that can be used for `youtube-dl`


funky-monkey:
👍

Issue #17051, opened 2018-07-23
max-filesize and match-filter not followed for certain streamed videos


amishxda:
### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2018.07.21*. If it's not, read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2018.07.21**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
- [x] Checked that provided video/audio/playlist URLs (if any) are alive and playable in a browser

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### Verbose output as follows:

```
/usr/bin/youtube-dl -v --max-filesize 101m --match-filter '!is_live & !live & duration <= 960' -i --no-warnings --no-mark-watched 'https://www.youtube.com/watch?v=pbk3RR0uF0c'
[debug] System config: []
[debug] User config: ['--format', 'best[height<=480]/bestvideo[height<=480]+bestaudio/best[height<=720]/bestvideo[height<=720]+bestaudio/best', '--merge-output-format', 'mp4', '--restrict-filenames', '--no-mtime', '--console-title', '--write-sub', '--proxy', '127.0.0.1:8080', '--no-check-certificate', '--no-call-home']
[debug] Custom config: []
[debug] Command-line args: ['-v', '--max-filesize', '101m', '--match-filter', '!is_live & !live & duration <= 960', '-i', '--no-warnings', '--no-mark-watched', 'https://www.youtube.com/watch?v=pbk3RR0uF0c']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2018.07.21
[...]

PR #17073, opened 2018-07-27
Add support for minimum and maximum duration

amishxda:

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

This adds the option of minimum and maximum duration for a video. (Similar to min and max filesize)

Often filesize is not an indicator of video duration but video selection by exact duration may be what is desired.

There are multiple reason why I am adding this:

1. For my youtube subscription - I see small videos (< 5 mins) directly on youtube without downloading. So downloading them makes no sense. (hence --min-duration)
2. Most of the times I am not interested in long duration videos (> 45 mins). Hence --max-duration.
3. See issue #17051. If this feature is passed - we could also add "-fs" (max filesize) and "-t" (max duration) option to ffmpeg processor. So that "recently" streamed video which is "no more live" but still under process by youtube (hence does not have file size and duration available) will not run away downloading 2-3 hour video and occupy GBs of space. In short --max-filesize and --max-duration would be respected even by ffmpeg.

Please consider.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,10 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
     --reject-title REGEX             Skip download for matching titles (regex or
                                      caseless sub-string)
     --max-downloads NUMBER           Abort after downloading NUMBER files
+    --min-duration SECONDS           Do not download any videos smaller than
+                                     duration, in seconds
+    --max-duration SECONDS           Do not download any videos larger than
+                                     duration, in seconds
     --min-filesize SIZE              Do not download any videos smaller than
                                      SIZE (e.g. 50k or 44.6m)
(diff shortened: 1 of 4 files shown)


dstftw:
`--match-filter FILTER`.

amishxda:
I know about the match filter. But as I said - match filter does not work when video gets downloaded via for ffmpeg.

For example --max-filesize is not passed to ffmpeg (it should be passed via -fs option to ffmpeg)

Similarly I can also pass -t option to ffmpeg if --max-duration option is supported.

Hence there is need for additional argument.

dstftw:
As already responded in another PR, `ffmpeg`'s `-fs` has nothing to do with `--max-filesize` option. What you are talking about is a different option that should have different name and should be implemented across all downloaders.

PR #17074, opened 2018-07-27
ffmpeg should respect --max-filesize too

amishxda:

---

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Currently --max-filesize means that do not download any videos larger than SIZE

Which in most cases implies that one does not want output file to grow larger than filesize.

But if my "subscriptions" has a live video OR "under-process" video where size is not known yet then youtube-dl keeps on downloading video forever (till streaming stops or end of the video is reached).

I believe that when --max-filesize is supplied, we must supply "-fs" option to ffmpeg so that output file does not go out of bounds.

This patch fixes that.

Please consider.

Thank you

### DIFF
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -296,8 +296,11 @@ class FFmpegFD(ExternalFD):
 
         args += ['-i', url, '-c', 'copy']
 
+        max_filesize = self.params.get('max_filesize')
         if self.params.get('test', False):
             args += ['-fs', compat_str(self._TEST_FILE_SIZE)]
+        elif max_filesize is not None:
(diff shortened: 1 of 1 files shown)


dstftw:
>Which in most cases implies that one does not want output file to grow larger than filesize.

Absolutely does not imply that. I'd expect it behave exactly as it's documented and not having half-downloaded cropped files.

What are you talking about is completely different option that should be implemented in downloaders separately.

As for live videos you can easily skip them with `--match-filter`.

amishxda:
What if I do not want to skip live video but I just want first 100MB say for promoting purpose?

PR #17075, opened 2018-07-27
Better min/max-filesize checking

amishxda:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

When using --min-filesize OR --max-filesize, youtube-dl first invokes downloader (i.e sends http download request) and then checks "Content-length".

This may not be necessary always. In most cases we already know filesize from meta information.

This PR improves the filesize checking.

If "filesize" is available via info_dict then it uses that and avoids invoking downloader unnecessarily.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -299,7 +299,7 @@ class YoutubeDL(object):
 
     The following parameters are not used by YoutubeDL itself, they are used by
     the downloader (see youtube_dl/downloader/common.py):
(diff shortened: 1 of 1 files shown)


dstftw:
This is the whole point of these options: to use **actual filesize** reported by HTTP header rather than metadata that may be incorrect.
If your fine with `filesize` metadata based check you use `--match-filter` that may not be 100% correct. If not you use `--max-filesize` that uses exact filesize but only works for plain HTTP downloads.

PR #17118, opened 2018-07-31
Close socket after downloading page.

bato3:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Close socket after downloading page.
On windows, when I make many requests it gives me error: **Python urllib2. URLError: <urlopen error [Errno 10048] Only one usage of each socket address (protocol/network address/port) is normally permitted>**
according to: https://stackoverflow.com/questions/4526933/python-urllib2-urlerror-urlopen-error-errno-10048-only-one-usage-of-each-so

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -769,6 +769,7 @@ class InfoExtractor(object):
             return res
         else:
             content, _ = res
+            _.close()
             return content
 
     def _download_xml_handle(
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/common.py
```diff
@@ -807,7 +808,12 @@ def _download_xml(
             transform_source=transform_source, fatal=fatal, encoding=encoding,
             data=data, headers=headers, query=query,
             expected_status=expected_status)
-        return res if res is False else res[0]
+        if res is False:
+            return res
+        else:
+            content, _ = res
```
dstftw: `_` is a throwaway name for a variable in python and should not be accessed.

bato3:
@dstftw Re: cant-reporoduce
People on Stackoverflow have this problem (as I looked, it was windows + python 2) And `close` is right solution

PR #17327, opened 2018-08-25
[olympicchannel] Add extractor

michaelforney:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This pull request adds an extractor for videos from https://www.olympicchannel.com. All of the necessary metadata is present in a `data-widget-data` JSON property, so I used that as the primary source, falling back to `_og_search_*`.


youtube_dl/extractor/olympicchannel.py
```diff
@@ -0,0 +1,50 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import parse_duration, unified_strdate
+
+
+class OlympicChannelIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?olympicchannel\.com/../playback/(?P<id>.*)'
```
dstftw: Matching empty id is senseless.

youtube_dl/extractor/olympicchannel.py
```diff
@@ -0,0 +1,50 @@
...
+            'release_date': '20160601',
+            'thumbnail': 'https://images.olympicchannel.com/media/images/stills/content/56721/a929eb66b550de98a453ef4d87c7856c.jpg'
+        }
+    }
+
+    def _real_extract(self, url):
+        display_id = self._match_id(url)
+        webpage = self._download_webpage(url, display_id)
+        data = self._parse_json(
+            self._html_search_regex(r'data-widget-data=["\']([^"\']*)["\']', webpage, 'data'), display_id)
```
dstftw: Matching empty data is senseless.

youtube_dl/extractor/olympicchannel.py
```diff
@@ -0,0 +1,50 @@
...
+    _TEST = {
+        'url': 'https://www.olympicchannel.com/en/playback/snowboard-mens-halfpipe',
+        'info_dict': {
+            'id': 'epis_63b711050c9749ab8ca3a6fa07f4cff6',
+            'ext': 'mp4',
+            'title': 'Men\'s Halfpipe Finals - Snowboard | PyeongChang 2018 Replays',
+            'description': 'The men\'s halfpipe finals were held at the Phoenix Snow Park on 14 February 2018.',
+            'duration': 6780,
+            'release_date': '20160601',
+            'thumbnail': 'https://images.olympicchannel.com/media/images/stills/content/56721/a929eb66b550de98a453ef4d87c7856c.jpg'
```
dstftw: No direct URLs here.

youtube_dl/extractor/olympicchannel.py
```diff
@@ -0,0 +1,50 @@
...
+    def _real_extract(self, url):
+        display_id = self._match_id(url)
+        webpage = self._download_webpage(url, display_id)
+        data = self._parse_json(
+            self._html_search_regex(r'data-widget-data=["\']([^"\']*)["\']', webpage, 'data'), display_id)
+        video_id = data['uid']
+        title = data.get('title')
+        description = data.get('synopsis')
+        thumbnail = data.get('images', {}).get('main', {}).get('url')
+        formats = None
```
dstftw: Must not be `None`.

youtube_dl/extractor/olympicchannel.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def _real_extract(self, url):
+        display_id = self._match_id(url)
+        webpage = self._download_webpage(url, display_id)
+        data = self._parse_json(
+            self._html_search_regex(r'data-widget-data=["\']([^"\']*)["\']', webpage, 'data'), display_id)
+        video_id = data['uid']
+        title = data.get('title')
+        description = data.get('synopsis')
+        thumbnail = data.get('images', {}).get('main', {}).get('url')
```
dstftw: `try_get`.

youtube_dl/extractor/olympicchannel.py
```diff
@@ -0,0 +1,50 @@
...
+            m3u8_url = ovps.get('stream_url') or ovps.get('streaming_url')
+            if m3u8_url:
+                formats = self._extract_m3u8_formats(m3u8_url, video_id, 'mp4')
+                self._sort_formats(formats)
+                break
+
+        return {
+            'id': video_id,
+            'display_id': display_id,
+            'title': title or self._og_search_title(webpage),
```
dstftw: Move to initial title assignment.

youtube_dl/extractor/olympicchannel.py
```diff
@@ -0,0 +1,50 @@
...
+            'formats': formats,
+            'description': description or self._og_search_description(webpage),
+            'duration': parse_duration(data.get('duration')),
```
dstftw: >"duration_in_seconds":6801,

youtube_dl/extractor/olympicchannel.py
michaelforney: Done

youtube_dl/extractor/olympicchannel.py
michaelforney: Done

youtube_dl/extractor/olympicchannel.py
michaelforney: Done

youtube_dl/extractor/olympicchannel.py
michaelforney: I figured it would be caught as an error later on, but an explicit error message is probably better so I added one.

youtube_dl/extractor/olympicchannel.py
michaelforney: Done

youtube_dl/extractor/olympicchannel.py
michaelforney: Done

youtube_dl/extractor/olympicchannel.py
michaelforney: Thanks, missed that.

youtube_dl/extractor/olympicchannel.py
```diff
@@ -0,0 +1,55 @@
...
+        display_id = self._match_id(url)
+        webpage = self._download_webpage(url, display_id)
+        data = self._parse_json(
+            self._html_search_regex(r'data-widget-data=["\']([^"\']+)["\']', webpage, 'data'), display_id)
+        item = data['items'][0]
+
+        video_id = data['uid']
+        title = data.get('title') or self._og_search_title(webpage)
+        description = data.get('synopsis') or self._og_search_description(webpage)
+        duration = item.get('duration_in_seconds')
```
dstftw: `int_or_none`.

youtube_dl/extractor/olympicchannel.py
```diff
@@ -0,0 +1,55 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import ExtractorError, try_get, unified_strdate
+
+
+class OlympicChannelIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?olympicchannel\.com/../playback/(?P<id>.+)'
```
dstftw: Should only capture path part and should not capture query or fragment parts of URL.

dstftw: Also looks like they redesigned the site so that extractor does not work any longer.

youtube_dl/extractor/olympicchannel.py
michaelforney: After redesign, there is no more duration.

youtube_dl/extractor/olympicchannel.py
michaelforney: Done

michaelforney: Updated for new layout. There is still a `<div id='ocs-video' data-options='...'>`, but it no longer contains all the metadata (missing description). However, they added new meta fields with everything we need, so I went with those instead which simplified things a bit.

michaelforney:
Updated for the new layout.

One thing I was not sure about was the thumbnail regex for the `_TEST`. It looks like the image URLs have no extension, so I just used `r're:^https?://.*'`. Is this okay, or is something else preferred?

michaelforney:
Ping. Anything else I need to do here? The extractor still seems to work with the current site.

PR #17414, opened 2018-09-01
Added info extractor for www.tele5.de, a German TV station.

jensrutschmann:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This adds support for www.tele5.de, which is a German TV station providing temporary online access to movies and TV shows after broadcast.

I have added tests, but since the referenced videos will only be available for limited time, these tests will fail in the future. They all were successful at the time this PR has been created, though. Unfortunately the site does not provide videos that are available for a longer time.

### DIFF
--- a/youtube_dl/extractor/tele5.py
+++ b/youtube_dl/extractor/tele5.py
@@ -1,0 +1,58 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from .nexx import NexxIE
+
+
+class Tele5IE(InfoExtractor):
+    _VALID_URL = r'https://www.tele5.de/(mediathek/filme-online/videos|tv/).*'
+
+    _TESTS = [{
+        'url': 'https://www.tele5.de/mediathek/filme-online/videos?vid=1550589',
+        'info_dict': {
+            'id': '1550589',
+            'ext': 'mp4',
+            'upload_date': '20180822',
+            'timestamp': 1534927316,
+            'title': 'SchleFaZ: Atomic Shark'
+        }
+    }, {
+        'url': 'https://www.tele5.de/tv/dark-matter/videos',
+        'info_dict': {
+            'id': '1548206',
+            'ext': 'mp4',
+            'title': 'Folge Sechsundzwanzig',
+            'timestamp': 1533664358,
+            'upload_date': '20180807'
+        }
+    }, {
+        'url': 'https://www.tele5.de/tv/relic-hunter/videos',
+        'info_dict': {
+            'id': '1548034',
+            'ext': 'mp4',
+            'timestamp': 1533577964,
+            'upload_date': '20180806',
+            'title': 'Mr. Right'
+        }
+    }, {
+        'url': 'https://www.tele5.de/tv/buffy-im-bann-der-daemonen/videos',
+        'info_dict': {
+            'id': '1547129',
+            'ext': 'mp4',
+            'upload_date': '20180730',
+            'timestamp': 1532967491,
+            'title': 'Der Höllenhund'
+        }
+    }]
+
+    def _real_extract(self, url):
+        webpage = self._download_webpage(url, 'N/A')
+
+        id = self._html_search_regex(
+            r'class="ce_videoelementnexx-video__player"\sid="video-player"\sdata-id="(?P<id>[0-9]+)"',
+            webpage, 'id')
+
+        return self.url_result(
+            'https://api.nexx.cloud/v3/759/videos/byid/%s'
+            % id, ie=NexxIE.ie_key())
--- a/youtube_dl/extractor/tele5.py
+++ b/youtube_dl/extractor/tele5.py
@@ -1,12 +1,14 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
+import re
+
 from .common import InfoExtractor
 from .nexx import NexxIE
 
 
 class Tele5IE(InfoExtractor):
-    _VALID_URL = r'https://www.tele5.de/[mediathek/filme-online/videos|tv/]'
+    _VALID_URL = r'https://www\.tele5\.de/(?:mediathek/filme-online/videos\?vid=|tv/)(?P<display_id>[\w-]+)'
 
     _TESTS = [{
         'url': 'https://www.tele5.de/mediathek/filme-online/videos?vid=1550589',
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/tele5.py
```diff
@@ -0,0 +1,58 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from .nexx import NexxIE
+
+
+class Tele5IE(InfoExtractor):
+    _VALID_URL = r'https://www.tele5.de/(mediathek/filme-online/videos|tv/).*'
```
dstftw: 1. `.*` at the end is senseless.
2. Don't capture groups you don't use.
3. Capture and use display id.

youtube_dl/extractor/tele5.py
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }, {
+        'url': 'https://www.tele5.de/tv/buffy-im-bann-der-daemonen/videos',
+        'info_dict': {
+            'id': '1547129',
+            'ext': 'mp4',
+            'upload_date': '20180730',
+            'timestamp': 1532967491,
+            'title': 'Der Höllenhund'
+        }
```
dstftw: Remove all duplicate tests.

youtube_dl/extractor/tele5.py
```diff
@@ -0,0 +1,58 @@
...
+    }]
+
+    def _real_extract(self, url):
+        webpage = self._download_webpage(url, 'N/A')
+
+        id = self._html_search_regex(
+            r'class="ce_videoelementnexx-video__player"\sid="video-player"\sdata-id="(?P<id>[0-9]+)"',
```
dstftw: Should be more relaxed.

youtube_dl/extractor/tele5.py
```diff
@@ -0,0 +1,58 @@
...
+            'upload_date': '20180730',
+            'timestamp': 1532967491,
+            'title': 'Der Höllenhund'
+        }
+    }]
+
+    def _real_extract(self, url):
+        webpage = self._download_webpage(url, 'N/A')
+
+        id = self._html_search_regex(
```
dstftw: Don't shadow built-in names.

youtube_dl/extractor/tele5.py
```diff
@@ -0,0 +1,58 @@
...
+            r'class="ce_videoelementnexx-video__player"\sid="video-player"\sdata-id="(?P<id>[0-9]+)"',
+            webpage, 'id')
+
+        return self.url_result(
+            'https://api.nexx.cloud/v3/759/videos/byid/%s'
+            % id, ie=NexxIE.ie_key())
```
dstftw: Pass video id.

jensrutschmann:
@dstftw I've implemented the requested changes - sorry for the beginner bugs - I've never done python programming before.

The only requested change I've not implemented is 'Capture and use display id.' because not all URLs contain the actual video ID.
For example the URL from the second test always points to the current (and only available) episode of a series, so the actual video reachable via this URL will change - for example each week, when a new episode is made available.

dstftw:
I'm not talking about video id, capture part of URL as display id.

youtube_dl/extractor/tele5.py
```diff
@@ -0,0 +1,40 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from .nexx import NexxIE
+
+
+class Tele5IE(InfoExtractor):
+    _VALID_URL = r'https://www.tele5.de/[mediathek/filme-online/videos|tv/]'
```
dstftw: Incorrect.

youtube_dl/extractor/tele5.py
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }, {
+        'url': 'https://www.tele5.de/tv/dark-matter/videos',
+        'info_dict': {
+            'id': '1548206',
+            'ext': 'mp4',
+            'title': 'Folge Sechsundzwanzig',
+            'timestamp': 1533664358,
+            'upload_date': '20180807'
+        }
```
dstftw: I've already pointed out: **remove all duplicate tests**. Or make them only_matching.

youtube_dl/extractor/tele5.py
```diff
@@ -0,0 +1,40 @@
...
+    }]
+
+    def _real_extract(self, url):
+        webpage = self._download_webpage(url, 'N/A')
+
+        video_id = self._html_search_regex(
+            r'id="video-player"\sdata-id="(?P<id>[0-9]+)"',
```
dstftw: Must allow arbitrary whitespace not only `\s`. Must allow single quotes. Group name is senseless when there is only one group.

youtube_dl/extractor/tele5.py
```diff
@@ -0,0 +1,40 @@
...
+            webpage, 'id')
+
+        return self.url_result(
+            'https://api.nexx.cloud/v3/759/videos/byid/%s'
+            % video_id, ie=NexxIE.ie_key())
```
dstftw: My comment here was ignored.

youtube_dl/extractor/tele5.py
jensrutschmann: Why is this a duplicate test although it tests another URL pattern?

dstftw: Because all of them use the same extraction scenario.

youtube_dl/extractor/tele5.py
dstftw: Read some doc on regular expressions. Namely what does `[]` and `()` mean and how it should be used.

jensrutschmann:
@dstftw I think I've implemented all requested changes.
Let me know if there's anything missing or if I should squash the commits before merging.

mrtnmtth:
Here is my take on it:  
https://github.com/mrtnmtth/youtube-dl/blob/master/youtube_dl/extractor/tele5.py

My regex is more relaxed by making `www` optional and allowing `http` without `s`.  
Also I only download parse the html if there is no id in the url.

Feel free to copy parts of it as you like.

youtube_dl/extractor/tele5.py
```diff
@@ -0,0 +1,39 @@
...
+            'title': 'SchleFaZ: Atomic Shark',
+        }
+    }, {
+        'url': 'https://www.tele5.de/tv/dark-matter/videos',
+        'only_matching': True,
+    }]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        display_id = mobj.group('display_id')
```
dstftw: `self._match_id`.

youtube_dl/extractor/tele5.py
```diff
@@ -0,0 +1,39 @@
...
+
+        webpage = self._download_webpage(url, display_id)
+
+        video_id = self._html_search_regex(
```
dstftw: `_search_regex`.

youtube_dl/extractor/tele5.py
```diff
@@ -0,0 +1,39 @@
...
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from .nexx import NexxIE
+
+
+class Tele5IE(InfoExtractor):
+    _VALID_URL = r'https://www\.tele5\.de/(?:mediathek/filme-online/videos\?vid=|tv/)(?P<display_id>[\w-]+)'
```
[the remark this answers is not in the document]
mrtnmtth: My take is:  
`_VALID_URL = r'https?://(?:www\.)?tele5\.de/(?:mediathek/filme-online/videos(/|\?(.*&)?vid=)|tv/)(?P<id>[\w-]+)'`

It matches all of the following:
```
https://www.tele5.de/mediathek/filme-online/videos?blah=1&vid=1550589
https://www.tele5.de/mediathek/filme-online/videos/schlefaz-atomic-shark
https://www.tele5.de/mediathek/filme-online/videos?vid=1549415
https://www.tele5.de/tv/digimon/videos
https://www.tele5.de/tv/digimon/videos/rikas-krise
https://www.tele5.de/tv/kalkofes-mattscheibe/video-clips/politik-und-gesellschaft?ve_id=1551191
```

PR #17498, opened 2018-09-08
[Instagram] Support for Stories videos

Tatsh:

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

This pull request adds support for individual Instagram story URLs such as `https://www.instagram.com/stories/charmedplayhouses/`. Also supports individual highlighted stories found on user pages.

Every request must have cookies specified (`--cookies`).

I have not decided on the best title format. For now it is *Stories - user_name*.

### DIFF
--- a/youtube_dl/extractor/instagram.py
+++ b/youtube_dl/extractor/instagram.py
@@ -9,6 +9,7 @@ from .common import InfoExtractor
 from ..compat import (
     compat_str,
     compat_HTTPError,
+    compat_urlparse,
 )
 from ..utils import (
     ExtractorError,
@@ -259,7 +260,7 @@ class InstagramPlaylistIE(InfoExtractor):
                 'first': 12,
                 'after': cursor,
             }
-            variables.update(self._query_vars_for(data))
+            variables.update(self._query_vars_for(data, url))
             variables = json.dumps(variables)
 
             if self._gis_tmpl:
@@ -274,6 +275,10 @@ class InstagramPlaylistIE(InfoExtractor):
 
             # try all of the ways to generate a GIS query, and not only use the
             # first one that works, but cache it for future requests
+            if hasattr(self, '_get_query_hash'):
+                query_hash = self._get_query_hash(url)
+            else:
+                query_hash = self._QUERY_HASH
             for gis_tmpl in gis_tmpls:
                 try:
                     json_data = self._download_json(
@@ -283,10 +288,10 @@ class InstagramPlaylistIE(InfoExtractor):
                             'X-Instagram-GIS': hashlib.md5(
                                 ('%s:%s' % (gis_tmpl, variables)).encode('utf-8')).hexdigest(),
                         }, query={
-                            'query_hash': self._QUERY_HASH,
+                            'query_hash': query_hash,
                             'variables': variables,
                         })
-                    media = self._parse_timeline_from(json_data)
+                    media = self._parse_timeline_from(json_data, url)
                     self._gis_tmpl = gis_tmpl
                     break
                 except ExtractorError as e:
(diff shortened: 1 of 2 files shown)


tordans:
It would be awesome to have this!

youtube_dl/extractor/instagram.py
```diff
@@ -359,3 +360,132 @@ def _real_extract(self, url):
...
+                                                    highlight_reel_ids=[reel_id],
+                                                    precomposed_overlay=False)))
+            data = self._download_json(query_url, reel_id, query=params)
+            reels_media = data.get('data', {'reels_media': []}).get('reels_media')
+
+            return self._entries(reels_media, username, title=title)
+
+        webpage = self._download_webpage('https://www.instagram.com',
+                                         username_or_id)
+        url = self._search_regex(r'<link .*href="(/graphql/query/.*only_stories[^"]+)"',
```
dstftw: 1. Do not shadow `url`.
2. Regex should not match across several tags.

youtube_dl/extractor/instagram.py
```diff
@@ -359,3 +360,132 @@ def _real_extract(self, url):
...
+                        width=res['config_width'],
+                        height=res['config_height'],
+                        format_note=res['mime_type'],
+                    ))
+
+                entries.append(dict(thumbnails=thumbnails,
+                                    thumbnail=item['display_url'],
+                                    id=item['id'],
+                                    title=title,
+                                    uploader=item['owner']['username'],
```
dstftw: Read coding conventions on optional meta fields. Fix all such meta fields added in this PR.

youtube_dl/extractor/instagram.py
```diff
@@ -359,3 +360,132 @@ def _real_extract(self, url):
...
+        if '/highlights/' in url:
+            webpage = self._download_webpage(url, username_or_id)
+            data = self._parse_json(
+                self._search_regex(
+                    r'sharedData\s*=\s*({.+?})\s*;\s*[<\n]', webpage, 'data'),
+                username_or_id)
+            uploader_id = data['entry_data']['StoriesPage'][0]['user']['id']
+            username = data['entry_data']['StoriesPage'][0]['user']['username']
+            reel_id = username_or_id
+            params = dict(query_hash='7c16654f22c819fb63d1183034a5162f',
```
dstftw: Use `{}` dicts instead.

dstftw:
Currently merging this breaks tag and user extractors.

Tatsh:
@dstftw requesting review again. This is a newly made branch and it uses the GraphQL code already in master so there's no duplicate code for that. Also doesn't break tags or user extractors. Tested all scenarios on my machine (using `--cookies` where necessary).

dstftw:
Tests are broken.

Tatsh:
@dstftw Please verify my fix. I made the tests get skipped because they require cookies. I did test both URLs in the tests locally with the `--cookies ...` argument.

dstftw:
Stories don't seem to work:
```
> py -3.7 -m youtube_dl https://www.instagram.com/stories/jillianbmele/ -v --cookies '...'
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['https://www.instagram.com/stories/jillianbmele/', '-v', '--cookies', '...']
[debug] Encodings: locale cp1251, fs utf-8, out utf-8, pref cp1251
[debug] youtube-dl version 2020.03.24
[debug] Git HEAD: 0534ab6
[debug] Python version 3.7.0 (CPython) - Windows-10-10.0.10240-SP0
[debug] exe versions: ffmpeg N-85653-gb4330a0, ffprobe N-85653-gb4330a0, phantomjs 2.1.1, rtmpdump 2.4
[debug] Proxy map: {}
[instagram:stories] jillianbmele: Downloading webpage
[download] Downloading playlist: jillianbmele
[instagram:stories] jillianbmele: Downloading JSON page 1
[instagram:stories] playlist jillianbmele: Downloading 0 videos
[download] Finished downloading playlist: jillianbmele
```

Tatsh:
Did you try the highlights URL?

I got this same response for this link after once downloading successfully at least one video. I feel it has something to do with whether or not you have viewed the video in your browser since Instagram has to have a way to know that you already viewed a story for its indicator. I'll look into this because there should still be a way even if you have viewed the story before.

Tatsh:
Also it's always possible a user has no video story objects up in which case finding zero videos is the correct output. The reason I picked this user is because she tends to post a lot of videos in her story timeline so it's good for testing.

dstftw:
Highlights URL does work. Both URLs work from browser though.

Tatsh:
This should fix it. Had to request the video data based on the reel ID. This is much more reliable than attempting to read from the first request.

Tatsh:
@dstftw bump

Tatsh:
The main stories URI can take an identifier after /stories/ and it appears this points to a graph object, which could be a video. So I have one more type of URI to support.

In the Instagram app these are generated looking at a story then tapping the ... button and tapping 'Copy link'.

Tatsh:
The type of URL mentioned in the last comment is now supported. @dstftw I think I am truly done here let's hope.

doonline-official:
can we please get this merged?.... stories would be so nice to have...

Tatsh:
@dstftw If you want me to make changes, I'll need to re-make this PR. I asked GitHub and they cannot fix branch references.

PR #17619, opened 2018-09-20
ABC: add support for retrieving the URL of the latest episode on a show page

pabs3:


dstftw:
1. Bother to read and fill PR template.
2. Clearly explain what it's supposed to fix.
3. Provide example URLs and add tests.

pabs3:
Sorry, about the template, I guess the command-line client I use doesn't support templates.

PR #17625, opened 2018-09-21
liveleak: add original format without watermark

jomo:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

LiveLeak video filenames commonly look somewhat like `something.mp4.zxcvbn.mp4`, which has a watermark and lower quality compared to `something.mp4`.

This PR adds an additional format that uses the original filename where possible.

### DIFF
--- a/youtube_dl/extractor/liveleak.py
+++ b/youtube_dl/extractor/liveleak.py
@@ -120,7 +120,17 @@ class LiveLeakIE(InfoExtractor):
             }
 
         for idx, info_dict in enumerate(entries):
+            source_found = False
             for a_format in info_dict['formats']:
+                if not source_found:
+                    source_url = re.sub(r'(https?://(?:\w+\.)?liveleak\.com/.*?\.\w+)\.\w+\.mp4((?:\?.+)?)', r'\1\2', a_format['url'])
+                    if source_url != a_format['url']:
+                        source_found = True
+                        info_dict['formats'].append({
+                            'url': source_url,
+                            'format_note': 'original, without watermark',
+                            'format_id': 'source'
+                        })
                 if not a_format.get('height'):
                     a_format['height'] = int_or_none(self._search_regex(
                         r'([0-9]+)p\.mp4', a_format['url'], 'height label',


youtube_dl/extractor/liveleak.py
```diff
@@ -120,6 +120,19 @@ def _real_extract(self, url):
             }
 
         for idx, info_dict in enumerate(entries):
+
+            for f in info_dict['formats']:
```
dstftw: There is already a for loop below you must integrate into.

youtube_dl/extractor/liveleak.py
```diff
@@ -120,6 +120,19 @@ def _real_extract(self, url):
...
+                match = re.match(r'https?://(?:\w+\.)?liveleak\.com/.*?\.(\w+)(\.\w+\.mp4)(?:\?.+)?', f['url'])
```
dstftw: `re.sub`.

youtube_dl/extractor/liveleak.py
```diff
@@ -120,6 +120,19 @@ def _real_extract(self, url):
...
+                if match:
+                    ext = match.group(1)
+                    new_url = f['url'].replace(match.group(2), '')
+                    info_dict['formats'].append({
+                        'url': new_url,
+                        'ext': ext,
+                        'format_note': 'original, without watermark',
+                        'format_id': '%s-orig' % (f['format_id'] or 'default')
+                    })
```
dstftw: 1. This will duplicate source format when multiple non source formats are available.
2. Source format must be named `source`.

youtube_dl/extractor/liveleak.py
jomo: I'm using results of the match groups here, how would I do this with `re.sub`?

dstftw: By read how `re.sub` works in the first place?

jomo: I did read into it, but as I don't want to replace the first group (which is the original file extension), I don't see how `re.sub` would give me the replaced string *and* the content of the group. Perhaps I'm overlooking it?

dstftw: Don't replace it then. What's the problem?

jomo: Would you prefer to use `re.sub` for the replacement and then `re.match` to grab the extension?

dstftw: No. You don't need to grab any extension in the first place.

jomo: Isn't that what the `ext` is meant for?

dstftw: Again: you should not bother with that. It will be automatically extracted from the download URL.

jomo: Ah, that wasn't clear to me.

jomo:
I think I have addressed all your change requests, @dstftw. Can you take a look?

youtube_dl/extractor/liveleak.py
```diff
@@ -120,7 +120,17 @@ def _real_extract(self, url):
             }
 
         for idx, info_dict in enumerate(entries):
+            source_found = False
             for a_format in info_dict['formats']:
+                if not source_found:
+                    source_url = re.sub(r'(https?://(?:\w+\.)?liveleak\.com/.*?\.\w+)\.\w+\.mp4((?:\?.+)?)', r'\1\2', a_format['url'])
+                    if source_url:
+                        source_found = True
```
dstftw: This is pointless. You already have `source_url` as marker.

youtube_dl/extractor/liveleak.py
```diff
@@ -120,7 +120,17 @@ def _real_extract(self, url):
             }
 
         for idx, info_dict in enumerate(entries):
+            source_found = False
             for a_format in info_dict['formats']:
+                if not source_found:
+                    source_url = re.sub(r'(https?://(?:\w+\.)?liveleak\.com/.*?\.\w+)\.\w+\.mp4((?:\?.+)?)', r'\1\2', a_format['url'])
```
dstftw: Just replace the part appended to non source formats with empty string.

jomo: Do you mean the URL parameters? They're required as some sort of temporary access key.

youtube_dl/extractor/liveleak.py
jomo: Oh, this should have been `if source_url != a_format['url']`. Fixed now.

jomo:
@dstftw anything else you want me to change?

PR #17823, opened 2018-10-07
Add support for oload.cc

yonaikerlol:

Add support for oload.cc, other of many domains of openload


wapkinglive:
yeah it will be great if you add openload

PR #17856, opened 2018-10-10
[MallTV] Add new extractor

schunka:

---

- [ ] Bug fix
- [x] Improvement
- [x] New extractor
- [ ] New feature

---

Extractor for new streaming service [Mall.tv](https://www.mall.tv). There was already a pull request for this extractor, but it was closed.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -1052,7 +1052,7 @@ class InfoExtractor(object):
     @staticmethod
     def _og_regexes(prop):
         content_re = r'content=(?:"([^"]+?)"|\'([^\']+?)\'|\s*([^\s"\'=<>`]+?))'
-        property_re = (r'(?:name|property)=(?:\'og:%(prop)s\'|"og:%(prop)s"|\s*og:%(prop)s\b)'
+        property_re = (r'(?:name|property)=(?:\'og[:-]%(prop)s\'|"og[:-]%(prop)s"|\s*og[:-]%(prop)s\b)'
                        % {'prop': re.escape(prop)})
         template = r'<meta[^>]+?%s[^>]+?%s'
         return [
--- a/youtube_dl/extractor/malltv.py
+++ b/youtube_dl/extractor/malltv.py
@@ -1,0 +1,67 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class MallTVIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?mall\.tv/(?:.+/)?(?P<id>.+)'
+    _TESTS = [
+        {
+            'url': ('https://www.mall.tv/18-miliard-pro-neziskovky'
+                    '-opravdu-jsou-sportovci-nebo-clovek-v-tisni-pijavice'),
+            'md5': '5235290504d20a27a19dd3915b1167b4',
+            'info_dict': {
+                'id': ('18-miliard-pro-neziskovky-opravdu-jsou-sportovci-nebo-'
+                       'clovek-v-tisni-pijavice'),
+                'ext': 'mp4',
+                'title': ('18 miliard pro neziskovky. Opravdu jsou sportovci '
+                          'nebo Člověk v tísni pijavice?'),
+                'description': ('Pokud někdo hospodaří s penězmi daňových '
+                                'poplatníků, pak logicky chceme vědět, jak s '
+                                'nimi nakládá. Objem dotací pro neziskovky '
+                                'roste, ale opravdu jsou tyto organizace '
+                                '„pijavice", jak o nich hovoří And')
+                }
+        },
+        {
+            'url': ('https://www.mall.tv/kdo-to-plati/18-miliard-pro-neziskovky'
+                    '-opravdu-jsou-sportovci-nebo-clovek-v-tisni-pijavice'),
+            'md5': '5235290504d20a27a19dd3915b1167b4',
+            'info_dict': {
+                'id': ('18-miliard-pro-neziskovky-opravdu-jsou-sportovci-nebo-'
+                       'clovek-v-tisni-pijavice'),
+                'ext': 'mp4',
+                'title': ('18 miliard pro neziskovky. Opravdu jsou sportovci '
+                          'nebo Člověk v tísni pijavice?'),
+                'description': ('Pokud někdo hospodaří s penězmi daňových '
+                                'poplatníků, pak logicky chceme vědět, jak s '
+                                'nimi nakládá. Objem dotací pro neziskovky '
+                                'roste, ale opravdu jsou tyto organizace '
+                                '„pijavice", jak o nich hovoří And')
+                }
+        },
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        self.report_extraction(video_id)
+
+        format_url = self._html_search_regex(
+            r'<source src=(.+index)\s',
+            webpage,
+            'm3u8 URL')
+        formats = self._extract_m3u8_formats(format_url+'.m3u8',
+                                             video_id, 'mp4')
+        self._sort_formats(formats)
+
+        return {
+            'id': video_id,
+            'title': self._og_search_title(webpage),
+            'thumbnail': self._og_search_thumbnail(webpage),
+            'description': self._og_search_description(webpage),
+            'title': self._og_search_title(webpage),
+            'formats': formats
+        }
--- a/youtube_dl/extractor/malltv.py
+++ b/youtube_dl/extractor/malltv.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 from .common import InfoExtractor
+from ..utils import parse_duration
 
 
 class MallTVIE(InfoExtractor):
@@ -22,7 +23,7 @@ class MallTVIE(InfoExtractor):
                                 'nimi nakládá. Objem dotací pro neziskovky '
                                 'roste, ale opravdu jsou tyto organizace '
                                 '„pijavice", jak o nich hovoří And')
-                }
+                },
         },
         {
             'url': ('https://www.mall.tv/kdo-to-plati/18-miliard-pro-neziskovky'
@@ -46,22 +47,29 @@ class MallTVIE(InfoExtractor):
     def _real_extract(self, url):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
-
-        self.report_extraction(video_id)
+        JSON_LD_RE = (r'(?is)<script[^>]+type=([\"\'])?application/ld\+json>.*'
+                      '(?P<json_ld>{.+}).*</script>')
+        json_ld = self._search_regex(JSON_LD_RE, webpage, 'JSON_LD',
+                                     group='json_ld')
+        if not json_ld:
+            info = {}
+        else:
+            info = self._json_ld(json_ld, video_id)
 
         format_url = self._html_search_regex(
-            r'<source src=(.+index)\s',
+            r'<source src=([\"\'])?(.+index)\s',
             webpage,
             'm3u8 URL')
         formats = self._extract_m3u8_formats(format_url+'.m3u8',
                                              video_id, 'mp4')
         self._sort_formats(formats)
-
+        title = info.get('title', self._og_search_title(webpage, fatal=False))
+        thumbnail = info.get('thumbnailUrl', self._og_search_thumbnail(webpage))
         return {
             'id': video_id,
-            'title': self._og_search_title(webpage),
-            'thumbnail': self._og_search_thumbnail(webpage),
+            'title': title,
+            'thumbnail': thumbnail,
             'description': self._og_search_description(webpage),
-            'title': self._og_search_title(webpage),
+            'duration': parse_duration(info.get('duration')),
             'formats': formats
         }
--- a/youtube_dl/extractor/malltv.py
+++ b/youtube_dl/extractor/malltv.py
@@ -1,14 +1,13 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
-import hashlib
-from .common import InfoExtractor
-from ..utils import parse_duration
 import re
+from .common import InfoExtractor
+from ..utils import parse_duration, merge_dicts
 
 
 class MallTVIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?mall\.tv/(?:.+/)?(?P<id>.+)'
+    _VALID_URL = r'https?://(?:www\.)?mall\.tv/(?:.+/)?(?P<id>.+)(?:\?.*$|$)'
     _TESTS = [
         {
             'url': 'https://www.mall.tv/18-miliard-pro-neziskovky-opravdu-jsou-sportovci-nebo-clovek-v-tisni-pijavice',
@@ -17,17 +16,22 @@ class MallTVIE(InfoExtractor):
                 'id': 't0zzt0',
                 'ext': 'mp4',
                 'title': '18 miliard pro neziskovky. Opravdu jsou sportovci nebo Člověk v tísni pijavice?',
-                'description': "Pokud někdo hospodaří s penězmi daňových poplatníků, pak logicky chceme vědět, jak s nimi nakládá. Objem dotací pro neziskovky roste, ale opravdu jsou tyto organizace „pijavice', jak o nich hovoří And"
+                'description': 'Pokud někdo hospodaří s penězmi daňových poplatníků, pak logicky chceme vědět, jak s nimi nakládá. Objem dotací pro neziskovky roste, ale opravdu jsou tyto organizace „pijavice", jak o nich hovoří And',
+                'upload_date': '20181007',
+                'timestamp': 1538870400
                 }
         },
         {
             'url': 'https://www.mall.tv/kdo-to-plati/18-miliard-pro-neziskovky-opravdu-jsou-sportovci-nebo-clovek-v-tisni-pijavice',
             'md5': '9ced0de056534410837077e23bfba796',
+            'only_matching': 1,
             'info_dict': {
                 'id': 't0zzt0',
                 'ext': 'mp4',
                 'title': '18 miliard pro neziskovky. Opravdu jsou sportovci nebo Člověk v tísni pijavice?',
-                'description': "Pokud někdo hospodaří s penězmi daňových poplatníků, pak logicky chceme vědět, jak s nimi nakládá. Objem dotací pro neziskovky roste, ale opravdu jsou tyto organizace „pijavice', jak o nich hovoří And"
+                'description': 'Pokud někdo hospodaří s penězmi daňových poplatníků, pak logicky chceme vědět, jak s nimi nakládá. Objem dotací pro neziskovky roste, ale opravdu jsou tyto organizace „pijavice", jak o nich hovoří And',
+                'upload_date': '20181007',
+                'timestamp': 1538870400
                 }
         }
     ]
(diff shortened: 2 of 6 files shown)


schunka:
this resolves #18058

dstftw:
Read coding conventions and fix code.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,65 @@
...
+
+from .common import InfoExtractor
+from ..utils import parse_duration
+
+
+class MallTVIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?mall\.tv/(?:.+/)?(?P<id>.+)'
+    _TESTS = [
+        {
+            'url': ('https://www.mall.tv/18-miliard-pro-neziskovky-opravdu-jsou-sportovci-nebo-clovek-v-tisni-pijavice'),
```
dstftw: Remove ().

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,65 @@
...
+            'info_dict': {
+                'id': ('18-miliard-pro-neziskovky-opravdu-jsou-sportovci-nebo-clovek-v-tisni-pijavice'),
+                'ext': 'mp4',
+                'title': ('18 miliard pro neziskovky. Opravdu jsou sportovci nebo Člověk v tísni pijavice?'),
+                'description': ('Pokud někdo hospodaří s penězmi daňových '
```
dstftw: md5.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,65 @@
...
+        webpage = self._download_webpage(url, video_id)
+        info = self._search_json_ld(webpage, video_id, default={})
+
+        format_url = self._html_search_regex(
+            r'<source src=([\"\'])?(?P<src>.+?index)\1?[^>]*?>',
+            webpage, 'm3u8 URL', group='src')
+        formats = self._extract_m3u8_formats(format_url+'.m3u8',
+                                             video_id, 'mp4')
+        self._sort_formats(formats)
+        title = info.get('title', self._og_search_title(webpage, fatal=False))
```
dstftw: Mandatory.

dstftw: Won't work for `info = {'title': None}`.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,65 @@
...
+        thumbnail = info.get('thumbnailUrl', self._og_search_thumbnail(webpage))
```
dstftw: Same.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,65 @@
...
+        return {
+            'id': video_id,
+            'title': title,
+            'thumbnail': thumbnail,
+            'description': self._og_search_description(webpage),
+            'duration': parse_duration(info.get('duration')),
```
dstftw: No `parse_duration` here.

youtube_dl/utils.py
```diff
@@ -183,7 +183,7 @@ def register_socks_protocols():
 ])
 
 PACKED_CODES_RE = r"}\('(.+)',(\d+),(\d+),'([^']+)'\.split\('\|'\)"
-JSON_LD_RE = r'(?is)<script[^>]+type=(["\'])application/ld\+json\1[^>]*>(?P<json_ld>.+?)</script>'
+JSON_LD_RE = r'(?is)<script[^>]+type=(["\'])?application/ld\+json\1?[^>]*>(?P<json_ld>.+?)</script>'
```
dstftw: Invalid. Either both quotes must be present or none. Not one of them.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,65 @@
...
+                'title': ('18 miliard pro neziskovky. Opravdu jsou sportovci nebo Člověk v tísni pijavice?'),
+                'description': ('Pokud někdo hospodaří s penězmi daňových '
+                                'poplatníků, pak logicky chceme vědět, jak s '
+                                'nimi nakládá. Objem dotací pro neziskovky '
+                                'roste, ale opravdu jsou tyto organizace '
+                                '„pijavice", jak o nich hovoří And')
+                },
+            'params': {
+                'skip_download': True
+            }
```
dstftw: Remove.

youtube_dl/extractor/malltv.py
schunka: This I find very complicated. I've originally got it here, but then after a while without any change on my side the tests failed due to different hash. So after some digging I fuond they changed a video (maybe a commercial or whatever).
What should be the procedure here then?

dstftw: Procedure for what? Once description changes test will fail regardless of whether it's an md5 or complete description.

schunka: Then I don't understand necessity of md5 other than try tu run the test and don't care for outcome.

dstftw: Removing useless noise.

schunka: done

youtube_dl/extractor/malltv.py
schunka: done

youtube_dl/extractor/malltv.py
schunka: solved

schunka: solved by `fatal=True`

youtube_dl/extractor/malltv.py
schunka: solved by puting thumbnail as "optional"

youtube_dl/utils.py
schunka: resolved for all three posibilities.

youtube_dl/extractor/malltv.py
schunka: done

youtube_dl/extractor/malltv.py
schunka: moved outside `return` and added only when present.

schunka:
How this works?
You point out something, I fix/change it and then wait for month or more for someone to review the changes. And then another round, until everybody is satisfied?
What bothers me is the long time between replies.

dstftw:
1. We do not track every commit in every PR. It's technically not feasible. Once you think you're done you should leave the message.
2. Not all issues I've mentioned are fixed.

youtube_dl/utils.py
```diff
@@ -183,7 +183,7 @@ def register_socks_protocols():
 ])
 
 PACKED_CODES_RE = r"}\('(.+)',(\d+),(\d+),'([^']+)'\.split\('\|'\)"
-JSON_LD_RE = r'(?is)<script[^>]+type=(["\'])application/ld\+json\1[^>]*>(?P<json_ld>.+?)</script>'
+JSON_LD_RE = r'(?is)<script[^>]+type=(["\']|)application/ld\+json\1[^>]*>(?P<json_ld>.+?)</script>'
```
dstftw: `r'(?is)<script[^>]+type=(["\']?)application/ld\+json\1[^>]*>(?P<json_ld>.+?)</script>'`.

schunka: that will fail misserably
your condition (AKA original one) does not match when there is no quotes at all (eg. this site)
"|" will ensure match even for empty string (meaning no quotes around)
if you know about a better solution that works let me know, but original regex does not match this:
```
<script type=application/ld+json>
            {
  "@context": "http://schema.org",
  "@type": "VideoObject",
  "name": "Umělé oplodnění. Kolik stojí naděje mít vlastní dítě?",
  "description": "Naše pořady můžete nově sledovat také v aplikacích pro \r\niOS\r\n a \r\nAndroid\r\n.\r\nZplození dítěte je posvátným okamžikem, k němuž však čím dál častěji nedochází přirozenou cestou. Změny životního prostře",
  "thumbnailUrl": "https://peshkaqeni.gjirafa.com/api/media/malltv/t0g01y/retina.jpg",
  "duration": "PT15M27S",
  "uploadDate": "2019-02-01T08:30:00.0000000",
  "interactionCount": 5080
}
        </script>
```
which is what comes from the server

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,66 @@
...
+        webpage = self._download_webpage(url, video_id)
+        info = self._search_json_ld(webpage, video_id, default={})
+
+        format_url = self._html_search_regex(
+            r'<source src=([\"\'])?(?P<src>.+?index)\1?[^>]*?>',
+            webpage, 'm3u8 URL', group='src')
+        formats = self._extract_m3u8_formats(format_url+'.m3u8',
+                                             video_id, 'mp4')
+        self._sort_formats(formats)
+        thumbnail = info.get('thumbnailUrl', self._og_search_thumbnail(webpage))
```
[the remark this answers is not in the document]
schunka: as you can see in this excerpt, there actually is thumbnailUrl, which will be there after _parse_json made in _search_json_ld

```
<script type=application/ld+json>
            {
  "@context": "http://schema.org",
  "@type": "VideoObject",
  "name": "Umělé oplodnění. Kolik stojí naděje mít vlastní dítě?",
  "description": "Naše pořady můžete nově sledovat také v aplikacích pro \r\niOS\r\n a \r\nAndroid\r\n.\r\nZplození dítěte je posvátným okamžikem, k němuž však čím dál častěji nedochází přirozenou cestou. Změny životního prostře",
  "thumbnailUrl": "https://peshkaqeni.gjirafa.com/api/media/malltv/t0g01y/retina.jpg",
  "duration": "PT15M27S",
  "uploadDate": "2019-02-01T08:30:00.0000000",
  "interactionCount": 5080
}
        </script>
```

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,66 @@
...
+        duration = parse_duration(info.get('duration'))
+        result = {
+            'id': hashlib.md5(video_id).hexdigest().decode('utf8'),
+            'title': info.get('name', self._og_search_title(webpage)),
```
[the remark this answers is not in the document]
schunka: as you can see in this excerpt, there actually is "name", which will be there after _parse_json made in _search_json_ld

so same as before
```
<script type=application/ld+json>
            {
  "@context": "http://schema.org",
  "@type": "VideoObject",
  "name": "Umělé oplodnění. Kolik stojí naděje mít vlastní dítě?",
  "description": "Naše pořady můžete nově sledovat také v aplikacích pro \r\niOS\r\n a \r\nAndroid\r\n.\r\nZplození dítěte je posvátným okamžikem, k němuž však čím dál častěji nedochází přirozenou cestou. Změny životního prostře",
  "thumbnailUrl": "https://peshkaqeni.gjirafa.com/api/media/malltv/t0g01y/retina.jpg",
  "duration": "PT15M27S",
  "uploadDate": "2019-02-01T08:30:00.0000000",
  "interactionCount": 5080
}
        </script>
```

youtube_dl/utils.py
dstftw: No, it won't. Bother to read it carefully. `(["\']?)` is idiomatic and correct way to say `(["\']|)`.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,66 @@
...
+        format_url = self._html_search_regex(
+            r'<source src=([\"\'])?(?P<src>.+?index)\1?[^>]*?>',
+            webpage, 'm3u8 URL', group='src')
+        formats = self._extract_m3u8_formats(format_url+'.m3u8',
+                                             video_id, 'mp4')
+        self._sort_formats(formats)
+        thumbnail = info.get('thumbnailUrl', self._og_search_thumbnail(webpage))
+        duration = parse_duration(info.get('duration'))
+        result = {
+            'id': hashlib.md5(video_id).hexdigest().decode('utf8'),
```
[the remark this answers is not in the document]
schunka: Well that's what I initially thought too. But after considerable amount of tries you find out it is not. And that it actually rarely changes for different videos.

youtube_dl/extractor/malltv.py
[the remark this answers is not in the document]
dstftw: Again: original keys from `ld+json` have nothing to do with info dict returned by `_search_json_ld`. Bother to check its code.

youtube_dl/extractor/malltv.py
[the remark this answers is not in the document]
dstftw: I've browsed quite some videos and haven't found any duplicate ids. Provide example URLs wth the same ids.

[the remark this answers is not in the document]
schunka: I'm sorry, you are right. That was my mistake made in the past.

youtube_dl/extractor/malltv.py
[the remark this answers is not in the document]
schunka: I did, and missed `extract_video_object` part. My bad. You're right.

youtube_dl/extractor/malltv.py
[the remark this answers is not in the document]
schunka: as before

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        info = self._search_json_ld(webpage, video_id, default={})
+
+        format_url = self._html_search_regex(
+            r'<source src=([\"\'])?(?P<src>.+?index)\1?[^>]*?>',
+            webpage, 'm3u8 URL', group='src')
+        formats = self._extract_m3u8_formats(format_url+'.m3u8',
```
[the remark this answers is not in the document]
schunka: what is the benefit of such thing?
I would have to search for source url to add ".m3u8" (like searching for format_url now)
and then use _parse_html5_media_entries, (which uses _extract_m3u8_formats anyway)

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,66 @@
...
+                                'poplatníků, pak logicky chceme vědět, jak s '
+                                'nimi nakládá. Objem dotací pro neziskovky '
+                                'roste, ale opravdu jsou tyto organizace '
+                                '„pijavice", jak o nich hovoří And')
+                }
+        }
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
```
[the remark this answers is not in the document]
schunka: OK will be fixed

youtube_dl/extractor/malltv.py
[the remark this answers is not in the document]
dstftw: The benefit in not wasting time reinventing the wheel and code reusage as it already extracts formats, subtitles, thumbnail and possibly more in future.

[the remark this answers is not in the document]
schunka: And tinkering with the processed data this way is ok?
Is there a tool for manipulating html in youtube-dl so I wouldn't "reinvent the wheel"?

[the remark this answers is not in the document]
dstftw: I don't see any technical difference. Both ways are "tinkering" with `m3u8` extension. The only difference in one case is that you use already written and tested code and get metadata extracted for free.
`re.sub` is the tool.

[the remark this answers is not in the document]
schunka: say no more then

youtube_dl/utils.py
schunka: have it your way, it probably is more readable.

schunka:
Well this should be all.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,56 @@
...
+                'title': '18 miliard pro neziskovky. Opravdu jsou sportovci nebo Člověk v tísni pijavice?',
+                'description': "Pokud někdo hospodaří s penězmi daňových poplatníků, pak logicky chceme vědět, jak s nimi nakládá. Objem dotací pro neziskovky roste, ale opravdu jsou tyto organizace „pijavice', jak o nich hovoří And"
+                }
+        }
+    ]
+
+    def _real_extract(self, url):
+        display_id = self._match_id(url)
+        webpage = self._download_webpage(url, display_id)
+        src_id_regex = r'(?P<before><source src=([\"\'])?)(?P<src>.+?/(?P<id>\w{6,}?)/index)(?P<after>\1?[^>]*?>)'
```
dstftw: There is no point in before and after parts. Matching one capturing group from `<source` to `index` is enough.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,56 @@
...
+        video_id = self._html_search_regex(src_id_regex, webpage, 'ID',
+                                           group='id')
+        info = self._search_json_ld(webpage, video_id, default={})
+        html = re.sub(src_id_regex, r'\g<before>\g<src>.m3u8\g<after>', webpage)
+        media = self._parse_html5_media_entries('', html, video_id)
```
dstftw: `url` should be passed as first arg.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,56 @@
...
+        thumbnail = info.get('thumbnail', self._og_search_thumbnail(webpage))
+        duration = parse_duration(info.get('duration'))
+        result = {
+            'id': video_id,
+            'title': info.get('title', self._og_search_title(webpage)),
+            'description': self._og_search_description(webpage)
+        }
+        if media:
```
dstftw: This check does not make sense since `media` contains mandatory data continuing without which is pointless. Read coding conventions.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,56 @@
...
+            result.update(media[0])
+        result.update({'thumbnail': thumbnail})
+        result.update({'duration': duration})
```
dstftw: Use `merge_dicts` to cleverly merge metadata from `media[0]`, `info` and `result`.

youtube_dl/extractor/common.py
```diff
@@ -1052,7 +1052,7 @@ def _get_tfa_info(self, note='two-factor verification code'):
     @staticmethod
     def _og_regexes(prop):
         content_re = r'content=(?:"([^"]+?)"|\'([^\']+?)\'|\s*([^\s"\'=<>`]+?))'
-        property_re = (r'(?:name|property)=(?:\'og:%(prop)s\'|"og:%(prop)s"|\s*og:%(prop)s\b)'
+        property_re = (r'(?:name|property)=(?:\'og[:-]%(prop)s\'|"og[:-]%(prop)s"|\s*og[:-]%(prop)s\b)'
```
dstftw: Add some [tests](https://github.com/rg3/youtube-dl/blob/5dda1edef93d94c9a49672f905df0c49c75c5739/test/test_InfoExtractor.py#L54-L74).

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,56 @@
...
+            'info_dict': {
+                'id': 't0zzt0',
+                'ext': 'mp4',
+                'title': '18 miliard pro neziskovky. Opravdu jsou sportovci nebo Člověk v tísni pijavice?',
+                'description': "Pokud někdo hospodaří s penězmi daňových poplatníků, pak logicky chceme vědět, jak s nimi nakládá. Objem dotací pro neziskovky roste, ale opravdu jsou tyto organizace „pijavice', jak o nich hovoří And"
+                }
+        },
+        {
+            'url': 'https://www.mall.tv/kdo-to-plati/18-miliard-pro-neziskovky-opravdu-jsou-sportovci-nebo-clovek-v-tisni-pijavice',
+            'md5': '9ced0de056534410837077e23bfba796',
```
dstftw: Duplicate tests must be `only_matching`.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,56 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import hashlib
+from .common import InfoExtractor
+from ..utils import parse_duration
+import re
```
dstftw: Std lib imports should go before youtube-dl imports.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class MallTVIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?mall\.tv/(?:.+/)?(?P<id>.+)'
```
dstftw: Still does not handle aforementioned URLs.

youtube_dl/extractor/malltv.py
```diff
@@ -0,0 +1,56 @@
...
+                'description': "Pokud někdo hospodaří s penězmi daňových poplatníků, pak logicky chceme vědět, jak s nimi nakládá. Objem dotací pro neziskovky roste, ale opravdu jsou tyto organizace „pijavice', jak o nich hovoří And"
+                }
+        }
+    ]
+
+    def _real_extract(self, url):
+        display_id = self._match_id(url)
+        webpage = self._download_webpage(url, display_id)
+        src_id_regex = r'(?P<before><source src=([\"\'])?)(?P<src>.+?/(?P<id>\w{6,}?)/index)(?P<after>\1?[^>]*?>)'
+        video_id = self._html_search_regex(src_id_regex, webpage, 'ID',
```
dstftw: `_search_regex` is enough here.

youtube_dl/extractor/malltv.py
schunka: got it.

youtube_dl/extractor/malltv.py
dstftw: `_parse_html5_media_entries` has nothing to do with this regex and has its own parsing routines. All you need to do is to add `.m3u8` after index, you don't need to even bother about what follows it.

schunka: well in that case, I'm lost. What I understand it does is this:
This regex is used in two places, one is for video_id, where everything other does not matter.
And second place is for replacing `source` tag in `re.sub`.
Result of that replacement is then passed to `_parse_html5_media_entries` in which the `source` tag must be complete otherwise it fails. It can include other source attributes.

One of us is missing something. It is probably me, but I cannot comprehend what it could be.

schunka:
except for that one thing I'm not sure about this is done.

PR #17904, opened 2018-10-16
pass on extractor's info to postprocessors

mx1up:
**pass on extractor's info to postprocessors so video metadata can be used in postprocessing_args**

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Hello there,

I am extracting a youtube playlist to mp3's and wanted to set the Album tag to the playlist title using the postprocessor args like this:

youtube-dl -x --audio-format mp3 --embed-thumbnail --postprocessor-args "-id3v2_version 3 -metadata **album=%(playlist_title)s**" --batch-file $BASEPATH/urls.txt --add-metadata -o "$BASEPATH/$outFile"

the `%(playlist_title)s` variable does get not resolved so I assumed this is not implemented. To implement this, I passed on the extractor info dict to the postprocessor so the variable could be substituted with its value in the postprocessor code. 

I am new to python so please bear with me. Also I'm not sure whether the map function I used is available in python 2.6?

thanks for this amazing project


youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -175,13 +175,13 @@ def get_audio_codec(self, path):
...
 
-    def run_ffmpeg_multiple_files(self, input_paths, out_path, opts):
+    def run_ffmpeg_multiple_files(self, input_paths, out_path, opts, info):
         self.check_version()
 
         oldest_mtime = min(
             os.stat(encodeFilename(path)).st_mtime for path in input_paths)
 
-        opts += self._configuration_args()
+        opts += map(lambda s: s % info, self._configuration_args())
```
dstftw: Won't work. See how this is done for output template.

mx1up: I indeed had a look at the output template code but had a hard time understanding it. In my case, the code works, but I guess you're talking about more complex cases? Are you referring to the following code? (in prepare_filename function)

```
template_dict = dict(info_dict)
template_dict['epoch'] = int(time.time())
autonumber_size = self.params.get('autonumber_size')
if autonumber_size is None:
    autonumber_size = 5
template_dict['autonumber'] = self.params.get('autonumber_start', 1) - 1 + self._num_downloads
if template_dict.get('resolution') is None:
    if template_dict.get('width') and template_dict.get('height'):
        template_dict['resolution'] = '%dx%d' % (template_dict['width'], template_dict['height'])
    elif template_dict.get('height'):
        template_dict['resolution'] = '%sp' % template_dict['height']
    elif template_dict.get('width'):
        template_dict['resolution'] = '%dx?' % template_dict['width']
sanitize = lambda k, v: sanitize_filename(
    compat_str(v),
    restricted=self.params.get('restrictfilenames'),
    is_id=(k == 'id' or k.endswith('_id')))
template_dict = dict((k, v if isinstance(v, compat_numeric_types) else sanitize(k, v))
                     for k, v in template_dict.items()
                     if v is not None and not isinstance(v, (list, tuple, dict)))
template_dict = collections.defaultdict(lambda: 'NA', template_dict)
outtmpl = self.params.get('outtmpl', DEFAULT_OUTTMPL)
# For fields playlist_index and autonumber convert all occurrences
# of %(field)s to %(field)0Nd for backward compatibility
field_size_compat_map = {
    'playlist_index': len(str(template_dict['n_entries'])),
    'autonumber': autonumber_size,
}
FIELD_SIZE_COMPAT_RE = r'(?<!%)%\((?P<field>autonumber|playlist_index)\)s'
mobj = re.search(FIELD_SIZE_COMPAT_RE, outtmpl)
if mobj:
    outtmpl = re.sub(
        FIELD_SIZE_COMPAT_RE,
        r'%%(\1)0%dd' % field_size_compat_map[mobj.group('field')],
        outtmpl)
# Missing numeric fields used together with integer presentation types
# in format specification will break the argument substitution since
# string 'NA' is returned for missing fields. We will patch output
# template for missing fields to meet string presentation type.
for numeric_field in self._NUMERIC_FIELDS:
    if numeric_field not in template_dict:
        # As of [1] format syntax is:
        #  %[mapping_key][conversion_flags][minimum_width][.precision][length_modifier]type
        # 1. https://docs.python.org/2/library/stdtypes.html#string-formatting
        FORMAT_RE = r'''(?x)
            (?<!%)
            %
            \({0}\)  # mapping key
            (?:[#0\-+ ]+)?  # conversion flags (optional)
            (?:\d+)?  # minimum field width (optional)
            (?:\.\d+)?  # precision (optional)
            [hlL]?  # length modifier (optional)
            [diouxXeEfFgGcrs%]  # conversion type
        '''
        outtmpl = re.sub(
            FORMAT_RE.format(numeric_field),
            r'%({0})s'.format(numeric_field), outtmpl)
# expand_path translates '%%' into '%' and '$$' into '$'
# correspondingly that is not what we want since we need to keep
# '%%' intact for template dict substitution step. Working around
# with boundary-alike separator hack.
sep = ''.join([random.choice(ascii_letters) for _ in range(32)])
outtmpl = outtmpl.replace('%%', '%{0}%'.format(sep)).replace('$$', '${0}$'.format(sep))
# outtmpl should be expand_path'ed before template dict substitution
# because meta fields may contain env variables we don't want to
# be expanded. For example, for outtmpl "%(title)s.%(ext)s" and
# title "Hello $PATH", we don't want `$PATH` to be expanded.
filename = expand_path(outtmpl).replace(sep, '') % template_dict
```

I guess most of it will be needed for my PR too? (except for example the expand_path stuff). But if it shares quite some code with the prepare_filename function, should it be put in a separate function? but in which file would that function belong then? 

I will create a first effort by copying the code, ok? after that, we can still generalize.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -5,7 +5,23 @@
 import subprocess
 import time
 import re
-
+import collections
+from string import ascii_letters
+import random
+from ..utils import expand_path
+# from youtube_dl import YoutubeDL
```
mx1up: when I do this import, I can no longer run the youtube_dl module: "can't find '__main__' module"
how to fix?

for now I duplicated the _NUMERIC_FIELDS field

youtube_dl/postprocessor/ffmpeg.py
mx1up: I updated the code to do similar stuff as for output template

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -214,6 +230,83 @@ def _ffmpeg_filename_argument(self, fn):
...
         return 'file:' + fn if fn != '-' else fn
 
+    def _resolve_postprocessor_arg_var(self, arg, info):
+        # map(lambda s: s % info, args)
+        if "%(" not in arg:
+            return arg
+
+        template_dict = dict(info)
+        template_dict['epoch'] = int(time.time())
+        # autonumber_size = self.params.get('autonumber_size')
```
mx1up: I copied the code from YoutubeDL and commented the lines that I thought were not applicable anymore
Just for easier review, I cleanup afterwards

ZedeN1:
Thanks, works great for putting full date, %(upload_date)s, as the date id3 tag

mx1up:
@dstftw / devteam, could you review my questions, please?

brandazzle:
This seems useful - has it been merged yet?

hardcore-coder:
Could anyone tell what's missing here? It would be so helpful to customize the meta data for post processing.

raine:
Just had a use case for this.

I'm downloading youtube videos to be listened as podcasts, and in order for the mp3 to be recognized as a specific podcast, it needs to have an album metatag.

The ideal way to do that would be passing `--postprocessor-args "-metadata album=%(channel)s"` to youtube-dl.

MrCyjaneK:
I'd like to have this in youtube-dl too... Seems really useful.

Chad90b:
any news about this or are we going to be stuck in do-not-merge limbo?

geotheory:
Github island on XKCD's internet map needs a sinkhole of forgotten PRs

alpe12:
No way to add url/channel to metadata yet?

brandazzle:
Welp, now this breaks the main program when used, causing `Exception: Unsupported JS expression in: u'throw d'`.

pirate:
Bump, this would be awesome to have still

PR #18086, opened 2018-11-06
skip-http-head flag feature

Quitten:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [x] New feature

---

Today I wanted to download mdp file. so i have a sepcific URL refer to this mdp file, this URL is one time URL, and its expires after 1 request.
youtube-dl is sending HEAD request before the GET one, cause that my URL is always expires and I wasn't able to download it.
By skipping the head request you can actually make it work and download the requested resource.

### DIFF
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -403,7 +403,7 @@ def _real_main(argv=None):
         'prefer_ffmpeg': opts.prefer_ffmpeg,
         'include_ads': opts.include_ads,
         'default_search': opts.default_search,
-        'skip_http_head': opts.skip_http_head,
+        'skip_http_head': opts.generic_skip_http_head,
         'youtube_include_dash_manifest': opts.youtube_include_dash_manifest,
         'encoding': opts.encoding,
         'extract_flat': opts.extract_flat,
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -403,6 +403,7 @@ def _real_main(argv=None):
         'prefer_ffmpeg': opts.prefer_ffmpeg,
         'include_ads': opts.include_ads,
         'default_search': opts.default_search,
+        'skip_http_head': opts.skip_http_head,
         'youtube_include_dash_manifest': opts.youtube_include_dash_manifest,
         'encoding': opts.encoding,
         'extract_flat': opts.extract_flat,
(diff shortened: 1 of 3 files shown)


dstftw:
I won't accept this until solid rationale is provided.

Quitten:
@dstftw - You are right, I didn't explain the reason I needed it.
Today I wanted to download mdp file. so i have a sepcific URL refer to this mdp file, this URL is one time URL, and its expires after 1 request.
youtube-dl is sending HEAD request before the GET one, cause that my URL is always expires and I wasn't able to download it.
By skipping the head request you can actually make it work and download the requested resource.

Quitten:
@dstftw  - WDYT???

dstftw:
This is too specific and rare for general case and does not fit into Network Options category. Rather it's solely an option of the generic extractor thus:
1. Option should be prefixed with generic_.
2. It should start a new category Generic Extractor options.

Quitten:
@dstftw - Files changed according your comments.

keithah:
I have a url which said:
WARNING: Could not send HEAD request to http://(urlwithtoken): HTTP Error 405: Method Not Allowed

But once I applied this patch and used -skip-http-head, it worked fine. I've experienced this a few times where this patch has been useful, esp for CNBC/theplatform videos.

davispuh:
Getting same warning, there are streaming servers that doesn't support `HEAD` so this PR would be useful to avoid that warning.

devnoname120:
I also encountered this issue with a server that took forever to reply with `504: Gateway Time-out` to the `HEAD` request.

I was trying to record a live stream so it was extremely annoying. I missed the start due to the fact that I couldn't skip this `HEAD` request (which in my case is useless).

pukkandan:
The head request is really just legacy code, and should be fully removed imo - no need of any option. See https://github.com/yt-dlp/yt-dlp/commit/61544381781d35276e1e7831456c653107ac8909

devnoname120:
@pukkandan Awesome. I guess that this PR can be closed now as it's superseded by https://github.com/yt-dlp/yt-dlp/commit/61544381781d35276e1e7831456c653107ac8909

pukkandan:
Not really - that commit is from yt-dlp, not this project. I just linked it for reference. It'll need to be ported and merged here

Issue #18202, opened 2018-11-16
NBC 


lincalinca:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like this: `[x]`)
- Use the *Preview* tab to see what your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2018.11.07*. If it's not, read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2018.11.07**
(I wasn't, then I upgraded from 2018-10-05, and the issue persisted; the logs below are from after upgrading)

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
- [ ] Checked that provided video/audio/playlist URLs (if any) are alive and playable in a browser (I couldn't do this, because I was using a proxy, which I haven't activated - I have had people geolocated in the US check, and the video works

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add the `-v` flag to **your command line** you run youtube-dl with (`youtube-dl -v <your command line>`), copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):

```
E:\OURS\The Good Place\nbcdl>youtube-dl --proxy "23.101.121.11:3128" -u ###.###@###.com -p ############ https://www.nbc.com/the-good-place/video/michaels-gambit-extended-cut/3454089 --verbose
[debug] System config: []
[debug] User config: []
[...]

PR #18233, opened 2018-11-18
Fix NBC extractor

josejuan05:
Edit: Realized my bug was separate from OP in #18202; Created a separate Issue.

See Issue <s>#18202</s>#18236

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

NBC extractor is failing on some videos, as seen in Issue <s>#18202</s>#18236. Proposed change fixes problem.

### DIFF
--- a/youtube_dl/extractor/nbc.py
+++ b/youtube_dl/extractor/nbc.py
@@ -93,6 +93,9 @@ class NBCIE(AdobePassIE):
                 'fields[shows]': 'shortTitle',
                 'include': 'show.shortTitle',
             })
+        if len(response['data']) is 0:
+          response = self._download_json(response['links']['self'],video_id)
+        
         video_data = response['data'][0]['attributes']
         query = {
             'mbr': 'true',


remitamine:
this does not fix anything, it requests the same URL twice.

josejuan05:
This is an incorrect assessment. The URL in request['links']['self'] is not the same URL that was initially requested. See #18236. The bug is transient but predictable. In my case, it seems to occur for new episodes of Saturday Night Live on the NBC.com website the morning after the episode initially airs. It has occurred consistently for me for the past 3 weeks: If I try in the morning it will fail with the error output as given in the linked bug report. The patch given here fixes it. By the afternoon something on the NBC.com backend has been changed and youtube-dl works without the patch.

Maybe the more correct solution is to change the initial request to api.nbc.com. I don't know. But this patch consistently fixes a consistent problem.

remitamine:
> Maybe the more correct solution is to change the initial request to api.nbc.com.

yes, from the output you have posted in https://github.com/rg3/youtube-dl/issues/18236#issuecomment-443537052, the request adds `page%5Bnumber%5D=1` to the api URL query, so if adding this to the iapi request(without making the api request twice) fixes the problem you explaining and does not effect downloading other videos, then i would accept the change even thought i can't reproduce the problem locally.

tych0:
Hi, I can confirm that this patch fixes this issue for me; any chance we can re-open and merge it?

PR #18259, opened 2018-11-21
fix now-obfuscated CDN URL

heisenberg33:


dstftw:
Current master code works just fine.

PR #18437, opened 2018-12-07
Add age-limit option on dailymotion to download explicit

coffeecore:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

https://github.com/rg3/youtube-dl/issues/18398 On Dailymotion, you can't download Explicit Content without cookie. It's because metadata page has GET params which raises error. If you remove these params, you can download. So, I added test on age_limit CLI option to generate good url. (Sorry, I'm not Python developer, I hope no bullshit in my little add)


youtube_dl/extractor/dailymotion.py
```diff
@@ -182,13 +182,18 @@ def _real_extract(self, url):
...
-                        'https://www.dailymotion.com/player/metadata/video/%s'
-                        % video_id, {
-                            'embedder': url,
-                            'integration': 'inline',
-                            'GK_PV5_NEON': '1',
-                        })
+                    if (17 < age_limit) :
+                        metadata_url = update_url_query(
+                            'https://www.dailymotion.com/player/metadata/video/%s'
+                            % video_id, {})
```
mdaniel: By far the better solution is to pull the extra `dict` out into a variable guarded by this `if` rather than duplicating the "meat" of that call; something like:

```python
                    if 17 < age_limit:
                        metadata_extra = {}
                    else:
                        metadata_extra = {
                                'embedder': url,
                                'integration': 'inline',
                                'GK_PV5_NEON': '1',
                        }
                    metadata_url = update_url_query(
                            'https://www.dailymotion.com/player/metadata/video/%s'
                            % video_id, metadata_extra)
```

youtube_dl/extractor/dailymotion.py
```diff
@@ -182,13 +182,18 @@ def _real_extract(self, url):
...
                     metadata_url = metadata_url.replace(':videoId', video_id)
                 else:
-                    metadata_url = update_url_query(
-                        'https://www.dailymotion.com/player/metadata/video/%s'
-                        % video_id, {
-                            'embedder': url,
-                            'integration': 'inline',
-                            'GK_PV5_NEON': '1',
-                        })
+                    if (17 < age_limit) :
```
mdaniel: Please continue to make the code [PEP8](https://www.python.org/dev/peps/pep-0008/) compliant, which the extra space between the condition and the `:` violate, as does the extra space between your `else` and `:` below

The parens around the test are also unnecessary, but I don't believe they're strictly speaking wrong

PR #18568, opened 2018-12-18
CS: DRY: use url_result

glensc:
This is code cleanup PR.

Use `url_result` method where possible, rather creating appropriate structure manually.

cc @dstftw


glensc:
@dstftw fixed error that caused travis to fail

PR #18569, opened 2018-12-18
[chore] DRY: move common _VALID_URL_RE caching to a method

glensc:
When poking around the code, noticed copy-pasted code. moved it to subroutine.

cc @dstftw


glensc:
@dstftw rebased against 00eb865b3. seems nothing changed in codebase meanwhile :)

Issue #10336, opened 2016-08-13
youtube-dl doesn't use locking to prevent concurrent writing to output files


ssokolow:
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.08.12**
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

```
% youtube-dl -v
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2016.08.12
[debug] Python version 2.7.12 - Linux-3.13.0-91-generic-x86_64-with-Ubuntu-14.04-trusty
[debug] exe versions: avconv 2.8.1-1, avprobe 2.8.1-1, ffmpeg 2.8.1-1, ffprobe 2.8.1-1, rtmpdump 2.4
[debug] Proxy map: {}
Usage: youtube-dl [OPTIONS] URL [URL...]

youtube-dl: error: You must provide at least one URL.
Type youtube-dl --help to see a list of all options.
```

It doesn't matter which URLs it's being run on and I couldn't find any evidence of concurrency-aware operation in the `--help` output or FAQ, so I didn't think it was necessary to get verbose output from the actual operation.

---

### Description of your *issue*, suggested solution and other information

I run youtube-dl via a cronscript to download new videos on certain YouTube channels for offline viewing. However, I sometimes also run the script manually. This can result in two copies of `youtube-dl` trying to download the same URL into the same folder at the same time.

When that happens, the interactions between the two copies of youtube-dl cause the tail end of the video to be repeated in the file. youtube-dl should use some form of locking to prevent this.

[...]

Issue #485, opened 2012-10-22
processes bidding for the same partially downloaded file ...


Albretch:
~ 
 at the risk of giving you enough reasons to start hating me ;-) ...
~ 
[use case]
~ 
 the only way I have seen you can speed up downloads for many media files is by running a number of processes at the same time. So I have a script to get a lot of youtube URIs, which I then sort, split in smaller files and reorder (one process processes the list from a-z the other from z-a ...) in order to make sure that if a feed is partially downloaded, options are the other process will pick it up and finish the download
~ 
[/use case]
~ 
 Now, if a process is downloading a file (so it temporarily suffixes it .part) and another process gets a chance to download the same file, it will 'think' (sorry for the anthropomorphizing) that it was left partially downloaded by some previous process (and not that is being downloaded by a current one)
~ 
 Short of checking with the OS to see if any process is currently writing to that file, I think, a delay of, say, 5 seconds without any increase of data would be a good bet that file is not being currently worked on, otherwise yo "say something" and keep going
~ 
 thanks
 lbrtchx

yan12125:
I'd like to hijack this thread. Since #1562, there is a class ```locked_file``` in ```youtube_dl/utils.py```. Implementing this should be much easier now.

PR #18690, opened 2018-12-29
modify sanitize_open() to use locked_file()

jakeogh:
modify sanitize_open() to use locked_file(), preventing silent corruption when a second youtube-dl instance is attempting to write the same file. There is still a corner case, if a .part file is being used (--no-part is not enabled), in that the .part file is closed before it's renamed to remove the .part, in that window, another process could modify the .part file before it's renamed. Using --no-part prevents this corner case.

related bugs #10336 #485 #1562

---

- [X] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Normally, if a second process is launched with the same target url, both processes will append to the same temp file, causing the temp file to be inconsistent beacuse both processes are appending bytes, while the 2nd process decided to start appending at byte x, the first process has already witten bytes after byte x.

Importantly, the advisory lock is made nonblocking, so the second process exits with an error instead of waiting to write to a complete file that the first process just dropped the lock on.

### DIFF
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -455,7 +455,8 @@ def sanitize_open(filename, open_mode):
                 import msvcrt
                 msvcrt.setmode(sys.stdout.fileno(), os.O_BINARY)
             return (sys.stdout.buffer if hasattr(sys.stdout, 'buffer') else sys.stdout, filename)
-        stream = open(encodeFilename(filename), open_mode)
+        stream = locked_file(encodeFilename(filename), open_mode, block=False)
+        stream = stream.__enter__()
         return (stream, filename)
     except (IOError, OSError) as err:
         if err.errno in (errno.EACCES,):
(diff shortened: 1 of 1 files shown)


jakeogh:
Would be happy to update this, I deliberately attempted to submit the minimal change set with the assumption that would be easier to get merged. I'm running this patch with the current master head, no problems. Did I miss something important?

dstftw:
>Did I miss something important?

Yes, working code. Have you even tested it on Windows?

jakeogh:
Implemented in yt-dlp:
https://github.com/yt-dlp/yt-dlp/pull/1066
https://github.com/yt-dlp/yt-dlp/pull/2714
https://github.com/yt-dlp/yt-dlp/pull/2994
https://github.com/yt-dlp/yt-dlp/pull/3610

There is still a rare corner case, even when locking is working, because locks are released before file media audio/video merge happens. https://gist.github.com/jakeogh/dd44d2f998b7c6eac9e1985f3db9fc59

PR #18696, opened 2018-12-30
[hketv] Add new extractor

anthonyfok:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

The Educational Bureau of the Government of Hong Kong has been producing HKETV (香港教育電視), i.e. Educational Television (ETV, eTV), for students from kindergarten to secondary schools in Hong Kong for over 40 years.  Their website is at https://etv.edb.gov.hk/ .

This new extractor pulls information from the website through https://apps.hkedcity.net/media/play/handler.php which has recently (mid-2018?) become geo-restricted, i.e. viewable in Hong Kong only.  The final URL to the MPEG-4 video looks something like https://video1.hkedcity.net/streaming/79ea4e9feec5e8291a7ffc72632ce19e/5c28ab09/channels/etv/201810/20181024230027_355992_720.mp4 which is not geo-restricted, but due to URL (Token) signing, the URL expires after an hour or so.

Note that the official ETV [Android app](https://play.google.com/store/apps/details?id=com.hketv.app) and [iOS app](https://itunes.apple.com/hk/app/etv/id563107065?mt=8) are not currently geo-restricted, though they have not been studied in depth for this extractor.

Please do not hesitate to let me know if you have any questions.

Cheers,
Anthony

### DIFF
--- a/youtube_dl/extractor/hketv.py
+++ b/youtube_dl/extractor/hketv.py
@@ -13,7 +13,6 @@ from ..utils import (
     try_get,
     unified_strdate,
     urlencode_postdata,
-    urljoin,
 )
 
 
@@ -59,9 +58,10 @@ class HKETVIE(InfoExtractor):
     def _real_extract(self, url):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
+        title = self._html_search_meta('ed_title', webpage, fatal=True)
 
-        file_id = self._html_search_regex(r'post_var\["file_id"\] = ([0-9]+);', webpage, 'file ID')
-        curr_url = self._html_search_regex(r'post_var\["curr_url"\] = "(.+?)";', webpage, 'curr URL')
+        file_id = self._html_search_regex(r'post_var\["file_id"\]\s*=\s*(.+?);', webpage, 'file ID')
+        curr_url = self._html_search_regex(r'post_var\["curr_url"\]\s*=\s*"(.+?)";', webpage, 'curr URL')
         data = {
             'action': 'get_info',
             'curr_url': curr_url,
@@ -69,7 +69,7 @@ class HKETVIE(InfoExtractor):
             'video_url': file_id,
         }
         _APPS_BASE_URL = 'https://apps.hkedcity.net'
-        handler_url = urljoin(_APPS_BASE_URL, '/media/play/handler.php')
+        handler_url = _APPS_BASE_URL + '/media/play/handler.php'
 
         response = self._download_json(
             handler_url, video_id,
@@ -77,14 +77,10 @@ class HKETVIE(InfoExtractor):
             headers=merge_dicts({'Content-Type': 'application/x-www-form-urlencoded'},
                                 self.geo_verification_headers()))
 
-        result = try_get(response, lambda x: x['result'], dict)
+        result = response.get('result')
 
         formats = []
         subtitles = {}
-
-        thumbnail_php = urljoin(_APPS_BASE_URL, result.get('image'))
-        thumbnail_urlh = self._downloader.urlopen(thumbnail_php)
-        thumbnail = thumbnail_urlh.geturl()
 
         if response.get('success') and response.get('access'):
             width = str_to_int(result.get('width'))
@@ -99,7 +95,7 @@ class HKETVIE(InfoExtractor):
                 elif label == 'SD':
                     h = 360
                 w = h * width // height
-                urlh = self._downloader.urlopen(urljoin(_APPS_BASE_URL, fmt.get('file')))
+                urlh = self._downloader.urlopen(_APPS_BASE_URL + fmt.get('file'))
                 formats.append({
                     'format_id': label,
                     'ext': fmt.get('type'),
@@ -117,7 +113,7 @@ class HKETVIE(InfoExtractor):
                     continue
                 if track_kind.lower() not in ('captions', 'subtitles'):
                     continue
-                track_url = urljoin(_APPS_BASE_URL, track.get('file'))
+                track_url = _APPS_BASE_URL + track.get('file')
                 if not track_url:
                     continue
                 track_label = track.get('label')
@@ -128,7 +124,7 @@ class HKETVIE(InfoExtractor):
 
         else:
             error = clean_html(response.get('access_err_msg'))
-            if error.find('Video streaming is not available in your country'):
+            if 'Video streaming is not available in your country' in error:
                 raise GeoRestrictedError(error)
             else:
                 raise ExtractorError(error)
@@ -142,19 +138,20 @@ class HKETVIE(InfoExtractor):
                 'data[bucket_id]': 'etv',
                 'data[identifier]': video_id,
             }),
-            headers={'Content-Type': 'application/x-www-form-urlencoded'})
+            headers={'Content-Type': 'application/x-www-form-urlencoded'},
+            fatal=False)
         if emotion.get('result'):
             like_count = str_to_int(try_get(emotion, lambda x: x['data']['emotion_data'][0]['count'], str))
 
         return {
             'id': video_id,
-            'title': self._html_search_meta('ed_title', webpage),
+            'title': title,
             'description': self._html_search_meta('description', webpage, fatal=False),
             'upload_date': unified_strdate(self._html_search_meta('ed_date', webpage, fatal=False), day_first=False),
             'duration': int_or_none(result.get('length')),
             'formats': formats,
             'subtitles': subtitles,
-            'thumbnail': thumbnail,
+            'thumbnail': _APPS_BASE_URL + result.get('image'),
             'view_count': str_to_int(result.get('view_count')),
             'like_count': like_count,
         }
--- a/youtube_dl/extractor/hketv.py
+++ b/youtube_dl/extractor/hketv.py
@@ -1,0 +1,160 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..compat import compat_str
+from ..utils import (
+    ExtractorError,
+    GeoRestrictedError,
+    clean_html,
+    int_or_none,
+    merge_dicts,
+    str_to_int,
+    try_get,
+    unified_strdate,
+    urlencode_postdata,
+    urljoin,
+)
+
+
+class HKETVIE(InfoExtractor):
+    IE_NAME = 'hketv'
+    IE_DESC = '香港教育局教育電視 (HKETV) Educational Television, Hong Kong Educational Bureau'
+    _VALID_URL = r'https?://(?:www\.)?hkedcity\.net/etv/resource/(?P<id>[0-9]+)'
+    _TESTS = [{
+        'url': 'https://www.hkedcity.net/etv/resource/2932360618',
+        'md5': 'f193712f5f7abb208ddef3c5ea6ed0b7',
+        'info_dict': {
+            'id': '2932360618',
+            'ext': 'mp4',
+            'title': '喜閱一生(共享閱讀樂)',
+            'description': '本節目輯錄了「閱讀滿Fun嘉年華」和「二○一八響應世界閱讀日――悅愛閱讀・愈讀愈愛」的活動花絮，並由學者、作家、演藝界人士等，分享培養子女閱讀興趣和習慣的方法，以及呼籲大家一同分享閱讀的樂趣。',
+            'upload_date': '20181024',
+            'duration': 900,
+            'subtitles': {
+                'en': [{
+                    'url': 'https://apps.hkedcity.net/media/mediaplayer/caption.php?f=74395&lang=en',
+                    'ext': 'srt',
+                }],
+                'zh-Hant': [{
+                    'url': 'https://apps.hkedcity.net/media/mediaplayer/caption.php?f=74395&lang=qmt',
+                    'ext': 'srt',
+                }],
+            }
+        },
+    }]
+
+    _CC_LANGS = {
+        '中文（繁體中文）': 'zh-Hant',
+        '中文（简体中文）': 'zh-Hans',
+        'English': 'en',
+        'Bahasa Indonesia': 'id',
+        '\u0939\u093f\u0928\u094d\u0926\u0940': 'hi',
+        '\u0928\u0947\u092a\u093e\u0932\u0940': 'ne',
+        'Tagalog': 'tl',
+        '\u0e44\u0e17\u0e22': 'th',
+        '\u0627\u0631\u062f\u0648': 'ur',
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        file_id = self._html_search_regex(r'post_var\["file_id"\] = ([0-9]+);', webpage, 'file ID')
+        curr_url = self._html_search_regex(r'post_var\["curr_url"\] = "(.+?)";', webpage, 'curr URL')
+        data = {
+            'action': 'get_info',
+            'curr_url': curr_url,
+            'file_id': file_id,
+            'video_url': file_id,
+        }
+        _APPS_BASE_URL = 'https://apps.hkedcity.net'
+        handler_url = urljoin(_APPS_BASE_URL, '/media/play/handler.php')
+
+        response = self._download_json(
+            handler_url, video_id,
+            data=urlencode_postdata(data),
+            headers=merge_dicts({'Content-Type': 'application/x-www-form-urlencoded'},
+                                self.geo_verification_headers()))
+
+        result = try_get(response, lambda x: x['result'], dict)
+
+        formats = []
+        subtitles = {}
+
+        thumbnail_php = urljoin(_APPS_BASE_URL, result.get('image'))
+        thumbnail_urlh = self._downloader.urlopen(thumbnail_php)
+        thumbnail = thumbnail_urlh.geturl()
+
+        if response.get('success') and response.get('access'):
+            width = str_to_int(result.get('width'))
+            height = str_to_int(result.get('height'))
+
+            playlist0 = try_get(result, lambda x: x['playlist'][0], dict)
+            fmts = try_get(playlist0, lambda x: x['sources'], list)
+            for fmt in fmts:
+                label = fmt.get('label')
+                if label == 'HD':
+                    h = 720
+                elif label == 'SD':
+                    h = 360
+                w = h * width // height
+                urlh = self._downloader.urlopen(urljoin(_APPS_BASE_URL, fmt.get('file')))
+                formats.append({
+                    'format_id': label,
+                    'ext': fmt.get('type'),
+                    'url': urlh.geturl(),
+                    'width': w,
+                    'height': h,
+                })
+
+            tracks = try_get(playlist0, lambda x: x['tracks'], list)
+            for track in tracks:
+                if not isinstance(track, dict):
+                    continue
+                track_kind = track.get('kind')
+                if not track_kind or not isinstance(track_kind, compat_str):
+                    continue
+                if track_kind.lower() not in ('captions', 'subtitles'):
+                    continue
+                track_url = urljoin(_APPS_BASE_URL, track.get('file'))
+                if not track_url:
+                    continue
+                track_label = track.get('label')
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,160 @@
...
+        file_id = self._html_search_regex(r'post_var\["file_id"\] = ([0-9]+);', webpage, 'file ID')
+        curr_url = self._html_search_regex(r'post_var\["curr_url"\] = "(.+?)";', webpage, 'curr URL')
+        data = {
+            'action': 'get_info',
+            'curr_url': curr_url,
+            'file_id': file_id,
+            'video_url': file_id,
+        }
+        _APPS_BASE_URL = 'https://apps.hkedcity.net'
+        handler_url = urljoin(_APPS_BASE_URL, '/media/play/handler.php')
```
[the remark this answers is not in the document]
anthonyfok: Thanks for your review.  I did have simple string concatenation with `+` originally, but saw some other extractors using `urljoin` and I changed the code shortly before submission.  I will change them back.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,160 @@
...
+            headers=merge_dicts({'Content-Type': 'application/x-www-form-urlencoded'},
+                                self.geo_verification_headers()))
+
+        result = try_get(response, lambda x: x['result'], dict)
+
+        formats = []
+        subtitles = {}
+
+        thumbnail_php = urljoin(_APPS_BASE_URL, result.get('image'))
+        thumbnail_urlh = self._downloader.urlopen(thumbnail_php)
```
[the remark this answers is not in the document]
anthonyfok: `urljoin` replaced with `+`.

Or did you mean the use of `self._downloader.urlopen()`?  That was intentionally done so that the download link of the thumbnail could be downloaded without geo-restriction, i.e. without proxy, and without the PHPSESSID cookie.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,160 @@
...
+            playlist0 = try_get(result, lambda x: x['playlist'][0], dict)
+            fmts = try_get(playlist0, lambda x: x['sources'], list)
+            for fmt in fmts:
+                label = fmt.get('label')
+                if label == 'HD':
+                    h = 720
+                elif label == 'SD':
+                    h = 360
+                w = h * width // height
+                urlh = self._downloader.urlopen(urljoin(_APPS_BASE_URL, fmt.get('file')))
```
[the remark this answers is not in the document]
anthonyfok: `urljoin` replaced with `+`.

Or did you mean the use of `self._downloader.urlopen()`?  That was intentionally done so that the download link of the video could be downloaded without geo-restriction, i.e. without proxy, and without the PHPSESSID cookie.  (The geo-restriction is done within their PHP script only.)

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,160 @@
...
+
+        else:
+            error = clean_html(response.get('access_err_msg'))
+            if error.find('Video streaming is not available in your country'):
+                raise GeoRestrictedError(error)
+            else:
+                raise ExtractorError(error)
+
+        # Likes
+        emotion = self._download_json(
```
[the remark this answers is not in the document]
anthonyfok: Thanks for catching that.  `fatal=False` added to the end of `emotion = self._download_json()`.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,160 @@
...
+                'data[bucket_id]': 'etv',
+                'data[identifier]': video_id,
+            }),
+            headers={'Content-Type': 'application/x-www-form-urlencoded'})
+        if emotion.get('result'):
+            like_count = str_to_int(try_get(emotion, lambda x: x['data']['emotion_data'][0]['count'], str))
+
+        return {
+            'id': video_id,
+            'title': self._html_search_meta('ed_title', webpage),
```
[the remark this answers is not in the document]
anthonyfok: Thanks!  The title is now extracted right after the webpage is downloaded.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,160 @@
...
+                    continue
+                track_label = track.get('label')
+                subtitles.setdefault(self._CC_LANGS.get(track_label, track_label), []).append({
+                    'url': self._proto_relative_url(track_url),
+                    'ext': 'srt',
+                })
+
+        else:
+            error = clean_html(response.get('access_err_msg'))
+            if error.find('Video streaming is not available in your country'):
```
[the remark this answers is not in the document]
anthonyfok: Thank you for catching it!  I didn't know about this till now, many thanks!
(You can probably tell I don't program often in Python...)

youtube_dl/extractor/hketv.py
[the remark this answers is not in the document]
dstftw: Yes. It does not work as you expect, since you don't specify geo verification headers for this request.
Moreover `thumbnail_php` works fine for me even without proxy and without any cookies, so all this is pointless.

youtube_dl/extractor/hketv.py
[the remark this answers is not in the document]
dstftw: Idiomatic way to check this is `'Video streaming is not available in your country' in error`.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,160 @@
...
+        _APPS_BASE_URL = 'https://apps.hkedcity.net'
+        handler_url = urljoin(_APPS_BASE_URL, '/media/play/handler.php')
+
+        response = self._download_json(
+            handler_url, video_id,
+            data=urlencode_postdata(data),
+            headers=merge_dicts({'Content-Type': 'application/x-www-form-urlencoded'},
+                                self.geo_verification_headers()))
+
+        result = try_get(response, lambda x: x['result'], dict)
```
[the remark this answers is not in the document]
anthonyfok: Thanks

youtube_dl/extractor/hketv.py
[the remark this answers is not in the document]
anthonyfok: You are right.  I have now simplified the code accordingly.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,160 @@
...
+        '\u0e44\u0e17\u0e22': 'th',
+        '\u0627\u0631\u062f\u0648': 'ur',
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        file_id = self._html_search_regex(r'post_var\["file_id"\] = ([0-9]+);', webpage, 'file ID')
+        curr_url = self._html_search_regex(r'post_var\["curr_url"\] = "(.+?)";', webpage, 'curr URL')
```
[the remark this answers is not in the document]
anthonyfok: I changed the regex to `r'post_var\["file_id"\]\s*=\s*(.+?);'` for example.
Hope that is relax enough.

youtube_dl/extractor/hketv.py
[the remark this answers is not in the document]
anthonyfok: So, may I keep using self._downloader.urlopen() for the MPEG-4 video file?  :-)

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
...
+            playlist0 = try_get(result, lambda x: x['playlist'][0], dict)
+            fmts = try_get(playlist0, lambda x: x['sources'], list)
+            for fmt in fmts:
+                label = fmt.get('label')
+                if label == 'HD':
+                    h = 720
+                elif label == 'SD':
+                    h = 360
+                w = h * width // height
+                urlh = self._downloader.urlopen(_APPS_BASE_URL + fmt.get('file'))
```
dstftw: Here you must use `urljoin`.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
...
+            tracks = try_get(playlist0, lambda x: x['tracks'], list)
+            for track in tracks:
+                if not isinstance(track, dict):
+                    continue
+                track_kind = track.get('kind')
+                if not track_kind or not isinstance(track_kind, compat_str):
+                    continue
+                if track_kind.lower() not in ('captions', 'subtitles'):
+                    continue
+                track_url = _APPS_BASE_URL + track.get('file')
```
dstftw: Same.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
...
+
+        return {
+            'id': video_id,
+            'title': title,
+            'description': self._html_search_meta('description', webpage, fatal=False),
+            'upload_date': unified_strdate(self._html_search_meta('ed_date', webpage, fatal=False), day_first=False),
+            'duration': int_or_none(result.get('length')),
+            'formats': formats,
+            'subtitles': subtitles,
+            'thumbnail': _APPS_BASE_URL + result.get('image'),
```
dstftw: Same.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
...
+            'https://emocounter.hkedcity.net/handler.php',
+            video_id,
+            data=urlencode_postdata({
+                'action': 'get_emotion',
+                'data[bucket_id]': 'etv',
+                'data[identifier]': video_id,
+            }),
+            headers={'Content-Type': 'application/x-www-form-urlencoded'},
+            fatal=False)
+        if emotion.get('result'):
```
dstftw: `emotion` is not guaranteed to be a dict.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
...
+            like_count = str_to_int(try_get(emotion, lambda x: x['data']['emotion_data'][0]['count'], str))
```
dstftw: compat_str.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
...
+                track_label = track.get('label')
+                subtitles.setdefault(self._CC_LANGS.get(track_label, track_label), []).append({
+                    'url': self._proto_relative_url(track_url),
+                    'ext': 'srt',
+                })
+
+        else:
+            error = clean_html(response.get('access_err_msg'))
+            if 'Video streaming is not available in your country' in error:
+                raise GeoRestrictedError(error)
```
dstftw: `raise_geo_restricted` and specify `countries`.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
...
+                formats.append({
+                    'format_id': label,
+                    'ext': fmt.get('type'),
+                    'url': urlh.geturl(),
+                    'width': w,
+                    'height': h,
+                })
+
+            tracks = try_get(playlist0, lambda x: x['tracks'], list)
+            for track in tracks:
```
dstftw: `tracks` is not guaranteed to be iterable.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
...
+                if not isinstance(track, dict):
+                    continue
+                track_kind = track.get('kind')
```
dstftw: `str_or_none`.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
...
+        _APPS_BASE_URL = 'https://apps.hkedcity.net'
+        handler_url = _APPS_BASE_URL + '/media/play/handler.php'
+
+        response = self._download_json(
+            handler_url, video_id,
+            data=urlencode_postdata(data),
+            headers=merge_dicts({'Content-Type': 'application/x-www-form-urlencoded'},
+                                self.geo_verification_headers()))
+
+        result = response.get('result')
```
dstftw: `get` is pointless since availability of result key is mandatory.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..compat import compat_str
+from ..utils import (
+    ExtractorError,
```
dstftw: Alphabetic.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
...
+
+        formats = []
+        subtitles = {}
+
+        if response.get('success') and response.get('access'):
+            width = str_to_int(result.get('width'))
+            height = str_to_int(result.get('height'))
+
+            playlist0 = try_get(result, lambda x: x['playlist'][0], dict)
+            fmts = try_get(playlist0, lambda x: x['sources'], list)
```
dstftw: try_get is pointless here.

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
...
+            for fmt in fmts:
+                label = fmt.get('label')
+                if label == 'HD':
+                    h = 720
+                elif label == 'SD':
+                    h = 360
+                w = h * width // height
```
dstftw: Breaks on uninitialized `h`. Breaks on None width and height.

youtube_dl/extractor/hketv.py
dstftw: Breaks on missing file key.

youtube_dl/extractor/hketv.py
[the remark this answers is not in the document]
dstftw: No, what for?

youtube_dl/extractor/hketv.py
```diff
@@ -0,0 +1,157 @@
...
+                                self.geo_verification_headers()))
+
+        result = response.get('result')
+
+        formats = []
+        subtitles = {}
+
+        if response.get('success') and response.get('access'):
+            width = str_to_int(result.get('width'))
+            height = str_to_int(result.get('height'))
```
dstftw: `int_or_none`.

youtube_dl/extractor/hketv.py
anthonyfok: Thanks!  Will be changing tracks to this instead:
```
            tracks = playlist0.get('tracks', [])
```

youtube_dl/extractor/hketv.py
anthonyfok: But I copied that few lines verbatim from `_parse_jwplayer_data` in youtube_dl/extractor/common.py!  Doesn't that mean `str_or_none` needs to be added there too?

youtube_dl/extractor/hketv.py
anthonyfok: Thanks.  I thought it were supposed to be in ASCII order, i.e. upper-case first, but then checking other extractors reveals that my assumption was wrong.  Fixed.

youtube_dl/extractor/hketv.py
anthonyfok: Thanks!  Fixed.

youtube_dl/extractor/hketv.py
anthonyfok: Thanks!  Indeed, some older videos were uploaded without width and height information, as in they are set to "0".  Changed to the following:

```
            for fmt in fmts:
                label = fmt.get('label')
                w = None
                h = None
                if label == 'HD':
                    h = 720
                elif label == 'SD':
                    h = 360
                if h:
                    if width and height:
                        w = h * width // height
                    else:
                        w = h * 4 // 3
```

youtube_dl/extractor/hketv.py
anthonyfok: Thanks!  Done.

youtube_dl/extractor/hketv.py
anthonyfok: Thanks!  Changed to:

```python
        result = response['result']
```

youtube_dl/extractor/hketv.py
anthonyfok: Changed to
```python
        if emotion.get('result'):
            like_count = int_or_none(try_get(emotion, lambda x: x['data']['emotion_data'][0]['count']))
```

youtube_dl/extractor/hketv.py
anthonyfok: Thanks!  Line removed.

youtube_dl/extractor/hketv.py
anthonyfok: Done.

youtube_dl/extractor/hketv.py
anthonyfok: Done

youtube_dl/extractor/hketv.py
anthonyfok: Done.

youtube_dl/extractor/hketv.py
anthonyfok: Fixed:
```diff
-                track_kind = track.get('kind')
+                track_kind = str_or_none(track.get('kind'))
```

youtube_dl/extractor/hketv.py
anthonyfok: Thanks!  After consulting other extractors, I am now using this:
```python
                self.raise_geo_restricted(msg=error, countries=self._GEO_COUNTRIES)
```

with the following definitions near the top:
```python
    _GEO_BYPASS = False
    _GEO_COUNTRIES = ['HK']
```

youtube_dl/extractor/hketv.py
anthonyfok: > Thanks!  Line removed.

As in
```python
    like_count = int_or_none(try_get(emotion, lambda x: x['data']['emotion_data'][0]['count']))
```
alone will work just fine without the extraneous `if emotion.get('result'):` check.

youtube_dl/extractor/hketv.py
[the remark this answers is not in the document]
anthonyfok: So that no cookie is needed with the link given by `-g`.  The link given by `urljoin(_APPS_BASE_URL, fmt.get('file'))` looks something like this:

https://apps.hkedcity.net/media/mediaplayer/filename.php?t=85ee74dae3477d95b05cc3091f26ff5a&m=mp4&p=fc61b1acf92b52238ad5c616409759360b4d914f183691d79c03e27a335a1c7e99a0f5d12b7183530daad9edc79fb17d97c3b6c0d3c445d68b96c92fa88b3a2f&file=551cc229795bb70003e9962bda09899612a7ae10b183be5e84e136419d91c684c63c863a8e3237a753ed31921500c2d28aa4db43a8684e37caac000571948110&e=.mp4

which requires the correct cookies to work.  It gives a "HTTP/1.1 302 Found" redirecting to a final URL that looks like https://video1.hkedcity.net/streaming/b3848d0a3296b85d67d88b749642c29a/5c3738d1/channels/etv/201507/20150716151039_989507_720.mp4 which requires no cookie.  This final URL is easier for me to copy-and-paste to other programs or even to stream to other devices.

That said, I have decided to take your advice and not "fully resolve" to the final URL because the intermediate URL, with the appropriate cookies saved with `--cookies`, remains valid for many hours, whereas the fully resolved URL would be "410 Gone" within one or two hours.  Also, I realized `self._downloader.urlopen` was being run twice in the `for fmt in fmts:` loop, leading to quite a few seconds of delay.  (The HKETV server seems rather slow in responding to this request.)

youtube_dl/extractor/hketv.py
anthonyfok: Thanks!  Sorry for my oversight.  I have restructured that code a little bit to check that fmt.get('file') is not None before continuing with appending to formats.

anthonyfok:
Thank you for your second round of thorough review.

I apologize for taking so long to complete the revision as I had been busy, and also it took me a while to think through the various issues you pointed out in your review.

I think I have finally fixed all the issues that you pointed out, and additionally added an additional test case of an 2007 HKETV video which has no subtitle, has width and height set as "0", and used to fail with my earlier code due to insufficient error checking.  Thank you indeed for correctly pointing out these issues in your review!

dstftw:
Cherry-picked, thanks.

anthonyfok:
Thank you for your review, improvement and cherry-pick!
Much appreciated!

PR #18710, opened 2019-01-01
[Streamango] Added support to new domain

yonaikerlol:
- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

Streamango has a new domain, similar to what Openload does. So I add support to this domain.

### DIFF
--- a/youtube_dl/extractor/streamango.py
+++ b/youtube_dl/extractor/streamango.py
@@ -14,7 +14,7 @@ from ..utils import (
 
 
 class StreamangoIE(InfoExtractor):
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/streamango.py
```diff
@@ -14,7 +14,7 @@
 
 
 class StreamangoIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?streamango\.com/(?:f|embed)/(?P<id>[^/?#&]+)'
+    _VALID_URL = r'https?://(?:www\.)?(streamango|fruithosts)\.(com|net)/(?:f|embed)/(?P<id>[^/?#&]+)'
```
dstftw: Don't capture groups you don't use.

PR #18728, opened 2019-01-03
Allow post processing a file in the same format when post processor args are passed

mdawar:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

A small change to allow post processing a downloaded file even if the requested format is the same but with post processor args passed using `--postprocessor-args`, a use case is passing arguments to ffmpeg to cut the downloaded file using these arguments `-ss, -to, -t`, or basically any other arguments that will give more control to the user, for example:

    youtube-dl URL --recode-video mp4 --format mp4 --postprocessor-args "-ss 10 -to 120 -c copy"

Currently if we try to `--recode-video` to the same format we will get this message:

    [ffmpeg] Not converting video file FILE_PATH - already is in target format FILE_FORMAT

So I made a change to the `FFmpegVideoConvertorPP` class, there was a check for the same file extension before showing this message, so I added another check for empty post processor args obtained with `self._configuration_args()` which basically returns the option `postprocessor_args` or an empty list.

I also made a change to `YoutubeDL.post_process` to prevent the processed file from being deleted with a check to see if the file path to be deleted is the same returned from the post processor.

### DIFF
--- a/test/test_YoutubeDL.py
+++ b/test/test_YoutubeDL.py
@@ -627,6 +627,16 @@ class TestYoutubeDL(unittest.TestCase):
         self.assertTrue(os.path.exists(audiofile), '%s doesn\'t exist' % audiofile)
         os.unlink(audiofile)
 
+        class SameFilePP(PostProcessor):
+            def run(self, info):
+                with open(info['filepath'], 'wt') as f:
+                    f.write('EXAMPLE')
+                return [info['filepath']], info
+
+        run_pp({'keepvideo': False}, SameFilePP)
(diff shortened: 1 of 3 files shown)


mdawar:
@dstftw can you please take a look at this change it's small and won't affect the any component.

dstftw:
This will introduce unwanted recoding in existing scenarios and configuration setups. What should be done is the new explicit option `--force-recode-video` to force recoding in all cases.

mdawar:
@dstftw thank you for reviewing, I have created a new pull request #19156 to add this option can you please review it ? Thanks.

PR #18757, opened 2019-01-06
[Instagram] Add new [:tag] extractor, refactor [:user]

jhwgh1968:

- [ ] Bug fix
- [ ] Improvement
- [X] New extractor
- [ ] New feature

---

This adds an extractor for Instagram videos. It downloads all videos that match a given hashtag as a playlist. It is heavily based on the Instagram user extractor.

I did consider refactoring common methods instead of copy-pasting, but decided it was more important to avoid touching the other code. This way is more difficult to maintain, but I hope is easier to review and avoids the risk of regressions in the other extractors.

If you would prefer a refactor -- either now or in a separate PR -- I would be happy to do it.

### DIFF
--- a/youtube_dl/extractor/instagram.py
+++ b/youtube_dl/extractor/instagram.py
@@ -227,44 +227,37 @@ class InstagramIE(InfoExtractor):
         }
 
 
-class InstagramUserIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?instagram\.com/(?P<id>[^/]{2,})/?(?:$|[?#])'
-    IE_DESC = 'Instagram user profile'
-    IE_NAME = 'instagram:user'
-    _TEST = {
-        'url': 'https://instagram.com/porsche',
-        'info_dict': {
-            'id': 'porsche',
-            'title': 'porsche',
-        },
-        'playlist_count': 5,
-        'params': {
-            'extract_flat': True,
-            'skip_download': True,
-            'playlistend': 5,
-        }
-    }
+class InstagramPlaylistIE(InfoExtractor):
+    # A superclass for handling any kind of query based on GraphQL which
+    # results in a playlist.
 
-    _gis_tmpl = None
+    _gis_tmpl = None  # used to cache GIS request type
 
-    def _entries(self, data):
+    def _parse_graphql(self, webpage, id):
+        # Reads a webpage and returns its GraphQL data.
+        return self._parse_json(
+            self._search_regex(
+                r'sharedData\s*=\s*({.+?})\s*;\s*[<\n]', webpage, 'data'),
+            id)
+
+    def _extract_graphql(self, data, url):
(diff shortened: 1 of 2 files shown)


dstftw:
No copy pasting here.

youtube_dl/extractor/instagram.py
```diff
@@ -227,6 +227,112 @@ def get_count(key, kind):
...
+                    'Downloading JSON page %d' % page_num, headers={
+                        'X-Requested-With': 'XMLHttpRequest',
+                        'X-Instagram-GIS': hashlib.md5(
+                            ('%s:%s' % (gis_tmpl, variables)).encode('utf-8')).hexdigest(),
+                    }, query={
+                        'query_hash': query_hash,
+                        'variables': variables,
+                    })
+                media = timeline_fn(query_json)
+                gis_tmpl = gis_tmpl
```
dstftw: `_gis_tmpl` broken.

youtube_dl/extractor/instagram.py
```diff
@@ -245,117 +351,65 @@ class InstagramUserIE(InfoExtractor):
...
+        assert tag
 
-        webpage = self._download_webpage(url, username)
+        webpage = self._download_webpage(url, tag)
 
         data = self._parse_json(
             self._search_regex(
                 r'sharedData\s*=\s*({.+?})\s*;\s*[<\n]', webpage, 'data'),
-            username)
+            tag)
```
dstftw: DRY.

youtube_dl/extractor/instagram.py
```diff
@@ -245,117 +351,65 @@ class InstagramUserIE(InfoExtractor):
...
         return self.playlist_result(
-            self._entries(data), username, username)
+            extract_graphql(self, data,
+                            _id=tag,
+                            query_hash='f92f56d47dc7a55b606908374b43a314',
+                            query_vars={'tag_name': tag},
+                            timeline_fn=lambda j:
+                                j['data']['hashtag']['edge_hashtag_to_media']
+                            ),
+            tag, tag)
```
dstftw: DRY.

youtube_dl/extractor/instagram.py
jhwgh1968: I didn't understand it was a persistent cache at first. Fixed and commented.

youtube_dl/extractor/instagram.py
jhwgh1968: Where is the repetition? All of the call arguments are different between the two classes.

jhwgh1968:
I have fixed two of the issues @dstftw, but I need clarification on the third one. Where is the repetition?

jhwgh1968:
I have reworked the classes as you requested, @dstftw. I am pretty sure there is nothing left to repeat now.

jhwgh1968:
Ping @dstftw, I think I have fixed everything. Is there anything else I need to do?

PR #18785, opened 2019-01-08
Remove useless `len(url_list) > 1` in YoutubeDL.py

nasyxx:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Remove useless `len(url_list) > 1` in YoutubeDL.py so that `download` function can support not only list but also iterable object.  The reasons are here:

1. It seems does not cause additional test errors.
2. People can directly pass an iterable object like map, and do not need `list(map)` any more.
3. When you have a very large url list, using a generator instead of a list can effectively reduce consumption.


dstftw:
This breaks the whole point of this check: in advance preventing of downloading multiple URLs to the same location specified with output template.

nasyxx:
Sorry for that.

PR #18805, opened 2019-01-10
New "does not contain" operator

slamkode:
- [x] New feature

---

Adding a "does not contain" operator allows to filter out unwanted format_id's that are not possible with the current operators. As an example using drtv there are streams with spoken subtitles containing the string SpokenSubtitles, with this new operator you can prevent them from being selected.


dstftw:
Instead of introducing a new operator the negation `!` should be implemented that will apply to any operator when prefixed.

Issue #19020, opened 2019-01-27
Handle Facebook redirects


nyuszika7h:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like this: `[x]`)
- Use the *Preview* tab to see what your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2019.01.27*. If it's not, read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2019.01.27**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
- [x] Checked that provided video/audio/playlist URLs (if any) are alive and playable in a browser

### What is the purpose of your *issue*?
- [ ] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [x] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add the `-v` flag to **your command line** you run youtube-dl with (`youtube-dl -v <your command line>`), copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):

```
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[...]

PR #19092, opened 2019-02-01
Fix for issue #19020

adi928:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Fixes the issue where facebook redirects to a warning page, causing youtube-dl to fail, instead of redirecting to the external link. Fix for issue #19020.

### DIFF
--- a/youtube_dl/extractor/facebook.py
+++ b/youtube_dl/extractor/facebook.py
@@ -39,15 +39,16 @@ class FacebookIE(InfoExtractor):
                                 photo\.php|
                                 video\.php|
                                 video/embed|
-                                story\.php
-                            )\?(?:.*?)(?:v|video_id|story_fbid)=|
+                                story\.php|
+                                flx/warn/
+                            )\?(?:.*?)(?:u|v|video_id|story_fbid)=|
                             [^/]+/videos/(?:[^/]+/)?|
                             [^/]+/posts/|
                             groups/[^/]+/permalink/
                         )|
                     facebook:
                 )
-                (?P<id>[0-9]+)
+                (?P<id>[0-9]+|https://.*)
                 '''
     _LOGIN_URL = 'https://www.facebook.com/login.php?next=http%3A%2F%2Ffacebook.com%2Fhome.php&login_attempt=1'
     _CHECKPOINT_URL = 'https://www.facebook.com/checkpoint/?next=http%3A%2F%2Ffacebook.com%2Fhome.php&_fb_noscript=1'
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/facebook.py
```diff
@@ -39,15 +39,16 @@ class FacebookIE(InfoExtractor):
                                 photo\.php|
                                 video\.php|
                                 video/embed|
-                                story\.php
-                            )\?(?:.*?)(?:v|video_id|story_fbid)=|
+                                story\.php|
+                                flx/warn/
+                            )\?(?:.*?)(?:u|v|video_id|story_fbid)=|
```
dstftw: No.

youtube_dl/extractor/facebook.py
```diff
@@ -39,15 +39,16 @@ class FacebookIE(InfoExtractor):
...
                             [^/]+/videos/(?:[^/]+/)?|
                             [^/]+/posts/|
                             groups/[^/]+/permalink/
                         )|
                     facebook:
                 )
-                (?P<id>[0-9]+)
+                (?P<id>[0-9]+|https://.*)
```
dstftw: No.

youtube_dl/extractor/facebook.py
```diff
@@ -455,6 +456,13 @@ def _real_extract(self, url):
         if info_dict:
             return info_dict
 
+        # In case the url is a redirect and contains an external link.
+        if video_id.startswith("https://"):
+            return {
+                        '_type': 'url',
+                        'url': video_id,
+                    }
```
dstftw: Must be separate extractor.

adi928: So a separate new extractor with a regex for the Facebook 'warning' URL? Any suggestions for the name for this new extractor? I was going to go with facebookWarning extractor.

Also, in the new extractor, won't we be ultimately passing the redirected URL to the generic.py, as the code is doing here?

dstftw: Name essentially does not matter. `FacebookExternalLinkIE` is a reasonable name. "The code here" does not pass to the generic extractor in the first place.

adi928: I'm sorry, I meant 'common.py'. Then would that be fine to pass the link to common.py for download?

dstftw: It's already fine. More idiomatic way is to use `url_result`.

adi928: Created another PR (#19116) with the correct branch as per the developer instructions.

PR #19125, opened 2019-02-03
[pornhd] added like count to info dict

JChris246:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

### Added like count to info dict

### DIFF
--- a/youtube_dl/extractor/pornhd.py
+++ b/youtube_dl/extractor/pornhd.py
@@ -86,7 +86,7 @@ class PornHdIE(InfoExtractor):
             'thumbnail', fatal=False, group='url')
 
         like_count = int_or_none(self._search_regex(
-            r'class="save-count">(\d+)<', webpage, 'like_count', fatal=False))
+            r'save-count[\'"]>(\d+)', webpage, 'like count', fatal=False))
 
         return {
             'id': video_id,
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/pornhd.py
```diff
@@ -85,6 +85,9 @@ def _real_extract(self, url):
             r"poster'?\s*:\s*([\"'])(?P<url>(?:(?!\1).)+)\1", webpage,
             'thumbnail', fatal=False, group='url')
 
+        like_count = int_or_none(self._search_regex(
+            r'class="save-count">(\d+)<', webpage, 'like_count', fatal=False))
```
dstftw: 1. Relaxl regex.
2. Should be "view count".

youtube_dl/extractor/pornhd.py
```diff
@@ -94,4 +97,5 @@ def _real_extract(self, url):
             'view_count': view_count,
             'formats': formats,
             'age_limit': 18,
+            'like_count': like_count,
```
dstftw: Move right after view_count.

youtube_dl/extractor/pornhd.py
```diff
@@ -85,13 +85,17 @@ def _real_extract(self, url):
...
+            r'save-count[\'"]>(\d+)', webpage, 'like count', fatal=False))
+
         return {
             'id': video_id,
             'display_id': display_id,
             'title': title,
             'description': description,
             'thumbnail': thumbnail,
             'view_count': view_count,
+            'like_count': like_count,
```
dstftw: Add to tests.

PR #19145, opened 2019-02-05
[SpankBangPlaylist] Add new extractor

JChris246:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

### [SpankBangPlaylist] New extractor 

New extractor to pull spankbang playlists
Adjusted spankbang regex to recognize single playlist videos

### DIFF
--- a/youtube_dl/extractor/spankbang.py
+++ b/youtube_dl/extractor/spankbang.py
@@ -8,11 +8,12 @@ from ..utils import (
     parse_duration,
     parse_resolution,
     str_to_int,
+    urljoin,
 )
 
 
 class SpankBangIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:(?:www|m|[a-z]{2})\.)?spankbang\.com/(?P<id>[\da-z-]+)/(?:video|playlist)'
+    _VALID_URL = r'https?://(?:(?:www|m|[a-z]{2})\.)?spankbang\.com/(?P<id>[\da-z]+)/video'
     _TESTS = [{
         'url': 'http://spankbang.com/3vvn/video/fantasy+solo',
         'md5': '1cc433e1d6aa14bc376535b8679302f7',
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/spankbang.py
```diff
@@ -12,7 +12,7 @@
 
 
 class SpankBangIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:(?:www|m|[a-z]{2})\.)?spankbang\.com/(?P<id>[\da-z]+)/video'
+    _VALID_URL = r'https?://(?:(?:www|m|[a-z]{2})\.)?spankbang\.com/(?P<id>[\da-z-]+)/(?:video|playlist)'
```
dstftw: This should not match playlist URLs.

youtube_dl/extractor/spankbang.py
```diff
@@ -94,3 +94,34 @@ def _real_extract(self, url):
...
+            'id': 'ug0k',
+            'title': 'Big Ass Titties playlist',
+        },
+        'playlist_mincount': 2,
+    }
+
+    def _extract_entries(self, webpage):
+        return [
+            self.url_result(
+                'http://www.%s/%s' % ('spankbang.com', video_url),
```
dstftw: `urljoin` or source `url`.

youtube_dl/extractor/spankbang.py
```diff
@@ -94,3 +94,34 @@ def _real_extract(self, url):
...
+                SpankBangIE.ie_key())
+            for video_url in re.findall(
+                r'href="/?([\da-z-]+/playlist/[^"]+)', webpage)
```
dstftw: 1. This is too broad, regex should be restricted by playlist id.
2. This captures duplicates.

youtube_dl/extractor/spankbang.py
```diff
@@ -94,3 +94,34 @@ def _real_extract(self, url):
...
+        ]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        playlist_id = mobj.group('id')
+        webpage = self._download_webpage(url, playlist_id)
+
+        entries = self._extract_entries(webpage)
+        title = self._search_regex(r'<h1>(.+)</h1>', webpage, 'playlist_title')
```
dstftw: Playlist title should not be fatal.

youtube_dl/extractor/spankbang.py
```diff
@@ -94,3 +94,34 @@ def _real_extract(self, url):
...
+            self.url_result(
+                'http://www.%s/%s' % ('spankbang.com', video_url),
+                SpankBangIE.ie_key())
+            for video_url in re.findall(
+                r'href="/?([\da-z-]+/playlist/[^"]+)', webpage)
+        ]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        playlist_id = mobj.group('id')
```
dstftw: `_match_id`.

youtube_dl/extractor/spankbang.py
```diff
@@ -94,3 +95,40 @@ def _real_extract(self, url):
...
+        video_items = re.findall(r'<div[^>]+class=[\'"].*?video-item[^>]*>\s*(.+?)>', webpage)
+
+        entries = []
+        if video_items:
+            for div in video_items:
+                page_url = self._search_regex(
+                    r'href="/?(' + id + '-[\da-z]+/playlist/[^"]+)', div, 'page url', default=None)
+
+                if page_url:
+                    page = self._download_webpage(urljoin('http://spankbang.com', page_url), id)
```
dstftw: No. It's a job if the video extractor.

youtube_dl/extractor/spankbang.py
```diff
@@ -94,3 +95,40 @@ def _real_extract(self, url):
...
+    }
+
+    def _extract_entries(self, webpage, id):
+        video_items = re.findall(r'<div[^>]+class=[\'"].*?video-item[^>]*>\s*(.+?)>', webpage)
+
+        entries = []
+        if video_items:
+            for div in video_items:
+                page_url = self._search_regex(
+                    r'href="/?(' + id + '-[\da-z]+/playlist/[^"]+)', div, 'page url', default=None)
```
dstftw: Don't shadow built-in names.

youtube_dl/extractor/spankbang.py
```diff
@@ -94,3 +95,40 @@ def _real_extract(self, url):
...
+            'title': 'Big Ass Titties playlist',
+        },
+        'playlist_mincount': 2,
+    }
+
+    def _extract_entries(self, webpage, id):
+        video_items = re.findall(r'<div[^>]+class=[\'"].*?video-item[^>]*>\s*(.+?)>', webpage)
+
+        entries = []
+        if video_items:
```
dstftw: This check is pointless.

youtube_dl/extractor/spankbang.py
JChris246: This is needed to get the canonical url for the single video extractor [SpankBang] to use

dstftw: Again: video URLs are already available on playlist page.

JChris246: these urls are in the for /playlist_id-playlist_item_id/playlist/playlistname
even if you replace playlist with video and only use the playlist_item_id it does point to the the actual video url

JChris246: *does not

dstftw: I'm not talking about **these** URLs.

JChris246: Where are they

PR #19156, opened 2019-02-06
Added option to force recode a video if it's already in the target format

mdawar:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [x] New feature

---

Added a new option `--force-recode-video` as suggested by @dstftw in pull request #18728 to force processing a video file even if it's already in the target format, this is useful if we're passing `--postprocessor-args` for example to cut the file, in the current version the conversion is skipped with the message:
```
[ffmpeg] Not converting video file FILE_PATH - already is in target format FILE_FORMAT
```

With this option we can force the recoding of the video:
```
youtube-dl URL --recode-video mp4 --format mp4 --postprocessor-args "-ss 10 -to 120 -c copy" --force-recode-video
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,6 +399,8 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
     --recode-video FORMAT            Encode the video to another format if
                                      necessary (currently supported:
                                      mp4|flv|ogg|webm|mkv|avi)
+    --force-recode-video             Force recoding a video if it is already in
+                                     the target format
     --postprocessor-args ARGS        Give these arguments to the postprocessor
     -k, --keep-video                 Keep the video file on disk after the post-
                                      processing; the video is erased by default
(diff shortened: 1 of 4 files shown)


mdawar:
@dstftw since we cannot read and write at the same time I have added a suffix `.recoded` for the output path, for example a `video.mp4` file will be recoded as `video.recoded.mp4`, if this is not the right way to do it I can change this behavior for example rename the old file before recoding.

ealgase:
Using `.temp.mp4` would probably make sense for consistency.

mdawar:
@ealgase does it make sense if a `video.mp4` is converted to `video.temp.mp4` and then `video.mp4` will be deleted if `--keep-video` was not passed and the `.temp` file returned ?

I think renaming the file that will be deleted before recoding is much better, we'll wait for @dstftw 's opinion and see.

mdawar:
@dstftw can you please review this request and let me know your opinion on renaming the original file before forcing a convert ? Thank you.

Issue #18514, opened 2018-12-13
dvtv extractor does not work


honzahosek:
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2018.12.09**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
- [x] Checked that provided video/audio/playlist URLs (if any) are alive and playable in a browser

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)

---

dvtv extractor stopped to work recently. I think the site changed player.
https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/dvtv.py
```
$ youtube-dl -v https://video.aktualne.cz/dvtv/havlikova-zeny-vydelavaji-mene-nez-muzi-a-doma-je-ceka-druha/r~70ab783ef7af11e8bf040cc47ab5f122/
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', 'https://video.aktualne.cz/dvtv/havlikova-zeny-vydelavaji-mene-nez-muzi-a-doma-je-ceka-druha/r~70ab783ef7af11e8bf040cc47ab5f122/']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2018.12.09
[debug] Python version 3.6.4 (CPython) - Linux-4.18.0-2-amd64-x86_64
[debug] exe versions: ffmpeg 4.0.3-1, ffprobe 4.0.3-1
[debug] Proxy map: {}
[dvtv] 70ab783ef7af11e8bf040cc47ab5f122: Downloading webpage
ERROR: Could not find neither video nor playlist; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.

```

---

eduardog3000:
Definitely looks like the extractor needs to be updated.

This looks like a good place to start:

```JavaScript
[...]

PR #19174, opened 2019-02-09
[dvtv] Fix extraction

jfriesse:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

As described in issue #18514 site player was changed and old extractor doesn't work any longer. I've tried to keep change as small as possible and keep all features working (live stream (added test case), playlist, ...)

### DIFF
--- a/youtube_dl/extractor/dvtv.py
+++ b/youtube_dl/extractor/dvtv.py
@@ -10,7 +10,9 @@ from ..utils import (
     int_or_none,
     js_to_json,
     mimetype2ext,
+    try_get,
     unescapeHTML,
+    parse_iso8601,
 )
 
 
@@ -28,6 +30,8 @@ class DVTVIE(InfoExtractor):
             'ext': 'mp4',
             'title': 'Vondra o Českém století: Při pohledu na Havla mi bylo trapně',
             'duration': 1484,
+            'upload_date': '20141217',
+            'timestamp': 1418792400,
         }
     }, {
         'url': 'http://video.aktualne.cz/dvtv/dvtv-16-12-2014-utok-talibanu-boj-o-kliniku-uprchlici/r~973eb3bc854e11e498be002590604f2e/',
@@ -84,6 +88,8 @@ class DVTVIE(InfoExtractor):
             'ext': 'mp4',
             'title': 'Zeman si jen léčí mindráky, Sobotku nenávidí a Babiš se mu teď hodí, tvrdí Kmenta',
             'duration': 1103,
+            'upload_date': '20170511',
+            'timestamp': 1494514200,
         },
         'params': {
             'skip_download': True,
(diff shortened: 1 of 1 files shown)


BKPepe:
@dstftw, I see you're doing a really good job for youtube-dl and also you reviewed a lot pull request in this repository. May I ask you to review this pull request as well? Would be really cool to have it in the next version of youtube-dl!

jfriesse:
@dstftw: Because it looks like you are doing most of the reviews. It's more than month since I've sent this PR. I would like to ask if there is something I can do to get it merged?

jfriesse:
@remitamine: It looks like you have successfully merged many commits lately, would you mind to review this PR?

youtube_dl/extractor/dvtv.py
```diff
@@ -145,32 +161,31 @@ def _real_extract(self, url):
...
         # playlist
         items = re.findall(
-            r"(?s)BBX\.context\.assets\['[0-9a-f]{32}'\]\.push\(({.+?})\);",
+            r"(?s)playlist\.push\(({.+?})\);",
             webpage)
-        if not items:
-            items = re.findall(r'(?s)var\s+asset\s*=\s*({.+?});\n', webpage)
 
         if items:
+            timestamp = parse_iso8601(self._html_search_meta(
```
remitamine: the timestamp is present in both playlist and single video pages.

youtube_dl/extractor/dvtv.py
```diff
@@ -91,43 +92,58 @@ class DVTVIE(InfoExtractor):
...
-                formats.append({
-                    'url': video_url,
-                    'format_id': '-'.join(format_id),
-                    'height': int_or_none(height),
-                })
+
+        tracks = data['tracks']
+
+        for track_type in tracks:
+            for video in tracks[track_type]:
```
remitamine: instead of:
```python
tracks = data['tracks']
for track_type in tracks:
    for video in tracks[track_type]:
```
you can just do something like:
```python
for tracks in data.get('tracks', {}).values():
    for video in tracks:
```

youtube_dl/extractor/dvtv.py
```diff
@@ -145,32 +161,31 @@ def _real_extract(self, url):
...
+                'entries': [self._parse_video_metadata(i, video_id, timestamp) for i in items]
             }
 
+        # single video
+        item = self._search_regex(
+            r'(?s)BBXPlayer.setup\((.+?)\);',
+            webpage, 'video', default=None)
+
+        # remove function calls (ex. htmldeentitize)
+        item_without_funcs = re.sub(r'\w+?\((.+)\)', r'\1', item)
```
remitamine: will break if `item` is `None`.

remitamine: should be handled someway in `js_to_json`, at least add a `TODO` to indicate that this should be fixed in a general way.

jfriesse:
@remitamine: First of all huge thank you for the review!

I've force-pushed fixed commit where all problems which you've found should be fixed.  Commit now also contains `closes` keyword with related issue.

Sadly the test which I've added (live stream) no longer fully works because video content is no longer accessible, but I think it still make sense to have it there.

remitamine:
Thanks, Do you want to be credited as one of the authors of youtube-dl? If so under what name?

jfriesse:
Thank you for review and merging. I think that one small change does not credit me as an youtube-dl author, but thanks a lot for an offer.

PR #19253, opened 2019-02-17
[youtube] Duplicate upload_date to release_date

Freso:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Currently if you use `%(release_date)s` in file naming templates you will get an "NA" from YouTube downloads, because YouTube doesn't distinguish between "release" and "upload" and only provides "upload" date. This can be annoying if you're also using `youtube-dl` for other sites that do distinguish.

This PR simply duplicate the found `upload_date` into `release_date` so both variables can be used for YouTube downloads.

It also happens to fix https://github.com/rg3/youtube-dl/issues/18704 even though that was marked as invalid… But I hope this PR will get accepted anyway! :bowing_man:

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -2025,6 +2025,10 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
                 video_webpage, 'upload date', default=None)
         upload_date = unified_strdate(upload_date)
 
+        # release date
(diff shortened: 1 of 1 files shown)


dstftw:
Again: youtube does not provide `release_date`. `upload_date` is not always == `release_date`.

PR #20219, opened 2019-03-02
[utils] strip #HttpOnly_ prefix from cookies files

remitamine:

- [ ] Bug fix
- [X] Improvement
- [ ] New extractor
- [ ] New feature

---

the code is a modification of [the original code for `load` method](https://github.com/python/cpython/blob/master/Lib/http/cookiejar.py#L1774-L1781), there were multiple similar changes that has been based on the original code of python in `compat.py`.
the code simply removes the `#HttpOnly_` prefix from begining of every line by using an in memory file(`StringIO`), it's the simplest solution that i can think of without using any temporary files or changing the `_really_load` method.

### DIFF
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -1141,6 +1141,8 @@ class YoutubeDLHTTPSHandler(compat_urllib_request.HTTPSHandler):
 
 
 class YoutubeDLCookieJar(compat_cookiejar.MozillaCookieJar):
+    _HTTPONLY_PREFIX = '#HttpOnly_'
+
     def save(self, filename=None, ignore_discard=False, ignore_expires=False):
         # Store session cookies with `expires` set to 0 instead of an empty
         # string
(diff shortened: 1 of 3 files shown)


dstftw:
1. Core tests are broken.
2. This changes should have a test as well.

youtube_dl/utils.py
```diff
@@ -1150,7 +1150,19 @@ def save(self, filename=None, ignore_discard=False, ignore_expires=False):
...
+        if filename is None:
+            if self.filename is not None:
+                filename = self.filename
+            else:
+                raise ValueError(compat_cookiejar.MISSING_FILENAME_TEXT)
+
+        cf = io.StringIO()
+        with open(filename) as f:
+            for line in f:
+                cf.write(remove_start(line, '#HttpOnly_'))
```
dstftw: There is no point to use `remove_start` since line is always a string.

remitamine:
> Core tests are broken.
> There is no point to use remove_start since line is always a string.

the core tests now pass after removing the use of `remove_start`.

> These changes should have a test as well.

done.

youtube_dl/utils.py
```diff
@@ -1150,7 +1150,21 @@ def save(self, filename=None, ignore_discard=False, ignore_expires=False):
...
+            if self.filename is not None:
+                filename = self.filename
+            else:
+                raise ValueError(compat_cookiejar.MISSING_FILENAME_TEXT)
+
+        cf = io.StringIO()
+        with open(filename) as f:
+            for line in f:
+                if line.startswith('#HttpOnly_'):
+                    line = line[10:]
```
dstftw: This is error prone, `len` of the actual string should be used instead.

remitamine: i don't see the problem here, the length of `#HttpOnly_`  is constant and won't change, besides the test is there to make sure that breakage in this part of code will be detected.

dstftw: >besides the test is there to make sure that breakage in this part of code will be detected

That's a doubtful argument considering broken core tests at the beginning of this PR.

The length of this string is const until one decides to refactor here something. Using 10 is a variation of code duplication since the length is already implicitly defined in the string literal itself. Also using 10 indicates no relation to the string literal so that one unfamiliar with code who decides to refactor it may forgot to change the number and may be unaware of the tests at all.

remitamine: ok, changed the code to use `len`.

Issue #14031, opened 2017-08-26
Can't automatically over write existing downloaded file


section83:
- [X] I've **verified** and **I assure** that I'm running youtube-dl **2017.08.23**

### Before submitting an *issue* make sure you have:
- [X] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [X] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [ ] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [X] Question
- [ ] Other

---

This is the full verbose output:

```
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'--restrict-filenames', u'--verbose', u'-o', u'%(title)s.%(ext)s', u'http://iview.abc.net.au/programs/amateur-hour-spotlight-shorts/ZW0776A006S00']
[debug] Encodings: locale US-ASCII, fs utf-8, out None, pref US-ASCII
[debug] youtube-dl version 2017.08.23
[debug] Python version 2.7.10 - Darwin-16.7.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg 3.3.2-tessus, ffprobe 3.3.2-tessus
[debug] Proxy map: {}
[abc.net.au:iview] ZW0776A006S00: Downloading webpage
[abc.net.au:iview] ZW0776A006S00: Downloading f4m manifest
[abc.net.au:iview] ZW0776A006S00: Downloading m3u8 information
[abc.net.au:iview] ZW0776A006S00: Downloading f4m manifest
[abc.net.au:iview] ZW0776A006S00: Downloading m3u8 information
[debug] Default format spec: bestvideo+bestaudio/best
[debug] Invoking downloader on u'http://iviewum-vh.akamaihd.net/i/playback/_definst_/_video/artsa/arts_amateur_01_06_20170814140030_,1500000,1000000,650000,500000,220000,.mp4.csmil/index_0_av.m3u8'
[download] Ash_Flanders.mp4 has already been downloaded

[download] 100% of 23.05MiB
[...]

PR #20405, opened 2019-03-18
Overwrite merged video without no-overwrites flag

alxnull:

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Currently, if a file with the same name already exists, the download of a video ends with this message, saying it's been downloaded already (be aware that this behaviour _does not_ change by adding the `--no-continue` flag): 

> [youtube] Jo_-KoBiBG0: Downloading webpage
[youtube] Jo_-KoBiBG0: Downloading video info webpage
[download] Anthem of Europe-Jo_-KoBiBG0.webm has already been downloaded and merged

However, to my understanding, _without_ the `--no-overwrites` flag which explicitly enforces such behaviour, all downloaded files _should_ be overwritten (thus redownloaded). That's already the case e.g. for info jsons, video descriptions or subtitles but not for video files themselves.

If this is an unwanted behaviour, I could add an optional flag (`--force-overwrites`) to implement it solely for video files. I currently did not do that because I think it would make the available options more confusing.

(This issue was raised e.g. in #14031 a while ago)


alxnull:
@dstftw or @remitamine as it seems you're the main developers on this project, I'd be really glad to receive some feedback from you on this.

dstftw:
I definitely won't accept current solution as it breaks the default behavior. Media files are considered big files and should not be overwritten until explicitly told to do so. I also won't accept `--force-overwrites`.

What I probably will accept is extending overwrites options to:
* No overwrites option defined - current default behavior, don't overwrite media files
* `--no-overwrites` - no overwrites for any file
* `--yes-overwrites` - overwrites for any file

Everything should be covered with tests.

DaveF63:
Until this essential command is added, could someone inform me: 
a) Where the record of downloaded files is stored.
b) Why does this appear to only occur with merged files. If I use `-f best` it overwrites quite happily.

alxnull:
As suggested by @dstftw, I added a new `--yes-overwrites` option, leaving the default behaviour untouched. As the available options and their consequences can get quite confusing now, here's a quick overview how the overwrite-related options behave in some cases (tested w. 2019.09.28 using a Youtube video):

| Additional options | _default_ | `--no-overwrites` | `--no-continue` | `--yes-overwrites` |
|----|----|----|----|----|
| _no options_ | ❌ | ❌ | ❌ | ✔️ |
| `-f best` | ❌ | ❌ | ✔️/ ❌ | ✔️ |
| `-x`/ `--recode-video mp4` | ✔️ | ✔️ | ✔️ | ✔️ |
| Metadata (e.g. `--write-thumbnail`) | ✔️ | ❌ | ✔️ | ✔️ |

(❌: no overwrite; ✔️: overwrite)

section83:
Can this change be implemented ?  It would greatly help.  Thanks.

altimumdelta:
I was also looking for such a feature too right now, I downloaded 720p, but I decided to use 1080p quality with another command, but I have to go navigate to the disk to delete it manually.

I do understand that it's better to be safe and not overwrite by accident, but in such a case where I know I don't need the old version and I'm really sure then why not have such an option. Thanks

section83:
> I do understand that it's better to be safe and not overwrite by accident, but in such a case where I know I don't need the old version and I'm really sure then why not have such an option. 

Those of us who use YTDL in our projects just have to build our own code to find duplicates and ask the user what to do. Challenge for me has been working out exactly what YTDL does.

alxnull:
> No overwrites option defined - current default behavior, don't overwrite media files
    --no-overwrites - no overwrites for any file
    --yes-overwrites - overwrites for any file

@dstftw Implemented as you suggested and added a test. Please review. Thanks!

subirax:
I just tested this and sadly, it doesn't work -- "no such option" with the latest update, 2020.09.20.
Here is the complete output of the attempt.  What am I doing wrong?

============
youtube-dl --yes-overwrites https://www.dailymotion.com/video/x17jdp7
Usage: youtube-dl [OPTIONS] URL [URL...]

youtube-dl: error: no such option: --yes-overwrites
==============

I hope this is helpful.
Thanks!

solbu:
@subirax 
I just tested the same, and it works.
Did you apply the patch?

Remember this Pull Request it isn't merged into the code yet, so you need to apply the patch before compiling/testing.

Arnauw:
Hello I have last version on "--yes-overwrites" doesn't work, is it possible to fix it or something ?

altimumdelta:
@Arnauw As stated above, this is still open, it's not merged yet, it's not part of youtube-dl yet, you would need to compile it with this patch yourself in the meantime while it gets all the tests/checks done to get in.

Arnauw:
@altimumdelta Ok thanks ;)

darbdenral:
The exe for Windows downloaded today youtube-dl version 2021.06.06 gives the same error..
youtube-dl.exe: error: no such option: --yes-overwrites

Serene-Arc:
Is this PR still active? Was considering coding a PR to do the same thing but this one looks like it would fulfill my use case.

PR #20479, opened 2019-03-26
[vlive:playlist] Fix failing test

biwubo:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Noticed that the `VLivePlaylistIE` test was failing. Looks like the playlist used in the test no longer exists so I've replaced it with a valid one.

### DIFF
--- a/youtube_dl/extractor/vlive.py
+++ b/youtube_dl/extractor/vlive.py
@@ -276,12 +276,12 @@ class VLivePlaylistIE(InfoExtractor):
     IE_NAME = 'vlive:playlist'
     _VALID_URL = r'https?://(?:(?:www|m)\.)?vlive\.tv/video/(?P<video_id>[0-9]+)/playlist/(?P<id>[0-9]+)'
     _TEST = {
-        'url': 'http://www.vlive.tv/video/22867/playlist/22912',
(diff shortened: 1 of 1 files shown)


dstftw:
You **must fix code** instead of hiding this bug behind working playlist URL.

biwubo:
The playlist in the current test doesn't seem to exist anymore, I can't see any reference to a playlist in the HTML (including the embedded JavaScript). There is just a suggestions box on the page where the playlist would otherwise be displayed. I have attached images of both below to show the difference.

www vlive tv_video_22867_playlist_22912
www vlive tv_video_117956_playlist_117963

dstftw:
It does not matter whether playlist exists or not. http://www.vlive.tv/video/22867/playlist/22912 opens a video in browser so the code must be fixed to be able to download from this URL.

biwubo:
Fair enough, but I still think if we're only limited to one test it should be a valid playlist to actually test the playlist extractor, I can add logic to redirect to the regular extractor if a playlist cannot be found, though.

dstftw:
We are not limited to one test.
Again: you can add another test for a valid playlist but you **must not** hide this bug by removing current test.

PR #20492, opened 2019-03-27
Update format example with parenthesis in readme.md, to prevent "audio not downloaded and merged" behavior

29988122:

---

- [x ] At least skimmed through [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) sections

- [ ] Bug fix
- [x ] Improvement
- [ ] New extractor
- [ ] New feature

---

When downloading with the intent of merging video+audio together, the lack of parehthesis would make the script fail to download the audio part.

Example:
```
youtube-dl -f 248/137/136+251/140
```
Whenever 248 and 251 do not exist, youtube-dl will only download 137, completely ignores 140, thus generated a file with video only.

Change it to (  (248/137/136)+(251/140)  ) solved the issue. A file with 137+140 can be downloaded and merged correctly.

Not so sure if I should change the code directly, so I changed readme.md first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -703,6 +703,9 @@ $ youtube-dl -f 'bestvideo[height<=480]+bestaudio/best[height<=480]'
 # Download best video only format but no bigger than 50 MB
 $ youtube-dl -f 'best[filesize<50M]'
 
+# Utilizing parenthesis for the + sign, in order to prevent audio from not being downloaded and merged
+$ youtube-dl -f ((248/247/137/136)+(251/140/171))
+
 # Download best format available via direct link over HTTP/HTTPS protocol
 $ youtube-dl -f '(bestvideo+bestaudio/best)[protocol^=http]'
 


dstftw:
1. This is expected behavior since `/` operator takes precedence over others.
2. Any example added must be reasonable, it should clearly demonstrate and explain some use case. This example does not.
3. Using concrete formats codes is discouraged and won't be accepted as example.

29988122:
Thank you for the quick comment, your every point is reasonable and fair.

Would you suggest me:
1. Remove the example, but add another paragraph in the format select section about logic operator sequences/priorities. 
2. Remove the example, but add another paragraph about merging behavior, mention the usage of parenthesis in it.
3. None of the above

I'd like to hear your advice, since the current readme in format selection can be improved to prevent confusion.

Thank you again, @dstftw .

dstftw:
There is already a paragraph about grouping with parentheses in FAQ that may be completed with mention of `,`, `/` and `+` operators precedence and several more examples demonstrating how to vary the behavior with different parentheses arrangement.

PR #20496, opened 2019-03-27
Added optional www prefix to pattern in weibo extractor

RexYuan:
It’s a tiny fix so I thought I’d just make a PR instead of going through filing an issue and all that.

### DIFF
--- a/youtube_dl/extractor/weibo.py
+++ b/youtube_dl/extractor/weibo.py
@@ -19,7 +19,7 @@ from ..utils import (
 
 
 class WeiboIE(InfoExtractor):
-    _VALID_URL = r'https?://(www\.)?weibo\.com/[0-9]+/(?P<id>[a-zA-Z0-9]+)'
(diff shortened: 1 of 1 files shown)


dstftw:
Do not capture groups you don't use.

RexYuan:
Weibo redirects you to the URL with the www record. For the test example ```https://weibo.com/6275294458/Fp6RGfbff?type=comment'```, Weibo redirects you to ```https://www.weibo.com/6275294458/Fp6RGfbff?type=comment%2527```. So if you try to copy and paste the URL from the browser and run youtube-dl with it directly, you'll get an Unsupported URL error. Sorry if I wasn't clear enough.

dstftw:
Did you even read my comment?

RexYuan:
I'm not sure if I understand. That was my response to your comment.

RexYuan:
Sorry I misunderstood your comment. Took another look at the file and fixed it now.

PR #20535, opened 2019-04-02
Initial version of Floatplane.com extractor.

Lamieur:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This is a very basic extractor for Floatplane.com.

It requires a Floatplane account and subscribing to a channel.

It requires a single cookie to be exported from a browser to work.

When used without subscription / with a cookie missing, a 403 is given, so a
simple error message was added, as some other extractors do.

Edit: I chose to only support 1080p resolution to speed up the extractor and reduce number of HTTP requests (URL for each quality has to be pulled separately). Re-enabling lower resolutions can be done by uncommenting a single line.

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -500,6 +500,12 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
 
         # RTMP (unnamed)
         '_rtmp': {'protocol': 'rtmp'},
+
+        # av01 video only formats sometimes served with "unknown" codecs
+        '394': {'acodec': 'none', 'vcodec': 'av01.0.05M.08'},
+        '395': {'acodec': 'none', 'vcodec': 'av01.0.05M.08'},
+        '396': {'acodec': 'none', 'vcodec': 'av01.0.05M.08'},
+        '397': {'acodec': 'none', 'vcodec': 'av01.0.05M.08'},
     }
     _SUBTITLE_FORMATS = ('srv1', 'srv2', 'srv3', 'ttml', 'vtt')
 


dstftw:
1. Read coding convetions and fix code.
2. All formats must be extracted.
3. You must provide example URLs.
4. You must provide account credentials.

Lamieur:
1. I thought I did? I made it not pass 80 columns and pass flake8.

2. No problem, the code is already there - but it does slow the extraction in the current form, which is why I've commented it.

Is there a way to tell ytdl a format list WITHOUT providing an URL for each one, and ONLY extracting the URL IF a user passes -f? I don't think there is, since user can say "-f bestvideo" and YTDL will decide based on the list that already includes URL-s?

If not, it will just be a bit slower, but I'll do it, no problem.

3. Is it acceptable to included test cases that can be tested only by subscribers? I've seen other platforms have a kind of a test video that is publicly accessible. As far as I can tell, Floatplane has no "speed test" video - everything is subscriber only.

If that's acceptable, I can include a URL from LTT channel which most people using Floatplane are subbed to, but no automated test (on a public test server) will pass. I wasn't sure how to solve that.

4. I don't understand. I took a good look on some other extractors for paywalled services and they all simply say to set cookies. I've seen no examples of this in the other extractors as to how to pass account credentials...

There's the usual -u/-p for YouTube, and --ap-username/password for some site. All other extractors simply say to use cookies. Can I add arguments like --fp-username/password or...?

gary-kim:
I think @dstftw is referring to using `info_json.get('thumbnail').get('path')` rather than `info_json['thumbnail']['path']` since the thumbnail is optional.

I'd also like to add support for downloading playlists once this gets merged.

PR #20598, opened 2019-04-07
fix(aol): fix the aol downloader

jrmlhermitte:

---

- [x ] At least skimmed through [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) sections

I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)

Bug Fix

---

The aol extractor is failing because the video id for the details API is not right (it likely has changed).
Here is a proposed fix. Unfortunately, to get the video url, we have to download the page and find it from within the page.

The id looks like a hash. I have tried just hashing different pieces of the url (manually) and didn't find a match. I might try to a few extra transformations, but it's likely that this hash is computed from other information. If we did find a way to transform from URL -> hash we would avoid the downloading and xpath selector step which would make this cleaner and require one less request.

I will try to clean this up a little more later but I do have some more higher level questions:
- For the page download request, am I running this in the right place, or is there a util function I should be calling?

### DIFF
--- a/youtube_dl/extractor/aol.py
+++ b/youtube_dl/extractor/aol.py
@@ -9,6 +9,9 @@ from ..utils import (
     int_or_none,
     url_or_none,
 )
+from urllib.parse import urlparse
+from urllib.request import urlopen
+import lxml.html
 
 
 class AolIE(InfoExtractor):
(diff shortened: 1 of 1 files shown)


dstftw:
1. All aol tests pass without any issue. Clearly describe what are you trying to fix in the first place, provide example URLs and add tests.
2. Read tutorial and coding conventions.

PR #20632, opened 2019-04-10
[iai] Add extractor

jwilk:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature


youtube_dl/extractor/iai.py
```diff
@@ -0,0 +1,53 @@
...
+        data = self._search_regex(r'\bwindow\.JS_VARS\s*=(.*)', webpage, 'JS_VARS')
+        data = self._parse_json(data, video_id)
+        formats = []
+        for widget in data['Widgets']['widgets']:
+            if widget.get('Type') != 'VideoPlayerWidget':
+                continue
+            sprops = widget['Sprops']
+            for format_id in 'sd', 'hd':
+                quality = 1 if format_id == 'hd' else 0
+                key = ('HighRes' if quality > 0 else '') + 'StreamingUrl'
```
dstftw: 34-35 can be easily moved into `for`.

youtube_dl/extractor/iai.py
```diff
@@ -0,0 +1,53 @@
...
+            'description': 'md5:1dbbaabf0609bcb45220b2b7d9ad063f',
+            'upload_date': '20190404',
+            'timestamp': 1554379576,
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        data = self._search_regex(r'\bwindow\.JS_VARS\s*=(.*)', webpage, 'JS_VARS')
```
dstftw: Capture dict if you expect dict.

youtube_dl/extractor/iai.py
```diff
@@ -0,0 +1,53 @@
...
+        data = self._parse_json(data, video_id)
+        formats = []
+        for widget in data['Widgets']['widgets']:
+            if widget.get('Type') != 'VideoPlayerWidget':
+                continue
+            sprops = widget['Sprops']
+            for format_id in 'sd', 'hd':
+                quality = 1 if format_id == 'hd' else 0
+                key = ('HighRes' if quality > 0 else '') + 'StreamingUrl'
+                url = sprops.get(key)
```
dstftw: 1. Don't shadow outer names.
2. `url_or_none`.

youtube_dl/extractor/iai.py
```diff
@@ -0,0 +1,53 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class IaiIE(InfoExtractor):
+    IE_NAME = 'IAI'
+    _VALID_URL = r'https://iai.tv/video/(?P<id>[^/]+)'
```
dstftw: Will capture incorrect id for https://iai.tv/video/darkness-authority-and-dreams#blah.

youtube_dl/extractor/iai.py
```diff
@@ -0,0 +1,53 @@
...
+                    continue
+                formats.append({
+                    'url': url,
+                    'format_id': format_id,
+                    'quality': quality,
+                })
+            break
+        self._sort_formats(formats)
+        info = self._search_json_ld(webpage, video_id, default={})
+        info['id'] = video_id
```
dstftw: Real id is in widget dict.

youtube_dl/extractor/iai.py
jwilk: Yes, but why?

youtube_dl/extractor/iai.py
jwilk: >     2. `url_or_none`.

I don't see the point.

youtube_dl/extractor/iai.py
dstftw: DRY.

youtube_dl/extractor/iai.py
dstftw: url is not validated and may be not a URL or not a string at all.

youtube_dl/extractor/iai.py
jwilk: I don't see anything wrong with this code.

youtube_dl/extractor/iai.py
jwilk: I don't think trying to recover from such fundamental breakage is a good use anybody's time.

jwilk:
I've lost interest in youtube-dl.

PR #20634, opened 2019-04-10
[giphy] Add extractor

jwilk:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

There are more formats available, but getting hands on them would require extra code.
I have have personally no use for the other formats, so I chose not do the work most likely nobody needs.

### DIFF
--- a/youtube_dl/extractor/giphy.py
+++ b/youtube_dl/extractor/giphy.py
@@ -1,0 +1,62 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from ..utils import (
+    int_or_none,
+    remove_end,
+    remove_start,
+)
+
+
+class GiphyIE(InfoExtractor):
+    _VALID_URL = r'https://(?:i|media)\.giphy\.com/media/(\w+)/|https://giphy\.com/gifs/[^/#?]+-(\w+)(?:[#?]|$)'
+    _TESTS = [{
+        'url': 'https://media.giphy.com/media/2rj8VysAig8QE/giphy.gif',
+        'md5': 'faa04a4e3134cb74005b78b28c34cc29',
+        'info_dict': {
+            'id': '2rj8VysAig8QE',
+            'ext': 'mp4',
+            'title': 'Css GIF',
+        }
+    }, {
+        'url': 'https://i.giphy.com/media/oY7zuBsNna7CM/giphy.mp4',
+        'only_matching': True
+    }, {
+        'url': 'https://giphy.com/gifs/quit-playin-MZAvNpYCrFUJi',
+        'only_matching': True
+    }]
+
+    def _real_extract(self, url):
+        [video_id] = filter(None, re.match(self._VALID_URL, url).groups())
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/giphy.py
```diff
@@ -0,0 +1,61 @@
...
+    }, {
+        'url': 'https://i.giphy.com/media/oY7zuBsNna7CM/giphy.mp4',
+        'only_matching': True
+    }, {
+        'url': 'https://giphy.com/gifs/quit-playin-MZAvNpYCrFUJi',
+        'only_matching': True
+    }]
+
+    def _real_extract(self, url):
+        [video_id] = filter(None, re.match(self._VALID_URL, url).groups())
```
dstftw: What's the point of all this mess? This can be easily expressed with a single capture group in regex.

youtube_dl/extractor/giphy.py
```diff
@@ -0,0 +1,61 @@
...
+        page_url = 'https://media.giphy.com/media/{id}/giphy.gif'.format(id=video_id)
```
dstftw: Inline.

youtube_dl/extractor/giphy.py
```diff
@@ -0,0 +1,61 @@
...
+        webpage = self._download_webpage(page_url, video_id)
+        title = self._og_search_title(webpage, fatal=True)
```
dstftw: This is already fatal.

youtube_dl/extractor/giphy.py
```diff
@@ -0,0 +1,61 @@
...
+        title = remove_end(title, ' - Find & Share on GIPHY')
+        formats = []
+        for prop in 'image', 'video':
+            url = self._og_search_property(prop, webpage)
```
dstftw: Should not be fatal.

youtube_dl/extractor/giphy.py
```diff
@@ -0,0 +1,61 @@
...
+            mime_type = self._og_search_property(prop + ':type', webpage)
```
dstftw: Should not be fatal.

youtube_dl/extractor/giphy.py
jwilk: What do you mean?

youtube_dl/extractor/giphy.py
jwilk: Yes, it should. There's no point in recovering from such fundamental breakage.

youtube_dl/extractor/giphy.py
jwilk: You are not being helpful.

youtube_dl/extractor/giphy.py
dstftw: Well, I'm not going to argue. You either fix it as requested or get PR rejected.

youtube_dl/extractor/giphy.py
dstftw: Inline variable at the place of usage.

jwilk:
I've lost interest in youtube-dl.

Issue #20540, opened 2019-04-03
Site support: Changba


AnthonyWC:
### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2019.04.01*. If it's not, read [this FAQ entry](https://github.com/ytdl-org/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2019.04.01**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/ytdl-org/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/ytdl-org/youtube-dl#faq) and [BUGS](https://github.com/ytdl-org/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
- [x] Checked that provided video/audio/playlist URLs (if any) are alive and playable in a browser

### What is the purpose of your *issue*?
- [ ] Bug report (encountered problems with youtube-dl)
- [x] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### If the purpose of this *issue* is a *site support request* please provide all kinds of example URLs support for which should be included (replace following example URLs by **yours**):
- Single video (music): `http://changba.com/s/nZqfbS_vCnieNNjJ7UiEGw?`

Note that **youtube-dl does not support sites dedicated to [copyright infringement](https://github.com/ytdl-org/youtube-dl#can-you-add-support-for-this-anime-video-site-or-site-which-shows-current-movies-for-free)**. In order for site support request to be accepted all provided example URLs should not violate any copyrights.

---

### Description of your *issue*, suggested solution and other information

This is a site that allows anyone to record their own singing of songs (e.g. from Karaoke).

jgalp:
Is anyone working on it? I am a student at the University of Michigan looking to contribute to an open source project for my upper-level software engineering class.

quinlander:
@jgalp and I are working on this issue.

PR #20694, opened 2019-04-17
Add changba.com extractor

rryanq:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Add new extractor for changba.com requested in issue #20540 

### DIFF
--- a/youtube_dl/extractor/changba.py
+++ b/youtube_dl/extractor/changba.py
@@ -30,6 +30,24 @@ class ChangbaIE(InfoExtractor):
             'ext': 'mp3',
             'title': '下雪 ',
         }
+    },
+    {
+        'url': 'http://changba.com/s/CPiNWbAa1qy0po0llqIJbg',
+        'md5': '',
+        'info_dict': {
+            'id': '136918054',
+            'ext': 'mp3',
+            'title': '红豆 ',
+        }
+    },
+    {
+        'url': 'http://changba.com/s/-N00JJ30YruunrER5eBcWw',
+        'md5': 'cd68f8da8d8c69afbb8e4dbbbfa8b277',
+        'info_dict': {
+            'id': '172671761',
+            'ext': 'mp3',
+            'title': '天与地 ',
+        }
     }]
 
     def _real_extract(self, url):
--- a/youtube_dl/extractor/changba.py
+++ b/youtube_dl/extractor/changba.py
@@ -1,0 +1,51 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+import re
+
+class ChangbaIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?changba\.com/s/(?P<id>[0-9A-Za-z-_]+)'
+    _TEST = {
+        'url': 'https://changba.com/s/0GHVw6vyXv9N2FhaFi2WJg',
+        'md5': 'TODO: md5 sum of the first 10241 bytes of the video file (use --test)',
+        'info_dict': {
+            'id': '1152860688',
+            'ext': 'mp4',
+            'title': 'Video title goes here',
+            'thumbnail': r're:^https?://.*\.jpg$',
+            # TODO more properties, either as:
+            # * A value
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/changba.py
```diff
@@ -0,0 +1,78 @@
...
+            'lzscuw',
+            'upscuw',
+            'aliuwmp3',
+            'upuwmp3',
+            'qiniuuwmp3'
+        ]
+
+        try:
+            src_url = self._search_regex(r'var a="([^"]*)', webpage, 'url')
+        except:
```
dstftw: Never use bare except.

youtube_dl/extractor/changba.py
```diff
@@ -0,0 +1,78 @@
...
+        SITE_SUBDOMAINS = [
+            'lzscuw',
+            'upscuw',
+            'aliuwmp3',
+            'upuwmp3',
+            'qiniuuwmp3'
+        ]
+
+        try:
+            src_url = self._search_regex(r'var a="([^"]*)', webpage, 'url')
```
dstftw: Use default. Read coding conventions and fix code.

youtube_dl/extractor/changba.py
```diff
@@ -0,0 +1,78 @@
...
+        except:
+            for subdomain in SITE_SUBDOMAINS:
+                try:
+                    src_url = 'http://{}.changba.com/{}.{}'.format(
+                        subdomain, str(id), ext
+                    )
+                except:
```
dstftw: 1. This does not make any sense at all since there will never be an exception.
2. You must extract media URL not guess it.

youtube_dl/extractor/changba.py
```diff
@@ -0,0 +1,78 @@
...
+            'upuwmp3',
+            'qiniuuwmp3'
+        ]
+
+        try:
+            src_url = self._search_regex(r'var a="([^"]*)', webpage, 'url')
+        except:
+            for subdomain in SITE_SUBDOMAINS:
+                try:
+                    src_url = 'http://{}.changba.com/{}.{}'.format(
```
dstftw: No `{}`.

youtube_dl/extractor/changba.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        id = self._search_regex(r'workid=([0-9]+)', webpage, 'id')
+        title = self._search_regex(
+            r'<div[^>]+class="title"[^>]*>([^<]+)', webpage, 'title'
+        )
+        isvideo = self._search_regex(r'&isvideo=([0-9])', webpage, 'isvideo')
+        ext = 'mp3' if int_or_none(isvideo) == 0 else 'mp4'
```
dstftw: Audio must have proper `vcodec` set.

youtube_dl/extractor/changba.py
```diff
@@ -0,0 +1,78 @@
...
+            'ext': 'mp3',
+            'title': '红豆 ',
+        }
+    }, {
+        'url': 'http://changba.com/s/-N00JJ30YruunrER5eBcWw',
+        'md5': 'cd68f8da8d8c69afbb8e4dbbbfa8b277',
+        'info_dict': {
+            'id': '172671761',
+            'ext': 'mp3',
+            'title': '天与地 ',
```
dstftw: Remove all duplicate tests.

youtube_dl/extractor/changba.py
jgalp: The reason we are creating the media URL is that Changba removes the URL from the metadata on some pages. We've found that the URL we create is still able to download those videos or songs.

dstftw: URL is always available. In case of video it's just hidden behind base64.

jgalp: Oh, I see. Thanks!

rryanq:
@dstftw We have made the requested changes. Please let us know if there are any further changes that need to be made. Thanks!

PR #20845, opened 2019-04-26
Dump video list only

peteruhrig:

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

Youtube-dl's capabilities of retrieving all the videos from a YouTube channel or playlist are excellent, but for very large channels (15k videos), scraping the channel with full downloads on one machine leads to problems, so we would like to distribute the download across multiple machines at various sites. For this, we simply need a list of all the videos in a channel or a playlist. Superficially, the current feature --get-id does this, but in fact it runs the full simulation with lots of requests, which makes it slow and again bears the risk of having the machine banned.
The new option --dump-video-list-only dumps the video list and is best used with -i and -q so the output is a parseable JSON Lines file. To this end, it simply goes through the ie_entries generator and prints every item from it. This has the effect of emptying the generator so that no further requests are done because the video list for processing is 0 elements long.


dstftw:
`--flat-playlist --dump-json`.

peteruhrig:
Thanks for the quick reply and sorry for bothering you with this. I can't believe I missed that option and spent quite some time trying to understand the code and adding a small feature that had been there all along...

PR #20863, opened 2019-04-27
[flake8] Ignore W504 (line break after binary operator)

jwilk:
Both W504 (line break after binary operator) and W503 (line break before binary operator) were implicitly enabled, even though conceptually they are mutually exclusive.

W504 had over 100 hits, while W503 had none, so let's disable the former.


dstftw:
Looks like W504 is now PEP 8 recommended style, so I guess it's better to fix the code and disable it.

PR #20964, opened 2019-05-05
[utils] Configurable SSL cipher list

jmakovicka:
Allows working around the weak ciphers by setting to DEFAULT@SECLEVEL=1

Needed for ceskatelevize.cz with recent Linux distros.

---

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

In latest Debian Linux (unstable), SSL library does not accept small Diffie-Hellman keys by default. Because of this, downloading from ceskatelevize.cz fails with the following error:

ERROR: Unable to download webpage: <urlopen error [SSL: DH_KEY_TOO_SMALL] dh key too small (_ssl.c:1056)> (caused by URLError(SSLError(1, '[SSL: DH_KEY_TOO_SMALL] dh key too small (_ssl.c:1056)')))

This patch allows to work around this by specifying the cipher list manually, eg. `youtube-dl --no-check-certificate --ciphers DEFAULT@SECLEVEL=1 https://www.ceskatelevize.cz/porady/11784224641-king-skate/`

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -354,6 +354,7 @@ class YoutubeDL(object):
         self.params = {
             # Default parameters
             'nocheckcertificate': False,
+            'ciphers': None,
         }
         self.params.update(params)
         self.cache = Cache(self)
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -397,6 +397,7 @@ def _real_main(argv=None):
         'download_archive': download_archive_fn,
         'cookiefile': opts.cookiefile,
         'nocheckcertificate': opts.no_check_certificate,
+        'ciphers': opts.ciphers,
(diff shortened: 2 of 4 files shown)


kparal:
Any hope that the developers could merge this?

Implementation wise, I think it would be better if the user didn't need to know how to specify `--ciphers`. Just using `--no-check-certificate` or `--prefer-insecure` should do the right thing and allow insecure ciphers. It should be easy for a user to download videos from websites using old certificates, not difficult.

cotwitch:
Parameter `--legacy-server-connect` was enough to resolve my issue (urlopen error [SSL: DH_KEY_TOO_SMALL] dh key too small) when downloading from weak host.

dirkf:
>Parameter `--legacy-server-connect` ...

... of _yt-dlp_, not yet implemented here.

PR #20738, opened 2019-04-21
[verystream] Add new extractor (is just a copy of Openload with little changes)

etmatrix:
### Before submitting a *pull request* make sure you have:
- [X] At least skimmed through [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) sections
- [X] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [X] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [X] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [ ] Improvement
- [X ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information
It's just a copy of Openload with some adjustment for new site.

Issue #20701, opened 2019-04-17
Site Support Request: verystream.com


ealgase:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like this: `[x]`)
- Use the *Preview* tab to see what your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2019.04.17*. If it's not, read [this FAQ entry](https://github.com/ytdl-org/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2019.04.17**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/ytdl-org/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/ytdl-org/youtube-dl#faq) and [BUGS](https://github.com/ytdl-org/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
- [x] Checked that provided video/audio/playlist URLs (if any) are alive and playable in a browser

### What is the purpose of your *issue*?
- [ ] Bug report (encountered problems with youtube-dl)
- [x] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add the `-v` flag to **your command line** you run youtube-dl with (`youtube-dl -v <your command line>`), copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):

```
(bionic)ealgase@localhost:~/Downloads$ youtube-dl -v https://verystream.com/stream/c1GWQ9ngBBx/Big_Buck_Bunny.mp4
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', 'https://verystream.com/stream/c1GWQ9ngBBx/Big_Buck_Bunny.mp4']
[...]

PR #20967, opened 2019-05-05
[openload] add support for verystream.com

ealgase:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This pull request adds support for verystream.com into the Openload extractor, closing #20701 (and making #20738 obsolete).

### DIFF
--- a/youtube_dl/extractor/openload.py
+++ b/youtube_dl/extractor/openload.py
@@ -254,7 +254,10 @@ class OpenloadIE(InfoExtractor):
                         (?:f|embed)/
                         (?P<id>[a-zA-Z0-9-_]+)
                     ''' % _DOMAINS
-
+    _EMBED_WORD = 'embed'
+    _STREAM_WORD = 'f'
+    _REDIR_WORD = 'stream'
+    _URL_IDS = ('streamurl', 'streamuri', 'streamurj')
     _TESTS = [{
         'url': 'https://openload.co/f/kUEfGclsU9o',
         'md5': 'bf1c059b004ebc7a256f89408e65c36e',
@@ -1954,6 +1957,11 @@ class OpenloadIE(InfoExtractor):
             r'<iframe[^>]+src=["\']((?:https?://)?%s/embed/[a-zA-Z0-9-_]+)'
             % OpenloadIE._DOMAINS, webpage)
 
+    def _extract_decrypted_page(self, page_url, webpage, video_id, headers):
+        phantom = PhantomJSwrapper(self, required_version='2.0')
+        webpage, _ = phantom.get(page_url, html=webpage, video_id=video_id, headers=headers)
+        return webpage
+
     def _real_extract(self, url):
         mobj = re.match(self._VALID_URL, url)
         host = mobj.group('host')
@@ -1964,9 +1972,9 @@ class OpenloadIE(InfoExtractor):
             'User-Agent': self._USER_AGENT_TPL % random.choice(self._CHROME_VERSIONS),
         }
 
-        for path in ('embed', 'f'):
+        for path in (self._EMBED_WORD, self._STREAM_WORD):
             page_url = url_pattern % path
-            last = path == 'f'
+            last = path == self._STREAM_WORD
             webpage = self._download_webpage(
                 page_url, video_id, 'Downloading %s webpage' % path,
                 headers=headers, fatal=last)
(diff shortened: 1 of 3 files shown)


remitamine:
no, take a look at `NovaMovIE` based extractors.

ealgase:
Ah, gotcha.  So I'll make an OpenloadBaseIE, then have a OpenloadIE and a VerystreamIE that both just wrap around that, correct?

remitamine:
for now, there is no need for a `OpenloadBaseIE`, just extend from `OpenloadIE` similar to `NovaMovIE`.

ealgase:
Ah, so set some defaults in OpenloadIE and overwrite them with VerystreamIE.  I'll get working.

ealgase:
OK, I've separated the extractors like you suggested.

ealgase:
> for now, there no need to prevent users without `PhantomJS` from using `VeryStream`.

Why not?  Is extraction possible without it?

remitamine:
> Why not? Is extraction possible without it?

yes, it's possible for `VeryStream`.

ealgase:
Ah, now I see that.  Does that mean this shouldn't be a wrapper, and should be a totally separate extractor? (I'm not understanding how this would be implemented).

remitamine:
> Does that mean this shouldn't be a wrapper, and should be a totally separate extractor?

the only part that needs to change is these two lines, so you can move them to a separate method and override it in `VeryStreamIE`.
also, the hardcoded `VeryStream` element ids(https://github.com/ytdl-org/youtube-dl/blob/master/youtube_dl/extractor/openload.py#L1984-L1986) should be different.

ealgase:
Verystream appears to only have on hardcoded id (`videolink`) so I'm unsure how to implement that.  (sorry for asking so many questions, I'm still learning how to effectively contribute here)

remitamine:
you can put the element IDs in a tuple and iterate over them checking for the existence of every value otherwise fallback to path regex.

ealgase:
OK, I've made the requested changed.

youtube_dl/extractor/openload.py
```diff
@@ -254,7 +254,15 @@ class OpenloadIE(InfoExtractor):
                         (?:f|embed)/
                         (?P<id>[a-zA-Z0-9-_]+)
                     ''' % _DOMAINS
+    _EMBED_WORD = 'embed'
+    _STREAM_WORD = 'f'
+    _REDIR_URL = 'stream'
```
remitamine: use an appropriate name for the variable(it's not a URL).

youtube_dl/extractor/openload.py
```diff
@@ -254,7 +254,15 @@ class OpenloadIE(InfoExtractor):
...
+    _URL_IDS = ('streamurl', 'streamuri', 'streamurj')
 
+    def _extract_decrypted_page(self, page_url, webpage, video_id, headers):
```
remitamine: incorrect order, put the methods after the properties.

youtube_dl/extractor/openload.py
```diff
@@ -2012,3 +2019,44 @@ def _real_extract(self, url):
...
+                        )/
+                        (?:stream|e)/
+                        (?P<id>[a-zA-Z0-9-_]+)
+                    ''' % _DOMAINS
+    _EMBED_WORD = 'e'
+    _STREAM_WORD = 'stream'
+    _REDIR_URL = 'gettoken'
+    _URL_IDS = ('videolink', )
+
+    def _extract_decrypted_page(self, page_url, webpage, video_id, headers):
```
remitamine: same.

youtube_dl/extractor/openload.py
```diff
@@ -2012,3 +2019,44 @@ def _real_extract(self, url):
...
+            'title': 'Big Buck Bunny.mp4',
+            'thumbnail': r're:^https?://.*\.jpg$',
+        },
+    }, {
+        'url': 'https://verystream.com/e/c1GWQ9ngBBx/',
+        'only_matching': True,
+    }]
+
+    @staticmethod
+    def _extract_urls(webpage):
```
dstftw: Unused.

etmatrix:
My patch will be obsolete but I tried this patch and doesn't working, I hope you can fix it.

youtube_dl/extractor/openload.py
ealgase: OK, I've added a reference in the generic extractor (so it now extracts Verystream embeds).

youtube_dl/extractor/openload.py
ealgase: Done

youtube_dl/extractor/openload.py
ealgase: Changed to `_REDIR_WORD`.

ealgase:
OK, I've made all the requested changed and squashed the commits.

youtube_dl/extractor/openload.py
```diff
@@ -2012,3 +2019,44 @@ def _real_extract(self, url):
...
+    }]
+
+    def _extract_decrypted_page(self, page_url, webpage, video_id, headers):
+        return webpage  # for Verystream, the webpage is already decrypted
+
+    @staticmethod
+    def _extract_urls(webpage):
+        return re.findall(
+            r'<iframe[^>]+src=["\']((?:https?://)?%s/e/[a-zA-Z0-9-_]+)'
+            % VerystreamIE._DOMAINS, webpage)
```
dstftw: Code duplication with base class method.

ealgase: What solution would you recommend?  Since it's a `staticmethod`, self._DOMAINS and self._EMBED_WORD can't be used.

dstftw: classmethod obviously.

youtube_dl/extractor/openload.py
```diff
@@ -2012,3 +2019,44 @@ def _real_extract(self, url):
             'subtitles': subtitles,
             'http_headers': headers,
         }
+
+
+class VerystreamIE(OpenloadIE):
+    IE_NAME = 'verystream'
+
+    _DOMAINS = r'(verystream\.com)'
```
dstftw: Must not capture.

ealgase:
OK, I've made the requested changes.

PR #21045, opened 2019-05-10
Cloudflare can use videodelivery.net subdomains

davex25:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [x] At least skimmed through [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

Was attempting to youtube-dl a video from a videodelivery.net stream, which did not work.  After a little research, discovered that it appears to be hosted by Cloudflare, and with a modification of the InfoExtractor regexes it appears to work.

PR #21049, opened 2019-05-10
Cloudflare can use videodelivery.net subdomains

davex25:
Another resubmission of https://github.com/ytdl-org/youtube-dl/pull/21045, without code duplication.

### DIFF
--- a/youtube_dl/extractor/cloudflarestream.py
+++ b/youtube_dl/extractor/cloudflarestream.py
@@ -10,7 +10,7 @@ class CloudflareStreamIE(InfoExtractor):
     _VALID_URL = r'''(?x)
                     https?://
                         (?:
-                            (?:watch\.)?cloudflarestream\.com/|
+                            (?:watch\.)?(?:cloudflarestream\.com|videodelivery\.net)/|
                             embed\.(?:cloudflarestream\.com|videodelivery\.net)/embed/[^/]+\.js\?.*?\bvideo=
                         )
                         (?P<id>[\da-f]+)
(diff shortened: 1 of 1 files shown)


dstftw:
FYI, you should not flood with PRs but push into the same branch.

youtube_dl/extractor/cloudflarestream.py
```diff
@@ -11,7 +11,7 @@ class CloudflareStreamIE(InfoExtractor):
                     https?://
                         (?:
                             (?:watch\.)?cloudflarestream\.com/|
```
dstftw: watch.videodelivery.net should also be added.

davex25: Ok.

youtube_dl/extractor/cloudflarestream.py
```diff
@@ -28,17 +28,26 @@ class CloudflareStreamIE(InfoExtractor):
     }, {
         'url': 'https://watch.cloudflarestream.com/9df17203414fd1db3e3ed74abbe936c1',
         'only_matching': True,
+    }, {
+        'url': 'https://watch.videodelivery.net/9df17203414fd1db3e3ed74abbe936c1',
+        'only_matching': True,
+    }, {
+        'url': 'https://videodelivery.net/9df17203414fd1db3e3ed74abbe936c1',
```
dstftw: No such URL exists.

davex25: True.  I don't have a working watch URL to add, should I just remove these?

dstftw: Yes.

davex25: Ok.

PR #21079, opened 2019-05-13
[Kaltura] Add support for an uncommon typo in a GDC Vault page

zorbathut:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

A few pages on the GDC Vault (https://www.gdcvault.com/play/1025699, https://www.gdcvault.com/play/1025739, possibly more) have a slightly-misformed Kaltura iframe link, where the URL has an unnecessary leading space. The Kaltura extractor isn't durable enough to ignore that so it fails to download the file. Web browsers seem happy to skip it, however.

This allows it to ignore that initial leading space when it exists.

### DIFF
--- a/youtube_dl/extractor/kaltura.py
+++ b/youtube_dl/extractor/kaltura.py
@@ -103,7 +103,21 @@ class KalturaIE(InfoExtractor):
         {
             'url': 'https://www.kaltura.com:443/index.php/extwidget/preview/partner_id/1770401/uiconf_id/37307382/entry_id/0_58u8kme7/embed/iframe?&flashvars[streamerType]=auto',
             'only_matching': True,
-        }
+        },
+        {
+            'url': 'https://www.gdcvault.com/play/1025699',
+            'info_dict': {
+                'id': '0_zagynv0a',
+                'ext': 'mp4',
+                'title': 'Tech Toolbox',
+                'upload_date': '20190408',
+                'uploader_id': 'joe@blazestreaming.com',
+                'timestamp': 1554764629,
+            },
+            'params': {
+                'skip_download': True,
+            },
+        },
     ]
 
     @staticmethod
--- a/youtube_dl/extractor/kaltura.py
+++ b/youtube_dl/extractor/kaltura.py
@@ -135,7 +135,7 @@ class KalturaIE(InfoExtractor):
                 ''', webpage)
             or re.search(
                 r'''(?xs)
-                    <(?:iframe[^>]+src|meta[^>]+\bcontent)=(?P<q1>["'])[ ]?
+                    <(?:iframe[^>]+src|meta[^>]+\bcontent)=(?P<q1>["'])\s*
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/kaltura.py
```diff
@@ -135,7 +140,7 @@ def _extract_url(webpage):
                 ''', webpage)
             or re.search(
                 r'''(?xs)
-                    <(?:iframe[^>]+src|meta[^>]+\bcontent)=(?P<q1>["'])
+                    <(?:iframe[^>]+src|meta[^>]+\bcontent)=(?P<q1>["'])[ ]?
```
dstftw: Superfluous brackets. Doesn't cover multiple spaces. Doesn't cover other kinds of whitespace.

youtube_dl/extractor/kaltura.py
```diff
@@ -103,6 +103,11 @@ class KalturaIE(InfoExtractor):
         {
             'url': 'https://www.kaltura.com:443/index.php/extwidget/preview/partner_id/1770401/uiconf_id/37307382/entry_id/0_58u8kme7/embed/iframe?&flashvars[streamerType]=auto',
             'only_matching': True,
+        },
+        {
+            # some GDC Vault pages embed a URL with extra leading space
+            'url': ' https://cdnapisec.kaltura.com/p/1670711/sp/167071100/embedIframeJs/uiconf_id/43558772/partner_id/1670711?iframeembed=true&playerId=kaltura_player_1547062087&cache_st=1547072087&width=1000&height=570&entry_id=1_gknidq66&videoid=ee6b8d51c7ebda0720a91025739',
```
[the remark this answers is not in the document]
zorbathut: I actually ended up removing the test because I couldn't figure out how to make it work properly - should it be reading off a live webpage? Is there an example of something that does the right thing?

youtube_dl/extractor/kaltura.py
zorbathut: Brackets are necessary because it has the X option set, which makes it ignore whitespace outside a character class.

I'll augment it to cover unlimited amounts of all whitespace, however.

youtube_dl/extractor/kaltura.py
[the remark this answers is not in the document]
zorbathut: There we go, figured it out.

youtube_dl/extractor/kaltura.py
```diff
@@ -103,7 +103,21 @@ class KalturaIE(InfoExtractor):
         {
             'url': 'https://www.kaltura.com:443/index.php/extwidget/preview/partner_id/1770401/uiconf_id/37307382/entry_id/0_58u8kme7/embed/iframe?&flashvars[streamerType]=auto',
             'only_matching': True,
-        }
+        },
+        {
+            'url': 'https://www.gdcvault.com/play/1025699',
```
dstftw: This must be in gdcvault extractor.

zorbathut:
Moved to https://github.com/ytdl-org/youtube-dl/pull/28969 .

PR #21155, opened 2019-05-19
[srgssr] add popupvideoplayer support

basicmaster:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This add support for video that are played in a popup video player, e.g. this one:
https://www.srf.ch/play/tv/popupvideoplayer?id=c4dba0ca-e75b-43b2-a34f-f708a4932e01

As the pattern group name can't be used twice in a regex, I surrounded it with "popup" and "player" which is very unlikely to lead to false positives.

### DIFF
--- a/youtube_dl/extractor/srgssr.py
+++ b/youtube_dl/extractor/srgssr.py
@@ -106,7 +106,7 @@ class SRGSSRIE(InfoExtractor):
 
 class SRGSSRPlayIE(InfoExtractor):
     IE_DESC = 'srf.ch, rts.ch, rsi.ch, rtr.ch and swissinfo.ch play sites'
-    _VALID_URL = r'https?://(?:(?:www|play)\.)?(?P<bu>srf|rts|rsi|rtr|swissinfo)\.ch/play/(?:tv|radio)/[^/]+/(?P<type>video|audio)/[^?]+\?id=(?P<id>[0-9a-f\-]{36}|\d+)'
+    _VALID_URL = r'https?://(?:(?:www|play)\.)?(?P<bu>srf|rts|rsi|rtr|swissinfo)\.ch/play/(?:tv|radio)/(?:[^/]+/|popup)(?P<type>video|audio)(?:/[^?]+|player)\?id=(?P<id>[0-9a-f\-]{36}|\d+)'
 
     _TESTS = [{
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/srgssr.py
```diff
@@ -106,7 +106,7 @@ def _real_extract(self, url):
 
 class SRGSSRPlayIE(InfoExtractor):
     IE_DESC = 'srf.ch, rts.ch, rsi.ch, rtr.ch and swissinfo.ch play sites'
-    _VALID_URL = r'https?://(?:(?:www|play)\.)?(?P<bu>srf|rts|rsi|rtr|swissinfo)\.ch/play/(?:tv|radio)/[^/]+/(?P<type>video|audio)/[^?]+\?id=(?P<id>[0-9a-f\-]{36}|\d+)'
+    _VALID_URL = r'https?://(?:(?:www|play)\.)?(?P<bu>srf|rts|rsi|rtr|swissinfo)\.ch/play/(?:tv|radio)/(?:[^/]+/|popup)(?P<type>video|audio)(?:/[^?]+|player)\?id=(?P<id>[0-9a-f\-]{36}|\d+)'
```
dstftw: Instead of such hacks you can name group differently and capture it without any issue.

youtube_dl/extractor/srgssr.py
```diff
@@ -163,6 +163,17 @@ class SRGSSRPlayIE(InfoExtractor):
...
+    }, {
+        'url': 'https://www.srf.ch/play/tv/popupvideoplayer?id=c4dba0ca-e75b-43b2-a34f-f708a4932e01',
+        'md5': 'f6247aa7c905b81c9ba7f50fb22e2fbd',
+        'info_dict': {
+            'id': 'c4dba0ca-e75b-43b2-a34f-f708a4932e01',
+            'ext': 'mp4',
+            'upload_date': '20190122',
+            'title': 'Erster Selfie-Stick (1983)',
+            'description': 'md5:23a6b40024e583137e4137f5946543c1',
+            'timestamp': 1548155133,
```
dstftw: `only_matching`.

youtube_dl/extractor/srgssr.py
basicmaster: I'm not that firm with Python yet, so I tried to stick with a single regex, but this is indeed more elegant.

Issue #17354, opened 2018-08-26
Unsupported URL: phoenix.de


mcnesium:
phoenix.de has had a relaunch recently. Now Downloads fail due to unsupported URL error:

```
~$ youtube-dl --version
2018.08.22
~$ youtube-dl https://www.phoenix.de/sendungen/dokumentationen/mein-ausland/gespaltenes-land---schweden-hat-die-wahl-a-324347.html
[generic] gespaltenes-land---schweden-hat-die-wahl-a-324347: Requesting header
WARNING: Falling back on generic information extractor.
[generic] gespaltenes-land---schweden-hat-die-wahl-a-324347: Downloading webpage
[generic] gespaltenes-land---schweden-hat-die-wahl-a-324347: Extracting information
ERROR: Unsupported URL: https://www.phoenix.de/sendungen/dokumentationen/mein-ausland/gespaltenes-land---schweden-hat-die-wahl-a-324347.html

```
the video is on the bottom of the page

h-2:
:+1:

majewsky:
This is still an issue with version 2019.05.20 and video URL https://www.phoenix.de/sendungen/gespraeche/phoenix-runde/nach-dem-wahldebakel---cdu-und-spd-im-taumel-a-1052290.html. On the website, the video player self-identifies as "ZDF player" in the right-click menu, so maybe logic could be shared with the code for ZDF Mediathek.

matthiasroos:
The player might be the same as for ZDF Mediathek, but it is loaded by javascript, rendering the ZDF code useless and making everything very complcated and tedious. Therefore, I am out.

mcnesium:
This issue is not solved yet. Why has it been closed?

    m@x ~/downloads youtube-dl --version
    2021.02.22
    
    m@x ~/downloads youtube-dl https://www.phoenix.de/sendungen/dokumentationen/mensch--mythos-a-1303895.html
    [generic] mensch--mythos-a-1303895: Requesting header
    WARNING: Falling back on generic information extractor.
    [generic] mensch--mythos-a-1303895: Downloading webpage
    [generic] mensch--mythos-a-1303895: Extracting information
    ERROR: Unsupported URL: https://www.phoenix.de/sendungen/dokumentationen/mensch--mythos-a-1303895.html

PR #21356, opened 2019-06-10
[3sat] extractor now included in zdf3sat.py (closes #21185)

### LINKED ISSUE #21185: [3sat] New URL
```
$ youtube-dl --ignore-config --verbose https://www.3sat.de/wissen/nano/nano-21-mai-2019-102.html
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'--ignore-config', u'--verbose', u'https://www.3sat.de/wissen/nano/nano-21-mai-2019-102.html']                                                                                                                   
[debug] youtube-dl version 2019.05.20
[debug] Python version 2.7.16 (CPython) - Linux-5.0.0-15-generic-x86_64-with-Ubuntu-19.04-disco
[debug] exe versions: ffmpeg 4.1.3, ffprobe 4.1.3, phantomjs 2.1.1, rtmpdump 2.4
[debug] Proxy map: {}
[generic] nano-21-mai-2019-102: Requesting header
[generic] nano-21-mai-2019-102: Downloading webpage
[generic] nano-21-mai-2019-102: Extracting information
ERROR: Unsupported URL: https://www.3sat.de/wissen/nano/nano-21-mai-2019-102.html
  File "/home/marc/bin/youtube-dl/youtube_dl/extractor/generic.py", line 2340, in _real_extract
    doc = compat_etree_fromstring(webpage.encode('utf-8'))
  File "/home/marc/bin/youtube-dl/youtube_dl/compat.py", line 2551, in compat_etree_fromstring
    doc = _XML(text, parser=etree.XMLParser(target=_TreeBuilder(element_factory=_element_factory)))
  File "/home/marc/bin/youtube-dl/youtube_dl/compat.py", line 2540, in _XML
    parser.feed(text)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1659, in feed
    self._raiseerror(v)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1523, in _raiseerror
    raise err
ParseError: not well-formed (invalid token): line 134, column 841
  File "/home/marc/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 796, in extract_info
    ie_result = ie.extract(url)
  File "/home/marc/bin/youtube-dl/youtube_dl/extractor/common.py", line 529, in extract
  File "/home/marc/bin/youtube-dl/youtube_dl/extractor/generic.py", line 3329, in _real_extract
    raise UnsupportedError(url)
UnsupportedError: Unsupported URL: https://www.3sat.de/wissen/nano/nano-21-mai-2019-102.html

```

matthiasroos:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

With a [relaunch of website and medai library](http://www.quotenmeter.de/n/109532/eins-fuer-alles-3sat-legt-homepage-und-mediathek-zusammen) the extractor stopped working ([#21185](https://github.com/ytdl-org/youtube-dl/issues/21185)).
As correctly pointed out in the discussion of the issue by [barsnick](https://github.com/ytdl-org/youtube-dl/issues/21185#issuecomment-500046024), 3sat now relies on the infrastructure of its main contributor zdf. I added support for 3sat to the zdf extractor. A separate 3sat is no longer needed and therefore removed.


matthiasroos:
I reverted my initial commit, because phoenix.py relies on dreisat.py, but is itself broken since a relaunch last year ([issue #17354](https://github.com/ytdl-org/youtube-dl/issues/17354)). Perhaps it is also possible to include phoenix in the zdf extractor (then as zdf3satphoenix.py), but it seems more complicated as for 3sat.

youtube_dl/extractor/extractors.py
```diff
@@ -1522,6 +1522,6 @@
     ZattooIE,
     ZattooLiveIE,
 )
-from .zdf import ZDFIE, ZDFChannelIE
+from .zdf3sat import ZDFIE, ZDFChannelIE
```
dstftw: Revert.

youtube_dl/extractor/zdf3sat.py
```diff
@@ -39,7 +39,7 @@ def _extract_player(self, webpage, video_id, fatal=True):
 
 
 class ZDFIE(ZDFBaseIE):
-    _VALID_URL = r'https?://www\.zdf\.de/(?:[^/]+/)*(?P<id>[^/?]+)\.html'
+    _VALID_URL = r'https?://www\.(?:zdf|3sat)\.de/(?:[^/]+/)*(?P<id>[^/?]+)\.html'
```
dstftw: Not all URLs are interchangeable so it's most likely should be separate extractor.

youtube_dl/extractor/extractors.py
matthiasroos: Done.

youtube_dl/extractor/zdf3sat.py
matthiasroos: Understood. Made new 3sat extractor based on zdf extractor.

matthiasroos:
@dstftw Is it okay like that?

dstftw:
No, it's not. Not copypasting here.

matthiasroos:
But what do we gain by developing new code instead of using perfectly fine code, because it is not a coincidence, 3sat works with ZDF code.

1. 3sat and ZDF belong to the same network,
2. they both use the same website structure,
3. they both use the same zdf player,
4. they both rely on the same streaming service in the background,
5. there are precedences in your program for shared code (prosiebensat1 is one, in this case there are even more channels of one network crammed into one regex and so one extractor, each channel with a different URL) 

So in this light, what does still speak against the shared regex and extractor, except the different URLs? I'm just curious. If there are good arguments, I will develop a better solution without copy/paste.

dstftw:
You **don't need** to develop a new code. You **must generalize and reuse** existing. Without copypasting.

matthiasroos:
@dstftw How about now? I reuse the ZDF code in a separate extractor without any copy paste.

youtube_dl/extractor/zdf3sat.py
FliegendeWurst: > Not all URLs are interchangeable so it's most likely should be separate extractor.

Interesting, would you mind giving an example?

youtube_dl/extractor/dreisat.py
```diff
@@ -1,193 +1,29 @@
...
-    unified_strdate,
-    xpath_text,
-    determine_ext,
-    float_or_none,
-    ExtractorError,
-)
-
-
-class DreiSatIE(InfoExtractor):
-    IE_NAME = '3sat'
```
barsnick: The reworked extractor should probably keep the IE_NAME.

youtube_dl/extractor/dreisat.py
```diff
@@ -1,193 +1,29 @@
+# coding: utf-8
 from __future__ import unicode_literals
 
-import re
+from .zdf import ZDFIE
+
+
+class DreiSatIE(ZDFIE):
+    _VALID_URL = r'https?://www\.3sat\.de/(?:[^/]+/)*(?P<id>[^/?]+)\.html'
+    _TESTS = [{
```
barsnick: If this passes the test, then: Nice! I have a different test case, but it's probably not worth the redundancy.

youtube_dl/extractor/dreisat.py
```diff
@@ -1,193 +1,29 @@
...
+         'url': 'https://www.3sat.de/dokumentation/natur/dolomiten-sagenhaftes-juwel-der-alpen-100.html',
+         'info_dict': {
```
barsnick: If this passes the test, then: Nice! I have a different test case, but it's probably not worth the redundancy.

basicmaster:
> But what do we gain by developing new code instead of using perfectly fine code, because it is not a coincidence, 3sat works with ZDF code.
> 
>     1. 3sat and ZDF belong to the same network,
> 
>     2. they both use the same website structure,
> 
>     3. they both use the same zdf player,
> 
>     4. they both rely on the same streaming service in the background,

Exactly. 3sat is a common program provided by ZDF (and ARD, ORF and SRG SSR). The technical processes like the linear TV channel, the website or these catch-up services are all maintained at ZDF in Mainz. So it would be absolutely reasonable to keep the name "ZDF" for the extractor; this would also keep the change smaller.

>     5. there are precedences in your program for shared code (prosiebensat1 is one, in this case there are even more channels of one network crammed into one regex and so one extractor, each channel with a different URL)

I agree - videos for the program SIXX (part of ProSiebenSAT.1) would also be downloaded by the prosiebensat1 extractor, though the name doesn't indicate any affiliation here.

PR #21372, opened 2019-06-12
[tf1] fix wat id extraction (closes ytdl-org#21365)

froiss:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

### DIFF
--- a/youtube_dl/extractor/tf1.py
+++ b/youtube_dl/extractor/tf1.py
@@ -43,12 +43,25 @@ class TF1IE(InfoExtractor):
     }, {
         'url': 'http://www.tf1.fr/hd1/documentaire/videos/mylene-farmer-d-une-icone.html',
         'only_matching': True,
+    }, {
+        'url': 'https://www.tf1.fr/tmc/quotidien-avec-yann-barthes/videos/quotidien-premiere-partie-11-juin-2019.html',
+        'info_dict': {
+            'id': '13641379',
+            'ext': 'mp4',
+            'title': 'Quotidien, première partie du 11 juin 2019',
+            'description': 'Retrouvez l’intégralité du replay de la première partie de Quotidien du 11 juin. On parle des enfants français rapatriés de Syrie avec Salhia Brakhlia, de la décision du New York Times d’arrêter les dessins politiques avec Lilia Hassaine, on part voir les Bleues à J-1 de leur rencontre avec la ...',
+            'upload_date': '20190611',
+        },
+        'params': {
+            # Sometimes wat serves the whole file with the --test option
+            'skip_download': True,
+        },
     }]
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
         wat_id = self._html_search_regex(
-            r'(["\'])(?:https?:)?//www\.wat\.tv/embedframe/.*?(?P<id>\d{8})\1',
+            r'"streamId":"(?P<id>\d{8})"',
             webpage, 'wat id', group='id')
         return self.url_result('wat:%s' % wat_id, 'Wat')
--- a/youtube_dl/extractor/tf1.py
+++ b/youtube_dl/extractor/tf1.py
@@ -3,6 +3,8 @@ from __future__ import unicode_literals
 
 from .common import InfoExtractor
 
+from ..utils import js_to_json
+
 
 class TF1IE(InfoExtractor):
     """TF1 uses the wat.tv player."""
@@ -61,16 +63,22 @@ class TF1IE(InfoExtractor):
     def _real_extract(self, url):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
-        slug = self._search_regex(
-            r'(?<=/)(?P<slug>[^/]+)(?=\.html$)',
-            url, 'slug', group='slug', default='')
-        wat_id = self._html_search_regex(
-            [r'(["\'])(?:https?:)?//www\.wat\.tv/embedframe/.*?(?P<id>\d{8})\1',
-             r'(["\']?)streamId\1\s*:\s*(["\']?)(?P<id>\d+)\2\s*,[^}}]*(["\']?)slug\4\s*:\s*(["\']){}\5'
-             .format(slug),
-             r'(["\']?)slug\1\s*:\s*(["\']){}\2,[^}}]*(["\']?)streamId\3\s*:\s*(["\']?)(?P<id>\d+)\4'
-             .format(slug),
-             r'(["\']?)streamId\1\s*:\s*(["\']?)(?P<id>\d+)\2'
-             ],
-            webpage, 'wat id', group='id')
+        vids_data_string = self._html_search_regex(
+            r'<script>\s*window.__APOLLO_STATE__\s*=\s*(?P<vids_data_string>.*?)\s*;*\s*</script>',
+            webpage, 'videos data string', group='vids_data_string', default=None)
+        wat_id = None
+        if vids_data_string is not None:
+            vids_data = self._parse_json(
+                vids_data_string, video_id,
+                transform_source=js_to_json)
+            video_data = [v for v in vids_data.values()
+                          if 'slug' in v and v['slug'] == video_id]
+            if len(video_data) > 0 and 'streamId' in video_data[0]:
+                wat_id = video_data[0]['streamId']
+        if wat_id is None:
+            wat_id = self._html_search_regex(
+                [r'(["\'])(?:https?:)?//www\.wat\.tv/embedframe/.*?(?P<id>\d{8})\1',
+                 r'(["\']?)streamId\1\s*:\s*(["\']?)(?P<id>\d+)\2'
+                 ],
+                webpage, 'wat id', group='id')
         return self.url_result('wat:%s' % wat_id, 'Wat')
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/tf1.py
```diff
@@ -43,12 +43,25 @@ class TF1IE(InfoExtractor):
...
+            'skip_download': True,
+        },
     }]
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
         wat_id = self._html_search_regex(
-            r'(["\'])(?:https?:)?//www\.wat\.tv/embedframe/.*?(?P<id>\d{8})\1',
+            r'"streamId":"(?P<id>\d{8})"',
```
dstftw: 1. Do not remove the old pattern.
2. Relax regex.

froiss: What should I do with the old pattern? Comment it out?

dstftw: No. As already said you must keep the old pattern along with the new.

youtube_dl/extractor/tf1.py
```diff
@@ -61,7 +64,13 @@ class TF1IE(InfoExtractor):
...
-            webpage, 'wat id', group='id')
+        try:
+            wat_id = self._html_search_regex(
+                # the old pattern. Should no longer work as of 2019-06-12
+                r'(["\'])(?:https?:)?//www\.wat\.tv/embedframe/.*?(?P<id>\d{8})\1',
+                webpage, 'wat id', group='id')
+        except RegexNotFoundError:
+            wat_id = self._html_search_regex(
+                r'\bstreamId\W+(?P<id>\d+)',
+                webpage, 'wat id', group='id')
```
dstftw: No. Use list of regexes.

youtube_dl/extractor/tf1.py
```diff
@@ -43,12 +46,31 @@ class TF1IE(InfoExtractor):
...
-            r'(["\'])(?:https?:)?//www\.wat\.tv/embedframe/.*?(?P<id>\d{8})\1',
-            webpage, 'wat id', group='id')
+        try:
+            wat_id = self._html_search_regex(
+                # the old pattern. Should no longer work as of 2019-06-12
+                r'(["\'])(?:https?:)?//www\.wat\.tv/embedframe/.*?(?P<id>\d{8})\1',
+                webpage, 'wat id', group='id')
+        except RegexNotFoundError:
+            wat_id = self._html_search_regex(
+                r'\bstreamId\W+(?P<id>\d+)',
```
dstftw: No `\W`. You must match arbitrary quotes and arbitrary whitespace.

youtube_dl/extractor/tf1.py
kylepw: @froiss Like he said, you can pass a list of regexes instead of a single regex: self._html_search_regex([r'old regex', r'new regex',], webpage, 'wat id')

youtube_dl/extractor/tf1.py
froiss: Fixed in latest commit

youtube_dl/extractor/tf1.py
```diff
@@ -3,6 +3,9 @@
 
 from .common import InfoExtractor
 
+from youtube_dl.utils import (
+    RegexNotFoundError,
```
dstftw: Unused.

youtube_dl/extractor/tf1.py
```diff
@@ -43,12 +46,26 @@ class TF1IE(InfoExtractor):
...
+        },
     }]
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
         wat_id = self._html_search_regex(
-            r'(["\'])(?:https?:)?//www\.wat\.tv/embedframe/.*?(?P<id>\d{8})\1',
+            [r'(["\'])(?:https?:)?//www\.wat\.tv/embedframe/.*?(?P<id>\d{8})\1',
+             r'(["\']?)streamId\1\s*:\s*["\']?(?P<id>\d+)'],
```
dstftw: This is ambiguous since regex matches multiple ids. Parse correct id from JSON matching by slug and only if it fails fallback to this regex pattern.

youtube_dl/extractor/tf1.py
```diff
@@ -43,12 +46,26 @@ class TF1IE(InfoExtractor):
     }, {
         'url': 'http://www.tf1.fr/hd1/documentaire/videos/mylene-farmer-d-une-icone.html',
         'only_matching': True,
+    }, {
+        'url': 'https://www.tf1.fr/tmc/quotidien-avec-yann-barthes/videos/quotidien-premiere-partie-11-juin-2019.html',
+        'info_dict': {
+            'id': '13641379',
+            'ext': 'mp4',
+            'title': 'Quotidien, première partie du 11 juin 2019',
+            'description': 'Retrouvez l’intégralité du replay de la première partie de Quotidien du 11 juin. On parle des enfants français rapatriés de Syrie avec Salhia Brakhlia, de la décision du New York Times d’arrêter les dessins politiques avec Lilia Hassaine, on part voir les Bleues à J-1 de leur rencontre avec la ...',
```
dstftw: Replace with `'md5:'`.

youtube_dl/extractor/tf1.py
froiss: Fixed in latest commit. I hope that the regexes are readable enough…

youtube_dl/extractor/tf1.py
```diff
@@ -43,12 +43,34 @@ class TF1IE(InfoExtractor):
...
+            'skip_download': True,
+        },
     }]
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
+        slug = self._search_regex(
+            r'(?<=/)(?P<slug>[^/]+)(?=\.html$)',
+            url, 'slug', group='slug', default='')
```
dstftw: It's already extracted as video_id.

youtube_dl/extractor/tf1.py
```diff
@@ -43,12 +43,34 @@ class TF1IE(InfoExtractor):
...
         wat_id = self._html_search_regex(
-            r'(["\'])(?:https?:)?//www\.wat\.tv/embedframe/.*?(?P<id>\d{8})\1',
+            [r'(["\'])(?:https?:)?//www\.wat\.tv/embedframe/.*?(?P<id>\d{8})\1',
+             r'(["\']?)streamId\1\s*:\s*(["\']?)(?P<id>\d+)\2\s*,[^}}]*(["\']?)slug\4\s*:\s*(["\']){}\5'
+             .format(slug),
+             r'(["\']?)slug\1\s*:\s*(["\']){}\2,[^}}]*(["\']?)streamId\3\s*:\s*(["\']?)(?P<id>\d+)\4'
+             .format(slug),
```
dstftw: As already said: parse as JSON not with regexes.

froiss: Fixed in latest commit.

rs:
Can we have this merged?

youtube_dl/extractor/tf1.py
```diff
@@ -43,12 +45,40 @@ class TF1IE(InfoExtractor):
...
     }]
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
-        wat_id = self._html_search_regex(
-            r'(["\'])(?:https?:)?//www\.wat\.tv/embedframe/.*?(?P<id>\d{8})\1',
-            webpage, 'wat id', group='id')
+        vids_data_string = self._html_search_regex(
+            r'<script>\s*window\.__APOLLO_STATE__\s*=\s*(?P<vids_data_string>\{.*?\})\s*;?\s*</script>',
```
dstftw: 1. Remove script tags, it's unique enough without them.
2. Do not use named group when there is only one group.
3. Curly braces don't need escaping.
4. Do not capture empty dict.

youtube_dl/extractor/tf1.py
```diff
@@ -43,12 +45,40 @@ class TF1IE(InfoExtractor):
...
+            webpage, 'videos data string', group='vids_data_string', default=None)
+        wat_id = None
+        if vids_data_string is not None:
+            vids_data = self._parse_json(
+                vids_data_string, video_id,
+                transform_source=js_to_json)
```
dstftw: Must not be fatal.

youtube_dl/extractor/tf1.py
```diff
@@ -43,12 +45,40 @@ class TF1IE(InfoExtractor):
...
+            video_data = [v for v in vids_data.values()
```
dstftw: `video_data` is totally useless. Write directly to id variable when found.

youtube_dl/extractor/tf1.py
```diff
@@ -43,12 +45,40 @@ class TF1IE(InfoExtractor):
...
+                          if 'slug' in v and v['slug'] == video_id]
```
dstftw: 1. `v` may not be dict.
2. `v.get('slug')`.

rs:
What's missing to get this merged?

Issue #21811, opened 2019-07-17
[twitch] New URLs for past-broadcasts, highlights and uploads


lel-amri:
## Checklist

- [x] I'm reporting a broken site support
- [x] I've verified that I'm running youtube-dl version **2019.07.16**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'--verbose', u'https://www.twitch.tv/spamfish/videos?filter=archives']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2019.07.16
[debug] Python version 2.7.15 (CPython) - Linux-4.19.57-gentoo-x86_64-AMD_Ryzen_5_2600_Six-Core_Processor-with-gentoo-2.6
[debug] exe versions: ffmpeg 4.1.3, ffprobe 4.1.3, rtmpdump 2.4
[debug] Proxy map: {}
[twitch:stream] spamfish: Downloading stream JSON
ERROR: spamfish is offline
Traceback (most recent call last):
  File "/home/leo/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 796, in extract_info
    ie_result = ie.extract(url)
  File "/home/leo/bin/youtube-dl/youtube_dl/extractor/common.py", line 530, in extract
    ie_result = self._real_extract(url)
  File "/home/leo/bin/youtube-dl/youtube_dl/extractor/twitch.py", line 581, in _real_extract
    raise ExtractorError('%s is offline' % channel_id, expected=True)
ExtractorError: spamfish is offline
```

## Description

It looks like Twitch changed its URLs for users's videos lists (ie. going from `https://www.twitch.tv/user/videos/past-broadcasts` to `https://www.twitch.tv/user/videos?filter=archives`).

lel-amri:
Opened PR #21812

PR #21812, opened 2019-07-17
Attempt to fix Twitch extractors for new URLs

lel-amri:
Twitch changed its URLs for user's videos lists recently. This commit includes
fixes for "past broadcasts", "hightlights" and "uploads".

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

See #21811

### DIFF
--- a/youtube_dl/extractor/twitch.py
+++ b/youtube_dl/extractor/twitch.py
@@ -205,6 +205,43 @@ class TwitchItemBaseIE(TwitchBaseIE):
         return self._extract_media(self._match_id(url))
 
 
+class TwitchVideoIE(TwitchItemBaseIE):
+    IE_NAME = 'twitch:video'
+    _VALID_URL = r'%s/[^/]+/b/(?P<id>\d+)' % TwitchBaseIE._VALID_URL_BASE
+    _ITEM_TYPE = 'video'
+    _ITEM_SHORTCUT = 'a'
+
+    _TEST = {
+        'url': 'http://www.twitch.tv/riotgames/b/577357806',
+        'info_dict': {
+            'id': 'a577357806',
+            'title': 'Worlds Semifinals - Star Horn Royal Club vs. OMG',
+        },
+        'playlist_mincount': 12,
+        'skip': 'HTTP Error 404: Not Found',
+    }
+
+
+class TwitchChapterIE(TwitchItemBaseIE):
+    IE_NAME = 'twitch:chapter'
+    _VALID_URL = r'%s/[^/]+/c/(?P<id>\d+)' % TwitchBaseIE._VALID_URL_BASE
+    _ITEM_TYPE = 'chapter'
+    _ITEM_SHORTCUT = 'c'
+
+    _TESTS = [{
+        'url': 'http://www.twitch.tv/acracingleague/c/5285812',
+        'info_dict': {
+            'id': 'c5285812',
+            'title': 'ACRL Off Season - Sports Cars @ Nordschleife',
+        },
+        'playlist_mincount': 3,
+        'skip': 'HTTP Error 404: Not Found',
+    }, {
+        'url': 'http://www.twitch.tv/tsm_theoddone/c/2349361',
+        'only_matching': True,
+    }]
+
+
 class TwitchVodIE(TwitchItemBaseIE):
     IE_NAME = 'twitch:vod'
     _VALID_URL = r'''(?x)
@@ -510,6 +547,8 @@ class TwitchStreamIE(TwitchBaseIE):
     def suitable(cls, url):
         return (False
                 if any(ie.suitable(url) for ie in (
+                    TwitchVideoIE,
+                    TwitchChapterIE,
                     TwitchVodIE,
                     TwitchProfileIE,
                     TwitchAllVideosIE,
--- a/youtube_dl/extractor/twitch.py
+++ b/youtube_dl/extractor/twitch.py
@@ -456,58 +456,49 @@ class TwitchAllVideosIE(TwitchVideosBaseIE):
 
 class TwitchUploadsIE(TwitchVideosBaseIE):
     IE_NAME = 'twitch:videos:uploads'
-    _VALID_URL = r'%s/uploads' % TwitchVideosBaseIE._VALID_URL_VIDEOS_BASE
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/twitch.py
```diff
@@ -456,58 +456,49 @@ class TwitchAllVideosIE(TwitchVideosBaseIE):
 
 class TwitchUploadsIE(TwitchVideosBaseIE):
     IE_NAME = 'twitch:videos:uploads'
-    _VALID_URL = r'%s/uploads' % TwitchVideosBaseIE._VALID_URL_VIDEOS_BASE
+    _VALID_URL = r'%s\/?\?.*filter=uploads.*' % TwitchVideosBaseIE._VALID_URL_VIDEOS_BASE
```
dstftw: Do not escape slash. Fix code duplication. Remove useless .* from the end. Do not use greedy mask.

lel-amri: * What code duplication ?
* The end `.*` is useful, because the url is actually looking like `https://www.twitch.tv/spamfish/videos?filter=archives&sort=time`. Should I change the test and include the "full" url ?

dstftw: filter.
Again: .* at the end does not make any sense since https://www.twitch.tv/spamfish/videos?
filter=archives&sort=time will match it without it.

lel-amri: * Do you mean I shouldn't repeat `%s/?\?.*filter=something.*'` ? (In this case, I'd still have to define another class variable and override the match method to generate a pattern on-the-fly)
* I actually tried this, but it didn't work. `re.match` match for the whole pattern. Anything appended to `filter=archives` would not match if the pattern was `.*filter=archives`. This is a naïve pattern that I have to change (the `.*` prefix is ugly, `\b` would be better for example).

dstftw: >`re.match` match for the whole pattern. Anything appended to `filter=archives` would not match if the pattern was `.*filter=archives`.

Wrong.

youtube_dl/extractor/twitch.py
```diff
@@ -456,58 +456,49 @@ class TwitchAllVideosIE(TwitchVideosBaseIE):
...
-        'url': 'https://www.twitch.tv/spamfish/videos/uploads',
+        'url': 'https://www.twitch.tv/spamfish/videos?filter=uploads',
         'info_dict': {
             'id': 'spamfish',
             'title': 'Spamfish',
         },
         'playlist_mincount': 0,
-    }, {
-        'url': 'https://m.twitch.tv/spamfish/videos/uploads',
-        'only_matching': True,
```
lel-amri: I removed the mobile view test, because it looks like Twitch removed specific pages for specific categories. Could someone double-check this ?

youtube_dl/extractor/twitch.py
lel-amri: Indeed, fixed it. A commit is on its way.

youtube_dl/extractor/twitch.py
```diff
@@ -394,37 +394,34 @@ class TwitchProfileIE(TwitchPlaylistBaseIE):
...
-    }, {
-        'url': 'https://m.twitch.tv/spamfish/videos/all',
-        'only_matching': True,
     }]
 
 
 class TwitchUploadsIE(TwitchVideosBaseIE):
     IE_NAME = 'twitch:videos:uploads'
-    _VALID_URL = r'%s\/?\?.*filter=uploads.*' % TwitchVideosBaseIE._VALID_URL_VIDEOS_BASE
+    _VALID_URL = TwitchVideosBaseIE._VALID_URL_VIDEOS_BASE % 'uploads'
```
lel-amri: Is it better ?

youtube_dl/extractor/twitch.py
```diff
@@ -431,83 +394,71 @@ class TwitchProfileIE(TwitchPlaylistBaseIE):
...
-        'url': 'https://www.twitch.tv/spamfish/videos/uploads',
+        'url': 'https://www.twitch.tv/spamfish/videos?filter=uploads&sort=time',
         'info_dict': {
             'id': 'spamfish',
             'title': 'Spamfish',
         },
         'playlist_mincount': 0,
-    }, {
-        'url': 'https://m.twitch.tv/spamfish/videos/uploads',
-        'only_matching': True,
```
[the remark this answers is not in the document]
lel-amri: This endpoint doesn't exists anymore. I asked for someone to double-check for this here: https://github.com/ytdl-org/youtube-dl/pull/21812#pullrequestreview-263639751

youtube_dl/extractor/twitch.py
```diff
@@ -205,43 +205,6 @@ def _real_extract(self, url):
         return self._extract_media(self._match_id(url))
 
 
-class TwitchVideoIE(TwitchItemBaseIE):
```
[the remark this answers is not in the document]
lel-amri: The `/c/` and `/b/` endpoints doesn't exists anymore. Should I create another PR for this ? (This one have a general (fitting, imo) name: _Attempt to fix Twitch extractors for new URLs_)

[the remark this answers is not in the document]
dstftw: As already said do not mix unrelated changes in single PR. Create PR per feature.

youtube_dl/extractor/twitch.py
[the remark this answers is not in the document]
dstftw: It opens fine in browser without being redirected to another location.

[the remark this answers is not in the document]
lel-amri: Yes, but it's not an URL you can obtain naturally from Twitch. Actually, you could put anything behind `https://m.twitch.tv/spamfish/videos/`, it would work the same. The "semantic" got lost sometimes in the past. The IEs also doesn't match on this URL anymore.

youtube_dl/extractor/twitch.py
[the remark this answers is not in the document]
lel-amri: Will do

youtube_dl/extractor/twitch.py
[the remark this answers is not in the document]
lel-amri: I bringed back the test on mobile URLs in ca7e9e1

lel-amri:
I'm allowing myself to bump this PR.
What should I do now @dstftw ?

Issue #15027, opened 2017-12-18
Site Support Request raiplayradio.it


mistnim:
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2017.12.14**

### What is the purpose of your *issue*?
- [x] Site support request (request for adding support for a new site)

---
This web interface of the main italian radio and television network is brand new. No wonder it is not supported in youtube-dl. Example of playlists to support are below. 

Playlist: http://www.raiplayradio.it/playlist/2017/12/Valentino-881b6a9f-644f-41cc-89dc-aaa010c11816.html
Playlist: http://www.raiplayradio.it/playlist/2017/12/Arancia-Meccanica-9d191ecb-23df-43fe-8e97-e423707f7de3.html

Vrihub:
The attached patch for `/extractor/rai.py` adds support for raiplayradio.it audio items.

For playlists, a new function should be added, to download the playlist json (append `?json` to the playlist URL), parse it, and extract the playlist episodes information.

[rai.py.raiplayradio.txt](https://github.com/rg3/youtube-dl/files/1587381/rai.py.raiplayradio.txt)

frafra:
I made a patch almost 2 years ago. It went under revision, I made the requested improvements, but the review process is very slow :(

PR #21837, opened 2019-07-20
Raiplayradio

### LINKED ISSUE #17141: [site support request] - **raiplayradio.it**
It would be nice to have a possibility to download audiobooks from **raiplayradio.it**,
for example - https://www.raiplayradio.it/playlist/2017/12/Centanni-di-solitudine-8f58783b-6ff4-4b26-b0ba-d0948b5701b2.html

frafra:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Support for raiplayradio.it audio playlists added.
Notes:
- I added an `outer` parameter to the  `get_element*` functions in `utils.py`to allow developers to look for a tag and extract attributes from it; I needed that in order to avoid code redundancy while keeping a robust extractor
- No support for individual audio, only playlists: downloading audio from their specific pages is very complicated (or the data should be extracted loading their playlists)

Fixes #17141 and #15027.


dstftw:
Read coding conventions.

frafra:
I am trying to figuring out what is the problem here:

- [x] id, title and url aree mandatory fields
- [x] Don't capture groups you don't use
- [x] Provide fallbacks (NA)
- [x] never split long string literals like URLs
- [x] Make regular expressions relaxed and flexible 
- [ ] Long lines policy
- [x] Inline values 
- [x] Collapse fallbacks
- [ ] Trailing parentheses
- [x] Use convenience conversion and parsing functions

frafra:
@dstftw Ready for the code review.

frafra:
Would it be possible to have it merged? Is there anything else that should be done/changed?

frafra:
@dstftw anything I can do to have this PR merged?

dstftw:
I've already pointed out: read coding conventions.

frafra:
I read them and I worked on my code to respect them. I can try to recap the conventions and double check them:

- **Mandatory and optional metafields**: mandatory fields are there
- **Provide fallbacks**: the rule says "try to" provide fallbacks to make the extractor robust; my code has been working for 9 months without any issue, and it looks solid to me
- **Regular expressions**
  - **Don't capture groups you don't use**: I do not
  - **Make regular expressions relaxed and flexible**: they look to me and the code worked for 9 months
- **Long lines policy**: lines are under 80 characters
- **Inline values**: it seems to me that it is OK
- **Collapse fallbacks**: NA
- **Trailing parentheses**: code is compliant
- **Use convenience conversion and parsing functions**: I reused what was already there and I had to make a small HTML attribute parser that was not available
- **flake8**: the code passes the test for the sections I added/changed

Could you please point out which rule my code is violating? So I can fix that and avoid bothering you further.

youtube_dl/extractor/rai.py
```diff
@@ -500,3 +502,108 @@ def _real_extract(self, url):
...
+        parser = HTMLListAttrsParser()
+        parser.feed(webpage)
+        parser.close()
+        return parser.items
+
+    def get_playlist_iter(self, url):
+        webpage = self._download_webpage(url, url)
+        for attrs in self.parse_list(webpage):
+            title = attrs['data-title'].strip()
+            webpage = urljoin(url, attrs['data-mediapolis'])
```
dstftw: If either of these attrs is missing whole playlist extraction is broken.

youtube_dl/extractor/rai.py
```diff
@@ -500,3 +502,108 @@ def _real_extract(self, url):
...
+            audio_url = compat_str(self._request_webpage(
+                webpage, title).geturl())
```
dstftw: This should not be here as done by downloader.

youtube_dl/extractor/rai.py
```diff
@@ -500,3 +502,108 @@ def _real_extract(self, url):
...
+            yield {
+                'url': audio_url,
+                'id': attrs['data-uniquename'].lstrip('ContentItem-'),
```
dstftw: Same.

youtube_dl/extractor/rai.py
```diff
@@ -500,3 +502,108 @@ def _real_extract(self, url):
...
+                'title': title,
+                'ext': determine_ext(audio_url),
+                'thumbnail': urljoin(url, attrs['data-image']),
```
dstftw: Same.

youtube_dl/extractor/rai.py
```diff
@@ -500,3 +502,108 @@ def _real_extract(self, url):
...
+        'url': 'https://www.raiplayradio.it/audio/2019/07/'
+               'RADIO3---LEZIONI-DI-MUSICA-'
+               '36b099ff-4123-4443-9bf9-38e43ef5e025.html',
+        'info_dict': {
+            'id': '36b099ff-4123-4443-9bf9-38e43ef5e025',
+            'ext': 'mp3',
+            'title': 'Dal "Chiaro di luna" al  "Clair de lune", '
+                     'prima parte con Giovanni Bietti',
+            'thumbnail': 'https://www.raiplayradio.it/cropgd/400x400/dl/'
+                         'img/2019/07/08/1562590329054_luna.jpg',
```
dstftw: Same.

youtube_dl/extractor/rai.py
```diff
@@ -500,3 +502,108 @@ def _real_extract(self, url):
...
+            'id': '72371d3c-d998-49f3-8860-d168cfdf4966',
+            'title': "Alice nel paese delle meraviglie",
+            'description': "di Lewis Carrol letto da Aldo Busi"},
+        'playlist_count': 11}
+
+    def _real_extract(self, url):
+        playlist_id = self._match_id(url)
+        playlist_webpage = self._download_webpage(url, playlist_id)
+        playlist_title = unescapeHTML(self._html_search_regex(
+            r'data-playlist-title="(.+?)"', playlist_webpage, 'title'))
```
dstftw: Playlist title is optional.

youtube_dl/extractor/rai.py
```diff
@@ -500,3 +502,108 @@ def _real_extract(self, url):
...
+                return entry
+
+
+class RaiPlayRadioPlaylistIE(RaiPlayRadioBaseIE):
+    _VALID_URL = r'%s/playlist/.+?-(?P<id>%s)\.html' % (
+        RaiPlayRadioBaseIE._BASE, RaiBaseIE._UUID_RE)
+    _TEST = {
+        'url': 'https://www.raiplayradio.it/playlist/2017/12/'
+               'Alice-nel-paese-delle-meraviglie-'
+               '72371d3c-d998-49f3-8860-d168cfdf4966.html',
```
dstftw: Same.

youtube_dl/extractor/rai.py
```diff
@@ -500,3 +502,108 @@ def _real_extract(self, url):
...
+    _BASE = 'https://www.raiplayradio.it'
+
+    def parse_list(self, webpage):
+        parser = HTMLListAttrsParser()
+        parser.feed(webpage)
+        parser.close()
+        return parser.items
+
+    def get_playlist_iter(self, url):
+        webpage = self._download_webpage(url, url)
```
dstftw: `url` is not a video id.

youtube_dl/extractor/rai.py
```diff
@@ -500,3 +502,108 @@ def _real_extract(self, url):
...
+            'title': 'Dal "Chiaro di luna" al  "Clair de lune", '
+                     'prima parte con Giovanni Bietti',
+            'thumbnail': 'https://www.raiplayradio.it/cropgd/400x400/dl/'
+                         'img/2019/07/08/1562590329054_luna.jpg',
+            'language': 'it'}}
+
+    def _real_extract(self, url):
+        list_url = url.replace('.html', '-list.html')
+        for entry in self.get_playlist_iter(list_url):
+            if entry['id'] == self._match_id(url):
```
dstftw: Extract id once before the loop.

youtube_dl/extractor/rai.py
```diff
@@ -500,3 +502,108 @@ def _real_extract(self, url):
...
+                return entry
```
dstftw: If nothing matches `None` will be returned.

youtube_dl/extractor/rai.py
dstftw: No direct URLs in tests.

youtube_dl/extractor/rai.py
```diff
@@ -500,3 +502,108 @@ def _real_extract(self, url):
...
+        return list(self.get_playlist_iter(*args, **kwargs))
+
+
+class RaiPlayRadioIE(RaiPlayRadioBaseIE):
+    _VALID_URL = r'%s/audio/.+?-(?P<id>%s)\.html' % (
+        RaiPlayRadioBaseIE._BASE, RaiBaseIE._UUID_RE)
+    _TEST = {
+        'url': 'https://www.raiplayradio.it/audio/2019/07/'
+               'RADIO3---LEZIONI-DI-MUSICA-'
+               '36b099ff-4123-4443-9bf9-38e43ef5e025.html',
```
dstftw: Are you trolling or what?
https://github.com/ytdl-org/youtube-dl#long-lines-policy
>For example, you should **never** split long string literals like URLs

docs/supportedsites.md
```diff
@@ -718,6 +718,7 @@
  - **RaiPlay**
  - **RaiPlayLive**
  - **RaiPlayPlaylist**
+ - **RaiPlayRadioPlaylist**
```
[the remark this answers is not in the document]
frafra: Removed.

youtube_dl/extractor/rai.py
```diff
@@ -500,3 +502,108 @@ def _real_extract(self, url):
...
+        'info_dict': {
+            'id': '36b099ff-4123-4443-9bf9-38e43ef5e025',
+            'ext': 'mp3',
+            'title': 'Dal "Chiaro di luna" al  "Clair de lune", '
+                     'prima parte con Giovanni Bietti',
+            'thumbnail': 'https://www.raiplayradio.it/cropgd/400x400/dl/'
+                         'img/2019/07/08/1562590329054_luna.jpg',
+            'language': 'it'}}
```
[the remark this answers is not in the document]
frafra: Fixed. `,` in the last element and `})` in the following line.

youtube_dl/extractor/rai.py
frafra: This was necessary to determine file format, as there is a redirect. I set the file format to `mp3` as the audio files seems to always be like that and returned a URL without using `self._request_webpage`.

youtube_dl/extractor/rai.py
frafra: Fixed.

youtube_dl/extractor/rai.py
frafra: Using the id extracted from the URL. Hope this is enough.

youtube_dl/extractor/rai.py
frafra: Done.

youtube_dl/extractor/rai.py
frafra: Replaced with a more relaxed regex.

youtube_dl/extractor/rai.py
frafra: No trolling of course. The PR had URL on the same line, but after not having feedback except "read the conventions" I double checked again trying to figure out what was wrong and I probably overlooked that.
Fixed.

youtube_dl/extractor/rai.py
frafra: The playlist webpage is very minimal (not even a complete HTML page) and the URL appears only in those attributes. This is how the website seems to work and it seems to be reliable.
I did not figured any other way to do that.

youtube_dl/extractor/rai.py
frafra: Same as above.

youtube_dl/extractor/rai.py
frafra: Right. Should an error be raised instead?

youtube_dl/extractor/rai.py
frafra: Fixed: it will not be added to the results if missing.

youtube_dl/extractor/rai.py
frafra: The extraction has been working as expected for years and the URL does not appear anywhere else, so a failure/error seems reasonable to me. Should I raise a custom error?

frafra:
Everything seems fine to me now. The only possible problem is how to handle critical error: should I raise a specific exception when a critical failure happens?

frafra:
I know that reviewing requires time, but I used a lot of my time too to adhere to the guidelines as much as possible. I only have an open question (should I let the code fail if something happens or should I catch the exception and raise a custom error?), but except that the code looks fine to me. I would like to move forward with this PR if possible. Another year passed from my last commit.

Issue #21294, opened 2019-06-04
youtube embeds on recent (2019?) squarespace sites fail to download


galgeek:
## Checklist

- [X] I'm reporting a new site support request
- [X] I've verified that I'm running youtube-dl version **2019.07.16**
- [X] I've checked that all provided URLs are alive and playable in a browser
- [X] I've checked that none of provided URLs violate any copyrights
- [X] I've searched the bugtracker for similar site support requests including closed ones

## Description

youtube-dl fails on youtube embeds on some recent squarespace sites, including 
http://www.ootboxford.com/
https://www.harvardballetcompany.org/past-productions
http://www.immediategratification.org/

`youtube-dl --write-pages http://www.ootboxford.com`
produces a dump showing the site's youtube embed like so:
```
[...]

PR #21859, opened 2019-07-22
download squarespace escaped video embeds

galgeek:

---

- [X] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Captures videos embedded in Squarespace sites in escaped html, like the youtube embeds in Squarespace sites noted in issue #21294.

This can replace #21802 already initially reviewed by @dstftw.

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -2076,6 +2076,17 @@ class GenericIE(InfoExtractor):
             'playlist_count': 6,
         },
         {
+            # Squarespace video embed, 2019-08-28
+            'url': 'http://ootboxford.com',
+            'info_dict': {
+                'id': 'Tc7b_JGdZfw',
+                'title': 'Out of the Blue, at Childish Things 10',
+            },
+            'params': {
+                'skip_download': True,
+            },
+        },
+        {
             # Zype embed
             'url': 'https://www.cookscountry.com/episode/554-smoky-barbecue-favorites',
             'info_dict': {
@@ -2394,6 +2405,13 @@ class GenericIE(InfoExtractor):
         # (e.g. https://github.com/ytdl-org/youtube-dl/issues/2448)
         # Unescaping the whole page allows to handle those cases in a generic way
         webpage = compat_urllib_parse_unquote(webpage)
+
+        # unescape re.sub replacement
+        def unescape_resub(m):
+            return unescapeHTML(m.group(0))
+
+        # unescape squarespace video embeds
+        webpage = re.sub(r'<div[^>]+class=[^>]*?sqs-video-wrapper[^>]*>', unescape_resub, webpage)
 
         # it's tempting to parse this further, but you would
         # have to take into account all the variations like


youtube_dl/extractor/generic.py
```diff
@@ -2395,6 +2395,9 @@ def _real_extract(self, url):
         # Unescaping the whole page allows to handle those cases in a generic way
         webpage = compat_urllib_parse_unquote(webpage)
 
+        # capture escaped squarespace and other embeds
+        webpage = unescapeHTML(webpage)
```
dstftw: Do not unescape the whole page. You must only unescape embedded piece.

galgeek:
Thanks, @dstftw!

I've updated, finding all squarespace video embeds, and escaping before searching for video URL.  

I've checked using these 3 URLs:

http://www.ootboxford.com/
https://www.harvardballetcompany.org/past-productions
http://www.immediategratification.org/

youtube_dl/extractor/generic.py
```diff
@@ -2532,6 +2532,18 @@ def _real_extract(self, url):
...
+        matches = re.findall(r'<div class="[^"]*?sqs-block-video[^>]*>', webpage)
+        if matches:
+            sqs_urls = set()
+            for m in matches:
+                sqs_url = re.search(r'<iframe[^>]+?src=[^"]*"(//[^\\]+)', unescapeHTML(m))
+                if sqs_url:
+                    sqs_urls.add(sqs_url.group(1))
+            if sqs_urls:
+                return self.playlist_from_matches(
+                    sqs_urls, video_id, video_title)
```
dstftw: One more time: you must unescape and expand embed code **in place** leaving the rest to the generic extractor. You **must not** extract iframes on your own.

galgeek: Thanks, @dstftw!
I've updated, and the new code should be pretty close to what you're looking for.

And I've tested again, using the three problem sites, and all again download with the new code.

galgeek:
Thanks, @dstftw!
I've updated, and the new code should be pretty close to what you're looking for.

galgeek:
I've verified that Issue 21294 is still present in youtube-dl version 2019.08.13, and this PR still fixes the issue.

youtube_dl/extractor/generic.py
```diff
@@ -2395,6 +2395,12 @@ def _real_extract(self, url):
         # Unescaping the whole page allows to handle those cases in a generic way
         webpage = compat_urllib_parse_unquote(webpage)
 
+        # unescape squarespace video embeds
+        sqs_videos = re.findall(r'<div class="[^"]*?sqs-video-wrapper[^>]*>', webpage)
```
dstftw: 1. `re.sub`.
2. Regex must match any quotes kind, must allow other attributes before class.
3. Add a test.

galgeek:
@dstftw thanks!

I've updated, and I've verified that the latest updates resolve #21294

youtube_dl/extractor/generic.py
```diff
@@ -2395,6 +2406,13 @@ def _real_extract(self, url):
         # Unescaping the whole page allows to handle those cases in a generic way
         webpage = compat_urllib_parse_unquote(webpage)
 
+        # unescape re.sub replacement
+        def unescape_resub(m):
+            return unescapeHTML(m.group(0))
```
dstftw: Just put in lambda, no need to extract it.

youtube_dl/extractor/generic.py
```diff
@@ -2075,6 +2075,17 @@ class GenericIE(InfoExtractor):
             },
             'playlist_count': 6,
         },
+        {
+            # Squarespace video embed, 2019-08-28
+            'url': 'http://ootboxford.com',
+            'info_dict': {
+                'id': 'Tc7b_JGdZfw',
+                'title': 'Out of the Blue, at Childish Things 10',
```
dstftw: Test does not work.

galgeek:
@dstftw thank you for your help with this!  

It's great to see the fix in master, and I learned a lot, too.

PR #22180, opened 2019-08-22
Feature metadata override

katieannbaker:

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

This pull request is for a new feature that modifies the FFmpegMetadataPP post-processor in order to accept an optional argument "preferredinfo". This goal of this is to support the post-processing overriding of metadata on the extracted file.

When present this dictionary will take precedence when populating the metadata dictionary that is used to update the metadata on the extracted file. If an attribute is not present in the preferredinfo dictionary, the info dictionary will be used.

As an additional step, the info dictionary is modified to reflect the information present in the preferredinfo dictionary if, and only if, this information was used to set the attribute on the metadata object. This behavior may or may not be preferred due to the impacts the modification of the info dictionary can have on post-processors executed after FFmpegMEtadataPP. Feedback on this update is requested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,11 +409,6 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
                                      webm and mkv videos)
     --embed-thumbnail                Embed thumbnail in the audio as cover art
     --add-metadata                   Write metadata to the video file
-    --preferred-info                 References a JSON file that contains the
-                                     metadata that will be used preferentially
-                                     over the extracted metadata and upated onto
-                                     the output file. This argument will only be
-                                     utilized when add-metadata is used
     --metadata-from-title FORMAT     Parse additional metadata like song title /
                                      artist from the video title. The format
                                      syntax is the same as --output. Regular
--- a/README.md
+++ b/README.md
@@ -409,6 +409,10 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
                                      webm and mkv videos)
     --embed-thumbnail                Embed thumbnail in the audio as cover art
     --add-metadata                   Write metadata to the video file
+    --preferred-info                 References a JSON file that contains the
+                                     metadata that will be used preferentially
+                                     over the extracted metadata and upated onto
+                                     the output file
     --metadata-from-title FORMAT     Parse additional metadata like song title /
                                      artist from the video title. The format
                                      syntax is the same as --output. Regular
--- a/README.md
+++ b/README.md
@@ -412,7 +412,8 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
     --preferred-info                 References a JSON file that contains the
                                      metadata that will be used preferentially
                                      over the extracted metadata and upated onto
-                                     the output file
+                                     the output file. This argument will only be
+                                     utilized when add-metadata is used
     --metadata-from-title FORMAT     Parse additional metadata like song title /
                                      artist from the video title. The format
                                      syntax is the same as --output. Regular
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -830,8 +830,7 @@ def parseOpts(overrideArguments=None):
              'inputted from the info_list, only one will be used. Fields '
              'curently supported for metadata override are '
              '((info_list) => (meta_list)):\n {}'.format(',\n'.join(
-                 ['({}) => ({})'.format(*[', '.join(a) for a in reversed(x)]) for x in get_metadata_override_elements()]
-                 )))
+                 ['({}) => ({})'.format(*[', '.join(a) for a in reversed(x)]) for x in get_metadata_override_elements()])))
     postproc.add_option(
         '--preferred-metadata-json'
     )
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -13,6 +13,9 @@ from .compat import (
     compat_kwargs,
     compat_shlex_split,
 )
+from .postprocessor.ffmpeg import (
+    get_metadata_override_elements
+)
 from .utils import (
     preferredencoding,
     write_string,
@@ -821,7 +824,14 @@ def parseOpts(overrideArguments=None):
     postproc.add_option(
         '--preferred-metadata-to-embed',
         dest='preferredmetadatatoembed', metavar='JSON', type=str,
-        help='Override metadata on the outputted file')
+        help='Override metadata on the outputted file. Pass in a JSON string '
+             'with one element from the info_list in order to override all '
+             'elements from the meta_list. Note, if multiple values are '
+             'inputted from the info_list, only one will be used. Fields '
+             'curently supported for metadata override are '
+             '((info_list) => (meta_list)):\n {}'.format(',\n'.join(
+                 ['({}) => ({})'.format(*[', '.join(a) for a in reversed(x)]) for x in get_metadata_override_elements()]
+                 )))
     postproc.add_option(
         '--preferred-metadata-json'
     )
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -819,6 +819,10 @@ def parseOpts(overrideArguments=None):
         action='store_true', dest='addmetadata', default=False,
         help='Write metadata to the video file')
     postproc.add_option(
+        '--preferred-info',
+        dest='preferredinfo', metavar='FILE',
+        help='Override metadata on the outputted file')
+    postproc.add_option(
         '--metadata-from-title',
         metavar='FORMAT', dest='metafromtitle',
         help='Parse additional metadata like song title / artist from the video title. '
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -819,9 +819,12 @@ def parseOpts(overrideArguments=None):
         action='store_true', dest='addmetadata', default=False,
         help='Write metadata to the video file')
     postproc.add_option(
-        '--preferred-info',
-        dest='preferredinfo', metavar='FILE',
+        '--preferred-metadata-to-embed',
+        dest='preferredmetadatatoembed', metavar='JSON', type=str,
         help='Override metadata on the outputted file')
+    postproc.add_option(
+        '--preferred-metadata-json'
+    )
     postproc.add_option(
         '--metadata-from-title',
         metavar='FORMAT', dest='metafromtitle',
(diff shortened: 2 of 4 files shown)


katieannbaker:
@dstftw Would you be able to review this pull request and provide feedback?

katieannbaker:
@dstftw following up on this request, again. Thanks!

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -431,21 +431,36 @@ def run(self, information):
 
 
 class FFmpegMetadataPP(FFmpegPostProcessor):
+
```
dstftw: Remove.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -431,21 +431,36 @@ def run(self, information):
...
+    def __init__(self, downloader=None, preferredinfo=None):
+        super(FFmpegMetadataPP, self).__init__(downloader)
+        self._preferredinfo = preferredinfo if isinstance(preferredinfo, dict) else None
```
dstftw: This must be asserted.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -431,21 +431,36 @@ def run(self, information):
...
-                if info.get(info_f) is not None:
-                    for meta_f in meta_list:
-                        metadata[meta_f] = info[info_f]
-                    break
+            preferred_key = add_info(meta_list, info_list, metadata, self._preferredinfo)
+            if preferred_key is None:
+                add_info(meta_list, info_list, metadata, info)
+            else:
+                for info_f in info_list:
+                    info[info_f] = self._preferredinfo[preferred_key]
```
dstftw: 1. No.
2. This won't have any effect anyway since post processors operate on info dict copy.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -431,21 +431,36 @@ def run(self, information):
...
             if not isinstance(info_list, (list, tuple)):
                 info_list = (info_list,)
-            for info_f in info_list:
-                if info.get(info_f) is not None:
-                    for meta_f in meta_list:
-                        metadata[meta_f] = info[info_f]
-                    break
+            preferred_key = add_info(meta_list, info_list, metadata, self._preferredinfo)
+            if preferred_key is None:
+                add_info(meta_list, info_list, metadata, info)
```
dstftw: `merge_dicts`.

katieannbaker: @dstftw My concern with a straight merge dict is if the info uses a different key from the info_list than the preferredinfo uses (e.g. info['track'] = 'a' and preferredinfo['title'] = 'b') - if they were merged without knowledge to the info_list, metadata['title'] would result in the track name from info instead of the title name from prefferredinfo. There are two options:

1. Leave as is
2. modify the preferredinfo object (on each add function call) to add the keys for all elements within the info_list, then preform a dictionary merge within each add onto a copied info

Let me know your preference and I will implement it.

katieannbaker:
@dstftw I've incorporated your updates, and have added this functionality for the CLI, as well as the README markdown for the additional option. I'm awaiting your feedback on how to proceed with the 'merge_dicts' feedback.

katieannbaker:
@dstftw following up on this review request.

youtube_dl/postprocessor/ffmpeg.py
dstftw: This is expected behavior, you named it preferredinfo not preferredmetadatatoembed. `track` from info must be preferred over `title` from preferredinfo.
Again: you merge `_preferredinfo` with `info` into separate merged info **not touching** the initial dicts then operate with old code on merged info instead of `info`.

README.md
```diff
@@ -409,6 +409,11 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
                                      webm and mkv videos)
     --embed-thumbnail                Embed thumbnail in the audio as cover art
     --add-metadata                   Write metadata to the video file
+    --preferred-info                 References a JSON file that contains the
+                                     metadata that will be used preferentially
+                                     over the extracted metadata and upated onto
+                                     the output file. This argument will only be
+                                     utilized when add-metadata is used
```
dstftw: This is autogenerated, don't touch this.

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -431,21 +431,34 @@ def run(self, information):
 
 
 class FFmpegMetadataPP(FFmpegPostProcessor):
+    def __init__(self, downloader=None, preferredinfo=None):
+        super(FFmpegMetadataPP, self).__init__(downloader)
+        if not (isinstance(preferredinfo, dict) or preferredinfo == None):
+            raise TypeError('preferredinfo must be a dictionary, if provided')
```
dstftw: Have you read previous comment at all?

youtube_dl/options.py
```diff
@@ -818,6 +818,10 @@ def _comma_separated_values_options_callback(option, opt_str, value, parser):
         '--add-metadata',
         action='store_true', dest='addmetadata', default=False,
         help='Write metadata to the video file')
+    postproc.add_option(
+        '--preferred-info',
+        dest='preferredinfo', metavar='FILE',
+        help='Override metadata on the outputted file')
```
dstftw: No, that won't work, nobody will bother crafting JSON file just to just override single info meta field. There should be interface for providing overrides for a single meta field from command line in the first place.

youtube_dl/postprocessor/ffmpeg.py
katieannbaker: @dstftw I will rename the object to preferredmetadatatoembed and propagate throughout the code as this is the functionality that I believe should be implemented.

dstftw: No. In this case the whole new notion of embeddable meta fields should be introduced.

youtube_dl/postprocessor/ffmpeg.py
katieannbaker: @dstftw I'm following an observed pattern (lines 352-355 and 586-589) and raising an exception on the input when necessary. Can you provide a pattern that you're expecting me to follow for this if these examples within the existing code base is not what you're looking for?

`class FFmpegVideoConvertorPP(FFmpegPostProcessor):
    def __init__(self, downloader=None, preferedformat=None):
        super(FFmpegVideoConvertorPP, self).__init__(downloader)
        self._preferedformat = preferedformat`

`class FFmpegSubtitlesConvertorPP(FFmpegPostProcessor):
    def __init__(self, downloader=None, format=None):
        super(FFmpegSubtitlesConvertorPP, self).__init__(downloader)
        self.format = format`

Or are you looking for unit tests to be added for this method?

youtube_dl/options.py
katieannbaker: @dstftw I'll add functionality for individual metadata overrides in additional to supporting a file based JSON override

youtube_dl/postprocessor/ffmpeg.py
dstftw: https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

katieannbaker: @dstftw assertions are only executed during debug and not during runtime. I added an exception error so that it is in effect 100% of time. Are you saying that your preference is that errors are not thrown consistently?

From the documentation you provided, assert is equivalent to:
`if __debug__:
   if not expression: raise AssertionError`

youtube_dl/postprocessor/ffmpeg.py
katieannbaker: This entire feature is about metadata override. The functionality is in place to do that. The name of the feature, the description of the feature, the class that it is being introduce into (FFmpegMetadataPP) are all about metadata. The only thing that is not in alignment with it is the name of the input parameter. What additional functionality and/or changes need to be put in place to allow for this to be the nature of the feature?

youtube_dl/postprocessor/ffmpeg.py
dstftw: I guess you misunderstand the whole point of assertions. You should distinguish cases when data comes externally and internally. In this case the latter takes place. The input data is build by our own code internally therefore we ourselves are responsible for guaranteeing its validity so that it's pointless to always perform this runtime check. If it turns out to be invalid then it's a bug in our code that is idiomatically as precondition covered by assert in debug.

katieannbaker: @dstftw A. I'm not seeing assertions in ffmpeg at all, so this pattern is not followed and B. I'm not seeing anywhere that datatypes are validated. This is an argument passed in externally. There isn't a pattern at this validation happens further up in the service. I'm happy to swap it out for an assert, but without upstream validation, it is an assert on an external data element

youtube_dl/postprocessor/ffmpeg.py
dstftw: It's about metadata however it does not reveal anything about its internals leaving it implementation detail and providing us more freedom. Once you start allowing to override concrete embeddable meta fields you must at least standardize and document the list of such fields in order for user to know what to actually override. That's actually not what I planned to maintain.

katieannbaker: @dstftw So the lists of metadata that are hard-coded within the Class right now are the same elements that this functionality allows to be overridden - no more and no less. My assumption was that this list was already being maintained?

youtube_dl/postprocessor/ffmpeg.py
dstftw: Again: asserts are for **invariant checks** during **development**. Once you're done you may remove them freely that's why there are almost no asserts in code base. The point is that there is no need to throw in such cases. If you want to check while developing add an assert. If you're sure enough you won't stumble while developing don't use asserts at all.

youtube_dl/postprocessor/ffmpeg.py
dstftw: No. As already said it's an implementation detail and may be changed any time.

katieannbaker: @dstftw Ok - I understand. Is there any compromise where I can make the help method for the CLI argument dynamically expose the list of meta-data fields that are available for override?

katieannbaker:
@dstftw Please see additions that add dynamic auto-documentation for which metadata elements can be overwritten by the FFmpegMetadataPP class.

PR #22309, opened 2019-09-04
[chaturbate] Allow alternative URLs at /fullvideo/

axelerometer:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

There are alternative Chaturbate URLs such as https://chaturbate.com/fullvideo/?b=caylin, which were previously not supported; this PR adds support for /fullvideo/ URLs.

### DIFF
--- a/youtube_dl/extractor/chaturbate.py
+++ b/youtube_dl/extractor/chaturbate.py
@@ -33,10 +33,10 @@ class ChaturbateIE(InfoExtractor):
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
-        url = 'https://chaturbate.com/%s/' % video_id
 
         webpage = self._download_webpage(
-            url, video_id, headers=self.geo_verification_headers())
+            'https://chaturbate.com/%s/' % video_id, video_id,
+            headers=self.geo_verification_headers())
 
         m3u8_urls = []
 
--- a/youtube_dl/extractor/chaturbate.py
+++ b/youtube_dl/extractor/chaturbate.py
@@ -7,7 +7,7 @@ from ..utils import ExtractorError
 
 
 class ChaturbateIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:[^/]+\.)?chaturbate\.com/(?P<id>[^/?#]+)'
+    _VALID_URL = r'^https?://(?:[^/]+\.)?chaturbate\.com/(?:fullvideo/\?b=)?(?P<id>[^/?&#]+)'
     _TESTS = [{
         'url': 'https://www.chaturbate.com/siswet19/',
         'info_dict': {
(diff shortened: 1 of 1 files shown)


axelerometer:
I also ran the extractor's tests, by substituting the room name with another one, and commenting out `skip`. In addition to testing by hand.

youtube_dl/extractor/chaturbate.py
```diff
@@ -21,6 +21,19 @@ class ChaturbateIE(InfoExtractor):
             'skip_download': True,
         },
         'skip': 'Room is offline',
+    }, {
+        'url': 'https://chaturbate.com/fullvideo/?b=caylin',
```
dstftw: Only matching.

youtube_dl/extractor/chaturbate.py
```diff
@@ -55,6 +68,15 @@ def _real_extract(self, url):
                     error = self._ROOM_OFFLINE
             if error:
                 raise ExtractorError(error, expected=True)
+
+            # Try the canonical URL (/fullvideo/?b=X canonical points to /X/)
+            canon_url = self._search_regex(
+                r'<link[^>]+rel=(["\']?)canonical\1[^>]+href=(["\'])(?P<url>[^>"\']+)\2',
+                webpage, 'canonical URL', group='url')
```
dstftw: No. Build initial URL from id.

youtube_dl/extractor/chaturbate.py
```diff
@@ -55,6 +68,15 @@ def _real_extract(self, url):
...
+            canon_url = urljoin(url, canon_url)
+            if canon_url != url and re.match(self._VALID_URL, canon_url):
+                return self._real_extract(canon_url)
```
dstftw: No.

youtube_dl/extractor/chaturbate.py
axelerometer: Doh, that's far simpler. Updated.

axelerometer:
@dstftw Thanks for the quick response. I fixed all the review items, although GitHub still displays "1 change requested".

axelerometer:
@dstftw Ping, any chance for re-review?

youtube_dl/extractor/chaturbate.py
```diff
@@ -7,7 +7,7 @@
 
 
 class ChaturbateIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:[^/]+\.)?chaturbate\.com/(?P<id>[^/?#]+)'
+    _VALID_URL = r'https?://(?:[^/]+\.)?chaturbate\.com/(?:fullvideo/\?b=)?(?P<id>[^/?&#]+)'
```
dstftw: Won't match https://chaturbate.com/fullvideo/?foo&b=... and similar.

axelerometer: It actually does, because the regexp is not anchored to end of string (does not end with `$`).

But now that you mention it, I added `^` to the beginning of this regex.

dstftw: Nothing to do with my comment. Have you read it at all?

axelerometer: Sorry, my bad, I misread your comment. But yes, I think that still counts as reading. Anyway it's a moot point now, thanks! :)

PR #22660, opened 2019-10-09
[hhu] Add new extractor

YtvwlD:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

[HHU Mediathek](https://mediathek.hhu.de) contains videos created by the Heinrich-Heine-Universität Düsseldorf (mostly recordings of lectures).
This extractor allows the download of single, publicly visible videos.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -2750,7 +2750,7 @@ class InfoExtractor(object):
                     # Often no height is provided but there is a label in
                     # format like "1080p", "720p SD", or 1080.
                     height = int_or_none(self._search_regex(
-                        r'^(\d{3,4})[pP]?(?:\b|$)', compat_str(source.get('label') or ''),
+                        r'(\d{3,4})[pP]?(?:\b|$)', compat_str(source.get('label') or ''),
                         'height', default=None))
                 a_format = {
                     'url': source_url,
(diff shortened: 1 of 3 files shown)


dstftw:
Read coding conventions.

YtvwlD:
I read the coding conventions before but I didn't notice the rule regarding closing braces/brackets/parentheses. (I'm used to the guidelines in PEP8, sorry about that.)

dstftw:
It does not say anything about brackets and braces. Also mostly it's nothing to do with closing parentheses which are perfectly valid PEP8.
As for extractor: jwplayer is already covered by generic extractor that should be extended to support such cases.

YtvwlD:
The generic extractor doesn't work here because `_find_jwplayer_data` doesn't recognize the player config because the used regex [fails](https://regexr.com/4mhpb) at the `encodeURI` call.

I modified my PR to at least use `_parse_jwplayer_data`.
Also, I had to touch the regex in `_parse_jwplayer_formats` to also accept format descriptions that are not at the beginning of the label text. I really hope that this won't break existing stuff - I tested a few extractors using these functions, but most of them seem have broken already before my change.

dstftw:
As already said jwplayer extraction code should be extended to support sanitizing function calls out of options JSON. HHU extractor must be removed. Extraction should be processed by generic extractor.

YtvwlD:
For finding the options there is already #21132 which works in this case (but isn't merged yet).

So, I'm guessing I should close this PR and open two new ones - one for dada9f6db9427f0e8738ecf6551a9a0fc865bbb0 and the other one for parsing this JS?

Issue #22094, opened 2019-08-14
10PLAY SOLUTION in #21446


klan30destine:
## Checklist

- [x ] I'm asking a question
- [ ] I've looked through the README and FAQ for similar questions
- [ ] I've searched the bugtracker for similar questions including closed ones

## Question

WRITE QUESTION HERE
10PLAY SOLUTION IN #21446. 
Just a note to indicate that an intermediate solution to the 10Play unsupported URL problem was added to Issue 21446. It contains some mods to the Youtube-dl code  that can be used as a temporary solution until the production code can be updated with the official solution.
Issue 21450 has some interim solutions for this problem as well.
Klan30destine

Issue #15980, opened 2018-03-24
Unable to download Vimeo private / password protected videos


emma-goldman:
Posted here and got a short unconstructive response:

https://github.com/rg3/youtube-dl/issues/7169

I can log in to the private video in my browser (using the same login info I copy pasted into yt-dl) and stream the video but I am not able to download the video using yt-dl so the answer "wrong login info" does not apply. @dstftw 

Previous post below:

I'm encountering this issue again, verbose output below:

`PS C:\Users\Xyz\Videos> youtube-dl https://vimeo.com/xxxxxxxx --video-password xxx -u 'PRIVATE' --verbose
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'https://vimeo.com/xxxxxxxx', u'--video-password', u'PRIVATE', u'-u', u'PRIVATE', u'--verbose']
Type account password and press [Return]:
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2018.03.03
[debug] Python version 2.7.14 (CPython) - Windows-10-10.0.16299
[debug] exe versions: ffmpeg 3.4.1, ffprobe 3.4.1
[debug] Proxy map: {}
[vimeo] Logging in
ERROR: Wrong login info: HTTP Error 418: I'm a teapot (caused by HTTPError()); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see https://yt-dl.org/update on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
File "c:\python27\lib\site-packages\youtube_dl\extractor\common.py", line 519, in _request_webpage
return self._downloader.urlopen(url_or_request)
File "c:\python27\lib\site-packages\youtube_dl\YoutubeDL.py", line 2199, in urlopen
return self._opener.open(req, timeout=self._socket_timeout)
File "c:\python27\lib\urllib2.py", line 435, in open
response = meth(req, response)
File "c:\python27\lib\urllib2.py", line 548, in http_response
'http', request, response, code, msg, hdrs)
File "c:\python27\lib\urllib2.py", line 473, in error
return self._call_chain(*args)
File "c:\python27\lib\urllib2.py", line 407, in _call_chain
result = func(*args)
File "c:\python27\lib\urllib2.py", line 556, in http_error_default
raise HTTPError(req.get_full_url(), code, msg, hdrs, fp)

`or

[...]

PR #22683, opened 2019-10-11
[Vimeo] Adds a specific hint for which param to use when a video is password protected

1RedOne:
#22094  Please follow the guide below

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

With Vimeo vids, we see lots of people raising issues (like this one, #15980 ) when they are trying to provide a -Password param, which is the wrong choice for the Vimeo domain for password protected videos.  If a user tries to download a password protected video and provides no password, there is a UI hint in the error message for that...but not if they use the patently wrong param like this.

This patch adds a specific hint on the Vimeo wrong username/password scenario to mitigate this in the future, like #15980

### DIFF
--- a/youtube_dl/extractor/vimeo.py
+++ b/youtube_dl/extractor/vimeo.py
@@ -68,7 +68,7 @@ class VimeoBaseInfoExtractor(InfoExtractor):
         except ExtractorError as e:
             if isinstance(e.cause, compat_HTTPError) and e.cause.code == 418:
                 raise ExtractorError(
-                    'Unable to log in: bad username or password',
+                    'Unable to log in: bad username or password, use the --video-password option for password protected videos',
                     expected=True)
             raise ExtractorError('Unable to log in')
 


dstftw:
If user downloads a password protected video or an album without providing a password he will be [explicitly](https://github.com/ytdl-org/youtube-dl/blob/c317b6163b294f4cdc2d1dff96e1a63da1bae910/youtube_dl/extractor/vimeo.py#L78) [prompted](https://github.com/ytdl-org/youtube-dl/blob/c317b6163b294f4cdc2d1dff96e1a63da1bae910/youtube_dl/extractor/vimeo.py#L526) [about that](https://github.com/ytdl-org/youtube-dl/blob/c317b6163b294f4cdc2d1dff96e1a63da1bae910/youtube_dl/extractor/vimeo.py#L836). No need to a create mess in unrelated places.

1RedOne:
Check out the related issue, people have mistakenly thought that -Password was the way to go (myself included).  In one of those threads, you yourself had to direct people on the right param to use.

Why not display a hint to the user in such a case?

dstftw:
Why do you even think writing this hint in another place will make any difference? If one does not bother reading output he won't read it from here either.
_mistakenly thought_ is a really bad excuse. What if one'd _mistakenly thought_ that `--username` is the way to go for custom user agent? Or similarly `--password` for `--ap-password`? Or more terminal scenarios? Make another additions to hint for these cases? What about another extractors? Add these hints to all other extractors with authentication? No, never happening.

PR #22773, opened 2019-10-20
Check nooverwrite in case postprocessors are to be applied

righthandabacus:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Usually I do something like this:

    python3 -m youtube_dl -w -x --audio-format=mp3 https://youtu.be/s0m3th1ng

and what it does behind the scene is usually download the video (e.g., in webm format), then run ffmpeg to convert the video into mp3, and then the webm file is deleted. If I run this command once again, the whole process goes over once more and effectively the `-w` (no overwrite) flag is ignored.

Indeed, I can avoid downloading again only if I did not remove the webm file. In this patch, at before proceed to download the webm file, we guess what the file would be (in the above example, it means the same filename but with .mp3 extension) and do `os.path.exists()` check to avoid download once the postprocessed file is believed to be exist.

I am not sure this is the best approach to do, especially on the filename guessing part. Comments and suggestions welcomed.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1922,6 +1922,18 @@ class YoutubeDL(object):
                         info_dict['__files_to_merge'] = downloaded
                 else:
                     # Just a single file
+                    if self.params.get('nooverwrites', False) and self.params['postprocessors']:
+                        # check if the postprocessed files already exists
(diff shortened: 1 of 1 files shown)


dstftw:
`--download-archive`.

PR #22844, opened 2019-10-25
[Bandcamp] Update extractor to match website changes

lyz-code:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

Bandcamp must have changed it's code because pulling from master made the first test fail.

I needed to change `(?:(?!\1).)` to `.` in the extract function because the artist name of the first test has a `"` and it wasn't captured otherwise.

### DIFF
--- a/youtube_dl/extractor/bandcamp.py
+++ b/youtube_dl/extractor/bandcamp.py
@@ -121,8 +121,7 @@ class BandcampIE(InfoExtractor):
         def extract(key):
             return self._search_regex(
                 r'\b%s\s*["\']?\s*:\s*(["\'])(?P<value>.+)\1' % key,
-                webpage, key, default=None, group='value'
-            )
+                webpage, key, default=None, group='value')
 
         artist = extract('artist')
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/bandcamp.py
```diff
@@ -117,7 +120,7 @@ def _real_extract(self, url):
 
         def extract(key):
             return self._search_regex(
-                r'\b%s\s*["\']?\s*:\s*(["\'])(?P<value>(?:(?!\1).)+)\1' % key,
+                r'\b%s\s*["\']?\s*:\s*(["\'])(?P<value>.+)\1' % key,
```
dstftw: No, as this will work incorrectly for empty `""` values.

lyz-code: I've made the modifications so it works as expected for empty values and with strings with `"` in the value.

lyz-code:
It seems that the master branch works as expected now

Issue #22897, opened 2019-10-29
SVT news not working


robalni:
## Checklist

- [x] I'm reporting a broken site support
- [x] I've verified that I'm running youtube-dl version **2019.10.29**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
$ youtube-dl -v 'https://www.svt.se/nyheter/lokalt/norrbotten/komplicerat-att-stanga-mr-kameran'
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', 'https://www.svt.se/nyheter/lokalt/norrbotten/komplicerat-att-stanga-mr-kameran']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2019.10.29
[debug] Python version 3.7.3 (CPython) - Linux-4.19.0-6-amd64-x86_64-with-debian-10.1
[debug] exe versions: ffmpeg 4.1.4-1, ffprobe 4.1.4-1, phantomjs 2.1.1, rtmpdump 2.4
[debug] Proxy map: {}
[SVTPage] komplicerat-att-stanga-mr-kameran: Downloading webpage
[download] Downloading playlist: MR-kameran fortfarande igång – komplicerad avstängning påbörjad
[SVTPage] playlist MR-kameran fortfarande igång – komplicerad avstängning påbörjad: Collected 0 video ids (downloading 0 of them)
[download] Finished downloading playlist: MR-kameran fortfarande igång – komplicerad avstängning påbörjad
```

## Description

SVT news videos have recently stopped working.

Example that doesn't work: https://www.svt.se/nyheter/lokalt/norrbotten/komplicerat-att-stanga-mr-kameran

The video that should be played can be found in this JSON file: https://api.svt.se/videoplayer-api/video/24256474

The id 4256474 can be found in the source of the page.

Here is an example with multiple videos on one page: https://www.svt.se/nyheter/utrikes/svenska-andrea-ar-en-mil-fran-branderna-i-kalifornien

The video ids are 24244546 and 24246318.

Rodondo2000:
I do not know whether this is the same problem as described in #22919 . As it too is about SVT.se I put it in here to avoid redundancy. 

Description of the problem:

[...]

PR #22919, opened 2019-10-30
[svt] fix issue #22897 : change process of video id extraction

uno20001:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This PR fixes issue #22897 by changing the process of extracting video ids.

### DIFF
--- a/youtube_dl/extractor/svt.py
+++ b/youtube_dl/extractor/svt.py
@@ -12,11 +12,11 @@ from ..utils import (
     determine_ext,
     dict_get,
     int_or_none,
-    orderedSet,
     strip_or_none,
     try_get,
     urljoin,
     compat_str,
+    js_to_json,
 )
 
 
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/svt.py
```diff
@@ -360,11 +364,26 @@ def _real_extract(self, url):
 
         webpage = self._download_webpage(url, playlist_id)
 
+        redux_state = self._parse_json(
+            self._search_regex(
+                r'\.\s*reduxState=\s*({.*})',
+                webpage,
+                'reduxState'),
+            None,
+            transform_source=js_to_json)
```
dstftw: Remove superfluous verbosity.

youtube_dl/extractor/svt.py
```diff
@@ -360,11 +364,26 @@ def _real_extract(self, url):
 
         webpage = self._download_webpage(url, playlist_id)
 
+        redux_state = self._parse_json(
+            self._search_regex(
+                r'\.\s*reduxState=\s*({.*})',
```
dstftw: Do not capture empty dict.

youtube_dl/extractor/svt.py
```diff
@@ -360,11 +364,26 @@ def _real_extract(self, url):
...
+                webpage,
+                'reduxState'),
+            None,
```
dstftw: None is not an id.

youtube_dl/extractor/svt.py
```diff
@@ -320,17 +320,24 @@ def _real_extract(self, url):
...
+        },
+        'playlist_count': 2,
     }, {
-        'url': 'https://www.svt.se/nyheter/inrikes/ebba-busch-thor-kd-har-delvis-ratt-om-no-go-zoner',
+        'url': 'https://www.svt.se/nyheter/lokalt/norrbotten/komplicerat-att-stanga-mr-kameran',
         'info_dict': {
-            'id': 'ebba-busch-thor-kd-har-delvis-ratt-om-no-go-zoner',
-            'title': 'Ebba Busch Thor har bara delvis rätt om ”no-go-zoner”',
+            'id': 'komplicerat-att-stanga-mr-kameran',
+            'title': 'MR-kameran fortfarande igång – komplicerad avstängning påbörjad',
```
dstftw: Remove duplicate tests.

youtube_dl/extractor/svt.py
```diff
@@ -360,11 +364,26 @@ def _real_extract(self, url):
...
 
+        redux_state = self._parse_json(
+            self._search_regex(
+                r'\.\s*reduxState=\s*({.*})',
+                webpage,
+                'reduxState'),
+            None,
+            transform_source=js_to_json)
+
+        media = try_get(redux_state, lambda x: x['componentState']['modal']['media']['article']['media'], list) or []
```
dstftw: Inline.

youtube_dl/extractor/svt.py
```diff
@@ -360,11 +364,26 @@ def _real_extract(self, url):
...
+        video_ids = []
+
+        for m in media:
+            if m.get('_type') == 'VIDEOCLIP':
+                video_ids.append(int_or_none(m.get('id')))
+
+        video_ids = filter(lambda x: x is not None, video_ids)
```
dstftw: Move inside a loop.

youtube_dl/extractor/svt.py
uno20001: I'm sorry I don't exactly see what you mean. The tests in `SVTPageIE` are all unique.

youtube_dl/extractor/svt.py
uno20001: I apologize, but your intent eludes me again, for I do not understand what you mean by *superfluous verbosity*.

youtube_dl/extractor/svt.py
uno20001: Could you please clarify what you mean by *inline*?

youtube_dl/extractor/svt.py
uno20001: Would a simple `{.+}` suffice as replacement?

youtube_dl/extractor/svt.py
uno20001: Upon further investigation of the *coding conventions*, I assume you mean that the *media* variable should be disposed of, is that correct?

uno20001:
I would really appreciate if I could get answers to my questions, because I still have no idea how I should modify the code.

Issue #23010, opened 2019-11-08
Chaturbate issue (room currently offline)


blastphemererik:
## Checklist

- [x] I'm reporting a broken site support
- [x] I've verified that I'm running youtube-dl version **2019.11.05**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar issues including closed ones

## Verbose log

[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', 'https://chaturbate.com/lovely_monic/']
[debug] Encodings: locale cp1250, fs mbcs, out cp852, pref cp1250
[debug] youtube-dl version 2019.11.05
[debug] Python version 3.4.4 (CPython) - Windows-7-6.1.7601-SP1
[debug] exe versions: ffmpeg N-94953-gef50cf7b32, ffprobe N-94953-gef50cf7b32
[debug] Proxy map: {}
[Chaturbate] lovely_monic: Downloading webpage
ERROR: Room is currently offline
Traceback (most recent call last):
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmp7hc5a0
9v\build\youtube_dl\YoutubeDL.py", line 796, in extract_info
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmp7hc5a0
9v\build\youtube_dl\extractor\common.py", line 530, in extract
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmp7hc5a0
9v\build\youtube_dl\extractor\chaturbate.py", line 61, in _real_extract
youtube_dl.utils.ExtractorError: Room is currently offline

## Description

It doesn't matter if a model is online, youtube-dl always says that the room is offline. Everything was fine today, the problems started a few hours ago. It seems the site got updated.

Ubernoob7:
Having the same issue, sucks :( Hopefully there is a fix soon

pcjamesy:
Same

altDark:
The same issue

cb123321:
Hello, when will official installation be updated with this fix?

silvermindphasar:
the same problem
the site chatrubate now reports that the room is offline but the cam is active and broadcasts

asdasddffgthhjtreresf:
[...]

PR #23011, opened 2019-11-08
[chaturbate] fix url extraction and parsing

throwaway396:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Fixes issue #23010 

The chaturbate extractor broke some time this evening. The regex looking for .m3u8 URLs no longer matched anything. Additionally, the URL now needs a bit of extra processing.

The page source contains a large JavaScript string containing an encoded JSON object. The nested quotes, as well as other special characters, are encoded as e.g. \u0022. So, we match the URL delimited by \u0022 or \u0027 (double or single quotes, respectively) and then decode any \uXXXX sequences in the match group.

### DIFF
--- a/youtube_dl/extractor/chaturbate.py
+++ b/youtube_dl/extractor/chaturbate.py
@@ -41,9 +41,9 @@ class ChaturbateIE(InfoExtractor):
         m3u8_urls = []
 
         for m in re.finditer(
-                r'(["\'])(?P<url>http.+?\.m3u8.*?)\1', webpage):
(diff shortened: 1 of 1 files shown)


throwaway396:
I just realized this doesn't work in Python 2. Not surprising, since it involves string decoding. I'll see if I can make this compatible using some future import or something.

throwaway396:
This patch was not compatible with python 2. I have a solution and will create a new pull request shortly.

PR #23486, opened 2019-12-23
README.md: clarify how to combine + and /

ProducerMatt:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

A while back I wanted to use both merging and precedence in format
selection. I encountered complaints from the program and eventually
figured out I was doing it backwards.

I was trying things like this:

`-f 136/mp4/bestvideo+140/m4a/bestaudio`

When what I needed to do was:

`-f 136+140/mp4+m4a/bestvideo+bestaudio`

I wrote a couple lines to try to clarify this. I hope this is helpful.

--Matt


dstftw:
You must use parentheses as clearly described in the next paragraph instead of rewriting it to another expression (that is non equal in your case by the way).

Issue #23191, opened 2019-11-24
Generic Extractor should add short-hash of URL to filename


d3fault:
## Checklist

- [x] I'm reporting a broken site support issue
- [x] I've verified that I'm running youtube-dl version **2019.11.22**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar bug reports including closed ones
- [x] I've read bugs section in FAQ

## Verbose log

```
n/a
```

## Description

Currently the generic extractor relies on the HTML Title and the last portion of the URL to generate a filename.

Ex:
`youtube-dl https://www.w3.org/2010/05/video/mediaevents.html`
gets you a file named:
`HTML5 Video-mediaevents.mp4`

This strategy could easily give filename collisions which result in an erroneous response of "this file has already been downloaded". The filename generated should contain a shorthash (first 8 characters of sha1) of the entire URL, or something similar.

Ex of what the filename should be:
`HTML5 Video-mediaevents-fbea3c2c.mp4`

PR #23589, opened 2020-01-01
Append sha1 hash of url to filename in Generic Extractor

mushifali:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Generic Extractor now appends sha1 hash of URL to filename to avoid filename collisions thus closing #23191.

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -2,6 +2,7 @@
 
 from __future__ import unicode_literals
 
+import hashlib
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


d3fault:
why is this "out-of-scope"? this pull lgtm (but then again I don't write python).

will #23191 ever be fixed? I've encountered real scenarios described in that bug report, it's not just theoretical.

mushifali:
@d3fault I agree! It would be useful for some real scenarios that you encountered. But unfortunately, this pull was closed immediately with "out-of-scope" label.

dstftw:
Id must be stable for identical URLs.

d3fault:
@dstftw sha1 is stable... you put a single URL through it multiple times, you always get same result. am I missing something?

dstftw:
>for identical URLs

d3fault:
@dstftw sha1 is stable for identical URLs

dstftw:
No, it's not. sha1 as well as any other hash is only stable for exactly same strings.

PR #23599, opened 2020-01-02
fixed error when parsing fm4 url

jonolt:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

removed if clause for special fm4 show_id handling, as url has been changed to a new format where the station is natively correct.

updated only test url of ORFRadio Extractors. While other extractors are not updated because they worked with a simple test in the console with random current links and the links would get invalid again quickly anyway in a short time.

### DIFF
--- a/youtube_dl/extractor/orf.py
+++ b/youtube_dl/extractor/orf.py
@@ -192,7 +192,7 @@ class ORFRadioIE(InfoExtractor):
 class ORFFM4IE(ORFRadioIE):
     IE_NAME = 'orf:fm4'
     IE_DESC = 'radio FM4'
-    _VALID_URL = r'https?://(?P<station>fm4)\.orf\.at/player/(?P<date>[0-9]+)/(?P<show>\w+)'
+    _VALID_URL = r'https?://(?P<station>fm4)\.orf\.at/player/(?P<date>[0-9]+)/(?P<show>4\w+)'
 
     _TEST = {
         'url': 'https://fm4.orf.at/player/20200101/4HOP',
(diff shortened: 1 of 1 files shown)


dstftw:
You have no guarantees that it will always start with 4.

jonolt:
you are right, i can't guarantee it, but
- all shows I used to download did never start with 4. but now do
- i checked different shows at multiple days which all start with 4
- i have recognized the update of the site by some minor changes on the interface, as well as a changed forwarding of expired links
- i can't use a if statement to check (see code block), because i don't know for sure that there is (will be) no show name that starts with "4". Which will yield the `show_id='4XY'` while the correct value is (may be) `show_id='44XY'`.
```pyhton
# cant do something like this:
if station == 'fm4':
    if not show_id.startswith("4"):
        show_id = '4%s' % show_id
```

dstftw:
URLs without 4 prefix should not be matched in the first place,

jonolt:
Then everything fits, or **did I misunderstand something?**

The current version inserts a "4" before station_id. 
```
OLD URL: https://fm4.orf.at/player/20191113/HOP

show_id = "HOP"
if station == 'fm4':
    show_id = '4%s' % show_id
show_id = "4HOP"
```
My commit removes this, because it is no longer necessary, since the URL-scheme of the website itself has changed and the "4" is already in the URL.

```
OLD URL: https://fm4.orf.at/player/20191113/HOP
NEW URL: https://fm4.orf.at/player/20200101/4HOP

show_id = "4HOP"
# no manipulation needed any more
```

dstftw:
No, it's not. Again: if you claim id to always start with 4 then extractor should not match URLs with ids not starting with 4.

jonolt:
> URLs without 4 prefix should not be matched in the first place.

I think I got it now :sweat_smile:. Updated the _VALID_URL regex string.

Issue #23200, opened 2019-11-25
eroprofile.com site updates from ~2wks back broke youtube-dl


ta-ytdl:
## Checklist

- [X] I'm reporting a broken site support
- [X] I've verified that I'm running youtube-dl version **2019.11.22**
- [X] I've checked that all provided URLs are alive and playable in a browser
- [X] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [X] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', 'https://www.eroprofile.com/m/videos/view/Wow-574']
[debug] Encodings: locale UTF-8, fs utf-8, out utf-8, pref UTF-8
[debug] youtube-dl version 2019.11.22
[debug] Python version 3.8.0 (CPython) - Linux-5.3.12-arch1-1-x86_64-with-glibc2.2.5
[debug] exe versions: ffmpeg 4.2.1, ffprobe 4.2.1, rtmpdump 2.4
[debug] Proxy map: {}
[EroProfile] Wow-574: Downloading webpage
ERROR: Unable to extract title; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/youtube_dl/YoutubeDL.py", line 796, in extract_info
    ie_result = ie.extract(url)
  File "/usr/lib/python3.8/site-packages/youtube_dl/extractor/common.py", line 530, in extract
    ie_result = self._real_extract(url)
  File "/usr/lib/python3.8/site-packages/youtube_dl/extractor/eroprofile.py", line 82, in _real_extract
    title = self._html_search_regex(
  File "/usr/lib/python3.8/site-packages/youtube_dl/extractor/common.py", line 1014, in _html_search_regex
    res = self._search_regex(pattern, string, name, default, fatal, flags, group)
  File "/usr/lib/python3.8/site-packages/youtube_dl/extractor/common.py", line 1005, in _search_regex
    raise RegexNotFoundError('Unable to extract %s' % _name)
[...]

PR #23626, opened 2020-01-04
[eroprofile] Update regexes

YetAnotherGhost:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

It updates the regular expressions following the eroprofile.com site update pointed out in #23200. The title is hopefully stable as long as they stick to `<h1>` elements; the thumbnail is a bit more fragile, but at least it's optional. (I could only find it in one place on the page, in the playlist bar above the video, together with other thumbnails.)

Note that the first video in `_TESTS` appears to no longer exist, but I assume that's unrelated.

### DIFF
--- a/youtube_dl/extractor/eroprofile.py
+++ b/youtube_dl/extractor/eroprofile.py
@@ -82,7 +82,7 @@ class EroProfileIE(InfoExtractor):
         title = self._html_search_regex(
             [r'<h1[^>]*>([^<]+)</h1>', r'Title:</th><td>([^<]+)</td>'],
             webpage, 'title')
-        thumbnail = self._search_regex(
+        thumbnail = self._html_search_regex(
             [r'<div class="playlistItem current">.*?<img src="([^"]+)"', r'onclick="showVideoPlayer\(\)"><img src="([^"]+)'],
             webpage, 'thumbnail', fatal=False)
 
--- a/youtube_dl/extractor/eroprofile.py
+++ b/youtube_dl/extractor/eroprofile.py
@@ -83,7 +83,7 @@ class EroProfileIE(InfoExtractor):
             [r'<h1[^>]*>([^<]+)</h1>', r'Title:</th><td>([^<]+)</td>'],
             webpage, 'title')
         thumbnail = self._search_regex(
-            [r'<div class="playlistItem current">.*<img src="([^"]+)"', r'onclick="showVideoPlayer\(\)"><img src="([^"]+)'],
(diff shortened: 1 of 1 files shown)


dstftw:
Keep old regexes as fallbacks.

YetAnotherGhost:
Done (as separate commit, since I'm using the web UI). I hope the lines aren't too long.

youtube_dl/extractor/eroprofile.py
```diff
@@ -80,9 +80,10 @@ def _real_extract(self, url):
         video_url = unescapeHTML(self._search_regex(
             r'<source src="([^"]+)', webpage, 'video url'))
         title = self._html_search_regex(
-            r'Title:</th><td>([^<]+)</td>', webpage, 'title')
+            [r'<h1[^>]*>([^<]+)</h1>', r'Title:</th><td>([^<]+)</td>'],
+            webpage, 'title')
         thumbnail = self._search_regex(
-            r'onclick="showVideoPlayer\(\)"><img src="([^"]+)',
+            [r'<div class="playlistItem current">.*<img src="([^"]+)"', r'onclick="showVideoPlayer\(\)"><img src="([^"]+)'],
```
dstftw: Part in the middle should not be greedy. Also use `_parse_html5_media_entries` as main path.

YetAnotherGhost: > Part in the middle should not be greedy.

Done, thanks.

> Also use `_parse_html5_media_entries` as main path.

I'm not sure what you mean… but this change doesn't touch the `video_url` so far?

YetAnotherGhost:
What’s the status on this? I’m done at the moment, as far as I’m concerned – if you’re waiting for anything else from me, let me know what’s still missing before merging this.

dstftw:
I've already pointed out to use `_parse_html5_media_entries` as main path for formats and thumbnail extraction.

YetAnotherGhost:
Yes, and I already said (or tried to say, I guess it wasn’t clear) that a) I wasn’t sure what you meant by that (the “for formats and thumbnail extraction” makes it a bit clearer, thanks), and b) if it means what I think it means, I don’t think it belongs in this pull request, since it’s unrelated beyond affecting the same extractor.

I tried to implement this in YetAnotherGhost/youtube-dl#1 (was supposed to be a pull request against this repository but I didn’t pay attention when creating the pull request; can be moved later). Is it more or less what you were thinking of?

PR #23657, opened 2020-01-07
[PeerTube] Various fixes and improvements

3risian:

- [X] Bug fix
- [X] Improvement
- [ ] New extractor
- [ ] New feature

---

Adds various functionalities and fixes to the PeerTube extractor. This is immediately important for PeerTube's 'Import with URL' feature for users importing their own videos to new instances.


youtube_dl/extractor/peertube.py
```diff
@@ -492,6 +493,12 @@ def _real_extract(self, url):
         video = self._download_json(
             'https://%s/api/v1/videos/%s' % (host, video_id), video_id)
 
+        video_description = self._download_json(
+            'https://%s/api/v1/videos/%s/description' % (host, video_id), video_id)
```
dstftw: Must not be fatal.

youtube_dl/extractor/peertube.py
```diff
@@ -492,6 +493,12 @@ def _real_extract(self, url):
...
+
+        video_captions = self._download_json(
+            'https://%s/api/v1/videos/%s/captions' % (host, video_id), video_id)
```
dstftw: Must not be extracted if no subtitles requested.

dstftw: Must be after title extraction.

youtube_dl/extractor/peertube.py
```diff
@@ -513,6 +520,13 @@ def _real_extract(self, url):
             formats.append(f)
         self._sort_formats(formats)
 
+        subtitles = {}
+        for entry in video_captions['data']:
```
dstftw: Must not be fatal. Read coding conventions.

youtube_dl/extractor/peertube.py
```diff
@@ -513,6 +520,13 @@ def _real_extract(self, url):
...
+            caption_url = 'https://%s%s' % (host, entry['captionPath'])
```
dstftw: Same.

youtube_dl/extractor/peertube.py
```diff
@@ -513,6 +520,13 @@ def _real_extract(self, url):
...
+            subtitles[entry['language']['id']] = [{
```
dstftw: Same.

dstftw: Does not handle multiple subtitle formats per language scenario.

youtube_dl/extractor/peertube.py
```diff
@@ -528,12 +542,14 @@ def account_data(field):
...
             'id': video_id,
             'title': title,
-            'description': video.get('description'),
+            'description': video_description.get('description'),
             'thumbnail': urljoin(url, video.get('thumbnailPath')),
             'timestamp': unified_timestamp(video.get('publishedAt')),
             'uploader': account_data('displayName'),
-            'uploader_id': account_data('uuid'),
-            'uploder_url': account_data('url'),
+            'uploader_id': '%s@%s' % (account_data('name'), account_data('host')),
```
dstftw: That's not a id.

youtube_dl/extractor/peertube.py
3risian: PeerTube does not currently enable this scenario, as only one subtitle file may exist for each language. Should I still change this code?

youtube_dl/extractor/peertube.py
3risian: There appears to be no in-built globally-unique id that is consistent across PeerTube instances.
If this is not an issue, I can use the int 'id' instead.

dstftw: It's not supposed to be globally unique. Read meta field description.

youtube_dl/extractor/peertube.py
```diff
@@ -528,12 +551,14 @@ def account_data(field):
         return {
             'id': video_id,
             'title': title,
-            'description': video.get('description'),
+            'description': video_description.get('description'),
```
dstftw: Breaks if description request fails.

youtube_dl/extractor/peertube.py
```diff
@@ -484,6 +485,23 @@ def _extract_urls(webpage, source_url):
                 entries = [peertube_url]
         return entries
 
+    def _get_subtitles(self, host, video_id):
+        video_captions = self._download_json(
+            'https://%s/api/v1/videos/%s/captions' % (host, video_id), video_id)
+
+        subtitles = {}
+        for entry in video_captions.get('data'):
```
dstftw: Breaks.

youtube_dl/extractor/peertube.py
```diff
@@ -484,6 +485,23 @@ def _extract_urls(webpage, source_url):
                 entries = [peertube_url]
         return entries
 
+    def _get_subtitles(self, host, video_id):
+        video_captions = self._download_json(
+            'https://%s/api/v1/videos/%s/captions' % (host, video_id), video_id)
```
dstftw: Must not be fatal.

youtube_dl/extractor/peertube.py
```diff
@@ -484,6 +485,23 @@ def _extract_urls(webpage, source_url):
...
+
+        subtitles = {}
+        for entry in video_captions.get('data'):
+            captions_language = entry.get('language')
```
dstftw: str_or_none

youtube_dl/extractor/peertube.py
```diff
@@ -484,6 +485,23 @@ def _extract_urls(webpage, source_url):
...
+            if captions_language is not None:
```
dstftw: `if not captions_language`

youtube_dl/extractor/peertube.py
```diff
@@ -484,6 +485,23 @@ def _extract_urls(webpage, source_url):
...
+                language_id = captions_language.get('id')
```
dstftw: Same.

youtube_dl/extractor/peertube.py
```diff
@@ -484,6 +485,23 @@ def _extract_urls(webpage, source_url):
...
+                caption_path = entry.get('captionPath')
```
dstftw: Same.

youtube_dl/extractor/peertube.py
```diff
@@ -484,6 +485,23 @@ def _extract_urls(webpage, source_url):
...
+                if language_id is not None and caption_path is not None:
```
dstftw: Same.

youtube_dl/extractor/peertube.py
rigelk: Each PeerTube server has a domain and locally ensures user uniqueness, since instances do not coordinate for id generation. Hence, much like with email, the `name@host` is the closest we will get to an id.

@3risian note the `uploader_id` should be the channel's, not the account's.

youtube_dl/extractor/peertube.py
dstftw: You should as already pointed out.

youtube_dl/extractor/peertube.py
3risian: Noted.
The current implementation uses the account name and account URL instead of channel name and URL. Should I change those?

youtube_dl/extractor/peertube.py
```diff
@@ -492,10 +514,10 @@ def _real_extract(self, url):
         video = self._download_json(
             'https://%s/api/v1/videos/%s' % (host, video_id), video_id)
 
-        title = video['name']
+        title = video.get('name')
```
dstftw: No.

youtube_dl/extractor/peertube.py
```diff
@@ -492,10 +514,10 @@ def _real_extract(self, url):
...
 
         formats = []
-        for file_ in video['files']:
+        for file_ in video.get('files'):
```
dstftw: No.

youtube_dl/extractor/peertube.py
dstftw: Nothing changed.

youtube_dl/extractor/peertube.py
```diff
@@ -423,26 +424,27 @@ class PeerTubeIE(InfoExtractor):
...
+            'uploader_url': 'https://framatube.org/accounts/framasoft',
+            'language': 'en',
+            'license': 'Attribution - Share Alike',
+            'duration': 113,
             'view_count': int,
             'like_count': int,
             'dislike_count': int,
-            'tags': list,
-            'categories': list,
+            'tags': ["framasoft", "peertube"],
```
dstftw: Single quotes.

youtube_dl/extractor/peertube.py
```diff
@@ -484,6 +486,26 @@ def _extract_urls(webpage, source_url):
...
+            if language_id and caption_path:
+                caption_url = 'https://%s%s' % (host, caption_path)
+                if language_id in subtitles:
+                    subtitles[language_id].append({
+                        'url': caption_url
+                    })
+                else:
+                    subtitles[language_id] = [{
+                        'url': caption_url
+                    }]
```
dstftw: `subtitles.setdefault`.

youtube_dl/extractor/peertube.py
```diff
@@ -513,9 +535,21 @@ def _real_extract(self, url):
             formats.append(f)
         self._sort_formats(formats)
 
+        video_description = self._download_json(
+            'https://%s/api/v1/videos/%s/description' % (host, video_id), video_id, fatal=False)
+
+        description = ""
+        if video_description:
+            description = video_description.get('description')
```
dstftw: Breaks if `video_description` is not a dict.

youtube_dl/extractor/peertube.py
```diff
@@ -513,9 +535,21 @@ def _real_extract(self, url):
...
+
+        subtitles = self.extract_subtitles(host, video_id)
+
         def account_data(field):
             return try_get(video, lambda x: x['account'][field], compat_str)
 
+        def channel_data(field):
+            return try_get(video, lambda x: x['channel'][field], compat_str)
```
dstftw: DRY.

youtube_dl/extractor/peertube.py
```diff
@@ -423,26 +424,27 @@ class PeerTubeIE(InfoExtractor):
...
         'info_dict': {
-            'id': '2790feb0-8120-4e63-9af3-c943c69f5e6c',
+            'id': '9c9de5e8-0a1e-484a-b099-e80766180a6d',
             'ext': 'mp4',
-            'title': 'wow',
-            'description': 'wow such video, so gif',
+            'title': 'What is PeerTube?',
+            'description': '**[Want to help to translate this video?](https://weblate.framasoft.org/projects/what-is-peertube-video/)**\r\n\r\n**Take back the control of your videos! [#JoinPeertube](https://joinpeertube.org)**\r\n*A decentralized video hosting network, based on free/libre software!*\r\n\r\n**Animation Produced by:** [LILA](https://libreart.info) - [ZeMarmot Team](https://film.zemarmot.net)\r\n*Directed by* Aryeom\r\n*Assistant* Jehan\r\n**Licence**: [CC-By-SA 4.0](https://creativecommons.org/licenses/by-sa/4.0/)\r\n\r\n**Sponsored by** [Framasoft](https://framasoft.org)\r\n\r\n**Music**: [Red Step Forward](http://play.dogmazic.net/song.php?song_id=52491) - CC-By Ken Bushima\r\n\r\n**Movie Clip**: [Caminades 3: Llamigos](http://www.caminandes.com/) CC-By Blender Institute\r\n\r\n**Video sources**: https://gitlab.gnome.org/Jehan/what-is-peertube/',
```
dstftw: `md5:`.

youtube_dl/extractor/peertube.py
```diff
@@ -7,6 +7,7 @@
 from ..compat import compat_str
 from ..utils import (
     int_or_none,
+    str_or_none,
```
dstftw: Alphabetic.

youtube_dl/extractor/peertube.py
```diff
@@ -528,12 +562,17 @@ def account_data(field):
...
+            'description': description,
             'thumbnail': urljoin(url, video.get('thumbnailPath')),
             'timestamp': unified_timestamp(video.get('publishedAt')),
             'uploader': account_data('displayName'),
-            'uploader_id': account_data('uuid'),
-            'uploder_url': account_data('url'),
+            'uploader_id': '%s@%s' % (account_data('name'), account_data('host')),
+            'uploader_url': account_data('url'),
+            'channel': channel_data('displayName'),
+            'channel_id': '%s@%s' % (channel_data('name'), channel_data('host')),
```
dstftw: Similarly, that's not an id. Also breaks on `None`s.

3risian: Keeping the two fields consistent until the above discussion on id is resolved.
- The same nickname can possibly refer to different accounts on different hosts with the same username. Therefore nickname in isolation is not an id.
- The same account/channel will have different numeric ids when accessed from different instances. This _could_ be used, but to me it seems like a poor identifier. This PeerTube extractor could then extract videos from multiple accounts with the same id, or the same account with different ids.
- `username@host` is certainly an id string, and the most robust id of the three. I do not see how it is problematic.

Fixing breaks on None.

dstftw: I've already pointed out: this is not a global unique identifier and it's not supposed to be.

3risian:
Requesting another review, some changes in this pull are important in solving PeerTube issues.

youtube_dl/extractor/peertube.py
```diff
@@ -484,6 +489,25 @@ def _extract_urls(webpage, source_url):
...
+    def _get_subtitles(self, host, video_id):
+        video_captions = self._download_json(
+            'https://%s/api/v1/videos/%s/captions' % (host, video_id), video_id, fatal=False)
+        if not isinstance(video_captions, dict):
+            return None
+
+        subtitles = {}
+        for entry in video_captions.get('data'):
+            language_id = try_get(entry, lambda x: x['language']['id'], compat_str)
+            caption_path = str_or_none(entry.get('captionPath'))
```
dstftw: `caption_url = urljoin('https://%s' % host, entry.get('captionPath'))`

youtube_dl/extractor/peertube.py
```diff
@@ -484,6 +489,25 @@ def _extract_urls(webpage, source_url):
...
+            if language_id and caption_path:
+                caption_url = 'https://%s%s' % (host, caption_path)
+                caption_dict = {
+                    'url': caption_url
+                }
+                if subtitles.setdefault(language_id, [caption_dict]) != [caption_dict]:
+                    subtitles[language_id].append(caption_dict)
```
dstftw: What are you even doing?
```
subtitles.setdefault(language_id, []).append({
    'url': caption_url,
})
```

youtube_dl/extractor/peertube.py
```diff
@@ -513,10 +537,25 @@ def _real_extract(self, url):
...
+        video_description = self._download_json(
+            'https://%s/api/v1/videos/%s/description' % (host, video_id), video_id, fatal=False)
+
+        description = None
+        if isinstance(video_description, dict):
+            description = video_description.get('description')
+
+        subtitles = self.extract_subtitles(host, video_id)
+
+        def try_get_second_level_data(section, field, type_):
```
dstftw: `try_get_second_level_data` -> `data`

youtube_dl/extractor/peertube.py
```diff
@@ -528,14 +567,17 @@ def account_data(field):
...
+            'description': description,
             'thumbnail': urljoin(url, video.get('thumbnailPath')),
             'timestamp': unified_timestamp(video.get('publishedAt')),
-            'uploader': account_data('displayName'),
-            'uploader_id': account_data('uuid'),
-            'uploder_url': account_data('url'),
-            'license': try_get(
-                video, lambda x: x['licence']['label'], compat_str),
+            'uploader': account_data('displayName', compat_str),
+            'uploader_id': account_data('id', int),
```
dstftw: Must be string.

youtube_dl/extractor/peertube.py
```diff
@@ -528,14 +567,17 @@ def account_data(field):
...
+            'uploader_url': url_or_none(account_data('url', compat_str)),
+            'channel': channel_data('displayName', compat_str),
+            'channel_id': channel_data('id', int),
```
dstftw: Must be string.

youtube_dl/extractor/peertube.py
3risian: As in, rename the function? Alright.
Would this cause any readability issues for future contributors as it is a special case (only goes down two levels)?

3risian:
Thank you for the help.

tyrann0us:
https://www.youtube.com/watch?v=rR4n-0KYeKQ 
LOL 🤣

PR #23659, opened 2020-01-07
[Youporn] Regex fix for newer URLs

snizzl:

---

- [X] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---
The URLs of new videos have a different format. A slash in the regex prevented the information from being read.

### DIFF
--- a/youtube_dl/extractor/youporn.py
+++ b/youtube_dl/extractor/youporn.py
@@ -57,7 +57,27 @@ class YouPornIE(InfoExtractor):
         'params': {
             'skip_download': True,
         },
-    }]
+    }, {
+        # the source urls in the "formats" array of newer videos might have a new format
+        'url': 'https://www.youporn.com/watch/15573378/visit-x-unschuldiges-deutsches-luder-bekommt-hintern-versohlen/',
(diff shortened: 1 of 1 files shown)


dstftw:
Add a test.

youtube_dl/extractor/youporn.py
```diff
@@ -146,7 +167,7 @@ def _real_extract(self, url):
             webpage, 'uploader', fatal=False)
         upload_date = unified_strdate(self._html_search_regex(
             [r'Date\s+[Aa]dded:\s*<span>([^<]+)',
-             r'(?s)<div[^>]+class=["\']videoInfo(?:Date|Time)["\'][^>]*>(.+?)</div>'],
+             r'(?s)<div[^>]+class=["\']video-uploaded["\'][^>]*>[^<]+?<span>([^<]+)'],
```
dstftw: Do not remove previous pattern.

snizzl: Alright. As they changed their design, the old pattern fails so i removed it . 
I've now restored the old pattern.

https://regexr.com/4s83s (new pattern)
https://regexr.com/4s83m (old pattern)

PR #23715, opened 2020-01-12
[youtube] Fix the 429 error (partially)

aminprox:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This bugfix will help the 429 error.
Youtube is blocking requests to youtube.com/get_video_info and not other pages (even with captcha resolving), but as I checked we don't need "youtube.com/get_video_info" mostly and the `player_response` is enough.


dstftw:
1. `video_info` has nothing to do with `videoDetails`.
2.
>Youtube is blocking requests to youtube.com/get_video_info

You must provide clear evidence of that. From what I see 429 has nothing to do with `get_video_info`.
3. This may breaks some extraction scenarios.

aminprox:
I should mention that there are two types of 429 error, the one with the captcha which appears all over the youtube pages (and maybe other google services) and another one without any captchas. The first one would be fixed just with solving a captcha, but the second one would not go away and it's only on "/get_video_info" and of course, the usual player work just fine (with the solved captcha and no need for "/get_video_info").

I can't provide evidence as I just found it, and of course, I could provide logs if it helps.

There should be no conflict as I have tested different videos and passes the builtin tests,  but I'm pretty sure it would just be in use when the "/watch" is ok and "/get_video_info" sends 429 error.
I know video_info has nothing to do with videoDetails, but after this lines of code, there are variables like "title" and other things which are available in the player_response.
After all, it works just in the required situation and I hope it would help other developers to address this issue maybe.

nishannakarmi:
No player response is not enough. There are different varieties of youtube videos. such as,

- https://www.youtube.com/watch?v=mFHRYMdhM6w (which has player details in the page source)
- https://www.youtube.com/watch?v=DxMo75KYzMc (which does not)

I case if not player details in the source then we may need get_video_info or other mechanisms to get video details.

aminprox:
I have tried to add those two lines in case of failure by get_video_info (maybe 429 error) so it would just continue the download rather than stop the whole procedure.
But in normal situation it does not do anything at all and get_video_info would be called and the normal routine will be conrinued.

PR #24199, opened 2020-03-01
fix(extractor: odnoklassniki): for download URL use the same protocol as in original URL

actionless:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Some videos are not downloading well via http. While investigating the problem i've noticed what when i change the hardcoded protocol from http to https those videos downloading well.

### DIFF
--- a/youtube_dl/extractor/odnoklassniki.py
+++ b/youtube_dl/extractor/odnoklassniki.py
@@ -121,6 +121,10 @@ class OdnoklassnikiIE(InfoExtractor):
         # Paid video
         'url': 'https://ok.ru/video/954886983203',
         'only_matching': True,
+    }, {
+        # Downloads only via https
+        'url': 'https://ok.ru/video/1704000096969',
+        'only_matching': True,
     }]
 
     @staticmethod
--- a/youtube_dl/extractor/odnoklassniki.py
+++ b/youtube_dl/extractor/odnoklassniki.py
@@ -123,7 +123,7 @@ class OdnoklassnikiIE(InfoExtractor):
         'only_matching': True,
     }, {
         # Downloads only via https
-        'url': 'https://ok.ru/video/1704000096969',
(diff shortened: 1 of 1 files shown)


dstftw:
1. Provide concrete example URLs.
2. Add a test.

actionless:
@dstftw done

youtube_dl/extractor/odnoklassniki.py
```diff
@@ -121,6 +121,10 @@ class OdnoklassnikiIE(InfoExtractor):
         # Paid video
         'url': 'https://ok.ru/video/954886983203',
         'only_matching': True,
+    }, {
+        # Downloads only via https
+        'url': 'https://ok.ru/video/1704000096969',
```
dstftw: >Video has been blocked due to author's rights infingement

actionless: updated the link

dstftw: https://ok.ru/video/1705664645833 downloads fine without these changes.

actionless: for me without this change it returns:
```
[Odnoklassniki] 1705664645833: Downloading webpage
ERROR: Unable to download webpage: HTTP Error 403: Forbidden (caused by <HTTPError 403: 'Forbidden'>); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

while for some other URLs i've tried it works

actionless: after the latest update (2020.03.01) the issue is still reproducible for me

actionless:
for some reason the problem is reproducible only from inside centralwoonen's network (who go outside through ziggo), but not even reproducible from other networks served by ziggo provider

PR #24232, opened 2020-03-03
Add feature: Option to choose temp folder

RasmusMathiesen:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [x] New feature

---

Adds a new option for the user to specify a folder where temporary files can be stored.
Motivation: If you are downloading a very large video that needs to be merged from a very large video and audio file, but want to minimize the number of write operations that will be performed on your drive, this feature enables that. (By letting a ramdisk be used as a temp folder for example)

Initially only used to determine where to store separate audio and video files before merging.

This feature is superior to using something like --exec "mv {} final_dir", because your ramdisk has to be twice the size of the final (maybe huge) video file this way.


dstftw:
1. Have you performed any I/O measures in the first? If not perform and provide clear evidence this results in tangible improvement.
2. Temp dir is not taken into account in cases when merging is not involved: direct files are still downloaded in non temp folder by chunks as well as fragmented downloads that's even worse so that your main aim to "minimize the number of write operations" is not met at all.
3. Why do you think that merging does not involve lots of small write operations? Have you measured it? Provide clear evidence that merging is comparable to bare copying in terms of I/O. If it's not then you have to merge in temp as well.
4. If you want this to be accepted you have to clearly define the notion of *temporary files* in the first place.

RasmusMathiesen:
Thank you for the feedback. Looking further into it, SSD longevity (which I also had in mind) is apparently not a big problem nowadays. For someone archiving huge amounts of video it could still be a consideration though. https://www.ontrack.com/blog/2018/02/07/how-long-do-ssds-really-last/
The only ubiquitous reason to use this then would be when you have limited space on a drive and merging requires too much space.
1. Don’t know the best way to measure this but will look into it. 
2. I imagined the option can be used in more ways in the future but wanted to keep the initial PR simple.
3. Also don’t know the proper way to measure this, but will look into it.
4. A temp file is a file expected to be deleted again because it’s irrelevant for the user. I’m aware of the -k option and not sure what effect these two options should have when combined.

PR #24285, opened 2020-03-07
Split Thumbzilla out into it's own extractor.

twaddington:

---

- [ ] Bug fix
- [x] Improvement
- [x] New extractor
- [ ] New feature

---

Split Thumbzilla out into it's own extractor.

This simplifies the pornhub extractor and makes it easier to maintain in
the future.

### DIFF
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -1129,6 +1129,7 @@ from .thisamericanlife import ThisAmericanLifeIE
 from .thisav import ThisAVIE
 from .thisoldhouse import ThisOldHouseIE
 from .threeqsdn import ThreeQSDNIE
+from .thumbzilla import ThumbzillaIE
 from .tiktok import (
     TikTokIE,
     TikTokUserIE,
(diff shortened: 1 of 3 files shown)


dstftw:
No, as this is the same extractor thus must have the same id for download archives.

twaddington:
> No, as this is the same extractor thus must have the same id for download archives. &mdash; @dstftw 

Can you elaborate on this? Wouldn't it have the same id since it delegates the download to the pornhub extractor via a url_result?

dstftw:
There is no point in extractor that only redirects and introduces unnecessary code duplication.

twaddington:
@dstftw code duplication isn't a bad thing if it makes the code more maintainable. Thumbzilla used to host videos from other tube sites that aren't Pornhub so this could would allow us to delegate out to different extractors depending on the source site.

twaddington:
Plus, if Thumbzilla ever goes away entirely, it makes it easier to delete the relevant code since it's not intermingled with the pornhub logic.

dstftw:
No, code duplication does not make code more maintainable, never.
>host videos from other tube sites that aren't Pornhub

Provide concrete URLs and add implement support for them.

twaddington:
lol, okay. Thanks for the warm welcome @dstftw.

Hrxn:
> No, as this is the same extractor thus must have the same id for download archives.

Do they really have the same video IDs? Why would that even make any sense?

PR #24558, opened 2020-03-31
srgssr: download metadata and stream via HTTPS instead of HTTP

vszakats:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [x] At least skimmed through [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [ ] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

Downloading an SRF video was downloading two metadata resources via cleartext HTTP while these resources are also available via HTTPS. Upgraded the protocol for these hard-coded URLs.

PR #24562, opened 2020-04-01
[srgssr] rework to use API v2 and HTTPS

vszakats:

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

* use v2 API endpoint
* use HTTPS protocol for metadata downloads
* prefer HTTPS streams
* add HTTPS test
* remove a test with URL no longer availabe
  (id `677f5829-e473-4823-ac83-a1087fe97faa`)
* fix test issues (e.g. id `6348260`)
* v2 API no longer offers `publishedDate`, only a `date` which may be slightly different. Tests updated accordingly.

### DIFF
--- a/youtube_dl/extractor/srgssr.py
+++ b/youtube_dl/extractor/srgssr.py
@@ -29,7 +29,7 @@ class SRGSSRIE(InfoExtractor):
     def _get_tokenized_src(self, url, video_id, format_id):
         sp = compat_urllib_parse_urlparse(url).path.split('/')
         token = self._download_json(
-            'http://tp.srgssr.ch/akahd/token?acl=/%s/%s/*' % (sp[1], sp[2]),
+            'https://tp.srgssr.ch/akahd/token?acl=/%s/%s/*' % (sp[1], sp[2]),
             video_id, 'Downloading %s token' % format_id, fatal=False) or {}
         auth_params = token.get('token', {}).get('authparams')
         if auth_params:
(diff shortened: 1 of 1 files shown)


vszakats:
Ref: https://github.com/ytdl-org/youtube-dl/pull/24558

vszakats:
@dstftw What do you think about this PR? Any of concern or something to fix?

youtube_dl/extractor/srgssr.py
```diff
@@ -56,42 +67,50 @@ def _real_extract(self, url):
...
-        title = metadata['title']
-        description = metadata.get('description')
-        created_date = media_data.get('createdDate') or metadata.get('createdDate')
+        chapter = media_data.get('_chapter', {})
+        if media_data.get('_segment'):
+            metadata = media_data['_segment']
+        else:
+            metadata = chapter
+
+        title = metadata.get('title')
```
dstftw, line 76: Read coding conventions on mandatory metadata.

youtube_dl/extractor/srgssr.py
```diff
@@ -56,42 +67,50 @@ def _real_extract(self, url):
...
+        description = metadata.get('description', '')
```
dstftw, line 77: Lack of data is denoted by None not empty string.

vszakats:
So I'm guessing this API doesn't fit your strict criteria. Let's leave it insecure and using the obsolete API.

PR #19173, opened 2019-02-09
[soundcloud] Use api-v2 to support AAC streams

ArrEssJay:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [x] At least skimmed through [adding new extractor tutorial](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/rg3/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [x] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

Soundcloud Go+ provides AAC streams at 256kbps. These are only available from the api-v2 endpoint as progressive https or hls. This change addresses adds support for these and retains support for existing https/hls mp3 from the same v2 endpoint, and downloads (when available) from the v1 endpoint.

The 256k streams are only available if a Go+ account is provided by way of an OAuth token:
```
--add-header "Authorization: OAuth <token>"
```
If an OAuth token is not supplied, only MP3 128kbps and Opus 64kbps streams are available. Incidentally, this change also adds support for downloading the low rate Opus streams.

The main downside to this change is a delay whilst all the of the stream URLs are retrieved and validated.

I have tested this with a Go+ account and 256kbps AAC streams are retrieved correctly. I did have to use ffmpeg on the HLS streams (```--hls-prefer-ffmpeg```) for a valid seek table to be present but I have not investigated the precise cause of this.

PR #24708, opened 2020-04-09
[soundcloud] added support for aac streams

tomsolarino:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [x] New feature

---

Soundcloud Go+ provides AAC streams at 256kbps. The 256k streams are only available if a Go+ account is provided by way of an OAuth token:

`--add-header "Authorization: OAuth <token>"`

If an OAuth token is not supplied, only MP3 128kbps and Opus 64kbps streams are available. If a Go+ account isn't used, the links to the AAC streams will not even be sent by the API. I have tested this with a Go+ account and 256kbps AAC streams are retrieved correctly. 

I did have to use ffmpeg on the HLS streams (`--hls-prefer-ffmpeg`) for a valid seek table to be present but I have not investigated the precise cause of this.

### DIFF
--- a/youtube_dl/extractor/soundcloud.py
+++ b/youtube_dl/extractor/soundcloud.py
@@ -350,6 +350,8 @@ class SoundcloudIE(InfoExtractor):
             format_id_list = []
             if protocol:
                 format_id_list.append(protocol)
+            if f.get('ext') == 'aac':
+                f['abr'] = '256'
             for k in ('ext', 'abr'):
                 v = f.get(k)
                 if v:


tomsolarino:
Credit to [@ArrEssJay](https://github.com/ArrEssJay) on their original [PR](https://github.com/ytdl-org/youtube-dl/pull/19173)

youtube_dl/extractor/soundcloud.py
```diff
@@ -350,6 +350,8 @@ def add_format(f, protocol, is_preview=False):
             format_id_list = []
             if protocol:
                 format_id_list.append(protocol)
+            if f.get('ext') == 'aac':
+                f['abr'] = '256'
```
dstftw, line 354: Must be numeric not string.

tomsolarino: @dstftw String is converted to int on line 364. At line 354 all other extensions' bit rates are as strings, that's how they're passed into `def add_format()`. Also by **account-needed** do you mean that Auth Token is not acceptable and that username and password also need supporting?

dstftw: Meaning that you must provide account credentials for testing or whatever else is needed.

eugeneware: I can confirm that this code works by providing the OAuth token.

eg:

``` sh
 youtube-dl -f bestaudio --add-header "Authorization:OAuth <OAUTH TOKEN>" https://soundcloud.com/turborecordings/azari-iii-hungry-for-the
-5
youtube-dl -f http_aac_256 --add-header "Authorization:OAuth <OAUTH TOKEN>" https://soundcloud.com/turborecordings/azari-iii-hungry-for-
the-5
```

You can get this code by opening up dev tools in the browser.

PR #24847, opened 2020-04-17
[README.md] Clarify docs re: --exec command (closes #19087)

### LINKED ISSUE #19087: Documentation: --exec does not execute POST download, but POST processing
Thanks for all the work :)
One thing, definitely being picky, but the --exec is documented as executing post download, and that happens if you specify "," (comma) between formats, because there's no merging.
But if there's a "+" (plus) then the exec happens NOT on the fragments (as I would expect) but on the final file, so it doesn't happen post download, but post processing.

willbeaufoy:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Clarify docs re: --exec command. As bug says, docs say --exec runs after file download, when actually it runs after download and processing (e.g. merging formats).

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,9 +434,9 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
                                      either the path to the binary or its
                                      containing directory.
     --exec CMD                       Execute a command on the file after
-                                     downloading, similar to find's -exec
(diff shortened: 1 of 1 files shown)


dstftw:
README is autogenerated.

PR #24848, opened 2020-04-17
[youtube] fix chapters extractor (fix #24819)

### LINKED ISSUE #24819: [youtube] video not getting video "chapters" properly
The 'chapters' value from the JSON (-J parameter), is usually present on some video. But since a few days, it is not available in ANY video. My guess is that's since the new chapter UI of youtube.

Here's a verbose excerpt from youtube version 23.04.2020, not that it looks very useful

```
[ ~ ] $ youtube-dl -vJ https://www.youtube.com/watch?v=gBRKnvK1JUE 
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'-vJ', u'https://www.youtube.com/watch?v=gBRKnvK1JUE']
[debug] youtube-dl version 2020.03.24
[debug] Python version 2.7.16 (CPython) - Linux-4.19.0-8-amd64-x86_64-with-debian-10.3
[debug] exe versions: ffmpeg 4.1.5, ffprobe 4.1.5
[debug] Proxy map: {}
[debug] Default format spec: bestvideo+bestaudio/best
{"upload_date": "20190921", "extractor": "youtube", "chapters": null, ...
```

jaimebl:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

That's a fix for issue https://github.com/ytdl-org/youtube-dl/issues/24819

It modifies the regular expression to capture chapters from description when they follow this pattern:
`pattern: 00:00 - 09:24 <title>` 

You can see an example of that here:
https://www.youtube.com/watch?v=gBRKnvK1JUE

### DIFF
--- a/test/test_youtube_chapters.py
+++ b/test/test_youtube_chapters.py
@@ -16,6 +16,41 @@ class TestYoutubeChapters(unittest.TestCase):
 
     _TEST_CASES = [
         (
+            # https://www.youtube.com/watch?v=gBRKnvK1JUE
+            # pattern: 00:00 - 09:24 <title>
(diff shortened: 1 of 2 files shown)


dstftw:
For this pattern you must extract start and end time from each entry.

anabis:
Hello, there is another way to get the chapters names and timestamp without hazardously parsing the description, it is now in an inline JSON in the page, I don't know Python but I you can get the info like so :
`curl -s "https://www.youtube.com/watch?v=AK9r27jWVrc" | grep "ytInitialData" | grep -o "{.*}" | jq '.playerOverlays.playerOverlayRenderer.decoratedPlayerBarRenderer.decoratedPlayerBarRenderer.playerBar.chapteredPlayerBarRenderer.chapters'`

someziggyman:
Looks like this solution fixes cases for timecodes formatted like this: "00:12 some title of the chapter"
However, it fails for timecode formats like this one video: https://www.youtube.com/watch?v=Le0-eaxQ1_k

And these are mystery, but for some reason timecodes are null as well:
https://www.youtube.com/watch?v=YhO_E8xm0aY
https://www.youtube.com/watch?v=7wriyeBB5hY

I guess it's some regex problem and it may be rather complicated to cover all the formatting cases, however decided t post this comment here. Thank you for making youtube-dl better!

PR #25063, opened 2020-04-29
Detect wrong video IDs for youtube and correct them

AStove:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Youtube video ids are the result of a base 64 encoding. 
For example DJztXj2GPfk is the result of encoding 0c 9c ed 5e 3d 86 3d f9
But due to padding video id DJztXj2GPfl will decode into this same binary value.
Due to this, youtube-dl will download both DJztXj2GPfk and DJztXj2GPfl, it will result in the same file.
So why would we need this change? This does not work for all youtube APIs, for example it doesn't work for age gated content. Does work for 6QFwo57WKwg but does not work for 6QFwo57WKwh. The 6QFwo57WKwh id gives an age gated error, while this is irrelevant, it's the id that is wrong.


dstftw:
You should not make any assumptions about id. Youtube redirects to the page with correct id that must be taken instead of this.

AStove:
The assumption is correct and it saves a bunch of time and requests.

oxguy3:
The assumption may be correct now, but there is no guarantee that it will remain correct in the future. There is little harm in downloading a duplicate file, especially given that this seems to be an extremely unusual edge case. However, there would be harm in failing to download two separate videos because we made a bad assumption about YouTube's IDs.

Do these wrongly-padded video IDs actually appear in the wild? Is this issue affecting any users?

AStove:
You are right.
> Do these wrongly-padded video IDs actually appear in the wild?

No.
>Is this issue affecting any users?

No.

Issue #12471, opened 2017-03-17
--write-info-json file is file peremissin 600 -rw--------


niammuddin:
Hi dear, I add --write-info-json, but file permission can be download, because file permisson is 600, I need to 644, so for all file .json to 644 permisson, please understand me, thanks

dstftw:
Modifying permissions is your responsibility not youtube-dl's. Out of scope.

aleksejrs:
Why is that different for JSON (600) and thumbnail/video (640)? It seems to only modify the video file.

```
test$ youtube-dl -v --ignore-config --write-all-thumbnails --write-info-json --write-sub --no-call-home https://www.youtube.com/watch?v=9y2N-p93kF4
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'-v', u'--ignore-config', u'--write-all-thumbnails', u'--write-info-json', u'--write-sub', u'--no-call-home', u'https://www.youtube.com/watch?v=9y2N-p93kF4']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2019.02.08
[debug] Python version 2.7.15+ (CPython) - Linux-4.19.0-1-amd64-x86_64-with-debian-buster-sid
[debug] exe versions: ffmpeg 4.1-1, ffprobe 4.1-1, rtmpdump 2.4
[debug] Proxy map: {}
[youtube] 9y2N-p93kF4: Downloading webpage
[youtube] 9y2N-p93kF4: Downloading video info webpage
WARNING: video doesn't have subtitles
[debug] Default format spec: bestvideo+bestaudio/best
[info] Writing video description metadata as JSON to: Рысь поймала зайца. Удачная съемка на фотоловушку (со звуком).-9y2N-p93kF4.info.json
[youtube] 9y2N-p93kF4: Downloading thumbnail ...
[youtube] 9y2N-p93kF4: Writing thumbnail to: Рысь поймала зайца. Удачная съемка на фотоловушку (со звуком).-9y2N-p93kF4.jpg
WARNING: Requested formats are incompatible for merge and will be merged into mkv.
[debug] Invoking downloader on u'https:// (…)'
[download] Destination: Рысь поймала зайца. Удачная съемка на фотоловушку (со звуком).-9y2N-p93kF4.f247.webm
[download] 100% of 9.52MiB in 00:01
[debug] Invoking downloader on u'(…)'
[...]

PR #25122, opened 2020-05-04
Fix issue 12471

ankenyr:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This is to fix https://github.com/ytdl-org/youtube-dl/issues/12471
Currently with the use of [tempfile](https://github.com/ytdl-org/youtube-dl/blob/master/youtube_dl/utils.py#L1826) on Linux it will create a file with only the user read write bit set for user and all other bits off. When renaming the file it keeps this restrictive permission setting. From my testing this seems to not have an affect in Windows.


ankenyr:
Seems I had some leftovers from years ago I forgot about. Removing those

youtube_dl/utils.py
```diff
@@ -1835,6 +1836,12 @@ def write_json_file(obj, fn):
                 os.unlink(fn)
             except OSError:
                 pass
+        os.chmod(tf.name,
+                 stat.S_IRUSR |
+                 stat.S_IWUSR |
+                 stat.S_IRGRP |
+                 stat.S_IROTH)
+
```
dstftw, line 1844: 1. No umask respected.
2. There is no `os.chmod` in python 3.2 according to python [docs](https://docs.python.org/3/library/os.html#os.chmod).
3. flake8.

ankenyr: umask should be good now.
flake8 no longer complaines (sorry about that one!)
Not sure what you mean about os.chmod not being in python3.2?
The link takes me to 3.8 and the python 3.2 docs [clearly show its there](https://docs.python.org/3.2/library/os.html)
If I remember when I last checked tf.name returned a string object so that should work.

ankenyr:
I believe this is good to be looked at again. @dstftw 
Please let me know if I am missing something.

ankenyr:
@dstftw I am hoping this can be reviewed and included in the next release.

youtube_dl/utils.py
```diff
@@ -1835,6 +1836,17 @@ def write_json_file(obj, fn):
...
+        mask = os.umask(0)
+        os.umask(mask)
+        os.chmod(tf.name,
+                 (stat.S_IRUSR
+                  | stat.S_IWUSR
+                  | stat.S_IRGRP
+                  | stat.S_IWGRP
+                  | stat.S_IROTH
+                  | stat.S_IWOTH)
+                 & ~mask)
```
dstftw, line 1848: This must be in a separate try-except so that it does not break renaming to correct name if any of these statements fails.

youtube_dl/utils.py
```diff
@@ -1835,6 +1836,17 @@ def write_json_file(obj, fn):
...
                 pass
+        mask = os.umask(0)
+        os.umask(mask)
+        os.chmod(tf.name,
+                 (stat.S_IRUSR
+                  | stat.S_IWUSR
+                  | stat.S_IRGRP
+                  | stat.S_IWGRP
+                  | stat.S_IROTH
+                  | stat.S_IWOTH)
```
dstftw, line 1847: Just `0o666`.

youtube_dl/utils.py
```diff
@@ -1835,6 +1836,17 @@ def write_json_file(obj, fn):
...
+                 & ~mask)
+
```
dstftw, line 1849: Remove unnecessary whitespace.

ankenyr: done

youtube_dl/utils.py
```diff
@@ -1835,6 +1835,12 @@ def write_json_file(obj, fn):
                 os.unlink(fn)
             except OSError:
                 pass
+        try:
+            mask = os.umask(0)
+            os.umask(mask)
+            os.chmod(tf.name, 0o666 & ~mask)
+        except OSError:
```
dstftw, line 1842: Are you sure no other exceptions possible here?

ankenyr: I believe that this module will only raise OSError per the docs
https://docs.python.org/3/library/os.html

> Note All functions in this module raise OSError (or subclasses thereof) in the case of invalid or inaccessible file names and paths, or other arguments that have the correct type, but are not accepted by the operating system.

PR #25184, opened 2020-05-07
[svt] Add timestamp and thumbnail in more cases

wader:
Add timestamp, set to "valid from" which i think could been seen as publish time.
Add thumbnail in more cases, seems to was only done in the embedded data case for some reason.
Update API URL (current one redirects to it).
Switch svtplay test url to an existing video and also one with no expire date.
Also add an additional thumbnail url test regex.

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

See commit messages

### DIFF
--- a/youtube_dl/extractor/svt.py
+++ b/youtube_dl/extractor/svt.py
@@ -9,6 +9,7 @@ from ..utils import (
     determine_ext,
     dict_get,
     int_or_none,
+    parse_iso8601,
     str_or_none,
     strip_or_none,
     try_get,
@@ -44,7 +45,8 @@ class SVTBaseIE(InfoExtractor):
                     'format_id': player_type,
                     'url': vurl,
                 })
-        if not formats and video_info.get('rights', {}).get('geoBlockedSweden'):
+        rights = video_info.get('rights', {})
+        if not formats and rights.get('geoBlockedSweden'):
             self.raise_geo_restricted(
                 'This video is only available in Sweden',
                 countries=self._GEO_COUNTRIES)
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/svt.py
```diff
@@ -141,21 +146,28 @@ class SVTPlayIE(SVTPlayBaseIE):
...
                 'sv': [{
+                    'ext': 'vtt',
+                }, {
                     'ext': 'wsrt',
                 }]
             },
         },
+        'params': {
+            'format': 'bestvideo',
+        },
```
wader, line 170: Without this i get `DownloadError: ERROR: requested format not available` when running test. Not sure why. I also noticed that with another test video https://www.svtplay.se/video/26595950/we-got-this/we-got-this-sasong-1-katten-rio i got an download error about `AssertionError: Expected test_SVTPlay_jNwpvBJ.mp4 to be at least 9.77KiB, but it's only 865.00B` which seems to be because the first init segment in the dash stream is 800 byte, tried to change `file_minsize` but the lowest seems to be 1000 because of https://github.com/ytdl-org/youtube-dl/blob/master/test/test_download.py#L219

wader:
Looks ok?

wader:
Ping?

wader:
Rebased on master and updated one test but i still have the issue i mentioned above about file_minsize

@dstftw could you take a look? no stress

wader:
@dstftw i've rebased on master. But i still have the issue i mentioned above about too small file size for the init segment

wader:
Hi again, something left to fix or changes not suitable to merge?

Issue #19241, opened 2019-02-16
[Feature Request] Add more operations to format selection


sgkoishi:
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2019.02.08**
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
- [x] Checked that provided video/audio/playlist URLs (if any) are alive and playable in a browser

### What is the purpose of your *issue*?
- [ ] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [x] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### Enhancement of Format Selection
[Format selection](https://github.com/rg3/youtube-dl#format-selection) now allow the best/worst format to download. However, the format selection only support conditions (e.g. `-f "best[height>=720]"`), as a filter. In some special case, people may want to download video of most/least certain field (e.g. highest resolution, smallest size) rather than default "best/worst". Youtube-dl does not support this without manually specify numeric format code or knowing the exact value.

For example, the `best` quality for [this youtube video](https://www.youtube.com/watch?v=d-nxW9qBtxQ) is a medium(640x360) resolution while highest resolution is 1080p(1920x1080).
```
format code  extension  resolution note
249          webm       audio only DASH audio   62k , opus @ 50k, 1.43MiB
250          webm       audio only DASH audio   82k , opus @ 70k, 1.91MiB
140          m4a        audio only DASH audio  128k , m4a_dash container, mp4a.40.2@128k, 3.32MiB
171          webm       audio only DASH audio  144k , vorbis@128k, 3.55MiB
251          webm       audio only DASH audio  159k , opus @160k, 3.77MiB
160          mp4        256x144    144p   51k , avc1.4d400c, 30fps, video only, 501.84KiB
278          webm       256x144    144p   95k , webm container, vp9, 30fps, video only, 1.47MiB
133          mp4        426x240    240p  147k , avc1.4d4015, 30fps, video only, 1.08MiB
242          webm       426x240    240p  169k , vp9, 30fps, video only, 1.59MiB
[...]

PR #25388, opened 2020-05-24
Iho/smallest

inferiorhumanorgans:

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [X] New feature

---

Per #19241 add `smallestvideo` and `smallestaudio` filters that select formats based on the smallest file size.


dstftw:
This must be solved by introducing a mechanism to manage sorting priorities not by introducing new selector names.

Issue #25594, opened 2020-06-09
KeyError('@id',) youtube-dl [nrk]


elvisandrei16:
C:\Users\Admin\Desktop\yt-dl>youtube-dl --write-sub --sub-format ttml https://tv.nrk.no/serie/fra-boelle-til-bestevenn/sesong/1/episode/1/avspiller [NRKTVEpisode] fra-boelle-til-bestevenn/sesong/1/episode/1: Downloading webpage ERROR: An extractor error has occurred. (caused by KeyError('@id',)); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Windows 7 64 bit. Youtube-dl is up-to-date 2020.06.06. I can download some videos from the same website.
Please help me!

barsnick:
Your subject line should mention that the extractor is "[nrk]".

I have encountered this problem since yesterday on some of NRK's video's, most notably [NRKTVEpisodeIE], including its provided tests.

It seems to be that "@id" is no longer provided in the embedded JSON.

I was able to extract the first provided _TEST video ("Hellum's Kro") as well as another such video by modifying the video_id extraction:
```
--- a/youtube_dl/extractor/nrk.py
+++ b/youtube_dl/extractor/nrk.py
@@ -425,9 +425,9 @@ class NRKTVEpisodeIE(InfoExtractor):

         webpage = self._download_webpage(url, display_id)

-        nrk_id = self._parse_json(
-            self._search_regex(JSON_LD_RE, webpage, 'JSON-LD', group='json_ld'),
-            display_id)['@id']
+        nrk_id = self._search_regex(
+            r'<meta[^>]+property=(["\'])nrk:program-id(["\'])[^>]+content=(["\']?)(?P<id>[^"\']+)(["\'])',
+            webpage, 'display_id', group='id')

         assert re.match(NRKTVIE._EPISODE_RE, nrk_id)
         return self.url_result(
```

Disclaimer: I cannot tell whether this breaks other videos, or whether both methods need to be provided, with one as fallback. It's just a proof of concept and "works for me"(TM) right now.

elvisandrei16:
Thank you for the answer! 

[...]

Issue #24221, opened 2020-03-02
[nrk:series] No video formats found / 404 not found


YumekuiNeru:
## Checklist

- [x] I'm reporting a broken site support
- [x] I've verified that I'm running youtube-dl version **2020.03.01**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
$ youtube-dl 'https://tv.nrk.no/serie/harald-eia-presenterer-saann-er-norge/sesong/1/episode/1/avspiller' --verbose
[debug] System config: []
[debug] User config: ['--buffer-size', '16K']
[debug] Custom config: []
[debug] Command-line args: ['https://tv.nrk.no/serie/harald-eia-presenterer-saann-er-norge/sesong/1/episode/1/avspiller', '--verbose']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2020.03.01
[debug] Python version 3.6.10 (CPython) - Linux-5.5.5-gentoo-pix-x86_64-Intel-R-_Core-TM-_i7-6700K_CPU_@_4.00GHz-with-gentoo-2.7
[debug] exe versions: ffmpeg 4.2.2, ffprobe 4.2.2, rtmpdump 2.4
[debug] Proxy map: {}
[NRKTVEpisode] harald-eia-presenterer-saann-er-norge/sesong/1/episode/1: Downloading webpage
[debug] Using fake IP 84.212.51.255 (NO) as X-Forwarded-For.
[NRK] KMTE31000118: Downloading mediaelement JSON
[NRK] KMTE31000118: Downloading f4m manifest
WARNING: Unable to download f4m manifest: HTTP Error 404: Not Found
[NRK] KMTE31000118: Downloading m3u8 information
WARNING: Failed to download m3u8 information: HTTP Error 404: Not Found
[NRK] KMTE31000118: Downloading f4m manifest
WARNING: Unable to download f4m manifest: HTTP Error 404: Not Found
[NRK] KMTE31000118: Downloading m3u8 information
WARNING: Failed to download m3u8 information: HTTP Error 404: Not Found
[...]

PR #25751, opened 2020-06-21
Better extraction for nrk

Roxedus:

---

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This fixes two open issues for the NRK extractor, #25594 and #24221 as well as providing `alt_title`, useful for shows that does indeed have unique episode names, rather than `<series name> <episode number>:<total episodes in season>`

In order to fully fix 24221 I had to use a additional endpoint to fetch a m3u8 playlist, this seems to be the endpoint which exposes 1080P quality.

Some test maintenance was also done, the description seems to change from time to time.

### DIFF
--- a/youtube_dl/extractor/nrk.py
+++ b/youtube_dl/extractor/nrk.py
@@ -40,6 +40,10 @@ class NRKBaseIE(InfoExtractor):
             self._api_host = api_host
             break
 
+        playback_manifest = self._download_json(
+            'http://%s/playback/manifest/program/%s' % (self._api_host, video_id),
+            video_id, 'Downloading manifest JSON', fatal=False)
+
         title = data.get('fullTitle') or data.get('mainTitle') or data['title']
         alt_title = data.get('mainTitle')
         video_id = data.get('id') or video_id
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/nrk.py
```diff
@@ -52,6 +57,28 @@ def _real_extract(self, url):
         def make_title(t):
             return self._live_title(t) if live else t
 
+        playback_convia = playback_manifest.get('statistics').get('conviva')
```
dstftw, line 60: Breaks if request fails. Read coding conventions on optional metadata.

youtube_dl/extractor/nrk.py
```diff
@@ -52,6 +57,28 @@ def _real_extract(self, url):
...
+        if playback_convia:
+            streamurl = playback_convia.get('streamUrl', None)
+            stream = self._extract_m3u8_formats(streamurl, video_id, 'mp4',
+                'm3u8_native', m3u8_id='hls', fatal=False)
+            custom = playback_convia.get('custom')
+
+            dur = parse_duration(playback_convia.get('duration'))
+            subs = {}
+            for sub_title in playback_manifest.get('playable').get('subtitles'):
```
dstftw, line 69: Ditto.

youtube_dl/extractor/nrk.py
```diff
@@ -139,6 +166,13 @@ def video_id_and_title(idx):
                 EPISODENUM_RE, _season_episode, 'episode number',
                 default=None, group='episode'))
 
+            if not season_number or episode_number:
+                programs = self._download_json(
+                    'http://%s/programs/%s' % (self._api_host, video_id),
+                    video_id, 'Downloading programs manifest JSON', fatal=False)
+                season_number = int_or_none(programs.get('seasonNumber'))
+                episode_number = int_or_none(programs.get('episodeNumber'))
```
dstftw, line 174: Ditto.

youtube_dl/extractor/nrk.py
```diff
@@ -427,7 +464,8 @@ def _real_extract(self, url):
 
         nrk_id = self._parse_json(
             self._search_regex(JSON_LD_RE, webpage, 'JSON-LD', group='json_ld'),
-            display_id)['@id']
+            display_id).get('@id') or \
```
dstftw, line 467: Breaks.

Roxedus:
> making them appear as playlists
They always were marked as playlist, the "duplication" is probably because it pulls from two sources

youtube_dl/extractor/nrk.py
Roxedus: Because of the lack of fallback?

PR #25778, opened 2020-06-24
Don't save fragment files unless --keep-fragments option is given

lightmare:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

When downloading HLS video, each fragment is saved in a separate file. Once the fragment is complete, this file is read and its contents appended to the final destination (or .part file). Without `--keep-fragments` option, the fragment file is then immediately deleted. This PR avoids creating the fragment file if `--keep-fragments` option was not specified, by holding fragment data only in memory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,8 +192,8 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
     --abort-on-unavailable-fragment  Abort downloading when some fragment is not
                                      available
     --keep-fragments                 Keep downloaded fragments on disk after
-                                     downloading is finished; fragments are
-                                     erased by default
+                                     downloading is finished; fragments are not
+                                     saved by default
     --buffer-size SIZE               Size of download buffer (e.g. 1024 or 16K)
                                      (default is 1024)
     --no-resize-buffer               Do not automatically adjust the buffer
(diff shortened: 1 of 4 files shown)


dstftw:
This is intentional. Keeping fragments in memory will make fragments resuming impossible.

lightmare:
> This is intentional. Keeping fragments in memory will make fragments resuming impossible.

Yes, but is the ability to resume a fragment worth writing everything to disk twice? If a video consists of hundreds of fragments, losing an incomplete fragment and re-downloading it from the start adds very little overhead -- and that's only after the download is interrupted, which I assume shouldn't happen to every fragment.

dstftw:
I don't see any problem in writing to disk twice. Disk speed is usually several orders of magnitude faster than network I/O. Also fragments may be of arbitrary size meaning that potentially substantial amount of memory will be constantly wasted for storing fragment in memory as well as the same amount of traffic wasted on each resume.

lightmare:
Alright, thanks for taking a look. I'll keep this in my fork, as I'm mostly dealing with fragments only a few MB large, so for me that extra memory and occasional re-download is an acceptable trade-off for reduced disk I/O (while running ytdl in background).

PR #25717, opened 2020-06-18
[postprocessor/embedthumbnail] Fix Youtube's WebP thumbnails by converting them to JPG

alexmerkel:
### Before submitting a *pull request* make sure you have:
- [x] At least skimmed through [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

This change should fix #25687:
Youtube started using the [WebP](https://en.wikipedia.org/wiki/WebP) image format for thumbnail images. However, both `ffmpeg` and `AtomicParsley` only support the embedding of _JPEG_ and _PNG_ files. Therefore, as @noname8753 suggested, a conversion step was added during which the thumbnail is converted to _JPEG_ using `ffmpeg`.

Issue #6724, opened 2015-09-01
youtube-dl does not respect --audio-format if -o is used.


rsd:
Somewhere in the last few weeks, something changed while using the -o option.

When extracting audio, if -o is used, the audio stream is copied (mp4) even when mp3 format is what was asked for.
This does not happen if -o is ommited.

$ youtube-dl --verbose --extract-audio --audio-format mp3 -o output.mp3 https://www.youtube.com/watch?v=NUbUen66Z0k

[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'--verbose', u'--extract-audio', u'--audio-format', u'mp3', u'-o', u'output.mp3', u'https://www.youtube.com/watch?v=NUbUen66Z0k']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.08.28
[debug] Python version 2.7.9 - Linux-3.19.0-25-lowlatency-x86_64-with-Ubuntu-15.04-vivid
[debug] exe versions: avconv 11.2-6, avprobe 11.2-6, ffmpeg 2.7.2-1, ffprobe 2.7.2-1, rtmpdump 2.4
[debug] Proxy map: {}
[youtube] NUbUen66Z0k: Downloading webpage
[youtube] NUbUen66Z0k: Downloading video info webpage
[youtube] NUbUen66Z0k: Extracting video information
[youtube] NUbUen66Z0k: Downloading DASH manifest
[youtube] NUbUen66Z0k: Downloading DASH manifest
[...]

PR #26110, opened 2020-07-25
Fix error when using --audio-format and specifying an extension for the output file.

CrypticSignal:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

When using `--audio-format` and specifying an extension for the output filename (e.g. `--audio-format mp3 -o test.mp3`), the user gets the following error:
`ERROR: Stream #1:0 -> #0:1 (copy)`

This has been mentioned in [#6724](https://github.com/ytdl-org/youtube-dl/issues/6724), and, more recently, [here](https://github.com/ytdl-org/youtube-dl/pull/25717#issuecomment-658729806).

Here's an example of the terminal output with my fix:
```
C:\Users\H\Desktop\youtube-dl>python -m youtube_dl -x --audio-format mp3 -o test.mp3 https://www.youtube.com/watch?v=gQjAEbWZEgU
[youtube] gQjAEbWZEgU: Downloading webpage
The Matroska container (.mka) will be used for the download, but the final file will have the .mp3 extension
[download] Destination: test.mka
[download] 100% of 2.51MiB in 00:01
[ffmpeg] Destination: test.mp3
Deleting original file test.mka (pass -k to keep)

C:\Users\H\Desktop\youtube-dl>
```


alexmerkel:
Thanks for creating the pull request. Unfortunately, your change only fixes the problem for .mp3 files. But the same error occurs when using an other extension, e.g:
```
$ python3 -m youtube_dl -x --embed-thumbnail --audio-format m4a --audio-quality 0 -o test.m4a https://www.youtube.com/watch?v=dQw4w9WgXcQ
[youtube] dQw4w9WgXcQ: Downloading webpage
[youtube] dQw4w9WgXcQ: Downloading thumbnail ...
[youtube] dQw4w9WgXcQ: Writing thumbnail to: test.webp
[download] Destination: test.m4a
[download] 100% of 3.28MiB in 00:00
[ffmpeg] Post-process file test.m4a exists, skipping
[ffmpeg] Converting thumbnail "test.webp" to JPEG
[atomicparsley] Adding thumbnail to "test.m4a"
ERROR: AtomicParsley error: bad mpeg4 file (ftyp atom missing or alignment error).
```
While the error message is different, the reason is the same: The audio file is not converted (in this case from webm to mp4). So I think a more general solution is required.

BassThatHertz:
@alexmerkel Ah, I assumed the issue was specific to MP3. In that case, using the Matroska container for all audio downloads should do the trick, but I will test with all of the supported audio formats before pushing a new commit.

BassThatHertz:
Done. Thanks for the heads up @alexmerkel

alexmerkel:
This seems to fix the error for me!
Unfortunately it breaks the test function for the `prepare_filename` method. 
Additionally as a small extra, I would suggest adding a `[download]` tag to the beginning of the printed comment just to make it more consistent.

BassThatHertz:
@alexmerkel 
>Unfortunately it breaks the test function for the `prepare_filename` method.

Not anymore :)
>Additionally as a small extra, I would suggest adding a [download] tag to the beginning of the printed comment just to make it more consistent.

Done.

dirkf:
This PR is based on a misunderstanding of the meaning of the `--output ...` option, I think. It's a template, not a "filename". If for some reason you want your MP3 file to have a `nq3` extension, use a shell rename command in `--exec ...`.

Issue #26002, opened 2020-07-14
PeerTube extractor: No video formats found


0l-l0:
Hi,

I tried to download videos from several PeerTube instances with youtube-dl and it worked fine except for one instance (https://spacepub.space). I contacted the instance owner but he claimed that this is a youtube-dl related issue.

Here is the command that I issued:
```
$ youtube-dl -j --print-traffic --verbose https://spacepub.space/videos/watch/1619c000-7c44-4330-9177-29a0854bd759
```

...and I got the following output:
```
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-j', '--print-traffic', '--verbose', 'https://spacepub.space/videos/watch/1619c000-7c44-4330-9177-29a0854bd759']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2020.06.16.1
[debug] Python version 3.7.7 (CPython) - Linux-5.6.18-156.current-x86_64-with-glibc2.2.5
[debug] exe versions: ffmpeg 4.3, ffprobe 4.3
[debug] Proxy map: {}
send: b'GET /api/v1/videos/1619c000-7c44-4330-9177-29a0854bd759/ HTTP/1.1\r\nHost: spacepub.space\r\nUser-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3767.0 Safari/537.36\r\nAccept-Charset: ISO-8859-1,utf-8;q=0.7,*;q=0.7\r\nAccept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8\r\nAccept-Encoding: gzip, deflate\r\nAccept-Language: en-us,en;q=0.5\r\nConnection: close\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server: nginx
header: Date: Tue, 14 Jul 2020 12:56:46 GMT
header: Content-Type: application/json; charset=utf-8
header: Content-Length: 6513
header: Connection: close
header: X-Powered-By: Express
header: Tk: N
header: Access-Control-Allow-Origin: *
header: Access-Control-Allow-Credentials: true
header: Access-Control-Expose-Headers: Retry-After
header: X-RateLimit-Limit: 50
header: X-RateLimit-Remaining: 49
[...]

PR #26302, opened 2020-08-11
[peertube] Extract files also from `streamingPlaylists`

0l-l0:
As mentioned in #26002 some PeerTube instances respond with a seemingly corrupted video descriptor JSON object and the extractor is unable to determine the available file formats in those cases. Turned out that an empty `"files"` array caused those kind of problems but, fortunately, there are another `"files"` arrays in `"streamingPlaylists"` array which can be used instead.

Here is the corresponding part of the API specification:
https://docs.joinpeertube.org/api-rest-reference.html#tag/Video/paths/~1videos~1%7Bid%7D/get

### DIFF
--- a/youtube_dl/extractor/peertube.py
+++ b/youtube_dl/extractor/peertube.py
@@ -451,6 +451,18 @@ class PeerTubeIE(InfoExtractor):
             'categories': ['Science & Technology'],
         }
     }, {
+        # Issue #26002
+        'url': 'peertube:spacepub.space:d8943b2d-8280-497b-85ec-bc282ec2afdc',
+        'info_dict': {
+            'id': 'd8943b2d-8280-497b-85ec-bc282ec2afdc',
+            'ext': 'mp4',
+            'title': 'Dot matrix printer shell demo',
+            'uploader_id': '3',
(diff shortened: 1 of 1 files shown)


tpikonen:
This PR indeed fixes #26002. Please merge.

youtube_dl/extractor/peertube.py
```diff
@@ -526,7 +538,15 @@ def _real_extract(self, url):
         title = video['name']
 
         formats = []
-        for file_ in video['files']:
+        files = video['files']
+        for playlist in video['streamingPlaylists']:
```
remitamine, line 542:
```suggestion
        files = video.get('files') or []
        for playlist in (video.get('streamingPlaylists') or []):
```

remitamine (changes requested):

0l-l0:
I made an another PR (#27728) because my original repo disappeared after the DMCA take-down.

youtube_dl/extractor/peertube.py
0l-l0: Corrections in #27728

Issue #22973, opened 2019-11-04
Support for stripchat.com


Stilian1:
## Checklist

- [x] I'm reporting a new site support request
- [x] I've verified that I'm running youtube-dl version **2019.10.29**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that none of provided URLs violate any copyrights
- [x] I've searched the bugtracker for similar site support requests including closed ones

## Example URLs

- Single video: https://www.youtube.com/watch?v=BaW_jenozKc
- Single video: https://youtu.be/BaW_jenozKc
- Playlist: https://www.youtube.com/playlist?list=PL4lCao7KL_QFVb7Iudeipvc2BCavECqzc

## Description

It is similar to chaturbate.

dirkf:
From #30258:
>
> *    Single Livestream: https://stripchat.com/AmiliYan
> *    Single Livestream: https://stripchat.com/RedAssHottie
> 
> ...
> Majority of the livestreams are free, without an account, to watch.

dirkf:
The _yt-dlp_ extractor can be back-ported but both result in `Model is in a private show` for the example models and others from the "Top Free Live Sex Cams" list. Examples of actual active non-private models are needed.

PR #26508, opened 2020-09-02
[Stripchat] add extractor

zulaport:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Adds support for Stripchat, an adult webcam site. This would fix https://github.com/ytdl-org/youtube-dl/issues/22973.

### DIFF
--- a/youtube_dl/extractor/stripchat.py
+++ b/youtube_dl/extractor/stripchat.py
@@ -1,0 +1,78 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..compat import (
+    compat_str,
+)
+from ..utils import (
+    ExtractorError,
+    lowercase_escape,
+    try_get,
+)
+
+
+class StripchatIE(InfoExtractor):
+    _VALID_URL = r'https?://stripchat\.com/(?P<id>[0-9A-Za-z-_]+)'
+    _TEST = {
+        'url': 'https://stripchat.com/feel_me',
+        'info_dict': {
+            'id': 'feel_me',
+            'ext': 'mp4',
+            'title': 're:^feel_me [0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}$',
+            'description': 're:^.*$',
+            'is_live': True,
+        },
+        'params': {
+            'skip_download': True,
+        },
+        'skip': 'Room is offline',
+    }
+
+    _ROOM_OFFLINE = 'Model is offline.'
+    _ROOM_PRIVATE = 'Model is in private show.'
+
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/stripchat.py
```diff
@@ -0,0 +1,78 @@
...
+        if not data:
+            raise ExtractorError('Unable to find configuration for stream.')
+
+        show = try_get(data, lambda x: x['viewCam']['show'], dict)
+        if show:
+            raise ExtractorError(self._ROOM_PRIVATE, expected=True)
+
+        is_live = try_get(data, lambda x: x['viewCam']['model']['isLive'], bool)
+        if not is_live:
+            raise ExtractorError(self._ROOM_OFFLINE, expected=True)
```
dirkf, line 55: Simplify, harmonise with yt-dlp pt5:
```suggestion
        if try_get(data, lambda x: x['viewCam']['show'], dict):
            raise ExtractorError('Model is in private show', expected=True)
        elif not try_get(data, lambda x: x['viewCam']['model']['isLive'], bool):
            raise ExtractorError('Model is offline', expected=True)
```

youtube_dl/extractor/stripchat.py
```diff
@@ -0,0 +1,78 @@
...
+        m3u8_url = 'https://b-%s.%s/hls/%d/%d.m3u8' % (server, host, model_id, model_id)
+
+        formats = self._extract_m3u8_formats(
+            m3u8_url,
+            video_id,
+            ext='mp4',
+            entry_protocol='m3u8_native',
+            m3u8_id='hls',
+            fatal=False,
+            live=True)
```
dirkf, line 70: Simplify, harmonise with yt-dlp pt6:
```suggestion
        formats = self._extract_m3u8_formats(
            'https://b-%s.%s/hls/%d/%d.m3u8' % (server, host, model_id, model_id),
            video_id, ext='mp4', m3u8_id='hls', fatal=False, live=True)
        self._sort_formats(formats)
```

youtube_dl/extractor/stripchat.py
```diff
@@ -0,0 +1,78 @@
...
+
+        return {
+            'id': video_id,
+            'title': self._live_title(video_id),
+            'description': self._og_search_description(webpage),
+            'is_live': True,
+            'formats': formats,
+        }
```
dirkf, line 78: Sarmonise with yt-dlp pt7:
```suggestion
            # Stripchat declares the RTA meta-tag, but in an non-standard format so _rta_search() can't be used
            'age_limit': 18,
        }
```

dirkf (changes requested):
This can be merged if someone will test it against live pages.

Issue #26594, opened 2020-09-15
Option to NOT check file existence?


CrendKing:
## Checklist

- [x] I'm asking a question
- [x] I've looked through the README and FAQ for similar questions
- [x] I've searched the bugtracker for similar questions including closed ones

## Question

Is there a way to tell youtube-dl NOT to check file existence and always tries to download? My scenario is this. I setup youtube-dl so that 
* the output format is just `C:/%(title)s.%(ext)s`, which is convenient but prone to duplicate.
* uses aria2c as external downloader. For aria2c, I enable the "auto-file-renaming" feature, so that if file exists it automatically rename it.

To test, I create an empty file `test.mp4` (imagine it as a completely unrelated video). Then invoke youtube-dl to download a video titled "test" (not Youtube). Here's output:
```
[debug] System config: []
[debug] User config: []
[debug] Custom config: ['--external-downloader', 'aria2c', '--external-downloader-args', '--conf-path=C:/aria2/aria2.conf', '--output', 'C:/%(title)s.%(ext)s', '-v']
[debug] Command-line args: ['--config-location', 'C:\\youtube-dl\\ydl.conf', 'https://www.test.com/test/']
[debug] Encodings: locale cp65001, fs utf-8, out utf-8, pref cp65001
[debug] youtube-dl version 2020.09.14
[debug] Python version 3.8.5 (CPython) - Windows-10-10.0.19041-SP0
[debug] exe versions: ffmpeg git-2020-08-31-4a11a6f, ffprobe git-2020-08-31-4a11a6f
[debug] Proxy map: {}
[generic] test-mp4: Requesting header
WARNING: Falling back on generic information extractor.
[generic] test-mp4: Downloading webpage
[generic] test-mp4: Extracting information
[download] Downloading playlist: test.mp4
[generic] playlist test.mp4: Collected 1 video ids (downloading 1 of them)
[download] Downloading video 1 of 1
[debug] Default format spec: bestvideo+bestaudio/best
[debug] Invoking downloader on 'https://www.test.com/video/test.mp4'
[download] C:\test.mp4 has already been downloaded
[download] 100% of 0.00B
[download] Finished downloading playlist: test.mp4
```
It seems youtube-dl checked the file existence by itself and never sent a request to aria2c. I see the "--no-overwrites" option, but I think what I need is the opposite of it. "--continue" also doesn't help. I know adding something like `%(epoch)s` to the template will break the duplicate, but I don't want random number in the filename.

[...]

PR #26607, opened 2020-09-16
Remove the continue argument to aria2c when --no-continue is specified

CrendKing:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

When youtube-dl is launched with `--no-continue` and configured to use aria2 as external downloader, it always passes `-c` (continue) to aria2c. Here is an use case where due to this behavior it won't work:

* Output template is something like `C:/%(title)s.%(ext)s`, which is convenient but prone to duplicate.
* Make sure aria2's `auto-file-renaming` is on (as default).
* Specify `--no-continue` and `--no-part`.
* Download unrelated videos from webpages that share the same title, using generic extractor. Expected result is, all videos are downloaded and auto renamed to different names. Actual result is, only the first video is downloaded.
* I had an issue thread at https://github.com/ytdl-org/youtube-dl/issues/26594 that contains more details.

A good example is the curl handler. It checks the `continuedl` param and put `--continue-at` conditionally:
```
cmd += self._bool_option('--continue-at', 'continuedl', '-', '0')
```
I think all handlers that have continue option should follow the same rule.

### DIFF
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -183,7 +183,7 @@ class Aria2cFD(ExternalFD):
 
     def _make_cmd(self, tmpfilename, info_dict):
         cmd = [self.exe]
-        cmd += self._valueless_option('-c', 'continuedl')
+        cmd += self._bool_option('--continue', 'continuedl', separator='=')
         cmd += self._configuration_args([
             '--min-split-size', '1M', '--max-connection-per-server', '4'])
         dn = os.path.dirname(tmpfilename)
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -184,6 +184,8 @@ class Aria2cFD(ExternalFD):
     def _make_cmd(self, tmpfilename, info_dict):
         cmd = [self.exe]
         cmd += self._bool_option('--continue', 'continuedl', separator='=')
+        cmd += self._bool_option('--allow-overwrite', 'continuedl', 'false', 'true', separator='=')
+        cmd += self._bool_option('--remove-control-file', 'continuedl', 'false', 'true', separator='=')
         cmd += self._configuration_args([
             '--min-split-size', '1M', '--max-connection-per-server', '4'])
         dn = os.path.dirname(tmpfilename)
--- a/youtube_dl/downloader/external.py
+++ b/youtube_dl/downloader/external.py
@@ -182,7 +182,8 @@ class Aria2cFD(ExternalFD):
     AVAILABLE_OPT = '-v'
 
     def _make_cmd(self, tmpfilename, info_dict):
-        cmd = [self.exe, '-c']
+        cmd = [self.exe]
+        cmd += self._valueless_option('-c', 'continuedl')
         cmd += self._configuration_args([
             '--min-split-size', '1M', '--max-connection-per-server', '4'])
         dn = os.path.dirname(tmpfilename)


youtube_dl/downloader/external.py
```diff
@@ -182,7 +182,8 @@ class Aria2cFD(ExternalFD):
     AVAILABLE_OPT = '-v'
 
     def _make_cmd(self, tmpfilename, info_dict):
-        cmd = [self.exe, '-c']
+        cmd = [self.exe]
+        cmd += self._valueless_option('-c', 'continuedl')
```
dstftw, line 186: This has no effect for youtube-dl's `--continue`/`--no-continue` - in both cases downloading is resumed. This is because you must also pass `--allow-overwrite=true` and `--remove-control-file=true` is case of youtube-dl's `--no-continue`.

dstftw: Use full `--continue=true|false` notation instead.

dstftw (changes requested):

youtube_dl/downloader/external.py
CrendKing: > This has no effect for youtube-dl's `--continue`/`--no-continue` - in both cases downloading is resumed. This is because you must also pass `--allow-overwrite=true` and `--remove-control-file=true` is case of youtube-dl's `--no-continue`.

Do you mean I should check for ydl's `no-overwrites` flag and add aria2's `allow-overwrite` here? I can do it, but isn't it better to change that in a new PR? That logic wasn't there before.

But if you mean to add `allow-overwrite` for `no-continue`, you misunderstand my intention. The scenario I'm having is actually not overwriting file with conflict output name, but instead saving those files with different names thanks to aria2's auto renaming feature. Imagine I have a file "test.mp4" and try to download another "test.mp4". Instead of overwriting or continuing, I want aria2 to download it and rename the new file to "test.1.mp4". If more, "test.2.mp4", "test.3.mp4".

If user wants to overwrite, they can always put the flags you mentioned in their aria2.conf. In my opinion, youtube-dl should pass as few hard coded flags to external downloader as possible, because user can't control them.

> Use full --continue=true|false notation instead.

Will do.

dstftw: >Do you mean I should check for ydl's `no-overwrites` flag and add aria2's `allow-overwrite` here?

No. `--no-overwrites` is for non media files.

>But if you mean to add `allow-overwrite` for `no-continue`, you misunderstand my intention.

The point is not your intention but replicating the actual behavior of youtube-dl itself. With `--continue` youtube-dl resumes download, aria2c as external downloader should do the same, so that a command should be passed to aria2c that will instruct it to resume. When `--no-continue` is used youtube-dl downloads from scratch, same should do aria2c.

CrendKing: > When --no-continue is used youtube-dl downloads from scratch, same should do aria2c.

I agree with this. But "from scratch" does not necessarily mean to always overwrite existing (unrelated) file just because they have same name. There is a good reason why aria2 provides `allow-overwrite` and default it to false in the first place.

If youtube-dl hardcode `--allow-overwrite=true` here, user who does not want to overwrite can't change the behavior. If youtube-dl does not hardcode, user who does want to overwrite can still provide that flag in their `--external-downloader-args`. Wouldn't the more flexible approach be better? Or maybe am I missing some project design principle that favors hardcoding commonly used combinations of flags?

dstftw: I've already pointed out: by default it should work the way youtube-dl works. If user wants to override external downloader behavior he should explicitly provide desired options via `--external-downloader-args` (these options should override the defaults by any suitable way external downloader supports - by putting at front/back or whetever).
As for link mentioned that's not a problem at all as it's easily solved by proper output template.

CrendKing: Updated. Please take a look.

> these options should override the defaults by any suitable way external downloader supports - by putting at front/back or whetever

I usually try to avoid passing conflict arguments to external programs because the order of override is generally implementation detail. For some first appeared wins while for others the last appeared wins (this is the case for aria2). They could change the behavior from version to version since it is not a contract.

PR #26629, opened 2020-09-19
[skyitalia] Add new extractor

nixxo:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

A new extractor for the websites sky.it (Sky Italy)

### DIFF
--- a/youtube_dl/extractor/skyitalia.py
+++ b/youtube_dl/extractor/skyitalia.py
@@ -1,0 +1,146 @@
+﻿# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import ExtractorError
+
+
+class SkyItaliaBaseIE(InfoExtractor):
+    _GET_VIDEO_DATA = 'https://apid.sky.it/vdp/v1/getVideoData?token={token}&caller=sky&rendition=web&id={id}'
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/skyitalia.py
```diff
@@ -0,0 +1,136 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import ExtractorError
+
+
+class SkyItaliaBaseIE(InfoExtractor):
+    _GET_VIDEO_DATA = 'https://apid.sky.it/vdp/v1/getVideoData?token={token}&caller=sky&rendition=web&id={id}'
```
dstftw, line 9: This must be a complete separate extractor. All other extractors must delegate to it not inherit.

dstftw (changes requested):

youtube_dl/extractor/skyitalia.py
nixxo: i'm new to python so i'm not sure i understood what is asked.

As in 6b427abc4dc5988a9baf0fde94b9b57c521aa174 I moved the _real_extract method to the "general" extractor and all the other "specific" ones uses the same instead of having one each. Is that ok?

PR #26637, opened 2020-09-20
Fix incorrect ANSI sequence for restoring console-window title

glenn-slayden:
The ANSI terminal sequence is incorrect:

https://github.com/ytdl-org/youtube-dl/blob/b55715934bb7f9474f69b99e4d51cc83dee7cbef/youtube_dl/YoutubeDL.py#L538

It must use the form `\033[` -- with a ***left*** (opening) square bracket... `[` , not the right/closing one.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -535,7 +535,7 @@ class YoutubeDL(object):
                 # already of type unicode()
                 ctypes.windll.kernel32.SetConsoleTitleW(ctypes.c_wchar_p(message))
         elif 'TERM' in os.environ:
-            self._write_string('\033]0;%s\007' % message, self._screen_file)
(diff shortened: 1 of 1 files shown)


h1z1:
That is a dangerous piece of code to begin with but logic is correct.  It's printing an escape, the callers are passing ansi to it.  Never udnerstood why they didn't use termios but guessing it was for portability reasons.

kikuyan:
It is OSC (Operating System Command), `ESC ]`. This PR should not be merged.
Refs: 
https://en.wikipedia.org/wiki/ANSI_escape_code#OSC_(Operating_System_Command)_sequences
https://www.xfree86.org/current/ctlseqs.html
https://man7.org/linux/man-pages/man4/console_codes.4.html

glenn-slayden:
Duly noted. Thank you.

PR #26643, opened 2020-09-20
Cache the download archive in a list (fast!)

rebane2001:

---

- [ ] Bug fix
- [X] Improvement
- [ ] New extractor
- [ ] New feature

---

Right now, youtube-dl's archive file is checked for every video it is being matched against. This is very inefficient and causes a lot of unnecessary IO. While not really a problem on smaller archives, it gets really slow if you have tens or hundreds of thousands of videos in a single archive. This pr adds a list that is used as a cache, leading to over 100x speed increase on my setup.

Comparisons: [original](https://i.imgur.com/vLV8ACV.mp4), [new](https://i.imgur.com/5lfU1UE.mp4).

There exists the pr #24852, but it has been in a pending-fixes/invalid state for months now, so I propose this pr as a simpler and smaller change to achieve the same goal, so it could be merged straight away.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -339,6 +339,7 @@ class YoutubeDL(object):
     _download_retcode = None
     _num_downloads = None
     _screen_file = None
+    archive_list = []
 
     def __init__(self, params=None, auto_init=True):
         """Create a FileDownloader object with the given options."""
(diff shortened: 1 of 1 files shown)


ibex-are-goats:
@dstftw Which algorithm do you like more, #24852 or #26643 ?

youtube_dl/YoutubeDL.py
```diff
@@ -339,6 +339,7 @@ class YoutubeDL(object):
     _download_retcode = None
     _num_downloads = None
     _screen_file = None
+    archive_list = []
```
serkef, line 342: a `set` will have significant better performance to check membership (line 2114)

rebane2001: I seem to be unable to edit the commit for some reason, perhaps it's because this PR is marked as closed?

Issue #6231, opened 2015-07-14
Bandcamp and Youtube-dl


NickWinston123:
Is there any way to do youtube-dl username.bandcamp.com and it download all of the albums? Is that supported yet?

When I tried it gave me this error.

nicks-mac-mini:~ noah$ youtube-dl http://liluglymane.bandcamp.com --verbose flag
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'http://liluglymane.bandcamp.com', u'--verbose', u'flag']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.07.07
[debug] Python version 2.7.10 - Darwin-13.3.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg 2.7.1, ffprobe 2.7.1
[debug] Proxy map: {}
[Bandcamp:album] liluglymane: Downloading webpage
ERROR: The page doesn't contain any tracks; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 654, in extract_info
    ie_result = ie.extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 273, in extract
    return self._real_extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/bandcamp.py", line 169, in _real_extract
    raise ExtractorError('The page doesn\'t contain any tracks')
ExtractorError: The page doesn't contain any tracks; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.

dstftw:
That is not supported. Only per album.

djvs:
A quick ruby script to hack this - requires curb and nokogiri libraries:

    #!/usr/bin/env ruby

    require 'curb'
    require 'nokogiri'

    targetdir = Dir.pwd
    url = ARGV[0]
    yta = 'youtube-dl --no-mtime -o "%(autonumber)s-%(title)s.%(ext)s"'
[...]

PR #26684, opened 2020-09-23
[bandcamp] fix regexp for JSON matching on bandcamp

### LINKED ISSUE #26681: Bandcamp is broken
## Checklist

- [x] I'm reporting a broken site support
- [x] I've verified that I'm running youtube-dl version **2020.09.20**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
youtube-dl -v https://proper-t.bandcamp.com/album/the-project-ivy-part-2-beauty-sleep
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', 'https://proper-t.bandcamp.com/album/the-project-ivy-part-2-beauty-sleep']
[debug] youtube-dl version 2020.09.20
[debug] Python version 3.8.5 (CPython) - macOS-10.15.6-x86_64-i386-64bit
[debug] exe versions: ffmpeg 4.3.1, ffprobe 4.3.1, rtmpdump 2.4
[debug] Proxy map: {}
[Bandcamp:album] the-project-ivy-part-2-beauty-sleep: Downloading webpage
WARNING: unable to extract title; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
[download] Downloading playlist: the-project-ivy-part-2-beauty-sleep
[Bandcamp:album] playlist the-project-ivy-part-2-beauty-sleep: Collected 7 video ids (downloading 7 of them)
[download] Downloading video 1 of 7
[Bandcamp] beauty-sleep: Downloading webpage
ERROR: No video formats found; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
  File "/usr/local/lib/python3.8/site-packages/youtube_dl/YoutubeDL.py", line 797, in extract_info
    ie_result = ie.extract(url)
  File "/usr/local/lib/python3.8/site-packages/youtube_dl/extractor/common.py", line 532, in extract
  File "/usr/local/lib/python3.8/site-packages/youtube_dl/extractor/bandcamp.py", line 196, in _real_extract
    self._sort_formats(formats)
  File "/usr/local/lib/python3.8/site-packages/youtube_dl/extractor/common.py", line 1359, in _sort_formats
    raise ExtractorError('No video formats found')
youtube_dl.utils.ExtractorError: No video formats found; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

## Description

Bandcamp links return error today

gilou:

- [X] Bug fix

---

Bandcamp changed their JSON format using `&quot;` entities all over, so this tries to match it so that youtube-dl can get the track info and work its magic. This most likely will fix #26681

### DIFF
--- a/youtube_dl/extractor/bandcamp.py
+++ b/youtube_dl/extractor/bandcamp.py
@@ -33,7 +33,7 @@ class BandcampIE(InfoExtractor):
         'info_dict': {
             'id': '1812978515',
             'ext': 'mp3',
-            'title': "youtube-dl  \"'/\\\u00e4\u21ad - youtube-dl  \"'/\\\u00e4\u21ad - youtube-dl test song \"'/\\\u00e4\u21ad",
+            'title': "youtube-dl  \"'/\\\u00e4\u21ad - youtube-dl test song \"'/\\\u00e4\u21ad",
             'duration': 9.8485,
             'uploader': "youtube-dl  \"'/\\\u00e4\u21ad",
             'timestamp': 1354224127,
@@ -99,7 +99,6 @@ class BandcampIE(InfoExtractor):
             webpage, 'track info', default='{}')
 
         track_info = self._parse_json(trackinfo_block, title)
-
         if track_info:
             file_ = track_info.get('file')
             if isinstance(file_, dict):
@@ -115,7 +114,7 @@ class BandcampIE(InfoExtractor):
                         'acodec': ext,
                         'abr': int_or_none(abr_str),
                     })
-            track = track_info.get('title')
+
             track_id = str_or_none(track_info.get('track_id') or track_info.get('id'))
             track_number = int_or_none(track_info.get('track_num'))
             duration = float_or_none(track_info.get('duration'))
@@ -126,6 +125,7 @@ class BandcampIE(InfoExtractor):
                 webpage, key, default=None, group='value')
             return data.replace(r'\"', '"').replace('\\\\', '\\') if data else data
 
+        track = extract('title')
         artist = extract('artist')
         album = extract('album_title')
         timestamp = unified_timestamp(
--- a/youtube_dl/extractor/bandcamp.py
+++ b/youtube_dl/extractor/bandcamp.py
@@ -316,10 +316,10 @@ class BandcampAlbumIE(InfoExtractor):
             if self._html_search_meta('duration', elem_content, default=None)]
 
         title = self._html_search_regex(
-            r'album_title\s*:\s*"((?:\\.|[^"\\])+?)"',
+            r'album_title\s*(?:&quot;|["\']):\s*(?:&quot;|["\'])((?:\\.|[^"\\])+?)(?:&quot;|["\'])',
             webpage, 'title', fatal=False)
         if title:
-            title = title.replace(r'\"', '"')
+            title = unescapeHTML(title)
         return {
             '_type': 'playlist',
             'uploader_id': uploader_id,
--- a/youtube_dl/extractor/bandcamp.py
+++ b/youtube_dl/extractor/bandcamp.py
@@ -33,8 +33,11 @@ class BandcampIE(InfoExtractor):
         'info_dict': {
             'id': '1812978515',
             'ext': 'mp3',
-            'title': "youtube-dl  \"'/\\\u00e4\u21ad - youtube-dl test song \"'/\\\u00e4\u21ad",
+            'title': "youtube-dl  \\ - youtube-dl  \"'/\\\u00e4\u21ad - youtube-dl test song \"'/\\\u00e4\u21ad",
             'duration': 9.8485,
+            'uploader': 'youtube-dl  \\',
+            'timestamp': 1354224127,
+            'upload_date': '20121129',
         },
         '_skip': 'There is a limit of 200 free downloads / month for the test song'
     }, {
--- a/youtube_dl/extractor/bandcamp.py
+++ b/youtube_dl/extractor/bandcamp.py
@@ -92,10 +92,10 @@ class BandcampIE(InfoExtractor):
 
         formats = []
         trackinfo_block = self._search_regex(
-            r'trackinfo&quot;:\[\s*({.+?})\s*\],&quot;',
+            r'trackinfo(?:["\']|&quot;):\[\s*({.+?})\s*\],(?:["\']|&quot;)',
             webpage, 'track info', default='{}')
-        quoted_json = trackinfo_block.replace('&quot;', '"')
-        track_info = self._parse_json(quoted_json, title)
+        unescaped_json = unescapeHTML(trackinfo_block)
+        track_info = self._parse_json(unescaped_json, title)
         if track_info:
             file_ = track_info.get('file')
             if isinstance(file_, dict):
@@ -118,7 +118,7 @@ class BandcampIE(InfoExtractor):
 
         def extract(key):
             return self._search_regex(
-                r',&quot;%s&quot;:(&quot;)(?P<value>(?:(?!&quot;).)+)&quot;' % key,
+                r',(["\']|&quot;)%s\1:\1(?P<value>(?:(?!\1).)+)\1' % key,
                 webpage, key, default=None, group='value')
 
         artist = extract('artist')
--- a/youtube_dl/extractor/bandcamp.py
+++ b/youtube_dl/extractor/bandcamp.py
@@ -91,10 +91,11 @@ class BandcampIE(InfoExtractor):
         duration = None
 
         formats = []
-        track_info = self._parse_json(
-            self._search_regex(
-                r'trackinfo\s*:\s*\[\s*({.+?})\s*\]\s*,\s*?\n',
-                webpage, 'track info', default='{}'), title)
+        trackinfo_block = self._search_regex(
+            r'trackinfo&quot;:\[\s*({.+?})\s*\],&quot;',
+            webpage, 'track info', default='{}')
+        quoted_json = trackinfo_block.replace('&quot;', '"')
+        track_info = self._parse_json(quoted_json, title)
         if track_info:
             file_ = track_info.get('file')
             if isinstance(file_, dict):
@@ -117,7 +118,7 @@ class BandcampIE(InfoExtractor):
 
         def extract(key):
             return self._search_regex(
-                r'\b%s\s*["\']?\s*:\s*(["\'])(?P<value>(?:(?!\1).)+)\1' % key,
+                r',&quot;%s&quot;:(&quot;)(?P<value>(?:(?!&quot;).)+)&quot;' % key,
                 webpage, key, default=None, group='value')
 
         artist = extract('artist')
--- a/youtube_dl/extractor/bandcamp.py
+++ b/youtube_dl/extractor/bandcamp.py
@@ -128,12 +128,12 @@ class BandcampIE(InfoExtractor):
         release_date = unified_strdate(extract('album_release_date'))
 
         download_link = self._search_regex(
-            r'freeDownloadPage\s*:\s*(["\'])(?P<url>(?:(?!\1).)+)\1', webpage,
+            r'freeDownloadPage(?:["\']|&quot;):\s*(["\']|&quot;)(?P<url>(?:(?!\1).)+)\1', webpage,
             'download link', default=None, group='url')
         if download_link:
             track_id = self._search_regex(
-                r'(?ms)var TralbumData = .*?[{,]\s*id: (?P<id>\d+),?$',
-                webpage, 'track id')
+                r'\?id=(?P<id>\d+)&',
+                download_link, 'track id')
 
             download_webpage = self._download_webpage(
                 download_link, track_id, 'Downloading free downloads page')
--- a/youtube_dl/extractor/bandcamp.py
+++ b/youtube_dl/extractor/bandcamp.py
@@ -33,9 +33,9 @@ class BandcampIE(InfoExtractor):
         'info_dict': {
             'id': '1812978515',
             'ext': 'mp3',
-            'title': "youtube-dl  \\ - youtube-dl  \"'/\\\u00e4\u21ad - youtube-dl test song \"'/\\\u00e4\u21ad",
+            'title': "youtube-dl  \"'/\\\u00e4\u21ad - youtube-dl  \"'/\\\u00e4\u21ad - youtube-dl test song \"'/\\\u00e4\u21ad",
             'duration': 9.8485,
-            'uploader': 'youtube-dl  \\',
+            'uploader': "youtube-dl  \"'/\\\u00e4\u21ad",
             'timestamp': 1354224127,
             'upload_date': '20121129',
         },
@@ -43,7 +43,7 @@ class BandcampIE(InfoExtractor):
     }, {
         # free download
         'url': 'http://benprunty.bandcamp.com/track/lanius-battle',
-        'md5': '853e35bf34aa1d6fe2615ae612564b36',
+        'md5': '5d92af55811e47f38962a54c30b07ef0',
         'info_dict': {
             'id': '2650410135',
             'ext': 'aiff',
@@ -94,11 +94,12 @@ class BandcampIE(InfoExtractor):
         duration = None
 
         formats = []
-        trackinfo_block = self._search_regex(
+        trackinfo_block = self._html_search_regex(
             r'trackinfo(?:["\']|&quot;):\[\s*({.+?})\s*\],(?:["\']|&quot;)',
             webpage, 'track info', default='{}')
-        unescaped_json = unescapeHTML(trackinfo_block)
-        track_info = self._parse_json(unescaped_json, title)
+
+        track_info = self._parse_json(trackinfo_block, title)
+
         if track_info:
             file_ = track_info.get('file')
             if isinstance(file_, dict):
@@ -120,9 +121,10 @@ class BandcampIE(InfoExtractor):
             duration = float_or_none(track_info.get('duration'))
 
         def extract(key):
-            return self._search_regex(
-                r',(["\']|&quot;)%s\1:\1(?P<value>(?:(?!\1).)+)\1' % key,
+            data = self._html_search_regex(
+                r',(["\']|&quot;)%s\1:\1(?P<value>(?:\\\1|((?!\1).))+)\1' % key,
                 webpage, key, default=None, group='value')
+            return data.replace(r'\"', '"').replace('\\\\', '\\') if data else data
 
         artist = extract('artist')
         album = extract('album_title')
@@ -319,10 +321,12 @@ class BandcampAlbumIE(InfoExtractor):
             if self._html_search_meta('duration', elem_content, default=None)]
 
         title = self._html_search_regex(
-            r'album_title\s*(?:&quot;|["\']):\s*(?:&quot;|["\'])((?:\\.|[^"\\])+?)(?:&quot;|["\'])',
-            webpage, 'title', fatal=False)
+            r'album_title\s*(?:&quot;|["\']):\s*(&quot;|["\'])(?P<album>(?:\\\1|((?!\1).))+)\1',
+            webpage, 'title', fatal=False, group='album')
+
         if title:
-            title = unescapeHTML(title)
+            title = title.replace(r'\"', '"')
+
         return {
             '_type': 'playlist',
             'uploader_id': uploader_id,
--- a/youtube_dl/extractor/bandcamp.py
+++ b/youtube_dl/extractor/bandcamp.py
@@ -93,12 +93,30 @@ class BandcampIE(InfoExtractor):
         track_number = None
         duration = None
 
-        formats = []
-        trackinfo_block = self._html_search_regex(
-            r'trackinfo(?:["\']|&quot;):\[\s*({.+?})\s*\],(?:["\']|&quot;)',
-            webpage, 'track info', default='{}')
+        scriptdatablocks = re.findall(r'<script type="text/javascript" src="[^"]+" nonce="[^"]+" (?:data-[a-z-]+="[^"]+" ?)+>', webpage)
+        bandcamp_data = {}
+        for block in scriptdatablocks:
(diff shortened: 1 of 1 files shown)


morrah:
some albums have free download urls (`download_link` regex) which key is quoted too now
test track for free download is http://benprunty.bandcamp.com/track/lanius-battle - .aiff format should be downloaded (but test fails anyway because md5 doesn't match for some reason)
also `track_id` regex changed

i tried to patch it myself, though not sure if using `unescapeHTML` on the whole page is the correct way https://github.com/morrah/youtube-dl/commit/8bc998aee583cba8589020e0ee1681c2be492097

gilou:
@morrah ah, I didn't know about that one, I'd stick with matching the actual original stuff, as there is only a low chance of them actually changing back to non-entities for that part, or we could just do something else I thought about: include a `(&quot;|["\'])` group, non-capturing as needed `(?:&quot;|["\'])`  so we match the entities as well as the quotes. But then I thought well, they're not going to switch back… dunno :P

morrah:
@gilou my point is that `test/test_download.py` for `BandcampIE` extractor should be done, and one of the tests is an ability to download higher quality tracks when an album is free for download;

regarding "regex first, then replace entities, then parse json" (as you did) vs "replace entities first, then regex, then parse json" (as i did) – yea, matching the original data is possibly better;

anyway there could be other entities (in track titles for example), not `&quot;` only, so imo it makes sense to use `unescapeHTML` instead of `replace`

gilou:
as for entities, agreed. I changed my code to match all quotes (i.e. normal or entities) even though I doubt that'll catch anything interesting, and I used the unescapeHTML function instead of the raw replace. 

I'm not sure about the download page, but it seems they changed their way… I'll look into it. Does your branch work for that usecase ?

iamleot:
When fetching albums, e.g.:

```
% youtube-dl -f bestaudio https://clowncore.bandcamp.com/album/toilet
```

...there is still a warning:

```
[Bandcamp:album] toilet: Downloading webpage
WARNING: unable to extract title; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  h
ttps://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
[download] Downloading playlist: toilet
```

@gilou can you please adjust `title` assignment in `BandcampAlbumIE` too?

gilou:
@iamleot there you go! That should be fixed ;)

iamleot:
Thanks!

In the meantime I was trying to think about such REs and... what would happens if the title is e.g. `"Quoted"`?  Doesn't the RE matches the end while it's actually the first `\&quot;` (please note the leading `\`!) or last `\&quot;` part of the title?

gilou:
https://github.com/ytdl-org/youtube-dl/pull/26684/commits/9385ec4b1c797ffab66b945f23fd4248c0c8a32e should fix the free download page for  http://benprunty.bandcamp.com/track/lanius-battle e.g. it seems to get the .aiff as wished. I need to test but I'm just starting to understand the way testing is done here…

gilou:
> In the meantime I was trying to think about such REs and... what would happens if the title is e.g. `"Quoted"`? Doesn't the RE matches the end while it's actually the first `\&quot;` (please note the leading `\`!) or last `\&quot;` part of the title?

I'll check, it seems that it won't be escaped that way, but if you have an example, it might be good to have it as a test ;)

iamleot:
I think this is a possible example for `album_title`: https://bloodntongue.bandcamp.com/album/quote

iamleot:
The previous title - with the current patch applied get parsed as `\"Quote\"` (the `\` are due JSON parsed as HTML and so `\"` not consumed as `"`).  (This isn't a good example to trigger the possibly problematic `\&quot;` though.)

youtube_dl/extractor/bandcamp.py
```diff
@@ -316,10 +316,10 @@ def _real_extract(self, url):
             if self._html_search_meta('duration', elem_content, default=None)]
 
         title = self._html_search_regex(
-            r'album_title\s*:\s*"((?:\\.|[^"\\])+?)"',
+            r'album_title\s*(?:&quot;|["\']):\s*(?:&quot;|["\'])((?:\\.|[^"\\])+?)(?:&quot;|["\'])',
             webpage, 'title', fatal=False)
         if title:
-            title = title.replace(r'\"', '"')
```
iamleot, line 321: The `title.replace(r'\"', '"')` is still needed.

iamleot (changes requested):
Please readd the `title = title.replace(r'\"', '"')`, as pointed out in the possible example case shared.

gilou:
Yeah, there is something off anyway with quoting, which may have lead me to do something weird in the tests.

I'll switch to html_search to clean up entities, and add back the quote replace. Also, this checksum thing is really weird, I can reproduce, but whatever it gets, the MD5 is never matching what I download "manually", that's weird.

gilou:
If I clean up the quoting regexp, I get \"Quote\" in that example, which is somehow correct, but it changes to Quote_ when fed through the template…

iamleot:
> Yeah, there is something off anyway with quoting, which may have lead me to do something weird in the tests.
> 
> I'll switch to html_search to clean up entities, and add back the quote replace. Also, this checksum thing is really weird, I can reproduce, but whatever it gets, the MD5 is never matching what I download "manually", that's weird.

Thanks!

Rearding the checksum, I think only `n` kb-s are actually fetched so I would just substitute the current one with a placeholder `XXX` or similar, rerun the tests and replace the `XXX` with the updated checksum and rerun the tests again.  If the checksum is still not matching probably the content is in some way dynamically generated.

gilou:
Well, unless it is, if I run `python3 test/test_download.py TestDownload.test_Bandcamp_1` I get
```
FAIL: test_Bandcamp_1 (__main__.TestDownload):
  File "test/test_download.py", line 228, in test_template
    self.assertEqual(tc['md5'], md5_for_file)
AssertionError: '853e35bf34aa1d6fe2615ae612564b36' != '5d92af55811e47f38962a54c30b07ef0'
- 853e35bf34aa1d6fe2615ae612564b36
+ 5d92af55811e47f38962a54c30b07ef0

```

If 5d92af55811e47f38962a54c30b07ef0 is the value you get as well, let's just update it ;)

gilou:
For those following, I force-pushed the PR branch because I forgot a flake8 cleanup (not in my hooks for youtube-dl, my bad)

iamleot:
> [...] 
> If 5d92af55811e47f38962a54c30b07ef0 is the value you get as well, let's just update it ;)

Yes, I can confirm that the updated checksum matches for me too.

Thanks!

youtube_dl/extractor/bandcamp.py
```diff
@@ -33,14 +33,17 @@ class BandcampIE(InfoExtractor):
         'info_dict': {
             'id': '1812978515',
             'ext': 'mp3',
-            'title': "youtube-dl  \"'/\\\u00e4\u21ad - youtube-dl test song \"'/\\\u00e4\u21ad",
+            'title': "youtube-dl  \"'/\\\u00e4\u21ad - youtube-dl  \"'/\\\u00e4\u21ad - youtube-dl test song \"'/\\\u00e4\u21ad",
```
iamleot, line 36: I think that's the title is incorrectly parsed, i.e. the test case shouldn't be adjusted.

iamleot (changes requested):
Why the title in the test-case was changed?  I think that the title it's incorrectly parsed and the code should be adjusted, not the test-case.

youtube_dl/extractor/bandcamp.py
gilou: OK, that's how trackinfo lists the title, I'll change how it was done to use the main one…

gilou: (there is no way that test was passing before using the track title from trackinfo, that's why I went for it, but I can fix this, hopefully)

iamleot:
> [...]
> regarding "regex first, then replace entities, then parse json" (as you did) vs "replace entities first, then regex, then parse json" (as i did) – yea, matching the original data is possibly better;
> [...]

IMO now the regular expressions are getting pretty tricky and I think that another possible approach is: "get the corresponding HTML tag attribute (probably no need for REs, just a `str.find(...)` or similar given that it should be `<attribute>="<quoted_json>"` and then parse json".

gilou:
> > [...]
> > regarding "regex first, then replace entities, then parse json" (as you did) vs "replace entities first, then regex, then parse json" (as i did) – yea, matching the original data is possibly better;
> > [...]
> 
> IMO now the regular expressions are getting pretty tricky and I think that another possible approach is: "get the corresponding HTML tag attribute (probably no need for REs, just a `str.find(...)` or similar given that it should be `<attribute>="<quoted_json>"` and then parse json".

I really don't know about the policy generally here, but I wouldn't use regexp at all here, would match using xpath or similar down to the `data-* ` attributes, JSON parse them and be done, but that will break every time they decide to update it…

morrah:
> IMO now the regular expressions are getting pretty tricky and I think that another possible approach is: "get the corresponding HTML tag attribute (probably no need for REs, just a `str.find(...)` or similar given that it should be `<attribute>="<quoted_json>"` and then parse json".

thought the same yesterday and tried to implement https://github.com/morrah/youtube-dl/commit/fdb591e2f16b1c23a9e2d573bfb0b1d47a426706 - using RE just to retrieve attributes' json-content

issues i've met:
1. `BandcampIE` doesn't pass one of the md5 tests (replaced it too);
2. `BandcampWeeklyIE` doesn't pass ext test (mp3 is a preferred option and opus is expected, possibly should be replaced too);
3. `BandcampWeeklyIE` doesn't have `episode_number` anymore, or it's just me who couldn't find it (removed all `episode_number` dependant code, not sure if it's correct).

all other tests are passed
hope it helps somehow

gilou:
I made a branch doing that as an alternative approach :
 https://github.com/ytdl-org/youtube-dl/commit/2f56e2c303b3904983e928058003b5c2c864ac90

https://github.com/gilou/youtube-dl/tree/data_approach

gilou:
ah @morrah  we used a slightly different approach, but we did it twice now… :P well, I don't know if it's better to parse the data-* attributes, anyway, it's a proposition

gilou:
since it's basically a dict merge, it could be done in a more pythonic way, let me know if I should switch to that approach!

gilou:
>     1. `BandcampIE` doesn't pass one of the md5 tests (replaced it too);

I'll update it as well, agreed it can be changed, I'll check that I have the same value as you
> 
>     2. `BandcampWeeklyIE` doesn't pass ext test (mp3 is a preferred option and opus is expected, possibly should be replaced too);
> 
>     3. `BandcampWeeklyIE` doesn't have `episode_number` anymore, or it's just me who couldn't find it (removed all `episode_number` dependant code, not sure if it's correct).

I see the info in the meta, so, might just be missing it ;)

Anyway the data-* approach if it makes us happy can be reused in the album / weekly extractor, but I'm not sure we want to go there yet?

morrah:
i'm neither maintainer nor collaborator, can just propose and wait as well :D

orbea:
This link still doesn't work with this PR applied to `2020.09.20`.

https://cloudsofficial.bandcamp.com/album/durere
```
$ youtube-dl --verbose 'https://cloudsofficial.bandcamp.com/album/durere'
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['--verbose', 'https://cloudsofficial.bandcamp.com/album/durere']
[debug] youtube-dl version 2020.09.20
[debug] Python version 3.8.5 (CPython) - Linux-5.4.58-x86_64-AMD_FX-tm-6350_Six-Core_Processor-with-glibc2.2.5
[debug] exe versions: ffmpeg git-2020-08-13-cfc6552, ffprobe git-2020-08-13-cfc6552, rtmpdump 2.4
[debug] Proxy map: {}
[Bandcamp:album] durere: Downloading webpage
ERROR: The page doesn't contain any tracks; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
  File "/usr/lib64/python3.8/site-packages/youtube_dl/YoutubeDL.py", line 797, in extract_info
    ie_result = ie.extract(url)
  File "/usr/lib64/python3.8/site-packages/youtube_dl/extractor/common.py", line 532, in extract
  File "/usr/lib64/python3.8/site-packages/youtube_dl/extractor/bandcamp.py", line 311, in _real_extract
    raise ExtractorError('The page doesn\'t contain any tracks')
youtube_dl.utils.ExtractorError: The page doesn't contain any tracks; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

ccfb3ee765a58cae:
Using 03edd545a9e14b0fbcb36574248d8cf0e7a224d6

Currently not working for this link: `https://sirensceol.bandcamp.com/album/the-method-to-love`

```
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['--add-metadata', '--verbose', 'https://sirensceol.bandcamp.com/album/the-method-to-love']
[debug] youtube-dl version 2020.09.20
[debug] Python version 3.8.5 (CPython) - Linux-5.8.13-arch1-1-x86
_64-with-glibc2.2.5
[debug] exe versions: ffmpeg 4.3.1, ffprobe 4.3.1, rtmpdump 2.4
[debug] Proxy map: {}
[Bandcamp:album] the-method-to-love: Downloading webpage
ERROR: The page doesn't contain any tracks; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
  File "/home/creato/repos/youtube-dl/youtube-dl/youtube_dl/YoutubeDL.py", line 797, in extract_info
    ie_result = ie.extract(url)
  File "/home/creato/repos/youtube-dl/youtube-dl/youtube_dl/extractor/common.py", line 532, in extract
  File "/home/creato/repos/youtube-dl/youtube-dl/youtube_dl/extractor/bandcamp.py", line 311, in _real_extract
    raise ExtractorError('The page doesn\'t contain any tracks')
youtube_dl.utils.ExtractorError: The page doesn't contain any tracks; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

vberlier:
I managed to make this PR work for albums by modifying the `_real_extract` method for `BandcampAlbumIE`:

```python
    def _real_extract(self, url):
        uploader_id = mobj.group('subdomain')
        album_id = mobj.group('album_id')
        playlist_id = album_id or uploader_id
        webpage = self._download_webpage(url, playlist_id)
        ld = re.findall(r'<script[^>]+type="application/ld\+json"[^>]*>(.*?)</script>', webpage, re.DOTALL)
        album = json.loads(ld[0])
        entries = [
            self.url_result(
                record['item']['@id'],
                ie=BandcampIE.ie_key(),
                video_title=record['item']['name'])
            for record in album['track']['itemListElement']
        ]
        title = album['name']
        return {
            '_type': 'playlist',
            'uploader_id': uploader_id,
            'id': playlist_id,
            'title': title,
            'entries': entries,
        }
```

The track info is retrieved directly from the json-ld embedded in the page.

cleslie577:
> I managed to make this PR work for albums by modifying the `_real_extract` method for `BandcampAlbumIE`:
>
I'm not quite sure where to paste this so I will wait for an official update but thanks a ton, hopefully there is a patch soon.

vberlier:
@cleslie577 no problem, the function is line 302 in [youtube_dl/extractors/bandcamp.py](https://github.com/ytdl-org/youtube-dl/blob/03edd545a9e14b0fbcb36574248d8cf0e7a224d6/youtube_dl/extractor/bandcamp.py#L302).

akadaedalus:
note for the python-challenged like me:  @vberlier's solution needs an "import json" near the top to work.  But it did work.

vberlier:
Yeah I forgot to mention the additional import. Just edited the original comment.

orbea:
Here is a patch applying @vberlier's changes on top of this PR. Seems to work for me, thanks!
```
diff -urN youtube-dl.orig/youtube_dl/extractor/bandcamp.py youtube-dl/youtube_dl/extractor/bandcamp.py
--- youtube-dl.orig/youtube_dl/extractor/bandcamp.py	2020-10-10 02:49:49.037550504 -0700
+++ youtube-dl/youtube_dl/extractor/bandcamp.py	2020-10-10 02:50:53.742191800 -0700
@@ -1,5 +1,6 @@
 
+import json
 import random
 import re
 import time
@@ -305,28 +306,16 @@
         album_id = mobj.group('album_id')
         playlist_id = album_id or uploader_id
         webpage = self._download_webpage(url, playlist_id)
-        track_elements = re.findall(
-            r'(?s)<div[^>]*>(.*?<a[^>]+href="([^"]+?)"[^>]+itemprop="url"[^>]*>.*?)</div>', webpage)
-        if not track_elements:
-            raise ExtractorError('The page doesn\'t contain any tracks')
-        # Only tracks with duration info have songs
+        ld = re.findall(r'<script[^>]+type="application/ld\+json"[^>]*>(.*?)</script>', webpage, re.DOTALL)
+        album = json.loads(ld[0])
         entries = [
             self.url_result(
-                compat_urlparse.urljoin(url, t_path),
+                record['item']['@id'],
                 ie=BandcampIE.ie_key(),
-                video_title=self._search_regex(
-                    r'<span\b[^>]+\bitemprop=["\']name["\'][^>]*>([^<]+)',
-                    elem_content, 'track title', fatal=False))
-            for elem_content, t_path in track_elements
-            if self._html_search_meta('duration', elem_content, default=None)]
-
-        title = self._html_search_regex(
-            r'album_title\s*(?:&quot;|["\']):\s*(&quot;|["\'])(?P<album>(?:\\\1|((?!\1).))+)\1',
-            webpage, 'title', fatal=False, group='album')
-
-        if title:
-            title = title.replace(r'\"', '"')
-
+                video_title=record['item']['name'])
+            for record in album['track']['itemListElement']
+        ]
+        title = album['name']
         return {
             '_type': 'playlist',
             'uploader_id': uploader_id,
```

gilou:
This however seems to break like all the tests?

gilou:
```
$ python3 test/test_download.py TestDownload.test_BandcampAlbum_1 -v 
test_BandcampAlbum_1 (__main__.TestDownload): ... [Bandcamp:album] hierophany-of-the-open-grave: Downloading webpage
[download] Downloading playlist: Nightbringer
[Bandcamp:album] playlist Nightbringer: Collected 9 video ids (downloading 9 of them)
[download] Downloading video 1 of 9
[download] Downloading video 2 of 9
[download] Downloading video 3 of 9
[download] Downloading video 4 of 9
[download] Downloading video 5 of 9
[download] Downloading video 6 of 9
[download] Downloading video 7 of 9
[download] Downloading video 8 of 9
[download] Downloading video 9 of 9
[download] Finished downloading playlist: Nightbringer
FAIL

FAIL: test_BandcampAlbum_1 (__main__.TestDownload):
  File "test/test_download.py", line 178, in test_template
    expect_info_dict(self, res_dict, test_case.get('info_dict', {}))
  File "/home/gilou/Code/tools/youtube-dl/test/helper.py", line 190, in expect_info_dict
  File "/home/gilou/Code/tools/youtube-dl/test/helper.py", line 186, in expect_dict
  File "/home/gilou/Code/tools/youtube-dl/test/helper.py", line 178, in expect_value
    self.assertEqual(
AssertionError: 'Hierophany of the Open Grave' != 'Nightbringer'
- Hierophany of the Open Grave
+ Nightbringer
 : Invalid value for field title, expected 'Hierophany of the Open Grave', got 'Nightbringer'

Ran 1 test in 2.207s

FAILED (failures=1)

```

vberlier:
Hmm probably because in the json data the album name is different. Actually it looks like  `album["name"]` is set to the artist name or something. I didn't really bother looking into it because it worked well enough for my need but I can investigate.

gilou:
It might be good to switch to the data approach used on the track extractor on my other branch, I think it's going to be easier… If someone comforts me in that direction, I'd just switch the PR to it ;)

vberlier:
What branch are you talking about exactly?

gilou:
That one: https://github.com/gilou/youtube-dl/tree/data_approach

hildebro:
any update on this?

iamleot:
@hildebro I have committed a patch based on the discussion we had here in pkgsrc: https://github.com/NetBSD/pkgsrc/blob/8ec948d0453d76b1602d5099b350c49212c385ff/net/youtube-dl/patches/patch-youtube__dl_extractor_bandcamp.py  Both my fork and @gilou's one seems still not accessible at the moment.  I will try to push it ASAP and coordinate with @gilou.

iamleot:
...and probably addressed differently via:

 https://github.com/ytdl-org/youtube-dl/commit/3f1748b9445e9d9367d29221c4b7bf9b88895e4e

iamleot:
JFTR, except for the possible semantic change of `track` all the problems raised in this issued should be solved by https://github.com/ytdl-org/youtube-dl/commit/3f1748b9445e9d9367d29221c4b7bf9b88895e4e so I think this PR and issue https://github.com/ytdl-org/youtube-dl/issues/26681 can be closed.

(Reverting back the old `track` change can be eventually discussed and addressed in a separate PR)

gilou:
#6231 should be closed as well, thanks @iamleot I'll keep an eye on this in the future, but this seems to work nicely for now ;)

PR #26778, opened 2020-10-02
[TubeTuGraz] Add new extractor

ruru4143:

- [ ] Bug fix
- [ ] Improvement
- [X] New extractor
- [ ] New feature

---

Add support for extracting videos from tube.tugraz.at.

### DIFF
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -1163,6 +1163,7 @@ from .tunein import (
     TuneInTopicIE,
     TuneInShortenerIE,
 )
+from .tubetugraz import TubeTuGrazIE
 from .tunepk import TunePkIE
 from .turbo import TurboIE
 from .tv2 import (
--- a/youtube_dl/extractor/tubetugraz.py
+++ b/youtube_dl/extractor/tubetugraz.py
@@ -1,0 +1,59 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+from urllib.error import HTTPError
+
+from .common import InfoExtractor
+
+
+class TubeTuGrazIE(InfoExtractor):
+    _VALID_URL = r'https?://tube.tugraz\.at/paella/ui/watch\.html\?id=(?P<video_id>[0-9a-f\-]+)'
+    _TEST = {
+        'url': 'https://tube.tugraz.at/paella/ui/watch.html?id=85dd2fa9-91ad-4d49-8c85-416fc357a3c3',
+        'md5': '9e6c926c2923ae090ce67c84e880cfbf',
+        'info_dict': {
+            'id': '85dd2fa9-91ad-4d49-8c85-416fc357a3c3',
+            'ext': 'mp4',
+            'title': '#13',
+            'episode': '#13',
+            'creator': 'Fuchs M',
+            'series': '439.002 18W Elektronische Schaltungstechnik 2',
+            'duration': 4790040,
+        }
+    }
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+
+        video_id = mobj.group('video_id')
+
+        information_json = self._download_json("https://tube.tugraz.at/search/episode.json", "information json")['search-results']['result']
+        information = list(filter(lambda video: video["id"] == video_id, information_json))[0]
+
+        m3u8_urls = []
+        for type_ in ("presentation", "presenter"):
+            m3u8_urls.append("https://wowza.tugraz.at/matterhorn_engage/smil:engage-player_%s_%s.smil/playlist.m3u8" % (video_id, type_))
+
+        formats = []
+        for m3u8_url in m3u8_urls:
+            try:
+                self._request_webpage(m3u8_url, video_id)
+                formats.append(self._extract_m3u8_formats(m3u8_url, video_id, ext="mp4")[0])
+            except HTTPError:
+                continue
+        try:
+            series = information['mediapackage']['seriestitle']
+        except KeyError:
+            series = None
+
+        return {
+            'id': video_id,
+            'title': information['mediapackage']['title'],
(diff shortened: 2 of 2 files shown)


Bernd-L:
Looks good

dstftw:
Read coding conventions.

Ferdi265:
I'm just now seeing this as I finished my own TUbe extractor. Mine has support for login (and therefore non-public videos) and downloading whole playlists, so it might make sense to merge some stuff from there.

My extractor is a little bit more hasitly put together though, since it extracts less metadata. Another tidbit is that my extractor uses the direct `.mp4` download instead of the `.m3u8` URL. I don't know which is better, or if it wouldn't be best to support format selection.

My extractor can be found [here](https://github.com/Ferdi265/youtube-dl).

Bernd-L:
My thoughts on this:

Would it be preferable to have a Paella/Opencast extractor, and subclass all other extractors of TU Graz and other universities from that one? Is there a common baseline API spec available?

"Opencast" (and its web-player "Paella") is the video system used by many universities, and it wouldn't make sense to re-implement it for every university.  
The only thing changing (I assume) would be the authentication.

Opencast offers both `mp4` and `m3u8` downloads, supporting both would probably be a good idea.  
Also noteworthy: Opencast has an API which can be queried for all `m3u8` and `mp4` (and other supported formats), so a typical "extraction" may not be necessary in most cases. (Edit: this is a (maybe custom?) API at TU Graz)

You can find more information about TUbe of TU Graz in this talk: <https://replay.teltek.es/video/58ab2c9aa7bc2836008b4571>. It describes the usage of Opencast with the Paella video player at TU Graz. (Other universities may use very similar setups).

FInd more information about Paella on its homepage: <https://paellaplayer.upv.es/features/> (for HTML-based extraction for a fallback if the API is (maybe in the future no longer) available)  
And while [there are REST-API docs for Opencast](https://stable.opencast.org/docs.html?path=/api), they are not very useful, as the API (at least for TU Graz) is not publicly accessible. Instead, they offer their own API, used for searching.

The Paella extractor may be able to handle the websites of many universities, if it's HTML based. And maybe some other institutions offer a limited public API of Opencast directly. May be worth a look.

While there may be many custom APIs, if a user provides a link to a webpage containing a Paella player, we could extract the underlying video source from there.

Ferdi265:
@Bernd-L thank you for these insights!

From my research on tube.tugraz.at up until now, there seems to be very little information in the HTML sent in response to the video URL. Most of the content gets loaded by JavaScript via the (most likely nonstandard) API. 

So while yes, having a generic Paella extractor would be great, it might be that tube.tugraz.at is almost too custom. But I do think it's worth looking into this.

As foe the format extraction: My branch currently queries the API and there definitely is a list of formats in there. The proof of concept version I have now just takes the first format, but this can be easily extended to multiple format choices, with a potential fallback to @ruru4143's m3u8 URL guessing if the API is not accessible (The m3u8 playlist manifest URL only depends on the video id and is often accessible even when the API is not).

Since paella seems to support quite a few formats (full mp4, HLS, HDS, RTMP, DASH, SMOOTH), which of these would be worth extracting into the format field?

- the full mp4 formats sometimes come in multiple bitrates, which could be used for high / low quality selection
- I understand too little of the other streaming formats to know whether they support quality selection, but they only show up once in the format list

tube.tugraz.at (maybe paella in general) also has one other peculiarty: Multiple video streams. For some lectures, the presentation and presenter camera are both used, and all of the format sources are then available twice. The simplest solution would be to default to the presentation if available, and offering the presenter view as a separate format option if both are available. A more complex solution would be to use ffmpeg to render a video with both views, but that is probably too complex (and will also take longer than the average user has time for. Two downloads _and_ ffmpeg rendering?)

Ferdi265:
@Bernd-L: The page with the talk you linked ([that one](https://replay.teltek.es/video/58ab2c9aa7bc2836008b4571) from replay.teltek.es) seems to use a paella player with a completely different API and response JSON structure. So having a common extractor for tugraz and that one does not seem feasible without effectively making a franken-extractor that is effectively two different extractors

Bernd-L:
> @Bernd-L: The page with the talk you linked ([that one](https://replay.teltek.es/video/58ab2c9aa7bc2836008b4571) from replay.teltek.es) seems to use a paella player with a completely different API and response JSON structure. So having a common extractor for tugraz and that one does not seem feasible without effectively making a franken-extractor that is effectively two different extractors

Well, that resolves this idea; but thanks for looking into it!

Ferdi265:
>     * I understand too little of the other streaming formats to know whether they support quality selection, but they only show up once in the format list

Follow-up on this one: some videos also have multiple RTMP streams with varying bitrate, and so far every (?) listed format has a video and audio bitrate field. I've had no success at all downloading from the RTMP streams though, so I'll be skipping those for now.

Ferdi265:
Quick organisational info: I'll be cleaning up my branch to conform to general coding style and add @ruru4143's additional metadata extraction and these new insights, and once that is done I can make a PR for that.

Bernd-L:
@Ferdi265 How would you handle multiple video streams (ie Presenter, Blackboard, Screen)?

The Paella website has [an example with four video streams](https://paellaplayer.upv.es/demos/n-stream-video/) and [another one with three streams](https://paellaplayer.upv.es/demos/three-video/).

Do you want to create separate `mp4` files? Should they be integrated into one file as separate video-tracks? What if only one of those videos contains an audio track with actual audio in it? How would `youtube-dl` find the one to use the audio from? Or should all audio tracks get put into one `mp4` file as separate audio-tracks?

Should an option to split the video streams into several files be provided to the user? And would one audio-track (which one?) be used in both files (assuming the recording only includes the actual microphone from the lecture hall in the presenter recording, but not in the screen capture)?

ruru4143:
closed due to:  #29160

PR #26794, opened 2020-10-03
Removed type ambiguity in one overload of YoutubeDL

VoxLight:
I don't believe there is a difference in functionality with using a set type as a default argument value. But keeping the argument type ambiguous with None isn't necessary.

- [ ] Bug fix
- [x] Improvement (Refactoring)
- [ ] New extractor
- [ ] New feature

---

Minor refactoring with argument type ambiguity.


dstftw:
This is intentional. Mutable default arguments should be avoided in python.

PR #26813, opened 2020-10-05
--batch-file URL enumeration improvements

glenn-slayden:
These four minor improvements apply to reading the list of URLs from the file supplied via the `--batch-file` (`-a`) command line option.

1. ***Exclude blank and empty lines*** in the file from being returned as enumerated URLs. Currently, lines with leading whitespace are only skipped when that whitespace is followed by a comment character `('#', ';', ']')`. This means that each empty line or line consisting only of whitespace currently appears in the list of URLs to process as an extraneous (trimmed) empty string which gets excluded later on.

2. **[bug fix]** Detect and ***correctly remove the Unicode BOM*** when the file descriptor is already decoding Unicode text lines. With Python 3, the `batch_fd` enumerator returns the lines of the file as Unicode. For UTF-8 BOM, this means that the raw BOM bytes from the file `\xef \xbb \xbf` show up converted into a single `\ufeff` character prefixed to the first enumerated text line. 

   This fix solves several buggy interactions between the presence of BOM, the skipping of comments and/or blank lines, and ensuring the list of URLs is consistently trimmed. For example, if the first line of the file is blank, the BOM would be returned as a standalone URL, but if the first line contains a URL, it will be prefixed with this unwanted single `\ufeff` character--which will have therefore inhibited the proper trimming of any leading whitespace. Currently, the `UnicodeBOMIE` helper attempts to recover from such error cases, but that approach is inherently flawed, because clearly it is illogical for a BOM to appear in a (non-batch) URL(s) specified directly on the command line... and for that matter, on any URL *after the first one* in a batch listing, as well. 

   Regardless of any subsequent BOM handling, this PR prevents the error from happening in the first place (at least on Python3). So unless `UnicodeBOMIE` is somehow relevant to Python 2.7 (which I wouldn't know about), I would suggest getting rid of that kludgy extractor entirely.

3. Per section 2.4.3 of [IETF RFC2396](https://www.ietf.org/rfc/rfc2396.txt), the `#` cannot appear in a URL "because it is used to delimit a URI from a fragment identifier in URI references." In other words, it delimits what is interpreted by the browser, so it cannot affect what gets downloaded online in any way. There is no special treatment in the `read_batch_urls` code, but because URLs have that design requirement, it has been previously possible to use the `#` character for end-of-line comments in the batch file. This PR adds ***proper trimming of the `#` character*** into the `read_batch_urls` processing so that the URLs it enumerates are cleaned and trimmed more consistently. For example, the current "accidental" use of `#` interferes with the proper end-trimming of any whitespace that precedes it.

4. Having fixed `read_batch_urls` so that it more consistently enumerates only properly trimmed URLs, it becomes sensible and trivial to do a quick on-the-fly ***elimination of exact duplicates*** so that the same URL is not returned multiple times. Of course this is done without disturbing the original order in which the URLs are listed.

### DIFF
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -3879,16 +3879,23 @@ def escape_url(url):
 
 
 def read_batch_urls(batch_fd):
+    seen = set()
     def fixup(url):
         if not isinstance(url, compat_str):
             url = url.decode('utf-8', 'replace')
         BOM_UTF8 = '\xef\xbb\xbf'
         if url.startswith(BOM_UTF8):
             url = url[len(BOM_UTF8):]
-        url = url.strip()
-        if url.startswith(('#', ';', ']')):
-            return False
-        return url
+        if url:
+            if url[0] == '\ufeff':
+                url = url[1:]
+            url = url.lstrip()
+            if url and not url[0] in ('#', ';', ']'):
+                url = url.split('#', 1)[0].rstrip()
+                if not url in seen:
+                    seen.add(url)
+                    return url
+        return False
 
     with contextlib.closing(batch_fd) as fd:
         return [url for url in map(fixup, fd) if url]


dstftw:
1. Blank and whitespace lines are excluded just fine. Provide concrete examples that does not work, add tests.
2. Add tests.
3. `#` is an indispensable part of URI and must not be touched. [Lots](https://github.com/ytdl-org/youtube-dl/blob/4eda10499e8db831167062b0e0dbc7d10d34c1f9/youtube_dl/extractor/bbc.py#L41) [of](https://github.com/ytdl-org/youtube-dl/blob/4eda10499e8db831167062b0e0dbc7d10d34c1f9/youtube_dl/extractor/appletrailers.py#L237) [extractors](https://github.com/ytdl-org/youtube-dl/blob/4eda10499e8db831167062b0e0dbc7d10d34c1f9/youtube_dl/extractor/kanalplay.py#L16) [already](https://github.com/ytdl-org/youtube-dl/blob/4eda10499e8db831167062b0e0dbc7d10d34c1f9/youtube_dl/extractor/nrk.py#L255) [use](https://github.com/ytdl-org/youtube-dl/blob/4eda10499e8db831167062b0e0dbc7d10d34c1f9/youtube_dl/extractor/rte.py#L134) [it](https://github.com/ytdl-org/youtube-dl/blob/4eda10499e8db831167062b0e0dbc7d10d34c1f9/youtube_dl/extractor/mailru.py#L25).
4. Duplicate entries may be intentional.

PR #26851, opened 2020-10-09
Don't attempt to coerce to JS string to numbers

kevinoconnor7:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

The current logic in `js_to_json` tries to rewrite octal/hex numbers to decimal. However, when the logic actually happens the `"` or `'` have already been trimmed off. This causes what were originally strings, that happen to look like octal/hex numbers, to get rewritten to decimal and returned as a number rather than a string.

In practive something like:

```js
{
  "0x40": "foo",
  "040": "bar",
}
```

would get rewritten as:

```json
{
  64: "foo",
  32: "bar
}
```

This is problematic since this isn't valid JSON as you cannot have non-string keys.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -994,6 +994,12 @@ class TestUtil(unittest.TestCase):
         on = js_to_json('{42:4.2e1}')
         self.assertEqual(json.loads(on), {'42': 42.0})
 
+        on = js_to_json('{ "0x40": "0x40" }')
+        self.assertEqual(json.loads(on), {'0x40': '0x40'})
+
+        on = js_to_json('{ "040": "040" }')
+        self.assertEqual(json.loads(on), {'040': '040'})
+
     def test_js_to_json_malformed(self):
         self.assertEqual(js_to_json('42a1'), '42"a1"')
         self.assertEqual(js_to_json('42a-1'), '42"a"-1')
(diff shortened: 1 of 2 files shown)


test/test_utils.py
```diff
@@ -994,6 +994,12 @@ def test_js_to_json_edgecases(self):
         on = js_to_json('{42:4.2e1}')
         self.assertEqual(json.loads(on), {'42': 42.0})
 
+        on = js_to_json('{ "0x40": true }')
```
dstftw, line 997: `true` is irrelevant here, keep tests minimal.

test/test_utils.py
```diff
@@ -994,6 +994,12 @@ def test_js_to_json_edgecases(self):
...
+        self.assertEqual(json.loads(on), {'0x40': True})
+
+        on = js_to_json('{ "040": true }')
```
dstftw, line 1000: Ditto.

dstftw (changes requested):

test/test_utils.py
kevinoconnor7: Do you have a preferred value? I don't think `{ "0x40": }` would be a meaningful test case since it's neither valid JS nor JSON.

dstftw: You can either use an array `["0x40"]` or a dict with both key and value set to `"0x40"`.

kevinoconnor7:
Fixes applied.

PR #26946, opened 2020-10-20
[gedi] Add new extractor

### LINKED ISSUE #7347: support request for video.repubblica.it
youtube-dl -v http://video.repubblica.it/spettacoli-e-cultura/pasolini-maestro-corsaro/216709/215893?ref=HREC1-10
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://video.repubblica.it/spettacoli-e-cultura/pasolini-maestro-corsaro/216709/215893?ref=HREC1-10']
[debug] youtube-dl version 2015.11.02
[debug] Python version 2.7.10 - Darwin-12.6.0-x86_64-i386-64bit
[debug] exe versions: avconv 11.4, avprobe 11.4, ffmpeg 2.8.1, ffprobe 2.8.1, rtmpdump 2.4
[debug] Proxy map: {}
[generic] 215893?ref=HREC1-10: Requesting header
[generic] 215893?ref=HREC1-10: Downloading webpage
[generic] 215893?ref=HREC1-10: Extracting information
ERROR: Unsupported URL: http://video.repubblica.it/spettacoli-e-cultura/pasolini-maestro-corsaro/216709/215893?ref=HREC1-10
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/generic.py", line 1241, in _real_extract
    doc = compat_etree_fromstring(webpage.encode('utf-8'))
  File "/usr/local/bin/youtube-dl/youtube_dl/compat.py", line 248, in compat_etree_fromstring
    doc = _XML(text, parser=etree.XMLParser(target=etree.TreeBuilder(element_factory=_element_factory)))
  File "/usr/local/bin/youtube-dl/youtube_dl/compat.py", line 237, in _XML
    parser.feed(text)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/xml/etree/ElementTree.py", line 1642, in feed
    self._raiseerror(v)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/xml/etree/ElementTree.py", line 1506, in _raiseerror
    raise err
ParseError: syntax error: line 1, column 0
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 661, in extract_info
    ie_result = ie.extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 290, in extract
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/generic.py", line 1839, in _real_extract
    raise UnsupportedError(url)
UnsupportedError: Unsupported URL: http://video.repubblica.it/spettacoli-e-cultura/pasolini-maestro-corsaro/216709/215893?ref=HREC1-10

nixxo:

- [x] New extractor

---

A new extractor for the various Gedi group italian newspaper websites like repubblica.it and lastampa.it among the biggest.

It closes #7347 (and #20835)

### DIFF
--- a/youtube_dl/extractor/gedidigital.py
+++ b/youtube_dl/extractor/gedidigital.py
@@ -1,0 +1,199 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+from var_dump import var_dump
+
+from .common import InfoExtractor
+
+
+class GediDigitalBaseIE(InfoExtractor):
+    @staticmethod
+    def _clean_audio_fmts(formats):
+        unique_formats = []
+        for f in formats:
+            if 'acodec' in f:
+                unique_formats.append(f)
+        formats[:] = unique_formats
+
+    def _real_extract(self, url):
+        u = re.match(self._VALID_URL, url)
+        self.IE_NAME = u.group('iename') if u.group('iename') else 'gedi'
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, video_id)
+        player_data = re.findall(
+            r'PlayerFactory\.setParam\(\'(?P<type>.+?)\',\s*\'(?P<name>.+?)\',\s*\'(?P<val>.+?)\'\);',
+            webpage)
+
+        formats = []
+        audio_fmts = []
+        hls_fmts = []
+        http_fmts = []
+        title = ''
+        thumb = ''
+
+        fmt_reg = r'(?P<t>video|audio)-(?P<p>rrtv|hls)-(?P<h>[\w\d]+)(?:-(?P<br>[\w\d]+))?$'
+        br_reg = r'video-rrtv-(?P<br>\d+)-'
+        for t, n, v in player_data:
+            if t == 'format':
+                m = re.match(fmt_reg, n)
+                if m:
+                    # audio formats
+                    if m.group('t') == 'audio':
+                        if m.group('p') == 'hls':
+                            audio_fmts.extend(self._extract_m3u8_formats(
+                                v, video_id, 'm4a', m3u8_id='hls', fatal=False))
+                        elif m.group('p') == 'rrtv':
+                            audio_fmts.append({
+                                'format_id': 'mp3',
+                                'url': v,
+                                'tbr': 128,
+                                'ext': 'mp3',
+                                'vcodec': 'none',
+                                'acodec': 'mp3',
+                            })
+
+                    # video formats
+                    elif m.group('t') == 'video':
+                        # hls manifest video
+                        if m.group('p') == 'hls':
+                            hls_fmts.extend(self._extract_m3u8_formats(
+                                v, video_id, 'mp4', m3u8_id='hls', fatal=False))
+                        # direct mp4 video
+                        elif m.group('p') == 'rrtv':
+                            if not m.group('br'):
+                                mm = re.search(br_reg, v)
+                            http_fmts.append({
+                                'format_id': 'https-' + m.group('h'),
+                                'protocol': 'https',
+                                'url': v,
+                                'tbr': int(m.group('br')) if m.group('br') else
+                                (int(mm.group('br')) if mm.group('br') else 0),
+                                'height': int(m.group('h'))
+                            })
+
+            elif t == 'param':
+                if n == 'videotitle':
+                    title = v
+                if n == 'image_full_play':
+                    thumb = v
+
+        title = re.sub(r'\s*-\s*La Stampa', '', self._og_search_title(webpage)) if title == '' else title
+
+        if audio_fmts:
+            self._clean_audio_fmts(audio_fmts)
+            self._sort_formats(audio_fmts)
+        if hls_fmts:
+            self._sort_formats(hls_fmts)
+        if http_fmts:
+            self._sort_formats(http_fmts)
+
+        formats.extend(audio_fmts)
+        formats.extend(hls_fmts)
+        formats.extend(http_fmts)
+
+        return {
+            'id': video_id,
+            'title': title,
+            'description': self._html_search_meta('twitter:description', webpage),
+            'thumbnail': thumb,
+            'formats': formats,
+        }
+
+
+class GediDigitalIE(GediDigitalBaseIE):
+    IE_NAME = ''
+    _VALID_URL = r'''(?x)https://video\.
+                    (?P<iename>
+                        (?:espresso\.)?repubblica
+                        |lastampa
+                        |ilsecoloxix
+                        |iltirreno
+                        |messaggeroveneto
+                        |ilpiccolo
+                        |gazzettadimantova
+                        |mattinopadova
+                        |laprovinciapavese
+                        |tribunatreviso
+                        |nuovavenezia
+                        |gazzettadimodena
+                        |lanuovaferrara
+                        |corrierealpi
+                        |lasentinella
+                    )
+                    (?:\.gelocal)?\.it/.+?/(?P<id>[\d/]+)$'''
+    _TESTS = [{
+        'url': 'https://video.lastampa.it/politica/il-paradosso-delle-regionali-la-lega-vince-ma-sembra-aver-perso/121559/121683',
+        'md5': '84658d7fb9e55a6e57ecc77b73137494',
+        'info_dict': {
+            'id': '121559/121683',
+            'ext': 'mp4',
+            'title': 'Il paradosso delle Regionali: ecco perché la Lega vince ma sembra aver perso',
+            'description': 'md5:de7f4d6eaaaf36c153b599b10f8ce7ca',
+            'thumbnail': 'https://www.repstatic.it/video/photo/2020/09/22/731397/731397-thumb-social-play.jpg',
+        },
+    }, {
+        'url': 'https://video.repubblica.it/motori/record-della-pista-a-spa-francorchamps-la-pagani-huayra-roadster-bc-stupisce/367415/367963',
+        'md5': 'e763b94b7920799a0e0e23ffefa2d157',
+        'info_dict': {
+            'id': '367415/367963',
+            'ext': 'mp4',
+            'title': 'Record della pista a Spa Francorchamps, la Pagani Huayra Roadster BC stupisce',
+            'description': 'md5:5deb503cefe734a3eb3f07ed74303920',
+            'thumbnail': 'https://www.repstatic.it/video/photo/2020/09/19/730799/730799-thumb-social-play.jpg',
+        },
+    }, {
+        'url': 'https://video.ilsecoloxix.it/sport/cassani-e-i-brividi-azzurri-ai-mondiali-di-imola-qui-mi-sono-innamorato-del-ciclismo-da-ragazzino-incredibile-tornarci-da-ct/66184/66267',
+        'md5': 'e48108e97b1af137d22a8469f2019057',
+        'info_dict': {
+            'id': '66184/66267',
+            'ext': 'mp4',
+            'title': 'Cassani e i brividi azzurri ai Mondiali di Imola: \\"Qui mi sono innamorato del ciclismo da ragazzino, incredibile tornarci da ct\\"',
+            'description': 'md5:fc9c50894f70a2469bb9b54d3d0a3d3b',
+            'thumbnail': 'https://www.repstatic.it/video/photo/2020/09/23/731520/731520-thumb-social-play.jpg',
+        },
+    }, {
+        'url': 'https://video.iltirreno.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/141059/142723',
+        'md5': 'a6e39f3bdc1842bbd92abbbbef230817',
+        'info_dict': {
+            'id': '141059/142723',
+            'ext': 'mp4',
+            'title': 'Dentro la notizia - Ferrari, cosa succede a Maranello',
+            'description': 'md5:9907d65b53765681fa3a0b3122617c1f',
+            'thumbnail': 'https://www.repstatic.it/video/photo/2020/09/23/731504/731504-thumb-social-play.jpg',
+        },
+    }, {
+        'url': 'https://video.messaggeroveneto.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }, {
+        'url': 'https://video.ilpiccolo.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }, {
+        'url': 'https://video.gazzettadimantova.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }, {
+        'url': 'https://video.mattinopadova.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }, {
+        'url': 'https://video.laprovinciapavese.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }, {
+        'url': 'https://video.tribunatreviso.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }, {
+        'url': 'https://video.nuovavenezia.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }, {
+        'url': 'https://video.gazzettadimodena.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }, {
+        'url': 'https://video.lanuovaferrara.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }, {
+        'url': 'https://video.corrierealpi.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }, {
+        'url': 'https://video.lasentinella.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }]
--- a/youtube_dl/extractor/gedidigital.py
+++ b/youtube_dl/extractor/gedidigital.py
@@ -108,7 +108,7 @@ class GediDigitalBaseIE(InfoExtractor):
 
 class GediDigitalIE(GediDigitalBaseIE):
     IE_NAME = ''
-    _VALID_URL = r'''(?x)https://video\.
+    _VALID_URL = r'''(?x)https?://video\.
                     (?P<iename>
                         (?:espresso\.)?repubblica
                         |lastampa
--- a/youtube_dl/extractor/gedidigital.py
+++ b/youtube_dl/extractor/gedidigital.py
@@ -2,9 +2,9 @@
 from __future__ import unicode_literals
 
 import re
-from var_dump import var_dump
 
 from .common import InfoExtractor
+from ..compat import compat_str
 
 
 class GediDigitalBaseIE(InfoExtractor):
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,180 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from ..compat import compat_str
+
+
+class GediDigitalBaseIE(InfoExtractor):
```
remitamine, line 10: what's the point of creating a base class here?

nixxo: because Gedi is a conglomerate media group that is buying various newspaper/wensites in italy and has various different websites so when I started writing the extractor I though that the various websites would require some sort of dedicated extractor for each different website.
While developing I saw that the situation was much easier and clean than expected and (as you can see in the GediDigitalIE extractor with the TESTS and the VALID_URL regex) only the Base Extractor basically was necessary without dedicated extractor for each website of the Gedi media group.

I kinda left the two classes sepatate because I thought about the possibility of a future aquisition by GediGroup of a website/newspaper that requires a dedicated Extractor, or the possibility for some other extractor to use the GediDigitalBase extractor if necessary.

remitamine: for now, merge with the main extractor and if a need for separation arises in the future, this can be done(only the common code).

nixxo: ok, merged the two classes.

and while I was there I added the possibility to generate http direct mp4 urls that I forgot to commit prior.

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,243 @@
...
+                        v, video_id, 'm4a', m3u8_id='audio-hls', fatal=False))
+                if n == 'audio-rrtv-mp3':
+                    audio_fmts.append({
+                        'format_id': 'mp3',
+                        'url': v,
+                        'tbr': 128,
+                        'ext': 'mp3',
+                        'vcodec': 'none',
+                        'acodec': 'mp3',
+                    })
```
remitamine, line 217: - extract all formats.
- try not complicating things, try to avoid using hardcoded values.

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,243 @@
...
+            'description': 'md5:5deb503cefe734a3eb3f07ed74303920',
+            'thumbnail': r're:^https://www\.repstatic\.it/video/photo/.+?-thumb-social-play\.jpg$',
+        },
+    }, {
+        'url': 'https://video.ilsecoloxix.it/sport/cassani-e-i-brividi-azzurri-ai-mondiali-di-imola-qui-mi-sono-innamorato-del-ciclismo-da-ragazzino-incredibile-tornarci-da-ct/66184/66267',
+        'md5': 'e48108e97b1af137d22a8469f2019057',
+        'info_dict': {
+            'id': '66184/66267',
+            'ext': 'mp4',
+            'title': 'Cassani e i brividi azzurri ai Mondiali di Imola: \\"Qui mi sono innamorato del ciclismo da ragazzino, incredibile tornarci da ct\\"',
```
remitamine, line 58: the title should'nt have an escaped double quotes.

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,243 @@
...
+    }]
+
+    @staticmethod
+    def _clean_audio_fmts(formats):
+        # remove duplicates audio formats
+        unique_formats = []
+        for f in formats:
+            if 'acodec' in f:
+                unique_formats.append(f)
+        formats[:] = unique_formats
```
remitamine, line 124: i'm not sure why you're adding this step.

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,243 @@
...
+            elif t == 'param':
+                if n == 'videotitle':
+                    title = v
+                if n in ['image_full_play', 'image_full', 'image']:
+                    thumb = v
+
+        title = self._og_search_title(webpage) if not title else title
+
+        # clean weird char
+        title = compat_str(title).encode('utf8', 'replace').replace(b'\xc3\x82', b'').decode('utf8', 'replace')
```
remitamine, line 227: do you have an example for this case?

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,243 @@
...
+
+        if mp4:
+            formats.extend(self._generate_http_urls(mp4, formats) or [])
+
+        self._sort_formats(formats)
+
+        return {
+            'id': video_id,
+            'title': title,
+            'description': self._html_search_meta('twitter:description', webpage),
```
remitamine, line 240: add more fallbacks and extract `timestamp`.

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,243 @@
...
+                'height': h or _QUALITY[q][1],
+                'tbr': t or int(q),
+                'protocol': 'https',
+                'format_id': 'https-%s' % q,
+            })
+        return http_formats
+
+    def _real_extract(self, url):
+        u = re.match(self._VALID_URL, url)
+        self.IE_NAME = u.group('iename') if u.group('iename') else 'gedi'
```
remitamine, line 184: no.

remitamine (changes requested):

youtube_dl/extractor/gedidigital.py
nixxo: the last test has this problem.

https://video.espresso.repubblica.it/embed/tutti-i-video/01-ted-villa/14772/14870

if you open in the webrowser, check the source and look for 'videotitle' you will se the error

the title in the page is ok but the title present in the video player parameter is wrongly encoded by the website so it needs to be cleaned up.

remitamine: i think the use of `videotitle` is not reliable(it also has a problem with escaped double quotes), instead of this cleanup, other sources should be prefered.

youtube_dl/extractor/gedidigital.py
nixxo: _extract_m3u8_formats returns for some reason multiple audio url that are in fact the same. so I remove the duplicates

without cleaning 
```
format code                       extension  resolution note
audio-hls-1@48000-64000-Audio_64  m4a        audio only
audio-hls-1@48000-80000-Audio_80  m4a        audio only
audio-hls-2@48000-96000-Audio_96  m4a        audio only
audio-hls-100                     m4a        audio only  100k , mp4a.40.2
audio-hls-116                     m4a        audio only  116k , mp4a.40.2
```

cleaned
```
format code                       extension  resolution note
audio-hls-100                     m4a        audio only  100k , mp4a.40.2
audio-hls-116                     m4a        audio only  116k , mp4a.40.2
```

youtube_dl/extractor/gedidigital.py
nixxo: ok

youtube_dl/extractor/gedidigital.py
remitamine: they have diffirent bitrate.

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,243 @@
...
+        video_id = self._match_id(url)
+
+        webpage = self._download_webpage(url, video_id)
+        player_data = re.findall(
+            r'PlayerFactory\.setParam\(\'(?P<type>.+?)\',\s*\'(?P<name>.+?)\',\s*\'(?P<val>.+?)\'\);',
```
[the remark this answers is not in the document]
nixxo: ok, reduced from 40+ to ~14 elements matched

youtube_dl/extractor/gedidigital.py
nixxo: I think if I remember correctly why i went with `videotitle` was because not always the title was present in the page (some ebeds that I tested ad it missing or something) and instead `videotitle` was always present because is uno of the player parameter (but it need to be cleaned because is messed up a bit)

I'll collect some urls to do some testing and check better.

youtube_dl/extractor/gedidigital.py
nixxo: 64000 is identical to hls-100
80000 is identical to hls-116

the only one different is 96000.

I'll improve the function.

remitamine: i think you should not remove the format, it's possible that they would be served from different server, protocol, etc...

youtube_dl/extractor/gedidigital.py
nixxo: done

youtube_dl/extractor/gedidigital.py
nixxo: done

youtube_dl/extractor/gedidigital.py
nixxo: ok, Ill keep everything. The format list will be long and a bit bloated but I understand your resoning.

youtube_dl/extractor/gedidigital.py
nixxo: fixed

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,170 @@
...
+    }, {
+        'url': 'https://video.lanuovaferrara.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }, {
+        'url': 'https://video.corrierealpi.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }, {
+        'url': 'https://video.lasentinella.gelocal.it/sport/dentro-la-notizia-ferrari-cosa-succede-a-maranello/133362/134466',
+        'only_matching': True,
+    }]
```
remitamine, line 123: use valid URLs.

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,170 @@
...
+                        v, video_id, 'm4a', m3u8_id='audio-hls', fatal=False))
+                if n == 'audio-rrtv-mp3':
+                    formats.append({
+                        'format_id': 'audio-mp3',
+                        'url': v,
+                        'tbr': 128,
+                        'ext': 'mp3',
+                        'vcodec': 'none',
+                        'acodec': 'mp3',
+                    })
```
remitamine, line 152: again, extract all formats, there is no guarentee that those formats are available in all videos.

remitamine (changes requested):

youtube_dl/extractor/gedidigital.py
nixxo: I have to rectify what I said in the previos comment as I checked again.

the function _extract_m3u8_formats extracts 6 audio formats but they are in reality 3 duplicate.

Like said before:
64000 is identical to hls-100
80000 is identical to hls-116
and 96000 is identical to hls-133

here is the m3u8 manifest:
```
#EXTM3U
#EXT-X-MEDIA:TYPE=AUDIO,URI="_64/index.m3u8",GROUP-ID="1@48000-64000",NAME="Audio 64",DEFAULT=NO,AUTOSELECT=NO
#EXT-X-MEDIA:TYPE=AUDIO,URI="_80/index.m3u8",GROUP-ID="1@48000-80000",NAME="Audio 80",DEFAULT=NO,AUTOSELECT=NO
#EXT-X-MEDIA:TYPE=AUDIO,URI="_96/index.m3u8",GROUP-ID="2@48000-96000",NAME="Audio 96",DEFAULT=NO,AUTOSELECT=NO
#EXT-X-STREAM-INF:BANDWIDTH=133336,CODECS="mp4a.40.2",AUDIO="2@48000-96000"
_96/index.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=100641,CODECS="mp4a.40.2",AUDIO="1@48000-64000"
_64/index.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=116989,CODECS="mp4a.40.2",AUDIO="1@48000-80000"
_80/index.m3u8
```
as you can see there are only 3 formats but the _extract_m3u8_formats outputs 6 formats for some reason.

youtube_dl/extractor/gedidigital.py
nixxo: ok

youtube_dl/extractor/gedidigital.py
nixxo: `video-hls-vod-ak` is always present and from that with _extarct_akamai it creates formats for hls, hds and http/https;
I think it's a cleaner solution that to add stuff from the website (like in some page there is also a youtube link of the same video as `video-youtube-pfp`) so I didn't want to pollute the format list with duplicates etc.

remitamine: as i said there is no guarantee that this will be the case all the time.

youtube_dl/extractor/gedidigital.py
remitamine: then you should use `_remove_duplicate_formats` method.

nixxo: I already tried it before writing my own method (which is a modified version of that one)

The "problem" of `_remove_duplicate_formats` is that it cheched the formats in order and therefore eliminated the "good" formats and kept the "bad" ones.
By "good" I mean:
```
audio-hls-100                     m4a        audio only  100k , mp4a.40.2
```
with a clean `name`, `bitrate` and `audio codec` and not:
```
audio-hls-1@48000-64000-Audio_64  m4a        audio only
```
with a rubbish name and without bitrate and audio codec.

Because they are listed after they get eliminated... so I written a coustom method to clean the formats in a way that I liked more.

remitamine: but it would always remove formats without `acodec` even if they are only present as `EXT-X-MEDIA` formats.

nixxo: ok...

btw, I've already removed it ia couple of commits ago https://github.com/ytdl-org/youtube-dl/pull/26946/commits/8f564a71d1669eaee696fc379eab52fbb712ac30

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,194 @@
...
+        'md5': '0391c2c83c6506581003aaf0255889c0',
+        'info_dict': {
+            'id': '14772/14870',
+            'ext': 'mp4',
+            'title': 'Festival EMERGENCY, Villa: «La buona informazione aiuta la salute»',
+            'description': 'md5:2bce954d278248f3c950be355b7c2226',
+            'thumbnail': r're:^https://www\.repstatic\.it/video/photo/.+?-thumb-social-play\.jpg$',
+            'timestamp': 1602159940,
+            'upload_date': '20201008',
+        },
```
remitamine, line 92: add a comment on why those tests have been extended(what's been test that is not covered by the first test case), if the reason is just that they have different domain/URL formats then mark them as `only_matching` tests.

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,194 @@
...
+                if n == 'audio-rrtv-mp3':
+                    formats.append({
+                        'format_id': 'audio-mp3',
+                        'url': v,
+                        'tbr': 128,
+                        'ext': 'mp3',
+                        'vcodec': 'none',
+                        'acodec': 'mp3',
+                    })
+            elif t == 'param' and n in ['image_full_play', 'image_full', 'image']:
```
remitamine, line 178: don't use the image that has the play icon(`image_full_play`).

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,194 @@
...
+        self._sort_formats(formats)
+
+        return {
+            'id': video_id,
+            'title': self._html_search_meta('twitter:title', webpage) or self._og_search_title(webpage),
+            'description': self._html_search_meta(
+                ['twitter:description', 'og:description', 'description'],
+                webpage, default=None),
+            'timestamp': parse_iso8601(self._og_search_property(
+                ['published_time', 'modified_time'],
```
remitamine, line 190: don't use `modified_time` as publish timestamp.

remitamine (changes requested):

youtube_dl/extractor/gedidigital.py
nixxo: i used it while writing the extarctor because they are different websites and kept it to see if the modification to the extractor would work on the other wenbsite as well.

Turns out the websites were a lot more similar then expected and so these test are not necessary, you're right. I'll make them only_matching.

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,158 @@
...
+            'ext': 'mp4',
+            'title': 'Il paradosso delle Regionali: ecco perché la Lega vince ma sembra aver perso',
+            'description': 'md5:de7f4d6eaaaf36c153b599b10f8ce7ca',
+            'thumbnail': r're:^https://www\.repstatic\.it/video/photo/.+?-thumb-full-.+?\.jpg$',
+            'timestamp': 1600788168,
+            'upload_date': '20200922',
+        },
+    }, {
+        'url': 'https://video.repubblica.it/motori/record-della-pista-a-spa-francorchamps-la-pagani-huayra-roadster-bc-stupisce/367415/367963',
+        'md5': 'c75ba5637a3c375a1b09062d7a7bd305',
```
remitamine, line 47: should be removed.

youtube_dl/extractor/gedidigital.py
```diff
@@ -0,0 +1,158 @@
...
+        self._sort_formats(formats)
+
+        return {
+            'id': video_id,
+            'title': self._html_search_meta('twitter:title', webpage) or self._og_search_title(webpage),
+            'description': self._html_search_meta(
+                ['twitter:description', 'og:description', 'description'],
+                webpage, default=None),
+            'timestamp': parse_iso8601(self._og_search_property(
+                'published_time', webpage, default='').strip()),
```
remitamine, line 155: - is it common that videos would not have `og:published_time`?
- do you have an example where the `strip` method call is needed?

nixxo: <meta property="og:published_time" content="2020-10-08T12:25:40 "/>
<meta property="og:modified_time" content="2020-10-08T12:25:40+02:00"/>

as you can see when the timezone is not set (not in every page) there is a space at the end and it makes the iso6601 method fail.

It's in one of the test that I put to only_matching. Should I restore it and comment it explaining the .split() thing?

youtube_dl/extractor/gedidigital.py
nixxo: ok

youtube_dl/extractor/gedidigital.py
remitamine: just add a comment above the related `only_matching` test case.

> is it common that videos would not have `og:published_time`?

nixxo: no, I guess it'a always there.

remitamine: then I think it would be better to use `fatal=False` instead of `default=None`.

nixxo: ok

remitamine: looking again at this, there are multiple problems with the `og:published_time`, the timezone is important to calculate the correct timestamp so the `og:published_time` value for embeds is malformed, and also there is a discrepancy between the values from the video page and the values from the embed page, so unless there is a way to determine the correct value, it might be better to drop it.

nixxo: well, if I simply remove the .strip() then if the value is malformed the timestamp is set to None.

Wouldn't that be ok? Set the timestamp only if available and only if not malformad.

remitamine: in the case of https://video.espresso.repubblica.it/tutti-i-video/01-ted-villa/14772/14870
the date extracted from `og:published_time` is `2020-10-15`.
while the one that is shown on the webpage `08 ottobre 2020`(i think this the right date).
so it appears that `og:published_time` is also unreliable similar to `videotitle`.

nixxo: well, even in the player setup script there are two dates:
```
"uploadDate": "2020-10-15T18:08:04+02:00",
"contentUrl": "https://media.gedidigital.it/repubblicatv/file/2020/10/08/.......
```

so I guess I'll remove it at that's it. right?

remitamine: > so I guess I'll remove it at that's it. right?

yes, it would be better to remove it.

nixxo:
@remitamine remember the looong conversation ;-) we had last month about the vvvvid extractor "season" title? ;-)

In the end I think the solution was to set the `playlist_title` tag as the more appropriate tag for the value.

When you have time can you give a look at it here #27618 so we can close the vvvvid extractor for good.

PR #27037, opened 2020-11-17
[iheartradio:podcast] Add new extractor

gardenappl:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

(continued from #26394)

Added support for podcasts from https://iheart.com/podcast. Supports individual episodes, as well as entire series of podcasts, using their internal API.

Note that IHeartRadio offers other services besides podcasts, but those seem to be only available in the US and/or Canada. That includes things like music playlists, and actual radio stations. I did not include support for them because these are not available where I live, and frankly, I really don't care about those features.

### DIFF
--- a/youtube_dl/extractor/iheart.py
+++ b/youtube_dl/extractor/iheart.py
@@ -35,7 +35,7 @@ class IHeartRadioIE(IHeartRadioBaseIE):
             'id': '70346499',
             'ext': 'mp3',
             'title': 'Part One: Alexander Lukashenko: The Dictator of Belarus',
-            'description': 'md5:66480b2d25ec93a5f60c0faa3275ce5c',
+            'description': 'md5:96cc7297b3a5a9ebae28643801c96fae',
             'timestamp': 1597741200,
             'upload_date': '20200818',
         }


remitamine:
you should split the extractor into:
- one that handles individual episodes 
- the other would handle a whole podcast.

gardenappl:
@remitamine Done, also made a few fixes.

gardenappl:
Made commit message a bit shorter.

remitamine:
still the episode and podcast code is mixed up, the code that epecific that specific and only related to podcasts should not be put in the base extractor, the same for the episode extractor.

gardenappl:
This last commit should be good.

youtube_dl/extractor/iheartradio.py
```diff
@@ -0,0 +1,213 @@
...
+
+from numbers import Number
+
+from ..utils import (
+    clean_html,
+    urlencode_postdata
+)
+
+
+# To get the audio files, we have to use their internal API
```
remitamine, line 20: no need to add comments that can be easily deduced from the code.

youtube_dl/extractor/iheartradio.py
```diff
@@ -0,0 +1,213 @@
...
+                'url': 'https://i.iheart.com/v3/catalog/podcast/%(id)s?ops=fit(%(size)d, %(size)d)' % {
+                    'id': podcast_id, 'size': size},
+                'width': size,
+                'height': size
+            })
+
+        # Release date timestamp is in milliseconds
+        release_date = content_info.get('startDate')
+        if isinstance(release_date, Number) and release_date > 2000000000:
+            release_date /= 1000
```
remitamine, line 85: use `int_or_none` function.

youtube_dl/extractor/iheartradio.py
```diff
@@ -0,0 +1,213 @@
...
+
+        # Remove analytics from stream URL (optional)
+        streamUrl = item_info['streamUrl']
+        streamUrl = re.sub(r'(?:www\.)?podtrac\.com/pts/redirect\.[\w]*/', '', streamUrl)
+        streamUrl = re.sub(r'chtbl\.com/track/[\w]*/', '', streamUrl)
+        streamUrl = re.sub(r'\?source=[\w]*', '', streamUrl)
```
remitamine, line 91: this should done separatly from this PR(it's used in multiple services that are supported by youtube-dl, so it should be done in generic way).

remitamine (changes requested):

youtube_dl/extractor/iheartradio.py
```diff
@@ -0,0 +1,213 @@
...
+            'deviceName': "web-desktop",
+            'host': "webapp.WW",
+            'oauthUuid': random_oauth_id,
+            'userName': 'anon' + random_oauth_id
+        })
+        return self._download_json(
+            'https://ww.api.iheart.com/api/v1/account/loginOrCreateOauthUser',
+            current_id, "Registering temporary user", data=register_user_values,
+            headers={'Accept': 'application/json, text/plain, */*',
+                     'X-hostName': 'webapp.WW'})
```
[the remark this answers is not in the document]
gardenappl: I'm not sure how I would do that, I was just looking at the network requests done by the web frontend and copying that behavior.

Also, what are we trying to avoid? Other API calls require the output from this.

youtube_dl/extractor/iheartradio.py
gardenappl: I couldn't find any other mentions of "podtrac" or "chbtl" in other extractors. But yeah, I guess I shouldn't be doing that in this extractor.

youtube_dl/extractor/iheartradio.py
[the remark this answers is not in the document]
remitamine: > I'm not sure how I would do that, I was just looking at the network requests done by the web frontend and copying that behavior.
> 
> Also, what are we trying to avoid? Other API calls require the output from this.

trying to avoid additional requests, I think to move quickly with this PR, I would make a simpler implementation, and then you can base your modification on it.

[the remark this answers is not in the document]
remitamine: the extractor that relies on API v3 has been added in 9c484c0019ba16cd4b4d686930bfa038f1bf9320.

gardenappl:
@remitamine Thanks for co-operating.
You said I could base my modifications on your commits, but your code seems perfectly fine and I don't think I can add much. The only thing I did was fix the test for the episode extractor. It was expecting the description to contain HTML tags, but that's cleaned now.

PR #27081, opened 2020-11-18
[youtube] Fix error reason extraction

dpeukert:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

In its current state, the YouTube error reason extraction logic only checks `player_response['playabilityStatus']['reason']`. Recently though, the actual reason started only getting sent as an array in `player_response['playabilityStatus']['errorScreen']['playerErrorMessageRenderer']['subreason']['runs']`. This PR changes the behaviour to check the array first and use the `reason` property as a fallback.

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -2006,6 +2006,21 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
             else:
                 error_message = extract_unavailable_message()
                 if not error_message:
+                    reason_list = try_get(
+                        player_response,
+                        lambda x: x['playabilityStatus']['errorScreen']['playerErrorMessageRenderer']['subreason']['runs'],
+                        list) or []
+                    for reason in reason_list:
+                        if not isinstance(reason, dict):
+                            continue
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/youtube.py
```diff
@@ -2005,6 +2005,21 @@ def _extract_filesize(media_url):
...
+                    reason_list = try_get(
+                        player_response,
+                        lambda x: x['playabilityStatus']['errorScreen']['playerErrorMessageRenderer']['subreason']['runs'],
+                        list) or []
+                    for reason in reason_list:
+                        if not isinstance(reason, dict):
+                            continue
+                        reason_text = try_get(reason, lambda x: x['text'], compat_str)
+                        if reason_text:
+                            if not error_message:
```
SuperSandro2000, line 2018:
```suggestion
if reason_text and not error_message:
```

dpeukert: That wouldn't work at all, those two conditions are separate for a reason, you can't use += to add to None, therefore the first condition sets the message to an empty string so that we can add to it. Your suggestion would also stop working after the first part of the reason, because the message would no longer be falsy.

PR #27092, opened 2020-11-19
[viki] fix stream extraction from mpd

beefchop:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This fixes a bug in parsing the streams json, where instead of downloading the video, it would only download the mpd file instead.

### DIFF
--- a/youtube_dl/extractor/viki.py
+++ b/youtube_dl/extractor/viki.py
@@ -296,6 +296,9 @@ class VikiIE(VikiBaseIE):
                         if f.get('acodec') == 'none' and f.get('vcodec') != 'none':
                             f['acodec'] = None
                     formats.extend(m3u8_formats)
+                elif format_id == 'mpd':
+                    formats.extend(self._extract_mpd_formats(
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/viki.py
```diff
@@ -296,6 +296,10 @@ def _real_extract(self, url):
                         if f.get('acodec') == 'none' and f.get('vcodec') != 'none':
                             f['acodec'] = None
                     formats.extend(m3u8_formats)
+                elif format_id == 'mpd':
+                    mpd_formats = self._extract_mpd_formats(
+                        format_url, video_id, mpd_id='DASH')
+                    formats.extend(mpd_formats)
```
remitamine, line 302:
```suggestion
                    formats.extend(self._extract_mpd_formats(
                        format_url, video_id, 'mpd-%s' % protocol, fatal=False))
```

remitamine (changes requested):

youtube_dl/extractor/viki.py
beefchop: done

Issue #18384, opened 2018-12-03
set mtime to video publish date


codl:
### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2018.12.03*. If it's not, read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2018.12.03**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
- [x] Checked that provided video/audio/playlist URLs (if any) are alive and playable in a browser

### What is the purpose of your *issue*?
- [ ] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [x] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### Description of your *issue*, suggested solution and other information

this is a followup to #18383 in which I found out that the mtime is only set according to the `Last-Modified` HTTP headers, and doesn't work for videos downloaded with DASH. I find this counter-intuitive: having downloaded videos and seen that the mtime was set to — as far as I could tell — the upload date, I expected this to happen all the time until it randomly didn't work for some videos.

I think that, whenever the upload time is available, it should be used for the final file's mtime in place of `Last-Modified`.

dotysan:
I just tested and it appears that all DASH *segments* indeed had the same/proper mtime when fetched via http.py. However when the segments are stitched together, the mtime is lost.

dotysan:
Not a pull request yet, because I'm new to the code in this project and haven't run the unit tests. But this seems to do the job!

```
diff --git a/youtube_dl/downloader/fragment.py b/youtube_dl/downloader/fragment.py
index 917f6dc01..37edcc900 100644
--- a/youtube_dl/downloader/fragment.py
+++ b/youtube_dl/downloader/fragment.py
@@ -116,6 +116,10 @@ class FragmentFD(FileDownloader):
         finally:
             if self.__do_ytdl_file(ctx):
[...]

PR #27138, opened 2020-11-21
set mtime for fragmented videos (such as DASH)

dotysan:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

set mtime for fragmented videos (such as DASH)
 - it is set to the mtime of the last fragment uploaded

See issue #18384.

### DIFF
--- a/youtube_dl/downloader/fragment.py
+++ b/youtube_dl/downloader/fragment.py
@@ -116,6 +116,10 @@ class FragmentFD(FileDownloader):
         finally:
             if self.__do_ytdl_file(ctx):
                 self._write_ytdl_file(ctx)
+            try:
+                os.utime(ctx['tmpfilename'], (time.time(), os.stat(ctx['fragment_filename_sanitized']).st_mtime))
+            except Exception:
+                pass
             if not self.params.get('keep_fragments', False):
                 os.remove(encodeFilename(ctx['fragment_filename_sanitized']))
             del ctx['fragment_filename_sanitized']
--- a/youtube_dl/downloader/fragment.py
+++ b/youtube_dl/downloader/fragment.py
@@ -116,10 +116,7 @@ class FragmentFD(FileDownloader):
         finally:
             if self.__do_ytdl_file(ctx):
                 self._write_ytdl_file(ctx)
-            try:
-                os.utime(ctx['tmpfilename'], (time.time(), os.stat(ctx['fragment_filename_sanitized']).st_mtime))
-            except Exception:
-                pass
+            self.mtime = os.stat(ctx['fragment_filename_sanitized']).st_mtime
             if not self.params.get('keep_fragments', False):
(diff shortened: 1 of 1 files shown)


dstftw:
This will set mtime for the final file *number-of-fragment* times. This is completely useless. mtime must be set one single time.

dotysan:
> This will set mtime for the final file _number-of-fragment_ times.

Agreed. Although the mtime of the last fragment is what is effective after the download is complete.

> This is completely useless. mtime must be set one single time.

Is the call to os.utime() for each fragment really that expensive? I agree it would be more efficient to do it only once.

> Another point that appears immediately - why the last fragment's mtime and not the first's or 42th's?

Because that's the last time the file was modified. I.e. when the last fragment finally uploaded.

dotysan:
OK @dstftw take another look at this PR. Indeed FileDownloader already calls os.utime() on every fragment, so there's no point in doing it twice on every fragment. I've stuffed it into a class variable and the syscall is just done once in _finish_frag_download() now.

dotysan:
> mtime must be set one single time.

@dstftw BTW, it already calls os.utime() on *every* fragment here:
https://github.com/ytdl-org/youtube-dl/blob/2cd43a00d1eec54c40e3fdf9a288f4cb391a8323/youtube_dl/downloader/common.py#L222

 This is in FileDownloader.try_utime() called from HttpFD.download(). And *in this case of fragmented files* it's just as wasteful, since we only need the mtime of the last fragment.

If you are truly concerned with eliminating wasteful syscalls, we should fix this too.

youtube_dl/downloader/fragment.py
```diff
@@ -116,6 +116,7 @@ def _append_fragment(self, ctx, frag_content):
         finally:
             if self.__do_ytdl_file(ctx):
                 self._write_ytdl_file(ctx)
+            self.mtime = os.stat(ctx['fragment_filename_sanitized']).st_mtime
```
dstftw, line 119: Nothing conceptually changed: you still request `st_mtime` unnecessarily *n-1* times.

dstftw (changes requested):

dstftw:
> > mtime must be set one single time.
> 
> @dstftw BTW, it already calls os.utime() on _every_ fragment here:
> https://github.com/ytdl-org/youtube-dl/blob/2cd43a00d1eec54c40e3fdf9a288f4cb391a8323/youtube_dl/downloader/common.py#L222
> 
> This is in FileDownloader.try_utime() called from HttpFD.download(). And _in this case of fragmented files_ it's just as wasteful, since we only need the mtime of the last fragment.
> 
> If you are truly concerned with eliminating wasteful syscalls, we should fix this too.

Calling `os.utime` on each separate fragment is perfectly legal since fragments are not necessarily removed after the download.

PR #27149, opened 2020-11-22
[MedalTV] Add new extractor

xenova:

---

- [ ] Bug fix
- [x] Improvement
- [x] New extractor
- [ ] New feature

---

Medal TV is a platform for gamers to record and share clips and videos. This is quite a popular site, so it was surprising that there is not an extractor for it yet. Although there is another pull request for Medal TV (https://github.com/ytdl-org/youtube-dl/pull/27118), this extractor provides all available formats (including source). The other pull request only supports 1080p or 720p. This extractor also provides error checking (in case a video does not exist for example). Furthermore, this extractor was tested on 1400 medal clips and every one produced expected behaviour.

### DIFF
--- a/youtube_dl/extractor/medaltv.py
+++ b/youtube_dl/extractor/medaltv.py
@@ -4,6 +4,12 @@ from __future__ import unicode_literals
 
 from .common import InfoExtractor
 from ..utils import ExtractorError
+from ..compat import compat_str
+from ..utils import (
+    try_get,
+    float_or_none,
+    int_or_none
+)
 
 
 class MedalTVIE(InfoExtractor):
@@ -43,17 +49,19 @@ class MedalTVIE(InfoExtractor):
         webpage = self._download_webpage(url, video_id)
 
         hydration_data = self._search_regex(r'<script>.*hydrationData\s*=\s*({.+?})\s*</script>', webpage, 'hydration data', default='{}')
-        parsed = self._parse_json(hydration_data, video_id, fatal=False)
+        parsed = self._parse_json(hydration_data, video_id)
 
-        author_info = next(iter(parsed['profiles'].values()))
-        clip_info = parsed['clips'][video_id]
+        author_info = next(iter(parsed.get('profiles').values()))
+        clip_info = parsed.get('clips').get(video_id)
 
         if('error' in clip_info):
-            if(clip_info['error'] == 404):
+            error_code = clip_info.get('error')
+            if(error_code == 404):
                 raise ExtractorError('That clip does not exist.', expected=True, video_id=video_id)
             else:
-                raise ExtractorError('An unknown error occurred ({}).'.format(clip_info['error']), video_id=video_id)
-        source_aspect_ratio = clip_info['sourceWidth'] / clip_info['sourceHeight']
+                raise ExtractorError('An unknown error occurred ({}).'.format(error_code), video_id=video_id)
+
+        source_aspect_ratio = clip_info.get('sourceWidth') / clip_info.get('sourceHeight')
 
         # ordered from worst to best quality
         heights = (144, 240, 360, 480, 720, 1080)
@@ -69,50 +77,51 @@ class MedalTVIE(InfoExtractor):
 
             # second condition needed as sometimes medal says
             # they have a format when in fact it is another format
-            if(video_key in clip_info and format_key in clip_info[video_key]):
+            if(video_key in clip_info and format_key in clip_info.get(video_key)):
                 formats.append({
-                    'url': clip_info[video_key],
+                    'url': clip_info.get(video_key),
                     'format_id': format_key,
                     'width': width,
                     'height': height
                 })
 
-            if(thumbnail_key in clip_info and format_key in clip_info[thumbnail_key]):
+            if(thumbnail_key in clip_info and format_key in clip_info.get(thumbnail_key)):
                 thumbnails.append({
                     'id': format_key,
-                    'url': clip_info[thumbnail_key],
+                    'url': clip_info.get(thumbnail_key),
                     'width': width,
                     'height': height
                 })
 
+        # add source to formats
         formats.append({
-            'url': clip_info['contentUrl'],
+            'url': clip_info.get('contentUrl'),
             'format_id': 'source',
-            'width': clip_info['sourceWidth'],
-            'height': clip_info['sourceHeight']
+            'width': clip_info.get('sourceWidth'),
+            'height': clip_info.get('sourceHeight')
         })
 
         return {
             'id': video_id,
-            'title': clip_info['contentTitle'] if clip_info['hasTitle'] else '',
+            'title': clip_info.get('contentTitle'),
 
             'formats': formats,
-            'url': clip_info['contentUrl'],
+            'url': clip_info.get('contentUrl'),
 
             'thumbnails': thumbnails,
-            'thumbnail': thumbnails[0]['url'],
+            'thumbnail': try_get(thumbnails, lambda x: x[-1]['url'], compat_str),
 
-            'description': clip_info['contentDescription'],
+            'description': clip_info.get('contentDescription'),
 
             # author information
-            'uploader': author_info['displayName'],
-            'timestamp': clip_info['created'] / 1000,
-            'uploader_id': author_info['id'],
-            'uploader_url': 'https://medal.tv/users/{}'.format(author_info['id']),
+            'uploader': author_info.get('displayName'),
+            'timestamp': float_or_none(clip_info.get('created') / 1000),
+            'uploader_id': author_info.get('id'),
+            'uploader_url': 'https://medal.tv/users/{}'.format(author_info.get('id')),
 
             # other clip information
-            'duration': clip_info['videoLengthSeconds'],
-            'view_count': clip_info['views'],
-            'like_count': clip_info['likes'],
-            'comment_count': clip_info['comments']
+            'duration': float_or_none(clip_info.get('videoLengthSeconds')),
+            'view_count': int_or_none(clip_info.get('views')),
+            'like_count': int_or_none(clip_info.get('likes')),
+            'comment_count': int_or_none(clip_info.get('comments'))
         }
--- a/youtube_dl/extractor/medaltv.py
+++ b/youtube_dl/extractor/medaltv.py
@@ -64,14 +64,6 @@ class MedalTVIE(InfoExtractor):
                 raise ExtractorError('An unknown error occurred ({0}).'.format(error),
                                      video_id=video_id)
 
-        # Necessary because the id of the author is not known in advance.
-        # Won't raise an issue if no profile can be found as this is optional.
-        author_info = try_get(parsed,
-                              lambda x: list(x['profiles'].values())[0], dict
-                              ) or {}
-        author_id = author_info.get('id')
-        author_url = 'https://medal.tv/users/{0}'.format(author_id) if author_id else None
-
         width = int_or_none(clip_info.get('sourceWidth'))
         height = int_or_none(clip_info.get('sourceHeight'))
 
@@ -118,6 +110,14 @@ class MedalTVIE(InfoExtractor):
                 'width': width,
                 'height': height
             })
+
+        # Necessary because the id of the author is not known in advance.
+        # Won't raise an issue if no profile can be found as this is optional.
+        author_info = try_get(parsed,
+                              lambda x: list(x['profiles'].values())[0], dict
+                              ) or {}
+        author_id = author_info.get('id')
+        author_url = 'https://medal.tv/users/{0}'.format(author_id) if author_id else None
 
         return {
             'id': video_id,
--- a/youtube_dl/extractor/medaltv.py
+++ b/youtube_dl/extractor/medaltv.py
@@ -55,15 +55,6 @@ class MedalTVIE(InfoExtractor):
             raise ExtractorError('Could not find video information.',
                                  video_id=video_id)
 
-        error = clip_info.get('error')
-        if error:
-            if(error == 404):
-                raise ExtractorError('That clip does not exist.',
-                                     expected=True, video_id=video_id)
-            else:
-                raise ExtractorError('An unknown error occurred ({0}).'.format(error),
-                                     video_id=video_id)
-
         width = int_or_none(clip_info.get('sourceWidth'))
         height = int_or_none(clip_info.get('sourceHeight'))
 
@@ -110,6 +101,15 @@ class MedalTVIE(InfoExtractor):
                 'width': width,
                 'height': height
             })
+
+        error = clip_info.get('error')
+        if not formats and error:
+            if(error == 404):
+                raise ExtractorError('That clip does not exist.',
+                                     expected=True, video_id=video_id)
+            else:
+                raise ExtractorError('An unknown error occurred ({0}).'.format(error),
+                                     video_id=video_id)
 
         # Necessary because the id of the author is not known in advance.
         # Won't raise an issue if no profile can be found as this is optional.
--- a/youtube_dl/extractor/medaltv.py
+++ b/youtube_dl/extractor/medaltv.py
@@ -47,16 +47,16 @@ class MedalTVIE(InfoExtractor):
 
         hydration_data = self._search_regex(
             r'<script[^>]*>\s*(?:var\s*)?hydrationData\s*=\s*({.+?})\s*</script>',
-            webpage, 'hydration data', default='{}', fatal=True)
-        parsed = self._parse_json(hydration_data, video_id, fatal=True)
+            webpage, 'hydration data', default='{}')
+        parsed = self._parse_json(hydration_data, video_id)
 
         clip_info = try_get(parsed, lambda x: x['clips'][video_id], dict) or {}
-        if(not clip_info):
+        if not clip_info:
             raise ExtractorError('Could not find video information.',
                                  video_id=video_id)
 
         error = clip_info.get('error')
-        if(error):
+        if error:
             if(error == 404):
                 raise ExtractorError('That clip does not exist.',
                                      expected=True, video_id=video_id)
--- a/youtube_dl/extractor/medaltv.py
+++ b/youtube_dl/extractor/medaltv.py
@@ -45,63 +45,78 @@ class MedalTVIE(InfoExtractor):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
 
-        hydration_data = self._search_regex(r'<script>.*hydrationData\s*=\s*({.+?})\s*</script>', webpage, 'hydration data', default='{}')
-        parsed = self._parse_json(hydration_data, video_id)
+        hydration_data = self._search_regex(
+            r'<script[^>]*>\s*(?:var\s*)?hydrationData\s*=\s*({.+?})\s*</script>',
+            webpage, 'hydration data', default='{}', fatal=True)
+        parsed = self._parse_json(hydration_data, video_id, fatal=True)
 
-        # This is necessary because the id of the author is not known in advance.
-        # Will not raise an issue if no profile can be found as this is optional.
-        author_info = try_get(parsed.get('profiles'), lambda x: list(x.values())[0], dict) or {}
-        author_url = 'https://medal.tv/users/{}'.format(author_info.get('id')) if author_info.get('id') else None
-
-        clip_info = try_get(parsed.get('clips'), lambda x: x.get(video_id), dict) or {}
+        clip_info = try_get(parsed, lambda x: x['clips'][video_id], dict) or {}
         if(not clip_info):
-            raise ExtractorError('Could not find video information.', video_id=video_id)
+            raise ExtractorError('Could not find video information.',
+                                 video_id=video_id)
 
-        if('error' in clip_info):
-            if(clip_info.get('error') == 404):
-                raise ExtractorError('That clip does not exist.', expected=True, video_id=video_id)
+        error = clip_info.get('error')
+        if(error):
+            if(error == 404):
+                raise ExtractorError('That clip does not exist.',
+                                     expected=True, video_id=video_id)
             else:
-                raise ExtractorError('An unknown error occurred ({}).'.format(clip_info.get('error')), video_id=video_id)
+                raise ExtractorError('An unknown error occurred ({0}).'.format(error),
+                                     video_id=video_id)
 
-        source_aspect_ratio = try_get(clip_info, lambda x: x.get('sourceWidth') / x.get('sourceHeight'), float) or (16 / 9)
+        # Necessary because the id of the author is not known in advance.
+        # Won't raise an issue if no profile can be found as this is optional.
+        author_info = try_get(parsed,
+                              lambda x: list(x['profiles'].values())[0], dict
+                              ) or {}
+        author_id = author_info.get('id')
+        author_url = 'https://medal.tv/users/{0}'.format(author_id) if author_id else None
 
-        # ordered from worst to best quality
+        width = int_or_none(clip_info.get('sourceWidth'))
+        height = int_or_none(clip_info.get('sourceHeight'))
+
+        aspect_ratio = (width / height) if(width and height) else (16 / 9)
+
+        # ordered from lowest to highest resolution
         heights = (144, 240, 360, 480, 720, 1080)
 
         formats = []
         thumbnails = []
 
         for height in heights:
-            format_key = '{}p'.format(height)
-            video_key = 'contentUrl{}'.format(format_key)
-            thumbnail_key = 'thumbnail{}'.format(format_key)
-            width = int(round(source_aspect_ratio * height))
+            format_key = '{0}p'.format(height)
+            video_key = 'contentUrl{0}'.format(format_key)
+            thumbnail_key = 'thumbnail{0}'.format(format_key)
+            width = int(round(aspect_ratio * height))
 
             # Second condition needed as sometimes medal says
             # they have a format when in fact it is another format.
-            if(video_key in clip_info and format_key in clip_info.get(video_key)):
+            format_url = clip_info.get(video_key)
+            if(format_url and format_key in format_url):
                 formats.append({
-                    'url': clip_info.get(video_key),
+                    'url': format_url,
                     'format_id': format_key,
                     'width': width,
                     'height': height
                 })
 
-            if(thumbnail_key in clip_info and format_key in clip_info.get(thumbnail_key)):
+            thumbnail_url = clip_info.get(thumbnail_key)
+            if(thumbnail_url and format_key in thumbnail_url):
                 thumbnails.append({
                     'id': format_key,
-                    'url': clip_info.get(thumbnail_key),
+                    'url': thumbnail_url,
                     'width': width,
                     'height': height
                 })
 
         # add source to formats
-        if('contentUrl' in clip_info):
+        source_url = clip_info.get('contentUrl')
+        if(source_url):
             formats.append({
-                'url': clip_info.get('contentUrl'),
+                'url': source_url,
                 'format_id': 'source',
-                'width': clip_info.get('sourceWidth'),
-                'height': clip_info.get('sourceHeight')
+                'width': width,
+                'height': height
             })
 
         return {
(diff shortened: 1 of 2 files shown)


remitamine:
read and apply [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions).

xenova:
@remitamine That should be good now. 👍

Let me know if anything else needs changing! Thanks.

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,127 @@
...
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        url = 'https://medal.tv/clips/{}'.format(video_id)
+        webpage = self._download_webpage(url, video_id)
+
+        hydration_data = self._search_regex(r'<script>.*hydrationData\s*=\s*({.+?})\s*</script>', webpage, 'hydration data', default='{}')
+        parsed = self._parse_json(hydration_data, video_id)
+
+        author_info = next(iter(parsed.get('profiles').values()))
```
remitamine, line 54: will break the extraction if `profiles` is empty or `None`.

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,127 @@
...
+        clip_info = parsed.get('clips').get(video_id)
```
remitamine, line 55: breaks the extraction if `clips` is `None`.

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,127 @@
...
+
+        if('error' in clip_info):
```
remitamine, line 57: same if `clip_info` is `None`.

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,127 @@
...
+            error_code = clip_info.get('error')
+            if(error_code == 404):
+                raise ExtractorError('That clip does not exist.', expected=True, video_id=video_id)
+            else:
+                raise ExtractorError('An unknown error occurred ({}).'.format(error_code), video_id=video_id)
+
+        source_aspect_ratio = clip_info.get('sourceWidth') / clip_info.get('sourceHeight')
```
remitamine, line 64: same if one of the values is `None`.

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,127 @@
...
+
+        return {
+            'id': video_id,
+            'title': clip_info.get('contentTitle'),
+
+            'formats': formats,
+            'url': clip_info.get('contentUrl'),
+
+            'thumbnails': thumbnails,
+            'thumbnail': try_get(thumbnails, lambda x: x[-1]['url'], compat_str),
```
remitamine, line 112: remove, youtube-dl will create the `thumbnail` key from the `thumbnails` list.

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,127 @@
...
+
+            'description': clip_info.get('contentDescription'),
+
+            # author information
```
remitamine, line 116: remove comments that can be easily deduced from the code.

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,127 @@
...
+            'uploader': author_info.get('displayName'),
+            'timestamp': float_or_none(clip_info.get('created') / 1000),
```
remitamine, line 118:
```suggestion
            'timestamp': float_or_none(clip_info.get('created'), 1000),
```

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,127 @@
...
+            'uploader_id': author_info.get('id'),
+            'uploader_url': 'https://medal.tv/users/{}'.format(author_info.get('id')),
```
remitamine, line 120: will return `https://medal.tv/users/None` if an `author id` couldn't be extracted.

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,127 @@
...
+            'timestamp': 1605580939,
+            'upload_date': '20201117',
+            'uploader_id': 5156321,
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        url = 'https://medal.tv/clips/{}'.format(video_id)
```
remitamine, line 48: why?

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,127 @@
...
+                thumbnails.append({
+                    'id': format_key,
+                    'url': clip_info.get(thumbnail_key),
+                    'width': width,
+                    'height': height
+                })
+
+        # add source to formats
+        formats.append({
+            'url': clip_info.get('contentUrl'),
```
remitamine, line 98: check the existence of the `contentUrl` before adding the format.

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,127 @@
...
+from ..compat import compat_str
+from ..utils import (
+    try_get,
+    float_or_none,
+    int_or_none
+)
+
+
+class MedalTVIE(InfoExtractor):
+    _VALID_URL = r'(?:https://)?(?:www\.)?medal\.tv(?:/clips)?/(?P<id>[0-9]+)'
```
remitamine, line 16: the `https?` part should not be optional.
do you have an example of clip URL without the `/clips` part?

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,127 @@
...
+            'width': clip_info.get('sourceWidth'),
+            'height': clip_info.get('sourceHeight')
+        })
+
+        return {
+            'id': video_id,
+            'title': clip_info.get('contentTitle'),
+
+            'formats': formats,
+            'url': clip_info.get('contentUrl'),
```
remitamine, line 109: remove, will be set by youtube-dl using `formats` dict.

remitamine (changes requested):
check the code with flake8.

youtube_dl/extractor/medaltv.py
xenova: Addressed this using `author_info = try_get(profiles, lambda x: list(x.values())[0], dict) or {}`

youtube_dl/extractor/medaltv.py
xenova: Addressed this by raising an error if there is no clip information (i.e. is fatal)

```python
clip_info = try_get(parsed.get('clips'), lambda x: x.get(video_id), dict) or {}
if(not clip_info):
    raise ExtractorError('Could not find video information.', video_id=video_id)
```

youtube_dl/extractor/medaltv.py
xenova: Above change also fixes this.

youtube_dl/extractor/medaltv.py
xenova: The intention behind this was to enforce that the HTML returned in always from the clip's main page. However, I can remove this because it probably isn't necessary.

youtube_dl/extractor/medaltv.py
xenova: Removed 👍

youtube_dl/extractor/medaltv.py
xenova: Resolved by using `try_get`:
```python
source_aspect_ratio = try_get(clip_info, lambda x: x.get('sourceWidth')/x.get('sourceHeight'), float) or (16/9)
```

Otherwise defaulting to standard aspect ratio 16:9

youtube_dl/extractor/medaltv.py
xenova: Removed 👍

youtube_dl/extractor/medaltv.py
xenova: Resolved by checking if author id is present:
```python
author_url = 'https://medal.tv/users/{}'.format(author_info.get('id')) if author_info.get('id') else None
```

youtube_dl/extractor/medaltv.py
xenova: Done 👍

youtube_dl/extractor/medaltv.py
xenova: As mentioned in the pull request, I tested the extractor against 1300 medal clips. Some of these did not include the https, so I decided to make it optional. However, after seeing how other extractors work, I will make it required.

The choice to have `/clips` optional was also to do with another comment you made ( changing `url = 'https://medal.tv/clips/{}'.format(video_id)`). As some of those 1300 people submitted links like: `https://cdn.medal.tv/10619174/1080p-34934644.mp4`, and I thought it would be good to extract all formats from it (i.e. pretending it is `https://medal.tv/clips/10619174`. But it would seem as the default extractor is prioritized when dealing with direct links. I will change the regex to have `clips` as required.

youtube_dl/extractor/medaltv.py
xenova: Removed 👍

youtube_dl/extractor/medaltv.py
xenova: Changed 👍

xenova:
Okay, third time's a charm!

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,123 @@
...
+            'upload_date': '20201117',
+            'uploader_id': 5156321,
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        hydration_data = self._search_regex(r'<script>.*hydrationData\s*=\s*({.+?})\s*</script>', webpage, 'hydration data', default='{}')
```
remitamine, line 48: `<script>.*` this part does not play a roll in the regex, may be change to `<script[^>]*>\s*(var\s*)?`.
Split Long lines(longer than 80 columns).
```suggestion
        hydration_data = self._search_regex(
            r'<script>.*hydrationData\s*=\s*({.+?})\s*</script>',
            webpage, 'hydration data', default='{}')
```

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,123 @@
...
+        parsed = self._parse_json(hydration_data, video_id)
+
+        # This is necessary because the id of the author is not known in advance.
+        # Will not raise an issue if no profile can be found as this is optional.
+        author_info = try_get(parsed.get('profiles'), lambda x: list(x.values())[0], dict) or {}
+        author_url = 'https://medal.tv/users/{}'.format(author_info.get('id')) if author_info.get('id') else None
```
remitamine, line 54: try put value that are used multiple times in a variable(ex: `author_info.get('id')`).
i think the `{}` format is not supported in python 2.6.

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,123 @@
...
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        hydration_data = self._search_regex(r'<script>.*hydrationData\s*=\s*({.+?})\s*</script>', webpage, 'hydration data', default='{}')
+        parsed = self._parse_json(hydration_data, video_id)
+
+        # This is necessary because the id of the author is not known in advance.
+        # Will not raise an issue if no profile can be found as this is optional.
+        author_info = try_get(parsed.get('profiles'), lambda x: list(x.values())[0], dict) or {}
```
remitamine, line 53:
```suggestion
        author_info = try_get(parsed, lambda x: list(x['profiles'].values())[0], dict) or {}
```

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,123 @@
...
+        author_url = 'https://medal.tv/users/{}'.format(author_info.get('id')) if author_info.get('id') else None
+
+        clip_info = try_get(parsed.get('clips'), lambda x: x.get(video_id), dict) or {}
```
remitamine, line 56:
```suggestion
        clip_info = try_get(parsed, lambda x: x['clips'][video_id], dict) or {}
```

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,123 @@
...
+        if(not clip_info):
+            raise ExtractorError('Could not find video information.', video_id=video_id)
+
+        if('error' in clip_info):
```
remitamine, line 60: error into a common variable and reuse it.
it would be better to check for error only after not been able extracted mandatory information(`video_id`, `title` and `formats`).
```suggestion
        error = clip_info.get('error')
        if error:
```

remitamine (changes requested):
check the code with `flake8`.

youtube_dl/extractor/medaltv.py
remitamine: you can return the part without the `/clips/` but make it more restrictive(check for `cdn.medal.tv`) and add `matching_only` test.

youtube_dl/extractor/medaltv.py
xenova: Changed to:
```python
        hydration_data = self._search_regex(
            r'<script[^>]*>\s*(?:var\s*)?hydrationData\s*=\s*({.+?})\s*</script>',
            webpage, 'hydration data', default='{}')
```

youtube_dl/extractor/medaltv.py
xenova: Oh yes you are correct, looks like `{}` requires a positional argument for python 2.6 and 3.0. Source (https://stackoverflow.com/questions/19668395/str-format-for-python-2-6-gives-error-where-2-7-does-not)

Fixed both issues as follows:
```python
author_id = author_info.get('id')
author_url = 'https://medal.tv/users/{0}'.format(author_id) if author_id else None
```

youtube_dl/extractor/medaltv.py
xenova: Changed 👍

youtube_dl/extractor/medaltv.py
xenova: Changed 👍

youtube_dl/extractor/medaltv.py
xenova: When you say "it would be better to check for error only after not been able extracted mandatory information(video_id, title and formats).", I believe I have already done this?

```python
        clip_info = try_get(parsed, lambda x: x['clips'][video_id], dict) or {}
        if(not clip_info):
            raise ExtractorError('Could not find video information.', video_id=video_id)

        error = clip_info.get('error')
        if(error):
            if(error == 404):
                raise ExtractorError('That clip does not exist.',
                                     expected=True, video_id=video_id)
            else:
                raise ExtractorError('An unknown error occurred ({0}).'.format(error),
                                     video_id=video_id)
```

I check for error `error = clip_info.get('error')` only after not being able to extract mandatory information: `if(not clip_info):`. `clip_info` contains the mandatory information, so if that does not exist, then the whole process should break.

youtube_dl/extractor/medaltv.py
xenova: However, is it not the case that the default extractor is prioritized when dealing with direct links? Links from `cdn.medal.tv` are always direct links, so, regardless of the regex, it would use the default extractor?

youtube_dl/extractor/medaltv.py
remitamine: even if `clip_info` exist, it's possibe that there is no format there(empty `contentUrl*` values).

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,123 @@
...
+            if(thumbnail_key in clip_info and format_key in clip_info.get(thumbnail_key)):
+                thumbnails.append({
+                    'id': format_key,
+                    'url': clip_info.get(thumbnail_key),
+                    'width': width,
+                    'height': height
+                })
+
+        # add source to formats
+        if('contentUrl' in clip_info):
```
remitamine, line 99: this does not handle the case where `contentUrl` value is `None`.

xenova:
When checking the code with `flake8`, it returns no errors:

```
$ flake8 youtube_dl/extractor/medaltv.py


```
However, when checking with `http://pep8online.com/checkresult#`, it returns this:
```
E501:49:80:line too long (82 > 79 characters)
E501:64:80:line too long (86 > 79 characters)
E501:73:80:line too long (90 > 79 characters)
```

which refer to the following lines (respectively):
```python
# E501:49:80:
            r'<script[^>]*>\s*(?:var\s*)?hydrationData\s*=\s*({.+?})\s*</script>',
# E501:64:80:
                raise ExtractorError('An unknown error occurred ({0}).'.format(error),
# E501:73:80:
        author_url = 'https://medal.tv/users/{0}'.format(author_id) if author_id else None
```

Each of these contain string literals, which we were advised not to split. I can always make them into variables, but we were advised against that too. I will commit the above changes, however.

remitamine:
> However, when checking with http://pep8online.com/checkresult#, it returns this:
just use the command locally(so, it reads the `setup.cfg` file and apply the rules there).

xenova:
Okay I see 👍 ... I've tested with flake8, but no errors appear. However, when tested with pep8 (which is now called pycodestyle), it returns the above recommendations.
```
joshu@DESKTOP-I39NJG7 MINGW64 /d/joshu/Xenova/Projects/youtube-dl (medaltv)
$ ls
AUTHORS  bin  build  ChangeLog  CONTRIBUTING.md  devscripts  dist  docs  LICENSE  Makefile  MANIFEST.in  README.md  setup.cfg  setup.py  test  tox.ini  youtube_dl  youtube-dl.plugin.zsh

joshu@DESKTOP-I39NJG7 MINGW64 /d/joshu/Xenova/Projects/youtube-dl (medaltv)
$ flake8 youtube_dl/extractor/medaltv.py

joshu@DESKTOP-I39NJG7 MINGW64 /d/joshu/Xenova/Projects/youtube-dl (medaltv)
$ pycodestyle youtube_dl/extractor/medaltv.py
youtube_dl/extractor/medaltv.py:49:80: E501 line too long (82 > 79 characters)
youtube_dl/extractor/medaltv.py:64:80: E501 line too long (86 > 79 characters)
youtube_dl/extractor/medaltv.py:73:80: E501 line too long (90 > 79 characters)

joshu@DESKTOP-I39NJG7 MINGW64 /d/joshu/Xenova/Projects/youtube-dl (medaltv)
$
```

However, like I said above, these lines contain string literals, which we were advised we should not change (https://github.com/ytdl-org/youtube-dl/blob/master/CONTRIBUTING.md#long-lines-policy)

> For example, you should never split long string literals like URLs or some other often copied entities over multiple lines to fit this limit

Should I use variables to shorten it instead?

youtube_dl/extractor/medaltv.py
xenova: Okay, I see what you mean, but the JSON returned for an error (when a clip does not exist) differs in structure from a normal clip:

WORKING CLIP:
```json
{
   "clips":{
      "34934644":{
         "contentId":34934644,
         "contentTitle":"Quad Cold",
         "contentDescription":"Medal,https://medal.tv/desktop/",
         "sourceWidth":1920,
         "sourceHeight":1080,
         "videoLengthSeconds":23,
         "contentUrl1080p":"https://cdn.medal.tv/10619174/1080p-34934644.mp4",
         "contentUrl720p":"https://cdn.medal.tv/10619174/720p-34934644.mp4",
         "contentUrl480p":"https://cdn.medal.tv/10619174/720p-34934644.mp4",
         "contentUrl360p":"https://cdn.medal.tv/10619174/360p-34934644.mp4",
         "contentUrl240p":"https://cdn.medal.tv/10619174/360p-34934644.mp4",
         "contentUrl144p":"https://cdn.medal.tv/10619174/360p-34934644.mp4",
         "thumbnail1080p":"https://cdn.medal.tv/10619174/thumbnail-34934644-720p.jpg",
         "thumbnail720p":"https://cdn.medal.tv/10619174/thumbnail-34934644-720p.jpg",
         "thumbnail480p":"https://cdn.medal.tv/10619174/thumbnail-34934644-720p.jpg",
         "thumbnail360p":"https://cdn.medal.tv/10619174/thumbnail-34934644-360p.jpg",
         "thumbnail240p":"https://cdn.medal.tv/10619174/thumbnail-34934644-360p.jpg",
         "thumbnail144p":"https://cdn.medal.tv/10619174/thumbnail-34934644-360p.jpg",
         "contentUrl":"https://cdn.medal.tv/source/10619174/1603165173-medaltvcallofdutyrblackopscoldwar20201019203926.mp4",
         "socialMediaVideo":"https://cdn.medal.tv/10619174/share-34934644.mp4",
         "created":1603165266000,
         "likes":80,
         "views":8702,
         "comments":18,
         "id":34934644
      }
   },
   "profiles":{
      "10619174":{
         "id":10619174,
         "userId":10619174,
         "displayName":"MowgliSB",
         "userName":"MowgliSB"
      }
   }
}
```

CLIP NOT FOUND:
```json
{
   "clips":{
      "12345":{
         "id":"12345",
         "error":404
      }
   },
   "profiles":{
      "undefined":{
         "status":"None",
         "error":{
            "message":"invalid json response body at http://api-lb.api/users/undefined reason: Unexpected end of JSON input",
            "type":"invalid-json"
         }
      }
   }
}
```

So, if `error` is in `clip_info`, then it automatically is invalid. This is why I check for `error` in `clip_info` before anything else. Because if `error` is in `clip_info`, there will never be format information.

You say that
> even if clip_info exist, it's possibe that there is no format there(empty contentUrl* values).

But if `clip_info` exists (i.e. `parsed['clips'][video_id]` exists) and `parsed['clips'][video_id]['error']` exists, then there will not be a `contentUrl` value.

Are you saying that I must check if at least one format is returned? E.g. in the case when there is `clip_info`, no `clip_info['error']` and no `clip_info['contentUrl']`? If so, surely that will be covered in the case when there are no formats retrieved? 
```ERROR: No video formats found!; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.``` Although it has never happened to me, if it does happen, youtube-dl will handle it correctly.

File: youtube_dl/extractor/medaltv.py (hunk 18 above)
Comment (xenova) [reply youtube_dl/extractor/medaltv.py]: Yes, I have changed this in the latest commit to be:
```python
		source_url = clip_info.get('contentUrl')
		if(source_url):
```

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,138 @@
...
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        hydration_data = self._search_regex(
+            r'<script[^>]*>\s*(?:var\s*)?hydrationData\s*=\s*({.+?})\s*</script>',
+            webpage, 'hydration data', default='{}', fatal=True)
```
remitamine, line 50: no need to set `fatal=True`, this is the default.

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,138 @@
...
+        parsed = self._parse_json(hydration_data, video_id, fatal=True)
```
remitamine, line 51: same.

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,138 @@
...
+
+        clip_info = try_get(parsed, lambda x: x['clips'][video_id], dict) or {}
+        if(not clip_info):
```
remitamine, line 54: no need for parenthesis(prefered way in python), unless it's needed.
```suggestion
        if not clip_info:
```

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,138 @@
...
+            raise ExtractorError('Could not find video information.',
+                                 video_id=video_id)
+
+        error = clip_info.get('error')
+        if(error):
```
remitamine, line 59: same.

remitamine (changes requested):

youtube_dl/extractor/medaltv.py
xenova: Changed 👍

remitamine:
> Should I use variables to shorten it instead?

the requirment is for the code to be checked with flake8, but if you want to get rid of these warning as well, then may be you can use a variable for user URL template, and split the `raise ExtractorError`, for the first line i think it can stay as it is.

youtube_dl/extractor/medaltv.py
xenova: Changed 👍

youtube_dl/extractor/medaltv.py
xenova: Changed 👍

youtube_dl/extractor/medaltv.py
xenova: Changed 👍

xenova:
Okay 👍 ... since the only requirement is flake8 (and it passes that), I will leave it as is. 

According to `CONTRIBUTING.md`, we are advised not to split strings (https://github.com/ytdl-org/youtube-dl/blob/master/CONTRIBUTING.md#long-lines-policy)
and we should use inline values where possible (https://github.com/ytdl-org/youtube-dl/blob/master/CONTRIBUTING.md#inline-values)

youtube_dl/extractor/medaltv.py
remitamine: check the other instances as well, those where just examples.

youtube_dl/extractor/medaltv.py
remitamine: it's also to cover the other possiblility, both `error` and `contentUrl*` exist, the response may change for different errors.
so in general it's prefered to check the existance of formats before raising an error(ex: [YahooIE](https://github.com/ytdl-org/youtube-dl/blob/master/youtube_dl/extractor/yahoo.py#L278-L279), [GoogleDriveIE](https://github.com/ytdl-org/youtube-dl/blob/master/youtube_dl/extractor/googledrive.py#L249), etc...).

xenova: Okay I see! Thanks for explaining 😃👍

youtube_dl/extractor/medaltv.py
```diff
@@ -0,0 +1,138 @@
...
+                })
+
+            thumbnail_url = clip_info.get(thumbnail_key)
+            if(thumbnail_url and format_key in thumbnail_url):
+                thumbnails.append({
+                    'id': format_key,
+                    'url': thumbnail_url,
+                    'width': width,
+                    'height': height
+                })
```
remitamine, line 93: the process to extract the format and the thumbnail is similar, so these part needs to be abstracted to remove duplication.

xenova: Okay, that is fair. Are we talking about adding something like:
```python
        info = {
            'formats': {
                'list': [],
                'key_name': 'contentUrl',
                'id_name': 'format_id'
            },
            'thumbnails': {
                'list': [],
                'key_name': 'thumbnail',
                'id_name': 'id'
            }
        }

        for height in heights:
            width = int(round(aspect_ratio * height))
            formatted_key = '{0}p'.format(height)

            for info_type in info:
                key = info[info_type]['key_name'] + formatted_key

                # Second condition needed as sometimes medal says
                # they have a format when in fact it is another format.
                url = clip_info.get(key)
                if(url and formatted_key in url):
                    info[info_type]['list'].append({
                        'url': url,
                        info[info_type]['id_name']: formatted_key,
                        'width': width,
                        'height': height
                    })
```

I originally chose to separate them, because youtube-dl requires different names for format ids, https://github.com/ytdl-org/youtube-dl/blob/7f41a598b3fba1bcab2817de64a08941200aa3c8/youtube_dl/extractor/common.py#L94-L303 as well as that the name templates were slightly different.

I could change the `info` dictionary to be a list, but the dictionary keys help name the items.

remitamine: kind of, i will try to abstract it further later(the `source` format also shares a bit code with this part).

pick-the-spiral-galaxy:
Thank you for a good Medal extractor! I closed my own PR for it (mine was just code laying around for personal use, needed 1080p only).

PR #27217, opened 2020-11-26
[generic] Use _meta_regex to extract twitter:player:stream tag

jetomit:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Use the standard (and more robust) regex when looking for `twitter:player:stream` meta tag. Some sites include additional attributes in the tag, which prevents the current, simpler regex from matching.

I don’t know if there’s always just one such tag; in that case, `_html_search_meta` could be used instead. This PR preserves current behavior and returns a list of all matching tags.

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -3335,8 +3335,8 @@ class GenericIE(InfoExtractor):
             # twitter:player:stream should be checked before twitter:player since
             # it is expected to contain a raw stream (see
             # https://dev.twitter.com/cards/types/player#On_twitter.com_via_desktop_browser)
-            found = filter_video(re.findall(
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/generic.py
```diff
@@ -3335,8 +3335,8 @@ def filter_video(urls):
             # twitter:player:stream should be checked before twitter:player since
             # it is expected to contain a raw stream (see
             # https://dev.twitter.com/cards/types/player#On_twitter.com_via_desktop_browser)
-            found = filter_video(re.findall(
-                r'<meta (?:property|name)="twitter:player:stream" (?:content|value)="(.+?)"', webpage))
+            found = filter_video(
+                m.group('content') for m in re.finditer(self._meta_regex("twitter:player:stream"), webpage))
```
dstftw, line 3339: These are not identical since the latter does not cover `value` attribute.

PR #27230, opened 2020-11-28
[nhk] Add support for NHK video programs

rolledback:

---

- [ ] Bug fix
- [x] Improvement
- [x] New extractor
- [ ] New feature

---

This change adds an extractor for NHK which handles video program pages. This extractor will allow users to download all available episodes of a video program at once instead of having to download each episode one at a time.

### DIFF
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -725,7 +725,10 @@ from .nexx import (
     NexxEmbedIE,
 )
 from .nfl import NFLIE
-from .nhk import NhkVodIE
+from .nhk import (
+    NhkVodIE,
+    NhkVodProgramIE,
+)
 from .nhl import NHLIE
 from .nick import (
     NickIE,
--- a/youtube_dl/extractor/nhk.py
+++ b/youtube_dl/extractor/nhk.py
@@ -3,14 +3,96 @@ from __future__ import unicode_literals
 import re
 
 from .common import InfoExtractor
+from ..utils import ExtractorError
 
 
-class NhkVodIE(InfoExtractor):
+class NhkBaseIE(InfoExtractor):
+    _API_URL_TEMPLATE = 'https://api.nhk.or.jp/nhkworld/%sod%slist/v7a/%s/%s/%s/all%s.json'
+
+    def _get_clean_field(self, episode, key):
+        return episode.get(key + '_clean') or episode.get(key)
+
+    def _list_episodes(self, m_id, lang, is_video, is_episode):
+        return self._download_json(
+            self._API_URL_TEMPLATE % (
+                'v' if is_video else 'r',
+                'clip' if m_id[:4] == '9999' else 'esd',
+                'episode' if is_episode else 'program',
+                m_id, lang, '/all' if is_video else ''),
+            m_id, query={'apikey': 'EJfK8jdS57GqlupFgAfAAwr573q01y6k'})['data']['episodes']
+
+    def _parse_episode_json(self, episode, lang, is_video):
+        title = episode.get('sub_title_clean') or episode['sub_title']
+
+        episode_id = None
+        if is_video:
+            pgm_id = episode.get('pgm_id')
+            pgm_no = episode.get('pgm_no')
+
+            if not (pgm_id and pgm_no):
+                missing_field = 'pgm_id' if not pgm_id else 'pgm_no'
+                raise ExtractorError('Cannot download episode. Field %s is missing from episode JSON.' % missing_field)
+
+            episode_id = pgm_id + pgm_no
+        else:
+            pgm_gr_id = episode.get('pgm_gr_id')
+            first_onair_date = episode.get('first_onair_date')
+            first_onair_no = episode.get('first_onair_no')
+
+            if not (pgm_gr_id and first_onair_date and first_onair_no):
+                missing_field = 'pgm_gr_id' if not pgm_gr_id else 'first_onair_date' if not first_onair_date else 'first_onair_no'
+                raise ExtractorError('Cannot download episode. Field %s is missing from episode JSON.' % missing_field)
+
+            episode_id = pgm_gr_id + '-' + first_onair_date + '-' + first_onair_no
+
+        series = self._get_clean_field(episode, 'title')
+
+        thumbnails = []
+        for s, w, h in [('', 640, 360), ('_l', 1280, 720)]:
+            img_path = episode.get('image' + s)
+            if not img_path:
+                continue
+            thumbnails.append({
+                'id': '%dp' % h,
+                'height': h,
+                'width': w,
+                'url': 'https://www3.nhk.or.jp' + img_path,
+            })
+
+        info = {
+            'id': episode_id + '-' + lang,
+            'title': '%s - %s' % (series, title) if series and title else title,
+            'description': self._get_clean_field(episode, 'description'),
+            'thumbnails': thumbnails,
+            'series': series,
+            'episode': title,
+        }
+
+        if is_video:
+            info.update({
+                '_type': 'url_transparent',
+                'ie_key': 'Piksel',
+                'url': 'https://player.piksel.com/v/refid/nhkworld/prefid/' + episode['vod_id'],
+            })
+        else:
+            audio = episode['audio']
+            audio_path = audio['audio']
+            info['formats'] = self._extract_m3u8_formats(
+                'https://nhkworld-vh.akamaihd.net/i%s/master.m3u8' % audio_path,
+                episode_id, 'm4a', entry_protocol='m3u8_native',
+                m3u8_id='hls', fatal=False)
+            for f in info['formats']:
+                f['language'] = lang
+
+        return info
+
+
+class NhkVodIE(NhkBaseIE):
     _VALID_URL = r'https?://www3\.nhk\.or\.jp/nhkworld/(?P<lang>[a-z]{2})/ondemand/(?P<type>video|audio)/(?P<id>\d{7}|[^/]+?-\d{8}-\d+)'
     # Content available only for a limited period of time. Visit
     # https://www3.nhk.or.jp/nhkworld/en/ondemand/ for working samples.
     _TESTS = [{
-        # video clip
+        # clip
         'url': 'https://www3.nhk.or.jp/nhkworld/en/ondemand/video/9999011/',
         'md5': '256a1be14f48d960a7e61e2532d95ec3',
         'info_dict': {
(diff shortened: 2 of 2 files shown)


youtube_dl/extractor/nhk.py
```diff
@@ -91,3 +100,23 @@ def get_clean_field(key):
...
+
+        is_video_program = m_type == 'program/video'
+        if is_video_program:
+            episodes = self._download_json(
+                self._API_URL_TEMPLATE % (
+                    'v',
+                    'esd',
+                    'program',
+                    episode_id, lang, '/all'),
+                episode_id, query={'apikey': 'EJfK8jdS57GqlupFgAfAAwr573q01y6k'})['data']['episodes']
```
remitamine, line 115: extract into a seperate method to reduce code duplication.

youtube_dl/extractor/nhk.py
```diff
@@ -6,7 +6,7 @@
 
 
 class NhkVodIE(InfoExtractor):
-    _VALID_URL = r'https?://www3\.nhk\.or\.jp/nhkworld/(?P<lang>[a-z]{2})/ondemand/(?P<type>video|audio)/(?P<id>\d{7}|[^/]+?-\d{8}-\d+)'
+    _VALID_URL = r'https?://www3\.nhk\.or\.jp/nhkworld/(?P<lang>[a-z]{2})/ondemand/(?P<type>video|audio|program/video)/(?P<id>\d{7}|[^/]+?-\d{8}-\d+|\w+)'
```
remitamine, line 9: create a seperate extractor.

remitamine (changes requested):

remitamine:
that's not what I meant, the request to create a separate extract is specifically for the new functionality(ex: [SoundCloudIE for tracks](https://github.com/ytdl-org/youtube-dl/blob/20c50c65566be31607e59d7ca2467c664a29c843/youtube_dl/extractor/soundcloud.py#L58), [SoundCloudPlaylistIE for Playlists](https://github.com/ytdl-org/youtube-dl/blob/20c50c65566be31607e59d7ca2467c664a29c843/youtube_dl/extractor/soundcloud.py#L733)).

youtube_dl/extractor/nhk.py
```diff
@@ -90,4 +66,81 @@ def get_clean_field(key):
...
+
+        if (len(episodes) > 0):
+            return self.playlist_result(
+                [self._parse_episode_json(episode, lang, True)
+                    for episode in episodes],
+                self._get_clean_field(episodes[0], 'pgm_gr_id'), self._get_clean_field(episodes[0], 'title'),
+                self._get_clean_field(episodes[0], 'title'))
+        else:
+            self.report_error('No episodes returned for program with ID: %s' % program_id)
+            return None
```
remitamine, line 146:
```suggestion
            raise ExtractrorError('No episodes returned for program with ID: %s' % program_id, expected=True)
```

youtube_dl/extractor/nhk.py
```diff
@@ -90,4 +66,81 @@ def get_clean_field(key):
...
+        'url': 'https://www3.nhk.or.jp/nhkworld/en/ondemand/program/video/10yearshayaomiyazaki/',
+        'only_matching': True,
+    }]
+
+    def _real_extract(self, url):
+        lang, m_type, program_id = re.match(self._VALID_URL, url).groups()
+
+        episodes = self._list_episodes(lang, 'program/video', program_id)
+
+        if (len(episodes) > 0):
```
remitamine, line 138:
```suggestion
        if episodes:
```

youtube_dl/extractor/nhk.py
```diff
@@ -90,4 +66,81 @@ def get_clean_field(key):
...
+            return self.playlist_result(
+                [self._parse_episode_json(episode, lang, True)
+                    for episode in episodes],
+                self._get_clean_field(episodes[0], 'pgm_gr_id'), self._get_clean_field(episodes[0], 'title'),
+                self._get_clean_field(episodes[0], 'title'))
```
remitamine, line 143: do not use the program title as a description.

youtube_dl/extractor/nhk.py
```diff
@@ -5,55 +5,30 @@
...
-        lang, m_type, episode_id = re.match(self._VALID_URL, url).groups()
-        if episode_id.isdigit():
-            episode_id = episode_id[:4] + '-' + episode_id[4:]
+    def _get_clean_field(self, episode, key):
+        return episode.get(key + '_clean') or episode.get(key)
 
-        is_video = m_type == 'video'
-        episode = self._download_json(
+    def _list_episodes(self, lang, m_type, m_id):
+        is_video = m_type == 'video' or 'program/video'
```
remitamine, line 15: I think you mean:
```suggestion
        is_video = m_type in ('video', 'program/video')
```

remitamine (changes requested):
make sure to check the code with flake8 and run all tests as described in [DEVELOPER INSTRUCTIONS](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)

youtube_dl/extractor/nhk.py
rolledback: @remitamine, If I shouldn't reuse the program title, would hardcoding something like "Episodes of <program title>" be ok? The reason I used the title as the description was because I thought it would be better to only use metadata that came from the source rather than inventing something that didn't exist, as there is no "description" returned from the NHK API.

remitamine: the playlist description is not mandatory, if it's not supplied by the website, then do not fill it.

rolledback:
@remitamine in my email, I saw you left another comment:
> that's not what I meant, the request to create a separate extract is specifically for the new functionality(ex: SoundCloudIE for tracks, SoundCloudPlaylistIE for Playlists).

But I don't see the comment here. Did you delete it? Just making sure I don't miss anything.

remitamine:
after looking at the code, it appeared to me that you just added a `BaseExtractor` without adding a new `NhkProgramIE` extractor, but after checking the code again, I saw the new extractor and removed my comment.

youtube_dl/extractor/nhk.py
```diff
@@ -3,57 +3,35 @@
...
+                'episode' if is_episode else 'program',
+                m_id, lang, '/all' if is_video else ''),
+            m_id, query={'apikey': 'EJfK8jdS57GqlupFgAfAAwr573q01y6k'})['data']['episodes']
+
+    def _parse_episode_json(self, episode, lang, is_video):
         title = episode.get('sub_title_clean') or episode['sub_title']
 
-        def get_clean_field(key):
-            return episode.get(key + '_clean') or episode.get(key)
+        episode_id = episode.get('pgm_id') + episode.get('pgm_no')
```
remitamine, line 32: this will break the extraction if `pgm_id` or `pgm_no` doesn't have a value(None)

youtube_dl/extractor/nhk.py
```diff
@@ -3,57 +3,35 @@
...
-        lang, m_type, episode_id = re.match(self._VALID_URL, url).groups()
-        if episode_id.isdigit():
-            episode_id = episode_id[:4] + '-' + episode_id[4:]
+    def _get_clean_field(self, episode, key):
+        return episode.get(key + '_clean') or episode.get(key)
 
-        is_video = m_type == 'video'
-        episode = self._download_json(
+    def _list_episodes(self, lang, m_type, m_id):
+        is_video = m_type in ('video', 'program/video')
```
remitamine, line 18: I think that should be outside of this method because the value is also recalculated before calling the `_parse_episode_json` method.

remitamine (changes requested):

rolledback:
> after looking at the code, it appeared to me that you just added a BaseExtractor without adding a new NhkProgramIE extractor, but after checking the code again, I saw the new extractor and removed my comment.

Ah ok, that's what I figured. The diff ended up being a little hard to read so I'm not surprised there was a moment of confusion. 😋

youtube_dl/extractor/nhk.py
```diff
@@ -29,48 +26,54 @@ def _list_episodes(self, lang, m_type, m_id):
...
-                'https://nhkworld-vh.akamaihd.net/i%s/master.m3u8' % audio_path,
-                episode_id, 'm4a', entry_protocol='m3u8_native',
-                m3u8_id='hls', fatal=False)
-            for f in info['formats']:
-                f['language'] = lang
+        pgm_id = episode.get('pgm_id')
+        pgm_no = episode.get('pgm_no')
 
-        return info
+        if (pgm_id == None or pgm_no == None):
```
remitamine:
```suggestion
        if not (pgm_id and pgm_no):
```

youtube_dl/extractor/nhk.py
```diff
@@ -29,48 +26,54 @@ def _list_episodes(self, lang, m_type, m_id):
...
+            raise ExtractorError('Cannot download episode: %s' % str(episode), expected=True)
```
remitamine, line 33: use `compat_str` instead of `str`, and I'm not what you would call `str` on a `dict`.

youtube_dl/extractor/nhk.py
```diff
@@ -29,48 +26,54 @@ def _list_episodes(self, lang, m_type, m_id):
...
+        else:
```
remitamine, line 34: there is no need for an `else` block.

youtube_dl/extractor/nhk.py
```diff
@@ -29,48 +26,54 @@ def _list_episodes(self, lang, m_type, m_id):
...
+            episode_id = episode.get('pgm_id') + episode.get('pgm_no')
```
remitamine, line 35: you already have `pgm_id` and `pgm_no` variables now.

remitamine (changes requested):

youtube_dl/extractor/nhk.py
```diff
@@ -3,57 +3,39 @@
...
+    def _parse_episode_json(self, episode, lang, is_video):
         title = episode.get('sub_title_clean') or episode['sub_title']
 
-        def get_clean_field(key):
-            return episode.get(key + '_clean') or episode.get(key)
+        pgm_id = episode.get('pgm_id')
+        pgm_no = episode.get('pgm_no')
+
+        if not (pgm_id and pgm_no):
+            raise ExtractorError('Cannot download episode: %s' % compat_str(episode), expected=True)
```
remitamine, line 34: i think my previous comment wasn't fully addressed, why are calling str/compat_str on a `dict` here?
after resolving this you can squash the commit into a single commit.

rolledback: Oh, I didn't realize you wanted no `str`/`compat_str`, I thought you were just asking for me to use `compat_str`. As for why I'm calling it, well typically when in a handle error code path, I become a little more careful about what I do. But you know this code base better than me, and I don't do python in my day to day, so if you don't think it is needed, I'm totally fine getting rid of it.

rolledback: Ok, `compat_str` gone and squashed down to one commit.

remitamine: we don't directly use `str` in the project, instead we use `compat_str` to keep the code compatible with all support version of Python, when you call `str/compat` on a dict it won't give a good result, so, i think it should be ommited.

youtube_dl/extractor/nhk.py
```diff
@@ -3,57 +3,38 @@
...
+    def _parse_episode_json(self, episode, lang, is_video):
         title = episode.get('sub_title_clean') or episode['sub_title']
 
-        def get_clean_field(key):
-            return episode.get(key + '_clean') or episode.get(key)
+        pgm_id = episode.get('pgm_id')
+        pgm_no = episode.get('pgm_no')
+
+        if not (pgm_id and pgm_no):
+            raise ExtractorError('Cannot download episode: %s' % episode, expected=True)
```
remitamine, line 33: you have to test the code, and see if the result is usefull, thos code with result in error message similar to this:
```
Cannot download episode: {'data': {'episodes': [{'pgm_gr_id': 'dwc', 'image': '/nhkworld/en/ondemand/video/9999011/images/buCOFSNvcN0ctbWvHm25sLdGTOKRUX32w7YLHHTS.jpeg', 'image_l': '/nhkworld/en/ondemand/video/9999011/images/dYFYZODb5cmlf6efJ9Cgb4bWIouBc6M83TEhRMFZ.jpeg', 'image_promo': '', 'voice_lang': 'en', 'caption_langs': [], 'voice_langs': ['en'], 'vod_id': 'lpZXIwaDE6_Z-976CPsFdxyICyWUzlT5', 'movie_lengh': '2:28', 'movie_duration': 148, 'analytics': "[nhkworld]clip;Dining with the Chef_Chef Saito's Family recipe: MENCHI-KATSU;en,001;9999-011-1970;", 'title': 'Dining with the Chef', 'title_clean': 'Dining with the Chef', 'sub_title': "Chef Saito's Family recipe: MENCHI-KATSU", 'sub_title_clean': "Chef Saito's Family recipe: MENCHI-KATSU", 'description': 'Umami juice splashes out from crunchy Panko batter! Menchi-katsu is a mouthwatering deep fried meat ball from Japan. [From "Dining with the Chef"]', 'description_clean': 'Umami juice splashes out from crunchy Panko batter! Menchi-katsu is a mouthwatering deep fried meat ball from Japan. [From "Dining with the Chef"]', 'url': '/nhkworld/en/ondemand/video/9999011/', 'category': [17], 'mostwatch_ranking': 136, 'related_episodes': [], 'tags': ['washoku'], 'chapter_list': [], 'life': 0, 'life_category': [], 'promotion': []}]}}
```
the error message is meant to be readable and usefull to the user, that is the point that i'm try pass.

rolledback: See latest revision. Does that seem more helpful?

remitamine: > Does that seem more helpful? If this error was to actually happen, it would mean that what the API returns has changed beyond what we expect. When the happens, I think all the user can do is report an issue here.

there is a generic option to do this(`--dump-pages`), and would ask to the user to provide this output if needed.

> Is it typical practice to say in error messages "report this error to youtube-dl"? If so I can add that as well.

you can do that, by removing `expected=True` from the options passed to `ExtractorError`,

rolledback: @remitamine, ok, expected is now false. Thanks for teaching me that. Also, I am choosing to leave the JSON in because if the user does report the error, I think it'd be helpful here to know what the API returned in case it is hard to reproduce.

rolledback: Oh I see that --dump-pages will print the API response. I guess that could mean I can remove the JSON from the string as well. Let me know what is considered best practice for youtube-dl and I'll go with that.

remitamine: > expected is now false

no need to type `expected=False`, it's the default.

> Oh I see what you mean about --dump-pages, it will print the API response. I guess that could mean I can remove the JSON from the string as well.

Exactly.

rolledback: 😊

remitamine:
checking the code again, it breaks the extraction for audio URLs.
```
python __main__.py -F https://www3.nhk.or.jp/nhkworld/en/ondemand/audio/j_art-20190711-1/
[NhkVod] j_art-20190711-1: Downloading JSON metadata
ERROR: Cannot download episode. Field pgm_id is missing from episode JSON.; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

rolledback:
Ok, I'll add an non-only_matching test for audio URLs, and get things working again for that.

rolledback:
Audio clips are working again.

youtube_dl/extractor/extractors.py
```diff
@@ -726,6 +726,7 @@
 )
 from .nfl import NFLIE
 from .nhk import NhkVodIE
+from .nhk import NhkVodProgramIE
```
remitamine, line 729:
```suggestion
from .nhk import (
    NhkVodIE,
    NhkVodProgramIE,
)
```

youtube_dl/extractor/nhk.py
```diff
@@ -3,14 +3,98 @@
 import re
 
 from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+)
```
remitamine, line 8:
```suggestion
from ..utils import ExtractorError
```

youtube_dl/extractor/nhk.py
```diff
@@ -3,14 +3,98 @@
...
 
 
-class NhkVodIE(InfoExtractor):
+class NhkBaseInfoExtractor(InfoExtractor):
```
remitamine, line 11:
```suggestion
class NhkBaseIE(InfoExtractor):
```

youtube_dl/extractor/nhk.py
```diff
@@ -3,14 +3,98 @@
...
+    _API_URL_TEMPLATE = 'https://api.nhk.or.jp/nhkworld/%sod%slist/v7a/%s/%s/%s/all%s.json'
+
+    def _get_clean_field(self, episode, key):
+        return episode.get(key + '_clean') or episode.get(key)
+
+    def _list_episodes(self, lang, is_video, is_episode, m_id):
```
remitamine, line 17: it would be better to order the function parameter based on their importance(i think the most important one is `m_id`).

remitamine (changes requested):
for the rest of the changes, I would take a more minimalistic approach based on your changes while adding more features.

rolledback:
Thanks for the review (and the patience)! I'll keep an eye out for any reports of issues with nhk in case this change causes any unforeseen problems.

PR #27247, opened 2020-11-29
Replace all os.remove & os.rename combos with single os.replace (Fix deletion of post-processed videos)

digitalsignalperson:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This pull request replaces the pattern of os.remove(original filename) then os.rename(temp filename, original filename) to single atomic os.replace(temp_filename, original filename).

The problem I was having was that certain files would download and I'd see a "[ffmpeg] Fixing malformed AAC bitstream in " message. In my file explorer (Dolphin) I could see the thumbnail of the original file appear and disappear at this operation. Youtube-dl would be done processing, but the end result was the downloaded file kept getting deleted!

After digging around, I found that the combo of os.remove followed by os.rename deleted both the original file, as well as the renamed file! This may be particular to my OS/filesystem (Arch Linux, downloading to a ZFS cifs/samba share). This PR fixes the problem and now my downloads now don't disappear. 

I don't fully understand the reason for the bug (atomic operations, race conditions...) but I found a similar bug here https://github.com/zarr-developers/zarr-python/issues/263

> Possible solutions:
> - Execute os.remove(file_path) on Windows only.
> - Use os.replace() instead of os.rename. This breaks Python 2.7 compatibility, though.

### DIFF
--- a/youtube_dl/postprocessor/embedthumbnail.py
+++ b/youtube_dl/postprocessor/embedthumbnail.py
@@ -85,8 +85,7 @@ class EmbedThumbnailPP(FFmpegPostProcessor):
 
             if not self._already_have_thumbnail:
                 os.remove(encodeFilename(thumbnail_filename))
-            os.remove(encodeFilename(filename))
-            os.rename(encodeFilename(temp_filename), encodeFilename(filename))
+            os.replace(encodeFilename(temp_filename), encodeFilename(filename))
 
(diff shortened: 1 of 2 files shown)


dstftw:
1. The code must work on python 2.6+ and 3.2+.
2. You must fix the core reason (bug in your FS implementation or whatsoever) not the consequences.

Issue #23035, opened 2019-11-10
match-filter with wildcard/regex


hollysheep:
I can download all videos of a playlist when i use the complete plalyist name in the match-filter command

`--match-filter "playlist_title = 'playlist name'"`

But when i only use a part of the title it doesn't work. Aren't wirdcard/regex not supported here? This can be useful when a new playlist with a slightly different playlist name is released from time to time. I want to download all the champions league hightlights from skyhd channel.

Issue #9092, opened 2016-04-06
--match-filter does not work on the "description" key


sebma:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like that [x])
- Use *Preview* tab to see how your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2016.04.05*. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2016.04.05**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add `-v` flag to **your command line** you run youtube-dl with, copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):
```
$ youtube-dl -v <your command line>
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://www.youtube.com/watch?v=BaW_jenozKcj']
[debug] Encodings: locale cp1251, fs mbcs, out cp866, pref cp1251
[debug] youtube-dl version 2016.04.05
[debug] Python version 2.7.11 - Windows-2003Server-5.2.3790-SP2
[debug] exe versions: ffmpeg N-75573-g1d0487f, ffprobe N-75573-g1d0487f, rtmpdump 2.4
[debug] Proxy map: {}
...
<end of log>
```

---

[...]

PR #27361, opened 2020-12-09
More string comparison operators for --match-filter

max-te:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This pull request allows to use string containment and regex match operators, introducing the operators `~=, !~=, *=` and `~*=`. Though regex support alone can handle all cases, the other additional operators may be more comfortable for users who are unfamiliar with regexes and also avoid having to escape certain characters. This PR adresses issues #23035 and #9092,

This PR also stops disallowing the use of `<=` and `>=` for strings. They are lexicographic comparison, this allows, e.g., to download videos in a playlist in alphabetic blocks A-F,G-K, etc. I don't have a specific use case for this, but it seems silly to disallow it when it requires no additional code.


dstftw (changes requested):
1. Operators must match exactly with operators used in format selection.
2. Negation must be separate operator.
3. Everything must be covered by tests.

max-te:
>     1. Operators must match exactly with operators used in format selection.
That is totally reasonable. I had failed to notice that format selection syntax already has a rich set of operators.
It doesn't have a regex-operator, though. What are your thoughts on that? Should we have one? If so, what should it be called? `*=` is taken for containment, would `~=` be fine?

I will take a look at how the operators for format selection are implemented and go from there.

dstftw:
The operator alias is discussable and secondary, it may be `~`/`!~` like in awk, or `=~`/`!~` like in perl, or suggested `~=`/`!~=` variant. What is important is to design its behavior, whether it will be used only in quoted variant or not (in this case there is an ambiguity appears immediately since regex may contain arbitrary characters), whether it will allow using captured groups in following clauses and so on.

max-te:
I was trying to make only minimal changes to the matching code in this PR. Maybe we should only do the `^=, *=, $=` operators as in format matching for this PR and have regexes as a second step. With the last commit I pushed, we have only the additional operators from format selection, and I added tests for each of them. Negation is now a separate operator.

Carrying over capture groups is interesting but not a standard feature of regexes and I don't believe it is supported by the `re` module.

A `/pattern/flags` syntax would require a bit of remodeling of `operator_rex` and the matching function in general. (I'm not sure which flags apart from `i` you'd realistically need, though.) The current quoting code handles escaped quotes, the only ambiguity comes from `&` characters, which already pose a problem for `=`, since the `match-filter` argument is split at all `&`s indiscriminantly.

PR #27362, opened 2020-12-09
[rbgtum] Add support for live.rbg.tum.de lecture platform

main--:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This pull request adds support for the https://live.rbg.tum.de/ lecture streaming site. There seems to be no identifier other than the page URL which is why I am using that.

### DIFF
--- a/youtube_dl/extractor/rbgtum.py
+++ b/youtube_dl/extractor/rbgtum.py
@@ -1,0 +1,29 @@
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class RbgTumIE(InfoExtractor):
+    _VALID_URL = r'https://live.rbg.tum.de/(?P<id>.*)'
+    _TESTS = [{
+        'url': 'https://live.rbg.tum.de/cgi-bin/streams/VOD/WiSe2021AutUFormSpr/2020_12_07_10_00/COMB',
+        'md5': '12de44b7ad4ef491c1cc952fd4a0adc9',
+        'info_dict': {
+            'id': 'cgi-bin/streams/VOD/WiSe2021AutUFormSpr/2020_12_07_10_00/COMB',
+            'ext': 'm3u8',
+            'title': 'Aufzeichnung aus IRH102'
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/rbgtum.py
```diff
@@ -0,0 +1,29 @@
...
+from .common import InfoExtractor
+
+
+class RbgTumIE(InfoExtractor):
+    _VALID_URL = r'https://live.rbg.tum.de/(?P<id>.*)'
+    _TESTS = [{
+        'url': 'https://live.rbg.tum.de/cgi-bin/streams/VOD/WiSe2021AutUFormSpr/2020_12_07_10_00/COMB',
+        'md5': '12de44b7ad4ef491c1cc952fd4a0adc9',
+        'info_dict': {
+            'id': 'cgi-bin/streams/VOD/WiSe2021AutUFormSpr/2020_12_07_10_00/COMB',
```
dstftw, line 12: `id` must not contain any irrelevant parts.

youtube_dl/extractor/rbgtum.py
```diff
@@ -0,0 +1,29 @@
...
+            'ext': 'm3u8',
+            'title': 'Aufzeichnung aus IRH102'
+        }
+    }]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        hlsurl = self._html_search_regex('MMstartStream\\( document\\.getElementById\\("video1"\\) , \'(.*)\' \\);', webpage, 'hlsurl')
```
dstftw, line 22: 1. Relax regex.
2. Do not capture empty string.

dstftw (changes requested):

main--:
Thank you for the fast review! I changed the regex accordingly. I am unsure how to deal with `id` though. While it looks like there may be a structure like `https://live.rbg.tum.de/cgi-bin/streams/VOD/{lecture}/{date}/COMB`, this structure is not strictly adhered to. For example, `https://live.rbg.tum.de/GC2020/` was in use earlier today. The HTML contains no ID whatsoever. This is why I chose to simply set the entire URL as `id`. Is there maybe a way to implement an extractor with no `id` at all?

PR #27391, opened 2020-12-12
[ruutu] Add extractor for hs.fi

tpikonen:

- [x] New extractor

---

Add extractors for videos in articles at hs.fi, the largest newspaper in the nordics. The videos are hosted in ruutu.fi and use the same ids.


dstftw:
You must delegate to ruutu not inherit it.

tpikonen:
Do you mean something like this (in the updated PR)?

dstftw:
No, use search this has been explained numerous times. Also read coding conventions.

tpikonen:
Ok, I think I got it now. 

Playlist for multiple videos per page will happen when I find a test case for one (I had one when I wrote this, but not any more...)

The second commit fixes Ruutu tests which were failing because of missing upload_date values.

PR #27405, opened 2020-12-13
[Generic] Fix RSS item thumbnail URL

renalid:
The thumbnail url of an item in RSS feed is in "href" attribute (not directly in the itunes:image tag)

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

The thumbnail url of an item in RSS feed is in "href" attribute (not directly in the itunes:image tag)

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -238,6 +238,9 @@ class GenericIE(InfoExtractor):
                     'thumbnail': r're:^https?://.*\.jpg$',
                 },
             }],
+            'params': {
+                'skip_download': True,
+            },
         },
         # RSS feed with enclosures and unsupported link URLs
         {
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -209,11 +209,11 @@ class GenericIE(InfoExtractor):
             'playlist': [{
                 'info_dict': {
                     'ext': 'mov',
-                    'id': 'pdv_maddow_netcast_mov-12-11-2020-225421',
+                    'id': 'pdv_maddow_netcast_mov-12-04-2020-224335',
                     'title': 're:MSNBC Rachel Maddow',
                     'description': 're:.*her unique approach to storytelling.*',
                     'timestamp': int,
-                    'upload_date': '20201211',
+                    'upload_date': compat_str,
                     'duration': float,
                 },
             }],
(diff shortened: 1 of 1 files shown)


dstftw:
Add a test.

youtube_dl/extractor/generic.py
```diff
@@ -208,12 +209,13 @@ class GenericIE(InfoExtractor):
...
                     'ext': 'mov',
-                    'id': 'pdv_maddow_netcast_mov-12-04-2020-224335',
+                    'id': 'pdv_maddow_netcast_mov-12-11-2020-225421',
                     'title': 're:MSNBC Rachel Maddow',
                     'description': 're:.*her unique approach to storytelling.*',
                     'timestamp': int,
-                    'upload_date': compat_str,
+                    'upload_date': '20201211',
                     'duration': float,
+                    'thumbnail': None,
```
dstftw, line 218: This make zero sense. You must add a test that will test the changes made.

dstftw (changes requested):

youtube_dl/extractor/generic.py
renalid: @dstftw yes, you are right. Can I add another RSS feed, because this one don't get thumbnail image ?
Or I can update this one http://www.hellointernet.fm/podcast?format=rss ?

dstftw: You can use any existing RSS test with thumbnail available or add a new if there are no such tests.

renalid:
@dstftw I added a new test. Thanks :-)

youtube_dl/extractor/generic.py
```diff
@@ -208,15 +209,36 @@ class GenericIE(InfoExtractor):
             'playlist': [{
                 'info_dict': {
                     'ext': 'mov',
-                    'id': 'pdv_maddow_netcast_mov-12-04-2020-224335',
+                    'id': 'pdv_maddow_netcast_mov-12-11-2020-225421',
                     'title': 're:MSNBC Rachel Maddow',
                     'description': 're:.*her unique approach to storytelling.*',
                     'timestamp': int,
-                    'upload_date': compat_str,
+                    'upload_date': '20201211',
```
dstftw, line 216: Do not change this.

dstftw (changes requested):

youtube_dl/extractor/generic.py
renalid: I rollbacked this issue

renalid:
@dstftw requested change done

Issue #27455, opened 2020-12-16
youtube-dl not working on StreetVoice


danny-laa:
## Checklist

- [x] I'm reporting a broken site support
- [x] I've verified that I'm running youtube-dl version **2020.12.14**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
$ youtube-dl https://streetvoice.com/lekimberleyy/songs/630281/ --verbose
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['https://streetvoice.com/lekimberleyy/songs/630281/', '--verbose']
[debug] Encodings: locale cp1252, fs mbcs, out cp437, pref cp1252
[debug] youtube-dl version 2020.12.14
[debug] Python version 3.4.4 (CPython) - Windows-10-10.0.19041
[debug] exe versions: ffmpeg 4.2
[debug] Proxy map: {}
[StreetVoice] 630281: Downloading JSON metadata
ERROR: Unable to download JSON metadata: HTTP Error 404: Not Found (caused by HTTPError()); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpwx5ikkp6\build\youtube_dl\extractor\common.py", line 632, in _request_webpage
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpwx5ikkp6\build\youtube_dl\YoutubeDL.py", line 2248, in urlopen
  File "C:\Python\Python34\lib\urllib\request.py", line 470, in open
  File "C:\Python\Python34\lib\urllib\request.py", line 580, in http_response
  File "C:\Python\Python34\lib\urllib\request.py", line 508, in error
  File "C:\Python\Python34\lib\urllib\request.py", line 442, in _call_chain
  File "C:\Python\Python34\lib\urllib\request.py", line 588, in http_error_default
```

## Description

404 HTTP Error when downloading JSON metadata

PR #27492, opened 2020-12-20
Fix for #27455. Adapted StreetVoice extractor to new API and fixed tests

PaaaulZ:
flake8

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Fix for issue #27455 . StreetVoice API changed and support for that site broke, I also fixed the test that pointed to an invalid url because the song was removed.

### DIFF
--- a/youtube_dl/extractor/streetvoice.py
+++ b/youtube_dl/extractor/streetvoice.py
@@ -3,27 +3,28 @@ from __future__ import unicode_literals
 
 from .common import InfoExtractor
 from ..compat import compat_str
-from ..utils import unified_strdate
+from ..utils import int_or_none, unified_timestamp
 
 
 class StreetVoiceIE(InfoExtractor):
     _VALID_URL = r'https?://(?:.+?\.)?streetvoice\.com/[^/]+/songs/(?P<id>[0-9]+)'
     _TESTS = [{
-        'url': 'http://streetvoice.com/skippylu/songs/94440/',
-        'md5': '15974627fc01a29e492c98593c2fd472',
+        'url': 'https://streetvoice.com/lekimberleyy/songs/630281/',
+        'md5': '515d4bfdfe337039d9efaa53be24484d',
         'info_dict': {
-            'id': '94440',
+            'id': '630281',
             'ext': 'mp3',
-            'title': '輸',
-            'description': 'Crispy脆樂團 - 輸',
-            'thumbnail': r're:^https?://.*\.jpg$',
-            'duration': 260,
-            'upload_date': '20091018',
-            'uploader': 'Crispy脆樂團',
-            'uploader_id': '627810',
+            'title': '郵票Good Times feat. 瘦子E.SO',
+            'description': 'Kimberley - 郵票Good Times feat. 瘦子E.SO',
+            'thumbnail': r're:^https?://.*\.jpg',
+            'uploader': 'lekimberleyy',
+            'uploader_id': '2674557',
+            'duration': 236,
+            'upload_date': '20201216',
+            'timestamp': 1608110177
         }
     }, {
-        'url': 'http://tw.streetvoice.com/skippylu/songs/94440/',
+        'url': 'https://tw.streetvoice.com/ConstantanChange/songs/628471/',
         'only_matching': True,
     }]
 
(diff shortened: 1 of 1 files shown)


remitamine:
> Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them

PaaaulZ:
I did check the guidelines before submitting the PR, considering that I changed as little as possible from the original code maybe I'm missing something. 
Could you point me to a section of the guidelines I did not follow or to where the code does not meet requirements so I can work on it? I'll check again tomorrow and fix it, thanks

remitamine:
Mainly, the [Mandatory and optional metafields](https://github.com/ytdl-org/youtube-dl#mandatory-and-optional-metafields) section, the extraction will break if optional fields are not available.

PaaaulZ:
Damn you are right I'm sorry, I'll fix it later and squash the commits

youtube_dl/extractor/streetvoice.py
```diff
@@ -3,47 +3,50 @@
...
-        'url': 'http://tw.streetvoice.com/skippylu/songs/94440/',
+        'url': 'https://tw.streetvoice.com/ConstantanChange/songs/628471/',
         'only_matching': True,
     }]
 
     def _real_extract(self, url):
         song_id = self._match_id(url)
 
+        song_details = self._download_json(
+            'https://streetvoice.com/api/v4/song/%s/?fields=id,name,image,likes_count,is_like,user,profile,nickname,type,public,username' % song_id, song_id, note="Downloading song details")
```
remitamine, line 31: use `query` argument.

youtube_dl/extractor/streetvoice.py
```diff
@@ -3,47 +3,50 @@
...
+
         song = self._download_json(
-            'https://streetvoice.com/api/v1/public/song/%s/' % song_id, song_id, data=b'')
+            'https://streetvoice.com/api/v4/song/%s/hls/file/' % song_id, song_id, data=b'', note="Downloading song data")
```
remitamine, line 34: extract all formats.

youtube_dl/extractor/streetvoice.py
```diff
@@ -3,47 +3,50 @@
...
-            'url': song['file'],
             'title': title,
             'description': '%s - %s' % (author, title),
-            'thumbnail': self._proto_relative_url(song.get('image'), 'http:'),
-            'duration': song.get('length'),
-            'upload_date': unified_strdate(song.get('created_at')),
-            'uploader': author,
-            'uploader_id': compat_str(song['user']['id']),
+            'thumbnail': self._proto_relative_url(song_details.get('image'), 'http:'),
+            'uploader': uploader_data.get('username'),
```
remitamine, line 49: still fails if `uploader_data` not available.

youtube_dl/extractor/streetvoice.py
```diff
@@ -3,47 +3,50 @@
...
+            song['file'], song_id, 'mp3', m3u8_id='hls'))
 
         return {
             'id': song_id,
-            'url': song['file'],
             'title': title,
             'description': '%s - %s' % (author, title),
-            'thumbnail': self._proto_relative_url(song.get('image'), 'http:'),
-            'duration': song.get('length'),
-            'upload_date': unified_strdate(song.get('created_at')),
```
remitamine, line 46: `created_at` still available in the API response, extract `timestamp` field from it.
and also there more metadata that can be extracted (`view_count`, `album`, etc...).

youtube_dl/extractor/streetvoice.py
```diff
@@ -3,47 +3,50 @@
...
     }, {
-        'url': 'http://tw.streetvoice.com/skippylu/songs/94440/',
+        'url': 'https://tw.streetvoice.com/ConstantanChange/songs/628471/',
         'only_matching': True,
     }]
 
     def _real_extract(self, url):
         song_id = self._match_id(url)
 
+        song_details = self._download_json(
```
remitamine, line 30: i think it would be better to keep the old variable name(`song`) as this response contain most of the data available in the old `API` response.

youtube_dl/extractor/streetvoice.py
```diff
@@ -3,47 +3,50 @@
...
+            'https://streetvoice.com/api/v4/song/%s/?fields=id,name,image,likes_count,is_like,user,profile,nickname,type,public,username' % song_id, song_id, note="Downloading song details")
+
         song = self._download_json(
-            'https://streetvoice.com/api/v1/public/song/%s/' % song_id, song_id, data=b'')
+            'https://streetvoice.com/api/v4/song/%s/hls/file/' % song_id, song_id, data=b'', note="Downloading song data")
+
+        uploader_data = song_details.get('user')
+        author = uploader_data.get('profile').get('nickname')
```
remitamine, line 37: fails if `profile` is not available.

youtube_dl/extractor/streetvoice.py
```diff
@@ -3,47 +3,50 @@
...
+        title = song_details['name']
 
-        title = song['name']
-        author = song['user']['nickname']
+        formats = []
+        formats.extend(self._extract_m3u8_formats(
```
remitamine, line 41: no need to call `_extract_m3u8_formats` method for HLS Media manifests.

remitamine (changes requested):
i might not be able to continue the review in a couple of days, will try to check it again when i can.

youtube_dl/extractor/streetvoice.py
PaaaulZ: Since I needed to add other metadata I removed the query parameters in the new commit so the API would send me all the data, most of it gets used so it's not really a waste of resources.

youtube_dl/extractor/streetvoice.py
PaaaulZ: I don't see any other formats and the json response isn't even an array I can iterate, seems like they only have one format even in the old api

PR #21962, opened 2019-07-31
[sproutvideo] Add new extractor (closes #7935)

TheZ3ro:
### Before submitting a *pull request* make sure you have:
- [x] At least skimmed through [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

*Sorry in advance for the long description*
This PR adds support for SproutVideo and every website/service that use it as a video provider.
This PR closes #7935, #16994, #16996 and #21333. 
As @remitamine explained in this comment https://github.com/ytdl-org/youtube-dl/issues/16996#issuecomment-406901324, sproutvideo uses 3 type of CloudFront policies/signatures, one for M3U8 manifests, one for segments and one for the HLS encryption key.

At the time of writing ffmpeg and FFmpegDL don't support different URL parameters for key, segment and manifests so I've added a `force_hlsdl` parameter in the `InfoExtractor` class so extractors can forse the use of the native hls downloader.

I've added SproutVideo inside the generic extractor since it can be embedded inside other websites with iframes (and most of the time sproutvideo's videos are distributed this way).

I've edited the HLS native downloader to support the new `extra_param_to_key_url` parameter in the cases where URL parameters for segments and key are different (like sproutvideo).
In the case that `extra_param_to_key_url` is not declared, `extra_param_to_segment_url` will be used so this change is backward compatible and should not affect other downloaders/extractors

Finally I've edited the HTTP downloader with a new `hls` parameter that will be passed by the HLS native downloader so the HTTP one has information about the encrypted stream.
[...]

PR #27620, opened 2021-01-02
Pad fragment data to the appropriate AES blocksize

kevinoconnor7:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Pads fragment data out to the AES-128 block size of 16 using PKCS#7 padding. This generally shouldn't be an issue for real downloads unless we've gotten truncated data somehow, but for tests the data is trimmed to exactly 10241 bytes and `10241 % 16 == 1` so we end with a malformed trailing block. This makes it impossible to run with `--test` or have extractor unit tests pass when the data being processed is encrypted with AES-128. Without this change you'll end up with an error like:

```
  File "/Users/kevinoconnor/Development/youtube-dl/youtube_dl/downloader/hls.py", line 182, in real_download
    decrypt_info['KEY'], AES.MODE_CBC, iv).decrypt(frag_content)
  File "/usr/local/lib/python3.7/site-packages/Crypto/Cipher/blockalgo.py", line 295, in decrypt
    return self._cipher.decrypt(ciphertext)
ValueError: Input strings must be a multiple of 16 in length
```

I didn't add explicit new tests for this as there's currently no unit tests for this file and (in theory) existing extractor tests will exercise this code (ex. #27621 tests only pass with this change).

### DIFF
--- a/youtube_dl/downloader/hls.py
+++ b/youtube_dl/downloader/hls.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 import re
+import struct
 import binascii
 try:
     from Crypto.Cipher import AES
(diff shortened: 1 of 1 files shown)


dstftw:
That's not acceptable. Data being downloaded must stay intact. What's really should be done in this case is disabling decryption while in test. There is a WIP PR #21962 with such changes.

kevinoconnor7:
I see two issues with those suggestions: the former doesn't actually verify that the decryption key is correct in tests, and the latter just punts on test coverage entirely.

Alternatively, I could amend this change to only pad for when `test` is `true`. Since it's the trailing block it shouldn't matter if we pad it out, especially if we do it consistently for tests.

dstftw:
This main point of md5 file tests is to test that downloaded data matches not to test how decryption works.

kevinoconnor7:
Fair enough. Given that #21962 seems to have stalled I'll split of that specific portion into a separate PR.

PR #27621, opened 2021-01-02
Add a PlutoTV extractor for on-demand content

### LINKED ISSUE #25626: Pluto TV on demand support
## Checklist

- [X ] I'm reporting a new site support request
- [X ] I've verified that I'm running youtube-dl version **2020.06.06**
- [X ] I've checked that all provided URLs are alive and playable in a browser
- [X ] I've checked that none of provided URLs violate any copyrights
- [X ] I've searched the bugtracker for similar site support requests including closed ones

## Example URLs

- Single video: https://pluto.tv/on-demand/series/engineering-connections/season/3/episode/burj-al-arab-2011-3-5
- Playlist: https://service-stitcher.clusters.pluto.tv/stitch/hls/episode/5ecd5f1c86f709001b51ad50/master.m3u8?advertisingId=&appName=web&appStoreUrl=&appVersion=5.4.0-03751b61882e52dadae8597d2215c0142b039e02&architecture=&buildVersion=&deviceDNT=0&deviceId=8f71fdc7-a160-4ba5-b37b-006ecbbda96a&deviceLat=39.095963&deviceLon=-77.329431
&deviceMake=Chrome&deviceModel=Chrome&deviceType=web&deviceVersion=83.0.4103.97&includeExtendedEvents=false&marketingRegion=US&sid=79371dcf-5c38-4368-9229-949350682341&userId=&serverSideAds=true

Still poking around so not positive how much of that playlist URL is actually necessary.

## Description

Currently youtube-dl is doing with Pluto TV on demand streams very similar to issue #15748 in that the in-stream ads mess up (choppy transitions around the ad insertion point, out of sync audio etc.) the final video.

Unlike the #15748 issue, I haven't found anything in the master.m3u8 that stands-out as an ad. 

[master-m3u8.txt](https://github.com/ytdl-org/youtube-dl/files/4766105/master-m3u8.txt)

I do notice while capturing the output that ad insertion happens like this:

[hls,applehttp @ 0xa4f760] Opening 'crypto+https://siloh-l3.plutotv.net/c6009f_p
luto/clip/5df29776642e4b9328052d27_Engineering_Connections_Burj_al_Arab_S3E5/720
p/20191212_113935/hls/0-516520/hls_2400-00102.ts' for reading
frame=15300 fps=187 q=-1.0 size=  151588kB time=00:08:29.99 bitrate=2434.9kbits/
[hls,applehttp @ 0xa4f760] Opening 'crypto+https://siloh-l3.plutotv.net/c6009f_p
luto/clip/5df29776642e4b9328052d27_Engineering_Connections_Burj_al_Arab_S3E5/720
p/20191212_113935/hls/0-516520/hls_2400-00103.ts' for reading
frame=15450 fps=187 q=-1.0 size=  153040kB time=00:08:35.00 bitrate=2434.3kbits/
[hls,applehttp @ 0xa4f760] Opening 'https://siloh-l3.plutotv.net/248093_59_ad/cr
eative/5ec5d94718e30b000b944683_ad/720p/20200520_182839/hls/hls_2400_keyfile_0.k
ey' for reading
[hls,applehttp @ 0xa4f760] Opening 'crypto+https://siloh-l3.plutotv.net/248093_5
9_ad/creative/5ec5d94718e30b000b944683_ad/720p/20200520_182839/hls/hls_2400-0000
0.ts' for reading
frame=15497 fps=185 q=-1.0 size=  153592kB time=00:08:36.50 bitrate=2436.1kbits/
[mp4 @ 0xdc1920] Non-monotonous DTS in output stream 0:0; previous: 46482000, cu
rrent: -6000; changing to 46482001. This may result in incorrect timestamps in t
he output file.

All of the in-stream ads seem to have a URL that contains "_ad" and "creative" while the fragment file seems so far to be ending with -00000.ts (crypto+https://siloh-l3.plutotv.net/**248093_5
9_ad/creative/5ec5d94718e30b000b944683_ad**/720p/20200520_182839/hls/**hls_2400-0000
0.ts**)

[youtube-dlout_ec.txt](https://github.com/ytdl-org/youtube-dl/files/4766152/youtube-dlout_ec.txt)

Thanks!

kevinoconnor7:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Adds an extractor for pluto.tv on-demand content supporting both movies and series. Series can be downloaded in their entirety or per season or episode, depending upon the scope of the URL provided.

Ads are currently retained in the downloads as there's no distinctive markers in the playlist. To prune them out we'll need to pipe in a URL blocklist regex to have it skip those fragments. Since that's a more invasive change to the native HLS downloader I have omitted them from this change.

Note: the tests won't pass until #27620 is merged, but the extractor works in non-test mode without it.

Fixes #25626.


youtube_dl/extractor/plutotv.py
```diff
@@ -0,0 +1,136 @@
...
+            if episode_slug is not None and len(videos) == 1:
+                return videos[0]
+            playlist_title = series_name
+            if season_number is not None:
+                playlist_title += ' - Season %d' % season_number
+            return self.playlist_result(videos,
+                                        playlist_id=video_json.get('_id', info_slug),
+                                        playlist_title=playlist_title)
+        else:
+            raise ExtractorError('Unsupported url')
```
fossdd, line 136: Does youtube-dl passes to this location? You already set here the supported url:

https://github.com/ytdl-org/youtube-dl/blob/b74e1295cfbd5c0a2d825053033af65285804246/youtube_dl/extractor/plutotv.py#L17

kevinoconnor7: Currently it should be an impossible state, but I figured I would add the extra guard here should that change in the future. At least you'll get an explicit error rather than some other obscure error (since the data won't be what we expect) or malformed data.

fossdd: Hmm, okay. Not the best soulution but it's also implemented in other extractores

kevinoconnor7: Is there a pattern you prefer? I'm open to discuss alternatives.

dstftw: This must be assert not exception.

kevinoconnor7: reworked it a bit. Done!

kevinoconnor7:
Rebased and updated the md5 tests, everything now pass that #27660 has been merged.

kevinoconnor7:
Added another commit to handle removal of ads. It ended up not being too complex, but I did need to do it via an extra attribute in the `info_dict`. @dstftw do you have any concerns on how I wired this up?

Danker1990:
how do you use pluto tv to download content.

dstftw:
1. New downloader options should be under `downloader_options` key and should be documented.
2. Option name should be generic, e.g. `fragment_skip_filters` or so.
3. It should apply to all fragment downloaders.

Danker1990:
> 1. New downloader options should be under `downloader_options` key and should be documented.
> 2. Option name should be generic, e.g. `fragment_skip_filters` or so.
> 3. It should apply to all fragment downloaders.

could you do a guide as im new to this?

kevinoconnor7:
@dstftw Done! Unfortunately there's not a great place to hook this since the subtypes of `FragmentFD` delegate responsibility back and forth. I opted to just leave it as the responsibility of the subtypes to check that if it should be skipped at the appropriate time.

I considered hook it into `_download_fragment` and treating it as a successful but empty download, however, subtypes may ends up processing that result (ex. the hls downloader may try to decrypt it) and barfing on that case, requiring explicit handling of this case anyway.

The only other weirdness is that I didn't wire up a command line option for this. I'm not sure how consequential that is, or if you have any strong feeling on this. The only other usage of `downloader_options` is for an option that also happens to be a command line option, so it's unclear if that intentional or not.

Finally, let me know if you think we should be logging these skipped fragments or not. I left it in for now for discussion.

kevinoconnor7:
@Danker1990 There's currently no support for PlutoTV in YoutubeDL (see #25626). This PR is adding support for it, but it's still a WIP. You could apply the PR locally yourself if you know how to test it out, otherwise you're better off waiting for this to be merged and for an official release to follow.

youtube_dl/downloader/fragment.py
```diff
@@ -65,6 +73,13 @@ def _prepare_url(self, info_dict, url):
         headers = info_dict.get('http_headers')
         return sanitized_Request(url, None, headers) if headers else url
 
+    def _should_skip_fragment(self, info_dict, url):
+        regexes = info_dict.get('downloader_options', {}).get('fragment_skip_filters', [])
+        should_skip = any([re.match(regex, url) for regex in regexes])
```
dstftw, line 78: `re.search` should be used instead as one may need partial matches.

youtube_dl/downloader/fragment.py
```diff
@@ -65,6 +73,13 @@ def _prepare_url(self, info_dict, url):
...
+        if should_skip:
+            self.to_screen('[download] Skipping filtered fragment: %s' % url)
```
dstftw, line 80: For now this should not be printed or only printed in `verbose` mode.

youtube_dl/downloader/ism.py
```diff
@@ -228,7 +228,7 @@ def real_download(self, filename, info_dict):
         frag_index = 0
         for i, segment in enumerate(segments):
             frag_index += 1
-            if frag_index <= ctx['fragment_index']:
+            if frag_index <= ctx['fragment_index'] or self._should_skip_fragment(info_dict, segment['url']):
```
dstftw, line 231: Split into two `if` statements for uniformity with other places in code and better readability.

dstftw (changes requested):
Could you also extract fragment skip filters feature into a separate PR?

remitamine:
just a note, it is possible to extract a raw m3u8 playlist that doesn't contain the Ads, example for the first test case:
`https://siloh.pluto.tv/9d6e28_Endemol_Shine/clip/5de6c593e9379ae4912def9c_I_Love_Money_Its_In_The_Cards_S2E3/720p/20200103_093628/hls/0-end/master.m3u8`

kevinoconnor7:
@remitamine ohh interesting, great find! I'll dig into that some more. I'll be a bit wonky since I need to crack open each of the playlists for the various bitrates first and then reassemble them into the formats list. Definitely should be doable though.

kevinoconnor7:
Was able to put something together so that we can just ad-free m3u8 playlists and not have to filter out specific URLs. It's a bit janky since I need to get the ad-supported playlists, parse them out as formats, get a fragment URL, transform that into the `master.m3u8`, and extract formats again. If that process fails then I fallback to the ad-supported playlist.

Unfortunately I couldn't find another path to getting these URLs :(

I resolved the outstanding review comments since they're no longer applicable. If I do need to go that route again I'll just create a separate PR for that.

kevinoconnor7:
@dstftw Do you need anything from me for the geo restriction? It seems to be accessible on all the VPN providers I tried.

youtube_dl/extractor/plutotv.py
```diff
@@ -0,0 +1,162 @@
...
+            if res is False:
+                continue
+            playlist_doc, _ = res
+            first_segment_url = re.search(
+                r'^(https?://.*/)0\-(end|[0-9]+)/[^/]+\.ts$', playlist_doc,
+                re.MULTILINE)
+            if not first_segment_url:
+                continue
+            m3u8_url = compat_urlparse.urljoin(
+                first_segment_url.group(1), '0-end/master.m3u8')
```
dstftw, line 94: If such reconstructed URL is already processed it should be skipped.

youtube_dl/extractor/plutotv.py
```diff
@@ -0,0 +1,162 @@
...
+            formats.extend(
+                self._extract_m3u8_formats(
+                    url, video_id, 'mp4', 'm3u8_native',
+                    m3u8_id='hls', fatal=False))
+        self._sort_formats(formats)
+        ad_free_formats = self._to_ad_free_formats(video_id, formats)
+        self._sort_formats(ad_free_formats)
+        info = {
+            'id': video_id,
+            'formats': ad_free_formats if ad_free_formats else formats,
```
dstftw, line 117: `ad_free_formats` is never empty here since `_sort_formats` will throw if's empty.

youtube_dl/extractor/plutotv.py
```diff
@@ -0,0 +1,162 @@
...
+        video_type, slug = re.match(self._VALID_URL, url).groups()
+        slug_components = slug.split('/')
+        info_slug = slug_components[0]
+        video_json = self._download_json(self._INFO_URL + info_slug, info_slug,
+                                         query=self._INFO_QUERY_PARAMS)
+
+        if video_type == 'series':
+            series_name = video_json.get('name', info_slug)
+            season_number = int_or_none(try_get(slug_components, lambda x: x[2]))
+            episode_slug = try_get(slug_components, lambda x: x[4])
```
dstftw, line 141: This won't work for URLs with query.

dstftw (changes requested):

kevinoconnor7:
Applied all requested changes.

felisucoibi:
thanks a lot testing

PR #27650, opened 2021-01-03
[fix] Additional installation info to Ubuntu 20.04 users

### LINKED ISSUE #27649: The script at https://yt-dl.org/downloads/latest/youtube_dl doesn't works at Ubuntu 20.04.
## Checklist

- [ ] I'm reporting a broken site support issue
- [x ] I've verified that I'm running youtube-dl version **2021.01.03**
- [x ] I've checked that all provided URLs are alive and playable in a browser
- [x ] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x ] I've searched the bugtracker for similar bug reports including closed ones
- [x ] I've read bugs section in FAQ

## Verbose log

```
/usr/bin/env: 'python': No such file or directory
```

## Description

The script at
https://yt-dl.org/downloads/latest/youtube_dl use 'python' binary which
is not existing at Ubuntu 20.04 and this introduces a bug that the
script cannot find the 'python' binary. The only python binaries by
default are 'python3, python3.8, python3-futurize, python3-pasteurize'.
So simple symbolic link can solve the issue.

Since there is no support for "python 2" anymore. I think this bug
exist in many OS's and further investigation is needed.

DNonov:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This additional information is required since the script at
https://yt-dl.org/downloads/latest/youtube_dl use 'python' binary which
is not existing at Ubuntu 20.04 and this introduces a bug that the
script cannot find the 'python' binary. The only python binaries by
default are 'python3, python3.8, python3-futurize, python3-pasteurize'.
So a simple symbolic link can solve the issue.

Since there is no support for "python 2" anymore. I think this bug
exist in many OS's and further investigation is needed.

This pull request closes #27649.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,12 +28,16 @@ If you do not have curl, you can alternatively use a recent wget:
     sudo wget https://yt-dl.org/downloads/latest/youtube-dl -O /usr/local/bin/youtube-dl
     sudo chmod a+rx /usr/local/bin/youtube-dl
 
+If you are Ubuntu 20.04 user you going to need one more step due to /usr/bin/python absence:
+
+    sudo ln -s /usr/bin/python3 /usr/local/bin/python
+
 Windows users can [download an .exe file](https://yt-dl.org/latest/youtube-dl.exe) and place it in any location on their [PATH](https://en.wikipedia.org/wiki/PATH_%28variable%29) except for `%SYSTEMROOT%\System32` (e.g. **do not** put in `C:\Windows\System32`).
 
 You can also use pip:
 
     sudo -H pip install --upgrade youtube-dl
-    
+
 This command will update youtube-dl if you have already installed it. See the [pypi page](https://pypi.python.org/pypi/youtube_dl) for more information.
 
 macOS users can install youtube-dl with [Homebrew](https://brew.sh/):


dstftw:
User may want to use `python2` instead thus you should not force using python3. Or may want to achieve this via installing `python-is-python3`/`python-is-python2` without creating symlinks by hand. So I'll leave this decision up to user.

Also this is youtube-dl installation instructions not instructions on how to setup your OS and not some particular OS either.

rautamiekka:
> User may want to use `python2` instead thus you should not force using python3. Or may want to achieve this via installing `python-is-python3`/`python-is-python2` without creating symlinks by hand. So I'll leave this decision up to user.

His instructions **are** optional for the User to implement for this very reason.

I had no idea `python-is-python`{2,3} existed, so that's indeed even better.

Kikobeats:
This PR is adding additional steps for Linux builds and that's good. Note how the readme is already referencing additional information for Windows users, so there is no reason for don't merge this.

DNonov:
@Kikobeats  After some of the points @dstftw made and the fact that probably this bug exists on other OS's as well, I think that slight wording might be needed. IMO this problem has to be addressed in the `README.md` in some form cause it's information on a problem that a user will have. Another may be a better option is to put this information in the FAQ. 

The only reason I did this PR was that is a fast fix. IMO the suggestion @Kikobeats had at #27649 is the right way to go.

PR #27742, opened 2021-01-09
[discoveryplus] Added "it" to match italian domain urls

nixxo:

- [x] Improvement

---

discoveryplus is now in italy as well so I added the "it" suffix fo match italian domain urls and a test.

### DIFF
--- a/youtube_dl/extractor/dplay.py
+++ b/youtube_dl/extractor/dplay.py
@@ -20,7 +20,7 @@ class DPlayIE(InfoExtractor):
             (?:www\.)?(?P<host>d
                 (?:
                     play\.(?P<country>dk|fi|jp|se|no)|
-                    iscoveryplus\.(?P<plus_country>dk|es|fi|it|se|no)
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/dplay.py
```diff
@@ -20,7 +20,7 @@ class DPlayIE(InfoExtractor):
             (?:www\.)?(?P<host>d
                 (?:
                     play\.(?P<country>dk|fi|jp|se|no)|
-                    iscoveryplus\.(?P<plus_country>dk|es|fi|it|se|no)
+                    iscoveryplus\.(?P<plus_country>dk|es|fi|it|se|no|it)
```
pzygielo, line 23: Is the first `it`, the one between `fi` and `se`, not working properly?

dstftw:
`it` is already matched.

nixxo:
that's weird. I must have made a mistake in typing the url the first time because it didn't match and gave me the generic extractor... so I added the it suffix in the regex and it worked.

For sure I made a mistake somehow, sorry.

PR #27786, opened 2021-01-12
Add --max-already-downloaded option

akx:
  - 👉   I don't have tests for this yet. I couldn't find a test for `max_downloads` either; any pointers on how to best test this?

- [x] New feature

---

This PR adds a new `--max-already-downloaded` switch, similar to `--max-downloads`, but it only takes into account files that would have been redownloaded.

To facilitate that,

* ⚠️ it minutely changes the API for `FileDownloader.download()` to, instead of True, return a string representing the reason a download succeeded; currently it's either `"ok"` or `"already_downloaded"`. I wouldn't expect any downstream users to use `if ...download() is True`, and strings are truthy anyway, so this should work just fine
* it adds a new super `DownloadLimitReached` exception class from whence `MaxDownloadsReached` and the new `MaxAlreadyDownloadedReached` exceptions are derived, and catches it wherever `MaxDownloadsReached` used to be caught before.
* it adds the new option to the argument parser and plops it into the ytdl instance's `options` as you might expect
* it adds the necessary counter to the ytdl object

As a side effect, it also removes an unused copy of `report_file_already_downloaded`.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -624,13 +624,6 @@ class YoutubeDL(object):
         error_message = '%s %s' % (_msg_header, message)
         self.trouble(error_message, tb)
 
-    def report_file_already_downloaded(self, file_name):
-        """Report file has already been fully downloaded."""
-        try:
-            self.to_screen('[download] %s has already been downloaded' % file_name)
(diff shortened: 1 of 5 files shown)


akx:
Guess not 😁

PR #27787, opened 2021-01-12
[Mixcloud] Harmonize ID generation from lists with full ID generation

akx:

  - This is already covered by the other Mixcloud tests.

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Mixcloud IDs are generated as `username_slug` when the full ID dict has been downloaded.  When downloading a list (e.g. uploads, favorites, ...), the temporary ID is just the `slug`.  This made e.g. archive file usage require the download of stream metadata before the download can be rejected as already downloaded.

This PR attempts to get the uploader username during the GraphQL query, so the temporary IDs are generated similarly, and archive file checks can occur before the full stream metadata is downloaded.

### DIFF
--- a/youtube_dl/extractor/mixcloud.py
+++ b/youtube_dl/extractor/mixcloud.py
@@ -251,8 +251,13 @@ class MixcloudPlaylistBaseIE(MixcloudBaseIE):
                 cloudcast_url = cloudcast.get('url')
                 if not cloudcast_url:
                     continue
+                video_id = cloudcast.get('slug')
+                if video_id:
+                    owner_username = try_get(cloudcast, lambda x: x['owner']['username'], compat_str)
+                    if owner_username:
+                        video_id = '%s_%s' % (owner_username, video_id)
                 entries.append(self.url_result(
-                    cloudcast_url, MixcloudIE.ie_key(), cloudcast.get('slug')))
+                    cloudcast_url, MixcloudIE.ie_key(), video_id))
 
             page_info = items['pageInfo']
             has_next_page = page_info['hasNextPage']
@@ -321,7 +326,8 @@ class MixcloudUserIE(MixcloudPlaylistBaseIE):
     _DESCRIPTION_KEY = 'biog'
     _ROOT_TYPE = 'user'
     _NODE_TEMPLATE = '''slug
-          url'''
+          url
+          owner { username }'''
 
     def _get_playlist_title(self, title, slug):
         return '%s (%s)' % (title, slug)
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/mixcloud.py
```diff
@@ -251,8 +251,11 @@ def _real_extract(self, url):
                 cloudcast_url = cloudcast.get('url')
                 if not cloudcast_url:
                     continue
-                entries.append(self.url_result(
-                    cloudcast_url, MixcloudIE.ie_key(), cloudcast.get('slug')))
+                id = cloudcast.get('slug')
```
remitamine, line 254: don't Python builtin function names as variable names(`id`).

youtube_dl/extractor/mixcloud.py
```diff
@@ -251,8 +251,11 @@ def _real_extract(self, url):
...
+                owner_username = try_get(cloudcast, lambda x: x['owner']['username'], compat_str)
+                if owner_username:
+                    id = '%s_%s' % (owner_username, id)
+                entries.append(self.url_result(cloudcast_url, MixcloudIE.ie_key(), id))
```
remitamine, line 258: keep the line the same as it was(broken in two lines).

youtube_dl/extractor/mixcloud.py
```diff
@@ -320,8 +323,11 @@ class MixcloudUserIE(MixcloudPlaylistBaseIE):
     _TITLE_KEY = 'displayName'
     _DESCRIPTION_KEY = 'biog'
     _ROOT_TYPE = 'user'
-    _NODE_TEMPLATE = '''slug
-          url'''
+    _NODE_TEMPLATE = '''
+    slug
+    url
+    owner { username }
```
remitamine, line 329: keep the same GQL query style(it was intentional).

remitamine (changes requested):

youtube_dl/extractor/mixcloud.py
```diff
@@ -251,8 +251,12 @@ def _real_extract(self, url):
                 cloudcast_url = cloudcast.get('url')
                 if not cloudcast_url:
                     continue
+                video_id = cloudcast['slug']
```
remitamine, line 254: this will break the extraction if `slug` is null.

akx: It's on purpose. If the slug is null, a proper video ID can't be generated anyway.

remitamine: the extraction can work just by having the track URL, so the `video_id` is an optional field, and the extraction should proceed even if an entry id has not been found.

akx: Okiedoke, I wasn't aware of that. Thanks :)

PR #27790, opened 2021-01-12
[twitch] Set OAuth for GraphQL requests to avoid ads with cookies

main--:

---

- [ ] Covered the code with tests (note that PRs without tests will be REJECTED)  *I mean fair, but this is impossible to write a test for because hardcoded cookies are just going to keep expiring*

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Twitch recently changed how streams are authenticated. It is now required to send an `Authorization: OAuth` header with the GraphQL request or else the server will always treat the request as anonymous, with the usual consequences of 1. injecting ads 2. no access to subscriber-only content.

The OAuth token is stored in the `auth-token` cookie, so we can simply set the authorization header based on this cookie and everything works again.

### DIFF
--- a/youtube_dl/extractor/twitch.py
+++ b/youtube_dl/extractor/twitch.py
@@ -143,7 +143,7 @@ class TwitchBaseIE(InfoExtractor):
             'Client-ID': self._CLIENT_ID,
         }
         gql_auth = self._get_cookies('https://gql.twitch.tv').get('auth-token')
-        if gql_auth is not None:
+        if gql_auth:
             headers['Authorization'] = 'OAuth ' + gql_auth.value
         return self._download_json(
             'https://gql.twitch.tv/gql', video_id, note,
--- a/youtube_dl/extractor/twitch.py
+++ b/youtube_dl/extractor/twitch.py
@@ -138,13 +138,17 @@ class TwitchBaseIE(InfoExtractor):
         self._sort_formats(formats)
 
     def _download_base_gql(self, video_id, ops, note, fatal=True):
+        headers = {
+            'Content-Type': 'text/plain;charset=UTF-8',
+            'Client-ID': self._CLIENT_ID,
+        }
+        gql_auth = self._get_cookies('https://gql.twitch.tv').get('auth-token')
+        if gql_auth is not None:
+            headers['Authorization'] = 'OAuth ' + gql_auth.value
         return self._download_json(
             'https://gql.twitch.tv/gql', video_id, note,
             data=json.dumps(ops).encode(),
-            headers={
-                'Content-Type': 'text/plain;charset=UTF-8',
-                'Client-ID': self._CLIENT_ID,
-            }, fatal=fatal)
+            headers=headers, fatal=fatal)
 
     def _download_gql(self, video_id, ops, note, fatal=True):
         for op in ops:


remitamine:
> the usual consequences of 1. injecting ads 2. no access to subscriber-only content.

you should provide a way that would allow verifying that this change actually does one of the mentioned consequences(or others if there are any).

main--:
>you should provide a way that would allow verifying that this change actually does one of the mentioned consequences(or others if there are any).

Sorry, I'm not sure I understand? Steps to reproduce:

* subscribe to a channel on Twitch (or buy Turbo)
* dump cookies
* download or watch the stream using `youtube-dl --cookies ...`
* the stream has pre-roll and mid-roll ads
* apply this patch
* try again
* the stream no longer has ads

youtube_dl/extractor/twitch.py
```diff
@@ -138,13 +138,17 @@ def _prefer_source(self, formats):
         self._sort_formats(formats)
 
     def _download_base_gql(self, video_id, ops, note, fatal=True):
+        headers = {
+            'Content-Type': 'text/plain;charset=UTF-8',
+            'Client-ID': self._CLIENT_ID,
+        }
+        gql_auth = self._get_cookies('https://gql.twitch.tv').get('auth-token')
+        if gql_auth is not None:
+            headers['Authorization'] = 'OAuth ' + gql_auth.value
```
remitamine, line 147: this should be done once(in `_real_initialize`).

remitamine (changes requested):

youtube_dl/extractor/twitch.py
main--: Sorry, I don't understand what exactly you're asking for. I can't find any header setting function in `common.py`. Also, this header is only needed for the gql request. While it would be possible to set something like `self.gql_auth_header` in `_real_initialize`, this feels to me like adding gratuitous complexity.

remitamine: > I can't find any header setting function in `common.py`.

`_real_initialize` is used for extractor initialization, it's up to every extrator to setup it's own initialization.

> While it would be possible to set something like self.gql_auth_header in _real_initialize, this feels to me like adding gratuitous complexity.

there is no complexty, you will set the header once and reuse them in every subsequent request, instead of constructing the same headers over and over again for a large collections.

dstftw: Technically, cookie may change between requests so that moving `Authorization` calculation in `_real_initialize` may result in expired token.

remitamine: the assumption was based on the fact that the cookie is set programatically and not using `Set-Cookie` headers, but as it's undefined wheather the value can change or not, i guess it's better to set the cookie value for every request.

youtube_dl/extractor/twitch.py
```diff
@@ -138,13 +138,17 @@ def _prefer_source(self, formats):
         self._sort_formats(formats)
 
     def _download_base_gql(self, video_id, ops, note, fatal=True):
+        headers = {
+            'Content-Type': 'text/plain;charset=UTF-8',
+            'Client-ID': self._CLIENT_ID,
+        }
+        gql_auth = self._get_cookies('https://gql.twitch.tv').get('auth-token')
+        if gql_auth is not None:
```
remitamine, line 146:
```suggestion
        if gql_auth:
```

PR #27618, opened 2021-01-01
[vvvvid] fix season title extraction (#27601)

nixxo:
### Before submitting a *pull request* make sure you have:
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [x] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [ ] Covered the code with tests (note that PRs without tests will be REJECTED)
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)

### What is the purpose of your *pull request*?
- [x] Bug fix
- [x] Improvement

---

### Description of your *pull request* and other information
fixes #27601

the only place where to find the season "title" is to call the same url used in the playlist extractor and scroll the file looking for the season that the current video beloongs end save the season "title" that way.

It's probably not super efficient or elegant

Or I can extract it in the playlist extractor and use url_transparent to pass it, it's a lot easier and cleaner:

```
'_type': 'url_transparent',
'ie_key': VVVVIDIE.ie_key(),
...
'season': str_or_none(season.get('name'))
```

but if you download the single episode you don't have the 'season title' info available.

Issue #27601, opened 2020-12-31
[vvvvid] Support for "season" name extraction


pjammo:
## Checklist

- [x] I'm reporting a site feature request
- [x] I've verified that I'm running youtube-dl version 2020.12.31
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
$ youtube-dl -v --get-filename -o "%(season)s - %(episode)s" https://www.vvvvid.it/show/404/one-punch-man
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', '--get-filename', '-o', '%(season)s - %(episode)s', 'https://www.vvvvid.it/show/404/one-punch-man']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2020.12.31
[debug] Proxy map: {}
[debug] Default format spec: bestvideo+bestaudio/best
Season 1 - L’uomo più forte
[debug] Default format spec: bestvideo+bestaudio/best
Season 1 - Il cyborg solitario
[debug] Default format spec: bestvideo+bestaudio/best
Season 1 - L’ossessione dello scienziato
[debug] Default format spec: bestvideo+bestaudio/best
Season 1 - Il ninja dei giorni nostri
[debug] Default format spec: bestvideo+bestaudio/best
Season 1 - L’insegnante definitivo
[debug] Default format spec: bestvideo+bestaudio/best
Season 1 - La città del terrore
[debug] Default format spec: bestvideo+bestaudio/best
Season 1 - L’allievo supremo
[debug] Default format spec: bestvideo+bestaudio/best
Season 1 - Il re degli abissi
[debug] Default format spec: bestvideo+bestaudio/best
Season 1 - La giustizia inflessibile
[debug] Default format spec: bestvideo+bestaudio/best
Season 1 - Una crisi senza precedenti
[debug] Default format spec: bestvideo+bestaudio/best
Season 1 - Il signore dell’universo
[debug] Default format spec: bestvideo+bestaudio/best
Season 1 - L’eroe più forte
[debug] Default format spec: bestvideo+bestaudio/best
Season 2 - L’uomo più forte
[debug] Default format spec: bestvideo+bestaudio/best
Season 2 - Il cyborg solitario
[...]

PR #27825, opened 2021-01-14
[vvvvid] added youtube support + test

nixxo:

- [x] Improvement

---

Some videos on the website are youtube embeds.
So I added the support for this kind of videos maintaining the infos extractions for consistency with playlist extraction

### DIFF
--- a/youtube_dl/extractor/vvvvid.py
+++ b/youtube_dl/extractor/vvvvid.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 import re
 
 from .common import InfoExtractor
+from .youtube import YoutubeIE
 from ..utils import (
     ExtractorError,
     int_or_none,
@@ -43,6 +44,23 @@ class VVVVIDIE(InfoExtractor):
             'id': '482493',
             'ext': 'mp4',
             'title': 'Episodio 01',
+        },
+        'params': {
+            'skip_download': True,
+        },
+    }, {
+        # video_type == 'video/youtube'
+        'url': 'https://www.vvvvid.it/show/404/one-punch-man/406/486683/trailer',
+        'md5': '33e0edfba720ad73a8782157fdebc648',
+        'info_dict': {
+            'id': 'RzmFKUDOUgw',
+            'ext': 'mp4',
+            'title': 'Trailer',
+            'season': 'EXTRA',
+            'upload_date': '20150906',
+            'description': 'md5:a5e802558d35247fee285875328c0b80',
+            'uploader_id': 'BandaiVisual',
+            'uploader': 'BANDAI NAMCO Arts Channel',
         },
         'params': {
             'skip_download': True,
@@ -154,6 +172,7 @@ class VVVVIDIE(InfoExtractor):
                     if season_number:
                         info['season_number'] = int(season_number)
 
+        youtube = False
         for quality in ('_sd', ''):
             embed_code = video_data.get('embed_info' + quality)
             if not embed_code:
@@ -172,19 +191,18 @@ class VVVVIDIE(InfoExtractor):
                     if kenc_message:
                         embed_code += '?' + ds(kenc_message)
                 formats.extend(self._extract_akamai_formats(embed_code, video_id))
+            elif video_type == 'video/youtube':
+                youtube = True
             else:
                 formats.extend(self._extract_wowza_formats(
                     'http://sb.top-ix.org/videomg/_definst_/mp4:%s/playlist.m3u8' % embed_code, video_id))
             metadata_from_url(embed_code)
-
-        self._sort_formats(formats)
 
         metadata_from_url(video_data.get('thumbnail'))
         info.update(self._extract_common_video_info(video_data))
         info.update({
             'id': video_id,
             'title': title,
-            'formats': formats,
             'duration': int_or_none(video_data.get('length')),
             'series': video_data.get('show_title'),
             'season_id': season_id,
@@ -193,6 +211,19 @@ class VVVVIDIE(InfoExtractor):
             'like_count': int_or_none(video_data.get('video_likes')),
             'repost_count': int_or_none(video_data.get('video_shares')),
         })
+
+        if youtube:
+            info.update({
+                '_type': 'url_transparent',
+                'ie_key': YoutubeIE.ie_key(),
+                'url': embed_code,
+            })
+        else:
+            self._sort_formats(formats)
+            info.update({
+                'formats': formats,
+            })
+
         return info
 
 
--- a/youtube_dl/extractor/vvvvid.py
+++ b/youtube_dl/extractor/vvvvid.py
@@ -56,7 +56,6 @@ class VVVVIDIE(InfoExtractor):
             'id': 'RzmFKUDOUgw',
             'ext': 'mp4',
             'title': 'Trailer',
-            'season': 'EXTRA',
             'upload_date': '20150906',
             'description': 'md5:a5e802558d35247fee285875328c0b80',
             'uploader_id': 'BandaiVisual',
--- a/youtube_dl/extractor/vvvvid.py
+++ b/youtube_dl/extractor/vvvvid.py
@@ -171,7 +171,6 @@ class VVVVIDIE(InfoExtractor):
                     if season_number:
                         info['season_number'] = int(season_number)
 
-        youtube = False
         for quality in ('_sd', ''):
             embed_code = video_data.get('embed_info' + quality)
             if not embed_code:
@@ -191,17 +190,26 @@ class VVVVIDIE(InfoExtractor):
                         embed_code += '?' + ds(kenc_message)
                 formats.extend(self._extract_akamai_formats(embed_code, video_id))
             elif video_type == 'video/youtube':
-                youtube = True
+                info.update({
+                    '_type': 'url_transparent',
+                    'ie_key': YoutubeIE.ie_key(),
+                    'url': embed_code,
+                })
+                formats = None
             else:
                 formats.extend(self._extract_wowza_formats(
                     'http://sb.top-ix.org/videomg/_definst_/mp4:%s/playlist.m3u8' % embed_code, video_id))
             metadata_from_url(embed_code)
+
+        if formats:
+            self._sort_formats(formats)
 
         metadata_from_url(video_data.get('thumbnail'))
         info.update(self._extract_common_video_info(video_data))
         info.update({
             'id': video_id,
             'title': title,
+            'formats': formats,
             'duration': int_or_none(video_data.get('length')),
             'series': video_data.get('show_title'),
             'season_id': season_id,
@@ -210,19 +218,6 @@ class VVVVIDIE(InfoExtractor):
             'like_count': int_or_none(video_data.get('video_likes')),
             'repost_count': int_or_none(video_data.get('video_shares')),
         })
-
-        if youtube:
-            info.update({
-                '_type': 'url_transparent',
-                'ie_key': YoutubeIE.ie_key(),
-                'url': embed_code,
-            })
-        else:
-            self._sort_formats(formats)
-            info.update({
-                'formats': formats,
-            })
-
         return info
 
 
--- a/youtube_dl/extractor/vvvvid.py
+++ b/youtube_dl/extractor/vvvvid.py
@@ -172,6 +172,7 @@ class VVVVIDIE(InfoExtractor):
                         info['season_number'] = int(season_number)
 
         video_type = video_data.get('video_type')
+        is_youtube = False
         for quality in ('', '_sd'):
             embed_code = video_data.get('embed_info' + quality)
             if not embed_code:
@@ -195,22 +196,22 @@ class VVVVIDIE(InfoExtractor):
                     'ie_key': YoutubeIE.ie_key(),
                     'url': embed_code,
                 })
-                formats = None
+                is_youtube = True
                 break
             else:
                 formats.extend(self._extract_wowza_formats(
                     'http://sb.top-ix.org/videomg/_definst_/mp4:%s/playlist.m3u8' % embed_code, video_id))
             metadata_from_url(embed_code)
 
-        if formats:
+        if not is_youtube:
             self._sort_formats(formats)
+            info['formats'] = formats
 
         metadata_from_url(video_data.get('thumbnail'))
         info.update(self._extract_common_video_info(video_data))
         info.update({
             'id': video_id,
             'title': title,
-            'formats': formats,
             'duration': int_or_none(video_data.get('length')),
             'series': video_data.get('show_title'),
             'season_id': season_id,
--- a/youtube_dl/extractor/vvvvid.py
+++ b/youtube_dl/extractor/vvvvid.py
@@ -171,12 +171,12 @@ class VVVVIDIE(InfoExtractor):
                     if season_number:
                         info['season_number'] = int(season_number)
 
-        for quality in ('_sd', ''):
+        video_type = video_data.get('video_type')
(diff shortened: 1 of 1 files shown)


remitamine:
```
python test_download.py TestDownload.test_VVVVID_2
[VVVVID] Downloading JSON metadata
[VVVVID] 486683: Downloading JSON metadata
[youtube] RzmFKUDOUgw: Downloading webpage
[info] Writing video description metadata as JSON to: test_VVVVID_2_RzmFKUDOUgw.info.json
F
FAIL: test_VVVVID_2 (__main__.TestDownload):
  File "/home/amine/youtube-dl/test/test_download.py", line 210, in test_template
    expect_info_dict(self, tc_res_dict, tc.get('info_dict', {}))
  File "/home/amine/youtube-dl/test/helper.py", line 190, in expect_info_dict
  File "/home/amine/youtube-dl/test/helper.py", line 186, in expect_dict
  File "/home/amine/youtube-dl/test/helper.py", line 178, in expect_value
    self.assertEqual(
AssertionError: 'EXTRA' != None : Invalid value for field season, expected 'EXTRA', got None

Ran 1 test in 1.988s

FAILED (failures=1)
```

remitamine:
as there is no relation between the two PRs, i don't see why this PR would depend on the other one.

youtube_dl/extractor/vvvvid.py
```diff
@@ -47,6 +48,23 @@ class VVVVIDIE(InfoExtractor):
...
+    }, {
+        # video_type == 'video/youtube'
+        'url': 'https://www.vvvvid.it/show/404/one-punch-man/406/486683/trailer',
+        'md5': '33e0edfba720ad73a8782157fdebc648',
+        'info_dict': {
+            'id': 'RzmFKUDOUgw',
+            'ext': 'mp4',
+            'title': 'Trailer',
+            'season': 'EXTRA',
+            'upload_date': '20150906',
```
nixxo, line 60:
```suggestion
            'title': 'Trailer',
            'upload_date': '20150906',
```

nixxo:
and then add it back after #27618 is merged

remitamine:
i think it would be better to be implemented this way:
- extract `video_type` outside the for loop, and check there for the youtube `video_type`.
- change the for loop iterated values order.
- after decoding the embed_code if the youtube URL has been extracted update the info dict and break there.
- keep the `_sort_formats` in it's place and guard it with a condition, and also set the `formats` value in the info dict.

nixxo:
> 
> i think it would be better to be implemented this way:
> 
>     * extract `video_type` outside the for loop, and check there for the youtube `video_type`.
> 
>     * change the for loop iterated values order.
I wantend to keep it simple and consistent with the code already written without changing a lot.

- why check `video_type` outside of the loop (for youtube) and then check again inside (for other type)?

>     * after decoding the embed_code if the youtube URL has been extracted update the info dict and break there.
> 
>     * keep the `_sort_formats` in it's place and guard it with a condition, and also set the `formats` value in the info dict.

if I do this then all infos extracted by the VVVVID extractor will not be available but used the one extracted by the YouTube extractor. I wanted to extract all the info with this exractor to maintain consistency when using the PlaylistExtractor.
Eg. Usually there are normal episodes and then extra content available with youtube embeds but all the info (season, episode, etc) should be extracted by this extractor and not by the YouTube extractor.

It's why I kept all the info extraction in common and just moved the 'formats' at the end if the video is not youtube.

remitamine:
> why check video_type outside of the loop (for youtube)

because it will used later outside the `for loop` for the formats condition.

> and then check again inside (for other type)?

to update the info dict, and break earlier.

nixxo:
> 
> 
> > why check video_type outside of the loop (for youtube)
> 
> because it will used later outside the `for loop` for the formats condition.
> 
> > and then check again inside (for other type)?
> 
> to update the info dict, and break earlier.

I pushed a commit after that comments... see if it's ok.

youtube_dl/extractor/vvvvid.py
```diff
@@ -172,12 +189,20 @@ def metadata_from_url(r_url):
...
+                })
+                formats = None
             else:
                 formats.extend(self._extract_wowza_formats(
                     'http://sb.top-ix.org/videomg/_definst_/mp4:%s/playlist.m3u8' % embed_code, video_id))
             metadata_from_url(embed_code)
 
-        self._sort_formats(formats)
+        if formats:
+            self._sort_formats(formats)
```
remitamine, line 205: `_sort_formats` should always be called for non youtube videos(to break the extraction with a proper message when no formats has been extracted).

remitamine:
something like that:
```diff
diff --git a/youtube_dl/extractor/vvvvid.py b/youtube_dl/extractor/vvvvid.py
index f4cae7fe9..be9332888 100644
--- a/youtube_dl/extractor/vvvvid.py
+++ b/youtube_dl/extractor/vvvvid.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 import re
 
 from .common import InfoExtractor
+from .youtube import YoutubeIE
 from ..utils import (
     ExtractorError,
     int_or_none,
@@ -154,12 +155,20 @@ class VVVVIDIE(InfoExtractor):
                     if season_number:
                         info['season_number'] = int(season_number)
 
-        for quality in ('_sd', ''):
+        video_type = video_data.get('video_type')
+        is_youtube = video_type == 'video/youtube'
+        for quality in ('', '_sd'):
             embed_code = video_data.get('embed_info' + quality)
             if not embed_code:
                 continue
             embed_code = ds(embed_code)
-            video_type = video_data.get('video_type')
+            if is_youtube:
+                info.update({
+                    '_type': 'url_transparent',
+                    'ie_key': YoutubeIE.ie_key(),
+                    'url': embed_code,
+                })
+                break
             if video_type in ('video/rcs', 'video/kenc'):
                 if video_type == 'video/kenc':
                     kenc = self._download_json(
@@ -177,14 +186,15 @@ class VVVVIDIE(InfoExtractor):
                     'http://sb.top-ix.org/videomg/_definst_/mp4:%s/playlist.m3u8' % embed_code, video_id))
             metadata_from_url(embed_code)
 
-        self._sort_formats(formats)
+        if not is_youtube:
+            self._sort_formats(formats)
+            info['formats'] = formats
 
         metadata_from_url(video_data.get('thumbnail'))
         info.update(self._extract_common_video_info(video_data))
         info.update({
             'id': video_id,
             'title': title,
-            'formats': formats,
             'duration': int_or_none(video_data.get('length')),
             'series': video_data.get('show_title'),
             'season_id': season_id,

```

nixxo:
I pushed a commit after that comments (and deleted the comment) ... see if it's ok.

remitamine:
i guess you've missed the inline comment, and the break has been added because there is no point in making two iteration while the URL can be extracted in on iteration.

nixxo:
yeah, that' probably my fault. I made the comment before lunch with an empty stomach ;-)

then after lunch with full stomach ;-) I modified the code, committed and deleted the comment that was no longer accurate or relevant.

For the loop break, I think is not an issue because there is already a check at the start of the loop if the `embed_code` is None https://github.com/ytdl-org/youtube-dl/blob/5d93a6bdbaa394722a0e2d74611820cd191cd98c/youtube_dl/extractor/vvvvid.py#L176

remitamine:
the review comment is still not addressed yet.

youtube_dl/extractor/vvvvid.py
nixxo: https://github.com/ytdl-org/youtube-dl/pull/27825/commits/c5966a4dca65001c23540dc1dff2926dc869787d

remitamine:
https://github.com/ytdl-org/youtube-dl/blob/ba15b2fee673f76c2cff2f193d5b4a19029b501f/youtube_dl/extractor/common.py#L107

nixxo:
@remitamine thanks for your time. Can you give a look at #27618 when you have time so we can complete the vvvvid extractor issues.

remitamine:
i don't see the other patch as something that should be merged, `VIDEO` or `EXTRA` are not values that fit the definition of a season, besides that an additional request to the video extractor has been added just to extract this value.

nixxo:
have you read the issue referenced in the PR? https://github.com/ytdl-org/youtube-dl/issues/27601

remitamine:
yes, i did, the same apply there as well, i don't consider `VERSIONE IN ITALIANO` and `VERSIONE IN GIAPPONESE` as values suitable to be considered as a season.

nixxo:
I disagree.
As in the example given in the issue the they are alternative version of the SAME season (one in italian, the other in japanese) and so the `season_number episode_number_ episode_title` are all the same, except for the `season` title that reflect this difference and, IMHO, is a necessary info to provide to the user.

remitamine:
the season field is about what season of a series, not what version of a series.
for a Series X that has two versions(in italian and japanese) and has 3 seasons:
- for episode 2 of season 3 of the italian version, i would expect the season value to be `Season 3` or `Staffel 3` not `VERSIONE IN ITALIANO`.
- for episode 5 of season 2 of the japanese version, i would expect the season value to be `Season 2` or `Staffel 2` not `VERSIONE IN GIAPPONESE`.
etc...

nixxo:
no, there are no multiple seasons, but multiple version of the same season.

the problem is that the website does not provide any indication of the `season_number` in the data available on the website.

As an example: https://www.vvvvid.it/show/1414/l-attacco-dei-giganti-la-stagione-finale

this is the 4th season of the show. The fact that is season 4 is not present in any data available on the website. The only data available is that on the page there are 4 playlists (italian version, jap version, commentary, extras)

Episode 1 in the jap version has the same title, ep_number etc of the italian version. The only difference is the playlist that represent, imho, the season title.

And as written in the referenced issue, if I want to download BOTH version of the show and put them in different directories (with the output template) this info is crucial in differentiating the episodes and to which season "version" they belong.

remitamine:
again, i'm not against extracting this value(except for the extra request), i'm against mixing fields(puting value in fields that they don't belong to), for examples:

- the `thumbnail` field should not be filled with story board URL.
- the `upload_date` field should not be filled with air date.
- the format `language` field should not be filled with hardsub language.
etc...

nixxo:
initially I thought about using the `alt_title` field and doing something like `"%s (%s)" % (title, season)` basically resulting in `Episode title blah blah (VERSIONE IN GIAPPONESE)`

but, again, the info should be available to the user if he needs it (like in the issue for creating a different folder for each version of the season) and the only field that it seems to me fits the criteria for such info was `season`.

Can you suggest a field that is more appropriate?

remitamine:
i think none of the already documented fields is suitable for this value.

nixxo:
what about the `alt_title` field that I described in the prev comment?

remitamine:
as far as I know, `alt_title` is used when the video is also known with another alternative title, as an example:
`Le Voyage Dans la Lune` title extracted from a french website can have an alternative title `A Trip to the Moon`.
so I don't think that `alt_title` is suitable to be used here.

nixxo:
sorry to bother you again...

i thought about another thing.
if the `season` field is something like `Season 1 (ita)` or `Season 1 (jap)` is that ok?

I can use the original value from the website `VERSIONE IN ITALIANO` to create a 'cleaner' season name.

remitamine:
> if the season field is something like Season 1 (ita) or Season 1 (jap) is that ok?

no, there is a language field that can used for this.
https://github.com/ytdl-org/youtube-dl/blob/8bf9591a70757c624a8ea5bf686040ed752246e0/youtube_dl/extractor/common.py#L183

nixxo:
so I'll have to modify the formats and add the language info when available...

```
if info.get('formats'):
    langs = {
        'ITALIANO': 'it',
        'GIAPPONESE': 'ja',
    }
    for reg in langs:
        if re.search(reg, season_title):
            for fmt in info['formats']:
                 fmt['language'] = langs[reg]
```

remitamine:
i still think that this should be in the series extractor not in the episode extractor(adding a extra request for every episode just to extract a non mandatory info), the information apply to the full season, and also i think you should start with a stricter match(check for `VERSIONE IN `) unless you have other counter examples.

nixxo:
well, if I cant's use the `season` field (or alt_title) and can only use the `language` field then it's a bit difficult to do it in the series extractor since the `language` field is nested inside the `formats` field...

(in the common.py file `language` is only listed inside `formats`)

remitamine:
the selected `format` field values will be copied to the info dict(to be used for output template), so it's technically valid to set a format value in the info dict when it apply to all formats, however it's not guaranteed that in the future that VVVVID would start serving formats as separate streams(`audio only` and `video only` formats) where the value will apply to the `video only` formats, which is incorrect in this case(`video only` formats don't have audio so the `language` value should not be set).

nixxo:
from what I gathered in this conversation the `season` field is just a textual translation of the numerical `season_number` field becuase the only accepted value is like "Season 1" and not something different. Am I wrong?

Final suggestion... the `playlist_title` field in the series extractor instead of just the title I set it as `Title blah blah (versione in italiano)`

and I basically return a list of playlist (instead of a single one) each one with a name corresponding to the content of that playlist.

result:

```
λ python -m youtube_dl -o "%(playlist_title)s - %(episode_number)s - %(title)s" --get-filename https://www.vvvvid.it/show/404/one-punch-man
One-Punch Man (VERSIONE IN ITALIANO) - 1 - L’uomo più forte
One-Punch Man (VERSIONE IN ITALIANO) - 2 - Il cyborg solitario
One-Punch Man (VERSIONE IN ITALIANO) - 3 - L’ossessione dello scienziato
One-Punch Man (VERSIONE IN ITALIANO) - 4 - Il ninja dei giorni nostri
One-Punch Man (VERSIONE IN ITALIANO) - 5 - L’insegnante definitivo
One-Punch Man (VERSIONE IN ITALIANO) - 6 - La città del terrore
One-Punch Man (VERSIONE IN ITALIANO) - 7 - L’allievo supremo
One-Punch Man (VERSIONE IN ITALIANO) - 8 - Il re degli abissi
One-Punch Man (VERSIONE IN ITALIANO) - 9 - La giustizia inflessibile
One-Punch Man (VERSIONE IN ITALIANO) - 10 - Una crisi senza precedenti
One-Punch Man (VERSIONE IN ITALIANO) - 11 - Il signore dell’universo
One-Punch Man (VERSIONE IN ITALIANO) - 12 - L’eroe più forte
One-Punch Man (VERSIONE IN GIAPPONESE) - 1 - L’uomo più forte
One-Punch Man (VERSIONE IN GIAPPONESE) - 2 - Il cyborg solitario
One-Punch Man (VERSIONE IN GIAPPONESE) - 3 - L'ossessione dello scienziato
One-Punch Man (VERSIONE IN GIAPPONESE) - 4 - Il ninja dei giorni nostri
One-Punch Man (VERSIONE IN GIAPPONESE) - 5 - L’insegnante definitivo
One-Punch Man (VERSIONE IN GIAPPONESE) - 6 - La città del terrore
One-Punch Man (VERSIONE IN GIAPPONESE) - 7 - L’allievo supremo
One-Punch Man (VERSIONE IN GIAPPONESE) - 8 - Il re degli abissi
One-Punch Man (VERSIONE IN GIAPPONESE) - 9 - La giustizia inflessibile
One-Punch Man (VERSIONE IN GIAPPONESE) - 10 - Una crisi senza precedenti
One-Punch Man (VERSIONE IN GIAPPONESE) - 11 - Il signore dell’universo
One-Punch Man (VERSIONE IN GIAPPONESE) - 12 - L’eroe più forte
One-Punch Man (EXTRA) - NA - Trailer
One-Punch Man (EXTRA) - 1 - Street re-view by Riccardo Greco
```

PR #27890, opened 2021-01-20
[njpwworld] new extraction

aarubui:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Player code changed. Old extractor no longer works.

### DIFF
--- a/youtube_dl/extractor/njpwworld.py
+++ b/youtube_dl/extractor/njpwworld.py
@@ -36,6 +36,9 @@ class NJPWWorldIE(InfoExtractor):
             'title': 'WORLD TAG LEAGUE 2020 & BEST OF THE SUPER Jr.27 2020年12月6日 福岡・福岡国際センター バックステージコメント（字幕あり）',
             'tags': ["福岡・福岡国際センター", "バックステージコメント", "2020", "20年代"],
         },
+        'params': {
+            'skip_download': True,
+        },
     }]
 
     _LOGIN_URL = 'https://front.njpwworld.com/auth/login'
--- a/youtube_dl/extractor/njpwworld.py
+++ b/youtube_dl/extractor/njpwworld.py
@@ -86,7 +86,7 @@ class NJPWWorldIE(InfoExtractor):
 
         tag_block = get_element_by_class('tag-block', webpage)
         tags = re.findall(
-            r'<a[^>]*\sclass="tag-[^>]*>([^<]+)</a>', tag_block
+            r'<a[^>]+class="tag-[^"]+"[^>]*>([^<]+)</a>', tag_block
         ) if tag_block else None
 
         return {
--- a/youtube_dl/extractor/njpwworld.py
+++ b/youtube_dl/extractor/njpwworld.py
@@ -87,7 +87,7 @@ class NJPWWorldIE(InfoExtractor):
         tag_block = get_element_by_class('tag-block', webpage)
         tags = re.findall(
             r'<a[^>]*>([^<]+)</a>', tag_block
-        )
+        ) if tag_block else None
 
         return {
             'id': video_id,
--- a/youtube_dl/extractor/njpwworld.py
+++ b/youtube_dl/extractor/njpwworld.py
@@ -85,10 +85,9 @@ class NJPWWorldIE(InfoExtractor):
         self._sort_formats(formats)
 
         tag_block = get_element_by_class('tag-block', webpage)
-        tag_list = get_element_by_class('tag-list', tag_block) if tag_block else None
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/njpwworld.py
```diff
@@ -64,35 +71,27 @@ def _real_extract(self, url):
...
 
         self._sort_formats(formats)
 
-        post_content = get_element_by_class('post-content', webpage)
+        tag_block = get_element_by_class('tag-block', webpage)
         tags = re.findall(
-            r'<li[^>]+class="tag-[^"]+"><a[^>]*>([^<]+)</a></li>', post_content
-        ) if post_content else None
+            r'<a[^>]*>([^<]+)</a>', tag_block
+        )
```
remitamine, line 90: keep similar checks for element class and `get_element_by_class` value.

youtube_dl/extractor/njpwworld.py
```diff
@@ -64,35 +71,27 @@ def _real_extract(self, url):
...
+            })
 
         self._sort_formats(formats)
 
-        post_content = get_element_by_class('post-content', webpage)
+        tag_block = get_element_by_class('tag-block', webpage)
         tags = re.findall(
-            r'<li[^>]+class="tag-[^"]+"><a[^>]*>([^<]+)</a></li>', post_content
-        ) if post_content else None
+            r'<a[^>]*>([^<]+)</a>', tag_block
```
remitamine, line 89: still the check for element class is missing.

aarubui: I'm not understanding what you want. What do you mean by "element class"?

remitamine: the `class` attribute of the `a` HTML element.

youtube_dl/extractor/njpwworld.py
```diff
@@ -64,35 +71,28 @@ def _real_extract(self, url):
...
+                'ext': 'mp4',
+                'protocol': 'm3u8',
+                'quality': 2 if kind == 'high' else 1,
+            })
 
         self._sort_formats(formats)
 
-        post_content = get_element_by_class('post-content', webpage)
+        tag_block = get_element_by_class('tag-block', webpage)
+        tag_list = get_element_by_class('tag-list', tag_block) if tag_block else None
```
remitamine, line 88: why is this extracted in two steps now, and not directly from `tag-block`?

youtube_dl/extractor/njpwworld.py
```diff
@@ -64,35 +71,28 @@ def _real_extract(self, url):
...
         tags = re.findall(
-            r'<li[^>]+class="tag-[^"]+"><a[^>]*>([^<]+)</a></li>', post_content
-        ) if post_content else None
+            r'<a[^>]*\sclass="tag-[^>]*>([^<]+)</a>', tag_list
```
remitamine, line 90:
```suggestion
            r'<a[^>]+class="tag-[^"]+"[^>]*>([^<]+)</a>', tag_list
```

youtube_dl/extractor/njpwworld.py
aarubui: Here's the HTML structure: Inside the `tag-block` div are the `tag-list` div (which has the `tag-` links) and a "More" link. This way was to exclude the "More" link in lieu of the "element class" check. Now we have a redundancy with both the `tag-list` and `tag-` filters.

remitamine: after the change to check the `a` HTML element `class` attribute, this is no longer needed.

aarubui: I already said it's a redundancy. It'd be clearer if you just said, "The `a` check is better than the `div` check. Go with that."

youtube_dl/extractor/njpwworld.py
```diff
@@ -6,30 +6,37 @@
...
         'skip': 'Requires login',
-    }
+    }, {
+        'url': 'https://front.njpwworld.com/p/s_series_00563_16_bs',
+        'info_dict': {
+            'id': 's_series_00563_16_bs',
+            'ext': 'mp4',
+            'title': 'WORLD TAG LEAGUE 2020 & BEST OF THE SUPER Jr.27 2020年12月6日 福岡・福岡国際センター バックステージコメント（字幕あり）',
+            'tags': ["福岡・福岡国際センター", "バックステージコメント", "2020", "20年代"],
+        },
```
remitamine, line 38: `skip_download` is needed for the test to pass similar to the first test.

Issue #27891, opened 2021-01-20
FranceCulture is broken


aurelg:
## Checklist

- [X] I'm reporting a broken site support
- [X] I've verified that I'm running youtube-dl version **2021.01.16**
- [X] I've checked that all provided URLs are alive and playable in a browser
- [X] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [X] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
[debug] System config: []
[debug] User config: ['--restrict-filenames']
[debug] Custom config: []
[debug] Command-line args: ['--verbose', 'https://www.franceculture.fr/emissions/la-methode-scientifique/la-methode-scientifique-emission-du-mercredi-20-janvier-2021']
[debug] Encodings: locale UTF-8, fs utf-8, out utf-8, pref UTF-8
[debug] youtube-dl version 2021.01.16
[debug] Python version 3.9.1 (CPython) - Linux-5.10.7-arch1-1-x86_64-with-glibc2.32
[debug] exe versions: ffmpeg 4.3.1, ffprobe 4.3.1, rtmpdump 2.4
[debug] Proxy map: {}
[FranceCulture] la-methode-scientifique-emission-du-mercredi-20-janvier-2021: Downloading webpage
ERROR: Unable to extract video data; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/home/aurelien/nosave/./youtube-dl/youtube_dl/YoutubeDL.py", line 805, in wrapper
    return func(self, *args, **kwargs)
  File "/home/aurelien/nosave/./youtube-dl/youtube_dl/YoutubeDL.py", line 826, in __extract_info
    ie_result = ie.extract(url)
  File "/home/aurelien/nosave/./youtube-dl/youtube_dl/extractor/common.py", line 532, in extract
    ie_result = self._real_extract(url)
  File "/home/aurelien/nosave/./youtube-dl/youtube_dl/extractor/franceculture.py", line 33, in _real_extract
    video_data = extract_attributes(self._search_regex(
  File "/home/aurelien/nosave/./youtube-dl/youtube_dl/extractor/common.py", line 1010, in _search_regex
    raise RegexNotFoundError('Unable to extract %s' % _name)
[...]

PR #27903, opened 2021-01-21
[franceculture] Fix extractor

aurelg:

---

- [X] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

- solves #27891
- test timestamp updated since the website changed it by a few hours

Thanks for this wonderful work!

### DIFF
--- a/youtube_dl/extractor/franceculture.py
+++ b/youtube_dl/extractor/franceculture.py
@@ -36,11 +36,11 @@ class FranceCultureIE(InfoExtractor):
                     </h1>|
                     <div[^>]+class="[^"]*?(?:title-zone-diffusion|heading-zone-(?:wrapper|player-button))[^"]*?"[^>]*>
                 ).*?
-                (<button[^>]+data-(asset-source|url)="[^"]+"[^>]+>)
+                (<button[^>]+data-(?:url|asset-source)="[^"]+"[^>]+>)
             ''',
             webpage, 'video data'))
 
-        video_url = video_data.get('data-asset-source') or video_data.get('data-url')
+        video_url = video_data.get('data-url') or video_data['data-asset-source']
         title = video_data.get('data-asset-title') or video_data.get('data-diffusion-title') or self._og_search_title(webpage)
 
         description = self._html_search_regex(
@@ -64,6 +64,6 @@ class FranceCultureIE(InfoExtractor):
             'ext': ext,
             'vcodec': 'none' if ext == 'mp3' else None,
             'uploader': uploader,
-            'timestamp': int_or_none(video_data.get('data-asset-created-date') or video_data.get('data-start-time')),
+            'timestamp': int_or_none(video_data.get('data-start-time')) or int_or_none(video_data.get('data-asset-created-date')),
             'duration': int_or_none(video_data.get('data-duration')),
         }
--- a/youtube_dl/extractor/franceculture.py
+++ b/youtube_dl/extractor/franceculture.py
@@ -20,7 +20,7 @@ class FranceCultureIE(InfoExtractor):
             'title': 'Rendez-vous au pays des geeks',
             'thumbnail': r're:^https?://.*\.jpg$',
             'upload_date': '20140301',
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/franceculture.py
```diff
@@ -36,12 +36,12 @@ def _real_extract(self, url):
                     </h1>|
                     <div[^>]+class="[^"]*?(?:title-zone-diffusion|heading-zone-(?:wrapper|player-button))[^"]*?"[^>]*>
                 ).*?
-                (<button[^>]+data-asset-source="[^"]+"[^>]+>)
+                (<button[^>]+data-url="[^"]+"[^>]+>)
```
dstftw, line 39: Do not remove the old pattern.

youtube_dl/extractor/franceculture.py
```diff
@@ -36,12 +36,12 @@ def _real_extract(self, url):
...
             ''',
             webpage, 'video data'))
 
-        video_url = video_data['data-asset-source']
-        title = video_data.get('data-asset-title') or self._og_search_title(webpage)
+        video_url = video_data['data-url']
```
dstftw, line 43: Ditto.

youtube_dl/extractor/franceculture.py
```diff
@@ -36,12 +36,12 @@ def _real_extract(self, url):
...
+        title = video_data.get('data-diffusion-title') or self._og_search_title(webpage)
```
dstftw, line 44: Ditto.

youtube_dl/extractor/franceculture.py
```diff
@@ -64,6 +64,6 @@ def _real_extract(self, url):
             'ext': ext,
             'vcodec': 'none' if ext == 'mp3' else None,
             'uploader': uploader,
-            'timestamp': int_or_none(video_data.get('data-asset-created-date')),
+            'timestamp': int_or_none(video_data.get('data-start-time')),
```
dstftw, line 67: Ditto.

dstftw (changes requested):

dstftw (changes requested):

youtube_dl/extractor/franceculture.py
```diff
@@ -36,12 +36,12 @@ def _real_extract(self, url):
                     </h1>|
                     <div[^>]+class="[^"]*?(?:title-zone-diffusion|heading-zone-(?:wrapper|player-button))[^"]*?"[^>]*>
                 ).*?
-                (<button[^>]+data-asset-source="[^"]+"[^>]+>)
+                (<button[^>]+data-(asset-source|url)="[^"]+"[^>]+>)
```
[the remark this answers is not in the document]
aurelg: I replace `data-(asset-source|url)` by `data-(?:asset-source|url)` in the next commit. Is that what you expect?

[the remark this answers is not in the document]
dstftw: Yes.

youtube_dl/extractor/franceculture.py
```diff
@@ -36,12 +36,12 @@ def _real_extract(self, url):
...
             ''',
             webpage, 'video data'))
 
-        video_url = video_data['data-asset-source']
-        title = video_data.get('data-asset-title') or self._og_search_title(webpage)
+        video_url = video_data.get('data-asset-source') or video_data.get('data-url')
```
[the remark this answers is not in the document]
aurelg: Thanks for your help, that's my first contribution to this project and I'm still a bit lost after reading the coding conventions.

As far as I know, either `data-asset-source` or `data-url` can be found in the HTML, on the previous and current version of the FranceCulture website, respectively. This is the root cause of the current extractor being broken, and the very reason of this PR. If both are missing, though, `video_url` will be `False` and I understand that's inappropriate and the need for a default value. Yet, I can't figure out what to use since none seem to make sense as the extractor will fail anyway.

What would be an acceptable solution?

Issue #26940, opened 2020-10-20
Couldn't Download. It told me to report to you.  Here it is:


abrogard:
S C:\youtubedl> youtube-dl --verbose https://www.patreon.com/posts/just-practicing-42940451?utm_medium=post_notification_email"&"utm_source=post_link"&"utm_campaign=patron_engagement"&"token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJyZWRpc19rZXkiOiJpbnN0YW50LWFjY2VzczoxM2ZlYjUyZi1jMzI1LTQ5ZTctYWUzNi00NWZiZTY4MGM5MzgifQ.MOm7IyWFjCD7MqlLWkSZzF5s7uLqQfJsoyVEj0y_xI4
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['--verbose', 'https://www.patreon.com/posts/just-practicing-42940451?utm_medium=post_notification_email&utm_source=post_link&utm_campaign=patron_engagement&token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJyZWRpc19rZXkiOiJpbnN0YW50LWFjY2VzczoxM2ZlYjUyZi1jMzI1LTQ5ZTctYWUzNi00NWZiZTY4MGM5MzgifQ.MOm7IyWFjCD7MqlLWkSZzF5s7uLqQfJsoyVEj0y_xI4']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2020.09.20
[debug] Python version 3.4.4 (CPython) - Windows-10-10.0.19041
[debug] exe versions: ffmpeg 4.1
[debug] Proxy map: {}
[Patreon] 42940451: Downloading JSON metadata
ERROR: An extractor error has occurred. (caused by KeyError('post_file',)); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpksi3o1r1\build\youtube_dl\extractor\common.py", line 532, in extract
[...]

PR #27917, opened 2021-01-22
[patreon] add support for patreon-hosted video embeds

putnam:

---

- [x ] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [x ] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [x ] Covered the code with tests (note that PRs without tests will be REJECTED)

- [x ] Bug fix
- [x ] Improvement
- [ ] New extractor
- [ ] New feature

---

The current Patreon extractor does not support videos that are embedded/encoded/hosted by Patreon itself. It only supports videos that are attached to posts, like file uploads. Patreon-native videos use a hash check on the CDN URL, requiring the user to log in for paywalled content. I have re-enabled (and updated/fixed) the login code, and added support for these types of embeds.

### DIFF
--- a/youtube_dl/extractor/patreon.py
+++ b/youtube_dl/extractor/patreon.py
@@ -16,6 +16,7 @@ from ..utils import (
 import json
 import re
 
+
 class PatreonIE(InfoExtractor):
     _VALID_URL = r'https?://(?:www\.)?patreon\.com/(?:creation\?hid=|posts/(?:[\w-]+-)?)(?P<id>\d+)'
     _TESTS = [{
@@ -150,16 +151,16 @@ class PatreonIE(InfoExtractor):
         if not info.get('url'):
             post_file = attributes['post_file']
             if post_file.get('name') == 'video':
-               # single video embed
-               info.update({
-                   'url': post_file['url']
-               })
+                # single video embed
+                info.update({
+                    'url': post_file['url']
+                })
             else:
-              # video is attached as a file 
-              ext = determine_ext(post_file.get('name'))
-              if ext in KNOWN_EXTENSIONS:
-                  info.update({
-                      'ext': ext,
-                      'url': post_file['url'],
-                  })
+                # video is attached as a file
+                ext = determine_ext(post_file.get('name'))
+                if ext in KNOWN_EXTENSIONS:
+                    info.update({
+                        'ext': ext,
+                        'url': post_file['url'],
+                    })
         return info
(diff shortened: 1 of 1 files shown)


putnam:
this fixes #26940

youtube_dl/extractor/patreon.py
```diff
@@ -11,11 +11,16 @@
     parse_iso8601,
     str_or_none,
     try_get,
+    sanitized_Request,
+    ExtractorError,
```
dstftw, line 15: Alphabetic.

youtube_dl/extractor/patreon.py
```diff
@@ -11,11 +11,16 @@
...
 )
+import json
+import re
```
dstftw, line 18: Move before youtube-dl imports.

youtube_dl/extractor/patreon.py
```diff
@@ -63,25 +68,32 @@ class PatreonIE(InfoExtractor):
...
+                'relationships': {}
+            }
         }
 
         request = sanitized_Request(
-            'https://www.patreon.com/processLogin',
-            compat_urllib_parse_urlencode(login_form).encode('utf-8')
+            'https://www.patreon.com/api/login?include=campaign,user_location&json-api-version=1.0',
+            json.dumps(login_form).encode('ascii')
         )
```
dstftw, line 97: Move this inside `_download_webpage` call.

youtube_dl/extractor/patreon.py
```diff
@@ -63,25 +68,32 @@ class PatreonIE(InfoExtractor):
...
+                    'password': password
+                },
+                'relationships': {}
+            }
         }
 
         request = sanitized_Request(
-            'https://www.patreon.com/processLogin',
-            compat_urllib_parse_urlencode(login_form).encode('utf-8')
+            'https://www.patreon.com/api/login?include=campaign,user_location&json-api-version=1.0',
```
dstftw, line 95: Move query to `query` parameter.

dstftw (changes requested):

putnam:
thanks for the quick review. fixed the convention issue and just dropped using sanitized_Request which is no longer needed with the more robust download_webpage helper function. also there is actually no need for the query params on the patreon login page so i dropped those.

dstftw:
For further review you have to provide account credentials with access to paywalled video.

putnam:
Can you provide more specific instructions? Posting private paid Patreon
account to a github page is not a good idea.

jeLee6gi:
I tested your paywalled test case providing patreon cookies with `--cookies` and I get a 404 from vimeo. I am subscribed to the creator of that video and I can view it in my webbrowser.

Issue #21787, opened 2019-07-14
Funimation extractor is broken


AlexBlandin:
## Checklist

- [x] I'm reporting a broken site support
- [x] I've verified that I'm running youtube-dl version **2019.07.14**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
[debug] System config: []
[debug] User config: ['--restrict-filenames', '-f', '299+251/bestvideo+251/bestvideo+bestaudio/best', '-o', 'C:\\Users\\me\\Videos\\%(title)s.%(ext)s']
[debug] Custom config: []
[debug] Command-line args: ['https://www.funimation.com/shows/death-parade/death-seven-darts/?qid=b97f9d167bf5c72b', '--verbose']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2019.07.14
[debug] Python version 3.4.4 (CPython) - Windows-10-10.0.17134
[debug] exe versions: ffmpeg N-92712-g011c9112a0, ffprobe N-92712-g011c9112a0
[debug] Proxy map: {}
[Funimation] death-seven-darts: Downloading webpage
[Funimation] 159505: Downloading JSON metadata
ERROR: An extractor error has occurred. (caused by KeyError('items',)); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmphq4ewlx_\build\youtube_dl\extractor\common.py", line 530, in extract
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmphq4ewlx_\build\youtube_dl\extractor\funimation.py", line 117, in _real_extract
KeyError: 'items'
Traceback (most recent call last):
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmphq4ewlx_\build\youtube_dl\extractor\common.py", line 530, in extract
[...]

PR #27918, opened 2021-01-22
[Funimation] Raise descriptive ExtractorError

NobleKangaroo:

---

- [ ] Bug fix
- [X] Improvement
- [ ] New extractor
- [ ] New feature

---

Adds a more descriptive error message to Funimation extractor.
\- Related to #21787 - Provides users with more info about encountered errors

**Previous nondescript error:**
```
C:\Users\NobleKangaroo\Downloads\youtube-dl\youtube-dl>python3 __main__.py -u XXXXXXXX@XXX.XXX -p XXXXXXXX --cookies cookies.txt https://www.funimation.com/shows/black-butler/his-butler-able/
[Funimation] Logging in
[Funimation] his-butler-able: Downloading webpage
[Funimation] 22209: Downloading JSON metadata
ERROR: An extractor error has occurred. (caused by KeyError('items')); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

**New descriptive error:**
```
C:\Users\NobleKangaroo\Downloads\youtube-dl\youtube-dl>python3 __main__.py -u XXXXXXXX@XXX.XXX -p XXXXXXXX --cookies cookies.txt https://www.funimation.com/shows/black-butler/his-butler-able/
[Funimation] Logging in
[Funimation] his-butler-able: Downloading webpage
[Funimation] 22209: Downloading JSON metadata
ERROR: Funimation said: Uh oh! Your account is being accessed by too many devices. You can only stream to 1 screens at a time. To continue watching this video, stop playing video on one of the devices below:
Web
```

### DIFF
--- a/youtube_dl/extractor/funimation.py
+++ b/youtube_dl/extractor/funimation.py
@@ -110,11 +110,24 @@ class FunimationIE(InfoExtractor):
             headers = {}
             if self._TOKEN:
                 headers['Authorization'] = 'Token %s' % self._TOKEN
-            sources = self._download_json(
-                'https://www.funimation.com/api/showexperience/%s/' % video_id,
-                video_id, headers=headers, query={
-                    'pinst_id': ''.join([random.choice(string.digits + string.ascii_letters) for _ in range(8)]),
-                })['items']
+            try:
+                sources = self._download_json(
+                    'https://www.funimation.com/api/showexperience/%s/' % video_id,
+                    video_id, headers=headers, query={
+                        'pinst_id': ''.join([random.choice(string.digits + string.ascii_letters) for _ in range(8)]),
+                    })
+                sources = sources['items']
+            except KeyError:
+                if 'errors' in sources:
+                    errors = sources['errors']
+                    if len(errors) > 0:
+                        error = errors[0]
+                        if 'detail' in error:
+                            detail = error['detail']
+                            raise ExtractorError('%s said: %s' % (
+                                self.IE_NAME, detail), expected=True)
+                        else:
+                            raise ExtractorError(error, expected=True)
         except ExtractorError as e:
             if isinstance(e.cause, compat_HTTPError) and e.cause.code == 403:
                 error = self._parse_json(e.cause.read(), video_id)['errors'][0]


youtube_dl/extractor/funimation.py
```diff
@@ -110,11 +110,24 @@ def _search_kane(name):
...
-                video_id, headers=headers, query={
-                    'pinst_id': ''.join([random.choice(string.digits + string.ascii_letters) for _ in range(8)]),
-                })['items']
+            try:
+                sources = self._download_json(
+                    'https://www.funimation.com/api/showexperience/%s/' % video_id,
+                    video_id, headers=headers, query={
+                        'pinst_id': ''.join([random.choice(string.digits + string.ascii_letters) for _ in range(8)]),
+                    })
+                sources = sources['items']
```
NobleKangaroo, line 119: Broke this out into a separate line so that `sources` is defined regardless of whether or not it contains a key `items`. This is useful because we can still grab the `errors` even if a `KeyError` is generated, which is what the code below does.

youtube_dl/extractor/funimation.py
```diff
@@ -110,11 +110,24 @@ def _search_kane(name):
...
+            except KeyError:
+                if 'errors' in sources:
```
NobleKangaroo, line 121: If the key `items` doesn't exist, we except this and try to figure out what the error is. If there is at least one error, try to get the `detail` and display it to the user. If there's an error without a `detail`, just spit out the whole error to the user. Could perhaps fall back even further if there's not an `errors[0]` at all and simply spit out the whole `sources` but that's probably a bit overkill? If there's not an `errors[0]`, odds are the user isn't going to have any idea how to fix it.

youtube_dl/extractor/funimation.py
dstftw, line 119: You should not throw everything under try/except. Also read coding conventions.

youtube_dl/extractor/funimation.py
dstftw, line 121: Read coding conventions.

dstftw (changes requested):

youtube_dl/extractor/funimation.py
NobleKangaroo: Understood. So in short, this extractor was written without "get()" - I'll try to clean it up a bit and push another commit.

NobleKangaroo: Before I commit, how does this look?

```python
        try:
            headers = {}
            if self._TOKEN:
                headers['Authorization'] = 'Token %s' % self._TOKEN

# + BEGIN EDITS +
            meta = self._download_json(
                'https://www.funimation.com/api/showexperience/%s/' % video_id,
                video_id, headers=headers, query={
                    'pinst_id': ''.join([random.choice(string.digits + string.ascii_letters) for _ in range(8)]),
                })
            sources = meta.get('sources')
            errors = meta.get('errors')
            if errors is not None:
                if len(errors) > 0:
                    error = errors[0]
                    detail = error.get('detail')
                    if detail is not None:
                        raise ExtractorError('%s said: %s' % (
                            self.IE_NAME, detail), expected=True)
                    else:
                        raise ExtractorError(error, expected=True)
# - END EDITS -

        except ExtractorError as e:
            if isinstance(e.cause, compat_HTTPError) and e.cause.code == 403:
                error = self._parse_json(e.cause.read(), video_id)['errors'][0]
                raise ExtractorError('%s said: %s' % (
                    self.IE_NAME, error.get('detail') or error.get('title')), expected=True)
            raise
```

NobleKangaroo: If errors is always a list and we want to iterate the whole list instead of just the first error:
```python3
        try:
            headers = {}
            if self._TOKEN:
                headers['Authorization'] = 'Token %s' % self._TOKEN

# + BEGIN NEW CODE +
            meta = self._download_json(
                'https://www.funimation.com/api/showexperience/%s/' % video_id,
                video_id, headers=headers, query={
                    'pinst_id': ''.join([random.choice(string.digits + string.ascii_letters) for _ in range(8)]),
                })
            sources = meta.get('items') or []
            errors = meta.get('errors')
            if errors:
                for error in errors:
                    detail = error.get('detail')
                    if detail:
                        raise ExtractorError('%s said: %s' % (
                            self.IE_NAME, detail), expected=True)
                    else:
                        raise ExtractorError(error, expected=True)
# - END NEW CODE -

        except ExtractorError as e:
            if isinstance(e.cause, compat_HTTPError) and e.cause.code == 403:
                error = self._parse_json(e.cause.read(), video_id)['errors'][0]
                raise ExtractorError('%s said: %s' % (
                    self.IE_NAME, error.get('detail') or error.get('title')), expected=True)
            raise
```

dstftw: More or less correct not taking code duplication into account.
>Aside from the fixes, should we loop and display all of the errors (instead of just `errors[0]`)?

You can if this make sense.
>Also, is `errors` always a list (remove the `isinstance` check)?

There is no such guarantee.

NobleKangaroo: Perfect. Yeah the code duplication was only there for reference of what is being changed. :) Submitting another commit now.

dstftw:
Can you name a location this can be reproduced from?

NobleKangaroo:
@dstftw
> Can you name a location this can be reproduced from?

Sorry I must have missed the email notification.

If you open a tab on Funimation, using a free account (one screen limit), then try to run youtube-dl with the same credentials, the error is produced.

joshbrown50:
If I use this modified code I get ERROR: No valid Free Avails for this experience.
Is that the same error as before? Or is that something new?

PR #27970, opened 2021-01-26
[AMP] Fix upload_date and timestamp extraction

adrianheine:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

### DIFF
--- a/youtube_dl/extractor/amp.py
+++ b/youtube_dl/extractor/amp.py
@@ -8,6 +8,7 @@ from ..utils import (
     int_or_none,
     mimetype2ext,
     parse_iso8601,
+    unified_timestamp,
     url_or_none,
 )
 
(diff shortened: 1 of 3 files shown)


youtube_dl/extractor/amp.py
```diff
@@ -9,6 +9,7 @@
     mimetype2ext,
     parse_iso8601,
     url_or_none,
+    unified_timestamp,
```
remitamine, line 12: Alphabetic.

remitamine (changes requested):
`BleacherReportCMSIE` test is also affected by this change.

youtube_dl/extractor/amp.py
adrianheine: Oh I tried

adrianheine:
> `BleacherReportCMSIE` test is also affected by this change.

That video seems to not be available anymore anyway.

remitamine:
> > `BleacherReportCMSIE` test is also affected by this change.
> 
> That video seems to not be available anymore anyway.

the extractor does not rely on the webpage.

adrianheine:
> the extractor does not rely on the webpage.

You're right, thanks for your diligence.

PR #28019, opened 2021-01-30
[youporn] Add duration

piplongrun:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Add video duration to output

### DIFF
--- a/youtube_dl/extractor/youporn.py
+++ b/youtube_dl/extractor/youporn.py
@@ -25,6 +25,7 @@ class YouPornIE(InfoExtractor):
             'title': 'Sex Ed: Is It Safe To Masturbate Daily?',
             'description': 'Love & Sex Answers: http://bit.ly/DanAndJenn -- Is It Unhealthy To Masturbate Daily?',
             'thumbnail': r're:^https?://.*\.jpg$',
+            'duration': 210,
             'uploader': 'Ask Dan And Jennifer',
             'upload_date': '20101217',
             'average_rating': int,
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/youporn.py
```diff
@@ -153,6 +153,8 @@ def _real_extract(self, url):
         thumbnail = self._search_regex(
             r'(?:imageurl\s*=|poster\s*:)\s*(["\'])(?P<thumbnail>.+?)\1',
             webpage, 'thumbnail', fatal=False, group='thumbnail')
+        duration = int_or_none(self._html_search_meta(
+            'video:duration', webpage, 'duration', default=False, fatal=False))
```
adrianheine, line 157: If you have `default`, `fatal` is not used.

adrianheine (changes requested):

youtube_dl/extractor/youporn.py
piplongrun: I did not know, thanks for the tip! I learn something new every time I do something with youtube-dl :) I've removed the `default`, following the line of the rest of the code in this file.

adrianheine: I know what you mean :) I think `default=None` is actually better (without `fatal=False`). `fatal=False` makes it emit a warning and return `None`, while `default=None` will just silently return `None`. For optional metadata like `duration`, I think it's better not to have a warning.

dstftw:
Add a test.

PR #28053, opened 2021-02-03
[GoogleDrive] fix non-video files extraction

kfur:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Currently google drive extractor can extract only video files, it's drawback of this [commit](https://github.com/ytdl-org/youtube-dl/commit/25a35cb38a472731d9f487309f6bcff94ee4918c).
This pr fix it.


remitamine:
Add a test.

kfur:
Added

remitamine:
youtube-dl is media downloader(video/audio), if this breaks the extraction for audio files, then it would be taken into consideration after extracting the rest of the information, otherwise it's not a goal of the tool to support non video/audio files.

kfur:
I understand, but it's just return a little bit of old code without increasing complexity

kfur:
By the way audio extracting doesn't work at this moment, but this commit fix it too

remitamine:
> By the way audio extracting doesn't work at this moment, but this commit fix it too

then you should add an audio test instead of the PDF test, and extract other available info that where accessible before(`thumbnail`, `duration`, ...).

kfur:
Changed test.
I don't know how to retrieve `duration` or other metadata.
I nothing found in network tab of web inspector.

youtube_dl/extractor/googledrive.py
```diff
@@ -173,7 +182,14 @@ def get_value(key):
         reason = get_value('reason')
         title = get_value('title')
         if not title and reason:
-            raise ExtractorError(reason, expected=True)
+            # try old style title extraction
+            webpage = self._download_webpage(
+                'http://docs.google.com/file/d/%s' % video_id, video_id)
```
remitamine, line 187: use the more generic domain(Google Drive's domain).

youtube_dl/extractor/googledrive.py
```diff
@@ -173,7 +182,14 @@ def get_value(key):
...
+            title = self._search_regex(
+                r'"title"\s*,\s*"([^"]+)', webpage, 'title',
```
remitamine, line 189: no longer exists in the webpage.

youtube_dl/extractor/googledrive.py
```diff
@@ -173,7 +182,14 @@ def get_value(key):
...
+                default=None) or self._og_search_title(webpage)
```
remitamine, line 190: add `name` meta tag value as fallback.

remitamine (changes requested):
set `vcodec` to `none` for audio files.

upintheairsheep:
> youtube-dl is media downloader(video/audio), if this breaks the extraction for audio files, then it would be taken into consideration after extracting the rest of the information, otherwise it's not a goal of the tool to support non video/audio files.

If the method supports non-media files, there is absolutely no reason not to support non-AV files, but not provide any support for it. See https://github.com/yt-dlp/yt-dlp/pull/5178 , they added a non-AV filter that crashes the thing even though extraction of other assets is possible

upintheairsheep:
Let's merge!

upintheairsheep:
Try instead, making the current video one the default, and the pull requested one only extract when the current one does not find any formats.

upintheairsheep:
@dirkf Can you test and merge?

upintheairsheep:
I tested it, and it works! Other for the fact it sometimes gives naming errors on some non-audio files.

upintheairsheep:
May somebody please move this pull request to yt-dlp? At least merge it? This pull request completely works and the issues could easily be fixed.

PR #2844, opened 2014-05-01
Added flag to merge videos from a playlist into one continuous file

amlweems:
#### Example Use-Case

```
youtube-dl http://thecolbertreport.cc.com/full-episodes/wllm5p/april-30--2014---audra-mcdonald
```

Usually, this produces four videos, one for each segment of the show. However...

```
youtube-dl --concat http://thecolbertreport.cc.com/full-episodes/wllm5p/april-30--2014---audra-mcdonald
```

Will produce one continuous video for the show.

PR #28153, opened 2021-02-13
Add concat-playlist flag to merge videos from a playlist into one continuous file

adammw:

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

### Prior art
https://github.com/ytdl-org/youtube-dl/issues/12385
https://github.com/ytdl-org/youtube-dl/pull/995
https://github.com/ytdl-org/youtube-dl/pull/2723
https://github.com/ytdl-org/youtube-dl/pull/2844

Added flag to merge videos from a playlist into one continuous file. 

In its current form, it uses the ffmpeg concat demux, which has limited support for mixed streams within a playlist, therefore is only really suited to the use case below.

The main use case is downloading episodes from MTV, as episodes are internally represented as a playlist of multiple segments, and downloading them as individual files becomes a pain for managing in other applications.

HELP WANTED - need to improve the filename/extractor logic as often the playlist name is not the best name for the video, in the example of MTV, the individual segments have more metadata about the show/season/episode than the playlist does.


youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -274,7 +273,8 @@ def run(self, information):
             raise PostProcessingError('WARNING: unable to obtain file audio codec with ffprobe')
 
         more_opts = []
-        if self._preferredcodec == 'best' or self._preferredcodec == filecodec or (self._preferredcodec == 'm4a' and filecodec == 'aac'):
+        if self._preferredcodec == 'best' or self._preferredcodec == filecodec or (
```
pukkandan, line 276: Best not to make unrelated changes in a PR. Same issue below. Also these changes often don't obey the linter

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -542,6 +545,29 @@ def can_merge(self):
...
+        file_opts = ['-f', 'concat', '-safe', '0']
+        opts = ['-c', 'copy']
+        self._downloader.to_screen('[ffmpeg] Concatenating files into "%s"' % filename)
+        self.run_ffmpeg(list_filename, temp_filename, opts, file_opts)
+        os.rename(encodeFilename(temp_filename), encodeFilename(filename))
+        return info['__files_to_concat'], info
+
+    def can_merge(self):
+        return True
+
```
pukkandan, line 569: Why?

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -542,6 +545,29 @@ def can_merge(self):
         return True
 
 
+class FFmpegConcatPP(FFmpegPostProcessor):
+    def run(self, info):
+        filename = info['filepath']
+
+        list_filename = prepend_extension(filename, 'list')
```
pukkandan, line 552: The filename will end up as `.list.mkv` which doesn't make sense. It should be either `.mkv.list` or just `.list`

pukkandan (commented):
DISCLAIMER: I am not a maintainer of this project. I am just putting up my suggestions since I find this PR interesting

tomck:
This does not work with using output titles, I think that it's expecting different filenames when concatenating and it misses. I have tested with using an output format: https://pastebin.com/jspQV0Q0 and without using an output format: https://pastebin.com/iAiMZAMB

The command I'm using is referenced from https://github.com/yt-dlp/yt-dlp/issues/447 - he also has a workaround for renaming the files after finished, but I think the --concat-playlist flag is failing if it doesn't work only when using the --output flag. Since these 'acts' are all of the same episode (there could be a check to see if they match for all x acts), it should work.

It is interesting to me that the %(title)s finds the %(episode)s part and that becomes the final output file.

PR #28223, opened 2021-02-21
[StreamwoIE] Add new extractor

aalsuwaidi:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This PR adds a new extractor to support [Streamwo](https://streamwo.com/), the video file downloaded via the generic extractor can't be played after download.

### DIFF
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -1159,6 +1159,7 @@ from .storyfire import (
 from .streamable import StreamableIE
 from .streamcloud import StreamcloudIE
 from .streamcz import StreamCZIE
+from .streamwo import StreamwoIE
 from .streetvoice import StreetVoiceIE
 from .stretchinternet import StretchInternetIE
 from .stv import STVPlayerIE
--- a/youtube_dl/extractor/streamwo.py
+++ b/youtube_dl/extractor/streamwo.py
@@ -1,0 +1,30 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class StreamwoIE(InfoExtractor):
+    _VALID_URL = r'https?://streamwo\.com/(?P<id>\w+)'
+    _TEST = {
+        "url": 'https://streamwo.com/4529aff4',
+        'md5': 'eaf1f163635c868ecbba95d23ba83448',
+        'info_dict': {
+            'id': '4529aff4',
+            'ext': 'mp4',
+            'title': 'Goal 2021 13-04-50'
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        self.report_extraction(video_id)
+        title = self._html_search_regex(r'<span class="titles">(.+?)</span>', webpage, 'title')
+        video_url = self._html_search_regex(r'<source src="(.+?)" type="video/mp4">', webpage, u'video URL')
+        return {
+            'id': video_id,
+            'url': video_url,
+            'ext': 'mp4',
+            "title": title
+        }


remitamine:
Not Playable in the browser.

remitamine:
this doesn't fix the real problem which can occur for other websites as well, the problem is that the generic extractor does not currently respect the referer policy set by the website, so this is what needs to be fixed.

aalsuwaidi:
Made changes to streamwo.py to extract the embedded media file and pass that to the Generic extractor. Similar to how it is done in [the Lenta Extractor](https://github.com/ytdl-org/youtube-dl/blob/416da574ec0df3388f652e44f7fe71b1e3a4701f/youtube_dl/extractor/lenta.py#L48)

remitamine:
> Made changes to streamwo.py to extract the embedded media file and pass that to the Generic extractor. Similar to how it is done in [the Lenta Extractor](https://github.com/ytdl-org/youtube-dl/blob/416da574ec0df3388f652e44f7fe71b1e3a4701f/youtube_dl/extractor/lenta.py#L48)

this would've been acceptable for non-standard embeding methods, but in the case of Streamwo website, it uses the standard html5 video element, the problem mentioned in my previous comment should be fixed at the generic extractor not by creating a new extractor.

aalsuwaidi:
Okay, now it is clear sorry for the confusion.

Would it be better to add this logic in the implementation of [_parse_html5_media_entries](https://github.com/ytdl-org/youtube-dl/blob/3c58f9e0b9d8471212406e012727374db084932b/youtube_dl/extractor/common.py#L2475)? 

For example this could be done like this:
```python
for f in media_info['formats']:
	parsed_url = compat_urlparse.urlparse(base_url)
	if referer_policy == "no-referrer":
		f.setdefault('http_headers', {})
	elif referer_policy == "no-referrer-when-downgrade" and "https" == parsed_url.scheme:
		f.setdefault('http_headers', {})['Referer'] = base_url
	elif referer_policy == "no-referrer-when-downgrade" and "http" == compat_urlparse.urlparse(f["url"]).scheme:
		f.setdefault('http_headers', {})
	elif referer_policy == "origin":
		f.setdefault('http_headers', {})['Referer'] = compat_urlparse.urljoin(base_url, "/")
	else:
		f.setdefault('http_headers', {})['Referer'] = base_url
```

Where **referer_policy** would be the value found with:
```python
def get_referer_policy(page):
	mobj = re.search(r'<meta name="referrer" content="(.+?)">', page)
	if mobj:
		return mobj.group(1
```

That way the correct referrer URL would be applied when parsing the standard HTML5 video elements rather than adding this logic to the Generic extractor.

remitamine:
you're in the right direction now.

> Would it be better to add this logic in the implementation of _parse_html5_media_entries?

i think it would be better to introduce a utility function(placed in `utils.py`) that can be reused in other places, and use that function in `_parse_html5_media_entries`.

remitamine (changes requested):

youtube_dl/utils.py
```diff
@@ -3901,10 +3954,10 @@ def urlencode_postdata(*args, **kargs):
 def update_url_query(url, query):
     if not query:
         return url
-    parsed_url = compat_urlparse.urlparse(url)
-    qs = compat_parse_qs(parsed_url.query)
+    parsed_referrer_source = compat_urlparse.urlparse(url)
+    qs = compat_parse_qs(parsed_referrer_source.query)
     qs.update(query)
-    return compat_urlparse.urlunparse(parsed_url._replace(
+    return compat_urlparse.urlunparse(parsed_referrer_source._replace(
```
remitamine, line 3960: don't make unrelated changes.

youtube_dl/extractor/common.py
```diff
@@ -2590,7 +2595,8 @@ def _media_formats(src, cur_media_type, type_info={}):
                             'url': absolute_url(src),
                         })
             for f in media_info['formats']:
-                f.setdefault('http_headers', {})['Referer'] = base_url
+                referrer = get_referrer_url(base_url, f["url"], referrer_policy)
+                f.setdefault('http_headers', {})['Referer'] = referrer
```
remitamine, line 2599: the `get_referrer_url` function can return None, there is not point in setting `Referer` to None.

remitamine (changes requested):

youtube_dl/utils.py
aalsuwaidi: This must have changed when updating `get_referrer_url` to match the specification.

aalsuwaidi:
@remitamine Any updates on this?

remitamine:
> @remitamine Any updates on this?

i will merge the changes when i check that they work as been described in Referer-Policy spec and other related specifications(such as the Web Origin spec).

remitamine:
the implementation of at least `origin-when-cross-origin` and `strict-origin` is incorrect.

Issue #21870, opened 2019-07-23
BBC /reel/video/ 


toyotabedzrock:
## Checklist

- [x] I'm reporting a broken site support
- [x] I've verified that I'm running youtube-dl version **2019.07.16**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['--verbose', 'https://www.bbc.com/reel/video/p07f3mvf/stonewall-riots-i-wanted-to-kill-the-police-']
[debug] Encodings: locale cp1252, fs mbcs, out cp437, pref cp1252
[debug] youtube-dl version 2019.07.16
[debug] Python version 3.4.4 (CPython) - Windows-10-10.0.10586
[debug] exe versions: ffmpeg N-93448-g5fceac1cdb, ffprobe N-93448-g5fceac1cdb
[debug] Proxy map: {}
[bbc] stonewall-riots-i-wanted-to-kill-the-police-: Downloading webpage
ERROR: no suitable InfoExtractor for URL https://www.bbc.co.uk/programmes/None
  File "__main__.py", line 19, in <module>
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpaxmfe_l6\build\youtube_dl\__init__.py", line 474, in main
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpaxmfe_l6\build\youtube_dl\__init__.py", line 464, in _real_main
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpaxmfe_l6\build\youtube_dl\YoutubeDL.py", line 2008, in download
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpaxmfe_l6\build\youtube_dl\YoutubeDL.py", line 807, in extract_info
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpaxmfe_l6\build\youtube_dl\YoutubeDL.py", line 868, in process_ie_result
[...]

PR #28268, opened 2021-02-25
Support BBC Reel URIs

dirkf:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Extracted from PR #23415.

Fixes Issue #21870 (etc).

Test URL: https://www.bbc.com/reel/video/p07c6sb6/how-positive-thinking-is-harming-your-happiness

### DIFF
--- a/youtube_dl/extractor/bbc.py
+++ b/youtube_dl/extractor/bbc.py
@@ -962,7 +962,7 @@ class BBCIE(BBCCoUkIE):
         # bbc reel (e.g. https://www.bbc.com/reel/video/p07c6sb6/how-positive-thinking-is-harming-your-happiness)
         initial_data = self._search_regex(
             r'<script[^>]+id="initial-data"[^>]+data-json=\'(.+)\'>',
-            webpage, 'initial data', fatal=False, default=None)
+            webpage, 'initial data', default=None)
         if initial_data:
             programme_id = self._search_regex(
                 r'"versionID":"(%s)"' % self._ID_REGEX,
--- a/youtube_dl/extractor/bbc.py
+++ b/youtube_dl/extractor/bbc.py
@@ -801,10 +801,11 @@ class BBCIE(BBCCoUkIE):
             'id': 'p07c6sb9',
             'ext': 'mp4',
             'title': 'How positive thinking is harming your happiness',
-            'description': 'BEST OF 2019: Illusionist and writer Derren Brown discusses why we should be rethinking our traditional concept of happiness.\n \nCould optimism and positive thinking actually have a harmful effect on our well-being?\n \nVideo by Griesham Taan\n\nAnimation by Jacqueline Robertson',
+            'description': 'md5:fad74b31da60d83b8265954ee42d85b4',
             'timestamp': 1559606400,
             'duration': 235,
             'thumbnail': 'https://ychef.files.bbci.co.uk/64x64/p07c9dsr.jpg',
+            'upload_date': '20190604',
         },
     }]
 
--- a/youtube_dl/extractor/bbc.py
+++ b/youtube_dl/extractor/bbc.py
@@ -965,25 +965,14 @@ class BBCIE(BBCCoUkIE):
                 r'<[sS][cC][rR][iI][pP][tT]\b[^>]+\bid\s*=\s*(["\'])initial-data\1[^>]+\bdata-json\s*=\s*(["\'])([^\']+)\2[^>]*>',
                 webpage, 'initial data', default=None, group=3)
             if initial_data:
-                # let's see if it actually is, or ExtractorError
-                initial_data = self._parse_json(initial_data, 'initial data')
-
-                def _extract_pid(data):
-                    if isinstance(data, dict):
-                        if data.get('kind', '') == 'programme' and 'versionID' in data:
-                            return data['versionID']
-                        else:
-                            for k in data:
-                                pid = _extract_pid(data[k])
-                                if pid:
-                                    return pid
-                    elif isinstance(data, list):
-                        for v in data:
-                            pid = _extract_pid(v)
-                            if pid:
-                                return pid
-                    return None
-                programme_id = _extract_pid(initial_data)
+                # let's see if it actually is, or warning
+                initial_data = self._parse_json(initial_data, 'initial data', fatal=False)
+                if initial_data:
+                    try:
+                        programme_id = initial_data['initData']['items'][0]['clipPID']
+                    except (TypeError, KeyError, IndexError) as ex:
+                        self.report_warning('initial data: exception %s getting pid from JSON' % str(ex))
+                        programme_id = None
 
         if programme_id:
             formats, subtitles = self._download_media_selector(programme_id)
--- a/youtube_dl/extractor/bbc.py
+++ b/youtube_dl/extractor/bbc.py
@@ -17,6 +17,7 @@ from ..utils import (
     parse_iso8601,
     try_get,
     unescapeHTML,
+    unified_timestamp,
     url_or_none,
     urlencode_postdata,
     urljoin,
@@ -968,11 +969,41 @@ class BBCIE(BBCCoUkIE):
                 # let's see if it actually is, or warning
                 initial_data = self._parse_json(initial_data, 'initial data', fatal=False)
                 if initial_data:
-                    try:
-                        programme_id = initial_data['initData']['items'][0]['clipPID']
-                    except (TypeError, KeyError, IndexError) as ex:
-                        self.report_warning('initial data: exception %s getting pid from JSON' % str(ex))
-                        programme_id = None
+                    initial_data = try_get(initial_data, lambda x: x['initData']
+                    if initial_data:
+                        smp_data = initial_data.get('smpData',{})
+                        clip_data = try_get(smp_data, lambda x: x['items'][0], dict)
+                        if clip_data:
+                            initial_data.update(smp_data)
+                            initial_data.update(clip_data)
+                            programme_id = initial_data.get('versionID')
+            if programme_id:               
+                 alt_title = initial_data.get('shortTitle', playlist_id)
+                 title = initial_data.get('title', alt_title)
+                 if title == alt_title:
+                     alt_title = None
+                 description = dict_get(initial_data, ['summary', 'shortSummary'])
+                 timestamp = unified_timestamp(initial_data.get('displayDate'))
+                 duration = initial_data.get('duration')
+                 thumbnail = initial_data.get('holdingImageURL')
+                 if thumbnail:
+                     thumbnail = thumbnail.replace('$recipe', '64x64')
+                     thumbnails = [{'url': thumbnail.replace('$recipe', '%dx%d' % (x, x)),
+                                    'width': x, 'height': x}
+                                   for x in [16, 32, 64, 128, 512]]
+                 else:
+                     thumbnails = None
+                 formats, subtitles = self._download_media_selector(programme_id)
+                 if formats:
+                     self._sort_formats(formats)
+                     vars = {'title', 'description', 'timestamp', 'formats', 
+                             'subtitles', 'duration', 'alt_title'}
+                     ret = {'id': programme_id}
+                     ret.update([[x, locals().get(x)] for x in vars if locals().get(x)])
+                     return ret
+                 else:
+                     self.report_warning('BBC Reel: no formats available', programme_id)
+                     programme_id = None
 
         if programme_id:
             formats, subtitles = self._download_media_selector(programme_id)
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/bbc.py
```diff
@@ -959,6 +959,16 @@ def _real_extract(self, url):
              r'videoId\s*:\s*["\'](%s)["\']' % self._ID_REGEX],
             webpage, 'vpid', default=None)
 
+        # bbc reel (e.g. https://www.bbc.com/reel/video/p07c6sb6/how-positive-thinking-is-harming-your-happiness)
+        initial_data = self._search_regex(
+            r'<script[^>]+id="initial-data"[^>]+data-json=\'(.+)\'>',
+            webpage, 'initial data', fatal=False, default=None)
```
remitamine, line 965: don't use both `fatal` and `default`.

youtube_dl/extractor/bbc.py
```diff
@@ -959,6 +959,16 @@ def _real_extract(self, url):
...
+        if initial_data:
+            programme_id = self._search_regex(
+                r'"versionID":"(%s)"' % self._ID_REGEX,
+                unescapeHTML(initial_data),
+                'programme id', fatal=False, default=None)
```
remitamine, line 970: parse json data and extract `versionID` from there.

remitamine (changes requested):

youtube_dl/extractor/bbc.py
dirkf: Sure thing. Looks like ajj8 didn't read the spec of `_search_regex`. Maybe I didn't either ...

Obviously not using `fatal=False` is what was intended: https://github.com/dirkf/youtube-dl-1/commit/b61ab55aacec21505c763b769d60a1c5ac8691a1.

youtube_dl/extractor/bbc.py
dirkf: The json is deeply nested. Something like JSONPath would have helped. I used an auxiliary function to parse it, but really it would have been easier just to pull the pid from the URI: https://github.com/dirkf/youtube-dl-1/commit/cccbb670f0fe1546c6e62f8da7fe545dc50bb9f1

youtube_dl/extractor/bbc.py
```diff
@@ -959,6 +959,32 @@ def _real_extract(self, url):
              r'videoId\s*:\s*["\'](%s)["\']' % self._ID_REGEX],
             webpage, 'vpid', default=None)
 
+        # bbc reel (e.g. https://www.bbc.com/reel/video/p07c6sb6/how-positive-thinking-is-harming-your-happiness)
+        if not programme_id:
+            initial_data = self._search_regex(
```
remitamine, line 964: use `_html_search_regex` and get rid of `unescapeHTML`.

youtube_dl/extractor/bbc.py
```diff
@@ -959,6 +959,32 @@ def _real_extract(self, url):
...
+                r'<script[^>]+id="initial-data"[^>]+data-json=\'(.+)\'>',
```
remitamine, line 965: this can match outside the script element and can be easily broken if an attribute follow `data-json`.

youtube_dl/extractor/bbc.py
```diff
@@ -959,6 +959,32 @@ def _real_extract(self, url):
...
+                webpage, 'initial data', default=None)
+            if initial_data:
+                # let's see if it actually is, or ExtractorError
+                initial_data = self._parse_json(unescapeHTML(initial_data), 'initial data')
```
remitamine, line 969: should not break the extraction if it wasn't able to parse json data.

youtube_dl/extractor/bbc.py
```diff
@@ -959,6 +959,32 @@ def _real_extract(self, url):
...
+                                pid = _extract_pid(data[k])
+                                if pid:
+                                    return pid
+                    elif isinstance(data, list):
+                        for v in data:
+                            pid = _extract_pid(v)
+                            if pid:
+                                return pid
+                    return None
+                programme_id = _extract_pid(initial_data)
```
remitamine, line 986: this is too generic and can lead to picking the wrong video.

remitamine (changes requested):
Add a test.

youtube_dl/extractor/bbc.py
dirkf: So ajj8 should have known about `_html_search_regex` and the regex attribute matching should at least use `/attr='[^']+'/`. I've given it what should be exact according to W3C HTML5 8.1.2, which might be wise as they're already using both "" and '' attribute values in the same tag.

https://github.com/dirkf/youtube-dl-1/commit/2dcd5c9437d20d43b94180ae57d4d0807f583cdc

youtube_dl/extractor/bbc.py
dirkf: I'm not convinced. At this point we have no pid, because of the test I added. Failing here means the user/developer will be told that the page looked like a Reel page (had a `<script>` with `id` `initial-data` and a `data-json` attribute -- but maybe that's not sufficiently distinctive?) but the content couldn't be extracted. Otherwise they get a more generic 'No extractor for page' error, no?

remitamine: > I'm not convinced. At this point we have no pid, because of the test I added. Failing here means the user/developer will be told that the page looked like a Reel page (had a `<script>` with `id` `initial-data` and a `data-json` attribute -- but maybe that's not sufficiently distinctive?) but the content couldn't be extracted. Otherwise they get a more generic 'No extractor for page' error, no?

no, they would get warning, and then proceed to try other methods that might success.
and you actually should extract more metadata from the `initial_data` not just the pid and then return immidiantly.

dirkf: OK, I've made that a warning and added one for not actually finding a pid in the JSON.

remitamine: no, as i said you would extract the metadata and return immediately.

youtube_dl/extractor/bbc.py
dirkf: Thanks, as far as I can see, we need the `clipPID` from the first element of the JSON, whose `items` seems to be a playlist starting at the current video. ajj8 went for `['smpData']['items'][0]['versionID'].

https://github.com/dirkf/youtube-dl-1/commit/35bb60e7488ae73db6c49737aaf98824086f1108

I've actually watched https://www.bbc.com/reel/video/p0931jtk/the-nuclear-lighthouses-built-by-the-soviets-in-the-arctic, downloaded with the code of the commit above and it matches the `['smpData']['items'][0]['duration']: I guess the `versionID` and `clipPID` both resolve to the same formats page. 

```
# ffprobe -hide_banner "The_nuclear_lighthouses_built_by_the_Soviets_in_the_Arctic_-_BBC_Reel.mp4"
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from '/media/drive1/Video The_nuclear_lighthouses_built_by_the_Soviets_in_the_Arctic_-_BBC_Reel.mp4':
  Metadata:
    major_brand     : isom
    minor_version   : 512
    compatible_brands: isomiso2avc1mp41
    encoder         : Lavf58.20.100
  Duration: 00:05:24.48, start: 0.000000, bitrate: 831 kb/s
    Stream #0:0(und): Video: h264 (Main) (avc1 / 0x31637661), yuv420p(tv, bt709), 704x396 [SAR 1:1 DAR 16:9], 730 kb/s, 25 fps, 25 tbr, 90k tbn, 50 tbc (default)
    Metadata:
      handler_name    : VideoHandler
    Stream #0:1(eng): Audio: aac (HE-AAC) (mp4a / 0x6134706D), 48000 Hz, stereo, fltp, 94 kb/s (default)
    Metadata:
      handler_name    : SoundHandler
```

Again, as the pid is in the URI, that might have been an easier target.

youtube_dl/extractor/bbc.py
dirkf: As below.

dirkf:
Here's a version with a test that succeeds and returns with metadata after parsing the JSON.

However this seems much simpler, although it finds no description:
```
        if not programme_id:
            programme_id = self._search_regex(
                r'/reel/video/(?P<id>%s)/' % self._ID_REGEX, url, 'Reel pid', default=None)
            if programme_id:
                return self.url_result(
                            'https://www.bbc.co.uk/programmes/%s' % programme_id,
                            ie=BBCCoUkIE.ie_key())

File: youtube_dl/extractor/bbc.py (hunk 5 above)
Comment (dirkf) [reply youtube_dl/extractor/bbc.py]: Implemented in  https://github.com/ytdl-org/youtube-dl/pull/28268/commits/db1096f4b92fbf1be6f8393f46b034f8bc09f8bb

File: youtube_dl/extractor/bbc.py (hunk 6 above)
Comment (dirkf) [reply youtube_dl/extractor/bbc.py]: On further investigation `versionID` is definitely the correct PID.

The implementation in https://github.com/ytdl-org/youtube-dl/pull/28268/commits/db1096f4b92fbf1be6f8393f46b034f8bc09f8bb goes for `initData['items'][0]`. If that is a `dict` and has `smpData` and that in turn has `['items'][0]` as a `dict`, the `smpData` and its child `dict` are merged into the parent `dict` from which the metadata is pulled.

File: youtube_dl/extractor/bbc.py [hunk 7]
```diff
@@ -959,6 +973,54 @@ def _real_extract(self, url):
...
+                if formats:
+                    self._sort_formats(formats)
+                    vars = {'title', 'description', 'timestamp', 'duration', 'alt_title',
+                            'thumbnail', 'thumbnails', 'formats', 'subtitles'}
+                    ret = {'id': programme_id}
+                    ret.update([[x, locals().get(x)] for x in vars if locals().get(x)])
+                    return ret
+                else:
+                    self.report_warning('BBC Reel: no formats available', programme_id)
+                    programme_id = None
```
remitamine, line 1022: not finding formats should and error, and it should be raised directly from `_sort_formats`.

youtube_dl/extractor/bbc.py
```diff
@@ -959,6 +973,54 @@ def _real_extract(self, url):
...
+                if thumbnail:
+                    thumbnail = thumbnail.replace('$recipe', '64x64')
+                    thumbnails = [
+                        {'url': thumbnail.replace('$recipe', '%dx%d' % (x, x)),
+                         'width': x, 'height': x}
+                        for x in [16, 32, 64, 128, 512]
+                    ]
+                else:
+                    thumbnails = None
+                formats, subtitles = self._download_media_selector(programme_id)
```
remitamine, line 1012: extract mandatory information before optional information.

youtube_dl/extractor/bbc.py
```diff
@@ -959,6 +973,54 @@ def _real_extract(self, url):
...
+                if formats:
+                    self._sort_formats(formats)
+                    vars = {'title', 'description', 'timestamp', 'duration', 'alt_title',
+                            'thumbnail', 'thumbnails', 'formats', 'subtitles'}
```
remitamine, line 1016: do not set both `thumbnail` and `thumbnail`.

remitamine (changes requested):

youtube_dl/extractor/bbc.py
dirkf: And it is, I see. You're saying that once we've got this far the extraction is committed and it doesn't make sense to trap the error as a warning and give further extraction techniques a chance?

I'll resolve this together with your next comment.

youtube_dl/extractor/bbc.py
dirkf: OK, let's get the `formats` and lose in `_sort_formats` if there aren't any. Then get the remaining metadata.

https://github.com/dirkf/youtube-dl-1/commit/b71cadbad4a0e2784248dbbae00f418bb34bcf12

Also includes the next comment.

youtube_dl/extractor/bbc.py
dirkf: OK -- maybe add that guidance in the comments under `class InfoExtractor` (common.py)?

Fixed above.

dirkf:
Thanks for the guidance and topping and tailing the result. I did have the feeling that you knew how to implement it yourself and actually that might have been easier.

PRs ought to come with tests but I couldn't see any spec of how to populate the `_TESTS` list. Extractor mods are more likely to come with tests if it's obvious how to do it -- maybe put it in the template?

remitamine:
> PRs ought to come with tests but I couldn't see any spec of how to populate the _TESTS list. Extractor mods are more likely to come with tests if it's obvious how to do it -- maybe put it in the template?

`_TESTS` is simply an array of test cases that has the same spec as `_TEST` dict that has an exmple in step 4 of [https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site](Adding support for a new site section).

PR #28332, opened 2021-03-03
Get HD formats for Reel

dirkf:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---
Per https://github.com/ytdl-org/youtube-dl/issues/21870#issuecomment-789266305, we don't get the highest resolution formats through the `BBCIE` extractor. This patch gets the metadata that is available from the Reel page and then uses the `transparent_url` mechanism to get the formats, which now include 1280x720 as well as the 960x540 and lower available from the BBCIE extractor.

### DIFF
--- a/youtube_dl/extractor/bbc.py
+++ b/youtube_dl/extractor/bbc.py
@@ -1012,6 +1012,7 @@ class BBCIE(BBCCoUkIE):
                 topic_title = init_data.get('topicTitle')
                 ret = {
                     'title': smp_data.get('title', playlist_id),
+                    'id': version_id,
                     'alt_title': init_data.get('shortTitle'),
                     'thumbnail': image_url.replace('$recipe', 'raw') if image_url else None,
                     'description': smp_data.get('summary') or init_data.get('shortSummary'),
@@ -1025,7 +1026,6 @@ class BBCIE(BBCCoUkIE):
                     formats, subtitles = self._download_media_selector(version_id)
                     self._sort_formats(formats)
                     ret.update({
-                            'id': version_id,
                             'formats': formats,
                             'subtitles': subtitles,
                         })
--- a/youtube_dl/extractor/bbc.py
+++ b/youtube_dl/extractor/bbc.py
@@ -995,6 +995,8 @@ class BBCIE(BBCCoUkIE):
             }
 
         # bbc reel (e.g. https://www.bbc.com/reel/video/p07c6sb6/how-positive-thinking-is-harming-your-happiness)
+        programme_id = self._search_regex( 
+            r'/reel/video/(?P<id>%s)/' % self._ID_REGEX, url, 'Reel pid', default=None)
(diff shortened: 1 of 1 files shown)


remitamine:
no, using the default `_MEDIA_SETS` from `BBCCoUkIE` should be enough instead of adding another unnecessary request.

youtube_dl/extractor/bbc.py
```diff
@@ -995,6 +995,8 @@ def _real_extract(self, url):
             }
 
         # bbc reel (e.g. https://www.bbc.com/reel/video/p07c6sb6/how-positive-thinking-is-harming-your-happiness)
+        programme_id = self._search_regex( 
+            r'/reel/video/(?P<id>%s)/' % self._ID_REGEX, url, 'Reel pid', default=None)
```
remitamine: unrelated.

youtube_dl/extractor/bbc.py
```diff
@@ -1006,12 +1008,13 @@ def _real_extract(self, url):
...
             if version_id:
                 title = smp_data['title']
+                # also try for higher resolutions
+                self._MEDIA_SETS.insert(0,'iptv-all')
                 formats, subtitles = self._download_media_selector(version_id)
                 self._sort_formats(formats)
                 image_url = smp_data.get('holdingImageURL')
                 display_date = init_data.get('displayDate')
                 topic_title = init_data.get('topicTitle')
-
```
remitamine, line 1014: unrelated.

dirkf:
I thought it would and that was my first pass, but it didn't work for me. It does when (logically enough) the value was set *before* `_download_media_selector`, as in the latest version.

Presumably there's a good reason why this isn't the default for BBCIE, relative to the comment at https://github.com/dirkf/youtube-dl/blob/49d83ee5022be269dbacc5e1300d331350f1a125/youtube_dl/extractor/bbc.py#L54.

Is this robust against failing to fetch the `iptv-all` sets?

Is there a danger of returning formats that the user can't actually reach?

youtube_dl/extractor/bbc.py
```diff
@@ -1006,12 +1008,13 @@ def _real_extract(self, url):
             version_id = clip_data.get('versionID')
             if version_id:
                 title = smp_data['title']
+                # also try for higher resolutions
+                self._MEDIA_SETS.insert(0, 'iptv-all')
```
remitamine, line 1012: as the `iptv-all` gives better quality in general, just drop `_MEDIA_SETS` from the extractor and just use the default value from `BBCCoUkIE`.

remitamine (changes requested):

Vangelis66:
Please kindly consider what's detailed in https://github.com/ytdl-org/youtube-dl/issues/21870#issuecomment-790001948 ; thanks in advance!

PR #23438, opened 2019-12-17
Fix bbc.co.uk playlist support

djcsdy:
### Before submitting a *pull request* make sure you have:
- [x] At least skimmed through [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

This PR fixes two problems:

 * Downloading playlists from BBC iPlayer did not work at all due to the page layout changing completely since the extractor was written.
 * Downloading playlists from other parts of bbc.co.uk was partially broken – only the first page would be downloaded due to some whitespace being inserted between elements since the extractor was written.

Example URLs:

 * https://www.bbc.co.uk/programmes/b00mfl7n/clips (All 142 clips should be downloaded, not just the first 24)
 * https://www.bbc.co.uk/programmes/b00mfl7n/clips?page=3 (Only the 24 clips on the specified page should be downloaded)
 * https://www.bbc.co.uk/iplayer/episodes/b07xd230/josh (All 18 episodes should be downloaded)
 * https://www.bbc.co.uk/iplayer/episodes/b07xd230/josh?seriesId=p047lvcs (Only the six episodes of season 2 should be downloaded)

Issue #15710, opened 2018-02-26
BBC iPlayer PlayList Unable to extract title/description


yepster:
## Please follow the guide below

- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *issue* (like this: `[x]`)
- Use the *Preview* tab to see what your issue will actually look like

---

### Make sure you are using the *latest* version: run `youtube-dl --version` and ensure your version is *2018.02.26*. If it's not, read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [x] I've **verified** and **I assure** that I'm running youtube-dl **2018.02.26**

### Before submitting an *issue* make sure you have:
- [x] At least skimmed through the [README](https://github.com/rg3/youtube-dl/blob/master/README.md), **most notably** the [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
- [x] Checked that provided video/audio/playlist URLs (if any) are alive and playable in a browser

### What is the purpose of your *issue*?
- [x] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---

### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your *issue*

---

### If the purpose of this *issue* is a *bug report*, *site support request* or you are not completely sure provide the full verbose output as follows:

Add the `-v` flag to **your command line** you run youtube-dl with (`youtube-dl -v <your command line>`), copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):

```
youtube-dl -v --yes-playlist  https://www.bbc.co.uk/iplayer/episodes/b088f8wd
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'-v', u'--yes-playlist', u'https://www.bbc.co.uk/iplayer/episodes/b088f8wd']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2018.02.26
[...]

PR #28360, opened 2021-03-05
[bbc] Fix BBCCoUkIPlayerPlaylistIE

dirkf:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

The `BBCCoUkIPlayerPlaylistIE` extractor was failing, whereas `BBCCoUkPlaylistIE` still worked for the same pids: eg, the `_TESTS`.

In the iPlayer series pages, the playlist metadata is now sent in a JSON expression within a <script> element, and this patch extracts the `title` and `description` from it, and modifies the `_VIDEO_ID_TEMPLATE` to match the episode links within the `entities` subobject of the JSON.

Also, the regex used to get the next page link for `BBCCoUkPlaylistBaseIE` didn't allow for a newline that is now sent between tags.

For multi-series iPlayer playlists, each series page is also processed.

For multi-page iPlayer playlists, the pagination regex is extended to match the "Next Page" link on these pages.

### DIFF
--- a/youtube_dl/extractor/bbc.py
+++ b/youtube_dl/extractor/bbc.py
@@ -1318,7 +1318,7 @@ class BBCCoUkPlaylistBaseIE(InfoExtractor):
             if single_page:
                 return
             next_page = self._search_regex(
-                r'<li[^>]+class=(["\'])pagination_+next\1[^>]*>\s*<a[^>]+href=(["\'])(?P<url>(?:(?!\2).)+)\2',
+                r'(?:<li[^>]+class=(["\'])pagination_+next\1[^>]*>\s*<a|<a[^>]+\baria-label=(["\'])Next Page\2)[^>]+href=(["\'])(?P<url>(?:(?!\3).)+)\3',
                 webpage, 'next page url', default=None, group='url')
             if not next_page:
                 break
@@ -1374,6 +1374,15 @@ class BBCCoUkIPlayerPlaylistIE(BBCCoUkPlaylistBaseIE):
             'description': 'A trusted GP sees her charmed life explode when she suspects her husband of an affair.',
         },
         'playlist_mincount': 10,
+    }, {
+        # Playlist with more than one page
+        'url': 'https://www.bbc.co.uk/iplayer/episodes/m0004c4v/beechgrove',
+        'info_dict': {
+            'id': 'm0004c4v',
+            'title': 'Beechgrove',
+            'description': 'Gardening show that celebrates Scottish horticulture and growing conditions.',
+        },
+        'playlist_mincount': 37,
     }]
 
     def _entries(self, webpage, url, playlist_id):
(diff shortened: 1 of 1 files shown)


remitamine:
do mix multiple changes in a single PR.

remitamine:
- does not handle programmes/series with multiple seasons.
- does not handle seasons/groups with multiple pages.

dirkf:
The current version should deal with your comments, but the download tests aren't yet validated: CI is failing because `Certificate did not match expected hostname: files.pythonhosted.org` when installing `nose` with `pip install nose` in the Jython configurations.

remitamine:
as i was looking at this, i think there is a better way to make an optimized version that would reduce the number of requests and still support all the expected features in addition to extracting more metadata for `--flat-playlist` option, i will try to implement it.

dirkf:
Feel free to use the test URLs from this PR for that. See also PR #23438 for a previous solution and set of test cases.

Generally, even though the BBC site support is split into different extractors, the BBCIE extractor in particular implements multiple extraction tactics and it's difficult to know which of these are still valid -- perhaps some telemetry would help, as well as refactoring the over-sized _real_extract() method.

It seems that two types of tactics are currently valid:
* find some JSON in JS that is part of the page; extract links and metadata from  the JSON;
* find certain links, normally with a regex search, but mostly trying to replicate a DOM search similar to JS document.querySelectorAll() without the overhead of DOM.

Maybe generic methods can be created to support these tactics?

I have further updates to fix some of the failing test cases but will hold off if you're reimplementing.

remitamine:
> Generally, even though the BBC site support is split into different extractors, the BBCIE extractor in particular implements multiple extraction tactics and it's difficult to know which of these are still valid -- perhaps some telemetry would help, as well as refactoring the over-sized _real_extract() method.
>
> It seems that two types of tactics are currently valid:
>
>- find some JSON in JS that is part of the page; extract links and metadata from the JSON;
>- find certain links, normally with a regex search, but mostly trying to replicate a DOM search similar to JS document.querySelectorAll() without the overhead of DOM.
>
>Maybe generic methods can be created to support these tactics?

the main BBC Extractor has to be split in the future, this can be done by slowly extracting code that is related to a specific URL pattern into it's own extractor, it would certainly make the code more manageable.

> I have further updates to fix some of the failing test cases but will hold off if you're reimplementing.

it always depend on whether there is a better way of doing thing, in the case of BBC Reel code, even thought i would preferred to introduce the code in a different way, but as the code can be reused in the future, i did accept it with only small changes, but in this case, the method that i added is quite different from the code here, so there aren't much code that can be reused(except for the tests).

dirkf:
The changes I mentioned aren't specific to the iPlayer extractor (fixes to tests, morph playlists, support simorgh playlists).

Regarding the new code, please see https://github.com/ytdl-org/youtube-dl/issues/15710#issuecomment-806033478 for a case where what started out as an iPlayer URL redirected to a `programmes` page in which `_call_api()` returned None and caused a crash.

PR #28361, opened 2021-03-05
rewrite zsh completion generation

pseyfert:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

So far the auto generated youtube-dl.zsh file takes little advantage of zsh's completion infrastructure. This PR changes the generated file to use `_arguments` directly and generate entries like

```
      '(-A --auto-number)'{-A,--auto-number}'' \
       '--write-annotations[Write video annotations to a .annotations.xml file]' \
       '--load-info-json[JSON file containing the video information (created with the "--write-info-json" option)]:FILE:_files' \
       '--cookies[File to read cookies from and dump cookie jar in]:FILE:_files' \
```
This does

 - `'(-A --auto-number)'` : if either `-A` or `--auto-number` are already present on the command line, neither is recommended again. (An option is by default anyway not suggested twice. The parentheses are to also drop the respective other).
  - `[Write video anotations to ...]` These are brief explanation that are shown in the list of completions (such that one doesn't have to switch back and forth between reading the man page and writing the command). It can look (depending on user settings) like that
  ```
  youtube-dl --write-TAB
  user@host:~/coding/youtube-dl > youtube-dl --write-TAB
option:
--write-all-thumbnails  -- Write all thumbnail image formats to disk
--write-annotations     -- Write video annotations to a .annotations.xml file
--write-auto-sub        -- Write automatically generated subtitle file (YouTube only)
--write-description     -- Write video description to a .description file
--write-info-json       -- Write video metadata to a .info.json file
```
 - `:FILE:_files` the trailing colon after the quick help indicates that the flag takes an argument, such that no other flags are suggested until the missing argument to the option is provided The entry between the first and second colon is a help that's maybe shown to the user (depending on settings…) and the zsh completion function after the colon - if given - provides further completion. (The man page zshcompsys explains that one should favour `_files` over `_path_files`, e.g. because a user can better configure their preferences.) If nothing follows the second colon, no automatic completion happens, the user has to type, as in the case of
 ```
       '--external-downloader-args[Give these arguments to the external downloader]:ARGS:' \
```

### DIFF
--- a/devscripts/zsh-completion.in
+++ b/devscripts/zsh-completion.in
@@ -3,26 +3,8 @@
 __youtube_dl() {
     local curcontext="$curcontext" fileopts diropts cur prev
     typeset -A opt_args
-    fileopts="{{fileopts}}"
-    diropts="{{diropts}}"
-    cur=$words[CURRENT]
-    case $cur in
-        :)
-            _arguments '*: :(::ytfavorites ::ytrecommended ::ytsubscriptions ::ytwatchlater ::ythistory)'
-        ;;
-        *)
-            prev=$words[CURRENT-1]
-            if [[ ${prev} =~ ${fileopts} ]]; then
-                _path_files
-            elif [[ ${prev} =~ ${diropts} ]]; then
-                _path_files -/
-            elif [[ ${prev} == "--recode-video" ]]; then
-                _arguments '*: :(mp4 flv ogg webm mkv)'
-            else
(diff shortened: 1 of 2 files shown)


pseyfert:
the completion file how it now gets generated for me
[youtube-dl.zsh.txt](https://github.com/ytdl-org/youtube-dl/files/6094110/youtube-dl.zsh.txt)

Issue #8366, opened 2016-01-29
youtube-dl skips post-processing for M4As


Kaos-Industries:
Hey there.

First of all, thanks for a great program, youtube-dl is the only program on the market I've come across that allows me to extract the audio from my huge YouTube playlists and convert them to M4A ready to be put on my phone. 

The problem I'm currently having is that after extracting the audio, ffmpeg skips the post-processing, leaving me with the original-quality audio files rather than converting them to the 192k bitrate that I need. After posting here in StackOverflow: http://stackoverflow.com/questions/34930699/cant-use-youtube-dl-to-download-specified-bitrate/34930832?noredirect=1#comment57644155_34930832, a user there informed me that I'm probably experiencing another instance of a previous bug that skipped post-processing on M4A files.

Here's what I'm getting when I try to extract and transcode:

```
B:\Users\Hashim>youtube-dl --extract-audio --audio-format m4a --audio-quality 19
2k --playlist-items 48 https://www.youtube.com/playlist?list=PLR
3nWwHlZ9WBpi3uWsjSe6r1PiA8MTbnE
[youtube:playlist] PLR3nWwHlZ9WBpi3uWsjSe6r1PiA8MTbnE: Downloading webpage
[download] Downloading playlist: Main Playlist
[youtube:playlist] PLR3nWwHlZ9WBpi3uWsjSe6r1PiA8MTbnE: Downloading page #1
[youtube:playlist] PLR3nWwHlZ9WBpi3uWsjSe6r1PiA8MTbnE: Downloading page #2
[youtube:playlist] PLR3nWwHlZ9WBpi3uWsjSe6r1PiA8MTbnE: Downloading page #3
[youtube:playlist] PLR3nWwHlZ9WBpi3uWsjSe6r1PiA8MTbnE: Downloading page #4
[youtube:playlist] playlist Main Playlist: Downloading 1 videos
[download] Downloading video 1 of 1
[youtube] -o36bO1XKnw: Downloading webpage
[youtube] -o36bO1XKnw: Downloading video info webpage
[youtube] -o36bO1XKnw: Extracting video information
[youtube] -o36bO1XKnw: Downloading DASH manifest
[download] Destination: B:\Users\Hashim\Desktop\New Folder\Have Mercy - Cigarett
es And Old Perfume.m4a
[download] 100% of 6.24MiB in 00:13
[ffmpeg] Correcting container in "B:\Users\Hashim\Desktop\New Folder\Have Mercy -
[...]

PR #20021, opened 2019-02-26
Handle Opus bitrate when quality is specified instead of bitrate

Hattshire:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [x] At least skimmed through [adding new extractor tutorial](https://github.com/rg3/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/rg3/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/rg3/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- ~~I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)~~

### What is the purpose of your *pull request*?
- ~~Bug fix~~
- [x] Improvement
- ~~New extractor~~
- ~~New feature~~

---

### Description of your *pull request* and other information

When converting formats the user would expect that with the default options the resulting file has at least a "normal quality", however when extracting audio from a source not in Opus, and specifying it as the resulting file format on youtube-dl without any other tuning, it produces a low quality file.

Here I used the [Xiph's Recomended Opus Bitrates](https://wiki.xiph.org/Opus_Recommended_Settings#Recommended_Bitrates) to emulate the mising -aq option on the opus codec.

PR #28430, opened 2021-03-12
Attempt to fix audio postprocessing bug (#8366)

cgillesp:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This PR is an attempt to fix issue #8366, a bug in FFmpegExtractAudioPP post-processor that makes it impossible to use youtube-dl to convert extracted audio to a different bitrate in the same format. 

The current behavior never converts a downloaded audio file if the desired format is the same as the original. However, if a user wants to do other post-processing, like changing the bitrate or mixing down a stereo audio file to mono while retaining the same format, the post-processor may still skip using FFmpeg altogether, or it may override their quality setting.

This PR changes the default value of the --audio-quality option (used only by FFmpegExtractAudioPP) from '5' to None, and only skips re-encoding between the same two formats if the --audio-quality option is None (i.e. the user has not explicitly set the audio quality) and the user has no custom --postprocessor-args. The default behavior (still VBR quality of 5) is now handled within the post-processor itself.  If the new file has the same filename as the old one, it will encode to filename.temp.mp3 and then overwrite the original filename.mp3. 

**I would like feedback on this PR, as I am still not entirely satisfied with it.**

In particular, I am not satisfied with the choice to simply re-encode whenever the user explicitly sets a quality, even if it is better than the current quality. This makes bigger files with quality that is actually worse. Fixing this would I think require mucking around with FFprobe to get the current quality, and even then I don't think it can account for VBR qualities.

I also don't know how overwriting the processed files should interact with the --keep-video (or audio in this case) flag. In the PR, if the new file would have the same name (and --no-post-overwrites is false), the old file is simply overwritten.

**Also:** in the course of writing this I discovered the bug fixed by the unmerged PR #20021. It causes opus files with a VBR quality setting to be encoded at that integer kbps (ie for the default of 5 it will encode at 5kbps). You should merge that PR too (it does conflict somewhat, so I can work on that if you want).

Thanks for all the great work,

- Charlie

### DIFF
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -792,7 +792,7 @@ def parseOpts(overrideArguments=None):
         help='Specify audio format: "best", "aac", "flac", "mp3", "m4a", "opus", "vorbis", or "wav"; "%default" by default; No effect without -x')
     postproc.add_option(
         '--audio-quality', metavar='QUALITY',
-        dest='audioquality', default='5',
+        dest='audioquality', default=None,
         help='Specify ffmpeg/avconv audio quality, insert a value between 0 (better) and 9 (worse) for VBR or a specific bitrate like 128K (default %default)')
     postproc.add_option(
         '--recode-video',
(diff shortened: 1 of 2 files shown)


pukkandan:
imo if the user wants to recode the video always, they should be using `--exec` and not `-x`

cgillesp:
> imo if the user wants to recode the audio always, they should be using `--exec` and not `-x`

Fair enough, but I don't think a pattern of use should break just because the extractor happens to give back a different format. I feel like a big part of the utility of youtube-dl is abstracting away those details and doing what the user wants.

Issue #23455, opened 2019-12-19
[BBC] no video formats found


basicmaster:
## Checklist

- [X] I'm reporting a broken site support
- [X] I've verified that I'm running youtube-dl version **2019.11.28**
- [X] I've checked that all provided URLs are alive and playable in a browser
- [X] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [X] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
$ youtube-dl -v https://www.bbc.com/news/av/embed/p07xmg48/50670843
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: [u'-v', u'https://www.bbc.com/news/av/embed/p07xmg48/50670843']
[debug] Encodings: locale UTF-8, fs UTF-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2019.11.28
[debug] Python version 2.7.12 (CPython) - Linux-4.15.0-72-generic-x86_64-with-Ubuntu-16.04-xenial
[debug] exe versions: avconv 2.8.15, avprobe 2.8.15, ffmpeg N-95296-g1cdb7c5, ffprobe N-95296-g1cdb7c5, rtmpdump 2.4
[debug] Proxy map: {}
[bbc] 50670843: Downloading webpage
[bbc] p07xmg48: Downloading media selection XML
ERROR: No video formats found; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/filme/yt/youtube-dl/youtube_dl/YoutubeDL.py", line 796, in extract_info
    ie_result = ie.extract(url)
  File "/filme/yt/youtube-dl/youtube_dl/extractor/common.py", line 530, in extract
    ie_result = self._real_extract(url)
  File "/filme/yt/youtube-dl/youtube_dl/extractor/bbc.py", line 937, in _real_extract
    self._sort_formats(formats)
  File "/filme/yt/youtube-dl/youtube_dl/extractor/common.py", line 1327, in _sort_formats
    raise ExtractorError('No video formats found')
ExtractorError: No video formats found; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
```

## Description

[...]

PR #28547, opened 2021-03-26
[bbc] Support embedded video with data in playlistObject of playerSettings

dirkf:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Supports av-embed URL as noted in issue #23455. The video data is encoded in the `playlistObject` element of a JS object assigned to `window.playerSettings`. After some processing, the object is transformed into JSON that can be extracted in the standard way. As it's very like the `data-playable` case, the common processing is separated into a new method.


youtube_dl/extractor/bbc.py
```diff
@@ -958,7 +977,31 @@ def _real_extract(self, url):
...
+        else:
+            # embed video with playerSettings, eg
+            # https://www.bbc.com/news/av/embed/p07xmg48/50670843
+            settings = self._html_search_regex(
+                r'<script\b[^>]+>.+\.playerSettings\s*=\s*(?P<json>\{.*\})\s*(?:,\s*function\s*\(\s*\)\s*\{\s*["\']use strict.+\(\s*\)\s*)?</script\b',
+                webpage, 'player settings', default='{}', group='json')
+            settings = re.sub(r'([{,]\s*)(\w+)(\s*:)', r'\1"\2"\3', settings)
+            for pling in (('!0', 'true'), ('!1', 'false')):
+                settings = re.sub(r'(:\s*)(%s)(\s*[,}])' % pling[0], r'\1%s\3' % pling[1], settings)
+            settings = self._parse_json(settings, playlist_id, fatal=False)
```
dstftw, line 989:
```suggestion
            settings = self._parse_json(settings, playlist_id, transform_source=js_to_json, fatal=False)
```

youtube_dl/extractor/bbc.py
```diff
@@ -912,23 +938,16 @@ def _real_extract(self, url):
...
+                        entry = self._extract_from_playlist_object(playlist_object)
+                        programme_id = entry.get('id')
+                        formats, subtitles = self._download_media_selector(programme_id)
+                        self._sort_formats(formats)
+                        entry.update({
+                            'timestamp': timestamp,
+                            'formats': formats,
+                            'subtitles': subtitles,
+                        })
+                        entries.append(entry)
```
dstftw, line 950: These looks like candidates for generalization and extracting into a separate method.

youtube_dl/extractor/bbc.py
```diff
@@ -958,7 +977,31 @@ def _real_extract(self, url):
...
+                    formats, subtitles = self._download_media_selector(programme_id)
+                    self._sort_formats(formats)
+                    thumbnail = playlist_object.get('holdingImageURL')
+                    entry.update({
+                        'timestamp': timestamp,
+                        'formats': formats,
+                        'subtitles': subtitles,
+                        'thumbnail': thumbnail.replace('$recipe', 'raw') if thumbnail else None,
+                    })
+                    entries.append(entry)
```
dstftw, line 1004: These looks like candidates for generalization and extracting into a separate method.

dstftw (changes requested):

youtube_dl/extractor/bbc.py
dirkf: That's a much better solution; now I know.

youtube_dl/extractor/bbc.py
dirkf: I created `_get_playlist_entry(entry) -> entry` method to encapsulate the key operations with the benefit of returning `None` when the given `entry` has no `'id'` and checking that result.

youtube_dl/extractor/bbc.py
dirkf: As above.

Issue #27780, opened 2021-01-12
BBC weather


SKWDiesel1:
## Checklist

- [ x] I'm reporting a broken site support
- [x ] I've verified that I'm running youtube-dl version **2021.01.08**
- [x ] I've checked that all provided URLs are alive and playable in a browser
- [x ] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x ] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
PASTE VERBOSE LOG HERE
```
youtube-dl --verbose https://www.bbc.co.uk/weather/features/55581056
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['--verbose', 'https://www.bbc.co.uk/weather/features/55581056']
[debug] Encodings: locale UTF-8, fs utf-8, out utf-8, pref UTF-8
[debug] youtube-dl version 2021.01.08
[debug] Python version 3.9.1 (CPython) - Linux-5.4.88-1-lts-x86_64-with-glibc2.32
[debug] exe versions: ffmpeg 4.3.1, ffprobe 4.3.1, rtmpdump 2.4
[debug] Proxy map: {}
[bbc] 55581056: Downloading webpage
ERROR: Unable to extract playlist data; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/youtube_dl/YoutubeDL.py", line 803, in wrapper
    return func(self, *args, **kwargs)
  File "/usr/lib/python3.9/site-packages/youtube_dl/YoutubeDL.py", line 824, in __extract_info
    ie_result = ie.extract(url)
  File "/usr/lib/python3.9/site-packages/youtube_dl/extractor/common.py", line 532, in extract
    ie_result = self._real_extract(url)
  File "/usr/lib/python3.9/site-packages/youtube_dl/extractor/bbc.py", line 1174, in _real_extract
    self._search_regex(
  File "/usr/lib/python3.9/site-packages/youtube_dl/extractor/common.py", line 1010, in _search_regex
    raise RegexNotFoundError('Unable to extract %s' % _name)
[...]

PR #28577, opened 2021-03-28
[bbc] Fix and improve extraction for pages with Morph-based data

dirkf:

---

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

In Morph-based pages the target is JSON in a <script> element containing `Morph.setPayload(arg1, arg2);`

The first argument of Morph.setPayload() has to be matched to pick the correct instance of the call. In that case, the second argument contains the target JSON.

Morph-based articles were previously implemented but looked for a `body.components` member of the JSON object that no longer seems to be sent. Instead `body` and `body.media` have to be used. A similar format is used on some BBC Weather pages, so these can also be processed.

Fixes these tests:
http://www.bbc.com/sport/0/football/34475836
http://www.bbc.co.uk/sport/rowing/35908187

Adds this test (fixes issue #27780):
https://www.bbc.co.uk/weather/features/55581056


remitamine:
@dirkf when the code needs to be reviewed, `Fix and improve extraction for pages with Morph-based data` is broad, BBC uses the morph code in multiple places in their website, and every part is different then the others(has different data layout), so to make things easier for you and me, it would be better to do one clear think in every PR.
the PR can be split into:
- fix support for Sport articles.
- fix support for Weather Features.
- add support for Bite-size guides.

you have also added some code related to `supportingMedia`, when introducing new changes you should point the test case that is related to this change or add a new one if it's not already covered, and as this change is also separate from the other ones listed above, it should be separated as well.

dirkf:
Just lost my response to your previous which would have helped to clarify the update. So again ...

You wrote:

>     breaks the old morph code(components).

The Olympics clip that was used to verify the original Morph.setPayload() tactic is no longer valid. This page https://www.bbc.co.uk/news/live/world-52466471 has the format,  but no `leadMedia` (ie, value false), only `supportingMedia`, a list of items of the same structure as media value of `leadMedia`: I couldn't find a page that used the same format with media in `leadMedia`, so by falling back to the first item in `supportingMedia` we can see that the components tactic is working. `supportingMedia` might also be extracted as a playlist _if there were a way to override the default for `noplaylist`_.

Given a test URL with valid `leadMedia`, that URL can be used instead of the Olympics clip, and the `supportingMedia` fallback could be removed if desired. Or a new test case could be added using a URL like the one quoted above.

>     BBC Weather still does not work[1].

It worked in Python 2.7 but weirdly not in 3.x. Normally this is a character code issue, but in this case the issue was `dict.values()`, which is a list in 2 and a view, synonymous with `dict.viewvalues()`, in 3.

>     extracted metadata differs from one run to the other[2].

As the format of the Klopp test video has changed since its test case was formulated, the title can be whatever we want from the titles available. In this case the title in the test case can be found in the document `<title>` element's innerText, while the extraction was returning the title as set in the `title` and `og:title` meta properties. Trivial to change either way, as in the revised version.

The problem with separating out the variations of the `Morph.setPayload()` tactic is that it's not clear which pages will be handled by each variation, and the processing for the variations is basically similar. I hope the revised change-set just pushed is clearer with this rubric.

dirkf:
Fixes second issue of #29926.

PR #28674, opened 2021-04-06
[bbc] Support BBC World (etc) pages with data in SIMORGH_DATA JSON

dirkf:

---

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Pages from BBC World Service (and possibly others) are being served using the BBC's open source [Simorgh](https://github.com/bbc/simorgh) web application framework. This includes two of the existing test case pages with a legacy format as well as pages created directly with Simorgh.

Media links and metadata can be extracted from a JSON object assigned to `window._SIMORGH_DATA`, which carries the application state from the server to the client in a similar way to `_REDUX_STATE` and `Morph.setPayload(uri,json)`.

In principle there could be several extractable media items but in this PR only the first is found. If pages with multiple items are found it would be appropriate and straightforward to modify the extraction to return a playlist in those cases.

Fixes (legacy format):
http://www.bbc.com/turkce/haberler/2015/06/150615_telabyad_kentin_cogu (but video now missing)
http://www.bbc.com/mundo/video_fotos/2015/06/150619_video_honduras_militares_hospitales_corrupcion_aw

Adds (direct format):
http://www.bbc.co.uk/scotland/articles/cm49v4x1r9lo (test page used for Simorgh, should be long-lasting)


youtube_dl/extractor/bbc.py
```diff
@@ -1074,6 +1086,85 @@ def _real_extract(self, url):
...
+                    bitrate = format.pop('bitrate')
+                    if bitrate:
+                        bitrate = re.sub(r'000\s*$', 'kbps', bitrate)
+                        format['tbr'] = parse_bitrate(bitrate)
+                    format['language'] = media.get('language')
+                    # format id: penultimate item from the url split on _ and .
+                    (fmt,) = re.split('[_.]', format['url'])[-2:][:1]
+                    format['format_id'] = '%s_%s' % (format['format'], fmt)
+                    if not format.get('resolution'):
+                        format['resolution'] = fmt
```
dstftw, line 1116: Resolution is used for textual representation of width and height and typically automatically deduced from these meta fields.

youtube_dl/extractor/bbc.py
```diff
@@ -1074,6 +1086,85 @@ def _real_extract(self, url):
...
+                for format in playlist:
+                    if not (format.get('url') and format.get('format')):
+                        continue
+                    bitrate = format.pop('bitrate')
+                    if bitrate:
+                        bitrate = re.sub(r'000\s*$', 'kbps', bitrate)
+                        format['tbr'] = parse_bitrate(bitrate)
+                    format['language'] = media.get('language')
+                    # format id: penultimate item from the url split on _ and .
+                    (fmt,) = re.split('[_.]', format['url'])[-2:][:1]
```
dstftw, line 1113: Why not just use regular expression? It's much easier to understand and figure out the original intention after URL pattern changes in future.

youtube_dl/extractor/bbc.py
```diff
@@ -1074,6 +1086,85 @@ def _real_extract(self, url):
...
+                media.update(media['media'])
+                formats = []
+                keys = {'url', 'format', 'format_id', 'language', 'quality', 'tbr', 'resolution'}
+                for format in playlist:
+                    if not (format.get('url') and format.get('format')):
+                        continue
+                    bitrate = format.pop('bitrate')
+                    if bitrate:
+                        bitrate = re.sub(r'000\s*$', 'kbps', bitrate)
+                        format['tbr'] = parse_bitrate(bitrate)
```
dstftw, line 1110: This looks like overkill. Why not just use `float_or_none`?

youtube_dl/extractor/bbc.py
```diff
@@ -1074,6 +1086,85 @@ def _real_extract(self, url):
...
+                    # format id: penultimate item from the url split on _ and .
+                    (fmt,) = re.split('[_.]', format['url'])[-2:][:1]
+                    format['format_id'] = '%s_%s' % (format['format'], fmt)
+                    if not format.get('resolution'):
+                        format['resolution'] = fmt
+                    format['quality'] = -1
+                    formats.append(dict((k, format[k]) for k in keys))
+                self._sort_formats(formats)
+                return {
+                    'id': media.get('id'),
```
dstftw, line 1121: `id` can't be `None`.

youtube_dl/extractor/bbc.py
```diff
@@ -1074,6 +1086,85 @@ def _real_extract(self, url):
...
+                    'title': (dict_get(media.get('headlines'),
+                                       ('shortHeadline', 'headline'))
+                              or playlist_title),
```
dstftw, line 1124: `title` can't be `None`.

youtube_dl/extractor/bbc.py
```diff
@@ -1074,6 +1086,85 @@ def _real_extract(self, url):
...
+                        vpid = try_get(block, lambda x: x['model']['versions'][0]['versionId'])
+                        if vpid:
+                            formats, subtitles = self._download_media_selector(vpid)
+                            self._sort_formats(formats)
+                            model = block['model']
+                            version = model['versions'][0]
+                            thumbnail = model.get('imageUrl')
+                            return {
+                                'id': vpid,
+                                'title': model.get('title') or 'unnamed clip',
```
dstftw, line 1150: Is this really possible?

dstftw (changes requested):

youtube_dl/extractor/bbc.py
dirkf: Compare `unified_timestamp` vs `parse_iso8601`, though? Who knows what the BBC will send here? But if it ends in `000`, they probably mean kbps ...

youtube_dl/extractor/bbc.py
dirkf: I'll look at that (time having passed), but what I wrote was my best attempt to be resilient against the format not having some resolution text (lo, med, hi) in that position.

youtube_dl/extractor/bbc.py
dirkf: Indeed not, good point.

youtube_dl/extractor/bbc.py
dstftw: Then something like `float_or_none(bitrate, scale=1000) or parse_bitrate(bitrate)` should be used.
Current approach also won't work if `bitrate` is not evenly divisible by 1000.

youtube_dl/extractor/bbc.py
dirkf: In the version that I was working on I think that I believed `playlist_title` was known to be set. I'll check if that's still the case.

youtube_dl/extractor/bbc.py
dirkf: Or throw on `model['title']`. But these pages may have a lot of clips and I don't know whether the BBC will set the title every time.

youtube_dl/extractor/bbc.py
dirkf: That's a good alternative, done.

youtube_dl/extractor/bbc.py
dirkf: I tried various `re.match()` constructions and couldn't reproduce what the comment says is being done by the slice + re syntax. I'd rather leave this as it is.

youtube_dl/extractor/bbc.py
dirkf: In this case the playlist data from the embedded JSON (unlike BBC `programme` playlists) doesn't appear to include video width or height, so these could only be found by parsing a selected media item. The test page has aspect ratio (global) and bitrate for each video. To populate the `resolution` column of `-F` output with the best information available, I try for a `resolution` (just in case) and otherwise use the resolution string `fmt` extracted as above. It might be more realistic to try for width/height instead, which I've implemented in the new version.

youtube_dl/extractor/bbc.py
dirkf: Changed to check for `title` and `id` up front, loop if not.

PR #28729, opened 2021-04-11
[YoutubeDL] Do not abandon playlist upon "format not available" error on video

glenn-slayden:
The downloading of a playlist is currently terminated upon encountering the first "format not available" error for any single video item in the list.

This should be a 'warning' instead, so that the remainder of the playlist or channel can continue downloading.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1671,18 +1671,17 @@ class YoutubeDL(object):
 
         formats_to_download = list(format_selector(ctx))
         if not formats_to_download:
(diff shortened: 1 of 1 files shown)


rautamiekka:
Would make more sense to 1st check if there are formats to download, then act diff when there ain't, rather than the current one where non-existence is checked, then acted diff when there are.

glenn-slayden:
@rautamiekka I agree, but I wanted to disturb the existing logic as little as possible in order to just fix the worst behavior.

dstftw:
This is expected behavior, `-i` is supposed to be used to not stop on errors.

PR #28744, opened 2021-04-12
Improve handling of SIGINT to ease shell scripting with ytdl

HendrikF:
- [ ] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site) - Not an extractor
- [x] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them - seems to focus on extractors
- [ ] Covered the code with tests (note that PRs without tests will be REJECTED) - i have no idea how to test this programmatically
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8) - no output

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This change allows compatible shells to stop script execution when youtube-dl exited because of a SIGINT (often Ctrl-C).
It enables the user to stop a shell script, which executes youtube-dl.

The behaviour is called "wait and cooperative exit" and described at https://www.cons.org/cracauer/sigint.html

Basically we signal the parent process that we exited because of SIGINT when that is the case.
Compatible shells will then stop script execution as the user apparently wishes so.
This works by resetting the signal handler to default and sending ourselves another SIGINT.

I was told that this behaviour might be part of the POSIX Standard, so it should work on other platforms too.
Worst case is no change to the current behaviour.

I only tested it on Ubuntu 18.04, **please test on Windows and Mac.**

Testcase:
```
python youtube-dl/youtube_dl/__main__.py -o test "https://www.youtube.com/watch?v=3QFVLZ1cnOs"
echo "FAILED - script should have exited on SIGINT"
```
When you interrupt this script, the second line should not be executed with this change.
Previously the shell would continue and you had to spam Ctrl-C to stop it - especially when executing a loop.

### DIFF
--- a/youtube_dl/__init__.py
+++ b/youtube_dl/__init__.py
@@ -478,7 +478,13 @@ def main(argv=None):
     except SameFileError:
         sys.exit('ERROR: fixed output name but more than one file to download')
     except KeyboardInterrupt:
-        sys.exit('\nERROR: Interrupted by user')
+        print('\nERROR: Interrupted by user')
+        # 'wait and cooperative exit' handling of SIGINT
+        # this signals our parent that we exited because of SIGINT
+        # the compatible parent shell can then stop script execution
+        import signal
+        signal.signal(signal.SIGINT, signal.SIG_DFL)
+        os.kill(os.getpid(), signal.SIGINT)
 
 
 __all__ = ['main', 'YoutubeDL', 'gen_extractors', 'list_extractors']


pukkandan:
I personally don't think this should be implemented. I use `Ctrl+C` often to exit out of `youtube-dl` without exiting out of the script itself. Pressing `Ctrl+C` twice can already be used to exit out of the script

HendrikF:
That is not a good argument against this change.

Using Ctrl-C twice does not always work, because youtube-dl might be called in a loop and you might be able not interrupt it fast enough, which leaves partial useless files.

You can just use `trap` to ignore SIGINT in your script. You can even check some conditions whether you want to ignore it or not.

You can change my example to this and it would behave like you want. The `echo` still executes then, even with this change to youtube-dl.

```bash
handler () {
  echo ignored
}
trap handler SIGINT

python youtube-dl/youtube_dl/__main__.py -o x "https://www.youtube.com/watch?v=3QFVLZ1cnOs"
echo "script will continue on SIGINT, as the handler ignores it"
```

```
[youtube] 3QFVLZ1cnOs: Downloading webpage
...
[download]  12.1% of 88.95MiB at  8.67MiB/s ETA 00:09^C
ERROR: Interrupted by user
ignored
script will continue on SIGINT, as the handler ignores it
```

pukkandan:
I primarily use windows and was referring to it in my previous comment.
However, after some testing, it doesn't seem like this PR has any effect on windows machines

> Using Ctrl-C twice does not always work, because youtube-dl might be called in a loop and you might not be able to interrupt it fast enough, which leaves partial useless files.

That is indeed problematic behavior. It doesn't happen on windows, so such an issue never crossed my mind

HendrikF:
> However, after some testing, it doesn't seem like this PR has any effect on windows machines

Hm, that is interesting. It is nice though, that my change does not break in a bad way.
I wonder how handling of SIGINT works on Windows in general and if this change would even be necessary.

PR #28831, opened 2021-04-22
Adding improvements to 9Now to handle episodic series

dalanmiller:

---

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

* Currently doesn't handle extracting the episode of season number that are both present in the available JSON. 
* Removes prior tests which no longer work.

### DIFF
--- a/youtube_dl/extractor/ninenow.py
+++ b/youtube_dl/extractor/ninenow.py
@@ -33,7 +33,6 @@ class NineNowIE(InfoExtractor):
             'timestamp': 1619002200,
             'upload_date': '20210421',
         },
-        # 'skip': 'Only available in Australia',
     }]
     BRIGHTCOVE_URL_TEMPLATE = 'http://players.brightcove.net/4460760524001/default_default/index.html?videoId=%s'
 
--- a/youtube_dl/extractor/ninenow.py
+++ b/youtube_dl/extractor/ninenow.py
@@ -10,6 +10,7 @@ from ..utils import (
     smuggle_url,
     unified_timestamp,
     unified_strdate,
+    try_get,
 )
 
 
@@ -51,14 +52,14 @@ class NineNowIE(InfoExtractor):
         'md5': 'e73b986e7efb81c90896f0806e56d7e4',
         'info_dict': {
             'id': '6249614030001',
-            'title': "Episode 3",
+            'title': 'Episode 3',
             'ext': 'mp4',
             'season_number': 3,
             'episode_number': 3,
             'description': 'In the first elimination of the competition, teams will have 10 hours to build a world inside a snow globe.',
-            'uploader_id': "4460760524001",
+            'uploader_id': '4460760524001',
             'timestamp': 1618989402,
-            'upload_date': "20210421",
+            'upload_date': '20210421',
         },
         'skip': 'Only available in Australia',
     }]
@@ -67,16 +68,9 @@ class NineNowIE(InfoExtractor):
     def _real_extract(self, url):
         display_id = self._match_id(url)
         webpage = self._download_webpage(url, display_id)
-        page_data = self._parse_json(
-            self._search_regex(
-                r'window\.__data\s*=\s*({.*?});',
-                webpage,
-                'page data',
-                default='{}'
-            ),
-            display_id,
-            fatal=False
-        )
+        page_data = self._parse_json(self._search_regex(
+            r'window\.__data\s*=\s*({.*?});', webpage,
+            'page data', default='{}'), display_id, fatal=False)
         if not page_data:
             page_data = self._parse_json(self._parse_json(self._search_regex(
                 r'window\.__data\s*=\s*JSON\.parse\s*\(\s*(".+?")\s*\)\s*;',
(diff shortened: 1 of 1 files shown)


dstftw:
Read coding conventions.

dalanmiller:
@dstftw ptal? 

Apologies on missing the `unified_$type` handlers. 

Let me know if you want me to randomly find some additional tests to replace the originals.

youtube_dl/extractor/ninenow.py
```diff
@@ -59,25 +90,41 @@ def _real_extract(self, url):
...
         if video_data.get('drm'):
             raise ExtractorError('This video is DRM protected.', expected=True)
-
         brightcove_id = video_data.get('brightcoveId') or 'ref:' + video_data['referenceId']
         video_id = compat_str(video_data.get('id') or brightcove_id)
-        title = common_data['name']
+
+        # Episode/Season data extraction
+        title = common_data['episode']['name']
+        season_number = common_data.get("season").get("seasonNumber", None)
```
dstftw, line 110: Breaks. Read coding conventions.

youtube_dl/extractor/ninenow.py
```diff
@@ -59,25 +90,41 @@ def _real_extract(self, url):
...
+        episode_number = common_data.get('episode').get("episodeNumber", None)
```
dstftw, line 111: Breaks.

youtube_dl/extractor/ninenow.py
```diff
@@ -8,44 +8,75 @@
...
+        page_data = self._parse_json(
+            self._search_regex(
+                r'window\.__data\s*=\s*({.*?});',
+                webpage,
+                'page data',
+                default='{}'
+            ),
+            display_id,
+            fatal=False
+        )
```
dstftw, line 79: Do not reformat code and remove irrelevant changes.

youtube_dl/extractor/ninenow.py
```diff
@@ -59,25 +90,41 @@ def _real_extract(self, url):
...
+        timestamp = None
+        upload_date = None
+
+        for key in ('airDate', 'availability'):
+            if 'episode' in common_data and key in common_data.get("episode"):
+                if key == 'airDate':
+                    timestamp = unified_timestamp(common_data.get('airDate'))
+
+                if key == 'availability':
+                    upload_date = unified_strdate(common_data.get('availability'))
```
dstftw, line 121: try_get.

youtube_dl/extractor/ninenow.py
```diff
@@ -59,25 +90,41 @@ def _real_extract(self, url):
...
 
         thumbnails = [{
             'id': thumbnail_id,
             'url': thumbnail_url,
-            'width': int_or_none(thumbnail_id[1:])
-        } for thumbnail_id, thumbnail_url in common_data.get('image', {}).get('sizes', {}).items()]
+            'width': int_or_none(thumbnail_id[1:]),
+        } for thumbnail_id, thumbnail_url in common_data.get('episode').get('image', {}).get('sizes', {}).items()]
```
dstftw, line 127: Breaks.

youtube_dl/extractor/ninenow.py
```diff
@@ -86,8 +133,12 @@ def _real_extract(self, url):
                 {'geo_countries': self._GEO_COUNTRIES}),
             'id': video_id,
             'title': title,
-            'description': common_data.get('description'),
+            'description': common_data.get('episode').get('description'),
```
dstftw, line 136: Breaks.

youtube_dl/extractor/ninenow.py
dstftw, line 110: Single quotes.

youtube_dl/extractor/ninenow.py
```diff
@@ -86,8 +133,12 @@ def _real_extract(self, url):
...
             'duration': float_or_none(video_data.get('duration'), 1000),
             'thumbnails': thumbnails,
             'ie_key': 'BrightcoveNew',
+            'season_number': season_number,
+            'episode_number': episode_number,
+            "timestamp": timestamp,
+            "upload_date": upload_date,
```
dstftw, line 143: `update_date` is pointless if `timestamp` is provided.

dstftw (changes requested):

youtube_dl/extractor/ninenow.py
dalanmiller: I see how they are essentially the same information, but this could be useful for easier formatting when templating the output file without having to do any conversion from `timestamp` to date string.

dalanmiller:
To clean it up further, I could also put all the `.get` attempts in the actual `dict` literal that is returned if that is also desirable!

dalanmiller:
ptal? @dstftw

dalanmiller:
Bump @dstftw 🙏🏻

aarubui:
I'd say just consistently use `try_get` instead of `.get().get()` or `common_data[][]`. Or better yet, not defining `common_data` as a dict to avoid nesting. Just use a `season_data` variable and you'll touch fewer lines. The deprecated tests probably aren't worth keeping, either.

dalanmiller:
> I'd say just consistently use try_get (with compat_str instead of str) instead of .get().get() or common_data[][]. Or better yet, not defining common_data as a dict to avoid nesting. Just use a season_data variable and you'll touch fewer lines. The deprecated tests probably aren't worth keeping, either.

Yeah some of these were from before my PR so wasn't sure if I was expected to clean up the whole file. But I will go ahead and do so. 

>  You should be using parse_iso8601

Good tip! 

> Remove the test skipping.

The requesting of this should be region bound. So probably not useful unless the tests can be run in 🇦🇺 .

dalanmiller:
ptal?

aarubui:
I actually meant consistency for the lines you touch. You're not expected to clean up someone else's work. Actually, the holy grail is to make changes that are so minimal that you can tell from reading the diff that, at worst, even if every new line failed, at least the extractor would work as it did previously.

As for the test skipping, `_GEO_COUNTRIES` gets the extractor working worldwide.

dalanmiller:
> As for the test skipping, _GEO_COUNTRIES gets the extractor working worldwide. You can try for yourself by changing 'AU' to 'US' to see it stops working.

Ahh interesting, I've removed the test skip @aarubui 👍

dalanmiller:
ptal? @aarubui

aarubui:
@dalanmiller I should have mentioned this at the start. I'm not on the team. I was just guessing what @dstftw wouldn't like to speed this along and maybe my pulls would get merged. That's not worked.

Issue #28671, opened 2021-04-06
Patreon all posts


makeworld-the-better-one:
## Checklist

- [x] I'm reporting a site feature request
- [x] I've verified that I'm running youtube-dl version **2021.04.01**
- [x] I've searched the bugtracker for similar site feature requests including closed ones

## Description

youtube-dl can currently download Patreon audio posts that are directly linked, like [this one](https://www.patreon.com/posts/memes-as-ep-01-34527413). It would be nice if the Patreon creator URL could be supplied, and all the posts would be found and downloaded, likely as a playlist object. An example of such a URL would be https://www.patreon.com/joshuacitarella. A URL like https://www.patreon.com/joshuacitarella/posts should also be considered valid.

KanuGaba:
@vishnunair6 and I would like to begin work on this issue!

PR #28852, opened 2021-04-23
[patreon:user] Add new extractor

KanuGaba:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Updated patreon.py with a new extractor in order to download a playlist of all the audio posts on a creator's page, given the creator's user-url. This resolves issue #28671 .


vishnunair6:
Pair programmed @vishnunair6

dstftw:
1. Headless browser engines should only be used in extreme cases when scraping is infeasible. In this case everything can be easily extracted via API calls.
2. It does not extract all posts.

vishnunair6:
Hi! We ran into a couple issues using the method described above, which is why we resorted to a headless browser engine. 
1. We looked into a couple API endpoints, the endpoints for this user were https://www.patreon.com/api/user/18719680 and https://www.patreon.com/api/campaigns/2799152, none of which provide valuable information on the posts themself. To get the post ID for any posts, we had to wait for dynamic content to load. The urllib library was very inconvenient and not cooperative while trying to wait for the posts to load. This occurred on both the campaign page and the posts page. In fact, when trying to use download_webpage and download_json on post pages, Patreon thought we were bots and redirected us to a reCaptcha page. We found the headless browser engine the most supportive in circumnavigating these issues. 
2. Some posts are hidden based on whether or not the user has paid for the service. This can be remedied by uncommenting the login method already available. Originally, the developer of this function mentioned that the code works for any audio links already, whether a person needs to pay for them or not. Unfortunately, we tested this and it does not work anymore for paid posts.

Issue #22338, opened 2019-09-08
--embed-thumbnail for Opus files


pylorak:
## Checklist

- [x] I'm reporting a feature request
- [x] I've verified that I'm running youtube-dl version **2019.09.01**
- [x] I've searched the bugtracker for similar feature requests including closed ones

## Description

Please add support for embedding cover art in opus files.

nullvideo:
You could externally, use `opusenc` to embed thumbnails into a OPUS files.
I’d try to implement that in YouTubeDL.

pylorak:
For the same reason that youtube-dl already supports embedding cover art into MP4, supporting it for Opus would be a lot more convenient for the user than having to post-process the file with a separate application like opusenc. Also, it would avoid having to re-encode the file just for the sake of adding a cover art.

nullvideo:
I was quite wrong, `opusenc` cannot post-process though, it requires a WAVE/PCM as input to encode to OPUS. This is results in lossy-to-lossy compression (quantization on already quantized signal) which degrades the quality even more.

> OPUS doesn't define it's own bit-stream format, it's usually contained inside a **OGG** container, `opusenc` follows this way.

---

MP4 (*MPEG-4 Part-II*) supports OPUS as a audio-codec. M4A is MP4 (Apple used this name to distinguish between Video and Audio files, besides that the structure is exactly the same).

Boils down to, `AtomicParsley` could be used to merge MP4 covers.

milahu:
.... or keep the ogg container, and embed the cover art image as base64 comment like

```
METADATA_BLOCK_PICTURE=base64_string
```

https://superuser.com/questions/169151/embed-album-art-in-ogg-through-command-line-in-linux

tryamid:
> 
> 
> .... or keep the ogg container
> and embed the cover art image as base64 comment like
> 
> ```
> METADATA_BLOCK_PICTURE=base64_string
> ```
> 
> https://superuser.com/questions/169151/embed-album-art-in-ogg-through-command-line-in-linux

Appending a *cover art* this way has its limitations (eg. argument list can become too long that the current shell might now allow it). Also, this isn’t automation in anyway, suppose you downloaded a bunch of files, you have to iterate through all the files and download it’s corresponding thumbnail from the site where it was downloaded so at the end *things become very messy*.

I recently focused a bit on this problem and found out that [`mutagen`][1] as optional PyPI dependency to embed thumbnails into OGGOpus files would be great.

[...]

PR #28894, opened 2021-04-27
Thumbnail embedding for OGG, OPUS and FLAC

tripulse:

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

Thumbnail embedding is as of now limited to **mp3**, **mp4**/**m4a** which is inconvenient for some people (eg. #22338, me). This is issue is insolvable by FFmpeg as [it is inable to write `METADATA_BLOCK_PICTURE` VorbisComment field](https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/vorbiscomment.c#L33-L39).

This PR adds support for embedding thumbnails into OGGVorbis, OGGOpus, FLAC files using [mutagen](https://github.com/quodlibet/mutagen) as optional dependency  which is easily installable from `pip`.

This pull request was again filed, due to my fork being taken down due to DMCA.

### DIFF
--- a/youtube_dl/postprocessor/embedthumbnail.py
+++ b/youtube_dl/postprocessor/embedthumbnail.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 
 import os
 import subprocess
+import imghdr
 
 from .ffmpeg import FFmpegPostProcessor
 
@@ -18,7 +19,6 @@ from ..utils import (
 )
 
 from base64 import b64encode
-from shutil import copyfile
 
 
 class EmbedThumbnailPPError(PostProcessingError):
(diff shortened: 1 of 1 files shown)


L33Tech (approved):

L33Tech:
+1, please add this change!

Luk164:
Any updates on this?

dirkf:
I think it would be better to implement [this work-around](https://superuser.com/questions/1708793/add-art-cover-in-ogg-audio-file/1816195#1816195) using _ffmpeg_, since the patch that would do that in _ffmpeg_ itself has still not been merged.

Luk164:
> I think it would be better to implement [this work-around](https://superuser.com/questions/1708793/add-art-cover-in-ogg-audio-file/1816195#1816195) using _ffmpeg_, since the patch that would do that in _ffmpeg_ itself has still not been merged.

Unfortunately I was converting mp3 files that already have thumbnails so I wanted to avoid having to extract the images and re-add them after conversion. Maybe we could replicate how the YTDLnis does it? Since it can embed thumbnails into its opec files by default.

dirkf:
Apparently YTDLnis wraps _yt-dlp_. You could just use that anyway.

Otherwise some script to
> ... extract the images and re-add them after conversion ...

would be a solution.

For yt-dl, the main use case is extracting audio from a video and adding the metadata (including thumbnails) from the video to the extracted audio. If _ffmpeg_ can't manage some conversion we're probably not going to do it either.

PR #29160, opened 2021-05-28
[TubeTuGraz] Add new extractor (rewrite of #26778)

Ferdi265:
  - #26778 is similar, but with less features. I have talked to @ruru4143, and he has suggested creating a new PR for this and closing the old one

  -  I have written the bulk of the extractor code from scratch
- [X] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)
  - Some of the code is inspired by @ruru4143's PR, who released it under the Unlicense according to that post. 

- [ ] Bug fix
- [ ] Improvement
- [X] New extractor (improvement to an existing proposed extractor)
- [ ] New feature

---

Add support for extracting videos from tube.tugraz.at.

Features:

- extract videos
- extract playlists
- extract restricted videos and playlists if login details are provided
- attempt fallback formats if metadata cannot be downloaded
- formats are assigned human-readable `format_id`s
- "presentation" formats are preferred over "presenter" formats

### DIFF
--- a/youtube_dl/extractor/tubetugraz.py
+++ b/youtube_dl/extractor/tubetugraz.py
@@ -108,8 +108,9 @@ class TubeTuGrazIE(InfoExtractor):
                 "count": 1,
                 "sort": "TITLE"
             })
-        series_info = try_get_or(series_info_data, [
-            lambda x: x["catalogs"][0]["http://purl.org/dc/terms/"]]) or {}
+        series_info = try_get(
+            series_info_data,
+            lambda x: x["catalogs"][0]["http://purl.org/dc/terms/"]) or {}
 
         if len(series_info) == 0:
             self.report_warning(
--- a/youtube_dl/extractor/tubetugraz.py
+++ b/youtube_dl/extractor/tubetugraz.py
@@ -2,9 +2,15 @@
 from __future__ import unicode_literals
 
 from .common import InfoExtractor
-from ..utils import ExtractorError, urlencode_postdata
+from ..utils import ExtractorError, urlencode_postdata, try_get
 
 import re
+
+def try_get_or(src, getters, expected_type=None):
+    for getter in getters:
+        v = try_get(src, getter, expected_type)
+        if v is not None:
+            return v
 
 class TubeTuGrazIE(InfoExtractor):
     IE_DESC = 'tube.tugraz.at'
@@ -27,15 +33,14 @@ class TubeTuGrazIE(InfoExtractor):
 
     def _login(self):
         username, password = self._get_login_info()
-
         if username is None:
             return
 
         result = self._download_webpage_handle(
             self._LOGIN_URL, None,
-            note = 'Downloading login page',
-            errnote = 'unable to fetch login page', fatal = False
-        )
+            note='downloading login page',
+            errnote='unable to fetch login page',
+            fatal=False)
         if result is False:
             return
         else:
@@ -43,18 +48,16 @@ class TubeTuGrazIE(InfoExtractor):
 
         result = self._download_webpage_handle(
             login_page_handle.url, None,
-            note = 'Logging in',
-            errnote = 'unable to log in', fatal = False,
-            data = urlencode_postdata({
+            note='logging in',
+            errnote='unable to log in',
+            fatal=False,
+            data=urlencode_postdata({
                 b"lang": "de",
                 b"_eventId_proceed": "",
                 b"j_username": username,
                 b"j_password": password
             }),
-            headers = {
-                "referer": login_page_handle.url
-            }
-        )
+            headers={ "referer": login_page_handle.url })
         if result is False:
             return
         else:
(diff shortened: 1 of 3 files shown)


Ferdi265:
A few things I still need to improve: for example, formats don't have format identifiers yet, which is needed to see which formats come from which video source

Ferdi265:
with the latest commit it also has readable `format_id`s

youtube_dl/extractor/tubetugraz.py
```diff
@@ -348,4 +349,5 @@ def _gen_format_id(self, format, type, transport, format_types):
         else:
             pre_index = ""
 
-        format["format_id"] = "%s-%s-%s%d" % (type, transport, pre_index, index)
+        format_id = "%s-%s-%s%d" % (type, transport, pre_index, index)
+        format["format_id"] = format_id
```
rautamiekka, line 349: What does this change do ?

Ferdi265: This change makes the line fit into the <80 character limit (it was exactly 80 characters long).

Ferdi265:
Since this PR has quite a lot of commits, compared to most of the other extractor commits, I will probably squash it down once this is reviewed.
For the review the history might be interesting, so I'll keep it for now.

dstftw:
Read coding conventions.

Ferdi265:
> Read coding conventions.

I definitely did that, and did my best to follow those conventions.
If you have more specific cases of style choices in my PR that are not according to project standards, then please mention these or mark them with a review comment so I can correct them. Otherwise I can only guess at what you mean.

Ferdi265:
Some things that you _might_ mean by that:

<details>
<summary>Detailed Summary of how I applied the Style Conventions</summary>
Mandatory fields:
- all mandatory fields are extracted

Provide fallbacks:
- multiple fallbacks for many fields are supported, and are extracted safely

Regular expressions:
- All fields of the regular expression are used. The style is inspired by the YouTube extractor.

Long lines policy:
- all lines except literal URLs are less than 80 characters long

Inline values:
- `_LOGIN_URL` and `_SUCCESS_URL` are only used once.
  I initially factored these out to make porting to a different site that uses a similar system easier, but since TuGraz uses a quite custom setup, I guess it might make sense to inline these
- same with `_API_SERIES`, but since `_API_EPISODE` is used twice, I didn't want to split them

Collapse fallbacks:
- I made `try_get_or` expressly for this purpose

Trailing parentheses:
- used everywhere

Use convenience conversion and parsing functions:
- I used `urlencode_postdata` and `try_get`, I have not seen a need for other parsing
</details>

So the only thing I can think of that might not be completely up to style standards is the 4 factored out URL constants.
Is that what you mean?

Ferdi265:
During testing, I have found a few corner cases that might need a bit more detailed design:

Some lectures have two different video streams. In this case, the presentation stream is preferred by the current implementation, since it contains most of the relevant content.

However, there are some lectures which switch between using the blackboard and the presentation, so ideally having both streams (or even merging them into one file) would be perfect. Currently the stream can be selected via format selection, but since the current format_id scheme is dependent on the bitrate, something like `-f presentation,presenter` cannot be done without checking the format list for the exact format_id.

To complicate matters further, in some cases only the presentation stream has audio. This is detected correctly by the DASH extractor, but the HLS extractor incorrectly assumes these have audio and video. I have found no way to detect whether these HLS streams contain audio without downloading them. One way to deal with this is to mark the presenter HLS streams as video-only if the presenter DASH streams are video-only.

This combined makes format selection cumbersome for the user. The default `bestvideo+bestaudio` and `best` formats work and download the presentation stream with audio (or the presenter stream if no presentation stream exists), but there's no easy way to download the presenter stream with audio instead. A workaround would be to add unlisted `presenter-best`, `presenter-bestvideo`, and `presenter-bestaudio` formats (and same with presentation) to be able to easily and reliably be able to download both streams.

A combined stream could also be merged into a single container file with two video streams and one audio stream, but that would require changes to the format spec code in youtube-dl (my syntax suggestion for combining video streams would be `video1*video2+audio`). Rendering a combined picture-in-picture format (single video stream, presenter video in the corner) with ffmpeg also works, but is infeasible for most users (render-times are at times significantly longer than download times).

The `_TEST` testcase in the PR contains an example for multiple video streams (where the presenter stream is not particularly important), but without HLS/DASH formats. I also have a testcase for multiple video streams with audio only in the presentation stream, but that one requires authentication for the metadata (HLS and DASH streams are still reachable without authentication and are correctly guessed by the extractor, but those being accessible might be a bug in tube.tugraz.at).

youtube_dl/extractor/tubetugraz.py
nao20010128nao: <80 character limit is a soft limit; you are allowed to over 80 characters limit since it doesn't make it hard to read

Ferdi265:
Closed in favor of [yt-dlp#2397](https://github.com/yt-dlp/yt-dlp/pull/2397)

PR #29175, opened 2021-05-30
Setting a cookie helps to avoid http 429 errors on the Youtube "watch?v=" page.

Stefan311:
Seems youtube now blocks IP's very fast when requesting there "watch?v="-page. This seems NOT the case if you add a cookie to this request.
This is not a production-ready patch. I know setting a static text for the consent-date and for the global zone is far from ideal. But it works for now.


Stefan311:
DAMN. Sent the wrong patch

PR #29281, opened 2021-06-13
[youtube] update invidious instance list

B0pol:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Updated the invidious instance list using https://github.com/iv-org/documentation/blob/master/Invidious-Instances.md


pukkandan:
Something like https://github.com/yt-dlp/yt-dlp/commit/df0c81513e0bb37986d00c532a5ad8cef31a24ea would be more beneficial than updating the list periodically

Ref: https://github.com/iv-org/invidious/pull/1730

youtube_dl/extractor/youtube.py
```diff
@@ -353,7 +353,7 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
         r'(?:www\.)?invidious\.13ad\.de',
         r'(?:www\.)?invidious\.mastodon\.host',
         r'(?:www\.)?invidious\.zapashcanon\.fr',
-        r'(?:www\.)?invidious\.kavin\.rocks',
+        r'(?:www\.)?(invidious(-us)?|piped)\.kavin\.rocks',
```
dstftw, line 356: Do not capture groups you don't use.

dstftw (changes requested):

youtube_dl/extractor/youtube.py
B0pol: ?

pukkandan: He means you are supposed to make it a non-capturing group. Like this:
`(?:invidious(?:-us)?|piped)`

PR #29383, opened 2021-06-24
[peertube] only call description endpoint if necessary

B0pol:

---

- [x] Covered the code with tests (note that PRs without tests will be REJECTED) ("what is peertube" video description is already in tests and is above 250 characters so an extra api call is needed).

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Prevent an useless call to API by checking if the description is shortened and only requesting the full description to the API if necessary.

### DIFF
--- a/youtube_dl/extractor/peertube.py
+++ b/youtube_dl/extractor/peertube.py
@@ -570,7 +570,7 @@ class PeerTubeIE(InfoExtractor):
         self._sort_formats(formats)
 
         description = video.get('description')
-        if len(description) == 250 and description.endswith('...'):
+        if len(description) >= 250:
             # description is shortened
             full_description = self._call_api(
                 host, video_id, 'description', note='Downloading description JSON',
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/peertube.py
```diff
@@ -569,15 +569,15 @@ def _real_extract(self, url):
             formats.append(f)
         self._sort_formats(formats)
 
-        full_description = self._call_api(
-            host, video_id, 'description', note='Downloading description JSON',
-            fatal=False)
+        description = video.get('description')
+        if len(description) == 250 and description.endswith('...'):
```
remitamine, line 573: it might be better to use the same condition that is used by [PeerTube's client code](https://github.com/Chocobozzz/PeerTube/blob/d55c466e1caff9082e8ca1734fbf079ac92e12b4/client/src/app/%2Bvideos/%2Bvideo-watch/video-watch.component.html#L216)

B0pol: The first check is just "is the button clicked", the second is >= 250.

From my test max is always 250, not above and it ends with `...` but I changed to >= 250 as you wanted in the last commit.

PR #29573, opened 2021-07-17
feat(arte_extractor): add alt_title for regular shows                …

casaper:

I changed somone elses extractor. However, I presume its ok. The checkboxes confused me.

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Currently a regular show like ARTE Reportage gets the same title
for all ARTE Reportage shows.
This change adds the more meaningful title additionally if available:

https://www.arte.tv/de/videos/030273-820-A/arte-reportage/

Now: title is `ARTE Reportage` and no alt_title is available
With this: title is `ARTE Reportage - Sudan: Die Tigray fliehen aus Äthiopien` and
           alt_title is `Sudan: Die Tigray fliehen aus Äthiopien`

### DIFF
--- a/youtube_dl/extractor/arte.py
+++ b/youtube_dl/extractor/arte.py
@@ -33,6 +33,15 @@ class ArteTVIE(ArteTVBaseIE):
                         /(?P<id>\d{6}-\d{3}-[AF])
                     ''' % {'langs': ArteTVBaseIE._ARTE_LANGUAGES}
     _TESTS = [{
+        'url': 'https://www.arte.tv/de/videos/092724-001-A/lasst-mich-schlafen/',
+        'info_dict': {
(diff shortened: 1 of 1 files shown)


casaper:
I could not use the example from the PR description for the tests, because this download will only be available until 2021-08-15.

Having said that, it is a general problem, though. The tests that were there and the one I added now are still prone to fail one day, because it seems most Arte.tv videos have an expiration date.

I'm not very fond with python, but in ruby I would use the VCR gem to solve that problem. Probably, python will have something that does the same: Record a http request in order to mock/stub a http call response behaviour for a test.
But that would be quite out of scope here... just mentioning it because I noticed it... ¯\＿(ツ)＿/¯

PR #29590, opened 2021-07-19
[YoutubeDL] Add --client-certificate option

dirkf:

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

Some sites only allow access to HTTPS connections that offer a valid client certificate. In some cases, the client certificate may be built into the firmware of a client device. While such certificates are typically not accessible without jail-breaking the device, it would be desirable to allow this means of authentication as well as username+password, etc.

This PR adds an option `--client-certificate` whose value should be the filename of a single certificate file containing the certificate and private key (as described in the Python library documentation for the `ssl` module under 'Combined key and certificate').

### DIFF
--- a/test/test_clientcert.py
+++ b/test/test_clientcert.py
@@ -39,7 +39,11 @@ class TestClientCert(unittest.TestCase):
 
     def test_check_clientcertificate(self):
         clientcertfn = os.path.join(TEST_DIR, 'testdata', 'clientcert', 'client.crt')
-        ydl = YoutubeDL({'logger': FakeLogger(), 'clientcertificate': clientcertfn})
+        ydl = YoutubeDL({'logger': FakeLogger(), 'clientcertificate': clientcertfn,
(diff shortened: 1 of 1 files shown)


dirkf:
The new test case `test_check_clientcertificate` failed for Python >= 2.7.8 because yt-dl sets the HTTPS client to validate the server certificate, and this failed with the test HTTPS server from `test/test_http.py` because self-signed certificates don't pass.

If `'nocheckcertificate': True` is passed along with the `clientcertificate` option in the `test_check_clientcertificate` test case, the test succeeds.

The tests for HTTPS in `test/test_http.py` don't check that HTTPS connections work in general, but only that  the `nocheckcertificate` option works. A similar test-case without that option fails, but this will happen unless the server certificate has a valid path to a root CA. Unfortunately, [arranging proper SSL certificates for testing seems to be a hopeless task](https://gist.github.com/andrewn/b30375074398ab94361dc607ac7c0e12).

PR #29688, opened 2021-07-29
[extractor/nbc] Fix NBC Olympics extractor

nchilada:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This might fix the extractor(s?) for NBC Olympics. It's a work in progress; I haven't finished debugging and testing it, just wanted to get this out there for anyone else who might be looking into this. I'll update this description when it's ready to go.

### DIFF
--- a/youtube_dl/extractor/nbc.py
+++ b/youtube_dl/extractor/nbc.py
@@ -495,12 +495,10 @@ class NBCOlympicsStreamIE(AdobePassIE):
         display_id = self._match_id(url)
         webpage = self._download_webpage(url, display_id)
         pid = self._search_regex(r'pid\s*=\s*(\d+);', webpage, 'pid')
-        resource = self._search_regex(
-            r"resource\s*=\s*'(.+)';", webpage,
-            'resource').replace("' + pid + '", pid)
         event_config = self._download_json(
             self._DATA_URL_TEMPLATE % ('event_config', pid),
             pid)['eventConfig']
+        resource = event_config.get('resourceId', 'NBCOlympics')
         title = self._live_title(event_config['eventTitle'])
         source_url = self._download_json(
             self._DATA_URL_TEMPLATE % ('live_sources', pid),


asedeno:
`-http_seekable` requires ffmpeg `n4.3` or newer; with older versions this errors out. Worked great once I built a newer ffmpeg than the one in Debian 10. Thanks!

asedeno:
It appears not all videos require getting a tokenized url, and for those that don't we get a `400 BAD REQUEST` while generating it. Using the non-tokenized `source_url` in place of the tokenized url seems to work for those though.

nchilada:
> `-http_seekable` requires ffmpeg `n4.3` or newer; with older versions this errors out. Worked great once I built a newer ffmpeg than the one in Debian 10. Thanks!

@asedeno good to know, thanks!

For people who are stuck with older versions of ffmpeg, one option would be to use a (local) proxy to strip request headers from the outbound requests. That's what I did initially:
Charles Proxy: fix ffmpeg requests to akamaized.net

> It appears not all videos require getting a tokenized url, and for those that don't we get a `400 BAD REQUEST` while generating it. Using the non-tokenized `source_url` in place of the tokenized url seems to work for those though.
>
> `event_config['cdnToken']` (boolean) is the indicator you want.

Hmm, I do see `event_config['cdnToken']: True` in the videos that I've been testing with. I think you're saying, if that property is `False`, then we must avoid [the code that computes `tokenized_url`](https://github.com/ytdl-org/youtube-dl/pull/29688/files/b675a6e6b9c4e2a54058746d1084dd67c8096dee#diff-47ee1d185ae7b1038bb5468a212ceabdd8e22c93ac45b55cbd0a29cb7ecdb1cfR524-R544) and simply use `source_url`?

It might be helpful if you could link to such a stream, but failing that, I'm happy to make the change and keep it if you say it works for you. I just don't want to submit a branch of code that hasn't been run by _someone_.

In any case, thank you very much for taking a look at this and providing important feedback! It's very late here but I will check back tomorrow.

asedeno:
> > It appears not all videos require getting a tokenized url, and for those that don't we get a `400 BAD REQUEST` while generating it. Using the non-tokenized `source_url` in place of the tokenized url seems to work for those though.
> > `event_config['cdnToken']` (boolean) is the indicator you want.
> 
> Hmm, I do see `event_config['cdnToken']: True` in the videos that I've been testing with. I think you're saying, if that property is `False`, then we must avoid [the code that computes `tokenized_url`](https://github.com/ytdl-org/youtube-dl/pull/29688/files/b675a6e6b9c4e2a54058746d1084dd67c8096dee#diff-47ee1d185ae7b1038bb5468a212ceabdd8e22c93ac45b55cbd0a29cb7ecdb1cfR524-R544) and simply use `source_url`?
> 
> It might be helpful if you could link to such a stream, but failing that, I'm happy to make the change and keep it if you say it works for you. I just don't want to submit a branch of code that hasn't been run by _someone_.
> 
> In any case, thank you very much for taking a look at this and providing important feedback! It's very late here but I will check back tomorrow.

Yeah, that's what I'm saying. Here's one that appears that way for me.

https://stream.nbcolympics.com/gymnastics-event-finals-mens-floor-pommel-horse-womens-vault-bars

nchilada:
@asedeno thanks for the super quick link! You were exactly right. Update incoming, with a new test case

wesnm:
I've been messing around with this for a couple hours, and I had to make these small changes to make it work... Note, I have NOT tried all manner of iterations, just made the requests look as close to what my browser makes as possible.

The data dictionary for _tokenized_url_ that works for me:
```
json.dumps({
                'application': 'NBCSports',
                'authentication-type': 'adobe-pass',
                'cdn': 'akamai',
                'inPath': 'false',
                'pid': pid,
                'platform': 'desktop',
                'requestorId': event_config.get('requestorId', 'NBCOlympics'),
                'resourceId': base64.b64encode(ap_resource.encode()).decode(),
                'token': base64.b64encode(media_token.encode()).decode(),
                'url': source_url,
                'version': 'v1',
}).encode()
```

Note that some of the keys are different, and I added _requestorId_, _platform_, _pid_, and _inPath_. Otherwise, the result is an error code. The _resourceId_ also needed to be _ap_resource_ rather than _resource_.

This got me to the point of ffmpeg starting, but it always died with a 403 error. After adding these extra headers, things started working:
```
--add-header Origin:https://stream.nbcolympics.com
--add-header Referer:https://stream.nbcolympics.com/
--add-header Sec-Fetch-Dest:empty
--add-header Sec-Fetch-Mode:cors
--add-header Sec-Fetch-Site:cross-site
```

Again, I didn't test different combinations to see if they were all required. Just added what was missing from the browser. After this, ffmpeg downloads the stream just fine.

wesnm:
This got me to the point of ffmpeg starting, but it always died with a 403 error. After adding these extra headers, things started working:
```
--add-header Origin:https://stream.nbcolympics.com
--add-header Referer:https://stream.nbcolympics.com/
--add-header Sec-Fetch-Dest:empty
--add-header Sec-Fetch-Mode:cors
--add-header Sec-Fetch-Site:cross-site
```

I didn't test different combinations to see if they were all required. Just added what was missing from the browser. After this, ffmpeg downloads the stream just fine.

Edit: Now it works fine this morning without the extra headers, strange.

Suggest using the `event_config["eventStatus"]` to determine if the event is live or not. The value is "live" for a live broadcast, otherwise set to "replay".

nchilada:
> This got me to the point of ffmpeg starting, but it always died with a 403 error. After adding these extra headers, things started working:
> 
> ```
> --add-header Origin:https://stream.nbcolympics.com
> --add-header Referer:https://stream.nbcolympics.com/
> --add-header Sec-Fetch-Dest:empty
> --add-header Sec-Fetch-Mode:cors
> --add-header Sec-Fetch-Site:cross-site
> ```
> 
> I didn't test different combinations to see if they were all required. Just added what was missing from the browser. After this, ffmpeg downloads the stream just fine.
> 
> Edit: Now it works fine this morning without the extra headers, strange.

@wesnm interesting, thanks for noting this. I will take a look later today. It’s conceivable that a server would restrict its activity based on `Origin` or partial `Referer`, though I haven't seen such behavior in the last couple days.

> Suggest using the `event_config["eventStatus"]` to determine if the event is live or not. The value is "live" for a live broadcast, otherwise set to "replay".

It makes sense that a live stream might behave differently than a reply! I’ll check whether the code works on a live stream, though I won’t be able to add a long-lasting test case for that.

nchilada:
> > This got me to the point of ffmpeg starting, but it always died with a 403 error. After adding these extra headers, things started working:
> > ```
> > --add-header Origin:https://stream.nbcolympics.com
> > --add-header Referer:https://stream.nbcolympics.com/
> > --add-header Sec-Fetch-Dest:empty
> > --add-header Sec-Fetch-Mode:cors
> > --add-header Sec-Fetch-Site:cross-site
> > ```
> > I didn't test different combinations to see if they were all required. Just added what was missing from the browser. After this, ffmpeg downloads the stream just fine.
> >
> > Edit: Now it works fine this morning without the extra headers, strange.
> 
> @wesnm interesting, thanks for noting this. I will take a look later today. It’s conceivable that a server would restrict its activity based on `Origin` or partial `Referer`, though I haven't seen such behavior in the last couple days.

I thought about this some more and I don't think it's worth adding prophylactic `Referer` and `Origin` headers, nor the preflight OPTIONS requests that are also part of CORS. When servers require `Referer` or `Origin`, I think it's usually a kind of CSRF protection, and I don't know if NBC and their technology partners would bother to implement _any_ kind of CSRF in the context of video streaming. I'm willing to wait and see.

> > Suggest using the `event_config["eventStatus"]` to determine if the event is live or not. The value is "live" for a live broadcast, otherwise set to "replay".
> 
> It makes sense that a live stream might behave differently than a reply! I’ll check whether the code works on a live stream, though I won’t be able to add a long-lasting test case for that.

Darn, I'm not able to load any live streams with my credentials, either via a web browser or via youtube-dl. I can't figure out if such pages even allow clients to go back to the very beginning of the stream. Can we skip this for now? I assume replays are the priority.

nchilada:
@remitamine @dstftw @others could someone please approve the workflows for this PR when you get a chance? I think this PR is ready to go!

wesnm:
> I thought about this some more and I don't think it's worth adding prophylactic `Referer` and `Origin` headers, nor the preflight OPTIONS requests that are also part of CORS. When servers require `Referer` or `Origin`, I think it's usually a kind of CSRF protection, and I don't know if NBC and their technology partners would bother to implement _any_ kind of CSRF in the context of video streaming. I'm willing to wait and see.

If it's not needed to work, I wouldn't include it. I don't know what was going on last night, maybe just needed sleep.

> 
> > > Suggest using the `event_config["eventStatus"]` to determine if the event is live or not. The value is "live" for a live broadcast, otherwise set to "replay".
> > 
> > 
> > It makes sense that a live stream might behave differently than a reply! I’ll check whether the code works on a live stream, though I won’t be able to add a long-lasting test case for that.
> 
> Darn, I'm not able to load any live streams with my credentials, either via a web browser or via youtube-dl. I can't figure out if such pages even allow clients to go back to the very beginning of the stream. Can we skip this for now? I assume replays are the priority.

One thing the "live" status does is determine if youtube-dl will delegate to alternate downloaders. If an event is flagged as live, it will _only_ download using ffmpeg with a single worker, which is much much slower. 

These live events in the browser do allow rewinding to any point in the stream.

Issue #10785, opened 2016-09-28
Add support for rumble.com


jegue:
## Please follow the guide below
- You will be asked some questions and requested to provide some information, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your _issue_ (like that [x])
- Use _Preview_ tab to see how your issue will actually look like

---
### Make sure you are using the _latest_ version: run `youtube-dl --version` and ensure your version is _2016.09.27_. If it's not read [this FAQ entry](https://github.com/rg3/youtube-dl/blob/master/README.md#how-do-i-update-youtube-dl) and update. Issues with outdated version will be rejected.
- [ ] I've **verified** and **I assure** that I'm running youtube-dl **2016.09.27**
### Before submitting an _issue_ make sure you have:
- [ ] At least skimmed through [README](https://github.com/rg3/youtube-dl/blob/master/README.md) and **most notably** [FAQ](https://github.com/rg3/youtube-dl#faq) and [BUGS](https://github.com/rg3/youtube-dl#bugs) sections
- [ ] [Searched](https://github.com/rg3/youtube-dl/search?type=Issues) the bugtracker for similar issues including closed ones
### What is the purpose of your _issue_?
- [ ] Bug report (encountered problems with youtube-dl)
- [ ] Site support request (request for adding support for a new site)
- [ ] Feature request (request for a new functionality)
- [ ] Question
- [ ] Other

---
### The following sections concretize particular purposed issues, you can erase any section (the contents between triple ---) not applicable to your _issue_

---
### If the purpose of this _issue_ is a _bug report_, _site support request_ or you are not completely sure provide the full verbose output as follows:

Add `-v` flag to **your command line** you run youtube-dl with, copy the **whole** output and insert it here. It should look similar to one below (replace it with **your** log inserted between triple ```):

```
$ youtube-dl -v <your command line>
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-v', u'http://www.youtube.com/watch?v=BaW_jenozKcj']
[debug] Encodings: locale cp1251, fs mbcs, out cp866, pref cp1251
[debug] youtube-dl version 2016.09.27
[debug] Python version 2.7.11 - Windows-2003Server-5.2.3790-SP2
[debug] exe versions: ffmpeg N-75573-g1d0487f, ffprobe N-75573-g1d0487f, rtmpdump 2.4
[debug] Proxy map: {}
...
<end of log>
```

---
[...]

PR #29758, opened 2021-08-09
Plugin rumble

Grabien:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Support of direct links in addition to embed links in the Rumble website plugin.

Link example:
https://rumble.com/vhlrar-mike-lindell-to-confront-brian-kemp-and-doug-ducey-over-election-fraud.html

### DIFF
--- a/youtube_dl/extractor/rumble.py
+++ b/youtube_dl/extractor/rumble.py
@@ -1,6 +1,8 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
+import re
+
 from .common import InfoExtractor
 from ..compat import compat_str
 from ..utils import (
@@ -9,8 +11,6 @@ from ..utils import (
     parse_iso8601,
     try_get,
 )
-
-import re
 
 
 class RumbleEmbedIE(InfoExtractor):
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/rumble.py
```diff
@@ -26,10 +28,18 @@ class RumbleEmbedIE(InfoExtractor):
...
         'url': 'https://rumble.com/embed/ufe9n.v5pv5f',
         'only_matching': True,
+    }, {
+        'url': 'https://rumble.com/vhlrar-mike-lindell-to-confront-brian-kemp-and-doug-ducey-over-election-fraud.html',
+        'only_matching': True,
     }]
 
     def _real_extract(self, url):
-        video_id = self._match_id(url)
+        if re.match(r'https?://(?:www\.)?rumble\.com/(.*?)\.html', url):
```
dirkf, line 37: This sequence would set a `video_id` for the debug log of `_download_webpage`:
```
if re.match(...):
    content, urlh = self._download_webpage_handle(url, video_id)
    video_id = ...
```

youtube_dl/extractor/rumble.py
```diff
@@ -26,10 +28,18 @@ class RumbleEmbedIE(InfoExtractor):
...
+            video_id = ""
+            content, urlh = self._download_webpage_handle(url, video_id)
+            video_id = re.findall(r'"embedUrl":"https://rumble\.com/embed/(.*?)/"', content)[0]
```
dirkf, line 40: Might the site start sending white space around the `:`, which would be matched by `\s*:\s*` in the regex?

Grabien:
I added extraction of direct video ID from the URL. Then it will be used to download the page to extract the embedded video ID. Direct ID and embed ID are different for each video, I hope it's hot a problem.

SimonEast:
@dirkf I would like to see this pull request get across the line, and I assume the the dozens of commenters in issue #10785 would also. Ideally we'd be able to download Rumble links using the main URL and not have to extract the embed URL manually.

I think your comments and suggestions on the pull request all seem reasonable. Can your suggested tweaks be made to get this closer to acceptance? I'd have a go myself, but I know only very little Python.

If there's aspects that are still broken/missing, let us know so we can understand what's needed to get it finished off.

PR #29807, opened 2021-08-17
[postprocessort/ffmpeg] Allow embedding subtitles in m4a

kikuyan:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Subtitles can be embedded in m4a same as mp4.

I imagine the reason m4a is not allowed now is because no one ever needed it or it is not video. However, given that webm may only contain audio, I think m4a should be allowed as well.


rautamiekka:
You can't embed subtitles into an audio codec, you need to create a separate vid to show the subtitles, at which point you no longer have an audio file but a vid file without anything but the text to see.

kikuyan:
mpv shows subtitles for audio only files on console, or if thumbnail is embedded (or with some options) on gui.

nao20010128nao:
@rautamiekka [Just like this one](https://en.wikipedia.org/wiki/File:Auld_Lang_Syne_-_U.S._Navy_Band.ogg), audio files can have subtitles, and ffmpeg supports this. Formats for subtitles are very limited for m4a (which is same as mp4 except that m4a lacks video), but this should work.

Issue #29539, opened 2021-07-14
Could not send HEAD request to https://playlist.megaphone.fm/ & Unable to download webpage: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] 


MagTun:
## Checklist

- [x ] I'm reporting a broken site support
- [ x] I've verified that I'm running youtube-dl version **2021.06.06**
- [ x] I've checked that all provided URLs are alive and playable in a browser
- [ x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x ] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
youtube-dl -ciw -f bestaudio --extract-audio --audio-format mp3 --audio-quality 0  "https://playlist.megaphone.fm/?e=FULLCAST5192859578" -v
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-ciw', '-f', 'bestaudio', '--extract-audio', '--audio-format', 'mp3', '--audio-quality', '0', 'https://playlist.megaphone.fm/?e=FULLCAST5192859578', '-v']
[debug] Encodings: locale cp1252, fs mbcs, out cp437, pref cp1252
[debug] youtube-dl version 2021.06.06
[debug] Python version 3.4.4 (CPython) - Windows-10-10.0.19041
[debug] exe versions: ffmpeg 4.2.2, ffprobe 4.2.2, phantomjs 2.1.1
[debug] Proxy map: {}
[generic] ?e=FULLCAST5192859578: Requesting header
WARNING: Could not send HEAD request to https://playlist.megaphone.fm/?e=FULLCAST5192859578: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:600)>
[generic] ?e=FULLCAST5192859578: Downloading webpage
ERROR: Unable to download webpage: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:600)> (caused by URLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:600)'),))
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpkqxnwl31\build\youtube_dl\extractor\common.py", line 634, in _request_webpage
[...]

PR #29826, opened 2021-08-19
[Megaphone.fm] Rewrite extractor with episode, playlist support

dirkf:

---

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

The site structure of megaphone.fm appears to have changed (issue #29539).

This PR updates the extraction to match the current structure, defaulting to the old extraction if the new one fails.

Also:
* adds extraction for episodes playlist.megaphone.fm/?e=...
* adds extraction for playlists playlist.megaphone.fm/?p=...
* adds extraction for channel playlists cms.megaphone.fm/channel/...

Playlists and channel playlists support `--no-playlist` to select the current show instead of the whole playlist.


youtube_dl/extractor/megaphone.py
```diff
@@ -45,11 +68,225 @@ def _real_extract(self, url):
...
+        playlist_json = self._parse_json(
+            playlist_json, video_id, 
+            transform_source=lambda s: self._parse_json(s, video_id, transform_source=js_to_json))
+        entries = []
+        ep_num = 0
+        ep_list = None
+        # Support --no-playlist to get the first item when no explicit selection
+        noplaylist = self._downloader.params.get('noplaylist')
+        for episode in playlist_json:
+            ep_num += 1
```
rautamiekka, line 245: Would using `for ep_num, episode in enumerate(playlist_json, 1)` and later `if len(...)` or simply `if:` make more sense since you'll be counting the episodes ?

youtube_dl/extractor/megaphone.py
```diff
@@ -45,11 +68,225 @@ def _real_extract(self, url):
...
-            'duration': episode_data['duration'],
+            'duration': episode_data.get('duration'),
             'formats': formats,
         }
 
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        episode_json = self._download_json(self._JSON_URL_TEMPL % video_id, video_id, fatal=False)
+        if episode_json is False:
```
rautamiekka, line 79: Here `if episode_json is False` can/should be replaced with `if not episode_json`, which is also the way you're doing it later.

dirkf: The test is intentional to distinguish between `False` (bad result from request) and a JSON object that converts to a falsy Python value (eg `{}`, `[]`).

rautamiekka: Oh, ok. That makes sense. Shoulda read the statement above it.

youtube_dl/extractor/megaphone.py
dirkf: Yes.

PR #29914, opened 2021-09-06
[snapchat] Add new extractor

robotman771:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This pull request adds an extractor for snapchat. Right now, only direct links to spotlight snaps such as [this one](ttps://story.snapchat.com/o/W7_EDlXWTBiXAEEniNoMPwAAYuz9_1mcdex8MAXndPyIOAXndPyFyAO1OAA) are supported, but support could be expanded to stories, like the ones that can be found [here](https://story.snapchat.com/) as well.

### DIFF
--- a/youtube_dl/extractor/snapchat.py
+++ b/youtube_dl/extractor/snapchat.py
@@ -1,0 +1,64 @@
+# coding: utf-8
+from __future__ import unicode_literals
+import re
+
+from ..utils import unified_timestamp, parse_duration
+from .common import InfoExtractor
+
+
+class SnapchatIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?story.snapchat\.com/o/(?P<id>\w+)'
+    _TEST = {
+        'url': 'https://story.snapchat.com/o/W7_EDlXWTBiXAEEniNoMPwAAYuz9_1mcdex8MAXndPyIOAXndPyFyAO1OAA',
+        'md5': 'ab1900981cadcd955aae32a526096cbd',
+        'info_dict': {
+            'id': 'W7_EDlXWTBiXAEEniNoMPwAAYuz9_1mcdex8MAXndPyIOAXndPyFyAO1OAA',
+            'ext': 'mp4',
+            'title': 'W7_EDlXWTBiXAEEniNoMPwAAYuz9_1mcdex8MAXndPyIOAXndPyFyAO1OAA',
+            'thumbnail': 'https://s.sc-cdn.net/p9mgs8YyMZzfFRls4HGZxwXWBGVF-cEa6ZYPuCT2xPk=/default/preview.jpg',
+            'description': '#spotlight',
+            'timestamp': 1622914559,
+            'upload_date': '20210605',
+            'view_count': 62100,
+            'uploader_id': None
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        schema_video_object_raw = self._html_search_regex(r'<script data-react-helmet="true" type="application/ld\+json">(.+)</script>',
+                                                          webpage, 'schema_video_object')
+        schema_video_object = self._parse_json(schema_video_object_raw, video_id, fatal=True)
+        video_url = schema_video_object.get('contentUrl')
+        title = schema_video_object.get('name')
+        if not title:
+            matching = re.match(r'sc-cdn.net/\w+/(\w+)', video_url)
+            if matching:
+                title = matching.group(1)
+        if not title:
+            title = video_id
+
+        try:
+            views = schema_video_object.get('interactionStatistic').get('userInteractionCount')
+        except AttributeError:
+            views = None
+
+        try:
+            uploader_id = schema_video_object.get('creator').get('alternateName')
+        except AttributeError:
+            uploader_id = None
+
+        return {
+            'id': video_id,
+            'title': title,
+            'url': video_url,
+            'ext': 'mp4',
+            'thumbnail': schema_video_object.get('thumbnailUrl'),
+            'timestamp': unified_timestamp(schema_video_object.get('uploadDate')),
+            'description': schema_video_object.get('description'),
+            'duration': parse_duration(schema_video_object.get('duration')),
+            'view_count': views,
+            'uploader_id': uploader_id
+        }
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -1109,6 +1109,7 @@ from .sky import (
 from .slideshare import SlideshareIE
 from .slideslive import SlidesLiveIE
 from .slutload import SlutloadIE
+from .snapchat import SnapchatIE
 from .snotr import SnotrIE
 from .sohu import SohuIE
 from .sonyliv import SonyLIVIE


youtube_dl/extractor/snapchat.py
```diff
@@ -0,0 +1,64 @@
...
+            'view_count': 62100,
+            'uploader_id': None
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        schema_video_object_raw = self._html_search_regex(r'<script data-react-helmet="true" type="application/ld\+json">(.+)</script>',
```
dirkf, line 31: * Consider non-greedy `...(.+?)...` to avoid matching multiple `<script..>...</script>` within the capture group.
* Isn't there a terminator like `;` that should be mentioned, so eg `...(.+?;)\s<script>'`?
* Instead of the spaces, `\s[^>]*?` is robust against changes of whitespace and unexpected new attributes.
* You could use the formulation `(?P<q1>"|\'|\b)attribute_name(?P=q1) to be robust against changes in or omission of quoting, but then you'd also have to use `...(?P<schema_obj>.+?)...` and add `group='schema_obj'` to the arguments.
* There could also be whitespace in the attribute expressions, eg `...data-react-helmet\s*=\s*"true"...`.
* Ideally the pattern would be robust against changes in attribute order, but there's no way to specify a match of two patterns A, B regardless of order, except to write `(?P<A>A)B(?(A)|A)`.

youtube_dl/extractor/snapchat.py
```diff
@@ -0,0 +1,64 @@
...
+class SnapchatIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?story.snapchat\.com/o/(?P<id>\w+)'
+    _TEST = {
+        'url': 'https://story.snapchat.com/o/W7_EDlXWTBiXAEEniNoMPwAAYuz9_1mcdex8MAXndPyIOAXndPyFyAO1OAA',
+        'md5': 'ab1900981cadcd955aae32a526096cbd',
+        'info_dict': {
+            'id': 'W7_EDlXWTBiXAEEniNoMPwAAYuz9_1mcdex8MAXndPyIOAXndPyFyAO1OAA',
+            'ext': 'mp4',
+            'title': 'W7_EDlXWTBiXAEEniNoMPwAAYuz9_1mcdex8MAXndPyIOAXndPyFyAO1OAA',
+            'thumbnail': 'https://s.sc-cdn.net/p9mgs8YyMZzfFRls4HGZxwXWBGVF-cEa6ZYPuCT2xPk=/default/preview.jpg',
```
dirkf, line 18: Could be better to use a pattern here in case Snapchat moves this image, eg:
```
            'thumbnail': 're:https://s\.sc-cdn\.net/.+\.jpg'
```

youtube_dl/extractor/snapchat.py
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+
+        schema_video_object_raw = self._html_search_regex(r'<script data-react-helmet="true" type="application/ld\+json">(.+)</script>',
+                                                          webpage, 'schema_video_object')
+        schema_video_object = self._parse_json(schema_video_object_raw, video_id, fatal=True)
+        video_url = schema_video_object.get('contentUrl')
```
dirkf, line 34: Strictly, the JSON may not have resulted in a `schema_video_object` that is a `dict`, even if that case is an error, and even if it is the required `video_url` might not be found. You could consider, eg
```
        try:
            video_url = str_or_none(schema_video_object['contentUrl'])
            if not video_url:
                raise ValueError('video_url must be non-empty string')
        except (TypeError, ValueError) as e:
            raise ExtractorError('Unexpected format for schema_video_object', cause=e, video_id=video_id)
```

youtube_dl/extractor/snapchat.py
```diff
@@ -0,0 +1,64 @@
...
+        title = schema_video_object.get('name')
+        if not title:
+            matching = re.match(r'sc-cdn.net/\w+/(\w+)', video_url)
+            if matching:
+                title = matching.group(1)
+        if not title:
+            title = video_id
```
dirkf, line 41: Or (probably the same result):
```
            title = self._generic_title(url)
```

youtube_dl/extractor/snapchat.py
```diff
@@ -0,0 +1,64 @@
...
+
+        try:
```
dirkf, line 43: Consider using the library routine `try_get` (`utils.py`) for ll. 43-6:
```
       views = try_get(schema_video_object.get('interactionStatistic'), lambda x: x['userInteractionCount'])

File: youtube_dl/extractor/snapchat.py [hunk 6]
```diff
@@ -0,0 +1,64 @@
...
+            views = schema_video_object.get('interactionStatistic').get('userInteractionCount')
+        except AttributeError:
+            views = None
+
+        try:
```
dirkf, line 48: Similarly for ll.48-51:
```
        uploader_id = try_get(schema_video_object.get('creator'), lambda x: x[alternateName'])

youtube_dl/extractor/snapchat.py
robotman771: - Done
- There is no terminator, just JSON
- Done
- I did not implement this as I do not understand it fully
- Done
- I did not implement this as I do not understand it fully

dirkf: >     * Done
> 
>     * There is no terminator, just JSON

Fine, then. 

>     * Done
> 
>     * I did not implement this as I do not understand it fully
The attribute values in HTML tags can be 'quoted' or "double-quoted" or (historically) not quoted at all.

To match all of these quoting styles, you could use a named capture group that matches any of the beginning quote styles `(?P<q1>"|\'|\b)`and then close the match with a back-reference `(?P=q1)`. However, a named capture group will count among the numbered groups considered by `_html_search_regex()`, so you'd also have to use a named capture group for the JSON that you're targeting `(?P<schema_obj>.+?)` instead of just `{.+)` and add `group='schema_obj'` to the method arguments to tell it to use that group. But I admit it's unlikely Snapchat will change its quoting style.
 
>     * Done
> 
>     * I did not implement this as I do not understand it fully

The HTML tag before the target JSON `<script data-react-helmet="true" type="application/ld\+json">` would be equally valid as `<script type="application/ld\+json" data-react-helmet="true">`. However regular expression syntax is not good at expressing this. The Perl-style REs supported in Python have an 'alternative' syntax `(?(A)match-if-A matched|match-if-A-didn't-match)`, where `A` is a capture group name or number. Thus `(?P<A>A-pattern)B-pattern(?(A)|A-pattern)` matches both `A-patternB-pattern` and `B-patternA-pattern`.

So your match (without any of the other suggested changes) could be `r'<script (?P<helmet>data-react-helmet="true" )type="application/ld\+json"(?(helmet)| data-react-helmet="true")>(.+)</script>'`

Introducing a named capture group means tweaking the method call as above, and again it's probably unlikely that Snapchat will change the order of the attributes, unless they upgrade their React toolset to something that accidentally generates a different order.

Thanks for listening!

PR #29916, opened 2021-09-06
pip to pip3

acetoneRu:

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Windows users can [download an .exe file](https://yt-dl.org/latest/youtube-dl.ex
 
 You can also use pip:
 
-    sudo -H pip install --upgrade youtube-dl
+    sudo -H pip3 install --upgrade youtube-dl
     
 This command will update youtube-dl if you have already installed it. See the [pypi page](https://pypi.python.org/pypi/youtube_dl) for more information.
 


dirkf:
1. You've ignored the template.
2. Why would this be a good example for a user who doesn't have Python3?

zulc22:
yeah, youtube-dl has support for python 2.

mara004:
+1 as there is no point at recommending the use of ancient Python 2 stuff.
I think on the long term it would make sense to drop support for Python 2 at some point so youtube-dl can make use of modern Python 3 features and libraries, such as f-strings, type hints, yield from, pathlib, and so on.

rautamiekka:
Note (source: official Python docs):
- F-strings require >=3.6.0 (but 3.8.0 would be optimal cuz >=3.7.0 allows using `await` expressions and comprehensions together with `async for`, and >=3.8.0 supports `=` to easily create debug `print`s showing the name+contents of a var). F-strings are very lovely and fast, but their realtime processing means they can only be used for strings which need to be created right now, while all templates need to use the somewhat slower `str.format()` way.
- [type hints](https://docs.python.org/3/library/typing.html) are truly useful >=3.9.0 due to many new features added since its introduction in 3.5.0, 1 being the standard types (list, tuple, ...) support being used directly as type hints instead of `import`ing them from `typing`. Not the biggest problem if you use PyCharm or some other editor to warn you of lack of `import`ed type hints being used, though. 3.9.1 introduced some new type hints, but I doubt those are gonna be used in yt-dl[p] anyway.
- `yield from` needs >=3.4.0 (which won't be a problem since 3.6.x has been the minimum for many projects, such as gallery-dl and yt-dlp).
- Same as above for pathlib (which according to/judging by https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/ is a very nice alternative, even if it was only used for writing less code).

PR #29944, opened 2021-09-11
[Test] Improve download test

dirkf:
---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This PR makes two small and one larger improvement to `test/test_download.py`, in 3 separate commits. 

#### Calculating the file hash

External downloaders may not support limiting the download exactly to a specified size. When running a test download, only a short initial portion of the file is supposed to be downloaded, and the hash is calculated from that portion. If the portion is longer than requested the hash will be incorrect. The PR explicitly uses the requested length, if specified, when calculating the file hash, so that the same value should be calculated regardless of downloader.

#### Header-related params in test cases

Header-related params (here: `--user-agent`, `--referer`, `--add-header`) are parsed into `std_headers` rather than the `params` dict in the main yt-dl program. In download tests, only params that are set into the `params` dict were supported, and header-related params in test cases were ignored. This PR changes the params processing for test downloads to match what happens in the main program.

As an example, the change makes it possible to specify a test case with a specific UA, such as this for an Invidious site fronted by Cloudflare
```
        ...
        'params': {
            # Cloudflare breaks HTTP if Chrome is mentioned in the UA (2021-08)
            'user_agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/6533.18.5 (KHTML, like Gecko) Safari/6533.18.5',
        },                                       
        ...
```
#### Add TestSuite test_..._all with all downloads for each extractor

When testing downloads for an extractor XXX, a set of test cases is created but it's not obvious which item in the extractor's `_TESTS` list corresponds to which `TestDownload.test_XXX...` test case. If there are many test cases one might fail to run some of the later tests, or run past the highest numbered case.

To facilitate running the download tests for an extractor in the same way that they run in the CI servers, the PR adds a further download test case with name suffix `_all` for each extractor class that has test case(s). The new test case is a `unittest.TestSuite` that runs all the other download tests for that extractor.

Thus to test `YoutubeIE`:
```
python test/test_download.py TestDownload.test_Youtube_all
```


test/test_download.py
```diff
@@ -254,12 +276,49 @@ def try_rm_tcs_files(tcs=None):
...
-    test_method.__name__ = str(tname)
-    ie_list = test_case.get('add_ie')
-    test_method.add_ie = ie_list and ','.join(ie_list)
-    setattr(TestDownload, test_method.__name__, test_method)
-    del test_method
-
+    ie_list = ','.join(test_case.get('add_ie', []))
+    TestDownload.addTest(test_method, tname, ie_list)
+
+# Py2 compat (should be in compat.py?)
```
Grub4K, line 282: These are already in compat as `compat_filter` and `compat_map` [here](https://github.com/ytdl-org/youtube-dl/blob/master/youtube_dl/compat.py#L2970).

test/test_download.py
```diff
@@ -254,12 +276,49 @@ def try_rm_tcs_files(tcs=None):
...
+
+
+for ie_key in set(
+    map(lambda x: x[0],
+        filter(
+            lambda x: callable(x[1]),
+            map(lambda a: (a[5:], getattr(TestDownload, a, None)),
+                filter(lambda t:
+                       re.match(r'test_.+(?<!(?:_all|.._\d|._\d\d|_\d\d\d))$', t),
+                       dir(TestDownload)))))):
```
Grub4K, line 319: Why do we need to create the intermediate tuple?
```suggestion
for ie_key in set(
    map(lambda a: a[5:],
        filter(
            lambda x: callable(getattr(TestDownload, x, None)),
            filter(
                lambda t: re.match(r"test_.+(?<!(?:_all|.._\d|._\d\d|_\d\d\d))$", t),
                dir(TestDownload))))):
```

test/test_download.py
dirkf: Yes, I subsequently needed them in some other PR (after I received commit privs), so they were added there. Here they should be imported.

test/test_download.py
dirkf: Looks like the result of Lego development. It can be simplified.

dirkf:
Note to self:

- [ ] also handle case where a fragment is downloaded that is smaller than expected by the original code.

test/test_download.py
```diff
@@ -23,6 +23,7 @@
 import io
 import json
 import socket
+import re
 
 import youtube_dl.YoutubeDL
 from youtube_dl.compat import (
```
dirkf, line 29:
```suggestion
from youtube_dl.compat import (
    compat_filter as filter,
    compat_map as map,
```

test/test_download.py
```diff
@@ -254,12 +276,48 @@ def try_rm_tcs_files(tcs=None):
...
-    del test_method
-
+    ie_list = ','.join(test_case.get('add_ie', []))
+    TestDownload.addTest(test_method, tname, ie_list)
+
+# Py2 compat (should be in compat.py?)
+try:
+    from itertools import (ifilter as filter, imap as map)
+except ImportError:
+    pass
```
dirkf, line 286:
```suggestion
```

test/test_download.py
```diff
@@ -84,6 +90,13 @@ def strclass(cls):
                                strclass(self.__class__),
                                ' [%s]' % add_ie if add_ie else '')
 
+    @classmethod
+    def addTest(cls, test_method, test_method_name, add_ie):
+        test_method.__name__ = str(test_method_name)
+        test_method.add_ie = add_ie
+        setattr(TestDownload, test_method.__name__, test_method)
```
dirkf, line 97:
```suggestion
        setattr(cls, test_method.__name__, test_method)
```

test/test_download.py
```diff
@@ -57,9 +62,10 @@ def process_info(self, info_dict):
         return super(YoutubeDL, self).process_info(info_dict)
 
 
-def _file_md5(fn):
+def _file_md5(fn, length=None):
     with open(fn, 'rb') as f:
-        return hashlib.md5(f.read()).hexdigest()
+        return hashlib.md5(
+            f.read(length) if length is not None else f.read()).hexdigest()
```
dirkf, line 68:
```suggestion
            f.read() if length is None else f.read(length)).hexdigest()
```

test/test_download.py
```diff
@@ -23,9 +23,12 @@
 import io
 import json
 import socket
+import re
 
 import youtube_dl.YoutubeDL
 from youtube_dl.compat import (
+    compat_filter as filter,
+    compat_map as map,
     compat_http_client,
```
dirkf, line 32:
```suggestion
    compat_http_client,
    compat_map as map,
```

PR #29947, opened 2021-09-12
[CGTN] Add new extractor

chao813:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---


youtube_dl/extractor/cgtn.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import unicode_literals
```
dirkf, line 1: You could put a `# coding: utf-8` line at the top in case someone adds a test case with non-ASCII characters. Currently about 1/3 of extractors do this.

chao813: done!

youtube_dl/extractor/cgtn.py
```diff
@@ -0,0 +1,46 @@
...
+
+        title = self._html_search_regex(r'<div class="news-title">(.+?)</div>', webpage, 'title')
+        download_url = self._html_search_regex(r'data-video ="(?P<url>.+m3u8)"', webpage, 'download_url')
+        thumbnail = self._html_search_regex(r'<div class="cg-player-container J_player_container" .*? data-poster="(?P<thumbnail>.+jpg)" (.*?)>', webpage, 'thumbnail', fatal=False)
+
+        formats = self._extract_m3u8_formats(
+            download_url, video_id, 'mp4',
+            entry_protocol='m3u8_native', m3u8_id='hls')
+
+        return {
```
dirkf, line 42: Perhaps gather `description` too? Say,
```
        description = self._html_search_meta(('og:description', 'twitter:description', ), webpage)
```
Also perhaps an item in the `categories` list, `timestamp`, `creator`,  from these elements:
```

        <div class="news-info">
            <div class="news-section-date news-text">
             <span class="section">
                 China
             </span>
            <span class="date">

    11:30, 06-Jun-2021

            </span>
            </div>

...
             <div class="news-author news-text news-statistics-info J_news-statistics">
                    <div class="news-author-name">
                             CGTN
                    </div>

             </div>
...
        </div>
```
For `timestamp`, apparently `unified_timestamp()` doesn't know about the date-time format used above (and confuses itself that the year is a timezone if it does), so either decode it yourself, or try also importing the `DATE_FORMATS_DAY_FIRST` list from `utils.py and appending '%H:%M %d/%b/%Y' to it, then replace `-` with `/` in the date-time string before passing it to `unified_timestamp()`.

Issue #28900, opened 2021-04-28
Default to https instead?


eiais:
## Checklist

- [ ] I'm reporting a broken site support issue
- [x ] I've verified that I'm running youtube-dl version **2021.04.26**
- [ x] I've checked that all provided URLs are alive and playable in a browser
- [x ] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x ] I've searched the bugtracker for similar bug reports including closed ones
- [x ] I've read bugs section in FAQ

## Verbose log

```
No log
```

## Description

I would like to send a PR changing the default protocol from http to https. This could cause breakage in the case of downloading from sites that only support http. However, as of april 2021 over 90% of internet traffic is https [1]. Would there be an ideological opposition to this change? I searched github and couldn't find anything discussing it, it looks like it hasn't been touched since the default_search was added.

[1] https://transparencyreport.google.com/https/overview?hl=en

rautamiekka:
This has been here before, but I can't find the earlier one.

The answer was something along the lines of: ytdl (and the underlying HTTP lib) always takes HTTPS if no known incompatibilities exist.

Every extractor's URL RegEx is created with just-in-case support for both HTTP and HTTPS cuz adding 2 chars (`s?`) doesn't hurt the performance. Plus it makes life easier by not having to fix the address.

dstftw:
What do you mean by "changing the default protocol" at all?

eiais:
https://github.com/ytdl-org/youtube-dl/blob/master/youtube_dl/extractor/generic.py#L2367 if it actually pulling it over https this is a confusing message.

dstftw:
This is expected behavior. Changing scheme to `https://` will break cases when URL is only available via `http://`. At the same time `http://` works fine for both cases since webservers are usually configured to redirect to https when https is preferred so that `http://` will end up processing https anyway.

PR #29950, opened 2021-09-12
Use HTTPS as default protocol instead of HTTP

YoshiRulz:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

HTTP bad

I doubt video downloads are getting MITM'd because of this fallback, but better safe than sorry.

### DIFF
--- a/youtube_dl/extractor/generic.py
+++ b/youtube_dl/extractor/generic.py
@@ -2344,8 +2344,8 @@ class GenericIE(InfoExtractor):
 
             if default_search in ('auto', 'auto_warning', 'fixup_error'):
                 if re.match(r'^[^\s/]+\.[^\s/]+/', url):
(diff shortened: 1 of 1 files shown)


rautamiekka:
Oddly I can't find the thread where this was already denied, but it was along the lines of HTTP websites still existing, and if a site doesn't wanna support HTTP they'll auto-redirect to HTTPS, and things like passwords are already encrypted.

Besides, MITM ? The chances are non-existent I think.

dirkf:
See, eg, issue #28900.

Maybe the PR would be more acceptable (supposing any PRs are ever accepted again) if it included an option such that HTTP remained the default but a user could make it HTTPS with a command-line or config option if desired.

rautamiekka:
> See, eg, issue #28900.

Yes, that's the one.

PR #29973, opened 2021-09-15
[aes] Add signature conversion decorator to `aes`

sulyi:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

It starts an overhaul of `aes`.
- [x] Simplifies signature of functions in the module, instead list of integers they accept bytes.
- [x] Adds function for AES-128-GCM  decrypt (uses already existing function for AES-CRT) and verification.
- [ ] Adds new functionality to existing code (i.e. AES-CRT nonce)
- [ ] Converts (internal) functions to also use mentioned types.
- [ ] Restructures the code according to OOP design
- [ ] Adds new tests cases to `test_aes` to safeguard code

I'd like to mention @pukkandan and the help he gave me during  the development of this code at https://github.com/yt-dlp/yt-dlp/pull/938.

### DIFF
--- a/youtube_dl/aes.py
+++ b/youtube_dl/aes.py
@@ -2,39 +2,70 @@ from __future__ import unicode_literals
 
 from math import ceil
 
-from .compat import compat_b64decode
+from .compat import compat_b64decode, compat_struct_pack
 from .utils import bytes_to_intlist, intlist_to_bytes
 
 BLOCK_SIZE_BYTES = 16
 
 
-def aes_ctr_decrypt(data, key, counter):
+def signature_conversion(args_conv=None, kwargs_conv=None, ret_conv=None):
+    def decorator(function):
+        def wrapper(*args, **kwargs):
+            result = function(*args_conv(args) if args_conv else args,
+                              **kwargs_conv(kwargs) if kwargs_conv else kwargs)
+            return ret_conv(result) if ret_conv else result
+        return wrapper
+    return decorator
+
+
+def convert_all_args_to_intlist(arguments):
+    return map(bytes_to_intlist, arguments)
+
+
+def convert_expanded_key_args(arguments):
+    data, key = arguments
+    return [bytes_to_intlist(data), key_expansion(bytes_to_intlist(key))]
+
+
+def _aes_ctr_decrypt(data, key, iv):
     """
     Decrypt with aes in counter mode
 
     @param {int[]} data        cipher
     @param {int[]} key         16/24/32-Byte cipher key
-    @param {instance} counter  Instance whose next_value function (@returns {int[]}  16-Byte block)
-                               returns the next counter block
+    @param {int[]} iv          16-Byte initialization vector
     @returns {int[]}           decrypted data
+    """
+    return _aes_ctr_encrypt(data, key, iv)
+
+
+def _aes_ctr_encrypt(data, key, iv):
+    """
+    Encrypt with aes in counter mode
+
+    @param {int[]} data        cleartext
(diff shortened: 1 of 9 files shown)


sulyi:
I have the test just haven't added them yet.

dirkf:
Why have the `aes_...` functions been given a leading `_`, contrary to PEP8?

Is this halfway to refactoring the code with these functions as static methods of some class?

The list `__all__` of functions in the module `aes.py` doesn't have the `_`.

sulyi:
> Why have the aes_... functions been given a leading _, contrary to PEP8?

There's no contradiction here, they are intended yo be private.

> Is this halfway to refactoring the code with these functions as static methods of some class?

I wouldn't use static but, kind of yes. Moving functions into a class is secondary, but rewriting them to use byte instead of list of ints is.

> The list __all__ of functions in the module aes.py doesn't have the _.

Private functions are exposed [here](https://github.com/sulyi/youtube-dl-rg3_epitaph/blob/aa6f0402766f74916a35932764997812c9568cbc/youtube_dl/aes.py#L232-L264).

youtube_dl/aes.py
```diff
@@ -80,50 +111,75 @@ def aes_cbc_encrypt(data, key, iv):
...
         previous_cipher_block = encrypted_block
 
     return encrypted_data
 
 
-def key_expansion(data):
+def _aes_gcm_decrypt_and_verify(data, key, tag, nonce):
     """
-    Generate key schedule
+    Decrypt with aes in GBM mode and checks authenticity using tag
```
pukkandan, line 124: Since #29201 is not merged, youtube-dl does not have any need for gcm

dirkf:
OK, now I see the new block (which after all is mentioned in the title of the PR!) that all makes sense, thanks.

Making the lists of int internal certainly seems a good idea. The decorator mechanism is slick and functional, and systematises the code, even if it doesn't always reduce it by much. Eg AIUI these are equivalent?
```
@signature_conversion(convert_all_args_to_intlist, None, intlist_to_bytes)
def aes_ctr_decrypt(data, key, iv):
    return _aes_ctr_decrypt(data, key, iv)
```
vs
```
def aes_ctr_decrypt(data, key, iv):
    return intlist_to_bytes(
        _aes_ctr_decrypt(bytes_to_intlist(data), bytes_to_intlist(key), bytes_to_intlist(iv)))
```
But some of the other interface signatures need more complex munging and so should be more simply expressed by the decorator.

sulyi:
Nope, that's is done [here](https://github.com/sulyi/youtube-dl-rg3_epitaph/blob/aa6f0402766f74916a35932764997812c9568cbc/youtube_dl/aes.py#L14-L15) and [here](https://github.com/sulyi/youtube-dl-rg3_epitaph/blob/aa6f0402766f74916a35932764997812c9568cbc/youtube_dl/aes.py#L21-L22).

But, @dirkf, you got me thinking whether it is possible to slap decorator directly on original functions.

youtube_dl/aes.py
sulyi: True, since I did it I just want it out somewhere...

pukkandan:
> you got me thinking whether it is possible to slap decorator directly on original functions.

If you never need access to the original functions (even internally), yes. You can just put the decorator on the original without needing a wrapper

sulyi:
My original assumption that using `bytes` would be better is a bust: https://bugs.python.org/issue19251.
But I was yet able to make some improvements, and imho, the new signature is still friendlier.

PR #29978, opened 2021-09-17
Youtube: Add broadcast start time (if it was a live stream)

arilou:
Signed-off-by: Jon Doron <arilou@gmail.com>

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Add to the youtube JSON the broadcast start time in case if it was a live stream upload.

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -1780,6 +1780,10 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
             or parse_duration(search_meta('duration'))
         is_live = video_details.get('isLive')
         owner_profile_url = microformat.get('ownerProfileUrl')
+        live_details = microformat.get('liveBroadcastDetails')
+        broadcast_start_timestamp = None
+        if live_details:
+            broadcast_start_timestamp = live_details.get('startTimestamp')
 
         info = {
             'id': video_id,
@@ -1809,6 +1813,7 @@ class YoutubeIE(YoutubeBaseInfoExtractor):
             'categories': [category] if category else None,
             'tags': keywords,
             'is_live': is_live,
+            'broadcast_start_timestamp': broadcast_start_timestamp,
         }
 
         pctr = try_get(


youtube_dl/extractor/youtube.py
```diff
@@ -1809,6 +1813,7 @@ def feed_entry(name):
             'categories': [category] if category else None,
             'tags': keywords,
             'is_live': is_live,
+            'broadcast_start_timestamp': broadcast_start_timestamp,
```
dirkf, line 1816: Rather than introducing a new metadata parameter, why shouldn't this just be the `timestamp` of the item, when it's a live show?
>    timestamp:      UNIX timestamp of the moment the video became available

arilou:
To be honest it would probably always be the same as upload_date but the thing I needed to have the exact time of day for the start in order to synchronize this later with the chat items.
And from what I saw upload_date is being set after being transformed via unified_strdate, which removes the full time and date information leaving you only with the YYYYMMDD.

So in order not to break the API I figured it will be better to introduce a new field which has the date in full form of RFC3399

dirkf:
> Or did you mean to simply rename the field to "timestamp" from "broadcast_start_timestamp"? and convert the RFC3399 to unix time?

Absolutely. If an extractor returns a `timestamp` but no `upload_date`, the core processing creates an `upload_date`; but not the other way round. You're quite safe to use `timestamp` as an epoch (UNIX) time. The `unified_timestamp()` utility routine can be used, if needed, to convert the value.

The existing API uses names ending in `_date` for YYYYMMDD strings and `timestamp` for UNIX time values.

pukkandan:
I recommend using `release_timestamp` instead, since `upload_date` and `timestamp` are supposed to point to the same data.

(This is how yt-dlp handles it, so I may be biased 😛)

arilou:
Heh ok renamed to release_timestamp and set it to be UNIX time

youtube_dl/extractor/youtube.py
```diff
@@ -1813,16 +1813,10 @@ def feed_entry(name):
...
             broadcast_start_timestamp = live_details.get('startTimestamp')
-
-        if broadcast_start_timestamp:
-            info['timestamp'] = parse_iso8601(broadcast_start_timestamp)
-        else:
-            info['upload_date'] = unified_strdate(
-                microformat.get('uploadDate')
-                or search_meta('uploadDate'))
+            if broadcast_start_timestamp:
+                info['release_timestamp'] = parse_iso8601(broadcast_start_timestamp)
```
dirkf, line 1819: Looks like a good solution.

PR #25475, opened 2020-05-31
Truncate filenames that are too long

TheStranjer:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [x] At least skimmed through [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

This fixes issues #7765 and #5908. I had a similar problem when I was trying to download a Twitter video and it tried to set the entire Tweet body as the filename. This does it in the common downloader, and if that doesn't give it _general_ applicability I'd love to hear how I can make it accomplish that.

PR #29989, opened 2021-09-19
[utils] Shorten proposed file name on create if too long

### LINKED ISSUE #5908: Vimeo: [Errno 36] File name too long
Filesystem: EXT4
OS: Ubuntu Linux 14.04 LTS
`youtube-dl --version`: 2015.06.04.1
When I try to download a Vimeo video:

`youtube-dl --ignore-errors --restrict-filenames https://vimeo.com/80352108`

I got the following error:

`ERROR: unable to open for writing: [Errno 36] File name too long: ...`

### LINKED ISSUE #7765: [Errno 36] File name too long
Hi,

./youtube-dl --ignore-errors --no-playlist --all-formats "http://www.hardsextube.com/video/1152690/Lovely-Arabian-Muslim-Kareemah-fucked-by-Small-Four-Inch-Oriental-Arab-Dick-on-Webcam"
[generic] Lovely-Arabian-Muslim-Kareemah-fucked-by-Small-Four-Inch-Oriental-Arab-Dick-on-Webcam: Requesting header
[generic] Lovely-Arabian-Muslim-Kareemah-fucked-by-Small-Four-Inch-Oriental-Arab-Dick-on-Webcam: Downloading webpage
[generic] Lovely-Arabian-Muslim-Kareemah-fucked-by-Small-Four-Inch-Oriental-Arab-Dick-on-Webcam: Extracting information
ERROR: unable to open for writing: [Errno 36] File name too long: 'Lovely Arabian Muslim Kareemah fucked by Small Four Inch Oriental Arab Dick on Webcam - Hardsextube-s7krq56ztqlqv0bv3bcnm5c0hht455zA5n6687k84bzc4qx3y2q84wchgbnrcj8qzvk1vtzlqkAxzA6y5qb57Aq2cw8rrkgb2t6cfvpl2blm0pcnA5A9zAgA04Ahs3kgs3tfv6ykyh6qjl92d4tzhb536036h8dn912skkd1zqq4jp1wv51y4n003vcrjgl2rjwq3n155222hkshrv5zfpx28560jwpl66jnsh7z1jj36mAjkjq9dAdzqgAm207x57bn8n05pmh68b5xq82n39kpzft8w44lj40z69sch6p98gwAghsrt8lx5vzmcd7n8mAtd3s07dc0hsxstnsk5tsq4k427y0fk95gqjqyqvs1xdhg1l00c45bf0hvtlc4.unknown_video.part'

### LINKED ISSUE #29627: file name too long，https://ok.ru/video/1632457787997?fromTime=459
youtube-dl https://ok.ru/video/1632457787997?fromTime=459

[Odnoklassniki] 1632457787997: Downloading webpage
[Odnoklassniki] 1632457787997: Downloading m3u8 information
[hlsnative] Total fragments: 516
[download] Destination: MARTY 1955 - Ernest Borgnine, Betsy Blair, Esther Minciotti, Augusta Ciolli, Joe Mantell, Karen Steele, Jerry Paris, James Bell, Joe Bell, John Beradino, Nick Brkich, Marvin Bryan, Charles Cane, Paddy Chayefsky, John Dennis, Steven Hecht, Paul Hoffman, Doris Kemper, Jack Klugman-1632457787997.mp4
ERROR: unable to download video data: [Errno 36] File name too long: 'MARTY 1955 - Ernest Borgnine, Betsy Blair, Esther Minciotti, Augusta Ciolli, Joe Mantell, Karen Steele, Jerry Paris, James Bell, Joe Bell, John Beradino, Nick Brkich, Marvin Bryan, Charles Cane, Paddy Chayefsky, John Dennis, Steven Hecht, Paul Hoffman, Doris Kemper, Jack Klugman-1632457787997.mp4.ytdl'

### LINKED ISSUE #29912: «File name too long» should not error
## Checklist

- [x] I'm reporting a feature request
- [x] I've verified that I'm running youtube-dl version **2021.06.06**
- [x] I've searched the bugtracker for similar feature requests including closed ones

## Description

To reproduce:
- try to download any video with a really long title
- you get `ERROR: unable to open for writing: [Errno 36] File name too long:`
- the video is not downloaded

My suggestion:
- instead of an `ERROR` put a `WARNING`
- download the video regardless, just truncate the title to an appropriate size, similar to what suggested in #29627

This will improve UX, check [bug tracker](https://github.com/ytdl-org/youtube-dl/issues?q=%22file+name+too+long%22) to see how common this issue is.

#7765 talks about this too (with a linked pull request) but:
- does not elaborate on whether it is actually a good thing / possible drawbacks
- is limited to collaborators

### LINKED ISSUE #29975: Excuse, when you use the >>title<< parameter to specify the output file name in windows, if you encounter a long video file name, a file system write error will be reported.
image

1. Excuse , when you use the >>title<< parameter to specify the output file name in windows, if you encounter a long video file name, a file system write error will be reported.
1.请问,在windows中使用>>title<<参数指定输出文件名的时如果碰到超长的视频文件名称会报出文件系统写入错误.

2. How do I get the local video name for youtube-dl? I use a proxy server to download a youtube video in China, where I see the video name is a Chinese name, but when I download it using youtube-dl, it becomes In English, I tried other methods such as using cookies and changing IP, and always youtubedl gets an English name. PS: As stated in 1, if you get some videos with very long title names, youtubedl cannot be downloaded under the windows system I guess it is the 255 directory/file name restrictions of the windows system. I can only try to output the download directory to the root directory of the disk to accommodate as many and longer file names as possible.
2.youtube-dl候如何获取本地视频名称?我在中国使用代理服务器下载一个youtube视频,在我这里我看到此视频名称是一个中文名称,但是使用youtube-dl下载的时候它就变成了英文的,我尝试使用了cookies和更换IP等其他办法,始终youtubedl获取的是一个英文名称.PS:在如1所表述,如果到一些title名字超长的视频,在windows系统下youtubedl就无法下载了.我猜测是windows系统的255目录/文件名等限制.我只能试着把下载目录输出到磁盘的根目录尽量容纳更多更长的文件名.

3.Functional requirements, you can limit the length of the file name when you want to use the title tag to output the file name.
3.功能要求,希望使用title标签输出文件名时可以限制其长度.

dirkf:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

If the destination filename constructed from the output template is too long for the filesystem, the yt-dl download fails. The user can't predict how long extracted fields (eg `title`) may be, so this may be unexpected. The user then has to work out how to modify the output template (eg `%(title)s` -> `%(title).25s`) and run the command again.

Consequently there have been suggestions (of which #29912 is the most recent) that the destination filename should be shortened automatically. This doesn't seem unreasonable, since `sanitize_path()` already has the potential to edit the filename. PR #25475 failed because it was platform-specific.

This PR adds a function `reduce_filename()` to `utils.py` that shortens the filename component by a specified  factor, and adds logic to `sanitize_open()` to detect `errno.ENAMETOOLONG` and iteratively call `reduce_filename()` until opening either succeeds, or fails for some other reason.

Resolves #29912, resolves #29975, resolves #29627, resolves  #7765,  resolves #5908, probably others.


dirkf:
Apparently the suspicion that I voiced [here](https://github.com/ytdl-org/youtube-dl/issues/29912#issuecomment-914371090) is valid.

For UNIX-like platforms, an over-long filename gives ENAMETOOLONG as expected.

For Windows, there may be two cases:
* if the over-long filename is caught by the [Windows API](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea), we get EINVAL (contrary to Microsoft's [`open()` documentation](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/open-wopen?view=msvc-160); perhaps it's in the CPython interface layer);
* possibly, if the over-long filename is detected at the filesystem level (eg, with the `\\?\...` filename syntax), we might still get ENAMETOOLONG.

Any test results from versions of Windows other than what's in the CI VMs, or from OSX, would be welcome.

dirkf:
The aim is to avoid what people have often complained about, namely having a download fail just because the proposed filename is too long for the file system:
* the `textwrap` shortening does occur in the middle;
* the `--download-archive` option exists for users who want to record the exact ID of each download, but perhaps there should be a variation, say `--download-archive-index`, which records both the ID and the filename under which the item was saved?
* possibly the default values for the `reduce_filename()` parameters should be varied? 
* if you can suggest some test cases for temp files I'll be happy to add them, but the `sanitize_open()` function is expected to change the proposed filename and it's up to the caller to manage the result.

tylerszabo:
I'm still concerned that this will create more bugs than it fixes - failing fast with a clear error message or altering default templates to reduce likelihood of exceeding length will both offer more predicable behavior without introducing additional edge cases without test coverage.

The same URL on the same OS with the same template will result in different filenames of saved on different filesystems with this change.

dirkf:
Again, many users don't consider it acceptable that a download should fail just because of filename length restrictions. Personally, I haven't suffered from this thanks to ext2/4, but I don't believe that it can be a bug if the download is successfully saved under some reasonable filename, actually the reverse.

Plainly the likelihood of needing to shorten filenames can be reduced if extractors don't (Twitter, eg) generate pointlessly long titles or IDs. But I do agree with users who are _disappointed_ when their download that appeared to work fails because of some filename issue. 

Already we have the `--restrict-filenames` option which will corrupt IDs containing `&`.
* Should the PR introduce `--[no-]restrict-filename-length` to en/[dis]/able its functionality?
* Should there be a "WARNING: yt-dl wanted to save this as 'xxx' but the file name was shortened to 'xx'"?

pukkandan:
If the actual filename is different from the one in the info-dict, postprocessors could fail

> Since the sanitize_open() function can change the proposed filename already, yt-dl has to be able to reassemble fragments whose filenames may not be as expected

Actually, no. It appears like that when looking at `sanitize_open`, but `sanitize_path` is also called when creating the filename. https://github.com/ytdl-org/youtube-dl/blob/af9e72507ea38e5ab3fa2751ed09ec88021260cb/youtube_dl/YoutubeDL.py#L721

So the except block in sanitize_open doesn't actually do anything. I haven't looked at the blame, but I assume it is a remnant from before output template was a thing

dirkf:
Thanks for the tip.

Looking again, it seems that `sanitize_open()` is called when downloading, though not necessarily with the exact filename that came from the quoted line of `prepare_filename()`. First of all `'f%s' % f['format_id']` is inserted into the prepared filename. Then in the HTTP downloader `.part` is added, and in the Fragment downloader `.ytdl` is added.

A `format_id` is forced to be a string and then sanitised to remove spaces and format selection syntax (but there was a case where a character generated by the extractor in a `format_id` still clashed with format selection syntax: =, BBC?). The characters added in the downloader are valid according to `sanitize_path()`, but if a `format_id` contained an invalid Windows filename character that wasn't sanitised (`:`, say), the except block of `sanitize_open()` could be entered. However the second call to `sanitize_path()` shouldn't affect the stem resulting from `prepare_filename()`. 

If the inserted and appended elements happen to cause the filename to exceed a length limit, the except block won't help, and the error will be re-raised.

So this is a problem:
* naïvely abbreviating the filename in the downloader could give a different stem according to the length of the `format_id`, invalidating the caller's `info_dict`;
* as there doesn't seem to be length limit for a `format_id`, either a limit has to be imposed, or a fake `format_id` could be used (`'f{i}' for i in enumerate(formats)`), or the maximum `format_id` length for the current item could be determined, so that a known additional length can be allowed for when abbreviating the filename;
* if the filename is abbreviated when `sanitize_open()` is called, the inserted/appended elements must have a known maximum length and be ignored when abbreviating, or 
* the filename must be abbreviated, allowing for the `format_id` and other elements potentially added by the downloader, before calling any downloader, say by creating a temporary file with dummy additional elements.

Gruesome, really.

pukkandan:
> as there doesn't seem to be length limit for a format_id, either a limit has to be imposed, or a fake format_id could be used ('f{i}' for i in enumerate(formats)), or the maximum format_id length for the current item could be determined, so that a known additional length can be allowed for when abbreviating the filename;

We could also hash the format id if it is too long. But there may be people who depend on the intermediate filenames in their scripts

I also thought about having the downloader return the final filename and then overwriting it in the infodict. But this would cause the `--get-filename` and actual filename to be different, potentially breaking people's workflows. To avoid that, it seems necessary that the final filename be determined in `prepare_filename` itself. We can sanitize the temporary filenames again in the downloader. It is mostly not an issue as long as the final filename can be fixed ahead of time.

akwiatkowski:
While I wait to be merged I added my own alias which helped. Keep in mind I'm not bash expert and truncation is not included here but this helped me.

```
alias ydl='function _ydl() { command="youtube-dl \"$1\" -o \"`youtube-dl --get-filename \"$1\" | perl -pe \"s/[^a-zA-Z0-9\. ]//g\" | perl -pe \"s/\s{2,}/ /g\" `\" "; echo "$command"; bash -c "$command"; }; _ydl'
```

rubyFeedback:
unfortunately I am really bad with bash code. Can we have some work around like an additional commandline flag? Something like "--safe-download" or so, that will download no matter what. Right now I don't know how to get the file name if the name is too long, yt-tdl just errors out on me and nothing gets downloaded as a result.

dirkf:
Just [put an output template specification](https://github.com/ytdl-org/youtube-dl/issues/30361#issuecomment-994233831) in your config file or on the command-line, replacing the default `%(title)s-%(id)s.%(ext)s`:
```
-o "%(title).200s-%(id)s.%(ext)s
```
Or change 200 to a smaller number if that doesn't work for you.

mzhboy:
for CJK users 200 is still too long.
my suggestion:
230/3=76 suitable for most cases
or
230/4=57 make sure filename alway fit

dirkf:
This is why a better solution is the approach attempted in the PR: ie, detecting what actual length is valid.

kapitaali:
> 

I would love to use this but EXT filesystem says that the file name is too long. Otherwise a great hack and I'd like to use it if it would shorten it. 

So to add an extra layer of regex, or how would you shorten the length to 255 characters?

dirkf:
Just use the [known work-around](https://github.com/ytdl-org/youtube-dl/issues/29912#issuecomment-985746897).

Issue #29969, opened 2021-09-15
Fravnce TV 2 journal not downloading any more


kdliss:
since about yesterday, the daily news journal does not download any more, from FRance TV2:
https://www.france.tv/france-2/journal-20h00/2743433-edition-du-mardi-14-septembre-2021.html

moviuro:
Also broken, the live video feed:

* https://www.france.tv/france-2/direct.html
* https://www.france.tv/france-3/direct.html
* https://www.france.tv/franceinfo/direct.html

```
% mpv --ytdl-raw-options-add=verbose=1 'https://www.france.tv/franceinfo/direct.html'
Andreas Waldetoft - Towards Utopia
[paused]  #60/189   0:14/8:07 (2%)
volume: 96%   repeat: on    random: off   single: off   consume: off
[ytdl_hook] [debug] System config: []
[ytdl_hook] [debug] User config: ['-o', '%(uploader)s-%(upload_date)s-%(title)s.%(ext)s']
[ytdl_hook] [debug] Custom config: []
[ytdl_hook] [debug] Command-line args: ['--no-warnings', '-J', '--flat-playlist', '--sub-format', 'ass/srt/best', '--format', 'bestvideo+bestaudio/best', '--verbose', '1', '--all-subs', '--no-playlist', '--', 'https://www.france.tv/franceinfo/direct.html']
[ytdl_hook] [debug] Encodings: locale UTF-8, fs utf-8, out utf-8, pref UTF-8
[ytdl_hook] [debug] youtube-dl version 2021.06.06
[ytdl_hook] [debug] Python version 3.9.7 (CPython) - Linux-5.12.18-1-ck-skylake-x86_64-with-glibc2.33
[ytdl_hook] [debug] exe versions: ffmpeg 4.4, ffprobe 4.4
[ytdl_hook] [debug] Proxy map: {}
[ytdl_hook] ERROR: '1' is not a valid URL. Set --default-search "ytsearch" (or run  youtube-dl "ytsearch:1" ) to search YouTube
[ytdl_hook] Traceback (most recent call last):
[ytdl_hook]   File "/usr/lib/python3.9/site-packages/youtube_dl/YoutubeDL.py", line 815, in wrapper
[ytdl_hook]     return func(self, *args, **kwargs)
[ytdl_hook]   File "/usr/lib/python3.9/site-packages/youtube_dl/YoutubeDL.py", line 836, in __extract_info
[...]

Issue #29965, opened 2021-09-14
Project dead?


cypheron:
Who is left maintaining still project? Even though there are more than 3,800 issues and almost 1k pull requests there hasn't been any commit for more than 2 months.

Also there has been no feedback on most of the proposed extractors by the community, for which I created an [overview](https://github.com/ytdl-org/youtube-dl/issues/28054) a while back. It would be nice to see at least a comment from the maintainers.

dirkf:
Supposing the current project owner and maintainers are not just on a summer break, or a death march project, it would have to be decided whether the project can attract enough support to continue with the same aims, either as-is, or in a fork, with new maintainers.

The yt-dlp project seems to be attracting most interest nowadays, but does not support Python 2, nor even Python < 3.6, which would seem to be the distinguishing feature between the two projects (there are quite few UI and other enhancements as well). Leading Linux distros that are still in extended support came with Python 2 and 3.5 as an option.

I would like to support a Python 2 platform (where the host is a set-top box for which Python 3.6+ may never be built), but it might be possible to do so by cherry-picking and adapting yt-dlp changes.

89z:
I have an alternative package/program here:

https://github.com/89z/mech

Its written in Go, so you dont need Python, and you dont even need Go if you just want to use the program.

dirkf:
No doubt a fine work, even if the site support is limited compared with yt-dl, but there are good reasons why @rg3's choice of Python for yt-dl was successful and they are still valid.

For instance, the platform to which I referred was frozen before Go was created and has trouble with GCC beyond v4.5. Even if there is some Py3 code feature it can usually be compat-ed out to run in older versions.

evilangelmd:
> there are quite few UI and other enhancements as well

didn't knew cli program has UI :D

dirkf:
It's difficult to use a program with no UI, even if not a GUI! But the point was to contrast new command-line options, which may just be a different way to set the program's parameters, with internal changes like multi-threaded download.

aeiouaeiouaeiouaeiouaeiouaeiou:
The lead maintainer is @dstftw. I began to believe the theory that he passed away from COVID.

fastily:
Wish I could sticky this somewhere.  The (actively maintained) spiritual successor is [yt-dlp](https://github.com/yt-dlp/yt-dlp)

ahm750:
The problem with yt-dlp is that it's flagged as a virus by some anti-virus programs.

[...]

Issue #247, opened 2011-12-10
myvideo.de download broken


strex:
Hi,
it seems that the myvideo.de download function is broken.

```
# youtube-dl http://www.myvideo.de/watch/8328930/Nickelback_When_We_Stand_Together_Official_Video
[myvideo] 8328930: Downloading webpage
[myvideo] 8328930: Extracting information
ERROR: Unable to download video

# youtube-dl http://www.myvideo.de/watch/8357965/Maria_Mena_Homeless
[myvideo] 8357965: Downloading webpage
[myvideo] 8357965: Extracting information
ERROR: Unable to download video
```

ilovelibertyfreedom:
You may also try this method to download video from myvideo at http://www.allavsoft.com/how-to/myvideo-downloader.html﻿

PR #29996, opened 2021-09-20
Sarnoud francetv

sarnoud:

---

- [ X] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [ X] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [ X] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [ X] Covered the code with tests (note that PRs without tests will be REJECTED)
- [ X] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

- [ X] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)

- [ X] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

The initial goal is to fix the newly broken francetv extractor (see issue https://github.com/ytdl-org/youtube-dl/issues/29956#issue-994870547)

Few changes:
1. extractor fixes for new backend
2. removed obsolete extractors in the francetv.py file (they now all redirect to the main site)
3. Fixed tests
4. implemented the ability to extract caption streams from m3u8 (behind include_subtitles flag)
5. implemented a new field 'download' that is a lambda function to retrieve subtitles (along 'data' and 'url')

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1879,6 +1879,8 @@ class YoutubeDL(object):
                         except (OSError, IOError):
                             self.report_error('Cannot write subtitles file ' + sub_filename)
                             return
+                    elif sub_info.get('downloader') is not None:
+                        sub_info.get('downloader')(self, encodeFilename(sub_filename))
                     else:
                         try:
                             sub_data = ie._request_webpage(
--- a/youtube_dl/extractor/francetv.py
+++ b/youtube_dl/extractor/francetv.py
@@ -50,14 +50,11 @@ class FranceTVIE(InfoExtractor):
     _TESTS = [{
         # without catalog
         'url': 'https://sivideo.webservices.francetelevisions.fr/tools/getInfosOeuvre/v2/?idDiffusion=162311093&callback=_jsonp_loader_callback_request_0',
-        'md5': 'c2248a8de38c4e65ea8fae7b5df2d84f',
+        'md5': '283491d723a14db7c4e10b887c4b475a',
         'info_dict': {
             'id': '162311093',
             'ext': 'mp4',
             'title': '13h15, le dimanche... - Les mystères de Jésus',
-            'description': 'md5:75efe8d4c0a8205e5904498ffe1e1a42',
-            'timestamp': 1502623500,
-            'upload_date': '20170813',
         },
     }, {
         # with catalog
@@ -252,9 +249,6 @@ class FranceTVSiteIE(FranceTVBaseInfoExtractor):
             'id': 'ec217ecc-0733-48cf-ac06-af1347b849d1',
             'ext': 'mp4',
             'title': '13h15, le dimanche... - Les mystères de Jésus',
-            'description': 'md5:75efe8d4c0a8205e5904498ffe1e1a42',
-            'timestamp': 1502623500,
-            'upload_date': '20170813',
         },
         'params': {
             'skip_download': True,
@@ -316,55 +310,26 @@ class FranceTVSiteIE(FranceTVBaseInfoExtractor):
         return self._make_url_result(video_id, catalogue)
 
 
-class FranceTVEmbedIE(FranceTVBaseInfoExtractor):
-    _VALID_URL = r'https?://embed\.francetv\.fr/*\?.*?\bue=(?P<id>[^&]+)'
-
-    _TESTS = [{
-        'url': 'http://embed.francetv.fr/?ue=7fd581a2ccf59d2fc5719c5c13cf6961',
-        'info_dict': {
-            'id': 'NI_983319',
-            'ext': 'mp4',
-            'title': 'Le Pen Reims',
-            'upload_date': '20170505',
-            'timestamp': 1493981780,
-            'duration': 16,
-        },
-        'params': {
-            'skip_download': True,
-        },
-        'add_ie': [FranceTVIE.ie_key()],
-    }]
-
-    def _real_extract(self, url):
-        video_id = self._match_id(url)
-
-        video = self._download_json(
-            'http://api-embed.webservices.francetelevisions.fr/key/%s' % video_id,
-            video_id)
-
-        return self._make_url_result(video['video_id'], video.get('catalog'))
-
-
 class FranceTVInfoIE(FranceTVBaseInfoExtractor):
     IE_NAME = 'francetvinfo.fr'
     _VALID_URL = r'https?://(?:www|mobile|france3-regions)\.francetvinfo\.fr/(?:[^/]+/)*(?P<id>[^/?#&.]+)'
 
     _TESTS = [{
-        'url': 'http://www.francetvinfo.fr/replay-jt/france-3/soir-3/jt-grand-soir-3-lundi-26-aout-2013_393427.html',
+        'url': 'https://www.francetvinfo.fr/replay-jt/france-3/soir-3/jt-grand-soir-3-lundi-26-aout-2019_3569073.html',
         'info_dict': {
-            'id': '84981923',
+            'id': 'e49f9ff0-2177-458e-830f-a28eccf19dd1',
             'ext': 'mp4',
             'title': 'Soir 3',
-            'upload_date': '20130826',
-            'timestamp': 1377548400,
             'subtitles': {
-                'fr': 'mincount:2',
+                'fr': 'mincount:1',
             },
         },
         'params': {
             'skip_download': True,
+            'format': 'dash-video=118000+dash-audio_fre=192000',
         },
         'add_ie': [FranceTVIE.ie_key()],
+        'expected_warnings': 'Unknown MIME type application/mp4 in DASH manifest',
     }, {
         'url': 'http://www.francetvinfo.fr/elections/europeennes/direct-europeennes-regardez-le-debat-entre-les-candidats-a-la-presidence-de-la-commission_600639.html',
         'only_matching': True,
@@ -389,6 +354,10 @@ class FranceTVInfoIE(FranceTVBaseInfoExtractor):
             'uploader_id': 'x2q2ez',
         },
         'add_ie': ['Dailymotion'],
+        'params': {
+            # TODO: the download currently fails (FORBIDDEN) - fix and complete the test
+            'skip_download': True,
+        },
     }, {
         'url': 'http://france3-regions.francetvinfo.fr/limousin/emissions/jt-1213-limousin',
         'only_matching': True,
@@ -417,139 +386,3 @@ class FranceTVInfoIE(FranceTVBaseInfoExtractor):
             webpage, 'video id')
 
         return self._make_url_result(video_id)
-
-
-class FranceTVInfoSportIE(FranceTVBaseInfoExtractor):
-    IE_NAME = 'sport.francetvinfo.fr'
-    _VALID_URL = r'https?://sport\.francetvinfo\.fr/(?:[^/]+/)*(?P<id>[^/?#&]+)'
-    _TESTS = [{
-        'url': 'https://sport.francetvinfo.fr/les-jeux-olympiques/retour-sur-les-meilleurs-moments-de-pyeongchang-2018',
-        'info_dict': {
-            'id': '6e49080e-3f45-11e8-b459-000d3a2439ea',
-            'ext': 'mp4',
-            'title': 'Retour sur les meilleurs moments de Pyeongchang 2018',
-            'timestamp': 1523639962,
-            'upload_date': '20180413',
-        },
-        'params': {
-            'skip_download': True,
-        },
-        'add_ie': [FranceTVIE.ie_key()],
-    }]
-
-    def _real_extract(self, url):
-        display_id = self._match_id(url)
-        webpage = self._download_webpage(url, display_id)
-        video_id = self._search_regex(r'data-video="([^"]+)"', webpage, 'video_id')
-        return self._make_url_result(video_id, 'Sport-web')
-
-
-class GenerationWhatIE(InfoExtractor):
-    IE_NAME = 'france2.fr:generation-what'
-    _VALID_URL = r'https?://generation-what\.francetv\.fr/[^/]+/video/(?P<id>[^/?#&]+)'
-
-    _TESTS = [{
-        'url': 'http://generation-what.francetv.fr/portrait/video/present-arms',
-        'info_dict': {
-            'id': 'wtvKYUG45iw',
-            'ext': 'mp4',
-            'title': 'Generation What - Garde à vous - FRA',
-            'uploader': 'Generation What',
-            'uploader_id': 'UCHH9p1eetWCgt4kXBYCb3_w',
-            'upload_date': '20160411',
-        },
-        'params': {
-            'skip_download': True,
-        },
-        'add_ie': ['Youtube'],
-    }, {
-        'url': 'http://generation-what.francetv.fr/europe/video/present-arms',
-        'only_matching': True,
-    }]
-
-    def _real_extract(self, url):
-        display_id = self._match_id(url)
-
-        webpage = self._download_webpage(url, display_id)
-
-        youtube_id = self._search_regex(
-            r"window\.videoURL\s*=\s*'([0-9A-Za-z_-]{11})';",
-            webpage, 'youtube id')
-
-        return self.url_result(youtube_id, ie='Youtube', video_id=youtube_id)
-
-
-class CultureboxIE(FranceTVBaseInfoExtractor):
-    _VALID_URL = r'https?://(?:m\.)?culturebox\.francetvinfo\.fr/(?:[^/]+/)*(?P<id>[^/?#&]+)'
-
-    _TESTS = [{
-        'url': 'https://culturebox.francetvinfo.fr/opera-classique/musique-classique/c-est-baroque/concerts/cantates-bwv-4-106-et-131-de-bach-par-raphael-pichon-57-268689',
-        'info_dict': {
-            'id': 'EV_134885',
-            'ext': 'mp4',
-            'title': 'Cantates BWV 4, 106 et 131 de Bach par Raphaël Pichon 5/7',
-            'description': 'md5:19c44af004b88219f4daa50fa9a351d4',
-            'upload_date': '20180206',
-            'timestamp': 1517945220,
-            'duration': 5981,
-        },
-        'params': {
-            'skip_download': True,
-        },
-        'add_ie': [FranceTVIE.ie_key()],
-    }]
-
-    def _real_extract(self, url):
-        display_id = self._match_id(url)
-
-        webpage = self._download_webpage(url, display_id)
-
-        if ">Ce live n'est plus disponible en replay<" in webpage:
-            raise ExtractorError(
-                'Video %s is not available' % display_id, expected=True)
-
-        video_id, catalogue = self._search_regex(
-            r'["\'>]https?://videos\.francetv\.fr/video/([^@]+@.+?)["\'<]',
-            webpage, 'video id').split('@')
-
-        return self._make_url_result(video_id, catalogue)
-
-
-class FranceTVJeunesseIE(FranceTVBaseInfoExtractor):
-    _VALID_URL = r'(?P<url>https?://(?:www\.)?(?:zouzous|ludo)\.fr/heros/(?P<id>[^/?#&]+))'
-
-    _TESTS = [{
-        'url': 'https://www.zouzous.fr/heros/simon',
-        'info_dict': {
-            'id': 'simon',
-        },
-        'playlist_count': 9,
-    }, {
-        'url': 'https://www.ludo.fr/heros/ninjago',
-        'info_dict': {
-            'id': 'ninjago',
-        },
-        'playlist_count': 10,
-    }, {
-        'url': 'https://www.zouzous.fr/heros/simon?abc',
-        'only_matching': True,
-    }]
-
-    def _real_extract(self, url):
-        mobj = re.match(self._VALID_URL, url)
-        playlist_id = mobj.group('id')
-
-        playlist = self._download_json(
-            '%s/%s' % (mobj.group('url'), 'playlist'), playlist_id)
-
-        if not playlist.get('count'):
-            raise ExtractorError(
-                '%s is not available' % playlist_id, expected=True)
-
-        entries = []
-        for item in playlist['items']:
-            identity = item.get('identity')
-            if identity and isinstance(identity, compat_str):
-                entries.append(self._make_url_result(identity))
-
-        return self.playlist_result(entries, playlist_id)
--- a/youtube_dl/extractor/francetv.py
+++ b/youtube_dl/extractor/francetv.py
@@ -6,7 +6,6 @@ import re
 
 from .common import InfoExtractor
 from ..compat import (
-    compat_str,
     compat_urlparse,
 )
 from ..utils import (
@@ -14,10 +13,8 @@ from ..utils import (
     determine_ext,
     ExtractorError,
     int_or_none,
-    parse_duration,
     try_get,
     url_or_none,
-    urljoin,
 )
 from .dailymotion import DailymotionIE
 from ..downloader import PROTOCOL_MAP
@@ -212,7 +209,7 @@ class FranceTVIE(InfoExtractor):
         for lang, sts in info['subtitles'].items():
             for st in sts:
                 st['downloader'] = lambda ydl, filename: PROTOCOL_MAP['m3u8_native'](ydl, ydl.params).download(filename, st)
-        
+
         return {
             'id': video_id,
             'title': self._live_title(info['title']) if is_live else info['title'],
--- a/youtube_dl/extractor/francetv.py
+++ b/youtube_dl/extractor/francetv.py
@@ -20,6 +20,7 @@ from ..utils import (
     urljoin,
 )
 from .dailymotion import DailymotionIE
+from ..downloader import PROTOCOL_MAP
 
 
 class FranceTVBaseInfoExtractor(InfoExtractor):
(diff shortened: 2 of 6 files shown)


youtube_dl/YoutubeDL.py
```diff
@@ -1879,6 +1879,8 @@ def ensure_dir_exists(path):
                         except (OSError, IOError):
                             self.report_error('Cannot write subtitles file ' + sub_filename)
                             return
+                    elif sub_info.get('downloader') is not None:
```
dirkf, line 1882: Would `callable(sub_info.get('downloader'))` be a safer test?

dirkf:
You might also consider [this](https://github.com/ytdl-org/youtube-dl/issues/29969#issuecomment-921299503), if your other changes haven't populated the properties mentioned.

lcheylus:
Hi @sarnoud, 

I reviewed and tested your code with fix for FranceTV extractor : 
- get GIT repository from your `sarnourd-francetv` branch
- install/test it in a virtual env (Python 3.9 on Debian testing).

Everything is OK : 
- extract and video formats (HLS and Dash)
- get subtitles
- tests with `test/test_download.py` for FranceTV, FranceTVSite and FranceTVInfo extractors.

But there is a bug with the infos returned as JSON => not serializable

```
$ youtube-dl -j https://www.francetvinfo.fr/replay-jt/france-3/soir-3/jt-grand-soir-3-lundi-26-aout-2019_3569073.html
WARNING: [FranceTV] Unknown MIME type application/mp4 in DASH manifest
  File "/home/fox/dev/youtube-dl.git/env/bin/youtube-dl", line 11, in <module>
    load_entry_point('youtube-dl==2021.6.6', 'console_scripts', 'youtube-dl')()
  File "/home/fox/dev/youtube-dl.git/env/lib/python3.9/site-packages/youtube_dl-2021.6.6-py3.9.egg/youtube_dl/__init__.py", line 475, in main
    _real_main(argv)
  File "/home/fox/dev/youtube-dl.git/env/lib/python3.9/site-packages/youtube_dl-2021.6.6-py3.9.egg/youtube_dl/__init__.py", line 465, in _real_main
  File "/home/fox/dev/youtube-dl.git/env/lib/python3.9/site-packages/youtube_dl-2021.6.6-py3.9.egg/youtube_dl/YoutubeDL.py", line 2070, in download
    res = self.extract_info(
  File "/home/fox/dev/youtube-dl.git/env/lib/python3.9/site-packages/youtube_dl-2021.6.6-py3.9.egg/youtube_dl/YoutubeDL.py", line 808, in extract_info
    return self.__extract_info(url, ie, download, extra_info, process)
(...)
  File "/home/fox/dev/youtube-dl.git/env/lib/python3.9/site-packages/youtube_dl-2021.6.6-py3.9.egg/youtube_dl/YoutubeDL.py", line 1774, in __forced_printings
    self.to_stdout(json.dumps(info_dict))
  File "/usr/lib/python3.9/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type function is not JSON serializable
```
This issue breaks Youtube-DL hook with mpv (Mplayer).

dirkf:
Presumably `YoutubeDL.py` l.1774 needs a `default` parameter to `json.dumps()` like the one added to `json.dump()` at `utils.py` l.1833?
```
            # json.dumps may need a default=lambda x: ...
            self.to_stdout(json.dumps(info_dict))
```
Also at l.2077:
```
                    self.to_stdout(json.dumps(res))
```

youtube_dl/YoutubeDL.py
sarnoud: Good one. Done.

sarnoud:
Good catches! Submitted changes.

lcheylus:
After these both commits, JSON output is correct.

But after some analysis, I don't understand your code to get subtitle/downloader : `youtube_dl/extractor/francetv.py` lines 209-211 : 

```
for lang, sts in info['subtitles'].items():
    for st in sts:
        st['downloader'] = lambda ydl, filename: PROTOCOL_MAP['m3u8_native'](ydl, ydl.params).download(filename, st)
```
With this lambda, `st['downloader']` is a reference to a Python function => not JSON serializable.

sarnoud:
Yes you are correct.
Check the change in YoutubeDL.py:1882.

There were 2 ways to get captions: via an url ('url' attrribute) or via providing the data directly ('data' attribute).

The issue with captions now is that there are delivered as a m3u8 stream that needs to be downloaded. So there would be 2 options: i- download captions all the time and put them in 'data', or ii- provide a lambda that will only do the download when needed. 
I implemented ii- and this is the 'downloader' attribute basically. 

                    elif callable(sub_info.get('downloader')):
                        sub_info.get('downloader')(self, encodeFilename(sub_filename))

The lambda is also the thing that was not getting properly serialized in the json.dump calls BTW.

Does that make sense?

dirkf:
Possibly the subtitle generator should be evaluated when generating JSON rather than just emitting "not serializable"?

sarnoud:
Technically possible, but the download itself is quite involved though. See for yourself with something like:
python3 -m youtube_dl --all-subs --skip-download https://www.france.tv/france-2/les-invisibles/les-invisibles-saison-1/2748331-pachelbel.html

I would advise against it for every JSON serialization?

renalid:
@sarnoud thanks a lot for the fix ! works well 👍 
Do you think there is a way to get (back) the description of the video ?

dirkf:
> ...
> There were 2 ways to get captions: via an url ('url' attrribute) or via providing the data directly ('data' attribute).
> 
> The issue with captions now is that they are delivered as a m3u8 stream that needs to be downloaded. So there would be 2 options: i- download captions all the time and put them in 'data', or ii- provide a lambda that will only do the download when needed.
> I implemented ii- and this is the 'downloader' attribute basically.
> ...

As this is adding a new element to the extractor-core API, I think it needs a bit more thought.

Presumably the reason for not downloading the captions in the extractor is to avoid unnecessarily downloading them without second-guessing the core logic (`YoutubeDL.py`) that makes that decision?

So instead of returning the actual subtitles, we return a downloader function, essentially a closure, and on first glance this seems like a cleaner solution than the `['url']` mechanism: as currently implemented, this has the core logic find some random extractor instance to use for downloading the subtitles.

However I believe that these are really two alternatives of the same sort. One is a URL that has to be downloaded as a plain text page; the new case is a URL that has to be downloaded as M3U8. Surely we should deal with both cases in the same logic? And, just as the core shouldn't be playing with random IEs, I'm not sure that an extractor should know about the downloader's `PROTOCOL_MAP`, as in the downloader lambda

Here's my proposal that addresses both concerns and (if it works) has some other benefits:
* scrap the `['downloader']` mechanism (sorry) -- benefit - no need to worry about unserializable items for JSON;
* from the extractor, pass the m3u8 URL as the `['url']` and also set `['protocol']` as `'m3u8_native'`: -- benefit - the subtitle link is available in JSON outputs;
* in the core method `_process_info()` where it handles subtitles, remove this
```
            ie = self.get_info_extractor(info_dict['extractor_key'])
```
* in the conditional branch where `sub_info['data'] is None`, replace the `try` statement with something like this:
```
                        # essentially the lambda downloader here
                        # pass 1st self.params to respect --external_downloader, --hls-prefer-native?
                        fd = get_suitable_downloader(sub_info, self.params)(self, self.params)
                        try:
                            if self.params.get('verbose'):
                                self.to_screen('[debug] Invoking subtitle downloader on %r' % sub_info.get('url'))
                            # the FD is supposed to encodeFilename() ?
                            if not fd.download(sub_filename, sub_info):
                                # depending on the FD, it may catch errors and return False, or not
                                raise DownloadError('subtitle download failed')
                        except (compat_urllib_error.URLError, compat_http_client.HTTPException, socket.error, OSError, IOError, YoutubeDLError) as err:
                            self.report_warning('Unable to download subtitle for "%s": %s' %
                                                (sub_lang, error_to_compat_str(err)))
                            continue
```
Provided that there aren't significant unwanted side-effects in using the `FileDownloader` (but it's basically working like the lambda was), I think that this would be a better structure.

chzurb:
@dirkf I understand your motivation to encourage the `download` aspect of PR changes from @sarnoud to be reengineered back into `['url']` but one benefit of leaving them as he has implemented them at least initially is that it doesn't break backward compatibility or introduce any potential regression with the existing use of ['url'] by every other IE.  Perhaps it's better to let it come in the way he has it now as separate `['download`] which allows it to get some real world production use in isolation from francetv IE users.  The  integration back into  `url ` might then be worked in parallel for a later PR in and of itself like you suggest.     Just my 2 cents.

dirkf:
Sure, understandable point. 

There's a halfway house: instead of replacing the existing `try` statement, insert my proposed replacement before with a test `elif sub_info['protocol']:`. Then, as you suggest, a subsequent PR could combine the last two conditional branches, based on the valuable debugging of FranceTV users.

sarnoud:
I implemented @dirkf proposal - which I agree makes sense.

@chzurb you're right that there is a risk of regression as it touches the core logic - even if it shouldn't? Tried downloading subtitles from a couple sites and it seems to work - but definitely not exhaustive.

dirkf:
Unfortunately it's not easy to scan the extractors for use of `['url']` in subtitles as the same key is heavily used throughout (131 of 787 mention the string). Maybe the old FranceTV was the only one, or one of a few?

sarnoud:
Yep. 
So are we willing to go with it?

dirkf:
It's good for me (if it works with FranceTV). If real maintainers reappear they might prefer the 2-stage approach but it would be easy to switch to that.

sarnoud:
Thanks @dirkf . What would be the next step? You would approve so that it can be merged?

dirkf:
If only. Please read #29965. Of course nothing stops anyone from merging anything into their own fork.

renalid:
@sarnoud Maybe could you propose this same PR to yt-dlp project ?

dirkf:
Indeed: see https://github.com/yt-dlp/yt-dlp/discussions/803.

dirkf:
The use of the `'protocol'` key for subtitles should be documented in `extractor/common.py l.247 ff., maybe like this:
```
     subtitles:      The available subtitles as a dictionary in the format
                     {tag: subformats}. "tag" is usually a language code, and
                     "subformats" is a list sorted from lower to higher
                     preference, each element is a dictionary with the "ext"
                     entry and one of:
                         * "data": The subtitles file contents
-                        * "url": A URL pointing to the subtitles file
+                        * "url": A URL pointing to the subtitles resource
+                   With "url", a "protocol" entry (as for "formats" above)
+                   may be provided to indicate how the URL should be
+                   processed; by default it is a file downloaded by HTTP(S)
                     "ext" will be calculated from URL if missing
```

ajt-en-france:
> I implemented @dirkf proposal - which I agree makes sense.
> 
> @chzurb you're right that there is a risk of regression as it touches the core logic - even if it shouldn't? Tried downloading subtitles from a couple sites and it seems to work - but definitely not exhaustive.

I found two issues. It's possible that I've not correctly built the binary from git, but:

1. It pulls the first audio file, which may be the audio description, not the regular audio track
2. The subtitles it pulls are malformed and I can't use them

For example:

dirkf:
>... 
> I found two issues. It's possible that I've not correctly built the binary from git, but:
> 
>     1. It pulls the first audio file, which may be the audio description, not the regular audio track

Does FranceTV have a separate programme page for the AD version (as with BBC)? If so, the AD should only be used with that page's URL. Otherwise the AD should only be fetched if selected with `-f`.

>     2. The subtitles it pulls are malformed and I can't use them

How does that manifest itself? Although I'm no VTT expert, the text you quote looks like what is specified in [RFC8216 section 3.5][(https://datatracker.ietf.org/doc/html/rfc8216#section-3.5) and its sources.

lcheylus:
> I found two issues. It's possible that I've not correctly built the binary from git, but:
> 
>     1. It pulls the first audio file, which may be the audio description, not the regular audio track

For this issue, we must set `format['preference']` before sorting `formats` : `extractor/francetv.py` line 202
```
for f in info['formats']:    
    if f['format_id'].startswith('dash-audio_qtz') or f['format_id'].find('Audio Description') > 0:    
        f['preference'] = -1    
        f['format_note'] = "Audio description"
    elif f['format_id'].startswith('dash-audio'):    
        f['preference'] = 10    
    elif f['format_id'].startswith('hls-audio'):    
        f['preference'] = 20    
    else:    
        f['preference'] = 100    
    
self._sort_formats(info['formats'])
```

With this modification, HLS-Audio with no audio description is always chosen for "Best Audio" format. I had also a format note for audio description to display it with `-F` (list formats) flag.

ajt-en-france:
> > ...
> > I found two issues. It's possible that I've not correctly built the binary from git, but:
> > ```
> > 1. It pulls the first audio file, which may be the audio description, not the regular audio track
> > ```
> 
> Does FranceTV have a separate programme page for the AD version (as with BBC)? If so, the AD should only be used with that page's URL. Otherwise the AD should only be fetched if selected with `-f`.

There could be several audio tracks for the same program on the same web pages, French, French with audio description and sometimes the original version "Version Originale" which could be English, Italian etc... On the France.tv web page you can make a selection, but if you use the browser based player it defaults to the French audio track even if the original program was in Italian for example. The audio described is never the default as far as I can tell.

When previously using the -F and then -f flags to pick streams I found it a bit unreliable, as you could often get French no matter which steam you picked with youtube-dl, even though the web player would pay the Italian for example...

boulderob:
> There could be several audio tracks for the same programme on the same web pages, French, French with audio description and sometimes the original version "_Version Originale_" which could be English, Italian etc... On the France.tv web page you can make a selection, but if you use the browser based player it defaults to the French audio track even if the original program was in Italian for example. The audio described is never the default as far as I can tell.
> 
> When previously using the **-F** and then **-f** flags to pick streams I found it a bit unreliable, as you could often get French no matter which steam you picked with youtube-dl, even though the web player would pay the Italian for example... _Montalbano sounds really strange in French..._

I can confirm in the *past* that the VO audio tracks were often NOT present in the -F listings or if it seemed like they were that when you downloaded and played them they were all french audio even if they said otherwise.  Yet the VO audio tracks were playable from the site which means they were available.  This just means that the files were not being picked up correctly by the francetv IE in the past.   

My thought is that the new PR introduced here will pick up these VO files in the new mpd formats that weren't being picked up by the old francetv IE.  So far with my limited testing of the new PR, `qtz` either picks up the AD or the VO depending on the video source.   That's a good sign.  

The bad part is that the current `qtz` format IDs don't always include the text "VO" or "Audio Description" in the format ids.  So you have to guess when downloading with `-f` after viewing the specs with `-F` what it's going to be.    Usually this is pretty easy because VO sources are usually not AD and vice versa based on my past usage on the site anyway.

The only way to solve this is to add "descriptive" text to the AD and VO format ids.  But that's only possible if that type of text is available in the source manifests.

boulderob:
> The subtitles look like this:
> 
> ```
> WEBVTT
> X-TIMESTAMP-MAP=MPEGTS:900000,LOCAL:00:00:00.000
> 
> 00:00:00.000 --> 00:00:33.033
> ...
> 
> WEBVTT
> X-TIMESTAMP-MAP=MPEGTS:1800000,LOCAL:00:00:00.000
> 
> WEBVTT
> X-TIMESTAMP-MAP=MPEGTS:2700000,LOCAL:00:00:00.000
> 
> WEBVTT
> X-TIMESTAMP-MAP=MPEGTS:3600000,LOCAL:00:00:00.000
> 
> 00:00:03.300 --> 00:00:05.733
> -Voilà.
> -Allez, les filles !
> 
> 00:00:06.166 --> 00:00:07.500
> -Hello !
> 
> 00:00:07.766 --> 00:00:10.366
> -Et voilà les plus belles !
> ```

I followed this issue and commit a bit on the `yt-dlp` site a few months ago to fix the broken subtitles but I haven't tested it yet.  If you can download subtitles but they are out of synch it really is still broken because they are unusable.  I would think the order of the subtitle segments is already fixed in the manifest so I'm not sure how they get out of synch during a download.

The ability to downloading subtitles isn't really a 'fix' if they're out of order and don't work.  I'm going to clone sarnaud's PR and see for myself where things stand in this regard.   I presume the old HLS subtitles work fine though since they were working before

dirkf:
> The ability to downloading subtitles isn't really a 'fix' if they're out of order and don't work. I'm going to clone sarnaud's PR and see for myself where things stand in this regard. I presume the old HLS subtitles work fine though since they were working before

Apparently yt-dlp has a [whole module for reassembling WebVTT from fragment downloads](https://github.com/yt-dlp/yt-dlp/commit/4a2f19abbd61274358211c2e3b1d9658cfbdcdde), which might address the several ToDo items in the code related to WebVTT.

pukkandan:
The subtitle downloading and reassembling code was submitted to youtube-dl first in https://github.com/ytdl-org/youtube-dl/pull/6144. Ofc, further improvements have been made to it in https://github.com/yt-dlp/yt-dlp/pull/247. Hope either of these PRs are helpful to you guys

sarnoud:
> +                        * "url": A URL pointing to the subtitles resource
> +                   With "url", a "protocol" entry (as for "formats" above)
> +                   may be provided to indicate how the URL should be
> +                   processed; by default it is a file downloaded by HTTP(S)

Done

boulderob:
> > The ability to downloading subtitles isn't really a 'fix' if they're out of order and don't work. I'm going to clone sarnaud's PR and see for myself where things stand in this regard. I presume the old HLS subtitles work fine though since they were working before
> 
> Apparently yt-dlp has a [whole module for reassembling WebVTT from fragment downloads](https://github.com/yt-dlp/yt-dlp/commit/4a2f19abbd61274358211c2e3b1d9658cfbdcdde), which might address the several ToDo items in the code related to WebVTT.

> The subtitle downloading and reassembling code was submitted to youtube-dl first in #6144. Ofc, further improvements have been made to it in yt-dlp/yt-dlp#247. Hope either of these PRs and the discussions in them are helpful to you guys

Thanks to both of you for that!  I just cloned sarnaud's PR repo but my guess is it would make more sense to clone yt-dlp and then try to merge sarnaud's PR #29996 into it to pick up the vtt ordering fixes vs the other way around due to the number of files changed in yt-dlp #247.   @pukkandan are you going to merge in sarnaud's #29996 to yt-dlp or wait on it?  

My merge skills are a little rusty.  I presume I just need to open up remote to sarnaud fork, fetch his PR branch, merge to yt-dlp master (or my own branch of it) and fix conflicts if they arise.  Sound right?

Thanks

youtube_dl/extractor/francetv.py
```diff
@@ -90,17 +91,47 @@ def _extract_video(self, video_id, catalogue=None):
...
+
+        info = {
+            'title': None,
+            'subtitle': None,
+            'image': None,
+            'subtitles': {},
+            'duration': None,
+            'videos': [],
+            'formats': [],
+        }
```
fstirlitz, line 100: It seems rather silly to stuff extracted information into an `info` dictionary only to extract each key back one by one at the end. The churn of changing `formats` into `info['formats']` later in the extractor inflates the diff size and makes code harder to review without actually accomplishing anything.

youtube_dl/extractor/francetv.py
```diff
@@ -167,56 +173,67 @@ def sign(manifest_url, manifest_id):
...
-            subtitles['fr'] = subtitles_list
+        if info['subtitle']:
+            info['title'] += ' - %s' % info['subtitle']
+        info['title'] = info['title'].strip()
 
         return {
             'id': video_id,
-            'title': self._live_title(title) if is_live else title,
+            'title': self._live_title(info['title']) if is_live else info['title'],
             'description': clean_html(info.get('synopsis')),
```
fstirlitz, line 230: This is no longer extracted. The `info` dictionary issue is masking a bug here.

youtube_dl/extractor/francetv.py
```diff
@@ -167,56 +173,67 @@ def sign(manifest_url, manifest_id):
...
-            'thumbnail': urljoin('https://sivideo.webservices.francetelevisions.fr', info.get('image')),
-            'duration': int_or_none(info.get('real_duration')) or parse_duration(info.get('duree')),
+            'thumbnail': info.get('image'),
+            'duration': int_or_none(info.get('duration')),
             'timestamp': int_or_none(try_get(info, lambda x: x['diffusion']['timestamp'])),
```
fstirlitz, line 233: Likewise.

youtube_dl/extractor/francetv.py
```diff
@@ -167,56 +173,67 @@ def sign(manifest_url, manifest_id):
...
-                    formats.append({
+                    info['formats'].append({
                         'url': video_url,
                         'format_id': format_id,
                     })
 
-        self._sort_formats(formats)
+        for f in info['formats']:
+            preference = 100
+            if f['format_id'].startswith('dash-audio_qtz=96000') or (f['format_id'].find('Description') >= 0):
```
fstirlitz, line 211:
```suggestion
            if f.get('language') == 'qtz':
```

should work as well, does it not?

Trit34:
@sarnoud We could have a fix for DRM videos here:
https://github.com/ytdl-org/youtube-dl/issues/29956#issuecomment-931559459

Trit34:
Argh: new problem…

```
$ youtube-dl -F https://www.france.tv/series-et-fictions/telefilms/2787667-meurtres-a-mulhouse.html --verbose
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-F', 'https://www.france.tv/series-et-fictions/telefilms/2787667-meurtres-a-mulhouse.html', '--verbose']
[debug] youtube-dl version 2021.06.06
[debug] Python version 3.9.7 (CPython) - Linux-5.14.8-arch1-1-x86_64-with-glibc2.33
[debug] exe versions: ffmpeg 4.4, ffprobe 4.4, rtmpdump 2.4
[debug] Proxy map: {}
[FranceTVSite] 2787667-meurtres-a-mulhouse: Downloading webpage
[FranceTV] 3b8da0af-a712-411c-bb42-b8f65c55d5d7: Downloading desktop video JSON
WARNING: Unable to download JSON metadata: HTTP Error 422: Unprocessable Entity
[FranceTV] 3b8da0af-a712-411c-bb42-b8f65c55d5d7: Downloading mobile video JSON
WARNING: Unable to download JSON metadata: HTTP Error 422: Unprocessable Entity
ERROR: 3b8da0af-a712-411c-bb42-b8f65c55d5d7: No video source has been found
  File "/usr/lib/python3.9/site-packages/youtube_dl/YoutubeDL.py", line 816, in wrapper
    return func(self, *args, **kwargs)
  File "/usr/lib/python3.9/site-packages/youtube_dl/YoutubeDL.py", line 837, in __extract_info
    ie_result = ie.extract(url)
  File "/usr/lib/python3.9/site-packages/youtube_dl/extractor/common.py", line 537, in extract
  File "/usr/lib/python3.9/site-packages/youtube_dl/extractor/francetv.py", line 251, in _real_extract
    return self._extract_video(video_id, catalog)
  File "/usr/lib/python3.9/site-packages/youtube_dl/extractor/francetv.py", line 127, in _extract_video
    raise ExtractorError(
youtube_dl.utils.ExtractorError: 3b8da0af-a712-411c-bb42-b8f65c55d5d7: No video source has been found
```

Trit34:
@sarnoud I know that youtube-dl has no vocation to bypass DRMs, but it did it before with France TV videos, so… Is there a way to code such a command in the francetv.py extractor?

`ffmpeg -i "https://replayftv-vh.akamaihd.net/i/streaming-adaptatif_media-secure_france-dom-tom/2021/S40/J3/1045549522-615df4c0e3747-,standard1,standard2,standard3,standard4,qaa,.mp4.csmil/master.m3u8?hdnea=exp=1633877480~acl=%2fi%2fstreaming-adaptatif_media-secure_france-dom-tom%2f2021%2fS40%2fJ3%2f1045549522-615df4c0e3747-,standard1,standard2,standard3,standard4,qaa,.mp4.csmil*~hmac=326b54f56efb3fcb82c8c94862f25cac02cc295ed78901c87e90f41cee7fee8a" -c copy -bsf:a aac_adtstoasc "Murdoch.mp4"`

(Based on the `https://replayftv-vh.akamaihd.net/i/streaming-adaptatif_media-secure_france-dom-tom/YEAR/SWEEK/JDAY/PLURIMEDIA_ID-MANIFEST_ID_REVERSED-,standard1,standard2,standard3,standard4,qaa,.mp4.csmil/master.m3u8` pattern given by TiA4f8R: https://github.com/ytdl-org/youtube-dl/issues/29956#issuecomment-931559459)

Trit34:
@sarnoud If you’re still around there, since I have updated Python to 3.10, your patch does not work anymore when I replace the original files by your patched ones:

```
$ youtube-dl -F https://www.france.tv/documentaires/art-culture/2941931-rochefort-noiret-marielle-les-copains-d-abord.html
  File "/usr/bin/youtube-dl", line 33, in <module>
    sys.exit(load_entry_point('youtube-dl==2021.6.6', 'console_scripts', 'youtube-dl')())
  File "/usr/bin/youtube-dl", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.10/importlib/metadata/__init__.py", line 162, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/usr/lib/python3.10/site-packages/youtube_dl/__init__.py", line 3, in <module>
    from .common import FileDownloader
ModuleNotFoundError: No module named 'youtube_dl.common'
```

PR #30021, opened 2021-09-25
Add new extractor(bh,pt) and update author name in sb extractor

helopy:

---

- [x] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

- Added new extractors for baddiehub and porntrex
- Fixed author name extraction in spankbang and add updated tests

### DIFF
--- a/youtube_dl/extractor/baddiehub.py
+++ b/youtube_dl/extractor/baddiehub.py
@@ -1,0 +1,66 @@
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+
+
+class BaddieHubIE(InfoExtractor):
+    _VALID_URL = r"https?://(?:www\.)?baddiehub\.com/\?p=(?P<id>[\w\d]+)"
+    _TESTS = [
+        {
+            "url": "https://baddiehub.com/?p=11436",
+            "only_matching": True,
+        },
+        {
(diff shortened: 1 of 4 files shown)


youtube_dl/extractor/baddiehub.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group("id") or mobj.group("name")
+
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._html_search_regex(r"<title>(.+?)</title>", webpage, "title")
+        content_url = self._html_search_regex(
+            r"<meta itemprop=\"contentURL\".*content=\"(?P<contentURL>.*?)\"\s\/>",
```
rautamiekka, line 35: Instead of escaping the inner double quotes you could single-quote the string.

dirkf: Or better, use the utility method that gets the value of the content attribute of a `<meta>` tag with any of these attributes having the specified value (or a value from an iterable): itemprop, name, property, id, http-equiv:
```
        content_url = self._html_search_meta('contentURL', webpage)
```
And also below for author.

helopy:
@rautamiekka please review again

PR #30033, opened 2021-09-29
[common] Test for valid URL with HEAD

dirkf:
---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

In the IE `_is_valid_url()` method, the URL is tested for validity. The resource linked may be a large media file which could be downloaded and then discarded. An early version of the method used a `HEAD` request, as would be expected. This was then changed to `GET` in 4069766c527d10b8e25b9262a3882101367deb3e, but there is no supporting rationale and no issues mentioning that `HEAD` requests were being rejected.

If that should be a problem with a specific site, we can set a class var, say `_test_valid_with GET = False` and then make the changed lines as follows, with the var set to `True` for the problem site:
```
            req = url if self._test_valid_with GET else HEADRequest(url)
            self._request_webpage(req, video_id, 'Checking %s URL' % item, headers=headers)
```
Comments welcome.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -53,6 +53,7 @@ from ..utils import (
     float_or_none,
     GeoRestrictedError,
     GeoUtils,
+    HEADRequest,
     int_or_none,
     js_to_json,
     JSON_LD_RE,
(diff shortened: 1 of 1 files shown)


Lee-Carre:
From dirkf/youtube-dl#4

> While generally sensible, I've encountered some sites in the past which didn't follow the HTTP spec, which states (paraphrased) that the response to a `HEAD` request should be the same as for a `GET` except that the body is omitted.
> 
> In some cases it wasn't just mismatched headers, but a very different response code. One case returned (if my memory is correct from so long ago) `405 Method Not Implemented` 🙄🤦‍♂️.
> 
> So, exceptions may well need to be handled.

dirkf:
OK, that's what the `expected_status` parameter of `_request_webpage()` is for.

Issue #30043, opened 2021-10-02
Cbc gems not working (used to)


crazytrain411:
.\youtube-dl.exe --verbose -f best https://gem.cbc.ca/media/dino-ranch/s01e18
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['--verbose', '-f', 'best', 'https://gem.cbc.ca/media/dino-ranch/s01e18']
[debug] Encodings: locale cp1252, fs mbcs, out cp850, pref cp1252
[debug] youtube-dl version 2021.06.06
[debug] Python version 3.4.4 (CPython) - Windows-10-10.0.19041
[debug] exe versions: ffmpeg 4.3.1-2021-01-01-full_build-www.gyan.dev, ffprobe 4.3.1-2021-01-01-full_build-www.gyan.dev
[debug] Proxy map: {}
[debug] Using fake IP 99.241.7.109 (CA) as X-Forwarded-For.
[cbc.ca:watch] d: Downloading XML
[cbc.ca:watch] d: Downloading XML
ERROR: cbc.ca:watch said: We are sorry, but the selected video is currently unavailable.; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpkqxnwl31\build\youtube_dl\YoutubeDL.py", line 815, in wrapper
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpkqxnwl31\build\youtube_dl\YoutubeDL.py", line 836, in __extract_info
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpkqxnwl31\build\youtube_dl\extractor\common.py", line 534, in extract
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpkqxnwl31\build\youtube_dl\extractor\cbc.py", line 436, in _real_extract
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpkqxnwl31\build\youtube_dl\extractor\cbc.py", line 261, in _call_api
[...]

PR #30046, opened 2021-10-03
[CBC] Fix extraction for Gem pages in line with yt-dlp#936

dirkf:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

According to #30043, CBC Gems pages were not being correctly extracted following a site change. The issue was also tracked in https://github.com/yt-dlp/yt-dlp/issues/936, resulting in a PR https://github.com/yt-dlp/yt-dlp/pull/1013. 

The PR author wrote
>This PR fixes https://github.com/yt-dlp/yt-dlp/issues/936. gem.cbc.ca videos, playlists, and livestreams can now be downloaded again. Content that requires login (which is free) can also be downloaded.
>
>I removed CBCOlympicsIE because the domain it was based on doesn't work anymore. I also removed some dead tests from the other classes.

This PR adapts the author's PR with a few changes for the conventions and platform requirements of this project.


youtube_dl/extractor/common.py
```diff
@@ -404,21 +404,27 @@ def __init__(self, downloader=None):
...
         if '_VALID_URL_RE' not in cls.__dict__:
             cls._VALID_URL_RE = re.compile(cls._VALID_URL)
-        return cls._VALID_URL_RE.match(url) is not None
+        return cls._VALID_URL_RE.match(url)
+
+    @classmethod
+    def suitable(cls, url):
+        """Receives a URL and returns True if suitable for this IE."""
+
+        return cls._match_valid_url(url) is not None
```
pukkandan, line 421: idk what the maintainers think about this, but I personally think this change is out of the scope of this PR. If this function is desired, it can be added separately. For now, you could just replace `self._match_valid_url(url)` with `re.match(self._VALID_URL, url)` as many other extractors already do.

dirkf: And I've done it myself. I thought this was a useful back-port although I agree that it might be controversial in the scope of the PR. I don't see any obvious downside to the change, although it would be a different case if yt-dlp was intending to revert this code (or wished it could).

One could just introduce a `CBCGemBaseIE` with the `_match_valid_url()` class method, as you've defined, so minimising code divergence from the yt-dlp PR, but then other extractors don't get the benefit.

Of course this is moot as long as no-one is accepting PRs (apparently there's almost a month of PRs to review at an average 15 minutes per PR in 8-hour days of nothing else but the odd coffee break), but I'll be merging it into my release.

PR #30048, opened 2021-10-03
[YouMaker] Add new extractor

flashdagger:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

New extractor for youmaker.com with hls support

### DIFF
--- a/youtube_dl/extractor/youmaker.py
+++ b/youtube_dl/extractor/youmaker.py
@@ -1,0 +1,100 @@
+import calendar
+import re
+from datetime import datetime
+
+from .common import InfoExtractor
+from .. import utils
+
+
+class YouMakerIE(InfoExtractor):
+    _VALID_URL = r"""(?x)
+                    https?://(?:www\.)?youmaker\.com/
+                    (?:v|video|embed)/
+                    (?P<id>[0-9a-zA-Z-]+)
+                    """
+
+    _TESTS = [
+        {
+            "url": "https://www.youmaker.com/v/Dnnrq0lw8062",
+            "info_dict": {
+                "id": "77c92592-57fa-4bfc-a5da-79e965667001",
+                "ext": "mp4",
+                "title": "Althistoriker Dr. David Engels im Interview: „Das ist der echte Untergang des Abendlandes“",
+                "description": "Im Interview mit Epoch Times führt der belgische Althistoriker Dr. David Engels aus, "
+                "wie sich der Zerfall der europäischen Staatengemeinschaft im Zuge der Corona-Krise "
+                "zugespitzt hat, und zeichnet Parallelen zu den letzten Atemzügen der spätrömischen "
+                "Republik. , , Der Artikel zu dem Interview folgt in Kürze "
+                "hier: https://www.epochtimes.de/wissen/gesellschaft/"
+                "das-ist-der-echte-untergang-des-abendlandes-a3613553.html - Youmaker",
+                "duration": 3507,
+                "upload_date": "20211001",
+                "timestamp": 1633046400,
+            },
+            "params": {
+                "skip_download": True,
+                "nocheckcertificate": True,
+            },
+        },
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        webpage = self._download_webpage(
+            "https://youmaker.com/video/%s" % video_id, video_id
+        )
+
+        info = self._parse_json(
+            self._search_regex(
+                (
+                    r'<script\s+type="application/ld\+json">s*'
+                    r"(?P<json>{[^}]+})"
+                    r"\s*</script>"
+                ),
+                webpage,
+                "json_info",
+                default="{}",
+                group="json",
+            ),
+            video_id,
+        )
+
+        thumbnail_urls = info.get("thumbnailUrl", [])
+        if not thumbnail_urls:
+            raise utils.ExtractorError("Resource not available", expected=False)
+
+        asset_base = thumbnail_urls[0].rsplit("/", maxsplit=1)[0]
+        match = re.match(r".*/assets/(\d\d\d\d)/(\d\d\d\d)/([^/]+)$", asset_base)
+        if match:
+            upload_date = "".join(match.groups()[0:2])
+            dt = datetime.strptime(upload_date, "%Y%m%d")
+            timestamp = calendar.timegm(dt.timetuple())
+            video_id = match.groups()[2]
+        else:
+            upload_date = utils.unified_strdate(info["uploadDate"])
+            timestamp = utils.parse_iso8601(info["uploadDate"])
+
+        formats = self._extract_m3u8_formats(
+            "/".join((asset_base, "playlist.m3u8")),
+            video_id,
+            ext="mp4",
+        )
+        parts_d = re.match(r"PT(?:(\d+)H)?(?:(\d+)M)?(\d+)S", info["duration"]).groups()
+        duration = sum((int(n) * l for n, l in zip(parts_d, (3600, 60, 1)) if n))
+        for item in formats:
+            item["format_id"] = "hls-%dp" % item["height"]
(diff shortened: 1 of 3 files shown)


youtube_dl/extractor/youmaker.py
```diff
@@ -0,0 +1,144 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from .. import utils
```
dirkf, line 5: Every existing extractor imports individual functions rather than the whole file.

youtube_dl/extractor/youmaker.py
```diff
@@ -0,0 +1,144 @@
...
+            self._sort_formats(formats)
+            for item in formats:
+                item["format_id"] = "hls-%dp" % item["height"]
+                if duration and item.get("tbr"):
+                    item["filesize_approx"] = 128 * item["tbr"] * duration
+
+        return {
+            "id": info["video_uid"],
+            "title": info["title"],
+            "description": info["description"],
```
dirkf, line 131: `info.get('description')` ? Not a mandatory item. Similarly with `info['uploaded_at']` below.

youtube_dl/extractor/youmaker.py
```diff
@@ -0,0 +1,144 @@
...
+            "%s/video/metadata/%s" % (self._api_url, video_id), video_id
+        )
+
+        status = info.get("status", "something went wrong")
+        if status != "ok":
+            raise utils.ExtractorError(status, expected=True)
+
+        info = info["data"]
+        info["tags"] = [
+            tag.strip() for tag in info.get("tag", "").strip("[]").split(",") if tag
```
dirkf, line 100: `()` is a better 0-length iterable than `""`, which implies text.

youtube_dl/extractor/youmaker.py
```diff
@@ -0,0 +1,144 @@
...
+            self.__protocol = "http"
+        info = self._download_json(
+            "%s/video/metadata/%s" % (self._api_url, video_id), video_id
+        )
+
+        status = info.get("status", "something went wrong")
+        if status != "ok":
+            raise utils.ExtractorError(status, expected=True)
+
+        info = info["data"]
```
dirkf, line 98: Now info['data'] needs to be a dict. As an `['id']` is mandatory (as is `['title']`), you could get it here and give up otherwise:
```
        display_id = video_id  # this can be included as the 'display_id' of the result 
        video_id, title = try_get(info, lambda x: (x['data']['id'], x['data']['title'], )
        title = str_or_none(title)
        if video_id is None or not title:
            raise ExtractorError('Unable to extract id/title')
```
Eventually `video_id` and `title` can be used in the result dict.

youtube_dl/extractor/youmaker.py
flashdagger: ok

youtube_dl/extractor/youmaker.py
dirkf: Yes, of course. Somehow I stopped reading the line too soon :-(

flashdagger: But this is a sign, that I should strife for readability rather than conciseness.
I pushed some more commits, adding categories and hopefully making the code more readable and robust.
Can you please check?

youtube_dl/extractor/youmaker.py
flashdagger: Ok. Please see new commits...

youtube_dl/extractor/youmaker.py
flashdagger: Ok. Now I check for certain mandatory keys, so I can safely use them in the code.

youtube_dl/extractor/youmaker.py
```diff
@@ -0,0 +1,196 @@
...
+    def _base_url(self):
+        return self._fix_url("//www.youmaker.com")
+
+    @property
+    def _asset_url(self):
+        # as this url might change in the future
+        # it needs to be extracted from some js magic...
+        return self._fix_url("//vs.youmaker.com/assets")
+
+    def _api(self, path, cache=False, **kwargs):
```
dirkf, line 71: Some 62 out of 64 other extractors that do a similar thing have called the corresponding method `_call_api()`. I'm only pointing this out in case you might want to do so.

youtube_dl/extractor/youmaker.py
```diff
@@ -0,0 +1,196 @@
...
+        key = hash((path, compat_urllib_parse_urlencode(kwargs.get("query", {}))))
+        if cache and key in self.__cache:
+            return self.__cache[key]
+
+        url = "%s/v1/api/%s" % (self._base_url, path)
+        info = self._download_json(url, self.__video_id, **kwargs)
+        status = info.get("status", "something went wrong")
+        data = info.get("data")
+        if status != "ok" or data is None:
+            raise ExtractorError(status, expected=True)
```
dirkf, line 87: Above, it's still possible that the JSON download works but doesn't result in a `dict`. So this would be better (setting `None` on error to help with the second point below):
```
        status = try_get(info, lambda x: x['status'])
```
Then, if the API makes a breaking change without us noticing, is that `expected` or not? As the site is unlikely to revert the change, it becomes our bug and so not `expected`. I suggest `expected` should correspond to the API returning an actual status that is not OK, and nothing else, like so:
```
             raise ExtractorError(status or 'something went wrong', expected=status not in ('ok', None))
```
But you are obviously familiar with the API and I'm not ...

youtube_dl/extractor/youmaker.py
```diff
@@ -0,0 +1,196 @@
...
+
+        if cache:
+            self.__cache[key] = data
+
+        return data
+
+    def _categories_by_id(self, cid):
+        category_map = self.__cache.get("category_map")
+        if category_map is None:
+            category_list = self._api(
```
dirkf, line 97: Should the extractor crash if the category download fails? It's not going to be used for a mandatory field.

youtube_dl/extractor/youmaker.py
dirkf: Actually I think the line I fingered was fine, but the replacement's fine too.

youtube_dl/extractor/youmaker.py
```diff
@@ -0,0 +1,196 @@
...
+        data = info.get("data")
+        if status != "ok" or data is None:
+            raise ExtractorError(status, expected=True)
+
+        if cache:
+            self.__cache[key] = data
+
+        return data
+
+    def _categories_by_id(self, cid):
```
dirkf, line 94: When this is called, `cid` might be `None`, so should the method return immediately in that case?

youtube_dl/extractor/youmaker.py
flashdagger: ok
consistency is king and tool-based refactoring is easy :)

youtube_dl/extractor/youmaker.py
flashdagger: status and data are minimal requirements. And don't worry, if expected errors happen repeatedly people will bother you anyway :)

youtube_dl/extractor/youmaker.py
```diff
@@ -0,0 +1,196 @@
...
+                if duration and item.get("tbr"):
+                    item["filesize_approx"] = 128 * item["tbr"] * duration
+
+        return {
+            "id": info["video_uid"],  # asserted before
+            "title": info["title"],  # asserted before
+            "description": info.get("description"),
+            "formats": formats,
+            "timestamp": parse_iso8601(info.get("uploaded_at")),
+            "upload_date": unified_strdate(info.get("uploaded_at")),
```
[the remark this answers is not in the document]
flashdagger: Good to know.

flashdagger:
I released the extractor as yt-dlp plugin under https://pypi.org/project/ytdlp-youmaker/

PR #30212, opened 2021-11-08
Add Glomex IEs

zmousm:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This patch adds support for Glomex videos, both embeddable and on the (demo) site video.glomex.com. Glomex operates a media exchange platform, which powers their video marketplace service, based on revenue sharing. This is used quite a lot in large parts of Europe, primarily by digital news outlets, which embed in their assets video clips provided by various content providers (for example TV channels), tied with advertising. 

* Add new IEs
  * GlomexBaseIE: Base IE class
  * GlomexIE: Extract videos from video.glomex.com (by deferring to
    glomex:embed)
  * GlomexEmbedIE: Extract Glomex videos by matching the player URL
* Query the API to extract metadata, detect video formats and get the
  respective (JWT protected) stream/source URLs
* The API query may return one or more videos: the latter case is
  treated as a playlist
  * As this is otherwise identically handled, a separate IE was not
    deemed necessary
  * However title and description fields are not set for playlist
    results
    * They do not exist in the parent object; obtaining them from the
      first entry is not indicative for the playlist content
    * As the playlist order is not always stable (this is true at least
      for related videos playlists), it makes writing test cases
      impossible
* Let GenericIE detect embeds by matching all three integration methods:
  * HTML: glomex-player tag or data attributes
  * Javascript: naive parsing of inline scripts for string constants
    assigned to integration parameters
  * Iframe: src attribute GlomexEmbedIE._VALID_URL
* Let GlomexIE and the former embed detection pass the origin URL to
  GlomexEmbedIE by smuggling it in the player URL, as this is an
  expected parameter in API requests
* Add test cases for both single videos and two playlist flavors

### DIFF
--- a/youtube_dl/extractor/glomex.py
+++ b/youtube_dl/extractor/glomex.py
@@ -1,0 +1,273 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from ..compat import (
+    compat_str,
+    compat_parse_qs,
+    compat_urllib_parse_urlparse,
+    compat_urllib_parse_urlencode,
+)
+from ..utils import (
+    ExtractorError,
+    determine_ext,
+    int_or_none,
+    try_get,
+    smuggle_url,
+    unsmuggle_url,
+    unescapeHTML,
+)
+
+
+class GlomexBaseIE(InfoExtractor):
+    _DEFAULT_ORIGIN_URL = 'https://player.glomex.com/'
+    _API_URL = 'https://integration-cloudfront-eu-west-1.mes.glomex.cloud/'
+
+    @staticmethod
+    def _smuggle_origin_url(url, origin_url):
+        return smuggle_url(url, {'origin': origin_url})
+
+    @classmethod
+    def _unsmuggle_origin_url(cls, url, fallback_origin_url=None):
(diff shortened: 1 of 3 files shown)


youtube_dl/extractor/glomex.py
```diff
@@ -0,0 +1,273 @@
...
+                if not video_id_js or not integration_js:
+                    continue
+                yield cls.build_player_url(video_id_js, integration_js, url)
+
+    def _real_extract(self, url):
+        url, origin_url = self._unsmuggle_origin_url(url)
+        embed_id = self._match_id(url)
+        query = compat_parse_qs(compat_urllib_parse_urlparse(url).query)
+        video_id = query['playlistId'][0]
+        # perhaps redundant
```
dirkf, line 269: Do you want the first &playlistId=... or the last? Maybe `query['playlistId'][-1]`? Obvs these are probably the same.

zmousm: You are right about the discrepancy between first and last match.

I thought parsing the URL query string would be more robust, yet it will not even get to this point if the `_VALID_URL` regex fails, as the IE will not match.

I will change it per your last suggestion, thus no `_match_id` call or URL parsing.

Thanks

zmousm:
Is there anything else I can do to have this patch reviewed?

PR #30224, opened 2021-11-10
fix tiktok when logged in

wranai:

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

TikTok now (since a few days ago?) uses a different format for metadata when the user is logged in (I used a cookie file exported from Chrome).

This patch just adds a parser for the new format: if the original fails, it tries the modified format. It's basically a copy/pasted/adapted version of the original few lines of a parser, nothing big.

### DIFF
--- a/youtube_dl/extractor/tiktok.py
+++ b/youtube_dl/extractor/tiktok.py
@@ -107,10 +107,18 @@ class TikTokIE(TikTokBaseIE):
     def _real_extract(self, url):
         video_id = self._match_id(url)
         webpage = self._download_webpage(url, video_id)
-        page_props = self._parse_json(self._search_regex(
-            r'<script[^>]+\bid=["\']__NEXT_DATA__[^>]+>\s*({.+?})\s*</script',
-            webpage, 'data'), video_id)['props']['pageProps']
-        data = try_get(page_props, lambda x: x['itemInfo']['itemStruct'], dict)
+        try:
+            page_props = self._parse_json(self._search_regex(
+                r'<script[^>]+\bid=["\']__NEXT_DATA__[^>]+>\s*({.+?})\s*</script',
+                webpage, 'data'), video_id)['props']['pageProps']
+            data = try_get(page_props, lambda x: x['itemInfo']['itemStruct'], dict)
+        except:
+            page_props = self._parse_json(self._search_regex(
+                r'<script[^>]+\bid=["\']sigi-persisted-data[^>]+>window\[\'SIGI_STATE\']=({.+?});window\[',
+                webpage, 'data'), video_id)
+            data = try_get(page_props, lambda x: x['ItemModule'][video_id], dict)
+            author = try_get(page_props, lambda x: x['UserModule']['users'][data['author']], dict)
+            data['author'] = author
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/tiktok.py
```diff
@@ -107,10 +107,18 @@ def _real_initialize(self):
...
+                r'<script[^>]+\bid=["\']__NEXT_DATA__[^>]+>\s*({.+?})\s*</script',
+                webpage, 'data'), video_id)['props']['pageProps']
+            data = try_get(page_props, lambda x: x['itemInfo']['itemStruct'], dict)
+        except:
+            page_props = self._parse_json(self._search_regex(
+                r'<script[^>]+\bid=["\']sigi-persisted-data[^>]+>window\[\'SIGI_STATE\']=({.+?});window\[',
+                webpage, 'data'), video_id)
+            data = try_get(page_props, lambda x: x['ItemModule'][video_id], dict)
+            author = try_get(page_props, lambda x: x['UserModule']['users'][data['author']], dict)
+            data['author'] = author
```
dirkf, line 121: Here `data` could be `None` if l.119 failed. Either add `or {}` to that line, or follow it with a test `if data:`

youtube_dl/extractor/tiktok.py
```diff
@@ -107,10 +107,18 @@ def _real_initialize(self):
...
         if not data and page_props.get('statusCode') == 10216:
             raise ExtractorError('This video is private', expected=True)
         return self._extract_video(data, video_id)
```
dirkf, line 124: Should we raise `ExtractorError('Unable to extract video data') here if `data` is falsey? Otherwise it would just crash in the called routine. If you say that's not really related to your PR, I would agree.

youtube_dl/extractor/tiktok.py
```diff
@@ -107,10 +107,18 @@ def _real_initialize(self):
...
-        page_props = self._parse_json(self._search_regex(
-            r'<script[^>]+\bid=["\']__NEXT_DATA__[^>]+>\s*({.+?})\s*</script',
-            webpage, 'data'), video_id)['props']['pageProps']
-        data = try_get(page_props, lambda x: x['itemInfo']['itemStruct'], dict)
+        try:
+            page_props = self._parse_json(self._search_regex(
+                r'<script[^>]+\bid=["\']__NEXT_DATA__[^>]+>\s*({.+?})\s*</script',
+                webpage, 'data'), video_id)['props']['pageProps']
+            data = try_get(page_props, lambda x: x['itemInfo']['itemStruct'], dict)
+        except:
```
dirkf, line 115: The style checker will complain that there's no exception class/tuple (eg, `(ExtractorError, RegexNotFoundError, KeyError, )`).

Otherwise you could make all the calls non-fatal and (as appropriate) test for falsity or supply defaults.

dirkf:
I don't think it's just logged-in access. In #30251, we found this new format in other pages. No doubt it will be rolled out to all pages eventually. You could also use the patch that I supplied in #30251, totally having forgotten your PR!

dirkf:
I updated the patch in the issue linked above.

PR #30227, opened 2021-11-11
[extractor/megatvcom] Add new extractor

zmousm:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This patch adds support for video extraction from megatv.com, a popular Greek content provider and (recently re-)established TV network.

* Add new IEs
  * MegaTVComBaseIE: Base IE class
  * MegaTVComIE: Extract from TV VOD pages and news articles, i.e. all sorts of pages showing videos on megatv.com
  * MegaTVComEmbedIE: Extract iframe-embeddable megatv.com videos
* When video_id is not matched in the URL, namely for news articles, extract it (article_id) from a particular element on the web page
* Derive metadata and sources directly from the web page, from data attributes of the player placeholder element and other commonly used elements
* Let MegaTVComEmbedIE defer to MegaTVComIE for extraction, as the metadata on the embeddable page are some times slightly different, for the same video

### DIFF
--- a/youtube_dl/extractor/megatvcom.py
+++ b/youtube_dl/extractor/megatvcom.py
@@ -1,0 +1,186 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import hashlib
+import re
+
+from .common import InfoExtractor
+from ..compat import (
+    compat_str,
+    compat_parse_qs,
+    compat_urllib_parse_urlparse,
+)
+from ..utils import (
+    HEADRequest,
+    ExtractorError,
+    determine_ext,
+    get_element_by_class,
+    unified_timestamp,
+    extract_attributes,
+    clean_html,
+    unescapeHTML,
+)
+
+
+class MegaTVComBaseIE(InfoExtractor):
+    _PLAYER_DIV_ID = 'player_div_id'
+
+    def _extract_player_attrs(self, webpage):
+        PLAYER_DIV_RE = r'''(?x)
+        <div(?:
+            id=(?P<_q1>[\"\'])(?P<%(pdi)s>%(pdi)s)(?P=_q1)|
+            [^>]*?
+        )+>
+        ''' % {'pdi': self._PLAYER_DIV_ID}
+        for mobj in re.finditer(PLAYER_DIV_RE, webpage):
+            if mobj.group(self._PLAYER_DIV_ID):
+                player_el = mobj.group(0)
+                break
+        else:
+            raise ExtractorError('no <div id="%s"> element found in webpage' %
+                                 self._PLAYER_DIV_ID)
+        return {
+            re.sub(r'^data-(?:kwik_)?', '', k): v
+            for k, v in extract_attributes(player_el).items()
+            if k not in ('id',)
+        }
+
+
+class MegaTVComIE(MegaTVComBaseIE):
+    IE_NAME = 'megatvcom'
+    IE_DESC = 'megatv.com videos'
+    _VALID_URL = r'https?://(?:www\.)?megatv\.com/(?:(?!\d{4})[^/]+/(?P<id>\d+)/[^/]+|\d{4}/\d{2}/\d{2}/.+)'
+
+    _TESTS = [{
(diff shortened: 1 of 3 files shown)


youtube_dl/extractor/megatvcom.py
```diff
@@ -0,0 +1,186 @@
...
+)
+
+
+class MegaTVComBaseIE(InfoExtractor):
+    _PLAYER_DIV_ID = 'player_div_id'
+
+    def _extract_player_attrs(self, webpage):
+        PLAYER_DIV_RE = r'''(?x)
+        <div(?:
+            id=(?P<_q1>[\"\'])(?P<%(pdi)s>%(pdi)s)(?P=_q1)|
```
rautamiekka, line 31: Since you're using `'''`, you don't need to escape any quotes (in fact this might break it) unless it's the same one in the start or end of the string, touching the same quote.

Same for the other such strings later in the code.

youtube_dl/extractor/megatvcom.py
```diff
@@ -0,0 +1,186 @@
...
+            'thumbnail': thumbnail,
+            'timestamp': timestamp,
+            'formats': formats,
+        }
+
+
+class MegaTVComEmbedIE(MegaTVComBaseIE):
+    IE_NAME = 'megatvcom:embed'
+    IE_DESC = 'megatv.com embedded videos'
+    _VALID_URL = r'https?://(?:www\.)?megatv.com/embed/?\?p=\d+'
```
rautamiekka, line 127: Missing dot escape.

youtube_dl/extractor/megatvcom.py
```diff
@@ -0,0 +1,186 @@
...
+    }]
+
+    @classmethod
+    def _extract_urls(cls, webpage, origin_url=None):
+        # make the scheme in _VALID_URL optional
+        _URL_RE = r'(?:https?:)?//' + cls._VALID_URL.split('://', 1)[1]
+        EMBED_RE = r'''(?x)
+            <iframe[^>]+?src=(?P<_q1>%(quot_re)s)
+                (?P<url>%(url_re)s)(?P=_q1)
+        ''' % {'quot_re': r'[\"\']', 'url_re': _URL_RE}
```
rautamiekka, line 160: No need to escape a double quote inside a single (in fact that might break this due to `r'`).

youtube_dl/extractor/megatvcom.py
zmousm: You are right that there is no need to escape quotes, so I will remove them. However I do not think there can be any breakage, at least not in this case, since the backslash makes no difference as to what the pattern matches:
```py3
patterns = (r'''["']''', r'''[\"']''', r'''["\']''', r'''[\"\']''',)
strings = ('"', '\"', "'", "\'")
ctx = 'div id=%swhatever'
all(re.match(ctx % p, ctx % s) for p in patterns for s in strings) # True
```

I noticed there are other such cases in the code base (there are more, these are on the same line):
```
youtube_dl/extractor/generic.py:3119:            r'''(?x)<iframe[^>]+src=(["\'])(?P<url>(?:https?:)?//
youtube_dl/extractor/myspace.py:113:                r'''<button.*data-song-id=(["\'])%s\1.*''' % video_id,
youtube_dl/extractor/peertube.py:498:            r'''(?x)<iframe[^>]+\bsrc=["\'](?P<url>(?:https?:)?//%s/videos/embed/%s)'''
youtube_dl/extractor/simplecast.py:103:            r'''(?x)<iframe[^>]+src=["\']
```

youtube_dl/extractor/megatvcom.py
zmousm: Right, thanks

youtube_dl/extractor/megatvcom.py
zmousm: Right. I am not sure about the breakage, again:

```py3
patterns = (r'["\']', r'[\"\']')
strings = ('"', '\"', "'", "\'")
ctx = 'iframe src=%shttps://whatever'
all(re.match(ctx % p, ctx % s) for p in patterns for s in strings) # True
```

zmousm:
Is there anything else I can do to have this patch reviewed?

Issue #29715, opened 2021-08-01
How to make header "Host:" the first header in headers list


emadmahdi:
## Checklist

- [ ] I'm asking a question
- [ ] I've looked through the README and FAQ for similar questions
- [ ] I've searched the bugtracker for similar questions including closed ones

## Question

WRITE QUESTION HERE

Some video websites started to check headers sequence ... how can i modify the sequence from python ... here is an example of this problem:

=== Resolve perfectly ===
GET   https://dood.la/e/vvdiu7jlpo2o
Host: dood.la
User-Agent: Mozilla/5.0

=== Does not resolve and returns error 403 ===
GET   https://dood.la/e/vvdiu7jlpo2o
User-Agent: Mozilla/5.0
Host: dood.la

kikuyan:
Headers are stored in dict (see std_headers in utils.py), and if the order is important, I think you should use Python 3.7+.

emadmahdi:
I actually found a workaround ... I can change the "User-Agent" and that will force the video server to expect the header "Host" in a different sequence.

I got that by testing CHROME Browser & EDGE Browser ... they both send "Host" as 1st header ... while MS IE Browser send it as 7th header and all the three browsers are resolving correctly ... which means that the video server "doodstream" is requesting the "Host" header location "sequence" based on the value of the "User-Agent" header.

I use a software named ... "Progress Telerik Fiddler Web Debugger v5.0" ... It allow me to send raw http requests and that was how i found this issue and how I debugged the three browsers.

I also tried the new version of "ResolveURL" v5.1.36  and it do resolve this video server "doodstream" correctly while v5.1.30 was unable to resolve it.

dirkf:
You may have found [this bug[(https://sansec.io/research/http-header-order-is-important).

[RFC2616 4.2](http://tools.ietf.org/html/rfc2616#section-4.2):
>The order in which header fields with differing field names are received is not significant. However, it is "good practice" to send general-header fields first, followed by request-header or response- header fields, and ending with the entity-header fields.

emadmahdi:
@dirkf  ... yes you are right ... it is the issue that i am talking about.

I also noticed that if the "Host" header is at the end ... this will trigger cloudflare protection with error 403 ... while if the "Host" header is at the beginning then it will resolve the video link correctly ... here is an example:


=== Works & Resolve correctly ===
Host: dood.la
[...]

PR #30240, opened 2021-11-15
[Podchaser] Add new extractor

connercsbn:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This adds support for **podchaser.com**, a podcast aggregator. All the podcast info is grabbed from JSON embedded in the HTML. If multiple episodes are found, it returns a playlist--otherwise, just the one result.

A couple notes:

 - Currently, the json is written in such a way that the only way I know to parse it is to parse it twice. Maybe it's double encoded? I tried running it through `_escape_html`, but that wasn't working, so right now if _parse_json returns a string, it parses again so it returns a dict.
 ```
        while isinstance(data, compat_str):
            data = self._parse_json(data, audio_id)
``` 

 - When I tested this on Python 2.7, it threw a `DownloadError: ERROR: Unable to download webpage: HTTP Error 403: Forbidden`. I'll keep working on fixing that, but until then, I thought I'd throw this up in case someone has a simple answer for that

### DIFF
--- a/youtube_dl/extractor/podchaser.py
+++ b/youtube_dl/extractor/podchaser.py
@@ -1,0 +1,98 @@
+# coding: utf-8
+from __future__ import unicode_literals
+import re
+from datetime import datetime
+from ..utils import float_or_none, try_get, str_to_int
+from ..compat import compat_str
+from .common import InfoExtractor
+
+
+class PodchaserIE(InfoExtractor):
+    _VALID_URL = r'''(?x)
+        https?://(?:www\.)?podchaser\.com/
+        (?:(?:podcasts)|(?:creators))
+        /[\w\-]+-
+        (?:
+            (?P<creator_id>[\d]+[\w]+)|(?P<podcast_id>[\d]+))
+        (?:/episodes/[\w\-]+-
+            (?P<id>[\d]+))?'''
+
+    _TESTS = [{
+        'url': 'https://www.podchaser.com/podcasts/cum-town-36924/episodes/ep-285-freeze-me-off-104365585',
+        'info_dict': {
+            'id': '104365585',
+            'title': "Ep. 285 – freeze me off",
+            'description': 'cam ahn',
+            'thumbnail': r're:^https?://.*\.jpg$',
+            'ext': 'mp3',
+            'categories': ['Comedy'],
+            'series': 'Cum Town',
+            'duration': 3708,
+            'timestamp': 1636560059,
+            'upload_date': '20211110'
+        }
+    }, {
+        'url': 'https://www.podchaser.com/podcasts/the-bone-zone-28853',
+        'info_dict': {
+            'id': '28853',
+            'title': 'The Bone Zone',
+            'description': 'md5:c39acd897170a8bf3ad94fc45dc25060',
+        },
+        'playlist_count': 6
+    }, {
+        'url': 'https://www.podchaser.com/creators/todd-glass-107ZzkFiEQ',
+        'info_dict': {
+            'id': '107ZzkFiEQ',
+            'title': 'Todd Glass',
+            'description': 'md5:f0302588c98d474d4e8551f4914042c9',
+        },
+        'playlist_mincount': 48
+    }, {
+        'url': 'https://www.podchaser.com/podcasts/sean-carrolls-mindscape-scienc-699349/episodes',
+        'info_dict': {
+            'id': '699349',
+            'title': "Sean Carroll's Mindscape: Science, Society, Philosophy, Culture, Arts, and Ideas",
+            'description': 'md5:a1ed5db77378a736b3a85a2331cbc05a'
+        },
+        'playlist_count': 25
+    }]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        audio_id, podcast_id, creator_id = compat_str(mobj.group('id')), \
+            compat_str(mobj.group('podcast_id')), compat_str(mobj.group('creator_id'))
+        webpage = self._download_webpage(url, audio_id)
+        page_title = self._html_search_meta(['title', 'og:title', 'twitter:title'], webpage, default=None) \
+            or self._search_regex(
+                r'<h1[^>]*>(.+?)</h1>', webpage, 'title', fatal=False, default="Podchaser Podcast")
+        page_description = self._html_search_meta(['description', 'og:description', 'twitter:description'], webpage, default=None)
+
+        data = self._search_regex(
+            r'window.__APP_STATE__\s*=\s*(["\']?{.+?}["\']?);</script>', webpage, 'app state')
+
+        while isinstance(data, compat_str):
+            data = self._parse_json(data, audio_id)
+
+        episodes = try_get(data, lambda x: x['podcast']['episodes']['entities'], dict) or {}
+        episode_list = [(episodes.get(episode_id), episode_id) for episode_id in episodes]
+
+        entries = [{
+            'id': episode_id,
+            'title': episode.get('title'),
+            'description': episode.get('description'),
+            'url': episode.get('audio_url'),
+            'thumbnail': try_get(episode, lambda x: x['podcast']['image_url']),
+            'duration': str_to_int(episode.get('length')),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: youtube_dl/extractor/podchaser.py [hunk 1]
```diff
@@ -0,0 +1,98 @@
...
+        audio_id, podcast_id, creator_id = compat_str(mobj.group('id')), \
+            compat_str(mobj.group('podcast_id')), compat_str(mobj.group('creator_id'))
+        webpage = self._download_webpage(url, audio_id)
+        page_title = self._html_search_meta(['title', 'og:title', 'twitter:title'], webpage, default=None) \
+            or self._search_regex(
+                r'<h1[^>]*>(.+?)</h1>', webpage, 'title', fatal=False, default="Podchaser Podcast")
+        page_description = self._html_search_meta(['description', 'og:description', 'twitter:description'], webpage, default=None)
+
+        data = self._search_regex(
+            r'window.__APP_STATE__\s*=\s*(["\']?{.+?}["\']?);</script>', webpage, 'app state')
```
dirkf, line 71: Move the capture group inside the quotes: then `_parse_json()` will work the first time:
```
        data = self._search_regex(
            r'window.__APP_STATE__\s*=\s*["\']?({.+?})["\']?;</script>', webpage, 'app state')
```

youtube_dl/extractor/podchaser.py
```diff
@@ -0,0 +1,98 @@
...
+from ..utils import float_or_none, try_get, str_to_int
+from ..compat import compat_str
+from .common import InfoExtractor
+
+
+class PodchaserIE(InfoExtractor):
+    _VALID_URL = r'''(?x)
+        https?://(?:www\.)?podchaser\.com/
+        (?:(?:podcasts)|(?:creators))
+        /[\w\-]+-
```
dirkf, line 14: To put `-` in `[]`, just make it the last item, hence `[\w-]`

youtube_dl/extractor/podchaser.py
```diff
@@ -0,0 +1,98 @@
...
+            'id': '699349',
+            'title': "Sean Carroll's Mindscape: Science, Society, Philosophy, Culture, Arts, and Ideas",
+            'description': 'md5:a1ed5db77378a736b3a85a2331cbc05a'
+        },
+        'playlist_count': 25
+    }]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        audio_id, podcast_id, creator_id = compat_str(mobj.group('id')), \
```
dirkf, line 62: The URL should have been `compat_str` already, so the regex groups will be too.

youtube_dl/extractor/podchaser.py
```diff
@@ -0,0 +1,98 @@
...
+            compat_str(mobj.group('podcast_id')), compat_str(mobj.group('creator_id'))
+        webpage = self._download_webpage(url, audio_id)
+        page_title = self._html_search_meta(['title', 'og:title', 'twitter:title'], webpage, default=None) \
+            or self._search_regex(
+                r'<h1[^>]*>(.+?)</h1>', webpage, 'title', fatal=False, default="Podchaser Podcast")
```
dirkf, line 67: According to the docs, not having a title should be an error, so scrap `fatal=False` in  the second case.

youtube_dl/extractor/podchaser.py
connercsbn: - [x] see 4d83c072b195f4deaf41f0ac08b909dfc87fa13a

youtube_dl/extractor/podchaser.py
connercsbn: - [x] see 89ef8b9b99a5cdf24a81dc3070106302bef64724

youtube_dl/extractor/podchaser.py
connercsbn: This causes an error:
```
 ERROR: 104365585: Failed to parse JSON  (caused by JSONDecodeError('Expecting property name enclosed in double quotes: line 1 column 2 (char 1)'))
```
Here are the first 100 characters of the data string being passed in, for reference.
```
{\"profile\":{\"account\":{\"error\":null}},\"modals\":{\"modalStack\":[],\"activity_modals\":{}},\"
```

dirkf: Sorry,  that's quite right. The framework in use is saving the page state in a stringified JSON and it needs two decodes to get it.

youtube_dl/extractor/podchaser.py
```diff
@@ -0,0 +1,98 @@
...
+        episode_list = [(episodes.get(episode_id), episode_id) for episode_id in episodes]
+
+        entries = [{
+            'id': episode_id,
+            'title': episode.get('title'),
+            'description': episode.get('description'),
+            'url': episode.get('audio_url'),
+            'thumbnail': try_get(episode, lambda x: x['podcast']['image_url']),
+            'duration': str_to_int(episode.get('length')),
+            'timestamp': datetime.strptime(episode.get('air_date'), '%Y-%m-%d %H:%M:%S').timestamp(),
```
dirkf, line 86: When I tried this it crashed ('datetime.datetime has no timestamp').

But we have `unified_timestamp()` in `utils.py` and `unified_timestamp(episode.get('air_date'))` seems to work OK.

youtube_dl/extractor/podchaser.py
```diff
@@ -0,0 +1,98 @@
...
+            'description': 'md5:a1ed5db77378a736b3a85a2331cbc05a'
+        },
+        'playlist_count': 25
+    }]
+
+    def _real_extract(self, url):
+        mobj = re.match(self._VALID_URL, url)
+        audio_id, podcast_id, creator_id = mobj.group('id'), \
+            mobj.group('podcast_id'), mobj.group('creator_id')
+        webpage = self._download_webpage(url, audio_id)
```
dirkf, line 64: In testing (py2/3), I got 403 Forbidden, depending on the UA used.

I believe that Cloudflare has broken the web by randomly forbidding certain UA strings (with no diagnostic). If the UA string only includes a 2-component version

youtube_dl/extractor/podchaser.py
connercsbn: - [x] see 6ddf44e2f31dc74d819cfe9fbc25b717e214e40e

youtube_dl/extractor/podchaser.py
connercsbn: On my end, this didn't solve the problem. I tried your solution, and I also tried removing the Chrome portion of the UA header altogether. Either way I got a 403 error unless I was using Python 3.7+

dirkf: Totally weird. My test versions are Py 2.7.17 and 3.5.2.

Do you ever get 403 in the browser? If not, what happens with the UA that your browser uses?

connercsbn: With Py 2.7.17 on my laptop and Py 2.7.18 on desktop, I get the 403 error, and I tried using my browser's UA. Never gotten 403 in browser.

I don't know if this is relevant, but when I tried running a curl request with only the UA header,
`curl 'https://www.podchaser.com/podcasts/world-record-podcast-with-bren-1020594/episodes/episode-57-geno-randazzo-82092499' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:94.0) Gecko/20100101 Firefox/94.0'`
it worked on my laptop (which I haven't been using for this project), but on my main PC, the response is always the Cloudflare captcha page.

dirkf: Well, that seems to confirm the responsible party.

In the linked issue it appeared that the **order** of headers might be important to Cloudflare. I suppose that could be different between the various web clients that you tested. It is [observed in the linked issue that the Python web client before 3.7 stores the headers in a `dict`](https://github.com/ytdl-org/youtube-dl/issues/29715#issuecomment-890611459) where the order can't be guaranteed.

connercsbn: Ok, then I think I'll try changing this so it downloads the JSON from their API. Seems like there's no problem getting data that way when using Py 2.7.

connercsbn: Done 6ddf44e

PR #30294, opened 2021-11-30
Strip path after sanitizing

ItsKlork:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

During sanitization, the path may break (For example, the title "? abc" will become " abc", breaking the path).
The issue can be fixed by calling `.strip()` when sanitizing.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -200,6 +200,7 @@ class TestUtil(unittest.TestCase):
             return
 
         self.assertEqual(sanitize_path('abc'), 'abc')
+        self.assertEqual(sanitize_path('? abc'), 'abc')
         self.assertEqual(sanitize_path('abc/def'), 'abc\\def')
         self.assertEqual(sanitize_path('abc\\def'), 'abc\\def')
         self.assertEqual(sanitize_path('abc|def'), 'abc#def')
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -2129,7 +2129,7 @@ def sanitize_path(s):
     if drive_or_unc:
         norm_path.pop(0)
     sanitized_path = [
-        path_part if path_part in ['.', '..'] else re.sub(r'(?:[/<>:"\|\\?\*]|[\s.]$)', '#', path_part)
+        path_part if path_part in ['.', '..'] else re.sub(r'(?:[/<>:"\|\\?\*]|[\s.]$)', '#', path_part).strip()
         for path_part in norm_path]
     if drive_or_unc:
         sanitized_path.insert(0, drive_or_unc + os.path.sep)


dirkf:
>Before submitting a pull request make sure you have:
>...
>    Covered the code with tests (note that PRs without tests will be REJECTED)

In this case an additional test in `test/testutils.py` would help to demonstrate what problem is being fixed, which isn't clear to me at least (quite possibly my fault).

ItsKlork:
@dirkf Just added it!
A file path such as "? hello" turns into " hello", something like that causes issues when later trying to access the path.

dirkf:
If the problem is that a sanitised path might have embedded spaces (welcome to MS-DOS or maybe Windows 3), why wouldn't any whitespace in a path be sanitised out? And then we have to ask why that wasn't done in the original code.

ItsKlork:
I'm not talking about all whitespace, I'm specifically referring to ones at the beginning

ItsKlork:
> If a file is saved as ' Foo.txt', where the leading character(s) is an ASCII Space (0x20), it will be saved to the file system as 'Foo.txt'.

From: [https://docs.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/file-folder-name-whitespace-characters](https://docs.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/file-folder-name-whitespace-characters)

After windows removes the whitespace from the beginning, future file accesses will break.

dirkf:
The linked support document is hopeless. It doesn't say what in Windows might remove initial space(s) when saving a file ("... will be saved ..." by what?). My bet is that it is talking about the common file dialog, where surrounding the filename to be saved in "" would preserve spaces (at least it did in 2011). Then it goes on to describe a thing called Object Manager that no non-programmer Windows user has ever heard of, where it says that leading spaces are preserved.

If yt-dl should create a file whose name begins with a space:
* will it be accessible from the Windows GUI? (I expect so)
* will it be accessible from PowerShell or cmd using appropriate quoting? (I'm sure so)
* will something in Windows change the name to remove the initial space without user intervention? (I don't believe so)

If my understanding is correct, the situation is the same between Windows and Unix-like platforms, ie that a path component with initial space(s) might be confusing, but not otherwise problematic.

ItsKlork:
Most external downloads and maybe the default one as well (Haven't tested it really) try accessing the "empty" created file to write the data and just throw a FileNotFoundError.

I attached a screenshot using a demo youtube video.
Screenshot_1

dirkf:
The external downloader code typically constructs the command as a list `[prog, arg1, arg2 ...]` which is then passed to `subprocess.Popen()`.

As Windows doesn't have an API like the `exec()` family in POSIX where the program to be executed and each of its command-line parameters are separate arguments, but only the Win32 `CreateProcess()` API that takes an entire command-line as its argument, Python has to stringify the command list such that Windows will correctly [split it into the program to be executed and its command-line parameters](https://docs.python.org/2/library/subprocess.html#converting-argument-sequence).

The Python manual from which that link comes says that on Windows a list of arguments is stringified with quoting like this:
```
>>> import subprocess
>>> x = ['prog', '-in', ' x y.z']
>>> subprocess.list2cmdline(x)
'prog -in " x y.z"'
```
So this ought to work. From the screenshot it seems like aria2c got the wrong filename. The downloader code that invokes aria2c splits the path into base and directory: perhaps that's breaking.

A text grab with `-v` should show how the command is being invoked. Or set a breakpoint at `downloader/external.py` l.185 and step through to the call to `subprocess.Popen()` to see how the argument list is constructed.

PR #30314, opened 2021-12-03
Remove 'URL' from usage description

clemencetbk:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Don't know if it's just me but the current command reads like `youtube-dl URL https://youtube.com/something` should be a valid usage when it actually needs to be `youtube-dl https://youtube.com/something`

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
 # DESCRIPTION
 **youtube-dl** is a command-line program to download videos from YouTube.com and a few more sites. It requires the Python interpreter, version 2.6, 2.7, or 3.2+, and it is not platform specific. It should work on your Unix box, on Windows or on macOS. It is released to the public domain, which means you can modify it, redistribute it or use it however you like.
 
-    youtube-dl [OPTIONS] URL [URL...]
+    youtube-dl [OPTIONS] [URL...]
 
 # OPTIONS
     -h, --help                           Print this help text and exit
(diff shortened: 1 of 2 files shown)


rautamiekka:
If there's only `[URL...]`, it means "there can be at least 0 URL's to download something". `<URL...>` would fix it to mean "there can be at least 1 URL to download something".

clemencetbk:
It seems that italics may be more standard than <>: https://man7.org/linux/man-pages/man1/man.1.html

rautamiekka:
Fair enough.

Ashish0804:
I personally prefer `<URL>` as well but I don't think so that anyone would mistake it with the current syntax.
And even if anyone does,
`youtube-dl URL https://youtube.com/something` SHOULD still work as it will simply spit out that `URL` is not a valid url.

rautamiekka:
> `youtube-dl URL https://youtube.com/something` SHOULD still work as it will simply spit out that `URL` is not a valid url.

Ya, it will.

clemencetbk:
> but I don't think so that anyone would mistake it with the current syntax.

I did 😅

It's true that you get a warning but I think it would be nice if it were clear what should be replaced with an argument and what should be there as-is

dirkf:
> It seems that italics may be more standard than <>: https://man7.org/linux/man-pages/man1/man.1.html

Not so standard that they aren't displayed as underlined text instead in a Linux console!

PR #30356, opened 2021-12-12
[options] modify --get-thumbnail option name to --get-thumbnail-url

kikuyan:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

From its name, users may expect that `--get-thumbnail` gets thumbnail, but it actually displays the thumbnail URL. It is peculiar among other `--get-...` option names.
Better to be renamed to `--get-thumbnail-url`.

`--get-thumbnail` will continue to be recognized by python optparse and work.

### DIFF
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -603,7 +603,7 @@ def parseOpts(overrideArguments=None):
         action='store_true', dest='getid', default=False,
         help='Simulate, quiet but print id')
     verbosity.add_option(
-        '--get-thumbnail',
+        '--get-thumbnail-url',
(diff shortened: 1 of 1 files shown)


dirkf:
yt-dlp has deprecated the `--get-...` options in favour of `--print ...`. In this case it has `--print thumbnail`.

Would it be better to take the same route and then introduce `--print thumbnail-url` as synonym of  `--print thumbnail`?

kikuyan:
I don't see the benefit of doing that, might be because I'm too accustomed with `--get-...`.

Issue #30427, opened 2021-12-26
Default installation instructions overwrite system files


lefth:
## Checklist
- [ ] I'm reporting a broken site support issue
- [n/a] I've verified that I'm running youtube-dl version **2021.12.17**
- [ ] I've checked that all provided URLs are alive and playable in a browser
- [ ] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar bug reports including closed ones
- [x] I've read bugs section in FAQ

## Verbose log
See explanation of the problem here: https://askubuntu.com/questions/1383688/after-installing-youtube-dl-on-ubuntu-21-10-with-the-instructions-on-github-many
After installing youtube-dl using the current instructions in the README (`curl -o`), this user lost a key system executable. This is the output of him trying to run a program that was not supposed to be related to youtube-dl (which had been overwritten):
```
$ gedit

Usage: gedit [OPTIONS] URL [URL...]

gedit: error: You must provide at least one URL.
Type youtube-dl --help to see a list of all options.
```

## Description
The installation instructions overwrite symlink referents. Some systems are organized with executables being symlinked to a multi-call binary like busybox, toolbox, or (in this case) firejail. `curl -o` does not overwrite the symlink, but instead overwrites the symlink referent. _The multi-call binary will be overwritten!_ I suggest a change to the installation instructions--instead of:
```
sudo curl -L https://yt-dl.org/downloads/latest/youtube-dl -o /usr/local/bin/youtube-dl
```
The program could safely be installed with:
```
curl -L https://yt-dl.org/downloads/latest/youtube-dl -o /tmp/youtube-dl && sudo mv /tmp/youtube-dl /usr/local/bin/
```

kikuyan:
I think it'd be better to do `sudo rm -f /usr/local/bin/youtube-dl` at the beginning, avoid using `/tmp`.

lefth:
@kikuyan I thought about that, but the failure case is too bad: it would have the side effect of uninstalling youtube-dl if the download fails.

PR #30428, opened 2021-12-26
Installation instructions: prevent multi-call binary from being overwritten

lefth:
This change prevents system damage when a multi-call binary is installed such as "/usr/local/bin/youtube-dl -> /usr/local/bin/firejail". This fixes issue #30427. (Yes, a user really reported this problem and had to ask for help in solving it.)

---

- [n/a] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [n/a] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [n/a] Covered the code with tests (note that PRs without tests will be REJECTED)
- [n/a] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This pull request fixes the installation instructions (based on either curl or wget). If a system uses a multi-call binary with symlinks, such as "/usr/local/bin/youtube-dl -> /usr/local/bin/firejail", the multi-call binary will be overwritten when a user follows the installation instructions. Fortunately a simple change to the installation command solves that issue. I tested and found that both wget and curl overwrite a symlink referent, and outputting to a temp directory and installing with `sudo mv` solves the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,12 +20,12 @@ youtube-dl - download videos from youtube.com or other video platforms
 
 To install it right away for all UNIX users (Linux, macOS, etc.), type:
 
-    sudo curl -L https://yt-dl.org/downloads/latest/youtube-dl -o /usr/local/bin/youtube-dl
+    curl -L https://yt-dl.org/downloads/latest/youtube-dl -o /tmp/youtube-dl && sudo mv /tmp/youtube-dl /usr/local/bin/
     sudo chmod a+rx /usr/local/bin/youtube-dl
 
 If you do not have curl, you can alternatively use a recent wget:
 
-    sudo wget https://yt-dl.org/downloads/latest/youtube-dl -O /usr/local/bin/youtube-dl
+    wget https://yt-dl.org/downloads/latest/youtube-dl -O /tmp/youtube-dl && sudo mv /tmp/youtube-dl /usr/local/bin/
     sudo chmod a+rx /usr/local/bin/youtube-dl
 
 Windows users can [download an .exe file](https://yt-dl.org/latest/youtube-dl.exe) and place it in any location on their [PATH](https://en.wikipedia.org/wiki/PATH_%28variable%29) except for `%SYSTEMROOT%\System32` (e.g. **do not** put in `C:\Windows\System32`).


dirkf:
According to the AskUbuntu issue, the problem was that `/usr/local/bin/youtube-dl` was a link to `/usr/bin/firejail` while the actual yt-dl program was `/usr/bin/youtube-dl`. And the problem was that the user re-installed instead of updating.

Maybe I'm being dense but I don't see how the proposed replacement command addresses that. In this scenario, doesn't the user have to know that the thing that looks like yt-dl (`echo $(which youtube-dl)`) isn't actually the program? Or does _firejail_ magically recognise a new non-link entry in `/usr/local/bin` and reconfigure the installation as above?

Otherwise, I suggest that this is an unusual case, where the user updated the program incorrectly, and it should be noted separately after showing the typical commands (as before)
```
--- old/youtube-dl/README.md
+++ new/youtube-dl/README.md
@@ -27,6 +27,12 @@
 
     sudo wget https://yt-dl.org/downloads/latest/youtube-dl -O /usr/local/bin/youtube-dl
     sudo chmod a+rx /usr/local/bin/youtube-dl
+
+To update the program, use the self-update option
+
+    sudo youtube-dl --update
+
+NB: in some configurations (in particular, using [firejail](https://man7.org/linux/man-pages/man1/firejail.1.html)) the installation commands may overwrite system files if used to update the program; the file /usr/local/bin/youtube-dl may be a symbolic link with the real program stored elsewhere, such as `/opt` or `/usr/bin`; use `which -a youtube-dl | xargs ls -l`, if available, to distinguish any symbolic link from the youtube-dl program to be updated. Use that path instead of `/usr/local/bin/youtube-dl` in the above commands.
 
 Windows users can [download an .exe file](https://yt-dl.org/latest/youtube-dl.exe) and place it in any location on their [PATH](https://en.wikipedia.org/wiki/PATH_%28variable%29) except for `%SYSTEMROOT%\System32` (e.g. **do not** put in `C:\Windows\System32`).
```

In the Unix-like self-update option, the file `sys.argv[0]` is overwritten; by the time _firejail_ has launched the actual yt-dl program, that should be the correct path.

Or the issue could be a new <s>Rare</s>Frequently Asked Question: "Why have all my programs turned into youtube-dl after updating?"

Possibly also relevant: _wget_ has a `--unlink` option, though _curl_ doesn't AFAIK.

lefth:
Should we be concerned with how common or uncommon this setup is, or if the user upgraded the wrong way? I think not. Overwriting undesired files should be considered a bad, embarrassing bug. Overwriting a link referent (with elevated permissions) is also a classic security vulnerability.

If an occasional user successfully overwrites /usr/local/bin/youtube-dl then realizes that was a bad idea on his system, it's a less serious issue (and more easily solved).

dirkf:
It is a very fair point that symlink referents should not be overwritten, especially given that the installing user is encouraged to become admin, even if some arguably over-tricksy security set-up might have inserted itself in the firing line.

However in the specific problem case, the modified commands don't seem to address the problem of updating the configuration magically transmuted by _firecfg_. Perhaps just add s/t like `! test -h /usr/local/bin/youtube-dl && ...` rather than trying to handle some non-standard set-up?

lefth:
That seems reasonable, depending on how important it is that non-developers understand the commands they are copying/pasting. It reminds me of another common installation procedure, for example the way to install Rust:

```
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
```

Would you want to make a simple installer script (or accept a pull request with one) that would allow a user to run this?:
```
curl <whatever flags> https://github.com/ytdl-org/youtube-dl/blob/master/install.sh | sh
```

That script could use `test -i` and `test -e` (giving a prompt depending on the result)--for example:

An existing youtube-dl executable was found. (Or "was found, linking to <link destination>.") You may abort installation and run "sudo youtube-dl --upgrade". Abort installation? [Y/n]

If the user doesn't abort, it would run curl or wget, using mktemp to ensure the temporary path is available and in the right place for that system. (I don't think there's any use case where overwriting the symlink referent makes sense, even if it points at another youtube-dl executable--because in that case it might be a custom fork or a wrapper script.)

dirkf:
Overwriting a symlink referent can be avoided just by deprecating _curl_ and having `wget --unlink ...`. Unfortunately I guess that the excellent _curl_ is probably more widely installed. Its maintainer seems responsive, though.

lefth:
Even if we convinced the curl maintainer to add a flag for this, it wouldn't be an immediate solution. Until OS maintainers upgrade to the newest version, the installation instructions wouldn't work for anybody. (Ubuntu and Debian are slow to upgrade packages, and Apple likes to ship very old versions of its software.)

dirkf:
I did put the question on the curl-users list, but then I actually did some tests.

Let `$y` represent the yt-dl download, let `tgt` be the destination, let `tgt` -> `precious`, a file of known size.
* `wget -O tgt $y`: the download overwrites precious
* `curl -o tgt $y`: the download overwrites precious
* `wget --unlink -O tgt $y`: the download **overwrites** precious (wget 1.21)
* `for f in tgt; do unlink $f;  wget -O $f $y; done`: precious is safe.

If _wget_ doesn't even get it right, we do need a work-around command-line.

PR #30453, opened 2022-01-02
[youtube] Fix bug where autogenerated subtitles could not be downloaded for videos without ASR

vladh:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Currently, youtube-dl only downloads automatically-generated subtitles from videos that also have ASR subtitles (“automatic speech recognition”, so subtitles automatically generated from speech). This is, however, a bug. There are videos (here's [an example](https://www.youtube.com/watch?v=6XiFntEQXmA)) that have autogenerated subtitles, but no ASR subtitles.

I've attempted to make the minimal possible change to fix this issue, and validate it using the tests. I can confirm that this change fixes the bug for me, and that it does not cause any subtitle tests to fail.

However, I am not familiar with the youtube-dl codebase, and I cannot say for sure that my change will not affect some other aspect of the code, so I'm very open to suggestions.

I think some version of this change should happen, because currently, attempting to download autogenerated subtitles on non-ASR videos simply silently fails without printing anything, which is very frustrating and confusing to the user.

### DIFF
--- a/test/test_subtitles.py
+++ b/test/test_subtitles.py
@@ -88,6 +88,13 @@ class TestYoutubeSubtitles(BaseTestSubtitles):
         subtitles = self.getSubtitles()
         self.assertTrue(subtitles['it'] is not None)
 
+    def test_youtube_automatic_captions_no_asr(self):
+        self.url = '6XiFntEQXmA'
+        self.DL.params['writeautomaticsub'] = True
+        self.DL.params['subtitleslangs'] = ['ro']
+        subtitles = self.getSubtitles()
+        self.assertTrue(subtitles['ro'] is not None)
+
     def test_youtube_translated_subtitles(self):
         # This video has a subtitles track, which can be translated
         self.url = 'Ky9eprVWzlI'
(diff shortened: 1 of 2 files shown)


dirkf:
The description would benefit from expanding the term 'ASR' on first use, especially as it isn't used in the manual.

vladh:
Thank you @dirkf, done!

I've run the tests here: https://github.com/vladh/youtube-dl/pull/1

It's not 100% clear to me what the results are. Everything seems to pass, all the runs are green, all the tests say OK. But somehow, I'm not sure all the tests are showing up? Could you confirm that the tests are running correctly?

dirkf:
As all the core tests are OK, I think the main thing is to confirm that you get, in the CI and/or by running it yourself, with various Python interpreters, the same results for `test/test_subtitles` before and after (except for the test of the change itself).

pukkandan:
With this PR, it is not possible for a user to distinguish between asr and "normal" auto-generated subs. You might want to port over yt-dlp's implementation - specifically, https://github.com/yt-dlp/yt-dlp/commit/ecdc9049c0d8c00ad9ea5218126eefb1e7049385

Issue #30451, opened 2022-01-02
Adding Year format name for RAI


AlexR00T:
## Checklist

- [x] I'm reporting a site feature request
- [x] I've verified that I'm running youtube-dl version **2021.12.17**
- [x] I've searched the bugtracker for similar site feature requests including closed ones

## Description

Hello, could we add a year format name at least for Rai. To use "%(title)s (%(year)s).%(ext)s" to get out filename in format MovieTitle (year).mp4.  After quickly scrolling through rai.py we could use the same json links to extract also year data and then use it to format files by it. It would be really useful for "Arrs" (Radarr, Sonarr) and other applications.

dirkf:
Use (eg) `%(upload_date).4s`, so `"%(title)s (%(upload_date).4s).%(ext)s"`.

For the next 7977 years, anyway.

AlexR00T:
the result for (%(upload_date).4s) is (NA). Didn't dive deep in the rai.py yet since im not really a master of python. I've seen tho that the links .json file on RaiPlays end does have a movie release year info in it. And I would assume theres also one for series. Thanks for the tip tho it's a good start I assume, maybe something is missing in rai.py to get that info from the .json link

dirkf:
```
$ youtube-dl --get-filename -o '%(upload_date).4s' 'http://www.raisport.rai.it/dl/raiSport/media/rassegna-stampa-04a9f4bd-b563-40cf-82a6-aad3529cb4a9.html'
2014
$
```
Plainly if the site provides no date string (`upload_date`, release_date`, not sure about any others), it can't be left-truncated to `.4s`.

You can generate the JSON with `-j` and inspect with a JSON viewer or _jq_ to see what is actually delivered for a particular video page.

PR #30459, opened 2022-01-03
[rai] Added fromat names for year and edit_year

AlexZigante:
Added this format names for automatic library management purposes.
year:  is the release year on Rai
edit_year:  is the first original release year of movie or series.
info usable for "*Arr"s and other similar collection managers

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

Added this two format names for automatic library management purposes.
year: i think is the release year on Rai/Italy
edit_year:  is the first original release year of movie or series which is info usable for automatic importing into "*Arr"s and other similar collection managers

#30451

### DIFF
--- a/youtube_dl/extractor/rai.py
+++ b/youtube_dl/extractor/rai.py
@@ -189,6 +189,10 @@ class RaiPlayIE(RaiBaseIE):
 
         program_info = media.get('program_info') or {}
         season = media.get('season')
+        
+        year = media.get('track_info').get('year')
+        edit_year = media.get('track_info').get('edit_year')
+
 
         info = {
             'id': remove_start(media.get('id'), 'ContentItem-') or video_id,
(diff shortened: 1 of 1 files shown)


AlexR00T:
#30451

dirkf:
I don't think this is the right solution. What you need to do is to make sure that `upload_date` and `release_date` are set if the necessary data is available. Then the year or release_year is part of a standard metadata field. There shouldn't be extractor-specific metadata.

The yt-dl core will set the second field if it's not extracted and the first is valid:
```
                ('timestamp', 'upload_date'),
                ('release_timestamp', 'release_date'),
                ('modified_timestamp', 'modified_date'),
```
So for `upload_date`, if that field might not be explicitly extracted, first check if `timestamp` is always extracted; if not, try to find some date value that can be transformed into a YYYYMMDD string (note that `unified_strdate()` doesn't work on just a 4 digit year); if there's no month or day information, fake it as `0101`, say.

Similarly, for `release_`xxx.

youtube_dl/extractor/rai.py
```diff
@@ -189,6 +189,10 @@ def _real_extract(self, url):
 
         program_info = media.get('program_info') or {}
         season = media.get('season')
+        
+        year = media.get('track_info').get('year')
```
dirkf, line 193: This isn't safe: if `media.get('track_info')` isn't the expected `dict`, it will crash.

PR #21774, opened 2019-07-13
[televizeseznam] add new extractor

arkamar:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [x] At least skimmed through [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site) and [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) sections
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [x] Improvement
- [x] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

This PR adds new extractor for televizeseznam.cz. It implements requests from issues #20900, #26490  and it is also related with #22114

PR #30462, opened 2022-01-03
[streamcz] Fixed Stream.cz extractor.

OndrejBakan:

---

- [x] Bug fix
- [x] Improvement
- [x] New extractor
- [ ] New feature

---

I updated the obsolete Stream.cz extractor. Separated code into classes, so support for playlists, etc. can be easily added. This is my first serious pull request on serious project, so any feedback is appreciated and welcome.


youtube_dl/extractor/streamcz.py
```diff
@@ -1,105 +1,112 @@
...
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
-        api_path = '/episode/%s' % video_id
 
-        req = sanitized_Request(self._API_URL + api_path)
-        req.add_header('Api-Password', _get_api_key(api_path))
-        data = self._download_json(req, video_id)
+        webpage = self._download_webpage(url, video_id)
+        webpage_data = self._html_search_regex(r'window\.APP_SERVER_STATE\s=\s(\{[^;]*)', webpage, 'webpage_data')
```
dirkf, line 83: Try `r'(?s)window\s*\.\s*APP_SERVER_STATE\s*=\s*(\{.+?\})\s*;'`:
* `(?s)` to make `.` match across lines (or pass `re.DOTALL` in `flags`)
* all the `\s*` are at points where arbitrary JS whitespace could occur
* `.+?` (1) `.` in case of `;` inside strings in the JSON (2) `+` definitely need some contents (3) `?` don't match `"{...}; ...x = {...};`
* finally force a teminating `}` inside the group and `;` afterwards.

youtube_dl/extractor/streamcz.py
```diff
@@ -1,105 +1,112 @@
...
-                    'format_id': '%s-%s' % (typ, f['quality']),
-                    'url': f['source'],
-                    'height': int_or_none(f['quality'].rstrip('p')),
-                    'quality': quality,
-                })
-        self._sort_formats(formats)
+        video_detail = try_get(webpage_json, lambda x: x['data']['fetchable']['episode']['videoDetail']['data'], dict)
+
+        # metadata
+        title = video_detail.get('name')
```
dirkf, line 89: Mandatory field: `video_detail.['name']`, or add `or self._generic_title(url)` (not generally done).

youtube_dl/extractor/streamcz.py
```diff
@@ -1,105 +1,112 @@
...
+        description = video_detail.get('perex')
+        duration = video_detail.get('duration')
+        published = video_detail.get('publishTime').get('timestamp')
```
dirkf, line 92: The `publishTime` isn't known to be a `dict`, so `try_get(video_detail, lambda x: x['publishTime']['timestamp'], int)`

youtube_dl/extractor/streamcz.py
```diff
@@ -1,105 +1,112 @@
...
-            title = data['name']
-
-        subtitles = {}
-        srt_url = data.get('subtitles_srt')
-        if srt_url:
-            subtitles['cs'] = [{
-                'ext': 'srt',
-                'url': srt_url,
-            }]
+        sdn_url = video_detail.get('spl') + 'spl2,3,VOD'
```
dirkf, line 98: If `video_detail.get('spl')` should be `None`, or something else that can't have a `compat_str` added, this will crash. The extraction would have failed, but it might be better to crash in `_extract_sdn_formats() ` instead. Try (eg) `'%sspl2,3,VOD' % (str_or_none(video_detail.get('spl')) or '', )`. Or make sure it does crash here with `['spl']` instead of `.get(...)`.

OndrejBakan:
@dirkf Thank you for your code review, I tried to fix the code according to your comments.

arkamar:
Hi, I am original author of https://github.com/ytdl-org/youtube-dl/pull/21774 PR that later evolved to https://github.com/yt-dlp/yt-dlp/pull/1616 which was merged to `yt-dlp`. I believe it make sense to consider backporting current state of `StreamCZ` extractor from `yt-dlp` because future improvements or fixes can be more easily shared between both projects.

dirkf:
That sounds sensible:
* are there any unique features in this PR that ought to be preserved (maybe the updated test)?
* are you interested in making the back-port PR?

With a quick review I see a couple of yt-dlp-isms as well as some f-strings:
* `traverse_obj()` needs to be back-ported, or replaced with a `try_get()` expression
* `_match_valid_url()`.

I have a draft back-port of `traverse_obj()` in some branch or stash. The problem is to turn `nonlocal` into down-level code, as in the `jsinterp.py` back-port.

dirkf:
Superseded by #30634.

PR #30520, opened 2022-01-15
[joj] Fixed extractor'

jastrab:

---

- [x] Bug fix
- [x] Improvement
- [x] New extractor
- [x] New feature

---

I updated the obsolete JOJ extractor.

### DIFF
--- a/youtube_dl/extractor/joj.py
+++ b/youtube_dl/extractor/joj.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 import re
+import json
 
 from .common import InfoExtractor
 from ..compat import compat_str
@@ -21,7 +22,8 @@ class JojIE(InfoExtractor):
                     (?P<id>[^/?#^]+)
                 '''
     _TESTS = [{
-        'url': 'https://media.joj.sk/embed/a388ec4c-6019-4a4a-9312-b1bee194e932',
+        'url':
+        'https://media.joj.sk/embed/a388ec4c-6019-4a4a-9312-b1bee194e932',
         'info_dict': {
             'id': 'a388ec4c-6019-4a4a-9312-b1bee194e932',
             'ext': 'mp4',
@@ -64,16 +66,20 @@ class JojIE(InfoExtractor):
                 r'(?s)(?:src|bitrates)\s*=\s*({.+?});', webpage, 'bitrates',
                 default='{}'),
             video_id, transform_source=js_to_json, fatal=False)
+        quality = self._search_regex(
+            r'var qualityLabels = ({.*?});', webpage, 'quality', default=None)
+        quality = json.loads(quality)
 
         formats = []
         for format_url in try_get(bitrates, lambda x: x['mp4'], list) or []:
             if isinstance(format_url, compat_str):
                 height = self._search_regex(
-                    r'(\d+)[pP]\.', format_url, 'height', default=None)
+                    r'\-(\d+)(p|)\.', format_url, 'height', default=None)
                 formats.append({
                     'url': format_url,
                     'format_id': '%sp' % height if height else None,
                     'height': int(height),
+                    'resolution': quality.get(str(height)+'p')
                 })
         if not formats:
             playlist = self._download_xml(
@@ -104,5 +110,5 @@ class JojIE(InfoExtractor):
             'title': title,
             'thumbnail': thumbnail,
             'duration': duration,
-            'formats': formats,
+            'formats': formats
         }
--- a/youtube_dl/extractor/markiza.py
+++ b/youtube_dl/extractor/markiza.py
@@ -1,21 +1,23 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
+# update 15.01.2022 jastrab
 import re
+import json
 
 from .common import InfoExtractor
-from ..compat import compat_str
 from ..utils import (
     orderedSet,
-    parse_duration,
-    try_get,
+    url_or_none,
+    determine_ext
 )
 
 
 class MarkizaIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?videoarchiv\.markiza\.sk/(?:video/(?:[^/]+/)*|embed/)(?P<id>\d+)(?:[_/]|$)'
+    _VALID_URL = r'https:\/\/(?:www\.)?videoarchiv\.markiza\.sk\/(?:video\/(?:[^\/]+\/)*|embed\/)epizoda\/(?P<id>\d+)(?:[\_\/\-]|$)'
     _TESTS = [{
-        'url': 'http://videoarchiv.markiza.sk/video/oteckovia/84723_oteckovia-109',
+        'url': 'http://videoarchiv.markiza.sk/video/oteckovia/\
+            84723_oteckovia-109',
         'md5': 'ada4e9fad038abeed971843aa028c7b0',
         'info_dict': {
             'id': '139078',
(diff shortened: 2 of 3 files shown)


youtube_dl/extractor/joj.py
```diff
@@ -64,16 +66,20 @@ def _real_extract(self, url):
...
+        quality = self._search_regex(
+            r'var qualityLabels = ({.*?});', webpage, 'quality', default=None)
+        quality = json.loads(quality)
 
         formats = []
         for format_url in try_get(bitrates, lambda x: x['mp4'], list) or []:
             if isinstance(format_url, compat_str):
                 height = self._search_regex(
-                    r'(\d+)[pP]\.', format_url, 'height', default=None)
+                    r'\-(\d+)(p|)\.', format_url, 'height', default=None)
```
dirkf, line 77: `r'-(\d+)p?\.' ?

youtube_dl/extractor/joj.py
```diff
@@ -104,5 +110,5 @@ def _real_extract(self, url):
             'title': title,
             'thumbnail': thumbnail,
             'duration': duration,
-            'formats': formats,
```
dirkf, line 107: The code typically leaves a trailing `'` in lists, dicts, tuples (even when not required) as it's valid syntax and avoids having to change two lines to insert a new line. You could leave this as-is.

youtube_dl/extractor/joj.py
```diff
@@ -64,16 +66,20 @@ def _real_extract(self, url):
...
         for format_url in try_get(bitrates, lambda x: x['mp4'], list) or []:
             if isinstance(format_url, compat_str):
                 height = self._search_regex(
-                    r'(\d+)[pP]\.', format_url, 'height', default=None)
+                    r'\-(\d+)(p|)\.', format_url, 'height', default=None)
                 formats.append({
                     'url': format_url,
                     'format_id': '%sp' % height if height else None,
                     'height': int(height),
+                    'resolution': quality.get(str(height)+'p')
```
dirkf, line 82: Use `compat_str(height)`.

youtube_dl/extractor/markiza.py
```diff
@@ -1,21 +1,23 @@
...
-    parse_duration,
-    try_get,
+    url_or_none,
+    determine_ext
 )
 
 
 class MarkizaIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?videoarchiv\.markiza\.sk/(?:video/(?:[^/]+/)*|embed/)(?P<id>\d+)(?:[_/]|$)'
+    _VALID_URL = r'https:\/\/(?:www\.)?videoarchiv\.markiza\.sk\/(?:video\/(?:[^\/]+\/)*|embed\/)epizoda\/(?P<id>\d+)(?:[\_\/\-]|$)'
```
dirkf, line 17: * it's not necessary to escape `/`. That's a JS (etc) thing `var regex = /https:\/\/.../;` -- I think it's clear why escaping is needed there;
* `[\_/\-]` can be `[_/-]` (`-` in a regex `[]` must be the last character);
* there's no need to unsupport http://.


Thus:
```
    _VALID_URL = r'https?://(?:www\.)?videoarchiv\.markiza\.sk/(?:video/(?:[^/]+/)*|embed/)epizoda/(?P<id>\d+)(?:[_/-]|$)'
```

youtube_dl/extractor/markiza.py
```diff
@@ -26,54 +28,89 @@ class MarkizaIE(InfoExtractor):
...
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
+        webpage = self._download_webpage(url, video_id)
+        embed = self._search_regex(
+            r'<iframe src="(https:\/\/media.*?)" style="',
+            webpage, 'embed', fatal=False)
+        webpage = self._download_webpage(embed, video_id)
+        data = self._search_regex(
+            r'processAdTagModifier\((\{.*)\), {"video":',
```
dirkf, line 62: The site could randomly insert syntactically valid white space in the target JS expression:
```
            r'processAdTagModifier\s*\(\s*(\{.*)\)\s*,\s*\{\s*"video"\s*:',
```
Also, consider these points:
* `.*` won't match newlines unless you set the `s` flag (`r'(?s)...'`)
* if the site should send several consecutive sequences like this `{...}, {"video: ...}, {...}, {"video: ...}, {...}, {"video: ...`, the pattern will match the entire string and not just the first `{...}, {"video: ...}, `: maybe use `.*?` so as to get the shortest match.

And at l.65 below.

youtube_dl/extractor/markiza.py
```diff
@@ -26,54 +28,89 @@ class MarkizaIE(InfoExtractor):
...
+            webpage, 'embed', fatal=False)
+        data2 = self._search_regex(
+            r'processAdTagModifier\(\{.*\), ({"video":.*)\);',
+            webpage, 'embed', fatal=False)
+        data = re.sub('\\\\/', '/', data)
```
dirkf, line 67: Here and in the next line, if the matches fail, `data`/`data2` will be `None` which isn't a valid string argument. In the `_search_regex()` calls, specify `default=''` (and remove `fatal=False` unless you want to generate a warning whenever the search fails -- I suppose that might be a reasonable thing to do if a failure is going to degrade the extraction).

youtube_dl/extractor/markiza.py
```diff
@@ -26,54 +28,89 @@ class MarkizaIE(InfoExtractor):
...
+        data2 = re.sub('\\\\/', '/', data2)
+        info = json.loads(data)
```
dirkf, line 69: Here and in the next line consider `self._parse_json()` instead of calling `json.loads()` directly.

youtube_dl/extractor/markiza.py
```diff
@@ -26,54 +28,89 @@ class MarkizaIE(InfoExtractor):
...
-        else:
-            info['duration'] = parse_duration(
-                try_get(data, lambda x: x['details']['duration'], compat_str))
-        return info
+        formats = []
+        for format_id, format_list in info['tracks'].items():
+            if not isinstance(format_list, list):
+                format_list = [format_list]
+            for format_dict in format_list:
+                if not isinstance(format_dict, dict):
```
dirkf, line 77: Here (ll.77-9) you can use `try_get()` to simplify and enforce the `format_url`:
```
                format_url = try_get(format_dict, lambda x: url_or_none(x['src']))
                if not format_url:
                    continue
                # now guaranteed: format_dict is a dict-like object, format_url is not None
```

youtube_dl/extractor/markiza.py
```diff
@@ -26,54 +28,89 @@ class MarkizaIE(InfoExtractor):
...
+                        fatal=False))
+                elif (format_type == 'application/dash+xml'
+                      or format_id == 'DASH' or ext == 'mpd'):
+                    formats.extend(self._extract_mpd_formats(
+                        format_url, video_id, mpd_id='dash', fatal=False))
+                else:
+                    formats.append({
+                        'url': format_url,
+                    })
+        thumbnail = info.get('plugins').get('thumbnails').get('url')
```
dirkf, line 96: Use `try_get(info, lambda x: x['plugins']['thumbnails']['url'], compat_str)` to avoid crashing on optional metadata.

youtube_dl/extractor/markiza.py
```diff
@@ -26,54 +28,89 @@ class MarkizaIE(InfoExtractor):
...
-                'id': video_id,
-                'title': try_get(
-                    data, lambda x: x['details']['name'], compat_str),
-            })
-        else:
-            info['duration'] = parse_duration(
-                try_get(data, lambda x: x['details']['duration'], compat_str))
-        return info
+        formats = []
+        for format_id, format_list in info['tracks'].items():
```
dirkf, line 73: Use `(try_get(info, lambda x: x['tracks'], dict) or {}).items()` to avoid crashing.

youtube_dl/extractor/markiza.py
```diff
@@ -26,54 +28,89 @@ class MarkizaIE(InfoExtractor):
...
+                    formats.extend(self._extract_mpd_formats(
+                        format_url, video_id, mpd_id='dash', fatal=False))
+                else:
+                    formats.append({
+                        'url': format_url,
+                    })
+        thumbnail = info.get('plugins').get('thumbnails').get('url')
+        thumbnail = re.sub('$Num$', '001', thumbnail)
+        duration = info.get('duration')
+        title2 = info2.get('video').get('title')
```
dirkf, line 99: Avoid crashing and skip the item on a missing title:
```
        title2 = try_get(info2, lambda x: x['video']['title'], compat_str)
        title = try_get(info2, lambda x: x['video']['custom']['show_title'], compat_str)
        title = ' - '.join(x for x in (title, title2, ) if x)
        if not title:
            continue
        return {
            'id': video_id,
            'title': title,
            ...
```
Add ` or title2` to the second `try_get() line if the value extracted as `title2` would be valid to replace a missing `title`.

You could also generate a diagnostic if this is a "really shouldn't happen" or `raise ExtractorError()` if it's a "can't happen" event.

youtube_dl/extractor/markiza.py
```diff
@@ -87,37 +124,39 @@ class MarkizaPageIE(InfoExtractor):
...
         'only_matching': True,
     }, {
-        'url': 'http://www.tvnoviny.sk/domace/1923887_po-smrti-manzela-ju-cakalo-poriadne-prekvapenie',
+        'url': 'https://www.tvnoviny.sk/domace/1923887_po-smrti-manzela-ju-cakalo-poriadne-prekvapenie',
         'only_matching': True,
     }]
 
     @classmethod
     def suitable(cls, url):
-        return False if MarkizaIE.suitable(url) else super(MarkizaPageIE, cls).suitable(url)
```
dirkf, line 108: The project doesn't enforce a line length and \-continuations are an accident waiting to happen. Consider using `()` if the line is really too long (eg Flake8 complains):
```
        return (
            False if MarkizaIE.suitable(url)
            else super(MarkizaPageIE, cls).suitable(url)
        )
```
Although I don't advise changing the line (a similar expression is used in quite a few other extractors), isn't `not MarkizaIE.suitable(url) and super(MarkizaPageIE, cls).suitable(url)` a shorter equivalent expression?

youtube_dl/extractor/rtvs.py
```diff
@@ -36,12 +43,66 @@ class RTVSIE(InfoExtractor):
     def _real_extract(self, url):
         video_id = self._match_id(url)
 
-        webpage = self._download_webpage(url, video_id)
+        if url.find('/radio/') != -1:
```
dirkf, line 46: `url.split('/')[-2]` is a valid expression, but the test and split can be combined with more assurance that a1, a2 aren't empty:
```
        mobj = re.search(r'(?=.*/radio/).+/([^/]+)/([^/]+)$', url)
        # or if 'radio' shouldn't match the a1/a2 groups
        # mobj = re.search(r'/radio(?:/[^/]+)*/([^/]+)/([^/]+)$', url)
        if mobj:
            # now mobj.groups() is essentially (a1, a2, )
            embed = self._download_webpage(
                'https://www.rtvs.sk/embed/radio/archive/%s/%s' % mobj.groups(),
                video_id)
            ...
```

youtube_dl/extractor/rtvs.py
```diff
@@ -36,12 +43,66 @@ class RTVSIE(InfoExtractor):
...
+            a2 = url.split('/')[-1]
+            a1 = url.split('/')[-2]
+            embed = self._download_webpage(
+                "https://www.rtvs.sk/embed/radio/archive/%s/%s" % (a1, a2),
+                video_id)
+            audio_id = re.search('audio5f.json?id=(?P<id>[^\"]+)', embed)
```
dirkf, line 52: Depending on where in the page the target may be, consider `self._html_search_regex()` which unescapes the returned match (eg, if it contains `&amp;` that should be `&`, or just `&#0049;` that should be `1`).

youtube_dl/extractor/rtvs.py
```diff
@@ -36,12 +43,66 @@ class RTVSIE(InfoExtractor):
...
+        else:
+            info = self._download_json(
+                "https://www.rtvs.sk/json/archive5f.json?id=%s" % video_id,
+                video_id)
+            info = info.get('clip')
 
-        data = self._download_json(
-            playlist_url, video_id, 'Downloading playlist')[0]
-        return self._parse_jwplayer_data(data, video_id=video_id)
+            formats = []
```
dirkf, line 77: Is this the same code as for Markiza? If so, are we looking at some well known framework? In that case we should consider creating a `_parse_whatever_data()` method of `InfoExtractor` that can be used elsewhere.

Issue #28873, opened 2021-04-26
Unable to extract


pepebc:
**Unable to extract**
youtube-dl version: 2021.04.26
Traceback:
_youtube-dl -v -o - "https://www.skylinewebcams.com/es/webcam/italia/lazio/latina/sperlonga-piazza-fontana.html"
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', '-o', '-', 'https://www.skylinewebcams.com/es/webcam/italia/lazio/latina/sperlonga-piazza-fontana.html']
[debug] Encodings: locale UTF-8, fs utf-8, out utf-8, pref UTF-8
[debug] youtube-dl version 2021.04.26
[debug] Python version 3.8.5 (CPython) - Linux-5.4.0-66-generic-x86_64-with-glibc2.29
[debug] exe versions: ffmpeg 4.2.4, ffprobe 4.2.4
[debug] Proxy map: {}
[SkylineWebcams] sperlonga-piazza-fontana: Downloading webpage
ERROR: Unable to extract stream url; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
Traceback (most recent call last):
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 806, in wrapper
    return func(self, *args, **kwargs)
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 827, in __extract_info
    ie_result = ie.extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 534, in extract
    ie_result = self._real_extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/skylinewebcams.py", line 28, in _real_extract
    stream_url = self._search_regex(
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 1012, in _search_regex
    raise RegexNotFoundError('Unable to extract %s' % _name)
youtube_dl.utils.RegexNotFoundError: Unable to extract stream url; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output._

mesosbrodleto:
same issue here, and same version of youtube.dl

[...]

PR #30542, opened 2022-01-20
fix: SkylineWebcams extractor

PieroMacaluso:

- [] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

SkylineWebcams extractor was not working.

Here are the list of changes to make it work:

1. Updated the regex to extract the stream URL from the webpage. In the webpage source code there is just a variable in the form `livee.m3u8?a=<id>` (multiple `e` is correct). The usage of that link by appending it to `https://hd-auth.skylinewebcams.com/` do not work. The solution was to extract just the `?a=<id>` area and appending it to `https://hd-auth.skylinewebcams.com/live.m3u8`. The extractor works properly.

2. Updated `_TEST` with a new `url`. The old one was disabled.

It solves #28873.

### DIFF
--- a/youtube_dl/extractor/skylinewebcams.py
+++ b/youtube_dl/extractor/skylinewebcams.py
@@ -7,7 +7,7 @@ from .common import InfoExtractor
 class SkylineWebcamsIE(InfoExtractor):
     _VALID_URL = r'https?://(?:www\.)?skylinewebcams\.com/[^/]+/webcam/(?:[^/]+/)+(?P<id>[^/]+)\.html'
     _TEST = {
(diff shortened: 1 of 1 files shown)


raspiuser1 (changes requested):
Hi,

its not working because there is a char in line 15 which is not UTF-8
in line 15 the letter à  in word Trinità is not UTF? i changed it to Trinita and now its working

Issue #29915, opened 2021-09-06
[Udemy] ERROR: Unable to download webpage: HTTP Error 403: Forbidden


fhcaglayan:
## Checklist

- [x] I'm reporting a broken site support
- [x] I've verified that I'm running youtube-dl version **2021.06.06**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```

C:\>youtube-dl.exe -u **** -p **** https://www.udemy.com/course/learn-advanced-java/ --verbose
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-u', 'PRIVATE', '-p', 'PRIVATE', 'https://www.udemy.com/course/learn-advanced-java/', '--verbose']
[debug] Encodings: locale cp1252, fs mbcs, out cp437, pref cp1252
[debug] youtube-dl version 2021.06.06
[debug] Python version 3.4.4 (CPython) - Windows-10-10.0.18362
[debug] exe versions: none
[debug] Proxy map: {}
[udemy:course] Downloading login popup
ERROR: Unable to download webpage: HTTP Error 403: Forbidden (caused by HTTPError()); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; type  youtube-dl -U  to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpkqxnwl31\build\youtube_dl\extractor\common.py", line 634, in _request_webpage
  File "C:\Users\dst\AppData\Roaming\Build archive\youtube-dl\ytdl-org\tmpkqxnwl31\build\youtube_dl\YoutubeDL.py", line 2288, in urlopen
  File "C:\Python\Python34\lib\urllib\request.py", line 470, in open
  File "C:\Python\Python34\lib\urllib\request.py", line 580, in http_response
  File "C:\Python\Python34\lib\urllib\request.py", line 508, in error
  File "C:\Python\Python34\lib\urllib\request.py", line 442, in _call_chain
  File "C:\Python\Python34\lib\urllib\request.py", line 588, in http_error_default```
```

## Description

[...]

PR #30573, opened 2022-01-30
Fix forbidden403

dmiraj:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

- Continue downloading nevertheless the HTTP Forbidden403 error.
- I created an exception Forbidden403 which will rise each time the error code returned by compat_urllib_error,HTTPError is equal 403 (compat_urllib_error.HTTPError.code == 403)
- The error will be caught in function YoutubeDl.__handle_extraction_exception.wrapper.
- Once error is caught, it will recurse in wrapper (YoutubeDl.__handle_extraction_exception.wrapper). This makes it attempt to download the webpage a second or more, times, at that same position in the playlist.


pukkandan:
> Fixes #29915 and many others

No it does not! All this will do is retry and fail again

dmiraj:
> > Fixes #29915 and many others
> 
> ~~No it does not! All this will do is retry and fail again~~ (Not even that 🤦 )

I am not too aware how extractors work, but I believe the error happens becomes information extracted becomes outdated. By re-extracting the information again, it will be able to continue downloading where it interrupted. Did you test it?
Me, I am able to download long playlist (+100) songs from YouTube Music.

youtube_dl/compat.py
```diff
@@ -2998,6 +2998,11 @@ def compat_ctypes_WINFUNCTYPE(*args, **kwargs):
         return ctypes.WINFUNCTYPE(*args, **kwargs)
 
 
+# Implement an error for which it is raise when error code 403 HTTPError
```
dirkf, line 3001: If this should be anywhere it should be in `utils.py`.

dirkf:
I tend to agree with the earlier comment.

From https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403:

>The HTTP 403 Forbidden response status code indicates that the server understands the request but refuses to authorize it. ... for the 403 Forbidden status code re-authenticating makes no difference. The access is permanently forbidden and tied to the application logic, such as insufficient rights to a resource.

However, the proof of the pudding is in the eating. If you had (as you claimed) 

>Covered the code with tests (note that PRs without tests will be REJECTED)

that showed this having an effect, that would be great.

Even so, the problem of HTTP 403 occurring during a playlist download may be a feature of relatively few sites and a work-around for those should be a feature of the extractor, or possibly also a site-specific downloader.

If the problem being addressed is that the download information expires before all the downloads in a playlist have been completed, there should be a difference between running the download with and without `--test` (which just downloads 10kB to check that the link is valid). Also `--playlist-start` should allow you to test whether the problem with (say) the 4th playlist item is that 3 previous items have been downloaded, or it's a problem with that item itself.

PR #30618, opened 2022-02-08
[teleportal] Add new extractor for teleportal.ua

prokopiff:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

New extractor for a Ukrainian site that allows to watch TV shows for free.

### DIFF
--- a/youtube_dl/extractor/teleportal.py
+++ b/youtube_dl/extractor/teleportal.py
@@ -1,0 +1,36 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+
+
+class TeleportalIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?teleportal\.ua(/ua)?/(?P<id>[0-9a-z-/]+)'
+    _TEST = {
+        'url': 'https://teleportal.ua/ua/show/stb/master-cheff/bitva-sezonov/vypusk-3',
+        'md5': '07bd056c45b515fa9cc0202b8403df41',
+        'info_dict': {
+            'id': 'show/stb/master-cheff/bitva-sezonov/vypusk-3',
+            'ext': 'mp4',
+            'title': 'МастерШеф. Битва сезонів 3 випуск: найогидніший випуск сезону!',
+            'thumbnail': r're:^https?://.*\.jpg$',
+            'description': r're:^<p>Не пропустіть.*',
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        backend_url = 'https://tp-back.starlight.digital/ua/{}'.format(video_id)
+        metadata = self._download_json(backend_url, video_id)
+        api_metadata = self._download_json('https://vcms-api2.starlight.digital/player-api/{}?referer=https://teleportal.ua/&lang=ua'.format(metadata["hash"]), video_id)
+
+        return {
+            'id': video_id,
+            'title': metadata['title'],
+            'description': metadata['description'],
+            'real_id': metadata['id'],
+            'hash': metadata['hash'],
+            'url': api_metadata['video'][0]['mediaHls'],
+            'thumbnail': api_metadata['video'][0]['poster'],
+            'formats': self._extract_m3u8_formats(api_metadata['video'][0]['mediaHls'], video_id, 'mp4'),
+        }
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/teleportal.py
```diff
@@ -0,0 +1,36 @@
...
+            'thumbnail': r're:^https?://.*\.jpg$',
+            'description': r're:^<p>Не пропустіть.*',
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        backend_url = 'https://tp-back.starlight.digital/ua/{}'.format(video_id)
+        metadata = self._download_json(backend_url, video_id)
+        api_metadata = self._download_json('https://vcms-api2.starlight.digital/player-api/{}?referer=https://teleportal.ua/&lang=ua'.format(metadata["hash"]), video_id)
```
camel2314, line 25: mixing quotes and double quotes

youtube_dl/extractor/teleportal.py
```diff
@@ -0,0 +1,36 @@
...
+
+        return {
+            'id': video_id,
+            'title': metadata['title'],
+            'description': metadata['description'],
```
camel2314, line 30: should use `metadata.get('description')` - see https://github.com/ytdl-org/youtube-dl#mandatory-and-optional-metafields
for optional.
for mandatory fields like `title` I guess the same approach would work and the caller is supposed to check their values.

prokopiff: Updated. It's fine if formats fail, because it's required, right?

youtube_dl/extractor/teleportal.py
```diff
@@ -0,0 +1,40 @@
...
+            thumbnail = None
+
+        return {
+            'id': video_id,
+            'title': metadata.get('title'),
+            'description': metadata.get('description'),
+            'real_id': metadata.get('id'),
+            'hash': metadata.get('hash'),
+            'thumbnail': thumbnail,
+            'formats': self._extract_m3u8_formats(api_metadata['video'][0]['mediaHls'], video_id, 'mp4'),
```
dirkf, line 39: Better to do this first, at your line 26 (I've added comments to explain that wouldn't be in the final code):
```
        title = metadata['title']  # mandatory field
        fmt_url = try_get(api_metadata, lambda x: x['video'][0]['mediaHls'])
        formats = []
        if fmt_url:
            formats.extend(self._extract_m3u8_formats(fmt_url, video_id, 'mp4')
        self._check_formats(formats)  #  include this line if the URLs in the m3u8 might sometimes fail (403, 404, eg)
        self._sort_formats(formats)  #  will raise exception if no formats
```

youtube_dl/extractor/teleportal.py
```diff
@@ -0,0 +1,40 @@
...
+        try:
+            thumbnail = api_metadata['video'][0]['poster']
+        except (KeyError, IndexError):
+            thumbnail = None
+
+        return {
+            'id': video_id,
+            'title': metadata.get('title'),
+            'description': metadata.get('description'),
+            'real_id': metadata.get('id'),
```
dirkf, line 36: This isn't a standard field: there's no prohibition against non-standard fields, but you could use `'display_id'`. `'hash'` doesn't have an obvious equivalent.

Are any of the other standard fields available from the main page (I know it would have to be downloaded) or the two metadata dicts?

youtube_dl/extractor/teleportal.py
```diff
@@ -0,0 +1,40 @@
...
+            'thumbnail': r're:^https?://.*\.jpg$',
+            'description': r're:^<p>Не пропустіть.*',
+        }
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        backend_url = 'https://tp-back.starlight.digital/ua/{}'.format(video_id)
+        metadata = self._download_json(backend_url, video_id)
+        api_metadata = self._download_json('https://vcms-api2.starlight.digital/player-api/{}?referer=https://teleportal.ua/&lang=ua'.format(metadata['hash']), video_id)
```
dirkf, line 25: Above you set the url in a var separately. Stick to one or the other: it's probably better to put it in two lines so that it's easy to see which line might have failed. So let's try something like this:
```
        api_hash = metadata['hash']
        api_metadata_url = 'https://vcms-api2.starlight.digital/player-api/{0}?referer=https://teleportal.ua/&lang=ua'.format(api_hash)
        api_metadata = self._download_json(api_metadata_url, video_id)
```
Also, after the first line succeeds, it's safe to use `metadata.get()` later.

dirkf (changes requested):
Thanks for the work.

PR #30629, opened 2022-02-10
Remove extractors for dead pages keezmovies, extremetube, mofosex, fix tube8

marieell:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

keezmovies, extremetube, mofosex are dead and tube8 extended keezmovies and was broken.


youtube_dl/extractor/tube8.py
```diff
@@ -37,44 +39,133 @@ def _extract_urls(webpage):
...
+
+        if not title:
+            title = self._html_search_regex(
+                r'<h1[^>]*>([^<]+)', webpage, 'title')
+
+        return webpage, {
+            'uploader': uploader,
+            'id': video_id,
+            'display_id': display_id,
+            'title': strip_or_none(title),
```
dirkf, line 132: `title` could be stripped to `None`: provide an alternative title, such as the last-but-one element of the URL path, capitalised (`... or url.rsplit('/', 2)[1].capitalize()`).

youtube_dl/extractor/tube8.py
```diff
@@ -37,44 +39,133 @@ def _extract_urls(webpage):
             r'<iframe[^>]+\bsrc=["\']((?:https?:)?//(?:www\.)?tube8\.com/embed/(?:[^/]+/)+\d+)',
             webpage)
 
+    def _extract_info(self, url, fatal=True):
+        mobj = re.match(self._VALID_URL, url)
+        video_id = mobj.group('id')
+        display_id = (mobj.group('display_id')
```
dirkf, line 45: How about:
```
       display_id = mobj.groupdict.get('display_id') or mobj.group('id')
```

youtube_dl/extractor/tube8.py
```diff
@@ -37,44 +39,133 @@ def _extract_urls(webpage):
...
 
         category = self._search_regex(
-            r'Category:\s*</dt>\s*<dd>\s*<a[^>]+href=[^>]+>([^<]+)',
+            r'videoCategoryByName\s*=\s*"([^"]+)";',
             webpage, 'category', fatal=False)
         categories = [category] if category else None
 
         tags_str = self._search_regex(
-            r'(?s)Tags:\s*</dt>\s*<dd>(.+?)</(?!a)',
+            r"(?s)<li class='video-tag' data-esp-node='tag'>(.*?)</div>",
```
dirkf, line 163: Use `\s+` for spaces in the pattern. Also possibly `\s*=\s*`.

marieell:
`_extract_info` is a verbatim copy from `KeezmoviesIE`, the previous superclass.

dirkf:
OK, I understand, but I hope you won't mind if I apply those changes before merging anyway.

marieell:
Not at all.

youtube_dl/extractor/tube8.py
marieell: `display_id` is never even empty

youtube_dl/extractor/tube8.py
marieell: There is a `fatal=True` fallback in `_real_extract` for `title`.

dirkf: Sure, but `strip_` **or_none**. The regex matches a whitespace title --> `None`.

marieell: Better like this?

its0x08 (approved):

Issue #30641, opened 2022-02-15
Multi-threading slowdown for YouTube


noembryo:
## Checklist

- [ ] I'm reporting a broken site support
- [x] I've verified that I'm running youtube-dl version **2021.12.17**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar issues including closed ones

## Verbose log

```
PASTE VERBOSE LOG HERE
```

## Description

Using `youtube_dl` with multiple threads to get information about multiple videos, is a lot slower after the last breakage.
Using the code below gives me ~70 sec for 42 videos.
With the `yt-dlp` the time it takes for the same videos is ~30 sec.
Before the last code change (I'm using the current git code), `youtube_dl` was faster than `yt-dlp`.
Changing the `THREAD_NR` didn't change the difference..

The processing is also much higher than before.
Because I use a similar strategy in an app I have, in my system (with an older i7) it got noticeable worst than before.
Trying it with an older laptop, made the app totally unusable..

The ids are some random links, you can use whatever you like.
```python
# coding=utf-8
from __future__ import absolute_import, division, print_function, unicode_literals
from queue import Queue
from threading import Thread, Event
import youtube_dl
# import yt_dlp as youtube_dl

ydl_opts = {"quiet": True, "no_warnings": True}
link_ids = ["4jduuQh-Uho", "9GNpv7QDvMY", "MbEOR2Flc-4", "ZKUzNF21n9w", "y-JqH1M4Ya8",
            "pUqfaiUb3l4", "bL5eqSOXMtE", "HyMm4rJemtI", "BU4kGkrrJEw", "wA1v207xlOw",
            "pFS4zYWxzNA", "aF6hDcAbSoE", "G1ckKDRc69w", "o9_jzBtdMZ4", "AGoQZx8Mn0g",
            "6W-pHCD6Tow", "kszLwBaC4Sw", "mwTd_PzGY-c", "iqLTYD_nhsU", "X335gdcPE7A",
            "z_54vDk8lWw", "8a82arE0JSQ", "tJmzQHWl9kc", "8jPQjjsBbIc", "ENJUB5thpB4",
            "dEhUMvjFuQY", "D6XyJh1tsGI", "tFCfb-Qqdz0", "UkafA6r1caQ", "OO8HtAXnRqQ",
[...]

PR #30188, opened 2021-11-02
[JSInterp] Back-port JS interpreter upgrade from yt-dlp PR #1437

dirkf:
## Please follow the guide below

---

### Before submitting a *pull request* make sure you have:
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [x] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [x] Covered the code with tests (note that PRs without tests will be REJECTED)
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [ ] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [x] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

As part of the PR "[[youtube] Fix throttling by decrypting n-sig](https://github.com/yt-dlp/yt-dlp/pull/1437)", @pukkandan implemented a more capable version of the JS interpreter module. Although it's not needed for the [equivalent fix in yt-dl](https://github.com/ytdl-org/youtube-dl/pull/30184), having the same level of capability in yt-dl would facilitate other back-ports, eg in case a more complex YouTube challenge should have to be decoded in future. It might also suffice for certain other extractors where some JS trickery needs to be emulated.

This PR back-ports the `jsinterp.py` from https://github.com/yt-dlp/yt-dlp/pull/1437 with changes to make it run in 2.6 <= Python < 3.6 as well.
* `yield from` becomes nested iterator
* `nonlocal` faked by a dummy namespace class
* f'...' replaced by various older format syntaxes
* replace `collections.abc` with `compat_collections_abc` added to `compat.py`.

All the tests in the unchanged back-ported `test/test_jsinterp.py` pass.

The `diff` below applies to the initial commit 7109df0.
<details>
[...]

PR #30184, opened 2021-11-01
[YouTube] Unthrottle downloads by responding to the "n" parameter challenge

dirkf:
## Please follow the guide below

---

### Before submitting a *pull request* make sure you have:
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [x] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [x] Covered the code with tests (note that PRs without tests will be REJECTED)
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

Since summer 2021, YouTube has started serving media URLs with a query parameter such as `...&n=SXiXBH-xzrjeioPN&...`. This now appears to be the default behaviour. Unless the value of this parameter is transformed according to an algorithm delivered in the site player JS, the download speed for the URL is throttled to ~50kB/s.

Solutions for this include:
* implementing a more complete JS interpreter, eg based on PR #11272;
* using an external JS interpreter (PhantomJS, now unmaintained, is used by some extractors);
* spoofing the Android or iOS client to acquire unthrottled links, as successfully implemented in [yt-dlp](https://github.com/yt-dlp/yt-dlp/commit/c888ffb95ab0ab4f4cd1d6c93eda014f80479551).

This PR now uses the first approach taking the enhanced JSInterpreter module from [yt-dlp PR #1437](https://github.com/yt-dlp/yt-dlp/pull/1437), as back-ported in PR #30188.

[...]

PR #30643, opened 2022-02-17
[JSInterpreter] Improve performance

dirkf:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

The limited JS interpreter used to decode YT signature URLs was enhanced in PR #30188, back-porting yt-dlp PR 1437, and merged from PR #30184 in order to respond to the n parameter challenge introduced by YT.

In #30641, a specific embedded application of yt-dl identified significantly reduced extraction performance compared with previous versions and even the stopgap n-parameter response.

A test application was profiled. After pulling out some constant regex expressions and changing a loop over assignment operators into a regex alternative match, the execution time (Py 2.7) went from 22s to 18s. The number of function calls was down by almost 1M, nearly 25% (application extracting 42 YT pages).

The regex optimisations apply to expressions that used f-strings in the yt-dlp version, which are said to be faster than formatting operations available to yt-dl; however, the loop optimisation, and one already made in the back-port, could equally apply to the yt-dlp version.

Currently constants used in the `JSInterpreter` class include both global and class vars:
* globals: `_OPERATORS`, `_ASSIGN_OPERATORS`, `_NAME_RE`, `_MATCHING_PARENS`
* class vars: `_EXPR_SPLIT_RE`, `_VARNAME_RE`, `_ARRAY_REF_RE`, `_FN_CALL_RE`, `_MEMBER_REF_RE`, `_FN_NAME_RE`, `_FN_DEF_RE`, `_ASSIGN_EXPR_RE`.

As to whether the constant regex expressions used for parsing should be class or global vars, probably the following heuristics should apply:
* constants that are characteristics of the JS language should be _vars of the module or the class
* constants that are specific to the internal workings of the JS interpreter should be __vars of the class.

### DIFF
--- a/youtube_dl/jsinterp.py
+++ b/youtube_dl/jsinterp.py
@@ -41,14 +41,19 @@ _NAME_RE = r'[a-zA-Z_$][a-zA-Z_$0-9]*'
 _MATCHING_PARENS = dict(zip(*zip('()', '{}', '[]')))
 
 
-class JS_Break(ExtractorError):
-    def __init__(self):
-        ExtractorError.__init__(self, 'Invalid break')
+class JSError(ExtractorError):
+    def __init__(self, msg):
+        super(JSError, self).__init__('JS: ' + msg)
 
 
-class JS_Continue(ExtractorError):
+class JS_Break(JSError):
     def __init__(self):
-        ExtractorError.__init__(self, 'Invalid continue')
+        super(JS_Break, self).__init__('Invalid break')
+
+
+class JS_Continue(JSError):
+    def __init__(self):
+        super(JS_Continue, self).__init__('Invalid continue')
 
 
 class LocalNameSpace(MutableMapping):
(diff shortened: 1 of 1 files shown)


Lee-Carre:
What about (optionally, if installed on the system) an external interpreter. Thus, youtube-dl could be simple and do only the basics, but if intensive users want more features or more performance, then then could ensure that suitable tools are available to youtube-dl.

dirkf:
At the moment PhantomJS, which was used by some extractors and is a potential dependency of of yt-dlp, similar to what you suggest, is unmaintained ("discontinued" is what Wikipedia says); owing to undisciplined feature creep a JS interpreter that worked last year will fail next year (for year in (2015, 2018, 2020)).

I had a look for other Python JS implementations, but didn't come up with anything that was (a) supported (b) compatible with Unlicense (c) compatible with yt-dl platforms; in fact I don't think any of those were satisfied.

So I think it's best is to stick with the "toy" interpreter and extend it when required.

Lee-Carre:
> I think it's best is to stick with the "toy" interpreter and extend it when required.

Good points. I suppose a dependency is only a good idea when the benefit it brings exceeds the cost of using it.

Pity that something so obvious is lacking in the libre world. Reminds me of `wget`; highly depended upon, but widely neglected (though, maintained enough to be useful).

If only the interpreters of browsers were available as standalone modules.

At least with rolling-your-own, you have control without scope-creep.

Kudos for improving the ‘toy’.

> PhantomJS, which was used by some extractors and is a potential dependency of of yt-dlp, similar to what you suggest, is unmaintained ("discontinued" is what Wikipedia says)

That certainly was the case, for a while. The [source repo](/ariya/phantomjs) seems to have been revived (no longer archived), with relatively recent (mid-2020) commits.

As usual, Wikipedia is lacking in factual accuracy.

dirkf:
>Kudos for improving the ‘toy’.

Pukkandan did the hard part.

Issue #30656, opened 2022-02-21
Unrecognized option 'x'


gbr000:
Two issues going on the first for a long time:

- `youtube-dl`, not able to download youtube stories anymore (a couple of months now)
- website Crunchyroll when listing with `youtube-dl -F`, `ffmpeg 5.0` throw two errors:
    - `Unrecognized option 'x'.` (which I have none in my scripts, I checked my script)
    - `WARNING: Unable to download XML: HTTP Error 404: Not Found`

I'm using Android 11, in `termux`, everything is updated!

The full output is here:

```bash
youtube-dl https://www.crunchyroll.com/blade-runner-black-lotus/episode-2-all-we-are-not-821199 -v
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-iwc', '-f', '247+251/(webm,mp4)[vcodec=vp9][height<=720]+bestaudio/best', '--external-downloader', 'aria2c', '--external-downloader-args', '-c -j 3 -x 3 -s 3 -k 1M', 'https://www.crunchyroll.com/blade-runner-black-lotus/episode-2-all-we-are-not-821199', '-v']
[debug] Encodings: locale UTF-8, fs utf-8, out utf-8, pref UTF-8
[debug] youtube-dl version 2021.12.17
[debug] Python version 3.10.2 (CPython) - Linux-4.4.177-22825772-aarch64-with-libc
[debug] exe versions: ffmpeg 5.0, ffprobe 5.0
[debug] Proxy map: {}
[crunchyroll] 821199: Downloading webpage
[crunchyroll] 821199: Downloading adaptive_hls-audio-jaJP-hardsub-itIT information
[crunchyroll] 821199: Downloading adaptive_hls-audio-jaJP-hardsub-deDE information
[crunchyroll] 821199: Downloading adaptive_hls-audio-jaJP-hardsub-ruRU information
[crunchyroll] 821199: Downloading adaptive_hls-audio-jaJP-hardsub-frFR information
[crunchyroll] 821199: Downloading adaptive_hls-audio-jaJP-hardsub-ptBR information
[crunchyroll] 821199: Downloading adaptive_hls-audio-jaJP-hardsub-esLA information
[crunchyroll] 821199: Downloading adaptive_hls-audio-jaJP-hardsub-arME information
[crunchyroll] 821199: Downloading adaptive_hls-audio-jaJP-hardsub-enUS information
[...]

Issue #2773, opened 2014-04-19
Unsupported site 'clubic.com' to download 'clubic-week'


skyghis:
```
http://www.clubic.com/video/clubic-week/video-clubic-week-2-0-le-fbi-se-lance-dans-la-photo-d-identite-448474.html
```

Hi, not sure if it's the good location or format to post this, but when I try to download the video from given URL, youtube-dl get me an empty ~700Ko swf file, I expect the ~60Mo mp4 file.

```
$ youtube-dl --version
2014.04.19
```

Thanks in advance.

PR #30664, opened 2022-02-22
[core] Ignore --external-downloader-args if --external-downloader was rejected

dirkf:
---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

In issue #30656 it was found that the yt-dl command failed because:
* the command used the option `--external-downloader aria2c`
* aria2c options were specified with `--external-downloader-args`
* yt-dl determined that aria2c was not able to download the specific target stream, and selected ffmpeg instead
* the aria2c options were **nonetheless passed to ffmpeg**.

As there is no significant commonality of options between possible external downloaders, any options specified with `--external-downloader-args` should only be used with the specified `--external-downloader`.

This PR applies a two-stage safeguard to ensure this:
* in `downloader/__init__.py`, the `external_downloader_args` of the `params` dict is nulled if it has an `external_downloader` that is rejected by `get_suitable_downloader()`;
* in `YoutubeDL.py`, the call to `get_suitable_downloader()` is wrapped to generate a warning in that case, since the function itself has no instance context from which to do so.

The PR does not affect the case where the user prophesies that yt-dl will invoke an external downloader (ffmpeg) and passes options for it with `--external-downloader-args`.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1906,8 +1906,17 @@ class YoutubeDL(object):
 
         if not self.params.get('skip_download', False):
             try:
+                def checked_get_suitable_downloader(info_dict, params):
+                    ed_args = params.get('external_downloader_args')
+                    dler = get_suitable_downloader(info_dict, params)
+                    if ed_args and not params.get('external_downloader_args'):
+                        # external_downloader_args was cleared because external_downloader was rejected
+                        self.report_warning('Requested external downloader cannot be used: '
+                                            'ignoring --external-downloader-args.')
+                    return dler
+
                 def dl(name, info):
-                    fd = get_suitable_downloader(info, self.params)(self, self.params)
+                    fd = checked_get_suitable_downloader(info, self.params)(self, self.params)
                     for ph in self._progress_hooks:
                         fd.add_progress_hook(ph)
                     if self.params.get('verbose'):
--- a/youtube_dl/downloader/__init__.py
+++ b/youtube_dl/downloader/__init__.py
@@ -50,6 +50,9 @@ def _get_suitable_downloader(info_dict, params={}):
         ed = get_external_downloader(external_downloader)
         if ed.can_download(info_dict):
             return ed
+        # Avoid using unwanted args since external_downloader was rejected
+        if params.get('external_downloader_args'):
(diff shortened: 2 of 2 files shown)


pukkandan:
Another solution would be to implement a syntax like #27723 for `--external-downloader-args`

dirkf:
Yes, that might be Phase 2. Would #2773 be better than pulling the implementation from yt-dlp?

pukkandan:
You can take code directly from yt-dlp too. Other than some refactoring, the main difference b/w the implementations is the ability to pass arguments to a specific position for ffmpeg

Here's the documentation from yt-dlp for reference:
```
    --postprocessor-args NAME:ARGS   Give these arguments to the postprocessors.
                                     Specify the postprocessor/executable name
                                     and the arguments separated by a colon ":"
                                     to give the argument to the specified
                                     postprocessor/executable. Supported PP are:
                                     Merger, ModifyChapters, SplitChapters,
                                     ExtractAudio, VideoRemuxer, VideoConvertor,
                                     Metadata, EmbedSubtitle, EmbedThumbnail,
                                     SubtitlesConvertor, ThumbnailsConvertor,
                                     FixupStretched, FixupM4a, FixupM3u8,
                                     FixupTimestamp and FixupDuration. The
                                     supported executables are: AtomicParsley,
                                     FFmpeg and FFprobe. You can also specify
                                     "PP+EXE:ARGS" to give the arguments to the
                                     specified executable only when being used
                                     by the specified postprocessor.
                                     Additionally, for ffmpeg/ffprobe, "_i"/"_o"
                                     can be appended to the prefix optionally
                                     followed by a number to pass the argument
                                     before the specified input/output file. Eg:
                                     --ppa "Merger+ffmpeg_i1:-v quiet". You can
                                     use this option multiple times to give
                                     different arguments to different
                                     postprocessors. (Alias: --ppa)
    --downloader-args NAME:ARGS      Give these arguments to the external
                                     downloader. Specify the downloader name and
                                     the arguments separated by a colon ":". For
                                     ffmpeg, arguments can be passed to
                                     different positions using the same syntax
                                     as --postprocessor-args. You can use this
                                     option multiple times to give different
                                     arguments to different downloaders (Alias:
                                     --external-downloader-args)
```

PR #30703, opened 2022-02-28
[core] Add option --get-manifest-url

### LINKED ISSUE #10976: [feature] manifest
is there a command to make it straight forward to find out the link for manifest?

uhle:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [x] New feature

---

For every item in the info dictionary, there is a respective `--get-*` option except for the `manifest_url`. This commit adds the new option `--get-manifest-url` simply in the same manner as for all the other `--get-*`options before. This new option can be used as you would use `--get-url` for instance but it would return the manifest URL instead of the video/audio URL.

Resolves #10976.

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,6 +301,7 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
                                          write anything to disk
     --skip-download                      Do not download the video
     -g, --get-url                        Simulate, quiet but print URL
+    --get-manifest-url                   Simulate, quiet but print manifest URL
     -e, --get-title                      Simulate, quiet but print title
     --get-id                             Simulate, quiet but print id
     --get-thumbnail                      Simulate, quiet but print thumbnail URL
--- a/test/parameters.json
+++ b/test/parameters.json
@@ -4,6 +4,7 @@
     "forcedescription": false, 
     "forcefilename": false, 
     "forceformat": false, 
+    "forcemanifesturl": false, 
     "forcethumbnail": false, 
     "forcetitle": false, 
     "forceurl": false, 
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -151,6 +151,7 @@ class YoutubeDL(object):
     quiet:             Do not print messages to stdout.
     no_warnings:       Do not print out anything for warnings.
     forceurl:          Force printing final URL.
+    forcemanifesturl:  Force printing manifest URL.
     forcetitle:        Force printing title.
     forceid:           Force printing ID.
     forcethumbnail:    Force printing thumbnail URL.
@@ -1763,6 +1764,12 @@ class YoutubeDL(object):
             else:
                 # For RTMP URLs, also include the playpath
                 self.to_stdout(info_dict['url'] + info_dict.get('play_path', ''))
+        if self.params.get('forcemanifesturl', False) and not incomplete:
+            if info_dict.get('requested_formats') is not None:
+                for f in info_dict['requested_formats']:
+                    self.to_stdout(f['manifest_url'])
+            else:
+                self.to_stdout(info_dict['manifest_url'])
         print_optional('thumbnail')
         print_optional('description')
         if self.params.get('forcefilename', False) and filename is not None:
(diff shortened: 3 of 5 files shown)


pukkandan:
There are a lot of fields in the infodict that doesn't have an equivalent `--get...` option. imo, it doesn't make sense to add such an option for each field. It would be more usefull to have a generic option to print whatever field (yt-dlp has `--print`).

Or, the user could just pluck the data from the `-j` output using something like jq. `yt-dlp URL -j |jq .manifest_url`

dirkf:
Thanks for the PR, but I agree with @pukkandan.

I'd be happy to merge a back-port of `--print`. However as the same thing can be achieved with `--get-filename -o '%(manifest_url)s'` (in this case), I would say that there are higher priorities.

Perhaps the above work-around (and/or use of _jq_, _[gron](https://github.com/tomnomnom/gron)_) should be a FAQ answer?

pukkandan:
> the same thing can be achieved with --get-filename -o '%(manifest_url)s' (in this case)

`--get-filename` sanitizes the field as-if it were a filename. So won't work

dirkf:
Yes, not for `manifest_url` (or any field that needs to be reproduced exactly) in general, I suppose. All the more reason for `--print`, then.

uhle:
Thank you both for the feedback.

I wasn't aware of the fork [yt-dlp](/yt-dlp/yt-dlp) and its capability to extract and get anything printed from the info dictionary. This is a nice feature indeed. And I can understand why you prefer a more generic approach.

Using `jq` as a work-around to filter out `manifest_url` was my "option" in the past. But that is not really always a good option and I don't want to elaborate why it's not.
Anyway, I remember situations when a simple `youtube-dl -j <url> | jq .manifest_url` would not do the work because there was no `manifest_url` at top-level but nested inside the array `requested_formats`. Then something like `youtube-dl -j <url> | jq '.requested_formats[0].manifest_url'` was needed. That was also one reason why I decided to create this PR and implement an additional option `--get-manifest-url` in much the same way as `--get-url`.
Unfortunately, I do not exactly remember those situations (i.e. from which website and under what circumstances). So I could not try but I guess that `yt-dlp --print '%(manifest_url)s' <url>` would fail as well in such a situation.
Today I spent quite some time to dive into the source code of `yt-dlp` (as it differs quite a lot from `youtube-dl`) to get an answer to this assumption and I found out that `url` is still specially handled, i.e. in the method `__forced_printings`, a local copy of `info_dict` gets a new entry with key `urls` and a value from what would be the concatenated output from `requested_formats`. So if you would use something like `--print '%(url)s'`, the value of `urls` is actually output instead of the value from a possibly (non-)existing `url` at top-level. As far as I can see, it is the only entry in `info_dict` that is handled in such a way. But `manifest_url` at least would need much the same special handling.

dirkf:
The processing for url/urls is to replace `--get-url`. For `manifest_url`, yt-dlp expects you to find it with an index path, if it's not at the top level.

error-reporting:
> The processing for url/urls is to replace `--get-url`. For `manifest_url`, yt-dlp expects you to find it with an index path, if it's not at the top level.

what's the index path I need to use to find manifest_url in yt-dlp

dirkf:
It'll be in whichever of `.requested_formats` you want, as in https://github.com/ytdl-org/youtube-dl/pull/30703#issuecomment-1063452989.

error-reporting:
can I get manifest_url of https formats

dirkf:
It'll be present for manifest formats and not for other protocols. Load the `-j` output into a JSON folding viewer to see, and/or review the extractor code.

error-reporting:
when there is more than one url in .requested_format in yt-dlp how do I select just one

dirkf:
For _yt-dlp_, read the manual there and raise a new issue if necessary/appropriate. For _yt-dl_, `man jq` (or equivalent).

error-reporting:
thanks
> *dirkf* for helping and guide me how-to getting the desired direct url

16 thg 2, 2025 19:07:50 dirkf ***@***.***>:

PR #30743, opened 2022-03-14
modify installation instructions for pip

MrRawes:


dirkf:
Should we also mention using _pip3_ instead of _pip_, or perhaps `python -m pip`? See also https://adamj.eu/tech/2019/03/11/pip-install-from-a-git-repository/.

MrRawes:
i agree (modified title and readme for the PR)

MrRawes

------- Original Message -------

rautamiekka:
For as long ytdl supports Py <3.x.x, the command can't be changed to Python 3.

MrRawes:
> For as long as ytdl supports Py <3.x.x, the command can't be changed to Python 3.

reverted that

dirkf:
Owing to the widespread apartheid policy where Python 3 gets installed as `python3` with no `python` alias, we should nonetheless mention that the user may need to say `python3` instead.

MrRawes:
added note

------- Original Message -------

MrRawes:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

i added a note
-----BEGIN PGP SIGNATURE-----
Version: ProtonMail

wnUEARYKAAYFAmI/hfAAIQkQ3GvPpUcI+NAWIQQzFMf/Dg/a/bqvGr/ca8+l
Rwj40DA5AQCpglIlNLOOAy2qUW4/qWm5QadWoz+G0sID7OELTMGn3QD9ExND
qqHJmckeO6og9dRyvwX2QtuSgx1cHB9qv7gk1QA=
=t3D4
-----END PGP SIGNATURE-----

PR #30763, opened 2022-03-22
[postprocessors] Added progress hook calls at the end of postprocessing

kerrynich:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This adds progress hooks to the post processors in addition to where they are already called in the downloaders. The specific use case this resolves is getting the finalized filename and final file size after merging or other post processing occurs. A new status option was added to the progress hooks called `postprocessed`.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -234,11 +234,13 @@ class YoutubeDL(object):
                        postprocessor.
     progress_hooks:    A list of functions that get called on download
                        progress, with a dictionary with the entries
-                       * status: One of "downloading", "error", or "finished".
+                       * status: One of "downloading", "error", "finished",
+                                 or "postprocessed".
                                  Check this first and ignore unknown values.
 
-                       If status is one of "downloading", or "finished", the
-                       following properties may also be present:
+                       If status is one of "downloading", "finished", or
+                       "postprocessed", the following properties may also be
+                       present:
                        * filename: The final filename (always present)
                        * tmpfilename: The filename we're currently writing to
                        * downloaded_bytes: Bytes on disk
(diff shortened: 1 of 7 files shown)


pukkandan:
If you only want the info at the end of ALL processing, you can register a custom `PostProcessor` class instead

```py
import youtube_dl

class MyCustomPP(youtube_dl.postprocessor.PostProcessor):
    def run(self, info):
        self.to_screen('Doing stuff')
        return [], info

with youtube_dl.YoutubeDL() as ydl:
    ydl.add_post_processor(MyCustomPP())
    ydl.download('https://www.youtube.com/watch?v=BaW_jenozKc')
```

Or, if a hook is to be added, I would suggest a separate `postprocessor_hooks`. There are a couple of reasons:
1. You may want hooks at the start of PPs as well. In future, we may wish to send even actual progress updates (eg: https://github.com/yt-dlp/yt-dlp/pull/2475)
2. Most of the fields of `progress_hook` doesn't make much sense for PPs. Having a separate param would allow us to define/document new fields

btw, defining a metaclass/wrapper is a much more elegant way to "install" the hooks ([example](https://github.com/yt-dlp/yt-dlp/blob/ea5ca8e7fcf348a40e3eda70738738739123bded/yt_dlp/postprocessor/common.py#L15-L32)). Generally, wrappers are prefered for such uses. But in this case, metaclass has the advantage that it will automatically work for any custom PPs third parties may add

PS: There is also some personal reason why I'd like to see a separate hook. https://github.com/yt-dlp/yt-dlp has already implemented `postprocessor_hooks` (that works like I described above). So there is no reasonable way for yt-dlp to port this PR, making this the first place where yt-dlp's API significantly diverges from youtube-dl. This is bound to happen eventually, but I would like to push that as far back as possible

kerrynich:
> https://github.com/yt-dlp/yt-dlp

Thanks for the detailed response. I agree with you about the postprocessor hooks and the metaclass design. My original intention was to keep the changes as minimal as possible while sticking to the design implemented in the downloaders for progress hooks. Are you of the opinion that the PR should be updated to have functionality more closely mimicking yt-dlp or should this just be paused/scrapped for now?

pukkandan:
> Are you of the opinion that the PR should be updated to have functionality more closely mimicking yt-dlp or should this just be paused/scrapped for now?

I am obv in favor following yt-dlp's design, but it is ultimately @dirkf's decision

PR #30791, opened 2022-03-29
[nbl] Add new extractor

alyssadev:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This is an extractor for game VODs, condensed game VODs, and highlight clips from the website of the National Basketball League of Australia (https://nbl.com.au). At this point it supports video watch URLs at the path /tv/[category]/[id]/. All tests are passing on 2.7.18 and 3.9.11 (I wasn't able to run 2.6.* on my dev machine).

I had plans to include the latter part of the url as the display_id, but I must be misunderstanding the named capture groups in _VALID_URL, as this does not seem to be getting added to the info dict.

I'd like to support the embed url format as well, e.g https://ott.nbl.com.au/en-int/embed/1303323, as this is used in og:video tags and might be a simplification for the extractor, but I didn't want to add that functionality at this time with my limited knowledge of what's available in ytdl.

### DIFF
--- a/youtube_dl/extractor/nbl.py
+++ b/youtube_dl/extractor/nbl.py
@@ -1,0 +1,69 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import unified_strdate
+
+
+class NBLIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?nbl\.com\.au/tv/(?:game-replays|highlights|condensed-games)/(?P<id>[0-9]+)'  # /(?P<display_id>.+)'
+    _TESTS = [
+        {
+            'url': 'https://nbl.com.au/tv/highlights/1310086/Perth-Wildcats-vs.-Sydney-Kings---Game-Highlights',
+            'md5': '684a1126879371c3137850b8474ae3c8',
+            'info_dict': {
+                'id': '1310086',
+                'ext': 'mp4',
+                'title': 'Perth Wildcats vs. Sydney Kings - Game Highlights',
+                'description': 'Watch the Game Highlights from Perth Wildcats vs. Sydney Kings, 03/26/2022',
+                'upload_date': '20220326'
+            }
+        },
+        {
+            'url': 'https://nbl.com.au/tv/condensed-games/1310087/Perth-Wildcats-vs.-Sydney-Kings---Condensed-Game',
+            'md5': '1f9ac9ea04dc4024e50b27593860a782',
+            'info_dict': {
+                'id': '1310087',
+                'ext': 'mp4',
+                'title': 'Perth Wildcats vs. Sydney Kings - Condensed Game',
+                'description': 'Watch the Condensed Game from Perth Wildcats vs. Sydney Kings, 03/26/2022',
+                'upload_date': '20220326'
+            }
+        },
+        {
+            'url': 'https://nbl.com.au/tv/game-replays/1303323/NBL22-Round-17-Replay---Perth-Wildcats-vs-Sydney-Kings',
+            'md5': '8505f02156f756e865a1aa80050eb768',
+            'expected_warnings': ['unable to extract OpenGraph description.+'],
+            'info_dict': {
+                'id': '1303323',
+                'ext': 'mp4',
+                'title': 'NBL22 Round 17 Replay - Perth Wildcats vs Sydney Kings',
+                'upload_date': '20220326'
+            }
+        }
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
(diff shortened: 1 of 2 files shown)


alyssadev:
One concern I had was disambiguation, as obviously there's other organisations with the initials NBL. The Australian league seems to be the most prominent one and other regions' instances of NBL seem to have more fitting initialisms, e.g NBLC. Of course there's no problem with renaming the extractor to avoid conflicts.

dirkf:
> One concern I had was disambiguation, as obviously there's other organisations with the initials NBL. ...

In this context, and I know little of basketball, it seems that the name you chose is uncontroversial, as the former US NBL won't be streaming anything from 70 years ago.

youtube_dl/extractor/nbl.py
```diff
@@ -0,0 +1,69 @@
...
+                'id': '1303323',
+                'ext': 'mp4',
+                'title': 'NBL22 Round 17 Replay - Perth Wildcats vs Sydney Kings',
+                'upload_date': '20220326'
+            }
+        }
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
```
dirkf, line 47: To access both groups from the URL match, use `mobj = re.match(self._VALID_URL, url)` (`import re`) and then access the matches as `video_id = mobj.group('id')`. In this case the fact that the extractor is running means that the `id` and `display_id` groups matched. If you had an optional group, like `(?P<display_id>.+)?`, something like `display_id = mobj.groupdict().get('display_id')` would be appropriate.

youtube_dl/extractor/nbl.py
```diff
@@ -0,0 +1,69 @@
...
+        webpage = self._download_webpage(url, video_id)
+
+        title = self._html_search_regex(r'<title>(.+?)</title>', webpage, 'title')
```
dirkf, line 50: This pattern automatically strips whitespace from the returned value, and allows for attributes in the `<title>`:
```py
        title = self._html_search_regex(r'<title\b[^>]*>\s*(.+?)\s*</title>', webpage, 'title')
```

youtube_dl/extractor/nbl.py
```diff
@@ -0,0 +1,69 @@
...
+        stream_url = self._download_json(
```
dirkf, line 51: By convention, and not necessarily as preferred elsewhere, yt-dl uses single quotes (or triple-quotes) for strings.

Is the `data=b""` necessary, or just to force a POST?

youtube_dl/extractor/nbl.py
```diff
@@ -0,0 +1,69 @@
...
+            data=b"",
+            headers={
+                "Referer": "https://ott.nbl.com.au/en-int/embed/" + video_id,
+                "Origin": "https://ott.nbl.com.au",
+                "Content-Length": "0"
+            }
+        )["data"]["stream"]
+        formats = self._extract_m3u8_formats(stream_url, video_id, ext="mp4")
+
+        return {
```
dirkf, line 63: Might there be further metadata items in the page?

youtube_dl/extractor/nbl.py
```diff
@@ -0,0 +1,69 @@
...
+            'id': video_id,
+            'title': title,
+            'description': self._og_search_description(webpage, default=None),
+            'upload_date': unified_strdate(self._og_search_property('video:release_date', webpage, 'upload_date')),
```
dirkf, line 67: As this isn't a required item, add `fatal=False` to the args of `_og_search_property()`.

dirkf (changes requested):
Thanks for your work!

Are there any geo-restriction, login, etc, issues with the NBL.com.au media?

youtube_dl/extractor/nbl.py
alyssadev: It's to force a POST with urllib, yeah. That request requires a POST with Content-Length: 0

youtube_dl/extractor/nbl.py
alyssadev: I looked through the available meta tags on the page and elsewhere in the source, and in the embed source, that seems to be it for video metadata. If I find the API endpoint for requesting information about a certain game (I've found the endpoint for searching for all matching games and vods so far), that could be used to add more details to the info dict in the future if that's relevant information.

alyssadev:
There doesn't seem to be any geo restriction with the media, I've successfully downloaded full games to a server hosted in the Netherlands.
The website requires login for normal users to view content, but the streaming service at ott.nbl.com.au does not, once a video ID is retrieved it can be used to download that content without authentication.

alyssadev:
I'm investigating allowing more lenient URLs and retrieving the title information from the embed page instead, please stand by.

PR #30793, opened 2022-03-29
[options] add an `add` flag for each `embed` flag

lucasew:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

I sometimes download some songs with this program and, like with bash primitives, I forget which flag is `embed` and which is `add` for metadata and thumbnail. This PR just add an alias for `--embed-subs` and `--embed-thumbnail` replacing the `embed` word to `add` to solve this problem. Now if you want you can use all add instead of some add and some embed.

### DIFF
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -811,15 +811,15 @@ def parseOpts(overrideArguments=None):
         action='store_true', dest='nopostoverwrites', default=False,
         help='Do not overwrite post-processed files; the post-processed files are overwritten by default')
     postproc.add_option(
-        '--embed-subs',
+        '--embed-subs', '--add-subs',
         action='store_true', dest='embedsubtitles', default=False,
         help='Embed subtitles in the video (only for mp4, webm and mkv videos)')
     postproc.add_option(
-        '--embed-thumbnail',
+        '--embed-thumbnail', '--add-thumbnail',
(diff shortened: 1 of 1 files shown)


dirkf:
If doing this, probably also make `--embed-metadata` an alias for `--add-metadata`. However I am not particularly enthusiastic about these changes in general. Maybe make them but don't advertise them, as with deprecated options?

lucasew:
IDK for real, there is no conflict AFAIK to keep both.

Would even help because it could avoid confusion.

A deprecation may cause breakage and the cost of keeping both flags seems cheap.

dirkf:
Basically the more options we have, the fewer people will make it through to the part of the manual they actually need. So my suggestion was to make these secret aliases that you can use if you know them, but not to advertise them in the documentation.

lucasew:
Perfect

If someone got confused then the program will just magically accept it and carry on xD

pukkandan:
The only `--add...` option that currently exists is `--add-metadata`. So why not just deprecate it in favor of `--embed-metadata`

dirkf:
Which would also match yt-dlp!

Given that, I'm happy to align the options.

However the yt-dlp manual hasn't been consistently updated to use `--embed-metadata`.

PR #30809, opened 2022-04-03
readme: make it more clear that regex ignores case

chapmanjacobd:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

I was confused whether the regex needed to be special when using the `OR` regex operator to list multiple substrings `Like|This`. I had to dig into the code to know what was going on. By moving the adjective in the README to the front it should be clear that both regex and substrings are treated with re.IGNORECASE

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,10 +141,10 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
                                          --playlist-items 1-3,7,10-13", it will
                                          download the videos at index 1, 2, 3,
                                          7, 10, 11, 12 and 13.
-    --match-title REGEX                  Download only matching titles (caseless 
-                                         regex or sub-string)
+    --match-title REGEX                  Download only matching titles (case-
+                                         insensitive regex or sub-string)
     --reject-title REGEX                 Skip download for matching titles
-                                         (caseless regex or sub-string)
+                                         (case-insensitive regex or sub-string)
     --max-downloads NUMBER               Abort after downloading NUMBER files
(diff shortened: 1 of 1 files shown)


dirkf:
We say "caseless" where the whole world says "case-insensitive". According to references "caseless" means "having no case", and for words everyone except the Unicode standard uses it to means "having no grammatical case".

How about:
```md
    --match-title REGEX                  Download only matching titles (case-
                                         insensitive regex or sub-string)
    --reject-title REGEX                 Skip download for matching titles
                                         (case-insensitive regex or sub-string)
```

chapmanjacobd:
yes I think that would be an improvement as well

README.md
```diff
@@ -141,10 +141,10 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
                                          --playlist-items 1-3,7,10-13", it will
                                          download the videos at index 1, 2, 3,
                                          7, 10, 11, 12 and 13.
-    --match-title REGEX                  Download only matching titles (regex or
-                                         caseless sub-string)
+    --match-title REGEX                  Download only matching titles (case-
+                                         insensitive regex or sub-string)
     --reject-title REGEX                 Skip download for matching titles
-                                         (regex or caseless sub-string)
```
pukkandan, line 147: these are auto-generated from `options.py`. You should edit that file instead

chapmanjacobd:
moved to https://github.com/ytdl-org/youtube-dl/pull/30854 because I'm on mobile and lazy (sorry)

PR #30854, opened 2022-04-13
readme: make it more clear that regex ignores case

chapmanjacobd:
https://github.com/ytdl-org/youtube-dl/pull/30809

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

I was confused whether the regex needed to be special when using the `OR` regex operator to list multiple substrings `Like|This`. I had to dig into the code to know what was going on. By moving the adjective in the README to the front it should be clear that both regex and substrings are treated with re.IGNORECASE

### DIFF
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -270,11 +270,11 @@ def parseOpts(overrideArguments=None):
     selection.add_option(
         '--match-title',
         dest='matchtitle', metavar='REGEX',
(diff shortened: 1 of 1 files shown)


dirkf:
Agreed, and there's no clash with yt-dlp where these options are reworked (`--match-filter ...` supports case-sensitive pattern matching with the `~=` operator and the `(?i)` flag has to be included in the pattern to make the match case-insensitive).

PR #30885, opened 2022-04-22
[KTH] Add new extractor for KTH play

ArniDagur:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This pull request adds an extractor for the KTH Play website. KTH Play is a website which KTH uses to publish lectures and other videos online. KTH is a university in Stockholm, Sweden.

### DIFF
--- a/youtube_dl/extractor/kaltura.py
+++ b/youtube_dl/extractor/kaltura.py
@@ -373,5 +373,5 @@ class KalturaIE(InfoExtractor):
             'duration': info.get('duration'),
             'timestamp': info.get('createdAt'),
             'uploader_id': info.get('userId') if info.get('userId') != 'None' else None,
-            'view_count': info.get('plays'),
+            'view_count': int_or_none(info.get('plays')),
         }
--- a/youtube_dl/extractor/kth.py
+++ b/youtube_dl/extractor/kth.py
@@ -1,0 +1,31 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import smuggle_url
+
+
+class KTHIE(InfoExtractor):
+    _VALID_URL = r'https?://play\.kth\.se/(?:.+/)+(?P<id>[a-z0-9_]+)'
+    _TEST = {
+        'url': 'https://play.kth.se/media/Lunch+breakA+De+nya+aff%C3%A4rerna+inom+Fordonsdalen/0_uoop6oz9',
+        'md5': 'd83ada6d00ca98b73243a88efe19e8a6',
+        'info_dict': {
+            'id': '0_uoop6oz9',
+            'ext': 'mp4',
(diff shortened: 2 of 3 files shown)


youtube_dl/extractor/kaltura.py
```diff
@@ -373,5 +373,5 @@ def sign_url(unsigned_url):
             'duration': info.get('duration'),
             'timestamp': info.get('createdAt'),
             'uploader_id': info.get('userId') if info.get('userId') != 'None' else None,
-            'view_count': info.get('plays'),
+            'view_count': int(info.get('plays')) if info.get('plays') is not None else None,
```
ArniDagur, line 376: This is needed because the play count was received as an integer string, e.g. `"47"` rather than `47` for some reason.

ArniDagur: Note that I've now changed this line of code to use `int_or_none`.

youtube_dl/extractor/kaltura.py
```diff
@@ -373,5 +373,5 @@ def sign_url(unsigned_url):
             'duration': info.get('duration'),
             'timestamp': info.get('createdAt'),
             'uploader_id': info.get('userId') if info.get('userId') != 'None' else None,
-            'view_count': info.get('plays'),
+            'view_count': int_or_none(info.get('plays')),
```
dirkf, line 376: Good catch.

youtube_dl/extractor/kth.py
```diff
@@ -0,0 +1,31 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import smuggle_url
+
+
+class KTHIE(InfoExtractor):
+    _VALID_URL = r'https?://play\.kth\.se/(?:.+/)+(?P<id>[a-z0-9_]+)'
```
dirkf, line 9: `.+/` might be rather greedy. Try `(?:[^/]+/)`.

Can the id use upper-case letters?

dirkf (changes requested):
Thanks for your work! Just a couple of nits.

youtube_dl/extractor/kth.py
ArniDagur: The ID never contains any upper case letters.

dirkf (approved):

ArniDagur:
This PR has been approved. Does that mean it is ready to be merged, or is there another step in the process?

dirkf:
After approving the changes I enabled the test workflow. I'm glad to see that `test_KTH (test.test_download.TestDownload): ... ok` for both Py 2.7 and 3.9, so I can now merge it with a good conscience.

ArniDagur:
I'm wondering when the next `youtube-dl` release will be made. I'm excited to see this go out :)

PR #30908, opened 2022-05-01
[cozytv] Add new extractor (replays)

xFA25E:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

This PR provides an extractor for (cozy.tv)[https://cozy.tv]. It is a streaming service. It can extract only saved streams. Live-stream download is not implemented yet.

Example url: `https://cozy.tv/<user>/replays/<id>


youtube_dl/extractor/cozytv.py
```diff
@@ -0,0 +1,95 @@
...
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from ..utils import unified_timestamp, unified_strdate
+
+
+class CozyTVReplayIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?cozy\.tv/(?P<user>[^/]+)/replays/(?P<id>[^/]+)'
```
dirkf, line 11: `(?P<id>[^/]+)` could capture `#anchor` or `?param=value` after the `id`. Is there a more specific pattern that would work, say `[\w-]+` (based on the test URLs)?

youtube_dl/extractor/cozytv.py
```diff
@@ -0,0 +1,95 @@
...
+                'duration': 313,
+                'view_count': 656,
+                'webpage_url': 'https://cozy.tv/beardson/replays/2022-05-01_2',
+            }
+        }
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        user = re.match(self._VALID_URL, url).group('user')
```
dirkf, line 67: You can do this in one line:
```suggestion
        video_id, user = re.match(self._VALID_URL, url).group('id', 'user')
```

youtube_dl/extractor/cozytv.py
```diff
@@ -0,0 +1,95 @@
...
+        data = self._download_json('https://api.cozy.tv/cache/' + user + '/replay/' + video_id, video_id)
+
+        cdn_url = data['cdns'][0] + '/replays/' + user + '/' + video_id
+        user_url = 'https://cozy.tv/' + user
+        timestamp = unified_timestamp(data.get('date'))
+        date = unified_strdate(data.get('date'))
```
dirkf, line 73: Move this to the dict, but you don't need to set upload_date as the core will calculate it from the `timestamp`:
```py
            'timestamp': unified_timestamp(data.get('date')),
```

youtube_dl/extractor/cozytv.py
```diff
@@ -0,0 +1,95 @@
...
+
+        return {
+            'id': data.get('_id'),
+            'title': data.get('title'),
+            'formats': self._extract_m3u8_formats(cdn_url + '/index.m3u8', video_id, ext='mp4', m3u8_id='hls'),
```
dirkf, line 78: Conversely, calculate these easy required fields; then get the formats:
```py
        info = {
            # if this might be missing, maybe data.get('_id') or video_id ?
            'id': data['_id'],
            'title': data['title'],
        }

        formats = self._extract_m3u8_formats(cdn_url + '/index.m3u8', video_id, ext='mp4', m3u8_id='hls')
        # then call this, which also raises an exception if there were no formats
        self._sort_formats(formats)

        info.update({
            'formats': formats,
            'display_id': video_id,
...
        })
        return info
```

youtube_dl/extractor/cozytv.py
```diff
@@ -0,0 +1,95 @@
...
+            'display_id': video_id,
+            'thumbnail': cdn_url + '/thumb.webp',
+            'uploader': user,
+            'creator': user,
+            'release_timestamp': timestamp,
+            'release_date': date,
```
dirkf, line 84: As above, you can omit this and let the core set it from `release_timestamp`:
```suggestion
```

youtube_dl/extractor/cozytv.py
```diff
@@ -0,0 +1,95 @@
...
+            'timestamp': timestamp,
+            'upload_date': date,
+            'uploader_id': user,
+            'uploader_url': user_url,
+            'channel': user,
+            'channel_id': user,
+            'channel_url': user_url,
+            'duration': data.get('duration'),
+            'view_count': data.get('peakViewers'),
```
dirkf, line 93: Wrap these to ensure the correct type:
```suggestion
            'duration': float_or_none(data.get('duration')),
            'view_count': int_or_none(data.get('peakViewers')),
```

youtube_dl/extractor/extractors.py
```diff
@@ -247,6 +247,9 @@
     CNNArticleIE,
 )
 from .coub import CoubIE
+from .cozytv import (
+    CozyTVReplayIE,
+)
```
dirkf, line 252: Whereas, if there's just one item (and you're not planning to add another extractor), this can be one line:
```suggestion
from .cozytv import CozyTVReplayIE,
```

youtube_dl/extractor/cozytv.py
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        user = re.match(self._VALID_URL, url).group('user')
+        data = self._download_json('https://api.cozy.tv/cache/' + user + '/replay/' + video_id, video_id)
+
+        cdn_url = data['cdns'][0] + '/replays/' + user + '/' + video_id
```
dirkf, line 70: Maybe better:
```suggestion
        cdn_url = '/'.join((data['cdns'][0], 'replays', user, video_id))
```

youtube_dl/extractor/cozytv.py
```diff
@@ -0,0 +1,95 @@
...
+                'view_count': 656,
+                'webpage_url': 'https://cozy.tv/beardson/replays/2022-05-01_2',
+            }
+        }
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        user = re.match(self._VALID_URL, url).group('user')
+        data = self._download_json('https://api.cozy.tv/cache/' + user + '/replay/' + video_id, video_id)
```
dirkf, line 68: Maybe better:
```suggestion
        data = self._download_json('https://api.cozy.tv/cache/%s/replay/%s' % (user, video_id), video_id)
```

xFA25E:
Thanks a lot! I really appreciate your detailed and helpful suggestions! I'm happy to contribute to one of my favorite open source projects

ant0n-anon:
is there anything i can do to help this get done? would really appreciate this merged.

youtube_dl/extractor/cozytv.py
```diff
@@ -0,0 +1,114 @@
...
+                'webpage_url': 'https://cozy.tv/beardson/replays/2022-05-01_2',
+            }
+        }
+    ]
+
+    def _real_extract(self, url):
+        video_id, user = re.match(self._VALID_URL, url).group('id', 'user')
+        data = self._download_json('https://api.cozy.tv/cache/%s/replay/%s' % (user, video_id), video_id)
+
+        cdn_url = '/'.join((data['cdns'][0], 'replays', user, video_id))
```
dirkf, line 73: Let's try to force the CI tests, and also ensure no final `/`:
```suggestion
        cdn_url = '/'.join((data['cdns'][0], 'replays', user, video_id)).rstrip('/')
```

youtube_dl/extractor/cozytv.py
dirkf: Although I think you may have taken "specific" more literally than I intended!

youtube_dl/extractor/cozytv.py
```diff
@@ -0,0 +1,114 @@
...
+            'formats': formats,
+            'release_timestamp': timestamp,
+            'thumbnail': cdn_url + '/thumb.webp',
+            'timestamp': timestamp,
+            'uploader': user,
+            'uploader_id': user,
+            'uploader_url': user_url,
+            'view_count': int_or_none(data.get('peakViewers')),
+            'webpage_url': url,
+        })
```
dirkf, line 112: Finally, let's check whether the metadata makes sense.

* is `user` actually the creator on the site, or just the uploader?
* according to the definition (for historical reasons yt-dl metadata is based on what YouTube offers) `uploader` is the user's actual name while `uploader_id` is the user's ID on the site: does it make sense to extract the same value to both?
* similarly for `channel`/`channel_id`?
* does the site have "channels" that justify extracting `channel*` with the same values as `uploader*` (considering what the IE doc says)?
 > ... channel fields may or may not repeat uploader fields. This depends on a particular extractor.
* should `release_timestamp` be set if it just repeats `timestamp`?

xFA25E:
I've added an extractor for livestreams

PR #30985, opened 2022-05-27
[dumpert] Fix videos downloading as m3u8 and added new-style valid URLs

DataGhost:

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Dumpert has changed playout to HLS a while ago (months or even over a year) which causes all videos since that change to download as .m3u8 files actually containing mp4 data. Older videos are unaffected, that's why the test never failed. Additionally, the site has received a new layout with some new possible URLs a while ago, those should now work as well.

### DIFF
--- a/youtube_dl/extractor/dumpert.py
+++ b/youtube_dl/extractor/dumpert.py
@@ -29,6 +29,16 @@ class DumpertIE(InfoExtractor):
     }, {
         'url': 'http://legacy.dumpert.nl/embed/6675421/dc440fe7',
         'only_matching': True,
+    }, {
+        'url': 'https://www.dumpert.nl/item/100031688_b317a185',
+        'md5': '98bece7add22c2645eb3906e6d55318b',
+        'info_dict': {
+            'id': '100031688/b317a185',
+            'ext': 'mp4',
+            'title': 'Epic schijnbeweging',
+            'description': '<p>Die zag je niet eh</p>',
+            'thumbnail': r're:^https?://.*\.(?:jpg|png)$',
+        },
     }]
 
     def _real_extract(self, url):
--- a/youtube_dl/extractor/dumpert.py
+++ b/youtube_dl/extractor/dumpert.py
@@ -50,6 +50,7 @@ class DumpertIE(InfoExtractor):
                 'url': uri,
                 'format_id': version,
                 'quality': quality(version),
+                'ext': 'mp4',
             })
         self._sort_formats(formats)
 
--- a/youtube_dl/extractor/dumpert.py
+++ b/youtube_dl/extractor/dumpert.py
@@ -9,7 +9,7 @@ from ..utils import (
 
 
 class DumpertIE(InfoExtractor):
-    _VALID_URL = r'(?P<protocol>https?)://(?:(?:www|legacy)\.)?dumpert\.nl/(?:mediabase|embed|item)/(?P<id>[0-9]+[/_][0-9a-zA-Z]+)'
+    _VALID_URL = r'(?P<protocol>https?)://(?:(?:www|legacy)\.)?dumpert\.nl(?:/(?:mediabase|embed|item)/|(?:/toppers|/latest|/?)\?selectedId=)(?P<id>[0-9]+[/_][0-9a-zA-Z]+)'
     _TESTS = [{
         'url': 'https://www.dumpert.nl/item/6646981_951bc60f',
         'md5': '1b9318d7d5054e7dcb9dc7654f21d643',
@@ -39,6 +39,15 @@ class DumpertIE(InfoExtractor):
             'description': '<p>Die zag je niet eh</p>',
             'thumbnail': r're:^https?://.*\.(?:jpg|png)$',
         },
+    }, {
+        'url': 'https://www.dumpert.nl/toppers?selectedId=100031688_b317a185',
+        'only_matching': True,
+    }, {
+        'url': 'https://www.dumpert.nl/latest?selectedId=100031688_b317a185',
+        'only_matching': True,
+    }, {
+        'url': 'https://www.dumpert.nl/?selectedId=100031688_b317a185',
+        'only_matching': True,
     }]
 
     def _real_extract(self, url):


youtube_dl/extractor/dumpert.py
```diff
@@ -50,6 +69,7 @@ def _real_extract(self, url):
                 'url': uri,
```
dirkf, line 69: Is the issue that the `uri` is a manifest (m3u8)? Or is it actually a direct mp4 link with a confusing format?

If the former, the typical pattern for handling format URLs is
```py
            ext = determine_ext(uri)
            if ext == 'm3u8':
                formats.extend(self._extract_m3u8_formats(uri, ...)
            elif ext == ...: # other manifests, eg mpd
                ...
            else:
               formats.append({
                   'url': uri,
                   ...
               })
```
This is so common that it should have been abstracted in a common routine, I think, but perhaps one with too many arguments. Anyhow it automatically extracts all the available formats from the manifest so that they can be listed. If appropriate, do the same here.

DataGhost: I have to admit I did see that going on in some other extractors but thought it wasn't meant to be used for non-master playlists (which is what they are here). Looking at the code, I saw no way to add in the quality argument through `_extract_m3u8_formats` so I did that in a separate loop afterwards.

youtube_dl/extractor/dumpert.py
```diff
@@ -39,18 +59,24 @@ def _real_extract(self, url):
...
+            if ext == 'm3u8':
+                formats.extend(self._extract_m3u8_formats(
+                    uri, video_id, ext='mp4', m3u8_id=version))
+            else:
+                formats.append({
+                    'url': uri,
+                    'format_id': version,
+                })
+            for format in formats:
+                format['quality'] = quality(format['format_id'])
```
dirkf, line 79: So, looking at (1 of) the manifests, they're playlists and you could pass `preference=quality(version)` into `_extract_m3u8_formats()`. Probably also `entry_protocol='m3u8_native' as the manifests look harmless.

Incidentally the API data appears to be in the webpage hydration JSON as well `window.__DUMPERT_STATE__ = JSON.parse(...)` but as you don't look at the page itself that doesn't matter.

DataGhost: Hm, for some weird reason I thought I'd seen somewhere that preference was supposed to be a float between 0.0 and 1.0 and that seemed annoying/ugly with a possibly longer list of qualities. Obviously it's free-form. I'll change it to preference in both cases, and add the suggested entry_protocol change.

DataGhost: For some reason, using entry_protocol='m3u8_native' changes the md5. Calling youtube-dl with --test uses ffmpeg to download a file of approx 93kB which is then used for the md5, rather than only the first 10241 bytes as claimed in the docs. After some debugging in the tests code it seems that it does use a 10241-bytes long file to calculate the md5 from, however it seems to have some kind of header prepended that the resulting video does not have. I'm just going to assume everything is all right as the resulting complete files obtained through both methods are identical.

dirkf: Generally, downloads using a manifest might be adaptive so the file MD5 check may not be reliable. Following precedent I put `'params': {'skip_download': True},` in the test case and comment out the `'skip_download'` line while testing locally.

DataGhost: In this case, keep it as is then?

youtube_dl/extractor/dumpert.py
```diff
@@ -3,13 +3,14 @@
...
 from ..utils import (
+    determine_ext,
     int_or_none,
     qualities,
 )
 
 
 class DumpertIE(InfoExtractor):
-    _VALID_URL = r'(?P<protocol>https?)://(?:(?:www|legacy)\.)?dumpert\.nl/(?:mediabase|embed|item)/(?P<id>[0-9]+[/_][0-9a-zA-Z]+)'
+    _VALID_URL = r'(?P<protocol>https?)://(?:(?:www|legacy)\.)?dumpert\.nl(?:/(?:mediabase|embed|item)/|(?:/toppers|/latest|/?)\?selectedId=)(?P<id>[0-9]+[/_][0-9a-zA-Z]+)'
```
dirkf, line 13: Also, support https://www.dumpert.nl/latest/?selectedId=100031688_b317a185.

```suggestion
    _VALID_URL = r'(?P<protocol>https?)://(?:(?:www|legacy)\.)?dumpert\.nl(?:/(?:mediabase|embed|item)/|(?:/toppers|/latest)?/?)\?selectedId=)(?P<id>[0-9]+[/_][0-9a-zA-Z]+)'
```
Are there any other "slugs" like `toppers`? We seem to get the same video (based on the `id`) regardless, but random choices of https://www.dumpert.nl/nieuw?selectedId=100031688_b317a185, etc, don't.

youtube_dl/extractor/dumpert.py
```diff
@@ -3,13 +3,14 @@
...
 from ..utils import (
+    determine_ext,
     int_or_none,
     qualities,
 )
 
 
 class DumpertIE(InfoExtractor):
-    _VALID_URL = r'(?P<protocol>https?)://(?:(?:www|legacy)\.)?dumpert\.nl/(?:mediabase|embed|item)/(?P<id>[0-9]+[/_][0-9a-zA-Z]+)'
+    _VALID_URL = r'(?P<protocol>https?)://(?:(?:www|legacy)\.)?dumpert\.nl(?:/(?:mediabase|embed|item)/|(?:/toppers|/latest)?/?)\?selectedId=)(?P<id>[0-9]+[/_][0-9a-zA-Z]+)'
```
dirkf, line 13:
```suggestion
    _VALID_URL = r'(?P<protocol>https?)://(?:(?:www|legacy)\.)?dumpert\.nl(?:/(?:mediabase|embed|item)/|(?:/toppers|/latest)?/?\?selectedId=)(?P<id>[0-9]+[/_][0-9a-zA-Z]+)'
```

dirkf (approved):
Good to go.

PR #31057, opened 2022-06-27
[livestreamfails] Add new extractor

nomevi:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Add an extractor for [Livestreamfails](https://livestreamfails.com), a Twitch clip mirror site for [r/livestreamfail](https://reddit.com/r/LivestreamFail/).

### DIFF
--- a/youtube_dl/extractor/livestreamfails.py
+++ b/youtube_dl/extractor/livestreamfails.py
@@ -1,0 +1,49 @@
+from .common import InfoExtractor
+import json
+import time
+import calendar
+
+
+class LivestreamfailsIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?livestreamfails\.com/clip/(?P<id>[0-9]+)'
+    _TESTS = [{
+        'url': 'https://livestreamfails.com/clip/139200',
+        'md5': '8a03aea1a46e94a05af6410337463102',
+        'info_dict': {
+            'id': '139200',
+            'ext': 'mp4',
+            'display_id': 'ConcernedLitigiousSalmonPeteZaroll-O8yo9W2L8OZEKhV2',
+            'title': 'Streamer jumps off a trampoline at full speed',
+            'creator': 'paradeev1ch',
+            'thumbnail': 'https://livestreamfails-image-prod.b-cdn.net/image/3877b1d38db083fa25c82685bbaf645637e575ea.png',
+            'timestamp': 1656271785,
+            'upload_date': '20220626',
+        }
+    }]
+
+    def _real_extract(self, url):
+        result = {}
+        result['id'] = self._match_id(url)
+
+        # https://livestreamfails.com/clip/id uses https://api.livestreamfails.com/clip/ to fetch the video metadata
+        # Use the same endpoint here to avoid loading and parsing the provided page (which requires JS)
+        apiResponse = json.loads(self._download_webpage('https://api.livestreamfails.com/clip/' + result['id'], result['id']))
+
+        # Twitch ID of clip
+        result['display_id'] = apiResponse.get('sourceId')
+
+        # Get the input timestamp (test case gives 2022-06-26T19:29:45.515Z)
+        result['timestamp'] = apiResponse.get('createdAt')
+        if(result.get('timestamp')):
+            # Parse it into a struct_time
+            result['timestamp'] = time.strptime(result['timestamp'], '%Y-%m-%dT%H:%M:%S.%fZ')
+            # Convert the struct_time to a UNIX timestamp while ignoring the local timezone attached by time.strptime()
+            result['timestamp'] = calendar.timegm(result['timestamp'])
+
+        # Other fields
+        result['url'] = 'https://livestreamfails-video-prod.b-cdn.net/video/' + apiResponse.get('videoId')
+        result['title'] = apiResponse.get('label')
+        result['creator'] = apiResponse.get('streamer', {}).get('label')
+        result['thumbnail'] = 'https://livestreamfails-image-prod.b-cdn.net/image/' + apiResponse.get('imageId')
+
+        return result
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -627,6 +627,9 @@ from .livestream import (
     LivestreamOriginalIE,
     LivestreamShortenerIE,
 )
+from .livestreamfails import (
+    LivestreamfailsIE,
+)
 from .lnkgo import LnkGoIE
 from .localnews8 import LocalNews8IE
 from .lovehomeporn import LoveHomePornIE


youtube_dl/extractor/livestreamfails.py
```diff
@@ -0,0 +1,49 @@
...
+            'creator': 'paradeev1ch',
+            'thumbnail': 'https://livestreamfails-image-prod.b-cdn.net/image/3877b1d38db083fa25c82685bbaf645637e575ea.png',
+            'timestamp': 1656271785,
+            'upload_date': '20220626',
+        }
+    }]
+
+    def _real_extract(self, url):
+        result = {}
+        result['id'] = self._match_id(url)
```
dirkf, line 26: The normal style is to stash results in variables and then construct the result dict as it's about to be returned. But if you're using this style, why not do:
```suggestion
        result = {
            'id': self._match_id(url),
        }
```

youtube_dl/extractor/livestreamfails.py
```diff
@@ -0,0 +1,49 @@
...
+
+        # https://livestreamfails.com/clip/id uses https://api.livestreamfails.com/clip/ to fetch the video metadata
+        # Use the same endpoint here to avoid loading and parsing the provided page (which requires JS)
+        apiResponse = json.loads(self._download_webpage('https://api.livestreamfails.com/clip/' + result['id'], result['id']))
```
dirkf, line 30: Use `IE._download_json()`.

youtube_dl/extractor/livestreamfails.py
```diff
@@ -0,0 +1,49 @@
...
+        # Get the input timestamp (test case gives 2022-06-26T19:29:45.515Z)
+        result['timestamp'] = apiResponse.get('createdAt')
+        if(result.get('timestamp')):
+            # Parse it into a struct_time
+            result['timestamp'] = time.strptime(result['timestamp'], '%Y-%m-%dT%H:%M:%S.%fZ')
+            # Convert the struct_time to a UNIX timestamp while ignoring the local timezone attached by time.strptime()
+            result['timestamp'] = calendar.timegm(result['timestamp'])
+
+        # Other fields
+        result['url'] = 'https://livestreamfails-video-prod.b-cdn.net/video/' + apiResponse.get('videoId')
```
dirkf, line 44: This should fail if the data isn't available. Eg:
```suggestion
        result['url'] = 'https://livestreamfails-video-prod.b-cdn.net/video/' + apiResponse['videoId']
```

youtube_dl/extractor/livestreamfails.py
```diff
@@ -0,0 +1,49 @@
...
+        result['title'] = apiResponse.get('label')
```
dirkf, line 45: Similarly, should fail if no `label`, or default to `IE._generic_title()`.

youtube_dl/extractor/livestreamfails.py
```diff
@@ -0,0 +1,49 @@
...
+        result['creator'] = apiResponse.get('streamer', {}).get('label')
```
dirkf, line 46: But this should not fail.
```suggestion
        result['creator'] = try_get(apiResponse, lambda x: x['streamer']['label'], compat_str)
```

youtube_dl/extractor/livestreamfails.py
```diff
@@ -0,0 +1,49 @@
...
+        result['thumbnail'] = 'https://livestreamfails-image-prod.b-cdn.net/image/' + apiResponse.get('imageId')
```
dirkf, line 47: Similarly:
```suggestion
        result['thumbnail'] = url_or_none(try_get(apiResponse, lambda x: 'https://livestreamfails-image-prod.b-cdn.net/image/' + x['imageId']))
```

youtube_dl/extractor/livestreamfails.py
```diff
@@ -0,0 +1,49 @@
...
+        # Twitch ID of clip
+        result['display_id'] = apiResponse.get('sourceId')
+
+        # Get the input timestamp (test case gives 2022-06-26T19:29:45.515Z)
+        result['timestamp'] = apiResponse.get('createdAt')
+        if(result.get('timestamp')):
+            # Parse it into a struct_time
+            result['timestamp'] = time.strptime(result['timestamp'], '%Y-%m-%dT%H:%M:%S.%fZ')
+            # Convert the struct_time to a UNIX timestamp while ignoring the local timezone attached by time.strptime()
+            result['timestamp'] = calendar.timegm(result['timestamp'])
```
dirkf, line 41:
```suggestion
        result['timestamp'] = parse_iso8601(apiResponse.get('createdAt'))
```

youtube_dl/extractor/livestreamfails.py
```diff
@@ -0,0 +1,49 @@
...
+    def _real_extract(self, url):
+        result = {}
+        result['id'] = self._match_id(url)
+
+        # https://livestreamfails.com/clip/id uses https://api.livestreamfails.com/clip/ to fetch the video metadata
+        # Use the same endpoint here to avoid loading and parsing the provided page (which requires JS)
+        apiResponse = json.loads(self._download_webpage('https://api.livestreamfails.com/clip/' + result['id'], result['id']))
+
+        # Twitch ID of clip
+        result['display_id'] = apiResponse.get('sourceId')
```
dirkf, line 33: This should fail if the data is not available, or default to `IE._generic_id()`:
```suggestion
        result['display_id'] = apiResponse.get['sourceId']
```
After this, `apiResponse` is known to be a dict that you can `.get()` from without crashing.

dirkf (changes requested):
Thanks for your work!

Apart from the detailed points I've made, consider re-working the last part of `_real_extract()` so that after setting `id`, `title` and `url` you just set the non-mandatory data in one block:
```py
        result.update({
            'timestamp': ...,
            ...,
        })
        return result
```
I've enabled the CI test, so please also check the results and address any failures related to the new extractor.

youtube_dl/extractor/livestreamfails.py
```diff
@@ -0,0 +1,49 @@
+from .common import InfoExtractor
```
dirkf, line 1:
```suggestion
# coding: utf-8
from __future__ import unicode_literals

from .common import InfoExtractor
```

youtube_dl/extractor/livestreamfails.py
nomevi: I would not value `sourceId` that highly since it links back to Twitch and therefore doesn't affect the downloaded video. Is it really worth breaking the entire extractor for what is arguably a niche situation?

youtube_dl/extractor/livestreamfails.py
```diff
@@ -0,0 +1,47 @@
...
+        id = self._match_id(url)
+
+        # https://livestreamfails.com/clip/id uses https://api.livestreamfails.com/clip/ to fetch the video metadata
+        # Use the same endpoint here to avoid loading and parsing the provided page (which requires JS)
+        api_response = self._download_json('https://api.livestreamfails.com/clip/' + id, id)
+
+        return {
+            'id': id,
+            'url': 'https://livestreamfails-video-prod.b-cdn.net/video/' + api_response['videoId'],
+            'title': api_response['label'],
```
dirkf, line 42: I'd bring these two out as separate statements before the return so that if one of them fails the diagnostics are clearer.
```suggestion
        video_url = 'https://livestreamfails-video-prod.b-cdn.net/video/' + api_response['videoId']
        title = api_response['label']

        return {
            'id': id,
            'url': video_url,
            'title': title,
```

youtube_dl/extractor/livestreamfails.py
```diff
@@ -0,0 +1,47 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+import time
+import calendar
```
dirkf, line 6: No longer used (Linter will complain)?

dirkf (changes requested):
Thanks.

youtube_dl/extractor/livestreamfails.py
dirkf: Right, that's fine for `display_id`.

youtube_dl/extractor/livestreamfails.py
nomevi: My bad, my IDE was misconfigured.

nomevi:
By the way, I appreciate the detailed review! Thank you for dealing with my basic mistakes.

PR #31141, opened 2022-08-07
[truth] Backport extractor from yt-dlp

palewire:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [x] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [x] Covered the code with tests (note that PRs without tests will be REJECTED)
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

This patch adds a new extractor for truthsocial.com, the microblogging service founded by former U.S. president Donald Trump. Due to its frequent use by the Trump family and other political actors, this would be a useful addition to the library for journalists and other researchers.

Issue #30976, opened 2022-05-24
[youtube] Unable to decode n-parameter: download likely to be throttled 


unityconstruct:
## Checklist
- [x] I'm reporting a broken site support

- [x] I've verified that I'm running youtube-dl version **2021.12.17**
```
./youtube-dl --version
2021.12.17
```
-- NOTE!!!!: Custom build from Master Working until recently. See: [Issue30583](https://github.com/ytdl-org/youtube-dl/issues/30583#issuecomment-1107068931)

- [x] I've checked that all provided URLs are alive and playable in a browser
-- URL plays normally in BraveBrowser: https://www.youtube.com/watch?v=lDqlpTcxLqg

- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
-- NOTE: using shell script so calls are consistent.
`COMMAND STRING: /media/media3/ytu/youtube-dl -v --console-title --verbose -f best[height<=?1080]/bestvideo[height<=?1080]+bestaudio --merge-output-format mp4  --write-thumbnail --write-description --write-info-json --ignore-errors --max-downloads 50 --no-overwrites --restrict-filenames --download-archive /media/media3/ytu/archive-video.txt --no-playlist  --output /home/uc/Videos/ytu/%(uploader)s/%(title)s-%(id)s.%(ext)s  https://www.youtube.com/watch?v=lDqlpTcxLqg`

- [x] I've searched the bugtracker for similar issues including closed ones
-- Search Term yielded no results: "Unable to decode n-parameter"


## Verbose log
```
Geting single URL ( dont get entire playlist )
NOT using BATCHFILE, using https://www.youtube.com/watch?v=lDqlpTcxLqg
COMMAND STRING: /media/media3/ytu/youtube-dl -v --console-title --verbose -f best[height<=?1080]/bestvideo[height<=?1080]+bestaudio --merge-output-format mp4  --write-thumbnail --write-description --write-info-json --ignore-errors --max-downloads 50 --no-overwrites --restrict-filenames --download-archive /media/media3/ytu/archive-video.txt --no-playlist  --output /home/uc/Videos/ytu/%(uploader)s/%(title)s-%(id)s.%(ext)s  https://www.youtube.com/watch?v=lDqlpTcxLqg
Timestamp: 
[R]un or Run & [T]ail? Options:[r/t]r
Typed: r
[debug] System config: []
[debug] User config: []                                                                                                                                                                                                                  
[...]

PR #31144, opened 2022-08-07
[parler] Backport extractor from yt-dlp

palewire:

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---


youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from urllib import parse
```
dirkf, line 4: See below.

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
...
+
+from .common import InfoExtractor
+
+from ..utils import clean_html, unified_timestamp
```
dirkf, line 8: Please format like this for mantainability; also, add routines per other suggestions:
```suggestion
from ..utils import (
    clean_html,
    strip_or_none,
    unified_timestamp,
    urlencode_postdata,
)
```

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class ParlerIE(InfoExtractor):
+    """Extract videos from posts on Parler."""
+
+    _VALID_URL = r"https://parler\.com/feed/(?P<id>[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})"
```
dirkf, line 14: Project convention is ''; also, as 5 other extractors have defined it already and probably others should have, prepare for utils.UUID_RE to be defined:
```suggestion
    _UUID_RE = r'[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
    _VALID_URL = r'https://parler\.com/feed/(?P<id>%s)' % (_UUID_RE, )
```

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
...
+                "uploader_id": "BennyJohnson",
+            },
+        },
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        # Get data from API
+        api_url = "https://parler.com/open-api/ParleyDetailEndpoint.php"
```
dirkf, line 48:
```suggestion
        api_url = 'https://parler.com/open-api/ParleyDetailEndpoint.php'
```

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
...
+        payload = parse.urlencode({"uuid": video_id}).encode()
+        status = self._download_json(api_url, video_id, data=payload)
+
+        # Pull out video
+        data = status["data"][0]["primary"]
+        video = data["video_data"]
+        url = video["videoSrc"]
```
dirkf, line 55: Although it's less neat, it's good to try for the mandatory field first, so that a log shows that failing rather than an intermediate stage:
```suggestion
        url = status['data'][0]['primary']['video_data']['videoSrc']
        # now we know this exists and is a dict
        data = status['data'][0]['primary']
```

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
...
+
+        # Pull out metadata
+        title = clean_html(data.get("full_body")).replace("\n", "")
```
dirkf, line 58: `clean_html()` removes newlines; also yt-dl expects a non-empty title:
```suggestion
        title = clean_html(data.get('full_body')) or self.generic_title(url)
```

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
...
+            },
+        },
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+
+        # Get data from API
+        api_url = "https://parler.com/open-api/ParleyDetailEndpoint.php"
+        payload = parse.urlencode({"uuid": video_id}).encode()
```
dirkf, line 49: Use `urlencode_postdata()` from `utils.py`:
```suggestion
        payload = urlencode_postdata({'uuid': video_id})
```

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
...
+            "url": "https://parler.com/feed/df79fdba-07cc-48fe-b085-3293897520d7",
+            "md5": "16e0f447bf186bb3cf64de5bbbf4d22d",
+            "info_dict": {
+                "id": "df79fdba-07cc-48fe-b085-3293897520d7",
+                "ext": "mp4",
+                "title": "Puberty-blocking procedures promoted by the Biden/Harris Admin are child abuse. The FDA has recently confirmed these hormones/drugs have extremely dangerous side effects, like brain swelling and vision loss.",
+                "timestamp": 1659744000,
+                "upload_date": "20220806",
+                "uploader": "Tulsi Gabbard",
+                "uploader_id": "TulsiGabbard",
```
dirkf, line 26: "" -> ''

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
...
+            "url": "https://parler.com/feed/a7406eb4-91e5-4793-b5e3-ade57a24e287",
+            "md5": "11687e2f5bb353682cee338d181422ed",
+            "info_dict": {
+                "id": "a7406eb4-91e5-4793-b5e3-ade57a24e287",
+                "ext": "mp4",
+                "title": "This man should run for office",
+                "timestamp": 1659657600,
+                "upload_date": "20220805",
+                "uploader": "Benny Johnson",
+                "uploader_id": "BennyJohnson",
```
dirkf, line 39: "" -> ''

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
...
+        # Pull out video
+        data = status["data"][0]["primary"]
+        video = data["video_data"]
+        url = video["videoSrc"]
+
+        # Pull out metadata
+        title = clean_html(data.get("full_body")).replace("\n", "")
+        timestamp = unified_timestamp(data.get("date_created"))
+        uploader = data.get("name")
+        uploader_id = data.get("username")
```
dirkf, line 61: For API string data you can condition the values with `strip_or_none()` from `utils.py`:
```suggestion
        uploader = strip_or_none(data.get('name'))
        uploader_id = strip_or_none(data.get('username'))
```

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
...
+        status = self._download_json(api_url, video_id, data=payload)
+
+        # Pull out video
+        data = status["data"][0]["primary"]
+        video = data["video_data"]
+        url = video["videoSrc"]
+
+        # Pull out metadata
+        title = clean_html(data.get("full_body")).replace("\n", "")
+        timestamp = unified_timestamp(data.get("date_created"))
```
dirkf, line 59:
```suggestion
        timestamp = unified_timestamp(data.get('date_created'))
```

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
...
+        uploader = data.get("name")
+        uploader_id = data.get("username")
+        uploader_url = "https://parler.com/" + uploader_id if uploader_id else None
```
dirkf, line 62: Parentheses make this clearer although not required:
```suggestion
        uploader_url = ('https://parler.com/' + uploader_id) if uploader_id else None
```

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,73 @@
...
+
+        # Return the result
+        return {
+            "id": video_id,
+            "url": url,
+            "title": title,
+            "timestamp": timestamp,
+            "uploader": uploader,
+            "uploader_id": uploader_id,
+            "uploader_url": uploader_url,
```
dirkf, line 72: "" -> ''

dirkf (changes requested):
Thanks for your work!

I've proposed some changes, mostly regarding project conventions.

If any comments apply to the other 2 new PRs that you've opened, please feel free to apply the equivalent changes; I'll review them in due course.

youtube_dl/extractor/parler.py
palewire: This is done.

youtube_dl/extractor/parler.py
palewire: Done.

palewire:
Thanks for all the advice. I've made your changes and tried too apply them to #31141 and #31143 as well.

palewire:
I've updated this patch to act as a backport from yt-dlp. You can find it here. https://github.com/yt-dlp/yt-dlp/pull/4616

I've left out the third test for a YouTube embed because I'm unsure how to handle that task in this package. If you'd like that included, if can give me some pointers on how to implement it I'd be happy to add.

pukkandan:
> I've left out the third test for a YouTube embed because I'm unsure how to handle that task in this package. If you'd like that included, if can give me some pointers on how to implement it I'd be happy to add.

It's the exact same. Just return a url_result. Only difference as I remember is that you have to do `YoutubeIE.ie_key()` instead of just `YoutubeIE` for the key

dirkf:
> > I've left out the third test for a YouTube embed because I'm unsure how to handle that task in this package. If you'd like that included, if can give me some pointers on how to implement it I'd be happy to add.
> 
> It's the exact same. Just return a url_result. Only difference as I remember is that you have to do `YoutubeIE.ie_key()` instead of just `YoutubeIE` for the key

In the third test, you need to have an `add_ie` item at the same level as the `info_dict`:
```
            'add_ie': ['Youtube'],
```
Apparently the test works without that in yt-dlp?

pukkandan:
Tests don't work in youtube-dl without add_ie? I haven't changed anything related to it in yt-dlp - and honestly have never understood what is the purpose of that key is. It could be a change inherited via yt-dl**c**, though I don't see any obvious commits for it in the history

dirkf:
I guess, something to do with lazy_extractors? What happened when @palewire ran the test in yt-dl? From the comment I inferred that it failed.

The `add_IE` thing has two effects, I find:
* ` [{self.addIE}]` appended to the test name
* check that the added IE is `.working()`.

palewire:
@dirkf, I added the YouTube passthrough test back in, as you suggested.

However, I'm getting #30976 when I try to test it locally. It's failing to dig out https://youtu.be/r5vkSaz8PxQ. 

If I run the vanilla CLI against that URL it does work. So I suspect there must be something off about how I'm accessing the YouTubeIE with this internal reference.

dirkf:
> getting #30976 when I try to test it locally

I think you need to update from the ytdl-org master branch, which fixed this a few days ago (after YT changed the challenge syntax in the latest player JS).

palewire:
Here's the error I get:

```bash
$ python test/test_download.py TestDownload.test_Parler_2                                                                                         
[Parler] f23b85c1-6558-470f-b9ff-02c145f28da5: Downloading JSON metadata                                                                          
[youtube] r5vkSaz8PxQ: Downloading webpage                                                                                                        
[youtube] r5vkSaz8PxQ: Downloading MPD manifest                                                                                                   
[youtube] r5vkSaz8PxQ: Downloading player 0c356943                                                                                                
ERROR: [youtube] Unable to decode n-parameter: download likely to be throttled (Failed to evaluate 0 + None (caused by TypeError("unsupported oper
and type(s) for +: 'int' and 'NoneType'")); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  h
ttps://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output. Traceback (most rec
ent call last):                                                                                                                                   
  File "/home/palewire/Code/youtube-dl/youtube_dl/jsinterp.py", line 203, in _operator                                                            
    return opfunc(left_val, right_val)                                                                                                            
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

dirkf:
That's the fixed error (actually a later player is fixed as well, and the latest player(s) have, so far, not broken our mini JS interpreter) .

palewire:
I pulled from master and still got it. So I think this bug is probably a phantom and the patch is ready for your review.

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,113 @@
...
+            'add_ie': ['Youtube'],
+        },
+    ]
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        data = self._download_json(
+            'https://parler.com/open-api/ParleyDetailEndpoint.php', video_id,
+            data=urlencode_postdata({'uuid': video_id}))['data'][0]
+        primary = data['primary']
```
dirkf, line 94:
```suggestion
        primary = try_get(data, lambda x: x['primary'], dict) or {}
```

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,113 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from .youtube import YoutubeIE
+from ..utils import (
+    clean_html,
+    int_or_none,
+    strip_or_none,
```
dirkf, line 9:
```suggestion
    strip_or_none,
    try_get,
```

dirkf (commented):
Trying to force CI tests.

youtube_dl/extractor/parler.py
```diff
@@ -0,0 +1,114 @@
...
+            'title': "Parler video #%s" % video_id,
+            'description': strip_or_none(clean_html(primary.get('full_body'))) or None,
+            'timestamp': unified_timestamp(primary.get('date_created')),
+            'uploader': strip_or_none(primary.get('name')),
+            'uploader_id': strip_or_none(primary.get('username')),
+            'uploader_url': 'https://parler.com/%s' % strip_or_none(primary.get('username')),
+            'view_count': int_or_none(primary.get('view_count')),
+            'comment_count': int_or_none(data['engagement']['commentCount']),
+            'repost_count': int_or_none(data['engagement']['echoCount']),
+        }
```
dirkf, line 114:
```suggestion
        }

```

palewire:
I'm withdrawing this request to tidy up my GitHub task panels. No sweat about it at all.

PR #31151, opened 2022-08-09
[twitter] Shorten title attribute for #29912

palewire:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

This request trims the standard Twitter ID field to workaround the bug reported in #29912

### DIFF
--- a/youtube_dl/extractor/twitter.py
+++ b/youtube_dl/extractor/twitter.py
@@ -198,7 +198,7 @@ class TwitterIE(TwitterBaseIE):
         'info_dict': {
             'id': '1555878318469091330',
             'ext': 'mp4',
-            'title': 'Tulsi Gabbard 🌺 - Puberty-blocking procedures promote[...]',
+            'title': 'Tulsi Gabbard 🌺 - Puberty-blocking ...',
             'description': 'Puberty-blocking procedures promoted by the Biden/Harris Admin are child abuse. The FDA has recently confirmed these hormones/drugs have extremely dangerous side effects, like brain swelling and vision loss. https://t.co/pcc70Uyt76',
             'thumbnail': r're:^https?://.*\.jpg',
             'uploader': 'Tulsi Gabbard 🌺',
@@ -213,7 +213,7 @@ class TwitterIE(TwitterBaseIE):
         'info_dict': {
             'id': '643211948184596480',
             'ext': 'mp4',
-            'title': 'FREE THE NIPPLE - FTN supporters on Hollywood Blvd to[...]',
+            'title': 'FREE THE NIPPLE - FTN supporters on Hollywood Blvd today!',
             'thumbnail': r're:^https?://.*\.jpg',
             'description': 'FTN supporters on Hollywood Blvd today! http://t.co/c7jHH749xJ',
             'uploader': 'FREE THE NIPPLE',
@@ -242,7 +242,7 @@ class TwitterIE(TwitterBaseIE):
         'info_dict': {
             'id': '665052190608723968',
             'ext': 'mp4',
-            'title': 'Star Wars | Andor Premieres Sept 21 on Disney+ - A new beginning is coming December [...]',
+            'title': 'Star Wars | Andor Premieres Sept 21 on Disney+ - A new beginning is coming December ...',
             'description': 'A new beginning is coming December 18. Watch the official 60 second #TV spot for #StarWars: #TheForceAwakens. https://t.co/OkSqT2fjWJ',
             'uploader_id': 'starwars',
             'uploader': 'Star Wars | Andor Premieres Sept 21 on Disney+',
@@ -271,7 +271,7 @@ class TwitterIE(TwitterBaseIE):
         'info_dict': {
             'id': '700207533655363584',
             'ext': 'mp4',
-            'title': 'jaydin donte geer - BEAT PROD: @suhmeduh  https://t.co/[...]',
+            'title': 'jaydin donte geer - BEAT PROD: @suhmeduh #Damndaniel',
             'description': 'BEAT PROD: @suhmeduh  https://t.co/HBrQ4AfpvZ #Damndaniel https://t.co/byBooq2ejZ',
             'thumbnail': r're:^https?://.*\.jpg',
             'uploader': 'jaydin donte geer',
@@ -298,7 +298,7 @@ class TwitterIE(TwitterBaseIE):
         'info_dict': {
             'id': '719944021058060289',
             'ext': 'mp4',
-            'title': 'Captain America - @King0fNerd Are you sure you made t[...]',
+            'title': 'Captain America - @King0fNerd ...',
             'description': '@King0fNerd Are you sure you made the right choice? Find out in theaters. https://t.co/GpgYi9xMJI',
             'uploader_id': 'CaptainAmerica',
             'uploader': 'Captain America',
@@ -338,7 +338,7 @@ class TwitterIE(TwitterBaseIE):
         'info_dict': {
             'id': '910031516746514432',
             'ext': 'mp4',
-            'title': 'Préfet de Guadeloupe - [Direct] #Maria Le centre se trouve[...]',
+            'title': 'Préfet de Guadeloupe - [Direct] #Maria Le centre se trouve...',
             'thumbnail': r're:^https?://.*\.jpg',
             'description': '[Direct] #Maria Le centre se trouve actuellement au sud de Basse-Terre. Restez confinés. Réfugiez-vous dans la pièce la + sûre. https://t.co/mwx01Rs4lo',
             'uploader': 'Préfet de Guadeloupe',
@@ -356,7 +356,7 @@ class TwitterIE(TwitterBaseIE):
         'info_dict': {
             'id': '1001551623938805763',
             'ext': 'mp4',
-            'title': 'Lis Power - Holy shit, Shep is on a roll today.[...]',
+            'title': 'Lis Power - Holy shit, Shep is on a roll today....',
             'thumbnail': r're:^https?://.*\.jpg',
             'description': 'md5:37b9f2ff31720cef23b2bd42ee8a0f09',
             'uploader': 'Lis Power',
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/twitter.py
```diff
@@ -440,14 +456,17 @@ def _real_extract(self, url):
...
         uploader = user.get('name')
-        if uploader:
-            title = '%s - %s' % (uploader, title)
         uploader_id = user.get('screen_name')
+        timestamp = unified_timestamp(status.get('created_at'))
+
+        # Keep the file name short so it doesn't exceed filesystem limits
+        title = self._generic_title(url)
+        if uploader_id:
+            title = '@%s-%s' % (uploader_id, title)
```
dirkf, line 469: I'd rather use the previous title == description and just truncate it, or otherwise abbreviate it. Thus:
* pick a maximum length that will be reasonable to use in the default output template `%(title)s-%(id)s.%(ext)s`, say 40;
* pick an ellipsis string, say `[...]`;
* if the putative title resulting from old line 445 is longer than the maximum length, replace n characters either at the end or in the middle of the putative title by the ellipsis string, where n is the excess length of the putative title plus the length of the ellipsis string;
* if the putative title will have `uploader - ` prepended, add the length of `uploader` plus 3 to n.

Eg:
```py
        title = description = status['full_text'].replace('\n', ' ')
        # strip  'https -_t.co_BJYgOjSeGA' junk from filenames
        title = re.sub(r'\s+(https?://\S+)', '', title)
        # Get metadata
        user = status.get('user') or {}
        uploader = user.get('name')

        # Keep the file name short so it doesn't exceed filesystem limits

        def abbrev(field, size, where=None, ellipsis_str='[...]'):
            ellipsis_len = len(ellipsis_str)
            where = (size - ellipsis_len + 1) // 2 if where is None else (where - ellipsis_len)
            to_strip = len(field) - size + ellipsis_len
            if to_strip > 0:
                rest_len = len(field) - where - to_strip
                return field[:where] + ellipsis_str + (field[-rest_len:] if rest_len > 0 else '')
            return field

        MAX_TITLE_LEN = 40
        if uploader:
            # also limit length of uploader
            uploader_n = abbrev(uploader, 12, 12, '...')
            title = '%s - %s' % (uploader_n, abbrev(title, MAX_TITLE_LEN - len(uploader_n) - 3))
        else:
            title = abbrev(title, MAX_TITLE_LEN)
```

palewire: Copy. I will make this change and let you know when it's done

palewire:
@dirkf, I believe I have achieved your request with a little less code. The uploader name is pretty strictly limited by Twitter, so I don't think we need to slice it. That means we can use a more straightforward slice like the one just pushed. Give it a look and let me know what you think.

youtube_dl/extractor/twitter.py
```diff
@@ -440,14 +456,19 @@ def _real_extract(self, url):
...
+        description = status['full_text'].replace('\n', ' ')
         user = status.get('user') or {}
         uploader = user.get('name')
+        uploader_id = user.get('screen_name')
+        timestamp = unified_timestamp(status.get('created_at'))
+
+        # Keep the file name short so it doesn't exceed filesystem limits
+        title = description
+        if len(title) > 40:
+            title = title[:35] + "[...]"
```
dirkf, line 469: If you're going to elide the end, I think `...` (not an actual Unicode ellipsis character) will be fine for the elision marker. Also, single quotes.

Looking at these test cases, it could be worth adding a heuristic to break at whitespace, say:
```suggestion
            trim = 35
            # allow up to 10 more characters to find whitespace 
            m = re.search(r'\s+', title, trim, trim + 10)
            if m:
                trim = m.start() + 1
            title = title[:trim] + '...'
```

BTW see #29989 for my general attempt to fix this problem that hasn't quite got off the ground.

dirkf (changes requested):
Thanks, LGTM, with comments as below. I'll enable the CI tests so that you can check the results.

youtube_dl/extractor/twitter.py
```diff
@@ -440,14 +456,19 @@ def _real_extract(self, url):
                 'tweet_mode': 'extended',
             })
 
-        title = description = status['full_text'].replace('\n', ' ')
-        # strip  'https -_t.co_BJYgOjSeGA' junk from filenames
-        title = re.sub(r'\s+(https?://[^ ]+)', '', title)
```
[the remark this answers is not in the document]
palewire: Copy. I've worked that back in.

palewire:
@dirkf, those changes are in.

palewire:
The flake errors in the test run appear to be in other files.

dirkf:
Apparently the default _flake8_ settings that the project uses changed in some way so that a keyword must be separated from the next token by a space, which is basically good but irritating. Otherwise GH somehow removed some spaces from the code, which seems unlikely and would be much worse.

If you update from the current master the 2 places where this caused problems are fixed.

palewire:
I think I've done that.

pukkandan:
imo, this should not be implemented.
* Would all extractors be expected to do this?
* title != filename. We shouldn't extract wrong metadata just because it is convenient for filenames. Output template should be used to control filename length instead. This is a poorly implemented bandaid. 

As I see, changing the default outtmpl to something like `%(title).200s-%(id)s.%(ext)s` would be much better than this solution.

dirkf:
This extractor is unusual in that there is no title in the plain webpage, although a title and an `og:title` are set by JS. The full tweet text is extracted in the `description`. So:

>* Would all extractors be expected to do this?

No. But Twitter was the source of some of the original issues complaining about filename length errors with the default output template, so this is an opportunity to avoid those in future, without expecting users to check for a FAQ that tells them to use `%(title).100s` or whatever, and without changing the core code.

Instead of fixing the title, we could generate a warning if the title is "long", or possibly include the FAQ text in the error message if we could reliably detect an exception caused by an over-long filename (a generic "solution" in place of #29989).

Changing the default output template could be an alternative but I would be concerned that setting a sufficiently low value to avoid filename length issues would affect filenames on filesystems with no effective limit, attracting a different set of complaints. The extractor could manipulate the output template by inserting a length field into any `%(title)s` but that would weirdly affect all URLs in the download session.

>* title != filename. We shouldn't extract wrong metadata just because it is convenient for filenames. ...

I believe that the default output template was chosen on the assumption that `title` would be a short text field, for some definition of "short". Implicitly, yt-dl tries to cast all sites in the mould of YouTube, where the title is [(a) required (b) not longer than 100 characters](https://developers.google.com/youtube/v3/docs/videos#snippet), although SEO fiends recommend no more than 70.

Apparently the page title for a tweet set by Twitter has the form (in terms of the extracted fields) `{og_title}: "{clean_html(description)}"`, where `og_title`, the `og:title` meta field, is `{uploader_id} on Twitter`. We emulate that but remove ` on Twitter`, presumably because, where else would it be, given the context? A profile page has `{uploader_id} (@{uploader}) / Twitter`, but we don't have an extractor for those [^note]. 

[^note]: We do have an extractor for `twitter:card` but it redirects to an invalid URL and all the test URLs but the single `only_matching` URL give a "nothing to see here" page; even that gets 403 for the API call to list video details, as does the only similar URL known to G.

[Nitter tries to improve Twitter](https://github.com/zedeus/nitter/issues/124) by making a tweet title thus `{profile.fullname} (@{profile.username}): "{stripHtml(tweet.text)}"`, combining the two methods used by Twitter. Perhaps we should also do that, as `{uploader_id} (@{uploader}): "{clean_html(description)}"` (and stripping unwanted Twitter links from the `description`, as now).

In the browser title bar I see the Twitter title presented with `...` replacing as much of the middle of the text as is needed to make the text fit the title bar. Something like that still seems appropriate.

>Output template should be used to control filename length instead. This is a poorly implemented bandaid.

After all, yt-dl itself is just a bandaid in some sense. What would make for a *better* implementation? Surely even a harsh critic would call the current implementation at least "adequate"?

pukkandan:
> What would make for a better implementation? Surely even a harsh critic would call the current implementation at least "adequate"?

I gave one suggestion above. As you pointed out, that has it's own issues too. (I still think this is a better solution though). If a proper solution was easy, I doubt this issue would have existed for so long

From a user perspective, I do agree this is better than nothing. But I also think that it'll become a headache to maintain any kind of consistency b/w extractors over time.

And I disagree that the twitter is special in this regard. There are plenty of other extractors that face this issue. [This search](https://cs.github.com/ytdl-org/youtube-dl?q=%22%27title%27%3A+%27md5%3A%22) is a quick way to get some examples

That said, since twitter is so popular, I can understand why you'd consider this technical debt to be acceptable.

And for the record, while an FAQ entry might help some users, I don't think it will reduce the number of issues opened. The kind of people who actually read the FAQ are the same people who'd look for duplicate issues.

palewire:
If you really want to get existential about it, I think the core problem is requiring a title for all sites. As detailed above, there's not a tidy title for a tweet analogous to a YouTube title. So there will never be tidy answer to this problem.

That said making the title field optional seems like a can of worms that isn't worth opening either. Some kind of compromise is necessary and, as documented in the issue that prompted this pull request, the current solution is bad for users. So, IMHO, some alternative is needed. Something along the lines of what's proposed here does seem like the simplest solution. If universality is what you're after, I think enforcing a global trim on all title attributes seems like the most natural extension.

dirkf:
The rationale for `title` is presumably what I said above, trying to make all sites fit the YT metadata schema. yt-dlp has allowed a `None` title while continuing to require the `title` field without the sky falling.

>... If a proper solution was easy, I doubt this issue would have existed for so long

Amen to that. If we can get something like #29989 to work, that would fix the problem for all sites. Meanwhile:

>... since twitter is so popular, I can understand why you'd consider this technical debt to be acceptable.

Exactly. The Twitter title field is up for grabs: its extraction shouldn't routinely crash the program. I guess that extractor is used more than all the other possibly problematic extractors together, which are also a small proportion of the ~800 (crikey) we have.

And
>...while an FAQ entry might help some users, I don't think it will reduce the number of issues opened

Absolutely, why I mentioned putting some FAQ text in error message appears, if only we knew reliably what that would be.

palewire:
I'm withdrawing this request.

PR #31155, opened 2022-08-10
[Tele5] Improve Extractor

menschel:
Return playlist instead of single video info in case of multiple
player elements.

Example with two parts.

https://tele5.de/mediathek/merlin-2-der-letzte-zauberer/

---

- [X] Improvement

---

The purpose of this PR is to get ytdl to download all videos instead only the first one which is annoying at best.

A note to the guy who wrote that nice style guide flake8 check box.
I would not have dared to write that line looking at your regular code.
PyCharm in combination with pytest, pytest-flak8 and pytest-cov is a guarantee for code quality.

### DIFF
--- a/youtube_dl/extractor/tele5.py
+++ b/youtube_dl/extractor/tele5.py
@@ -1,6 +1,9 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
+import re
+from typing import Optional
+
 from ..compat import compat_urlparse
 from ..utils import (
     ExtractorError,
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/tele5.py
```diff
@@ -1,10 +1,13 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
+import re
+from typing import Optional
```
rautamiekka, line 5: The `typing` module only exists as of Python 3.5.0 but ytdl must work with 2.6.x-2.7.x and 3.2+, if you can believe the [`/setup.py`](https://github.com/ytdl-org/youtube-dl/blob/master/setup.py).

menschel:
To be say it bluntly, Python2 is officially dead since 1st January 2020
and we are currently at python 3.10.

IMO the whole code in project needs some serious rework and the setup.py
is no exception.

rautamiekka:
I agree, but ytdl decided to stay with Py2.6+. yt-dlp uses 3.7+ since just recently, you probly wanna consider that.

menschel:
Thanks for the hint, but yt-dlp's pukkandan got a clear message from me
that there will be no further PRs towards his/her/its projects after the
last incident.

Thank you, I guess that closes this PR.

dirkf:
> Thanks for the hint, but yt-dlp's pukkandan got a clear message from me that there will be no further PRs towards his/her/its projects after the last incident. Thank you, I guess that closes this PR.

By all means re-open the PR. There are only minor syntactical issues that would need to be fixed for yt-dl's platform conventions. I'm happy to provide those if the code is fundamentally working.

PR #31158, opened 2022-08-11
Add an option to discard style information from TTML subtitles

filip-hejsek:


dirkf:
I think this needs some examples. I couldn't see any issue in the tracker that this would fix - is there one, or maybe in yt-dlp?

If the code is left as-is, I'd suggest changing the sense of `convert_style=True` to `strip_style=False`.

But ...

Which target sttl formats support styles? Rather than introducing a new option, could we just invent new sttl format names (eg with `-plain` appended) to stand for the named format with no styles? 

Then tests!

filip-hejsek:
Thanks for your feedback.

There is no issue in the tracker for this. The problem this is trying to solve is that some players or devices (TVs) don't support HTML in SRT and just display the tags.

I don't know whether this is also an issue for the other subtitle formats.

One problem with my patch is that it confusingly only removes styles for subtitles converted from TTML. A proper solution would also add support for stripping the styling from unconverted subtitle files. However, that would expand the scope significantly to support many formats. Alternatively, only stripping from SRT could be supported.

So i think maybe it would be best to add a `srt-plain` format and add code to strip HTML styles from SRT files (which i'm willing to implement as it's mostly just a regex replace). Stripping styles from other formats could be implemented in the future if/when someone needs it.

What do you think?

dirkf:
Sorry for this delayed response. I think srt-plain isn't a bad idea. Note the `clean_html()` function in utils.py

fenopa:
@dirkf any progress on this? styles on ttml subtitles are really a problem (especially their font size thing)

Issue #31164, opened 2022-08-12
[YouTube] When running without --verbose, … (No terminating paren } in {var b=a.split(""),…


mk-pmb:
## Checklist

- [x] I'm reporting a broken site support
- [x] I've verified that I'm running youtube-dl version **2021.12.17**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar issues including closed ones

## Verbose log

Half no. Here's the version info from verbose mode:

```text
[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['--no-call-home', '--abort-on-error', '--no-overwrites', '--keep-video', '--fixup=warn', '--restrict-filenames', '--output', '%(upload_dat>
[debug] Encodings: locale UTF-8, fs utf-8, out utf-8, pref UTF-8
[debug] youtube-dl version 2021.12.17
[debug] Git HEAD: e6a836d54
[debug] Python version 3.8.10 (CPython) - Linux-5.8.0-44-lowlatency-x86_64-with-glibc2.29
[debug] exe versions: ffmpeg 4.2.4, ffprobe 4.2.4
```
When running without --verbose, [youtube] Unable to decode n-parameter: … (No terminating paren } in {var b=a.split(""),…

However, fortunately, using `--verbose` works around the problem. :+1:
With `--verbose`, I get a few dozen lines of debug messages and then the download starts.
When running without `--verbose`, I see

```text
[youtube] 37gJCuf6UMY: Downloading webpage
[youtube] 37gJCuf6UMY: Downloading player 324f67b9
WARNING: [youtube] Unable to decode n-parameter: download likely to be throttled (No terminating paren } in {var b=a.split(""),
c=[1070485609,7,function(d,e){d.push(e)},
```

and then the terminal is busy spewing lots more seemingly minified JS code, until I send SIGINT.

When I redirect stdout and stderr to a file (`… |& tee -- nparam.log`), it writes the first two lines and then seems stuck for about a minute, after which I gave up and sent SIGINT.

## Description

On Ubuntu focal, downloading from YouTube stopped, and instead, my terminal is flooded.
It worked fine a few hours ago.
Downloading from Twitch works as expected, so probably not a network problem.

Thanks for still maintaining compatible with ancient pythons!

dirkf:
[...]

PR #31170, opened 2022-08-14
[jsinterp] Overhaul JSInterp to handle new YT players 4c3f79c5, 324f67b9

dirkf:
<details>
<summary>Boilerplate: mixed code, bug fix</summary>

---

- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/), except for code from yt-dlp for which either this or the below has been asserted

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

</details>

---

Issue #31164 reported errors with new players served by YouTube. More complex JS constructs confused yt-dl's JSInterp mini-JS interpreter.

Thanks mostly to @pukkandan, this PR fixes the problem code by back-porting from https://github.com/pukkandan/yt-dlp-dev/commit/8f53dc44a0cc1c2d98c35740b9293462c080f5d0 into the existing previously back-ported framework.

Additionally:
* supports `void` used in player 324f67b9
* improves `<<`/`>>` precedence (TODO item in back-ported code), by preventing an operator search for `<`/`>` from matching the operator with the same first character
* improves processing of expressions in comma-list expressions, where the last item of a multi-item list wasn't fully processed
* allows for `let`, `const` in case they should appear
* allows for further comparison operators in case they should appear
* adds more tests
* allows for variable in `new Date(...)`
* improves argument list parsing and processing.

### DIFF
--- a/youtube_dl/jsinterp.py
+++ b/youtube_dl/jsinterp.py
@@ -1,42 +1,87 @@
 from __future__ import unicode_literals
 
+import itertools
 import json
+import math
 import operator
 import re
 
 from .utils import (
+    NO_DEFAULT,
     ExtractorError,
+    js_to_json,
     remove_quotes,
+    unified_timestamp,
 )
 from .compat import (
-    compat_collections_abc,
+    compat_collections_chain_map as ChainMap,
+    compat_itertools_zip_longest as zip_longest,
     compat_str,
 )
-MutableMapping = compat_collections_abc.MutableMapping
 
+_NAME_RE = r'[a-zA-Z_$][\w$]*'
 
-class Nonlocal:
-    pass
+# (op, definition) in order of binding priority, tightest first
+# avoid dict to maintain order
+# definition None => Defined in JSInterpreter._operator
+_DOT_OPERATORS = (
+    ('.', None),
+    # TODO: ('?.', None),
+)
 
-
-_OPERATORS = [
+_OPERATORS = (
     ('|', operator.or_),
     ('^', operator.xor),
     ('&', operator.and_),
     ('>>', operator.rshift),
     ('<<', operator.lshift),
-    ('-', operator.sub),
     ('+', operator.add),
-    ('%', operator.mod),
-    ('/', operator.truediv),
+    ('-', operator.sub),
     ('*', operator.mul),
-]
-_ASSIGN_OPERATORS = [(op + '=', opfunc) for op, opfunc in _OPERATORS]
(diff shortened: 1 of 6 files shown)


richardcox69 (commented):
Ok

youtube_dl/jsinterp.py
```diff
@@ -254,110 +349,116 @@ def raise_constructor_error(c):
...
         sub_expressions = list(self._separate(expr))
-        expr = sub_expressions.pop().strip() if sub_expressions else ''
-        for sub_expr in sub_expressions:
-            self.interpret_expression(sub_expr, local_vars, allow_recursion)
+        if len(sub_expressions) > 1:
+            for sub_expr in sub_expressions:
+                ret, should_abort = self.interpret_statement(sub_expr, local_vars, allow_recursion)
+                if should_abort:
+                    return ret, True
+            return ret, False
```
pukkandan, line 369: Could you explain to me the reasoning behind this change? All your tests pass in yt-dlp without this

youtube_dl/jsinterp.py
```diff
@@ -534,25 +648,30 @@ def extract_function_from_code(self, argnames, code, *global_stack):
...
         return self.build_function(argnames, code, local_vars, *global_stack)
 
     def call_function(self, funcname, *args):
         return self.extract_function(funcname)(args)
 
+    @classmethod
+    def build_arglist(cls, arg_text):
+        if not arg_text:
+            return []
+        return list(filter(None, (x.strip() or None for x in cls._separate(arg_text))))
```
pukkandan, line 663: Same. The `filter` doesn't make sense to me

youtube_dl/jsinterp.py
pukkandan: nvm, I figured it out

dirkf: > ...the last item of a multi-item list wasn't fully processed
Before: process list, pop last item, continue processing last item instead of list or processing last item from the beginning (where's the `goto` when you need it!).

youtube_dl/jsinterp.py
dirkf: Removing empty items, but there shouldn't be any:
```js
function fred(a,b,c) {
  return b;
}
fred(1, ,2);
/*
Exception: SyntaxError: expected expression, got ','
@Scratchpad/1:4
*/
function fred(a,,c) {
  return c;
}
/*
Exception: SyntaxError: missing formal parameter
@Scratchpad/1:1
*/
```
So probably just strip and let it crash elsewhere.

PR #31179, opened 2022-08-17
[Youtube] Parse horizontalListRenderer entries the same way as the gr…

Konubinix:
…idRenderer ones (horizontalListRenderer)

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

I was getting the album from Sonata Arctica here
https://www.youtube.com/channel/UCyGTmzFc3YzefTe8mOQ-QgA/playlists and realized
it did only get the created playlists, not the albums and singles.

By debugging the code, I realized the album and single parts are in a
horizontalListRenderer. Parsing this renderer was not yet implemented.

By looking at the content, I realized that apart from being of type
horizontalListRenderer, the items were in the same format as the ones I found in
a gridRenderer.

So I just used the same code for a few lines above that deal with a
gridRenderer. I definitely don't know enough to ensure it will always be the
case, so I'm just starting this pull request to start the discussion.

Also, I wanted to copy the tests about the gridRenderer to make tests about the
horizontalListRenderer, but I could not find any. I don't have the skill to
write tests on my own, so I might need so help there.

Thank you for the awesome youtube-dl.

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -2594,8 +2594,8 @@ class YoutubeTabIE(YoutubeBaseInfoExtractor):
                 yield entry
         renderer = content.get('horizontalListRenderer')
         if renderer:
-            # TODO
-            pass
+            for entry in self._grid_entries(renderer):
+                yield entry
 
     def _shelf_entries(self, shelf_renderer, skip_channels=False):
         ep = try_get(


dirkf (commented):
Thanks for this idea!

I was going to ask you to check against the equivalent yt-dlp code (to save me the effort) but the TODO is identical.

Blame tells me that no-one has changed this since Sergey nearly 2 years ago and he had some similar code before that, which was replaced.

As an almost never user of the YT site I'm "uniquely qualified" to maintain this area of code. The change looks fine as code but we ought to understand what effect is being achieved.

One way to clarify this is to add and/or change tests (does the change affect exiting tests?); whatever, test coverage is needed.

Another is to invoke the spirits of @pukkandan and @coletdjnz for advice, which I hereby do, and any other help would be welcome.

coletdjnz:
Hm. These are shelves which have a direct link to get all the items, so you should prefer that than extracting from the horizontelListRenderer (which will only give a small subset of the actual results). This is what is currently happening.

```
yt-dlp --print url --flat-playlist https://www.youtube.com/channel/UCyGTmzFc3YzefTe8mOQ-QgA/playlists 
https://www.youtube.com/channel/UCyGTmzFc3YzefTe8mOQ-QgA/playlists?view=1&sort=dd&shelf_id=0
https://www.youtube.com/channel/UCyGTmzFc3YzefTe8mOQ-QgA/playlists?view=50&sort=dd&shelf_id=17666223384013636041
```

The issue in this case is the url for "Albums & Singles" is incorrect - it leads back to the same page. It seems to be inconsistent too. The correct url is `https://www.youtube.com/c/sonataarctica/playlists?view=71&shelf_id=507`, which appears to show up under the drop down menu? 

Seems to be some sort of YouTube bug, maybe something to do with the topic channel handling. If you click on the albums and singles it will take you to the same page but on the different channel.

(Also I notice that shelf_entries always try to extract entries from content when it has found a url. Not sure if that is intended, but could lead to duplicate entries.)

dirkf:
For comparison with another InnerTube application, at an Invidious site the shelf_id parameter added to the `/channel/UCyGTmzFc3YzefTe8mOQ-QgA/playlists` path has no effect. All such URLs just show the band bio and 9 playlists.

The `/c/sonataarctica/playlists?view=71&shelf_id=507` path resolves to `/channel/UC56F8c8khYkaA42xQhoS_gw/playlists` (and generally `UC56F8c8khYkaA42xQhoS_gw` is the resolution of `/c/sonataarctica`), showing what might be 2 albums and 2 singles playlists.

Prototaxis:
This could be caused by the url system failing to parse the long view_id

dirkf:
> This ...
This = ?

89z:
FYI here is a quote from @coletdjnz in another forum:

> youtube deploying broken shit to prod like usual

I thought maintainer here would be interested to know. I have dealt with many a difficult character in my times online, but the absolute worst type is the two face.

Untitled

Konubinix:
Hi,
dirkf ***@***.***> writes:

[...]

> One way to clarify this is to add and/or change tests (does the change affect exiting tests?); whatever, test coverage is needed.

I would gladly do that, but I could not find tests that deal with the
format of the shelves to use as example.

Could you provide me with a pseudo test example to help me write the appropriate
tests.

-- 
Konubinix
GPG Key    : 7439106A
Fingerprint: 5993 BE7A DA65 E2D9 06CE  5C36 75D2 3CED 7439 106A

Prototaxis:
> > This ...
> 
> This = ?

_if you click on albums and singles under the drop down, it will take you too the correct view under the topic channel, under the url https://www.youtube.com/channel/UCyGTmzFc3YzefTe8mOQ-QgA/playlists?view=50&shelf_id=1766622338401363604_

Konubinix:
Thank you all for your answers.

As you understand, my final goal here is to be able to have all the known albums
of an artist, provided with its channel url. Having horizontalListRenderer
working is only an instrumental goal here.

So, here I'm focusing a little bit more on that final goal, so that I can adapt
the pull request accordingly.

When using https://www.youtube.com/channel/UCyGTmzFc3YzefTe8mOQ-QgA/playlists,
youtubedl only extracts the so called "Created playlists" and not the "Albums &
Single". I found this out because I heard about the album "Acoustic Adventures -
Volume One" and realized it was not exported by youtube-dl.

By debugging the code, I realized that "Acoustic Adventures - Volume One" was
inside a document called horizontalListRenderer. That lead me to search why it
was not extracted and I eventually found out the part of the code that is the
subject of the PR.

If I understand correctly, @coletdjnz, you discourage the use of
horizontalListRenderer because it brings out only partial data. I had the
feeling that all the musics where listed when I tried it. Can you be more
specific about what should be missing?

Also, you point out that the UI of youtube is pretty weird and inconsistent and
suggest that we might not be able to do anything related to this issue.  I may
have to look at the code of youtube-dl to get my own feeling, but from a naive
point of view, I suppose that whatever youtube's UI does can be simulated as
well in youtube-dl. The fact that the UI is not consistent means that the code
will have to absorb this inconsistency. Thus, it looks to me like this makes the
data extraction more complicated, but not impossible, I mean theoretically.

Or maybe what you are suggesting is that the code to deal with this will be too
complicated to be worth it? As far I am concerned, I agree that I might not be
able to provide a pull request to do this on my own.

But, with a little help from guys like you that have experience, do you think it
would be possible?

Konubinix:
I spend some time with the code and playing with youtube and here is what I think
1. @coletdjnz I agree with you. It's quite a mess and it is likely not worth trying to simulate what youtube does. That is very unfortunate though...
2. when clicking on the singsle & albums link, the url followed is "https://www.youtube.com/pagead/viewthroughconversion/962985656/?backend=innertube&cname=1&cver=2_20220829_00_00&data=backend%3Dinnertube%3Bcname%3D1%3Bcver%3D2_20220829_00_00%3Bptype%3Dcview%3Btype%3Dcview%3Butuid%3D56F8c8khYkaA42xQhoS_gw&foc_id=56F8c8khYkaA42xQhoS_gw&label=followon_cvisit&ptype=cview&utuid=56F8c8khYkaA42xQhoS_gw", entailing your assumption. It sounds like it is a innertube migration stuff
3. by looking at the horizontalListRenderer, one can actually find the browse_id=UC56F8c8khYkaA42xQhoS_gw that, if you go to /channel/UC56F8c8khYkaA42xQhoS_gw, redirects to /c/sonataarctica
4. by playing with [innertube](https://pypi.org/project/innertube/), I find the same json that the one youtube-dl gets
5. looking at the content of the horizontalListRenderer, I actually found out that it was not thourough, as you suggested,

Therefore, I assume that there is not enough information in the youtubemusic url to get all the albums of a given artist, one should have to first find the appropriate channel (in this case UC56F8c8khYkaA42xQhoS_gw) before extracting the data.

As a conclusion, I suggest I close this pull request. 
To reach my final goal, which is to be able to list the albums from the youtubeMusic url, I will use an ad hoc script using innertube to find the appropriate browse_id, but that sounds to be out of the scope of youtube-dl.

7.

coletdjnz:
Yeah as I mentioned it's a bug on YouTube's end, which will hopefully be fixed shortly (unless they consider this a feature). 

A workaround in youtube-dl/p will likely require very specific handling for the bug as you have alluded to, rather than something generic, to avoid extraction issues on shelves that are not buggy.

Konubinix:
Thank you for the discussion. 

So far, as long as I can find (even with an adhoc way) to the "albums & songs" list, that's not ideal but that's ok.

However, and going slightly out of scope, if by chance you know how to get the "albums & songs" list from 2Cellos, that would be very much appreciated 

There is an horizontalListRenderer in there YTmusic page (https://music.youtube.com/channel/UCxwLkfMCfx2uSKO9w4iRfDQ) but the other one (taken using browse_id or simply clicking on "albums & singles") does not contain it (https://www.youtube.com/channel/UCxwLkfMCfx2uSKO9w4iRfDQ/playlists?view=50&sort=dd&shelf_id=17666223384013636040)

Konubinix:
Also, after having tried to get the redirection working for several artists using innertube, I decided to give up. There are so many particular cases to deal with that the task is pretty unbearable to me. Sometimes the browse_id works, some other times I need to follow something else.

So yeah, that was definitely a dead end, at least as far as I can understand.

Prototaxis:
Implementing protoc into youtube-dl and requesting the shelf/grid directly from innertube is the cleanest solution I can think of.
https://www.youtube.com/channel/UCxwLkfMCfx2uSKO9w4iRfDQ/playlists?view=50&shelf_id=17666223384013636041
```py
{
context: {...},
browseId: "UCxwLkfMCfx2uSKO9w4iRfDQ", #channelId
params: protobuf.encode.raw({ #a concept function which encodes a base64 protobuf from a raw object
    2: "playlists", #tab
    4: 50, #view
    14: 17666223384013636041 #shelf_id
 })
}
```
The current issue I see with the url system failing to navigate to the correct playlist section is that it fails to correctly parse the long shelf_id which results in an incorrect request, whereas the dropdown menu on the /playlist page doesn't rely on the url and requests the section directly from innertube, hence the correct result. By generating a raw protobuf parameter we bypass the faulty url system and reliably get the section response.

Konubinix:
@Prototaxis Thank you for the information. Where did you find out that the params part was protobuf encoded?

Prototaxis:
> @Prototaxis Thank you for the information. Where did you find out that the params part was protobuf encoded?

@Konubinix It's common knowledge. Most of the parameters used by innertube are base64 protobuf encoded

Konubinix:
Thank you for the information @Prototaxis

Issue #30661, opened 2022-02-21
Download webpage after executing JavaScript scripts


DarkFighterLuke:
## Checklist

- [x] I'm asking a question
- [x] I've looked through the README and FAQ for similar questions
- [x] I've searched the bugtracker for similar questions including closed ones

## Question

I am trying to make a new extractor and I would like to know if (and eventually how) is possible to download a webpage after executing its own javascript scripts. This because of some fields (in my case the video url) which are loaded by the scripts and then, if you download the page without executing them (just like wget does), you'll get an empty url.

Thanks.

89z:
please provide example URL

DarkFighterLuke:
> please provide example URL

https://sxyprn.com/post/5bb443225d5ae.html try this

dirkf:
First, don't worry about that site. It's handled by the YourPorn extractor (or can be be with some tweaks that I have at hand).

As to the larger question that you ask, the answer is that [the tools don't exist](https://github.com/ytdl-org/youtube-dl/pull/30643#issuecomment-1046339547) to enable it.

Actually, these days it's difficult enough to get even a browser that isn't the latest Chrome to handle general web JS. However, there are certain cases where we do need to run some specific JS for an extractor (so far, YouTube itself) and the JSInterpreter module enables that without introducing new dependencies; but it's always at the risk that some site for which it's used suddenly deploys a syntax that we haven't allowed for (or even new syntax that breaks all previous JS interpreters, like optional chaining or ES2018 regex literals).

Two extractors use the PhantomJS interpreter mentioned in the link, but that introduces platform and configuration problems, and I don't really know if it works since I don't have it installed.

DarkFighterLuke:
Thank you for the answer @dirkf , it was all I implicitly wanted to know. I'm happy that is possible to deal with pages using JS using youtube-dl (and i should have known since YouTube uses it).

From your answer I see that an extractor for that site already exists and from what I tried it works.
I am curious about the tweaks you have in mind.

PR #31187, opened 2022-08-20
[options] Added workaround option to execute "n_function"

u-n-k-n-o-w-n:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [x] New feature

---

In regards to the recent youtube "n_function" update, if another fancy way comes up that the current jsinterp can't handle, this patch can give users a workaround until it is fixed in jsinterp.
But of course, you need to have selenium installed($ pip install selenium) and webdriver for the browser you want to use. :p
- [options] Added workaround option to execute "n_function"


dirkf:
If we're doing this, let's align it with the PhantomJS support in yt-dlp so that either package can be used. Other extractors might have a use for a JS interpreter (currently PhantomJS has a class in the OpenLoad extractor module).

There should be a function that runs a JS function and returns the result plus some way of determining which "engine" to use, based on options and defaults.

What do you think?

u-n-k-n-o-w-n:
You mean to create something like WebDriverJSwrapper?
I think it is a good idea to make the parts universally usable. :)

P.S.
In my personal opinion, I would not recommend using phantomjs.

phantomjs is no longer maintained, and the executable that can be downloaded from the official website is statically linked to the 2016 webkit. In other words, the webkit vulnerabilities discovered since 2016 have not been fixed. Executing externally obtained JS is inherently dangerous, but even more so with a JS engine like this one.

Before that, phantomjs can read and write files and execute processes from JS, so the game is over when you load externally obtained JS. I don't think youtube would do such a thing, but I am not confident enough to say that all video sites would not do it.

When using selenium's webdriver, the JS engine is usually the browser the user normally uses, so there should be no increased risk. Also, even the average user is likely to update their browser, so it is always guaranteed that the latest JS engine can be used with the same level of security as the browser. For example, if a new JS syntax is defined, you do not need to consider it as long as you use the browser with webdriver.

Anyway, if you are going to use an external JS engine, you should use one that is maintained. :p

dirkf:
The [lack of a plausible alternative](https://github.com/ytdl-org/youtube-dl/issues/30661#issuecomment-1047297990) is why we have the built-in mini-JS interpreter. In some applications there is no "browser the user normally uses".

Also mentioned [here](https://github.com/yt-dlp/yt-dlp/commit/600f09249f64f0318ddd831032b47c1a05637165#r81105697) is this more plausible https://github.com/amol-/dukpy

u-n-k-n-o-w-n:
Yes, the best thing to do is to continually improve jsinperp so that it can do whatever the browser can do.
The original intent of this patch was
> this patch can give users a workaround until it is fixed in jsinterp.

so I totally agree with you on that point.

> You mean to create something like WebDriverJSwrapper?
> I think it is a good idea to make the parts universally usable. :)

Implemented at https://github.com/ytdl-org/youtube-dl/pull/31187/commits/3038610951027d76d8255030b4f80e1b44126126.

youtube_dl/extractor/youtube.py
```diff
@@ -1499,6 +1503,27 @@ def _extract_n_function(self, video_id, player_url):
...
+             'const d = decodeURIComponent(e);'
+             'const p = d.lastIndexOf("}}");'
+             'const th = d.substring(0, p);'
+             'const bh = d.substring(p);'
+             'const m = "var {0};" + th + ";{0} = {1};" + bh;'
+             'const s = document.createElement("script");'
+             's.innerHTML = m;'
+             'document.body.append(s);'
+             'return {0}("{2}");'
+             '}})("{3}");').format(dummyfunc, funcname, n_param, compat_urllib_quote(jscode))
```
pukkandan, line 1523: Doesn't this run the entire js?

pukkandan:
If we want to use an external binary, wouldn't `node` be better than selenium?

youtube_dl/extractor/youtube.py
u-n-k-n-o-w-n: Yes, when the JS is loaded, an anonymous function is called with _yt_player as an argument.
There is a lot of waste, but this is what the browser normally does when browsing youtube.
Also, the JS that allows the desired function to be referenced from the outside is inserted at the time of this call, so there is no need for complicated parsing. :)

u-n-k-n-o-w-n:
@pukkandan 
With Selenium, the browser's JS engine can be used simply by installing webdriver.
Most people probably use browsers as well, so there is often no need to go to the trouble of installing another JS engine.

In addition, since nodejs was originally designed for writing regular programs in JS, it does not seem like a good idea to pass externally derived JS to be executed(child_process, fs, etc...).
But I am not familiar with nodejs, so if nodejs can run as safely as the various restrictions that browsers impose on JS engines, then we would not need to consider this issue.

pukkandan:
> With selenium, the browser's JS engine can be used simply by installing webdriver.
> Most people probably use browsers as well, so there is often no need to go to the trouble of installing another JS engine.

youtube-dl is often used in servers/containers/embedded devices where a browser is not normally available. And installing chrome/firefox on a headless system is not trivial

PS: Also, puppeteer is generally preferred over selenium nowadays

u-n-k-n-o-w-n:
> youtube-dl is often used in servers/containers/embedded devices where a browser is not normally available. And installing chrome/firefox on a headless system is not trivial

Yes, so this is only an workaround.
My goal is not to help all users, but to provide workarounds, even if only to a subset of users.

> PS: Also, puppeteer is generally preferred over selenium nowadays

Thanks, is [this](https://github.com/pyppeteer/pyppeteer) it?
I'll check it out. :)

youtube_dl/options.py
```diff
@@ -572,6 +572,9 @@ def _comma_separated_values_options_callback(option, opt_str, value, parser):
             'Upper bound of a range for randomized sleep before each download '
             '(maximum possible number of seconds to sleep). Must only be used '
             'along with --min-sleep-interval.'))
+    workarounds.add_option(
+        '--webdriver', metavar='TYPE', dest='webdriver', default=None,
+        help='Specify webdriver type when you want to use selenium webdriver to execute "n_function" : "firefox", "chrome", "edge", or "safari"')
```
rautamiekka, line 577:
```suggestion
        help='Specify webdriver type when you want to use Selenium to execute YouTube's "n_function" in order to avoid throttling: "firefox", "chrome", "edge", or "safari"')
```

dirkf:
Now that _Deno_ (addressing security issues associated with node.js) exists, I'm more favourably inclined to this functionality. If we can ensure that it works like _yt-dlp_ (which does support _Deno_), it should be merged.

PR #31213, opened 2022-08-29
[youtube] add published_time for --flat_playlist JSON

blueowl04:
When getting JSON using "--flat-playlist", it misses any information about video age. It is very desirable to have such information. 
This PR adds "published_time" to the JSON.

Example:
  youtube-dl --flat-playlist -j ytsearch10:"egg recipe"

### DIFF
--- a/youtube_dl/extractor/youtube.py
+++ b/youtube_dl/extractor/youtube.py
@@ -330,6 +330,8 @@ class YoutubeBaseInfoExtractor(InfoExtractor):
             renderer,
             (lambda x: x['ownerText']['runs'][0]['text'],
              lambda x: x['shortBylineText']['runs'][0]['text']), compat_str)
+        published_time = try_get(
+            renderer, lambda x: x['publishedTimeText']['simpleText'], compat_str) or ''
         return {
             '_type': 'url',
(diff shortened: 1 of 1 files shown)


dirkf:
It would be useful. But...

The text value should be converted to a timestamp, as is done in yt-dlp: however there are several dependencies of that code; we'd have to pull in some further methods of YoutubeBaseInfoExtractor as well as some non-trivial routines from utils.py.

As other PRs need those routines, I'll try to get them merged to simplify the job.

blueowl04:
Well, it would definitely be very good to have a timestamp. Unfortunately, Youtube's publishedTimeText just contains broad terms like "5 weeks ago" or "4 months ago". So converting that to a timestamp doesn't make much sense. Anyway, it is still very useful to add this information (even as text) to know how old the video is.
(If there is a more precise date/time elsewhere it would be great).

dirkf:
The `publishedTimeText` can provide the highest resolution that YT offers if the video is new ("1 hour ago", eg). When each video page is actually processed there is an `upload_date`, but I believe the pTT is the only field available when the playlist is being processed.

Anyhow, converting this relative time text to a datetime is what the yt-dlp code that I mentioned does.

PR #31277, opened 2022-10-04
Fix Crucial guide command leads into errors

miguelgargallo:
At university, colleagues and I tried this tool, but all of us got the error at this point:

  youtube-dl [OPTIONS] URL [URL...]

  youtube-dl -f 'bestvideo[ext=mp4]+bestaudio[ext=m4a]/best[ext=mp4]/best' URL https://www.youtube.com/watch?v=Toon9VARRIU

youtube-dl-miguelgargallo-contribution

ERROR: You've asked youtube-dl to download the URL "URL". That doesn't make any sense. Simply remove the parameter in your command or configuration. Add -v to the command line to see what arguments and configuration youtube-dl got.

So, to avoid future devs into this error... we love the project and want to improve it!

By just removing the "URL" word.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
 # DESCRIPTION
 **youtube-dl** is a command-line program to download videos from YouTube.com and a few more sites. It requires the Python interpreter, version 2.6, 2.7, or 3.2+, and it is not platform specific. It should work on your Unix box, on Windows or on macOS. It is released to the public domain, which means you can modify it, redistribute it or use it however you like.
 
-    youtube-dl [OPTIONS] URL [URL...]
+    youtube-dl [OPTIONS] [URL...]
 
 # OPTIONS
     -h, --help                           Print this help text and exit


rautamiekka:
You misunderstood it: if you remove `URL`, the command now reads "optionally enter >=1 URL to use" instead of ">=1 URL required to use".

gamer191:
How about `youtube-dl [OPTIONS] URLS`

gamer191:
> You misunderstood it: if you remove `URL`, the command now reads "optionally enter >=1 URL to use" instead of ">=1 URL required to use".

Technically that's correct though, depending on which options are passed (eg: `--update`, `--batch-file FILE`, `--list-extractors`)

rautamiekka:
In that sense it's technically correct, true.

miguelgargallo:
Yeah! Totally true @rautamiekka 
I modified into #ec1c85432c0e171e9d4570dd7075d47d52f90f99

Lesmiscore:
I think you'll also want to modify here,

https://github.com/ytdl-org/youtube-dl/blob/d35557a75d943865e40410d51bfcc18276e98532/youtube_dl/options.py#L125

as youtube-dl gives:

```
$ youtube-dl
Usage: youtube-dl [OPTIONS] URL [URL...]

youtube-dl: error: You must provide at least one URL.
Type youtube-dl --help to see a list of all options.
```

dirkf:
Yes, I can just about see how the original help synopsis might be confusing. Anyhow it's just wrong. The options don't have to precede the URLs and the URLs don't have to be provided if the options mean that yt-dl doesn't expect them.

_curl_, which has a somewhat similar UI, has the right formulation. `man curl`:
```
curl(1)                           curl Manual                          curl(1)

NAME
       curl - transfer a URL

SYNOPSIS
       curl [options / URLs]

...
```

PR #31303, opened 2022-10-22
[netease] Add warnings for users outside of China

xyb:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Add warnings for users outside of China, add more diagnostic information, and give suggested advice.

### DIFF
--- a/youtube_dl/extractor/neteasemusic.py
+++ b/youtube_dl/extractor/neteasemusic.py
@@ -95,7 +95,7 @@ class NetEaseMusicBaseIE(InfoExtractor):
         url = 'https://interface3.music.163.com/eapi/song/enhance/player/url'
         data, headers = self.make_player_api_request_data_and_headers(song_id, bitrate)
         try:
-            msg = 'empty result, maybe you should use a proper proxy server located in China'
+            msg = 'empty result'
             result = self._download_json(
                 url, song_id, data=data.encode('ascii'), headers=headers)
             if result:
@@ -106,11 +106,12 @@ class NetEaseMusicBaseIE(InfoExtractor):
                 raise
         except Exception as e:
             msg = error_to_compat_str(e)
-            self.report_warning('API call (%s, %s, %s) failed: %s' % (
-                url, song_id, bitrate, msg))
+            self.report_warning('%s API call (%s) failed: %s' % (
+                song_id, bitrate, msg))
         return {}
 
     def extract_formats(self, info):
+        err = 0
         formats = []
         song_id = info['id']
         for song_format in self._FORMATS:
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/neteasemusic.py
```diff
@@ -94,14 +95,19 @@ def _call_player_api(self, song_id, bitrate):
         url = 'https://interface3.music.163.com/eapi/song/enhance/player/url'
         data, headers = self.make_player_api_request_data_and_headers(song_id, bitrate)
         try:
-            return self._download_json(
+            msg = 'empty result, maybe you should use a proper proxy server located in China'
```
dirkf, line 98: We have a method that tells people this, that can be called later.
```suggestion
            msg = 'empty result'
```

youtube_dl/extractor/neteasemusic.py
```diff
@@ -94,14 +95,19 @@ def _call_player_api(self, song_id, bitrate):
...
         except ExtractorError as e:
             if type(e.cause) in (ValueError, TypeError):
                 # JSON load failure
                 raise
-        except Exception:
-            pass
+        except Exception as e:
+            msg = error_to_compat_str(e)
+            self.report_warning('API call (%s, %s, %s) failed: %s' % (
+                url, song_id, bitrate, msg))
```
dirkf, line 110: Actually `url` is always the same, and other messages put the video/song ID first.
```suggestion
            msg = error_to_compat_str(e)
            self.report_warning('%s API call (%s) failed: %s' % (
                song_id, bitrate, msg))
```

youtube_dl/extractor/neteasemusic.py
```diff
@@ -94,14 +95,19 @@ def _call_player_api(self, song_id, bitrate):
...
         return {}
 
     def extract_formats(self, info):
```
dirkf, line 113:
```suggestion
    def extract_formats(self, info):
        err = 0
```

youtube_dl/extractor/neteasemusic.py
```diff
@@ -125,6 +133,8 @@ def extract_formats(self, info):
                         'filesize': int_or_none(song.get('size')),
                         'asr': int_or_none(details.get('sr')),
                     })
+        if not formats:
+            self.report_warning('Maybe you should use a proper proxy server located in China')
```
dirkf, line 137: If the site returns an error code (assume semantics like HTTP codes) and no formats, tell the user.

If no code and no formats, assume geo-blocking. The diagnostic will look like this:
```
ERROR: No media links found: probably this video is not available from your location due to geo restriction
This video is available in China.
You might want to use a VPN or a proxy server (with --proxy) to workaround.
```
If any other regions are served (HK? other E Asian countries?), their ISO 2-letter codes should be added to the list of countries.
```suggestion
                elif err == 0:
                    err = try_get(song, lambda x: x['code'], int)

        if not formats:
            msg = 'No media links found'
            if err != 0 and (err < 200 or err >= 400):
                raise ExtractorError(
                    '%s (site code %d)' % (msg, err, ), expected=True)
            else:
                self.raise_geo_restricted(
                    msg + ': probably this video is not available from your location due to geo restriction.',
                    countries=['CN'])

```

dirkf (changes requested):
We should use the `_raise_geo_restricted()` IE method and could also report if an error code was returned in the JSON.

dirkf:
Thanks!

Issue #28748, opened 2021-04-13
Telegram


QBcrusher:
## Checklist

- [ x] I'm reporting a new site support request
- [ x] I've verified that I'm running youtube-dl version **2021.04.07**
- [ x] I've checked that all provided URLs are alive and playable in a browser
- [ x] I've checked that none of provided URLs violate any copyrights
- [ x] I've searched the bugtracker for similar site support requests including closed ones

## Example URLs

- Single video: https://t.me/r_Damnthatsinteresting/23326
- Channel https://t.me/s/r_Damnthatsinteresting

## Description

Requesting support for native Telegram videos. Channels can be displayed in browser, groups cannot but being able to crawl all of the videos uploaded on a channel (instead of using Telegram's export tool which kindof blows) would be pretty sweet.

mayeaux:
Bro there is like no way I can find to do this online :'( wtf. I don't want to have to forward a link to a bot that's insane I just want to download the url directly..

89z:
https://github.com/ytdl-org/youtube-dl/issues/23376

89z:
@mayeaux please provide example link

mayeaux:
> @mayeaux please provide example link

https://t.me/disclosetv/5934

89z:
With all due respect, I am not helping anyone to get that bullshit. If you have another video thats not someone spreading conspiracy theories, I am glad to help.

mayeaux:
> With all due respect, I am not helping anyone to get that bullshit. If you have another video thats not someone spreading conspiracy theories, I am glad to help.

https://t.me/disclosetv/5947

baptx:
Related: https://github.com/yt-dlp/yt-dlp/issues/2910

PR #31335, opened 2022-11-07
[Telegram] Add new extractor

fluks:

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

Add very rudimentary extractor for Telegram. Supports only public and channels which are previewable with browser.

Implements #28748.

### DIFF
--- a/youtube_dl/extractor/telegram.py
+++ b/youtube_dl/extractor/telegram.py
@@ -1,0 +1,60 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import re
+
+from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    int_or_none,
+)
+
+
+class TelegramIE(InfoExtractor):
+    _VALID_URL = r'https://t\.me/(?P<user>[^/]+)/(?P<id>\d+)'
+    _TEST = {
+        'url': 'https://t.me/telegram/195',
+        'info_dict': {
+            'id': '195',
+            'ext': 'mp4',
+            'title': 'telegram',
+            'description': 'Telegram’s Bot Documentation has been completely overhauled –\xa0adding the latest info, along with detailed screenshots and videos.\n\nNewcomers now have an easy way to learn about all the powerful features, and can build a bot from our step-by-step tutorial with examples for popular programming languages.\n\nExperienced developers can explore recent updates and advanced features, ready for 2022 and beyond.',
+            'duration': 23,
+        },
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        m = re.match(r'https://t\.me/(?P<channel>[^/]+)/', url)
+        if m is None:
+            raise ExtractorError('Unable to find channel name')
+        title = m.group('channel')
+        embed_url = url + '?embed=1&mode=tme'
+        html = self._download_webpage(embed_url, video_id)
+
+        video_url = self._search_regex(r'<video src="([^"]+)"', html, 'video_url')
+        formats = [{'url': video_url}]
+
+        duration = self._search_regex(
+            r'<time class="message_video_duration.*?>(\d+:\d+)<', html,
+            'duration', fatal=False)
+        if duration:
+            try:
+                mins, secs = duration.split(':')
+                secs = int_or_none(secs)
+                mins = int_or_none(mins)
+                duration = None if secs is None or mins is None else secs + 60 * mins
+            except ValueError:
+                duration = None
+
+        description = self._html_search_regex(
+            r'<div class="tgme_widget_message_text.*?>(.+?)</div>', html,
+            'description', fatal=False)
+
+        return {
+            'id': video_id,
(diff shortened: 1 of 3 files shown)


youtube_dl/extractor/telegram.py
```diff
@@ -0,0 +1,60 @@
...
+            r'<time class="message_video_duration.*?>(\d+:\d+)<', html,
+            'duration', fatal=False)
+        if duration:
+            try:
+                mins, secs = duration.split(':')
+                secs = int_or_none(secs)
+                mins = int_or_none(mins)
+                duration = None if secs is None or mins is None else secs + 60 * mins
+            except ValueError:
+                duration = None
```
dirkf, line 48: Use `parse_duration()` from utils.py.

youtube_dl/extractor/telegram.py
```diff
@@ -0,0 +1,60 @@
...
+
+        description = self._html_search_regex(
+            r'<div class="tgme_widget_message_text.*?>(.+?)</div>', html,
```
dirkf, line 51: Possibly allow to match newlines?
```suggestion
            r'(?s)<div class="tgme_widget_message_text.*?>(.+?)</div>', html,
```

youtube_dl/extractor/telegram.py
```diff
@@ -0,0 +1,60 @@
...
+            'description', fatal=False)
+
+        return {
+            'id': video_id,
+            'title': title,
```
dirkf, line 56: What you extracted as `title` isn't really one, is it? More `channel` or `uploader`. If there really isn't a title in the page or associated metadata, make one up using `_generic_title()` or like `'Telegram video %s from %s' % (video_id, title)`

youtube_dl/extractor/telegram.py
```diff
@@ -0,0 +1,60 @@
...
+        title = m.group('channel')
+        embed_url = url + '?embed=1&mode=tme'
+        html = self._download_webpage(embed_url, video_id)
+
+        video_url = self._search_regex(r'<video src="([^"]+)"', html, 'video_url')
+        formats = [{'url': video_url}]
+
+        duration = self._search_regex(
+            r'<time class="message_video_duration.*?>(\d+:\d+)<', html,
+            'duration', fatal=False)
```
dirkf, line 40: Possibly use `get_element_by_class('message_video_duration', html)`.

youtube_dl/extractor/telegram.py
```diff
@@ -0,0 +1,60 @@
...
+        },
+    }
+
+    def _real_extract(self, url):
+        video_id = self._match_id(url)
+        m = re.match(r'https://t\.me/(?P<channel>[^/]+)/', url)
+        if m is None:
+            raise ExtractorError('Unable to find channel name')
+        title = m.group('channel')
+        embed_url = url + '?embed=1&mode=tme'
```
dirkf, line 32: A more general way of doing this, in case there are trailing parts of the URL already:
```suggestion
        embed_url = update_url_query(url, {'embed': 1, 'mode': 'tme'})
```

youtube_dl/extractor/telegram.py
```diff
@@ -0,0 +1,60 @@
...
+
+from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    int_or_none,
+)
+
+
+class TelegramIE(InfoExtractor):
+    _VALID_URL = r'https://t\.me/(?P<user>[^/]+)/(?P<id>\d+)'
```
dirkf, line 14: I suppose any http URL gets redirected:
```suggestion
    _VALID_URL = r'https?://t\.me/(?P<user>[^/]+)/(?P<id>\d+)'
```

youtube_dl/extractor/telegram.py
```diff
@@ -0,0 +1,60 @@
...
+        video_id = self._match_id(url)
+        m = re.match(r'https://t\.me/(?P<channel>[^/]+)/', url)
+        if m is None:
+            raise ExtractorError('Unable to find channel name')
+        title = m.group('channel')
+        embed_url = url + '?embed=1&mode=tme'
+        html = self._download_webpage(embed_url, video_id)
+
+        video_url = self._search_regex(r'<video src="([^"]+)"', html, 'video_url')
+        formats = [{'url': video_url}]
```
dirkf, line 36: Consider using `_parse_html5_media_entries()` method. Otherwise you can just put the `url` in the returned `info-dict`: probably best to condition the value using `url_or_none()` and raise `ExtractorError()` if the result is `None`.

dirkf (changes requested):
Thanks for your work!

I've made some suggestions and comments. Should be good to go soon.

youtube_dl/extractor/telegram.py
fluks: Yes, it isn't really a title. There is no meaningful title in the page anywhere. I will use one of those then.

Issue #31373, opened 2022-11-21
Unable to download video from www.senate.gov


tejaswi-pote:
## Checklist

- [X] I'm reporting a broken site support issue
- [X] I've verified that I'm running youtube-dl version **2021.12.17**
- [X] I've checked that all provided URLs are alive and playable in a browser
- [X] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [X] I've searched the bugtracker for similar bug reports including closed ones
- [X] I've read bugs section in FAQ

## Verbose log

```
PASTE VERBOSE LOG HERE
```
ydl=YoutubeDL(ydl_opts)
ydl.download([video_url])
ydl_dict = ydl.extract_info(self.source_video_url, download=False)

## Description

WRITE DESCRIPTION HERE

Unable to download Video from following URLS:

**Getting Error:** *** DownloadError: ERROR: Unable to download f4m manifest: HTTP Error 404: Not Found (caused by HTTPError()); please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.

https://www.senate.gov/isvp/?comm=govtaff&type=archv&stt=975&filename=govtaff111722&auto_play=false&poster=https%3A%2F%2Fwww%2Ehsgac%2Esenate%2Egov%2Fimages%2Fvideo%2Dposter%2Dflash%2Dfit%2Epng

 https://www.senate.gov/isvp/?type=arch&comm=energy&filename=energy111722&stt=00:22:30&auto_play=false&wmode=transparent&poster=https%3A%2F%2Fwww%2Eenergy%2Esenate%2Egov%2Fthemes%2Fenergy%2Fimages%2Fvideo%2Dposter%2Dflash%2Dfit%2Epng
 
 'https://www.senate.gov/isvp/?comm=foreign&type=archv&stt=0&filename=foreign080322&auto_play=false&wmode=transparent&poster=https%3A%2F%2Fwww%2Eforeign%2Esenate%2Egov%2Fthemes%2Fforeign%2Fimages%2Fvideo%2Dposter%2Dflash%2Dfit%2Epng'

calebvisage:
[...]

PR #31411, opened 2022-12-10
[senateisvp] Updated senateisvpie to allow downloading from new senate.gov video source

zhuasy:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

The senate.gov extractor has been broken as per [#31373](https://github.com/ytdl-org/youtube-dl/issues/31373). This is because the locations of the m3u8 files had been moved to different URLs, and the format of these URLs have also changed. The extraction of these formats can be found [here](https://thedailywtf.com/articles/enterprise-streaming).

Depending on the video, there are four possible locations for the m3u8 to be located at:
- https://www-senate-gov-media-srs.akamaized.net/hls/live/{stream_id}/{committee}/{filename}/master.m3u8
- https://www-senate-gov-msl3archive.akamaized.net/{msl3}/{filename}_1/master.m3u8
- {stream_domain}/i/{filename}_1@{stream_number}/master.m3u8
- https://ussenate-f.akamaihd.net/i/{video_id}

The above parameters are found in the following places:
- committee, filename - query parameters in the URL
- stream_number, stream_domain, stream_id, msl3 - a js table variable called streamInfo which has been hardcoded as a python dict in senateisvp.py

The new extractor will grab the parameters from the url and the lookup table, then iterate through the possible locations for the m3u8, returning it and its information when found.

Special thanks to [dirkf](https://github.com/dirkf) for the insights as well as code provided in the issue thread.


youtube_dl/extractor/senateisvp.py
```diff
@@ -5,149 +5,169 @@
...
 
-        video_type = qs['type'][0]
-        committee = video_type if video_type == 'arch' else qs['comm'][0]
-        stream_num, domain = self._get_info_for_comm(committee)
+        # the possible locations that the video could be contained at.
+        possible_manifest_urls = [
+            'https://www-senate-gov-media-srs.akamaized.net/hls/live/%d/%s/%s/master.m3u8' % (stream_id, committee, filename),
+            'https://www-senate-gov-msl3archive.akamaized.net/%s/%s_1/master.m3u8' % (msl3, filename),
+            '{stream_domain}/i/%s_1@%d/master.m3u8' % (filename, stream_number),
+            'https://ussenate-f.akamaihd.net/i/%s' % video_id,
```
dirkf, line 136: The page JS is using this (actually probably not since the page code is wrong), supposing that the part before `/i/` is the stream format for the `arch` row of the map:
```suggestion
            'https://ussenate-f.akamaihd.net/i/%s.mp4/master.m3u8' % video_id,
```
Or extract it from the page map:
```py
        _, stream_arch, _, _ = self._get_info_for_comm('arch')
...
            '%s/i/%s.mp4/master.m3u8' % (stream_arch, video_id),
```

youtube_dl/extractor/senateisvp.py
```diff
@@ -5,149 +5,169 @@
...
 
-        if smuggled_data.get('force_title'):
-            title = smuggled_data['force_title']
-        else:
-            title = self._html_search_regex(r'<title>([^<]+)</title>', webpage, video_id)
-        poster = qs.get('poster')
-        thumbnail = poster[0] if poster else None
+        # there is no point in pulling the title from the webpage since it always defaults to 'Integrated Senate Player'
+        title = smuggled_data.get('force_title') or filename
+        stream_number, stream_domain, stream_id, msl3 = self._get_info_for_comm(committee)
```
dirkf, line 129: If getting the committee to stream info map from the page:
```suggestion
        # get the lookup table from the page
        stream_info = '[%s]' % (self._search_regex(
            r'\bnew\s+Array\s*\(([^)]+)',
            webpage, 'stream info', fatal=False) or '', )
        if stream_info:
            stream_info = self._parse_json(stream_info, video_id, transform_source=js_to_json, fatal=False)
        if stream_info:
            self._COMM_MAP = stream_info
```

youtube_dl/extractor/senateisvp.py
```diff
@@ -5,149 +5,169 @@
...
             webpage)
         if mobj:
             return mobj.group('url')
 
+    # returns stream_number, stream_domain, stream_id, and msl3
     def _get_info_for_comm(self, committee):
-        for entry in self._COMM_MAP:
-            if entry[0] == committee:
-                return entry[1:]
+        return self._COMM_MAP[committee][0:]
```
dirkf, line 110: If using the original format:
```suggestion
        for entry in self._COMM_MAP:
            if entry[0] == committee:
                return (entry + ['', ''])[1:5]

dirkf (changes requested):
Thanks for your work! A few comments to address.

PR #31476, opened 2023-01-18
Rumble + UsaWatchdog - improves Rumble support and adds UsaWatchdog s…

epilitimus:
…upport

---

- [x ] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [x ] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [x ] Covered the code with tests (note that PRs without tests will be REJECTED)

- [ ] Bug fix
- [x ] Improvement
- [ x] New extractor
- [ ] New feature

---

This pull request improves the support for Rumble.com and adds support for UsaWatchdog.com which depends on the improved rumble support.

rumble.py:
Rumble offers two methods for embedding videos: JScript and iframes. The legacy iframes method was previously supported by RumbleEmbedIE. Rumble.com itself uses the JScript method.

1. Adds support for Rumble's JScript embedding method. (RumblePageIE)
2. Adds support for Rumble video pages. (via RumblePageIE)
3. Adds support for Rumble channel and user playlists. (RumblePlaylistIE)
4. Provides a function to extract both JScript and iframes video ids from non Rumble sites which use Rumble for video hosting. (e.g. UsaWatchdog)
5. Fixes the crash that occurs when an invalid Rumble video URL is provided by returning None from the actual data extraction function.
6. The actual data extraction function was moved to a base class which both RumbleEmbedIE and RumblePageIE now inherit.

usawatchdog.py
UsaWatchdog now uses Rumble for video hosting. The included extractor uses the function in rumble.py to extract the video ids from story pages (UsaWatchdogStoryIE). Also included is the ability to extract a playlist from the front page of UsaWatchdog.com

Note: I am the original author of all the new code. Some of the code in rumble.py existing previously as part of youtube-dl. I simply rearranged it somewhat.

### DIFF
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -1049,7 +1049,11 @@ from .rtve import RTVEALaCartaIE, RTVELiveIE, RTVEInfantilIE, RTVELiveIE, RTVETe
 from .rtvnh import RTVNHIE
 from .rtvs import RTVSIE
 from .ruhd import RUHDIE
-from .rumble import RumbleEmbedIE
+from .rumble import (
+    RumbleEmbedIE,
+    RumblePageIE,
+    RumblePlaylistIE,
+)
 from .rutube import (
     RutubeIE,
     RutubeChannelIE,
@@ -1414,6 +1418,10 @@ from .urort import UrortIE
 from .urplay import URPlayIE
 from .usanetwork import USANetworkIE
 from .usatoday import USATodayIE
+from .usawatchdog import (
+    UsaWatchdogStoryIE,
+    UsaWatchdogIE,
+)
 from .ustream import UstreamIE, UstreamChannelIE
 from .ustudio import (
     UstudioIE,
--- a/youtube_dl/extractor/rumble.py
+++ b/youtube_dl/extractor/rumble.py
@@ -1,5 +1,6 @@
 # coding: utf-8
 from __future__ import unicode_literals
+import re
 
 from .common import InfoExtractor
 from ..compat import compat_str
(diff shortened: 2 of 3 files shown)


dirkf:
Thanks for your work.

My initial thoughts:
1. `rumble_embedded_id()` should be a class method of one of the Rumble extractors and should be called in the enormous list of embedded tests in `extractor/generic.py`; this has been re-factored in yt-dlp to remove the hard-coded list in favour of run-time discoverable embed extractors, and we'll do the same soon.
2. Then the separate USAWatchdog extractors shouldn't be needed.
3. Please check that any tests that succeed in the yt-dlp Rumble extractor also succeed for you; pull in any code necessary to make them work.
4. As the value of the `rumble_video_info()` method is returned by `_real_extract()` methods, it should either return a valid info dict or raise `ExtractorError`.

epilitimus:
1. It could be a static method. The reason I made it a stand alone was because it uses REs from both RumbleEmbed and RumblePage. Any preference for which extractor to make it a static method of?
2.The UsaWatchdogIE extractor would still be needed as it pulls a UsaWatchdog based playlist not the rumble videos themselves.
3. Will work on this.
4. I was not able to find anything which said what to do if no video was found at the provided URL. Reading the code Youttube_DL.py:838 suggests that None is the proper response which is why I chose it, (though that is in conflict with extractor/common.py:537). Perhaps the proper return should be stated somewhere.

I will work on the above and get back to you.

epilitimus:
Just a thought: would it be better/acceptable to collapse the three Rumble extractors into one by overloading the `suitable()` method?
That way there is no confusion created by having `rumble_emedded_id()` as a method of one but using bits of the other. Would also eliminate RumbleBase. If RumbleEmbed and RumblePage are collapsed into one may as well collapse RumbePlaylist as well. Just an idea.

dirkf:
Actually, what I suggest is starting by from the [yt-dlp extractor](https://github.com/yt-dlp/yt-dlp/tree/master/yt_dlp/extractor/rumble.py), which seems to be a superset of the existing yt-dl extractor, and adding any of your improvements to that. `UsaWatchdogIE` could go in the rumble module.

epilitimus:
All the runmble extractorsa are now one class. generic.py now uses the static method to extract urls from webpages which allowed `UsaWatchdogStoryIE` to be eliminated. 
`UsaWatchdogIE` is still in a separate file because it is not rumble specific. He started on youtube and moved to rumble. There is no reason he couldn't move to another hoster later and he does on occasion embed videos from other hosters.

PR #31525, opened 2023-02-12
Add live stream support for TubiTv

wfaulk:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

The TubiTv extractor was modified to support TubiTv's live streams, specifically the "Live TV" feature at <https://tubitv.com/live>. The accepted URLs are the ones after you click on the video preview from that guide page.

In addition, the existing tests for the extractor didn't work because the videos referenced were no longer available, so I updated those to include a normal on-demand video, plus a live stream.

### DIFF
--- a/youtube_dl/extractor/tubitv.py
+++ b/youtube_dl/extractor/tubitv.py
@@ -13,38 +13,29 @@ from ..utils import (
 
 
 class TubiTvIE(InfoExtractor):
-    _VALID_URL = r'https?://(?:www\.)?tubitv\.com/(?:video|movies|tv-shows)/(?P<id>[0-9]+)'
+    _VALID_URL = r'https?://(?:www\.)?tubitv\.com/(?:video|movies|tv-shows|live)/(?P<id>[0-9]+)'
     _LOGIN_URL = 'http://tubitv.com/login'
     _NETRC_MACHINE = 'tubitv'
     _GEO_COUNTRIES = ['US']
     _TESTS = [{
-        'url': 'http://tubitv.com/video/283829/the_comedian_at_the_friday',
-        'md5': '43ac06be9326f41912dc64ccf7a80320',
+        'url': 'https://tubitv.com/movies/229877/carnival-of-souls',
+        'md5': '275a9f78428cfd6428520989408bcfe9',
         'info_dict': {
-            'id': '283829',
+            'id': '229877',
             'ext': 'mp4',
-            'title': 'The Comedian at The Friday',
-            'description': 'A stand up comedian is forced to look at the decisions in his life while on a one week trip to the west coast.',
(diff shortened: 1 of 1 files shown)


youtube_dl/extractor/tubitv.py
```diff
@@ -13,38 +13,29 @@
...
             'ext': 'mp4',
-            'title': 'Penitentiary',
-            'description': 'md5:8d2fc793a93cc1575ff426fdcb8dd3f9',
-            'uploader_id': 'd8fed30d4f24fcb22ec294421b9defc2',
-            'release_year': 1979,
-        },
-        'params': {
-            'skip_download': True,
+            'title': 'WB TV Watchlist',
+            'description': 're:.*',
```
wfaulk, line 36: The `description` field was required for the test, but the current value of that field ("Catch up on fan favorite series that are a must for your watch list – like Westworld, Raised by Wolves, Third Watch and more.") implies that it might be frequently updated.

dirkf: Just use `compat_str` as the test value:
```py
            'description': compat_str,
```

wfaulk: Ah, I misinterpreted the documentation surrounding that. Updated.

youtube_dl/extractor/tubitv.py
```diff
@@ -97,6 +90,10 @@ def _real_extract(self, url):
                 'url': self._proto_relative_url(sub_url),
             })
 
+        is_live = None
+        if video_data.get('tags'):
+            is_live = 'live' in video_data.get('tags')
+
```
dirkf, line 96:
```suggestion
        is_live = ('live' in video_data.get('tags') or []) or None

```
or, a more complex version that won't crash if `video_data['tags']` isn't iterable or `None`,
```suggestion
        is_live = traverse_obj(video_data, 'tags', expected_type=lambda x: 'live' in x) or None

```

youtube_dl/extractor/tubitv.py
```diff
@@ -76,7 +68,8 @@ def _real_extract(self, url):
         title = video_data['title']
 
         formats = self._extract_m3u8_formats(
-            self._proto_relative_url(video_data['url']),
+            self._proto_relative_url(video_data.get('url')
+                                     or video_data['video_resources'][0]['manifest']['url']),
```
dirkf, line 72:
```suggestion
            self._proto_relative_url(traverse_obj(video_data, 'url', ('video_resources', 0, 'manifest', 'url'))),
```

dirkf (changes requested):
Thanks, I think this is nearly ready to go.

PR #31781, opened 2023-03-06
[indavideo] Fix broken API URL, update examples

thlor:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Status quo: IndaVideo extractor is broken, the API URL seems to have changed since the last commit. I don't know how long it has been broken, I first noticed it ~2 weeks ago, the first time I ever tried the extractor.

My pull request contains the correct URL which I acquired from inspecting the in-browser behavior of the IndaVideo platform. The fix works for every video I tried. It's unclear why the platform added this seemingly arbitrary suffix.

I also updated the list of IndaVideo example videos as some of them have been deleted.

### DIFF
--- a/youtube_dl/extractor/indavideo.py
+++ b/youtube_dl/extractor/indavideo.py
@@ -42,10 +42,9 @@ class IndavideoEmbedIE(InfoExtractor):
 
     # Some example URLs covered by generic extractor:
     #   http://indavideo.hu/video/Vicces_cica_1
-    #   http://index.indavideo.hu/video/2015_0728_beregszasz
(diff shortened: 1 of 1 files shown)


dirkf:
Thanks. This seems to work for the test here according to the CI tests.

However, download test Generic_137 (Indavideo embed) is failing on the URL https://streetkitchen.hu/receptek/igy_kell_otthon_hamburgert_sutni/ which is giving "Unsupported URL". Would you be able to investigate this? It seems that this URL no longer has the expected embedded video. If it can't be fixed, please replace the test video with one of those mentioned in the newly added comments.

PR #31894, opened 2023-03-21
disable --mtime by default

### LINKED ISSUE #1709: Problems with automatically setting the video's mtime to the server's modification time
Youtube-DL recently changed its behaviour in a way that broke how I was using it. It used to set the `mtime` to when the file was downloaded; now it's set to when the file was uploaded to the video site.

This means that I can't sort my videos by download time anymore. It's problematic when you have a large download directory, or multiple collections, which you slowly process over time.

The only other downloader I've found with this default behaviour is `wget`. All the other downloaders -- browsers, torrent programs, curl, scp -- set the `mtime` to the time that you downloaded the file. (This is the [the definition of mtime](https://en.wikipedia.org/wiki/Mtime).)

I can see how people would want to save the video creation date. Luckily, there's a pull request in the pipeline (#1708) that writes this information -- _and more_ -- to the file's xattrs.

So, you can have it all: files sorted by download date, and their creation date!

How 'bout it?

flisk:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

the mtime of a file normally reflects the last modification time of the file on disk rather than what the contents of the file represent. youtube-dl goes against this convention by default, which can cause issues when users don't know about youtube-dl special mtime behavior, e.g. premature deletion of files by cleanup scripts or time-based file searches.

instead of behaving in this unconventional way by default, youtube-dl should not modify the mtime of its output files unless the user specifically asks for that.

fixes #1709

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,8 +263,10 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
                                          files (restart from beginning)
     --no-part                            Do not use .part files - write directly
                                          into output file
-    --no-mtime                           Do not use the Last-modified header to
+    --mtime                              Use the Last-modified header to
                                          set the file modification time
+    --no-mtime                           Do not use the Last-modified header to
+                                         set the file modification time (default)
     --write-description                  Write video description to a
                                          .description file
     --write-info-json                    Write video metadata to a .info.json
@@ -482,15 +484,12 @@ Alternatively, refer to the [developer instructions](#developer-instructions) fo
 
 You can configure youtube-dl by placing any supported command line option to a configuration file. On Linux and macOS, the system wide configuration file is located at `/etc/youtube-dl.conf` and the user wide configuration file at `~/.config/youtube-dl/config`. On Windows, the user wide configuration file locations are `%APPDATA%\youtube-dl\config.txt` or `C:\Users\<user name>\youtube-dl.conf`. Note that by default configuration file may not exist so you may need to create it yourself.
 
-For example, with the following configuration file youtube-dl will always extract the audio, not copy the mtime, use a proxy and save all videos under `Movies` directory in your home directory:
+For example, with the following configuration file youtube-dl will always extract the audio, use a proxy and save all videos under `Movies` directory in your home directory:
 ```
 # Lines starting with # are comments
 
 # Always extract audio
 -x
-
-# Do not copy the mtime
---no-mtime
 
 # Use this proxy
 --proxy 127.0.0.1:3128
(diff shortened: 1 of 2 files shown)


dirkf:
The problem is that the setting has been as it is for so long that if the default is changed people's workflows may be affected. The solution was already provided [10 years ago](https://github.com/ytdl-org/youtube-dl/issues/1709#issuecomment-27641159), ie `--no-mtime`.

yt-dlp also retains this default, presumably for the same reason.

flisk:
a bit of pedantry that's necessary in my view: `--no-mtime` is not a solution to the problem "youtube-dl's default behavior is surprising and unconventional". that flag shouldn't be necessary for youtube-dl to behave in a reasonable way.

yt-dlp retains this default for no other reason than youtube-dl retaining it. see also: https://github.com/yt-dlp/yt-dlp/pull/6585#issuecomment-1477813545

is this your final decision?

dirkf:
I don't think it's obvious that 
> ...default behavior is surprising

Personally I'm happy to see a downloaded file take on a modification time that matches the content. If you want to know when a file was put on the disk, use its creation time.

Also: https://github.com/yt-dlp/yt-dlp/pull/6585#issuecomment-1483808528

flisk:
there is no "creation time" on unix-like systems. you may be confusing "ctime" with "creation time". the latter is only used as such on windows. https://en.wikipedia.org/wiki/MAC_times#Change_time_and_creation_time_(ctime)

dirkf:
True enough, but dependent on filesystem: this just shows that most of my media collection is on NTFS external disks.

The WP article reflects exactly the issue being discussed in #1709.

flisk:
>I don't think it's obvious that

linux documents `st_mtime` in `stat(2)` as such:
>This is the time of last modification of file data.

`stat(2)` on macos documents `st_mtime` as such:
>Time when file data last modified.  Changed by the mknod(2), utimes(2) and write(2) system calls.

since this is the expected default behavior for `st_mtime`, i don't think it's reasonable to deviate from it unless the user has explicitly asked for that.

>If you want to know when a file was put on the disk, use its creation time.

`mtime` is not the same as creation time, it's for recording when the data of a file was last modified. the fact that `mtime` matches the time of file creation if the file hasn't been modified after its creation is incidental.

dirkf:
The controversial point is whether the "time of last modification of file data" (and the equivalent text for macOS) means when the current bits were written to that file, or when the file data that was written was modified.

The latter interpretation avoids the need for filesystem tools to parse file content or extended attributes to determine the "content modification time".

flisk:
there clearly isn't a single correct interpretation of what this attribute should be used for, and changing youtube-dl's default behavior now to align with a different interpretation now seems like a lateral move to me.

thanks for taking the time to discuss this.

PR #31976, opened 2023-03-31
Fix npo support

bartbroere:

- [x] Bug fix

---

### Fix support for NPO sites
~This fixes two things that have been changed on the NPO websites:~
+ ~The current video player no longer returns the token in the JSON body, but instead provides us with an XSRF token in the cookie.~
+ ~A second call changed from GET to POST and should include this XSRF token.~

This branch started out as a small fix, but in the ~11 months the PR was open the NPO site was updated heavily, so now it changes a lot more than the things above. Many of the broadcaster (NL: omroep) sites' extractors (vpro etc.) no longer worked, so these have been removed entirely. I'm always willing to look into re-implementing support for some of these, if that's still possible. However, typically the broadcaster's website just embeds the npo.nl player, so grabbing the same media from the npo.nl URL is recommended.


youtube_dl/extractor/npo.py
```diff
@@ -215,7 +225,10 @@ def _get_info(self, url, video_id):
                     'quality': 'npo',
                     'tokenId': player_token,
                     'streamType': 'broadcast',
-                })
+                },
+                data=b"")
```
dirkf, line 229: Project convention:
```suggestion
                data=b'')
```

youtube_dl/extractor/npo.py
```diff
@@ -6,6 +6,8 @@
 from ..compat import (
     compat_HTTPError,
     compat_str,
+    compat_urllib_parse_unquote_plus,
+    compat_cookies_SimpleCookie,
```
dirkf, line 10: Keep this sorted:
```suggestion
    compat_cookies_SimpleCookie,
    compat_HTTPError,
    compat_str,
    compat_urllib_parse_unquote_plus,
```

dirkf (changes requested):
Thanks for your work!

I've made a few suggestions, really just conventions, so I'll let the CI test run now.

bartbroere:
I have accepted all suggestions here. Sorry I missed some conventions.

bartbroere:
@dirkf Is it ready to merge now?

RickTimmer:
I have had great success using this fork to download NPO Start videos. Greatly appreciated @bartbroere. It would be nice to see these changes merged.

youtube_dl/extractor/npo.py
```diff
@@ -4,8 +4,10 @@
```
dirkf, line 4: Add this line at the top of the file:
```py
# coding: utf-8
```

youtube_dl/extractor/npo.py
```diff
@@ -184,21 +186,29 @@ def _real_extract(self, url):
...
 
     def _get_info(self, url, video_id):
-        token = self._download_json(
+        _, xsrf_token_response = self._download_webpage_handle(
             'https://www.npostart.nl/api/token', video_id,
             'Downloading token', headers={
                 'Referer': url,
                 'X-Requested-With': 'XMLHttpRequest',
-            })['token']
+            })
```
dirkf, line 194: Add a check for expired (?) content:
```suggestion
            })
        if 'npo.nl/start' in xsrf_token_response.geturl():
            raise ExtractorError(
                'Redirect to home page: content expired?',
                video_id=video_id, expected=True)
```

youtube_dl/extractor/npo.py
```diff
@@ -184,21 +186,29 @@ def _real_extract(self, url):
         return self._get_info(url, video_id) or self._get_old_info(video_id)
```
dirkf, line 186: The site used in `_get_old_info()` no longer exists. Make this change and also delete the old method.
 ```suggestion
         return self._get_info(url, video_id)
```

File: youtube_dl/extractor/npo.py (hunk 4 above)
Comment (dirkf) [inline youtube_dl/extractor/npo.py:194]: Avoid reading the page content:
```suggestion
        xsrf_token_response = self._request_webpage(
            'https://www.npostart.nl/api/token', video_id,
            'Downloading token', headers={
                'Referer': url,
                'X-Requested-With': 'XMLHttpRequest',
            })
```

youtube_dl/extractor/npo.py
```diff
@@ -184,21 +186,29 @@ def _real_extract(self, url):
...
+        cookies.load(xsrf_token_response.headers['Set-Cookie'])
+        cookies = dict((k, v.value) for k, v in cookies.items())
+        xsrf_token = cookies['XSRF-TOKEN']
 
         player = self._download_json(
             'https://www.npostart.nl/player/%s' % video_id, video_id,
-            'Downloading player JSON', data=urlencode_postdata({
+            'Downloading player JSON',
+            headers={
+                'x-xsrf-token': compat_urllib_parse_unquote_plus(xsrf_token)
```
dirkf, line 204: Use the built-in cookie jar:
```suggestion
        xsrf_token = next((c.value for c in self.cookiejar if c.name == 'XSRF-TOKEN'), '')
        if xsrf_token:
            xsrf_token = compat_urllib_parse_unquote_plus(xsrf_token)
        else:
            self.report_warning('Failed to get XSRF token', video_id=video_id)

        player = self._download_json(
            'https://www.npostart.nl/player/%s' % video_id, video_id,
            'Downloading player JSON',
            headers={
                'x-xsrf-token': xsrf_token,
```
I'm not sure about the unquoting here but don't have a working URL to test it.

dirkf (changes requested):
I'd like to merge this but it needs valid tests.

All the current test URLs redirect to the home page, but maybe that doesn't happen in NL. Please ensure that there is at least one valid test (with playable content, non-DRM) and mark any invalid URLs with
```py
        'skip': 'Content expired',
```
(or `...: 'only available in NL',`, or whatever is appropriate) after the `info_dict`. If an invalid test URL doesn't match the extractor's `_VALID_URL`, it should be deleted (but I don't think that applies here).

If URLs are only valid in NL, or with an account, please post a console log of successful tests.

I've made some additional suggestions that I might improve with the aid of a usable test URL.

bartbroere:
> I'd like to merge this but it needs valid tests.
> 
> All the current test URLs redirect to the home page, but maybe that doesn't happen in NL. Please ensure that there is at least one valid test (with playable content, non-DRM) and mark any invalid URLs with
> 
> ```python
>         'skip': 'Content expired',
> ```
> 
> (or `...: 'only available in NL',`, or whatever is appropriate) after the `info_dict`. If an invalid test URL doesn't match the extractor's `_VALID_URL`, it should be deleted (but I don't think that applies here).
> 
> If URLs are only valid in NL, or with an account, please post a console log of successful tests.
> 
> I've made some additional suggestions that I might improve with the aid of a usable test URL.

Thanks for the suggested improvements. I'll address all the feedback, and re-request review once I feel it's all done.

bartbroere:
@dirkf Looking into it I realised this is probably caused by [the recently released new NPO app and website](https://service.npo.nl/onderwerp/technische-zaken#extra-toelichting-bij-vernieuwing-npo-start). I'll check all tests and maybe add some new ones

bartbroere:
https://github.com/yt-dlp/yt-dlp/pull/9319 addresses the same issues. I think (speaking for the Dutch users a bit) that support for npo.nl is the main feature, and all the other sites often just embed the NPO player in some way.

Since almost all of the other sites no longer work in the same way as before, I would propose throwing away extractors for many other domains, since rebuilding them is probably a lot quicker.

dirkf:
It's a bit vexing that the yt-dlp PR has done the same job, but by all means plunder any useful stuff from there. I would generally merge/back-port an updated _yt-dlp_ extractor but generally also only when ours doesn't work.

bartbroere:
> It's a bit vexing that the yt-dlp PR has done the same job, but by all means plunder any useful stuff from there. I would generally merge/back-port an updated _yt-dlp_ extractor but generally also only when ours doesn't work.

I'm not convinced that the pull request on `yt_dlp` actually works for the new npo.nl player yet:
https://github.com/rvsit/yt-dlp/blob/c2d2b589588c7dbe44a97b4d29bad0d68922cd07/yt_dlp/extractor/npo.py#L122-L123

On the other hand, my branch probably doesn't work for some of the `yt_dlp` test cases yet.

So I think we can do a nice bit of "cross-plundering" between my `youtube-dl` pull request and @rsvit 's `yt-dlp` pull request.

dirkf:
Great.

Also, if there are standard patterns for the discarded sites where NPO videos are embedded, we can add a class variable `_EMBED_REGEX` that is a list of those patterns with the NPO URL found in regex group `url`. Add this to provide an entry-point that could be called from the generic extractor:
```py
    @classmethod
    def _extract_urls(cls, webpage):
        def yield_urls():
            for p in cls._EMBED_REGEX:
                for m in re.finditer(p, webpage):
                    yield m.group('url')

        return list(yield_urls())
```
Eventually the webpage extraction system from yt-dlp should be pulled in, with the method standardised in `InfoExtractor`, and the generic extractor refactored to blazes.

youtube_dl/extractor/npo.py
```diff
@@ -82,686 +49,77 @@ class NPOIE(NPOBaseIE):
...
+            url = url[:-9]
+        elif url.endswith('/afspelen/'):
+            url = url[:-10]
+        url = url.rstrip('/')
+        slug = url.split('/')[-1]
+        page = self._download_webpage(url, slug, 'Finding productId using slug: %s' % slug)
+        # TODO find out what proper HTML parsing utilities are available in youtube-dl
+        next_data = page.split('<script id="__NEXT_DATA__" type="application/json">')[1].split('</script>')[0]
+        # TODO The data in this script tag feels like GraphQL, so there might be an easier way
+        #      to get the product id, maybe using a GraphQL endpoint
```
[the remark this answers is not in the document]
bartbroere: Yes, sounds good to work together. I have added you as a collaborator to my fork, so you can push to the `fix-npo-support` branch directly.

And thanks for the information about the endpoint. That's a lot better than hoping it continues to predictably be in the same `script` tag.

bartbroere:
> My suggestion for a universal approach would be
> 
> * Extract POMS id from url with various implementations per url format/domain
> * They all call the same function that take a POMS id and get's the url's
> * Optionally enrich it with title and such if not already extracted from url
>   
>   * We could use the POMS api for this, but the url signing/api key is a little annoying even although those keys are public, `https://rs.poms.omroep.nl/v1/api/media/[id]` [example_response_POMS_NTR_388772.json](https://github.com/ytdl-org/youtube-dl/files/14463673/example_response_POMS_NTR_388772.json)

This indeed seems the best way to do this. I'll rework the code to fit in this pattern.

youtube_dl/extractor/npo.py
```diff
@@ -82,686 +27,253 @@ class NPOIE(NPOBaseIE):
...
+        results = re.findall("page: '(.+)'", page)
+        formats = []
+        for result in results:
+            formats.extend(self._download_by_product_id(result, video_id))
 
-        live_id = self._search_regex(
-            [r'media-id="([^"]+)"', r'data-prid="([^"]+)"'], webpage, 'live id')
+        if not formats:
+            raise ExtractorError('Could not find a POMS product id in the provided URL, '
+                                 'perhaps because all stream URLs are DRM protected.')
```
dirkf, line 161: `self._sort_formats(...)` should be called and will raise if there aren't any. If there is a way to identify that a stream has DRM, and given that unlike yt-dlp we're going to skip DRM formats, one could, eg, return a tuple '(formats, number_of formats_seen)` and then compare the sum of the total formats against `len(formats)`. If `not formats` and the values differ, `self.report_drm()` can be called.

youtube_dl/extractor/npo.py
```diff
@@ -82,686 +27,253 @@ class NPOIE(NPOBaseIE):
...
+                                        }).encode('utf8'),
+                                    headers={
+                                        'Content-Type': 'application/json',
+                                    })
+        product_id = media.get('data', {}).get('player', {}).get('pomsProductId')
+
+        formats = self._download_by_product_id(product_id, video_id)
 
         return {
```
dirkf, line 134: This could be re-worked like the previous `return`, with the `merge_dicts(traverse_obj(metadata, dict_construction), dict_of_known_vars)` pattern.

youtube_dl/extractor/npo.py
```diff
@@ -82,686 +27,253 @@ class NPOIE(NPOBaseIE):
...
+        for profile in (
+                'dash',
+                # 'hls' is available too, but implementing it doesn't add much
+                # As far as I know 'dash' is always available
+        ):
+            stream_link = self._download_json(
+                'https://prod.npoplayer.nl/stream-link', video_id=slug,
+                data=json.dumps({
+                    'profileName': profile,
+                    'drmType': 'widevine',
```
dirkf, line 89: Naïvely, I wonder any streams returned with this request are not WV-encrypted, and what happens if other or no values are passed.

youtube_dl/extractor/npo.py
```diff
@@ -114,12 +96,184 @@ def _real_extract(self, url):
...
+        'url': 'https://schooltv.nl/item/zapp-music-challenge-2015-zapp-music-challenge-2015',
+        # TODO fill in other test attributes
+    }]
+
+    def _real_extract(self, url):
+        video_id = url.rstrip('/').split('/')[-1]
+
+        # TODO Find out how we could obtain this automatically
+        #      Otherwise this extractor might break each time SchoolTV deploys a new release
+        build_id = 'b7eHUzAVO7wHXCopYxQhV'
```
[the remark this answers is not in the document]
dirkf: There is the `_search_nextjs_data()` method if needed.

youtube_dl/extractor/npo.py
```diff
@@ -82,686 +27,253 @@ class NPOIE(NPOBaseIE):
...
-            'display_id': 'ademhaling-de-hele-dag-haal-je-adem-maar-wat-gebeurt-er-dan-eigenlijk-in-je-lichaam',
-            'title': 'Ademhaling: De hele dag haal je adem. Maar wat gebeurt er dan eigenlijk in je lichaam?',
-            'ext': 'mp4',
-            'description': 'md5:abfa0ff690adb73fd0297fd033aaa631'
-        },
-        'params': {
-            # Skip because of m3u8 download
-            'skip_download': True
+            'id': video_id,
+            'title': metadata.get('title', '') + ' - ' + metadata.get('subtitle', ''),
```
dirkf, line 225: Or use `utils.join_nonempty()` for this effect:
* `'title'`, `'subtitle'` -> `'title - subtitle'`
* `'title'`, `''` -> `'title'`
* `''`, `'subtitle'` -> `'subtitle'`

youtube_dl/extractor/npo.py
```diff
@@ -114,12 +96,184 @@ def _real_extract(self, url):
...
+    _VALID_URL = r'https?://(?:www\.)?vpro.nl/.*'
+    _TESTS = [{
+        'url': 'https://www.vpro.nl/programmas/tegenlicht/kijk/afleveringen/2015-2016/offline-als-luxe.html',
+        # TODO fill in other test attributes
+    }]
+
+    def _real_extract(self, url):
+        video_id = url.rstrip('/').split('/')[-1]
+        page, _ = self._download_webpage_handle(url, video_id)
+        results = re.findall(r'data-media-id="(.+_.+)"\s', page)
```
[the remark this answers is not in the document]
dirkf: `'"([^"]+)"'`, maybe?

In principle the quoting could be `'` or `"` or absent, and there could be white space, so
```py
        results = [m.group('id') for m in re.finditer(
            r'''data-media-id\s*=\s*(?P<q>"|')?(?P<id>(?(q)[^(?P=q)]|[a-zA-Z0-9_])+)''',
            page)]
```
Or s/t like that. The pattern doesn't need to include the RH quote because the match of the quoted text automatically runs out when it's reached.

There is also the question of whether `\w` can or should be used instead of `a-zA-Z0-9_` (locale issues).

youtube_dl/extractor/npo.py
```diff
@@ -82,686 +27,253 @@ class NPOIE(NPOBaseIE):
...
-            for video_id in orderedSet(re.findall(self._PLAYLIST_ENTRY_RE, webpage))
-        ]
 
-        playlist_title = self._html_search_regex(
-            self._PLAYLIST_TITLE_RE, webpage, 'playlist title',
-            default=None) or self._og_search_title(webpage)
+class VPROIE(NPOIE):
+    IE_NAME = 'vpro'
+    IE_DESC = 'vpro.nl'
+    _VALID_URL = r'https?://(?:www\.)?vpro.nl/.*'
```
dirkf, line 241: Is this a tight enough pattern?

youtube_dl/extractor/npo.py
```diff
@@ -82,686 +27,253 @@ class NPOIE(NPOBaseIE):
...
 
-        return self.playlist_result(entries, playlist_id, playlist_title)
+    def _real_extract(self, url):
+        video_id = url.rstrip('/').split('/')[-1]
+        page, _ = self._download_webpage_handle(url, video_id)
+        results = re.findall(r'data-media-id="(.+_.+)"\s', page)
+        formats = []
+        for result in results:
+            formats.extend(self._download_by_product_id(result, video_id))
+            break  # TODO find a better solution, VPRO pages can have multiple videos embedded
```
dirkf, line 254: May this embedding occur in other pages (not vpro.nl)?

Are the second and up videos related (clips, trailers, etc), or is the case more like a series page with various episodes?

In the first case maybe skip the subsidiary videos; in the second normally return a `playlist` result whose entries are either the `url_result()`s of episode URLs constructed for each video, or `info_dict`s extracted from the page.

dirkf (commented):
Thanks, I can see this is coming along well. Here are a few comments and suggestions.

youtube_dl/extractor/npo.py
bartbroere: I closed some of these because XSRF is no longer necessary fortunately. The code this comments on is no longer is in the diff.

youtube_dl/extractor/npo.py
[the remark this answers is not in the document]
bartbroere: We no longer extract info from this script tag anymore, so resolved :-)

youtube_dl/extractor/npo.py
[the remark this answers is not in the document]
bartbroere: > There is the `_search_nextjs_data()` method if needed.

Thanks! I'll look into that and use it if I can make it work.

youtube_dl/extractor/npo.py
```diff
@@ -1,78 +1,23 @@
...
-                                npo\.nl/(?:[^/]+/)*|
-                                (?:ntr|npostart)\.nl/(?:[^/]+/){2,}|
-                                omroepwnl\.nl/video/fragment/[^/]+__|
-                                (?:zapp|npo3)\.nl/(?:[^/]+/){2,}
-                            )
-                        )
-                        (?P<id>[^/?#]+)
-                '''
+    IE_DESC = 'npo.nl'
+    _VALID_URL = r'https?://(?:www\.)?npo\.nl/.*'
```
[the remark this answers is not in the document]
bartbroere:
```suggestion
    _VALID_URL = r'https?://(?:www\.)?npo\.nl/start/serie/'
```

I'm adding this so it becomes a bit more specific. There may be some older players on this domain, for which I could try to add a different extractor (with a more specific regex as well).

rvsit (approved):

youtube_dl/extractor/npo.py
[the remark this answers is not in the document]
dirkf: In fact, all the information that you might have got with the JSON API is in the Next.js hydration JSON in the page, including the build ID that is no longer of interest. I have this working but will update once other issues have been cleared.

youtube_dl/extractor/npo.py
dirkf: Looking at the test video page, there is apparently a content video and a teaser video. The former can be detected because it's inside (preceded by) `<div class=grid>`.

As far as I can see other pages that might list multiple videos are playlist pages like https://www.vpro.nl/programmas/tegenlicht/kijk/afleveringen.html or https://www.vpro.nl/programmas/tegenlicht/kijk/afleveringen.html or https://www.vpro.nl/programmas/tegenlicht/categorieen/wereld.html that don't include `data-media-id`s but just have links to programme episodes. Counterexamples welcome.

bartbroere:
I'll be mostly offline during April, so this is just a heads up that this PR will not see much progress for a few weeks. I didn't forget about it though, and plan to continue working on it in May.

dirkf:
I may be able to save you much of that work. I've run through the d/l tests locally and adjusted the code somewhat to match. If anyone has an answer to https://github.com/ytdl-org/youtube-dl/pull/31976#discussion_r1548520050, I'll be able to settle that extractor and push an update that should be close to mergeable.

youtube_dl/extractor/npo.py
bartbroere: Done!

youtube_dl/extractor/npo.py
```diff
@@ -82,686 +27,253 @@ class NPOIE(NPOBaseIE):
...
+                                            },
+                                            'query': 'query getMedia($id: ID!, $mediaUrl: String, $hasAdConsent: Boolean!, $atInternetId: Int) {\n  player(\n    id: $id\n    mediaUrl: $mediaUrl\n    hasAdConsent: $hasAdConsent\n    atInternetId: $atInternetId\n  ) {\n    ... on PlayerSucces {\n      brand {\n        name\n        slug\n        broadcastsEnabled\n        __typename\n      }\n      title\n      programTitle\n      pomsProductId\n      broadcasters {\n        name\n        __typename\n      }\n      duration\n      classifications {\n        title\n        imageUrl\n        type\n        __typename\n      }\n      image {\n        title\n        url\n        __typename\n      }\n      cta {\n        title\n        url\n        __typename\n      }\n      genres {\n        name\n        __typename\n      }\n      subtitles {\n        url\n        language\n        __typename\n      }\n      sources {\n        name\n        url\n        ratio\n        __typename\n      }\n      type\n      token\n      __typename\n    }\n    ... on PlayerError {\n      error\n      __typename\n    }\n    __typename\n  }\n}'
```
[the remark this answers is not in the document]
bartbroere: I think we could also reduce the fields we request in this GraphQL query and remove the `\n` newlines. It would no longer match the behaviour of the original site however.

Moving it into a variable is a good idea even if I would do that of course.

youtube_dl/extractor/npo.py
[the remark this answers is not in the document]
bartbroere: I have used the suggested `_search_nextjs_data()` method, but I still use the JSON endpoint to get the metadata. 

My preference is to call the JSON endpoint since it requires a bit less parsing and searching, but I'm open to your solution as well @dirkf . Feel free to push it to the pull request's branch if you prefer your solution

youtube_dl/extractor/npo.py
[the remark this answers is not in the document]
bartbroere: I moved it into a multi-line string, and replaced the newlines with actual newlines to make it somewhat readable too.

youtube_dl/extractor/npo.py
bartbroere: I didn't want to make it a lot more specific, because this site is probably using a CMS that lets you embed in practically any page. By excluding pages in advance with the regex pattern we might hinder some people. I do remain open to suggestions to improve this regex of course.

youtube_dl/extractor/npo.py
bartbroere: I have removed the TODO, because I share your opinion. The cases where there's multiple embeds are often the teaser+episode scenario. Let's not make something for the eventuality that multiple videos are embedded, at least not while we don't know of any examples of this being done in practice.

bartbroere:
It successfully downloads all test URLs right now, but I'm not sure what it should say in the `ext` field in the `info_dict` of each test. I tried debugging it, but that gets tricky with the `format_selector` method being created at runtime.

The example URLs all download successfully with this command
```bash
python3.11 -m youtube_dl https://npo.nl/start/serie/vpro-tegenlicht/seizoen-11/zwart-geld-de-toekomst-komt-uit-afrika https://npo.nl/start/serie/zembla/seizoen-2015/wie-is-de-mol-2/ https://www.bnnvara.nl/videos/27455 https://ongehoordnederland.tv/2024/03/01/korte-clips/heeft-preppen-zin-betwijfel-dat-je-daar-echt-iets-aan-zult-hebben-bij-oorlog-lydia-daniel/ https://www.zapp.nl/programmas/zappsport/gemist/POMS_AT_811523 https://schooltv.nl/item/zapp-music-challenge-2015-zapp-music-challenge-2015 https://hetklokhuis.nl/dossier/142/zoek-het-uit/tv-uitzending/2987/aliens https://www.vpro.nl/programmas/tegenlicht/kijk/afleveringen/2015-2016/offline-als-luxe.html https://anderetijden.nl/programma/1/Andere-Tijden/aflevering/676/Duitse-soldaten-over-de-Slag-bij-Arnhem
```

but I can't get the test framework to pass. That gives me the error:
```
youtube_dl.utils.DownloadError: ERROR: requested format not available
```

So even with the failing tests I think this is an improvement over the current `master`, which no longer successfully downloads any NPO video. If anyone knows what it should say in `ext`, I can change it to the correct value.

victorhartman:
Hey, this is amazing work! I wanted to test it out, but it no longer seems to work unfortunately. Issue with getting the token.

```
py.exe -m youtube_dl https://www.bnnvara.nl/videos/27455
[bnnvara] 27455: Downloading JSON metadata
[bnnvara] VARA_101369808: Downloading token
ERROR: An extractor error has occurred. (caused by KeyError('token'));`
```

bartbroere:
> Hey, this is amazing work! I wanted to test it out, but it no longer seems to work unfortunately. Issue with getting the token.
> 
> ```
> py.exe -m youtube_dl https://www.bnnvara.nl/videos/27455
> [bnnvara] 27455: Downloading JSON metadata
> [bnnvara] VARA_101369808: Downloading token
> ERROR: An extractor error has occurred. (caused by KeyError('token'));`
> ```

Thanks for reporting this! I'll look into it.

mdbraber:
@bartbroere NPO has renamed 'token' to 'jwt'

mdbraber:
@bartbroere can this branch be (also) merged in yt-dlp and can development be synced there - or is there any specific reason why this is built on youtube-dl? (this was also referenced before in relation to https://github.com/yt-dlp/yt-dlp/pull/9319)

I might be off, but it seems that youtube-dl isn't considered upstream for yt-dlp (anymore) so any youtube-dl changes won't make it back to yt-dlp by default?

youtube_dl/extractor/npo.py
```diff
@@ -1,767 +1,387 @@
...
-    url_or_none,
-    urlencode_postdata,
-)
-
-
-class NPOBaseIE(InfoExtractor):
-    def _get_token(self, video_id):
-        return self._download_json(
-            'http://ida.omroep.nl/app.php/auth', video_id,
-            note='Downloading token')['token']
```
victorhartman, line 32: Changing this to 'jwt' makes it work again

mdbraber: `--no-check-certificate` was not needed for me. Created a PR for fixing the token error https://github.com/bartbroere/youtube-dl/pull/1

bartbroere:
> @bartbroere NPO has renamed 'token' to 'jwt'

Thanks! I changed it here too!

> @bartbroere can this branch be (also) merged in yt-dlp and can development be synced there - or is there any specific reason why this is built on youtube-dl? (this was also referenced before in relation to [yt-dlp/yt-dlp#9319](https://github.com/yt-dlp/yt-dlp/pull/9319))
> 
> I might be off, but it seems that youtube-dl isn't considered upstream for yt-dlp (anymore) so any youtube-dl changes won't make it back to yt-dlp by default?

I wouldn't be against merging this there as well. I did indeed think it would end up there automatically once merged here, but if that's not the case I could create a PR there too. There is an open PR there as well, let's start by tagging them so we can compare our work later: https://github.com/yt-dlp/yt-dlp/pull/11227

youtube_dl/extractor/npo.py
bartbroere: Thanks, I renamed it to jwt in commit [91e29f](https://github.com/ytdl-org/youtube-dl/pull/31976/commits/91e279fec128500fa5dc4ae17b986be8e5c6faf2)

bartbroere:
> btw, related question. Is all content uploaded nowadays drm protected? This is a great PR, but it seems like it can only be used on older content as the rest is drm protected.

Yes, unfortunately it seems practically all new content is now DRM protected.

I kind of understand adding DRM when a broadcaster buys the rights to an international television show, but for things like the news I think it is a bit much (especially for a publicly funded organisation). In my opinion there's lots of legitimate reasons to want to use a news fragment offline.

And you're also correct on the older content still being downloadable. I think this older content makes it worth the effort to fix the NPO extractor here.

youtube_dl/extractor/npo.py
```diff
@@ -1,767 +1,387 @@
...
-            for video_id in orderedSet(re.findall(self._PLAYLIST_ENTRY_RE, webpage))
-        ]
-
-        playlist_title = self._html_search_regex(
-            self._PLAYLIST_TITLE_RE, webpage, 'playlist title',
-            default=None) or self._og_search_title(webpage)
-
-        return self.playlist_result(entries, playlist_id, playlist_title)
+        video_id = url.rstrip('/').split('/')[-1]
+        page, _ = self._download_webpage_handle(url, video_id)
```
dirkf, line 359:
```suggestion
        page = self._download_webpage(url, video_id)
```

youtube_dl/extractor/npo.py
```diff
@@ -1,767 +1,387 @@
...
-            default=None) or self._og_search_title(webpage)
-
-        return self.playlist_result(entries, playlist_id, playlist_title)
+        video_id = url.rstrip('/').split('/')[-1]
+        page = self._download_webpage(url, video_id)
+        results = re.findall(r'data-media-id="([a-zA-Z0-9_]+)"\s', page)
+        formats = []
+        for result in results:
+            formats.extend(self._extract_formats_by_product_id(result, video_id))
+            break
```
dirkf, line 364: If the desired behaviour is to use just the first match, this is easier: 

```suggestion
        page = self._download_webpage(url, video_id)
        format = traverse_obj(re.search(r'data-media-id="([a-zA-Z0-9_]+)"\s', page), (
                1, T(lambda x: self._extract_formats_by_product_id(x, video_id)),
                Ellipsis))
```
Or if not getting `page` shouldn't crash:
```suggestion
        page = self._download_webpage(url, video_id, fatal=False) or ''
        format = traverse_obj(re.search(r'data-media-id="([a-zA-Z0-9_]+)"\s', page), (
            1, T(lambda x: self._extract_formats_by_product_id(x, video_id)),
            Ellipsis))
```
To be extra sure (if the target must be an attribute value of an HTML element), you can search for elements like `r'<[\w.-]+\s[^>]*(?<!-)\bdata-media-id\s*=\s*[^>]+>' and then get the first or all `extract_attributes(element).get('data-media-id') that is not `None`.

PR #32138, opened 2023-05-03
Df extractor miscbp 202305

dirkf:

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---


youtube_dl/utils.py
```diff
@@ -4262,16 +4263,38 @@ def try_get(src, getter, expected_type=None):
                 return v
 
 
-def merge_dicts(*dicts):
+def merge_dicts(*dicts, **kwargs):
```
pukkandan, line 4266: Why `kwargs` and not `*dicts, unblank=True, rev=False`? Is it some py2 compat thing I'm missing?

youtube_dl/utils.py
```diff
@@ -4262,16 +4263,38 @@ def try_get(src, getter, expected_type=None):
...
+    """
+        Merge the `dict`s in `dicts` using the first valid value for each key.
+        Normally valid: not None and not an empty string
+
+        Keyword-only args:
+        unblank:    allow empty string if False (default True)
+        rev:        merge dicts in reverse order (default False)
```
pukkandan, line 4273: I don't get the point. Instead of `unblank=False`, why not just do `**d1, **d2`? And instead of `rev=True`, you can just pass the dict in reverse order, no?

youtube_dl/utils.py
```diff
@@ -6062,122 +6159,173 @@ def from_iterable(iterables):
...
+        result = _traverse_obj(obj, path, index == len(paths), True)
         if result is not None:
             return result
 
     return None if default is NO_DEFAULT else default
 
 
+def exp(x):
+    """ For use in yt-dl instead of {type} or set((type,)) """
+    return set((x,))
```
pukkandan, line 6326: We chose sets specifically because of the syntactic simplicity. Another proposal was `traverse_obj.transform(f)`, which isn't really ideal either. But imo `exp` is too vague.

youtube_dl/utils.py
dirkf: Just so:
```
>>> def foo(*args, kw=None):
  File "<input>", line 1
    def foo(*args, kw=None):
                    ^
SyntaxError: invalid syntax
>>> 
```
Weirdly, you can put the keyword args **before** the `*args`, but then you have to call it like that too: otherwise 2.7 thinks that what you wanted to be `args[0]` is `kw` and complains about a duplicate `kw` argument. I don't have a 2.6 to test.

youtube_dl/utils.py
dirkf: 1. 2.7 knows not `{**d1, **d2}` (`**` is only expanded in the function call, I think).
2. (Hence) `rev` allows the list of `dict`s to be left in the same order when converting from `{**d1, **d2}` and minimising diffs.

There is some logic to this, given point 1.

youtube_dl/utils.py
dirkf: I don't really care for it, but that was the best name of 3 or fewer letters that I came up with. If it's too vague, `E()` is shorter and at least as vague. The `{type}` thing isn't entirely obvious but does have the idea of "must be a `type`".

Obviously, my aim was to (a) minimise diffs between `{type}` and `exp(type)` (or whatever) and (b) avoid writing `expected_type=type` a lot.

dirkf:
To do:
- [ ] pull `_search_nextjs_data()` and `_search_nuxt_data()` into `InfoExtractor`
- [ ] decide on name for `exp()`.

youtube_dl/utils.py
dirkf: In #32445, I committed `T()`.

youtube_dl/extractor/clipchamp.py
```diff
@@ -0,0 +1,76 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..compat import compat_str
+from ..utils import (
+    exp,
+    ExtractorError,
+    merge_dicts,
```
dirkf, line 9:
```suggestion
    ExtractorError,
    merge_dicts,
    T,
```

youtube_dl/extractor/globalplayer.py
```diff
@@ -0,0 +1,285 @@
...
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import (
+    clean_html,
+    exp,
+    join_nonempty,
+    merge_dicts,
+    parse_duration,
+    str_or_none,
```
dirkf, line 11:
```suggestion
    join_nonempty,
    merge_dicts,
    parse_duration,
    str_or_none,
    T,
```

youtube_dl/utils.py
```diff
@@ -4268,13 +4268,8 @@ def variadic(x, allowed_types=NO_DEFAULT):
...
 def dict_get(d, key_or_keys, default=None, skip_false_values=True):
-    if isinstance(key_or_keys, (list, tuple)):
-        for key in key_or_keys:
-            if key not in d or d[key] is None or skip_false_values and not d[key]:
-                continue
-            return d[key]
-        return default
-    return d.get(key_or_keys, default)
+    exp = (lambda x: x or None) if skip_false_values else IDENTITY
+    return traverse_obj(d, *variadic(key_or_keys), expected_type=exp, default=default)
```
dirkf, line 4272:
```suggestion
    return traverse_obj(d, *variadic(key_or_keys), expected_type=exp, default=default, get_all=False)
```

dirkf: Done.

Issue #7201, opened 2015-10-17
Support hqporner.com


codeballarina23:
There is a nice and clean website called hqporner. Sadly youtube-dl doesn't have a extractor at the moment and generic doesn't work too. 

Would be awesome of you if you put support for hqporner.com into the program.

Thanks

dstftw:
Post [an example URL](https://github.com/rg3/youtube-dl/blob/master/README.md#is-the-description-of-the-issue-itself-sufficient).

codeballarina23:
Thanks for fast answer. 

Here example URL: http://hqporner.com/hdporn/80371-Orgasmic_massage_for_Emily_Bloom.html

dstftw:
This looks like copyrighted content. youtube-dl does not support sites dedicated to [copyright infringement](https://github.com/rg3/youtube-dl#can-you-add-support-for-this-anime-video-site-or-site-which-shows-current-movies-for-free). Do they have rights to distribute this? If not provide [an example URL](https://github.com/rg3/youtube-dl/blob/master/README.md#is-the-description-of-the-issue-itself-sufficient) that does not violate any copyrights.

codeballarina23:
I looked at the ToS but they only talk about user submissions and what user "aren't allowed" to submit. But nothing that strong enough to fit yt-dls requirements i think. So if nobody else finds a sufficient notice...

PR #32245, opened 2023-05-28
[HQPorner] Add new extractor: video, playlist and search pages

dirkf:
<details>
<summary>Boilerplate: own code, new extractor</summary>

---

- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

</details>

---

This PR adds an extractor module for hqporner.com, as suggested in https://github.com/yt-dlp/yt-dlp/issues/7116.

### DIFF
--- a/youtube_dl/extractor/extractors.py
+++ b/youtube_dl/extractor/extractors.py
@@ -480,6 +480,11 @@ from .hotstar import (
 )
 from .howcast import HowcastIE
 from .howstuffworks import HowStuffWorksIE
+from .hqporner import (
+    HQPornerIE,
+    HQPornerListIE,
+    HQPornerSearchIE,
+)
 from .hrfernsehen import HRFernsehenIE
 from .hrti import (
     HRTiIE,
--- a/youtube_dl/extractor/hqporner.py
+++ b/youtube_dl/extractor/hqporner.py
@@ -1,0 +1,233 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+import itertools
+import re
+from string import capwords
+
+from .common import InfoExtractor
+from ..utils import (
+    clean_html,
+    date_from_str,
+    get_element_by_class,
+    get_element_by_id,
+    int_or_none,
+    join_nonempty,
+    parse_duration,
+    strip_or_none,
+    unified_strdate,
+    urljoin,
+)
+
+
+class HQPornerIE(InfoExtractor):
+    _VALID_URL = r'https?://(?:www\.)?hqporner\.com/hdporn/(?P<id>[\d]+)-'
+    _TESTS = [{
+        'url': 'https://hqporner.com/hdporn/110374-looking_for_a_change_of_pace.html',
+        'md5': '7eb7b791a1ce8a619bde603b2dc334b5',
+        'info_dict': {
+            'id': '110374',
+            'ext': 'mp4',
+            'title': 'Looking For A Change Of Pace',
+            'description': 'featuring Myra',
+            'upload_date': '20230227',
+            'age_limit': 18,
+            'tags': list,
+            'categories': list,
+            'duration': 3271,
+            'thumbnail': r're:https?://.*\.jpg',
+        },
+    }, {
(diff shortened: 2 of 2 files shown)


JChris246:
An [issue](https://github.com/ytdl-org/youtube-dl/issues/7201) was opened previously for this site and subsequently closed as it was determined that this site hosts copyrighted content. https://github.com/ytdl-org/youtube-dl/issues/7201

dirkf:
Since content is generally subject to copyright, that's not the deciding factor. Look at other sites supported by yt-dl and you will find copyrighted material that may be hosted with, or (shockingly) without, the copyright holder's permission: not just porn, but user-generated content sites like YouTube.

Also, #7201 applies to the site as it was in 2015, though there's no indication of whether the site has changed.

The project's policy has two parts:
>Make sure that site you are requesting is not dedicated to copyright infringement, see https://yt-dl.org/copyright-infringement. youtube-dl does not support such sites. 

>In order for site support request to be accepted all provided example URLs should not violate any copyrights.

The linked guidance is this:
>As a matter of policy (as well as legality), youtube-dl does not include support for services that specialize in infringing copyright. As a rule of thumb, if you cannot easily find a video that the service is quite obviously allowed to distribute (i.e. that has been uploaded by the creator, the creator's distributor, or is published under a free license), the service is probably unfit for inclusion to youtube-dl.
>
>A note on the service that they don't host the infringing content, but just link to those who do, is evidence that the service should not be included into youtube-dl. The same goes for any DMCA note when the whole front page of the service is filled with videos they are not allowed to distribute. A "fair use" note is equally unconvincing if the service shows copyright-protected videos in full without authorization.

I would say that "the whole front page of the service is filled with videos they are not allowed to distribute" could cover almost every porn site that yt-dl supports. I looked at a large sample of yt-dl[p] porn site extractors and most failed the test: eg, daft.sex.

The porn ecosystem is entirely different from, say, broadcasting and movie streaming sites. If a movie is available on BBC iPlayer, we can take it that the BBC is distributing it with permission. Sites like YouTube and Vimeo do distribute actual user-generated content alongside material for which the submitter may not have the rights. In fact, YT conspires with its users and the music industry to promote ad-supported content submitted by users who do not own the content.

A typical porn site provides no explicit copyright attribution. The user can't be expected to know which video might have been licensed by the copyright owner, even in the presence of a logo or watermark. Does the site have some sort of ad-revenue-sharing deal with content owners? Is the content being supplied legitimately as a promotion? No-one can tell for sure, and the waters are further muddied because the same company may operate both a paid content site and ad-supported "user-generated" sites that include old content from the paid site.

My interpretation of the policy is rather that yt-dl is like a web browser and it cannot be responsible for identifying whether any particular content is legitimate. If a user is concerned about some material, don't download it. The site must allow copyright owners to remove unauthorised content. If the site claims to follow DMCA, that should be enough, as it is for YouTube. Therefore, I don't entirely support the statement about DMCA in the guidance quoted above.

Also, the project requirement "example URLs should not violate any copyrights" seems to be very difficult to interpret in this context. One might think that a URL is a reference that cannot be restricted by copyright: otherwise no-one could write `the movie "Top Gun - Maverick"` without permission. If the policy is saying that no example URL should reference content that is explicitly being distributed without permission ("hey l33t guys, grab this warez movie I hacked"), that's clear enough. But anything more than that fails in the UGC or porn site context.

Regarding HQPorner, there are particular issues:
1. it claims to host user-submitted content, but there's no identification of the uploader (the site says it registers users manually)
2. its terms require that users have permission to post videos, but all videos that I checked appeared to be unedited commercial productions
3. its terms acknowledge DMCA and there is a dedicated email address for DMCA take-down requests, but DMCA is linked in the terms and contact pages, rather than on the main video and list pages.

Because of point 1 above, as well as the ecosystem context, it's difficult to tell whether any video is posted with valid permission. Maybe (point 2) the material is actually being provided by the owners as a sort of ad-supported front? The example URL in https://github.com/yt-dlp/yt-dlp/pull/7147 showed a gigantic banner ad for the apparent owner of the video.

In this PR I selected a more anonymous video for the main test.

PR #32310, opened 2023-06-16
[FFmpegMetadata] Store release_year in the metadata

unkernet:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Many music extractors returns only `release_year` field, and it is not stored in the metadata.

### DIFF
--- a/youtube_dl/postprocessor/ffmpeg.py
+++ b/youtube_dl/postprocessor/ffmpeg.py
@@ -459,7 +459,7 @@ class FFmpegMetadataPP(FFmpegPostProcessor):
         # 4. http://atomicparsley.sourceforge.net/mpeg-4files.html
 
         add('title', ('track', 'title'))
-        add('date', 'upload_date')
+        add('date', ('release_year', 'upload_date'))
         add(('description', 'comment'), 'description')
         add('purl', 'webpage_url')
         add('track', 'track_number')


youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -451,7 +451,7 @@ def add(meta_list, info_list=None):
         # 4. http://atomicparsley.sourceforge.net/mpeg-4files.html
 
         add('title', ('track', 'title'))
-        add('date', 'upload_date')
+        add('date', ('release_year', 'upload_date'))
```
dirkf, line 454: This will set a year rather than the yyyymmdd value if both items are available. Shouldn't they be reversed?

youtube_dl/postprocessor/ffmpeg.py
```diff
@@ -451,7 +451,7 @@ def add(meta_list, info_list=None):
         # 4. http://atomicparsley.sourceforge.net/mpeg-4files.html
```
dirkf, line 451: The `add()` function above could be improved based on yt-dlp (import `T()`, `traverse_obj()`, `txt_or_none()`, `variadic()` from `..utils`):
```py
        def add(meta_list, info_list=None):
            if not meta_list:
                return
            meta_list = variadic(meta_list)
            info_list = variadic(info_list) if info_list else meta_list
            value = traverse_obj(info, (info_list, T(txt_or_none), any))
            if value not in (None, ''):
                value = value.replace('\0', '') # nul character cannot be passed in command line
                metadata.update((meta_f, value) for meta_f in meta_list)
```

youtube_dl/postprocessor/ffmpeg.py
unkernet: I think not. This field called "Year" in most media players.

PR #32481, opened 2023-07-28
[Misc] Under-the-hood fixes, improvements and tests (pt 2)

dirkf:
<details>
<summary>Boilerplate: own/yt-dlp code, bug fix/improvement</summary>

---

- [x] Except for code taken from yt-dlp for which either this or the below has already been asserted, I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)

- [x] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

</details>

---

This PR encapsulates various under-the-hood fixes, improvements and tests.
* Process `Content-Encoding`s according to [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110):
  - support nested encodings
  - support `br` encoding, if _Brotli_ package is installed (`pip install brotli`, ["Only Python 2.7+ is supported."](https://github.com/google/brotli/blob/master/python/README.md) -- Only?)
  - support 'compress' encoding, if _ncompress_ package is installed (`pip install ncompress`, "Requires: Python >=3.8")
  - change: the response `Content-Encoding` header has only unprocessed encodings, or is removed if all encodings were decoded
  - change: the value in the `Content-Length` header is the decoded length
  - `zlib` is used for both deflate and gzip decompression (except for Jython which doesn't support this in 2.7.1 for reasons not investigated)
  - some elements taken from https://github.com/yt-dlp/yt-dlp/pull/7094: thx @coletdjnz.
* Rework URL path munging for ., .. components ( https://github.com/yt-dlp/yt-dlp/issues/3355):
  - revise e8de54b as part of `utils.YoutubeDLHandler`
  - using this, also process `Location` header for redirects
  - use the static (test_utils) and network (test_http) tests from yt-dlp/yt-dlp#7662.

### DIFF
--- a/youtube_dl/compat.py
+++ b/youtube_dl/compat.py
@@ -3200,6 +3200,18 @@ except AttributeError:
     def compat_datetime_timedelta_total_seconds(td):
         return (td.microseconds + (td.seconds + td.days * 24 * 3600) * 10**6) / 10**6
 
+# optional decompression packages
+# PyPi brotli package implements 'br' Content-Encoding
+try:
+    import brotli as compat_brotli
+except ImportError:
+    compat_brotli = None
+# PyPi ncompress package implements 'compress' Content-Encoding
+try:
+    import ncompress as compat_ncompress
+except ImportError:
+    compat_ncompress = None
+
 
 legacy = [
     'compat_HTMLParseError',
@@ -3234,6 +3246,7 @@ __all__ = [
     'compat_Struct',
     'compat_base64_b64decode',
     'compat_basestring',
+    'compat_brotli',
     'compat_casefold',
     'compat_chr',
     'compat_collections_abc',
@@ -3259,6 +3272,7 @@ __all__ = [
     'compat_itertools_zip_longest',
     'compat_kwargs',
     'compat_map',
+    'compat_ncompress',
     'compat_numeric_types',
     'compat_open',
     'compat_ord',
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -278,7 +278,7 @@ jobs:
     #-------- Jython ------
     - name: Set up Java 8
       if: ${{ matrix.python-impl == 'jython' }}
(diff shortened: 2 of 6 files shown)


youtube_dl/compat.py
```diff
@@ -3200,6 +3200,18 @@ def compat_register_utf8():
...
+# PyPi brotli package implements 'br' Content-Encoding
+try:
+    import brotli as compat_brotli
+except ImportError:
+    compat_brotli = None
+# PyPi ncompress package implements 'compress' Content-Encoding
+try:
+    import ncompress as compat_ncompress
+except ImportError:
+    compat_ncompress = None
```
pukkandan, line 3213: Has this ever been observed in the wild? I'm against adding a whole new dependency for a use case that's never encountered.

pukkandan: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding indicates `compress` has been deprecated for decades

dirkf: It says:
> ...  this content-encoding is not used by many browsers today ...

Not by Mozilla, eg. One guy on the Web seemed to [want it](http://www.sqaforums.com/forums/archive/index.php/t-26299.html?X-ANT-WAF-Attempt=1), though it's hardly a yt-dl application and it was indeed more than two decades ago.

However as `compress` is in RFC 9110 and easy* to support, I included it.

It's a fairly trivial point since encoding is negotiated and `compress` is arguably a different case from `br` (Myspace vs Instagram). We would never have seen a bug report if some site offered `compress` but not `gzip` or `deflate`. You could surely be absolved if yt-dlp failed to reflect this feature, even given the slicker dependency system.

Separately, I wonder what sites do today if the client says `Accept-Encoding: identity` (so no encoding). Would that provoke/alleviate fingerprinting issues?

\* Considering negotiation, the PR doesn't **advertise** any optional encodings, though, since I forgot to include that change.

PR #32507, opened 2023-08-15
[postprocessor/ffmpeg] Allow options for mp3

a-rich:
Why?
Options like `preferredquality` are ignored when postprocessing mp3s.

What?
Remove `mp3` from the list of `filecodec` values to check for in the "lossless" condition.

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Why?
Options like `preferredquality` are ignored when postprocessing mp3s.

What?
Remove `mp3` from the list of `filecodec` values to check for in the "lossless" condition.


dirkf:
I'm not convinced about this.

The existing logic tries to avoid re-encoding the audio unless the target container requires it. As I see it, this change on its own will generally result in worse audio quality than what the site offered: if you ask for lower quality you'll get it and if you ask for higher quality you'll get a degraded version of the original with a higher nominal bitrate. Unless you always want to dumb down the MP3 quality to reduce filesize, that doesn't sound good.

You could try this
```diff
-            elif filecodec in ['aac', 'flac', 'mp3', 'vorbis', 'opus']:
+            elif filecodec in ['aac', 'flac', 'vorbis', 'opus'] or (filecodec == 'mp3' and self._preferredquality is None):
```
If you specify `--audio-quality ...`, that should be passed into `self._preferredquality` and cause the re-encoding that you want to occur. A more complex change could extend the same processing to the other audio formats.

The FFMPEG audio option processing is definitely a target for updates. _yt-dlp_ has some more sophisticated processing but I haven't studied it fully.

a-rich:
I see. I've amended the commit to incorporate your suggestion which I feel achieves both aims...

> if you ask for lower quality you'll get it
which satisfies my need to ensure a standardized bit rate

> and [preventing asking] for higher quality [and getting] a degraded version of the original with a higher nominal bitrate
which avoids the needless and unexpected audio degradation

Issue #31462, opened 2023-01-11
Epidemic Sound


UltraBlackLinux:
## Checklist

- [x] I'm reporting a new site support request
- [x] I've verified that I'm running youtube-dl version **2021.12.17**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that none of provided URLs violate any copyrights
- [x] I've searched the bugtracker for similar site support requests including closed ones

## Example URLs

Search anything on https://www.epidemicsound.com/

Here is the audio link: https://dkihjuum4jcjr.cloudfront.net/ES_ITUNES/Think%20It's%20over%20Now%20(Instrumental%20Version)/ES_Think%20It's%20over%20Now%20(Instrumental%20Version).mp3

## Description

Epidemicsound allows the user to "preview" the tracks on the website. Whenever pressing the play button while inspecting the site traffic monitor, one can, see that the client receives a link that corresponds to the track's audio.

This should be reasonably easy to retrieve.

In case this is added, please consider that links from the search tool are an easy way to get a playlist of tracks in the user's taste, so being able to download from a search would be appreciated.

rautamiekka:
They also have multiple YouTube channels.

UltraBlackLinux:
it's hard to search the youtube channels for particular genres

rautamiekka:
Very true, at least if they didn't have genre-specific channels, a long list of them, at that: https://www.youtube.com/@epidemicsound/channels

And each one has playlists for even more specificity, like in the case of jazz that I randomly picked: https://www.youtube.com/@EpidemicJazz/playlists

UltraBlackLinux:
it all happens on teh same page:
1. You click the play button on a song
2. You get a response with the link and some metadata

I also found out that they have some kind of api at https://www.epidemicsound.com/json/search/tracks/ but there is no documentation on this, that's why I'm still sticking to yt-dl

realRobotix:
> This is not a real web site (as it doesn't have URLs for different resources within the site) and so isn't suitable for use with yt-dl. Unless anyone can persuade me otherwise this will be closed `wontfix`.

They do actually have URLs for all resources, but this might also be a recent addition. Just hover over a song/sound effect and on the right there is a link icon which copies the URL to the clipboard. Example: https://www.epidemicsound.com/track/yFfQVRpSPz/

PR #32628, opened 2023-11-06
[Epidemic Sound] Add new extractor

realRobotix:

---

- [ ] Bug fix
- [x] Improvement
- [x] New extractor
- [ ] New feature

---

see #31462

### DIFF
--- a/youtube_dl/extractor/epidemicsound.py
+++ b/youtube_dl/extractor/epidemicsound.py
@@ -18,7 +18,8 @@ class EpidemicSoundIE(InfoExtractor):
         'url': 'https://www.epidemicsound.com/track/yFfQVRpSPz/',
         'md5': 'd98ff2ddb49e8acab9716541cbc9dfac',
         'info_dict': {
-            'id': 'yFfQVRpSPz',
+            'id': '45014',
+            'display_id': 'yFfQVRpSPz',
             'ext': 'mp3',
             'tags': ['foley', 'door', 'knock', 'glass', 'window', 'glass door knock'],
             'title': 'Door Knock Door 1',
@@ -27,15 +28,16 @@ class EpidemicSoundIE(InfoExtractor):
             'timestamp': 1415320353,
             'upload_date': '20141107',
             'age_limit': None,
-              # check that the "best" format was found, since test file MD5 doesn't
-              # distinguish the formats
+            # check that the "best" format was found, since test file MD5 doesn't
+            # distinguish the formats
             'format': 'full',
         },
     }, {
         'url': 'https://www.epidemicsound.com/track/mj8GTTwsZd/',
         'md5': 'c82b745890f9baf18dc2f8d568ee3830',
         'info_dict': {
-            'id': 'mj8GTTwsZd',
+            'id': '148700',
+            'display_id': 'mj8GTTwsZd',
             'ext': 'mp3',
             'tags': ['liquid drum n bass', 'energetic'],
             'title': 'Noplace',
@@ -76,6 +78,7 @@ class EpidemicSoundIE(InfoExtractor):
         self._sort_formats(formats)
 
         info = traverse_obj(json_data, {
+            'id': ('id', T(txt_or_none)),
             'tags': ('metadataTags', Ellipsis, T(txt_or_none)),
             'title': ('title', T(txt_or_none)),
             'duration': ('length', T(float_or_none)),
@@ -91,7 +94,7 @@ class EpidemicSoundIE(InfoExtractor):
         }))
 
         info.update({
-            'id': video_id,
+            'display_id': video_id,
             'formats': formats,
         })
 
--- a/youtube_dl/extractor/epidemicsound.py
+++ b/youtube_dl/extractor/epidemicsound.py
@@ -28,23 +28,21 @@ class EpidemicSoundIE(InfoExtractor):
             'thumbnail': 'https://cdn.epidemicsound.com/curation-assets/commercial-release-cover-images/default-sfx/3000x3000.jpg',
             'timestamp': 1415320353,
             'upload_date': '20141107',
-        }
-    },
-        {
-            'url': 'https://www.epidemicsound.com/track/mj8GTTwsZd/',
-            'md5': 'c82b745890f9baf18dc2f8d568ee3830',
-            'info_dict': {
-                'id': 'mj8GTTwsZd',
-                'ext': 'mp3',
-                'tags': ["liquid drum n bass", "energetic"],
-                'title': 'Noplace',
-                'duration': 237,
-                'thumbnail': 'https://cdn.epidemicsound.com/curation-assets/commercial-release-cover-images/11138/3000x3000.jpg',
-                'timestamp': 1694426482,
-                'upload_date': '20230911',
-            }
         },
-    ]
+    }, {
+        'url': 'https://www.epidemicsound.com/track/mj8GTTwsZd/',
+        'md5': 'c82b745890f9baf18dc2f8d568ee3830',
+        'info_dict': {
+            'id': 'mj8GTTwsZd',
+            'ext': 'mp3',
+            'tags': ['liquid drum n bass', 'energetic'],
+            'title': 'Noplace',
+            'duration': 237,
+            'thumbnail': 'https://cdn.epidemicsound.com/curation-assets/commercial-release-cover-images/11138/3000x3000.jpg',
+            'timestamp': 1694426482,
+            'upload_date': '20230911',
+        },
+    }]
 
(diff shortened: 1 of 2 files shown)


youtube_dl/extractor/epidemicsound.py
```diff
@@ -0,0 +1,37 @@
...
+
+        return {
+            'id': video_id,
+            'url': json_data.get('stems').get('full').get('lqMp3Url'),
+            'tags': json_data.get('metadataTags'),
+            'title': json_data.get('title'),
+            'duration': json_data.get('length'),
+            'timestamp': unified_timestamp(json_data.get('added')),
+            'thumbnail': json_data.get('imageUrl'),
+        }
```
dirkf, line 37: The returned info_dict must have a valid format URL and by convention also a title.

Generally, the data can be type-checked/coerced too.
```suggestion
        info = traverse_obj(json_data, {
            'url': ('stems', 'full', 'lqMp3Url', T(url_or_none)),
            'tags': ('metadataTags', Ellipis, T(txt_or_none)),
            'title': ('title', T(txt_or_none)),
            'duration': ('length', T(float_or_none)),
            'timestamp': ('added', T(unified_timestamp)),
            'thumbnail': ('imageUrl', T(url_or_none)),
        })
        for required_key, target in (('url', 'valid audio URL'), ('title', 'title')):
            if not info.get(required_key):
                raise ExtractorError('No %s found' % (target,))

        info['id'] = video_id
        if info.get('tags') == []:
            del info['tags']

        return info     
```

youtube_dl/extractor/epidemicsound.py
```diff
@@ -0,0 +1,37 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import unified_timestamp
```
dirkf, line 5: To match the other suggestion:
```suggestion
from ..utils import (
    ExtractorError,
    float_or_none,
    txt_or_none,
    unified_timestamp,
    url_or_none,
)

**Review by dirkf — CHANGES_REQUESTED** [review]
Thanks for your work!

I've made a couple of suggestions. Please consider those and check the result of the CI tests too (NB the tests may appear to succeed even if individual extractor tests fail).

**Comment by realRobotix** [comment]
The Extractor is now using traverse_obj and type checking and I also added support for downloading separate tracks.
Example: https://www.epidemicsound.com/track/mj8GTTwsZd/ has separate tracks for Melody, Instruments, Bass, Drums and a Full Mix

File: youtube_dl/extractor/epidemicsound.py [hunk 3]
```diff
@@ -0,0 +1,87 @@
...
+                'ext': 'mp3',
+                'tags': ["liquid drum n bass", "energetic"],
+                'title': 'Noplace',
+                'duration': 237,
+                'thumbnail': 'https://cdn.epidemicsound.com/curation-assets/commercial-release-cover-images/11138/3000x3000.jpg',
+                'timestamp': 1694426482,
+                'upload_date': '20230911',
+            }
+        },
+    ]
```
dirkf, line 47:
```suggestion
        },
    }, {
        'url': 'https://www.epidemicsound.com/track/mj8GTTwsZd/',
        'md5': 'c82b745890f9baf18dc2f8d568ee3830',
        'info_dict': {
            'id': 'mj8GTTwsZd',
            'ext': 'mp3',
            'tags': ['liquid drum n bass', 'energetic'],
            'title': 'Noplace',
            'duration': 237,
            'thumbnail': 'https://cdn.epidemicsound.com/curation-assets/commercial-release-cover-images/11138/3000x3000.jpg',
            'timestamp': 1694426482,
            'upload_date': '20230911',
        },
    }]
```

youtube_dl/extractor/epidemicsound.py
```diff
@@ -0,0 +1,85 @@
...
+        'md5': 'c82b745890f9baf18dc2f8d568ee3830',
+        'info_dict': {
+            'id': 'mj8GTTwsZd',
+            'ext': 'mp3',
+            'tags': ['liquid drum n bass', 'energetic'],
+            'title': 'Noplace',
+            'duration': 237,
+            'thumbnail': 'https://cdn.epidemicsound.com/curation-assets/commercial-release-cover-images/11138/3000x3000.jpg',
+            'timestamp': 1694426482,
+            'upload_date': '20230911',
```
dirkf, line 43: We can't rely on the MD5, so check the `format`:
```suggestion
            'upload_date': '20230911',
            'format': 'full',
```

dirkf (commented):
With the new suggestions (etc) implemented I get this
```
$ python -m youtube_dl -F 'https://www.epidemicsound.com/track/mj8GTTwsZd/'
[EpidemicSound] mj8GTTwsZd: Downloading JSON metadata
[info] Available formats for 148700:
bass         mp3        unknown    track ID 433551 
drums        mp3        unknown    track ID 433547 
instruments  mp3        unknown    track ID 433550 
melody       mp3        unknown    track ID 433548 
full         mp3        unknown    track ID 433549  (best)
$
```
What do you think?

youtube_dl/extractor/epidemicsound.py
```diff
@@ -0,0 +1,85 @@
...
+            'duration': 1,
+            'thumbnail': 'https://cdn.epidemicsound.com/curation-assets/commercial-release-cover-images/default-sfx/3000x3000.jpg',
+            'timestamp': 1415320353,
+            'upload_date': '20141107',
+        },
+    }, {
+        'url': 'https://www.epidemicsound.com/track/mj8GTTwsZd/',
+        'md5': 'c82b745890f9baf18dc2f8d568ee3830',
+        'info_dict': {
+            'id': 'mj8GTTwsZd',
```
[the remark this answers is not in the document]
realRobotix: Not sure why a user would want the id of the json. I'd just leave it out for now.

youtube_dl/extractor/epidemicsound.py
```diff
@@ -0,0 +1,85 @@
...
+            } for key, value in json_data.get('stems').items()
+        ]))
+
+        for f in formats:
+            for key, value in f.items():
+                if value is None or key is None:
+                    del f
+
+        if len(formats) == 0:
+            raise ExtractorError('No downloadable content found')
```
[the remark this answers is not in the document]
realRobotix: I tried using preference but i thought it would sort it automatically. Thanks for pointing that out.

realRobotix:
this should be ready to merge now

youtube_dl/extractor/epidemicsound.py
```diff
@@ -0,0 +1,96 @@
...
+            'id': 'yFfQVRpSPz',
+            'ext': 'mp3',
+            'tags': ['foley', 'door', 'knock', 'glass', 'window', 'glass door knock'],
+            'title': 'Door Knock Door 1',
+            'duration': 1,
+            'thumbnail': 'https://cdn.epidemicsound.com/curation-assets/commercial-release-cover-images/default-sfx/3000x3000.jpg',
+            'timestamp': 1415320353,
+            'upload_date': '20141107',
+            'age_limit': None,
+            'format': 'full',
```
dirkf, line 30: Document the issue that I mentioned since it's unusual to match `format`:
```suggestion
              # check that the "best" format was found, since test file MD5 doesn't
              # distinguish the formats
            'format': 'full',
```

youtube_dl/extractor/epidemicsound.py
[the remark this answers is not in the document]
dirkf: The ID is meant to be a unique value that specifies the video (1-1 mapping). We can't really tell if the slug in the URL is going to be unique, but the numeric ID in the JSON looks like an internal index that ought to be unique. Putting the slug as `display_id` is consistent with the way other extractors work.

Here's another thing, though. Will people normally want the full version only, or the whole set of instrumental tracks? If the latter, the extractor should return all the formats as a playlist, and then the user would have to use `--no-playlist` to force the current behaviour. Or with the current behaviour, you can get multiple formats with comma-separated formats, like `-f "full,drums"`, but the user has to list the formats explicitly as there's no `-f all`.

dirkf (changes requested):
As you say, just about ready to merge, but see the 3 points I've listed.

realRobotix:
I think most people are just interested in the full mix but there is an issue with downloading multiple formats at the same time because the filenames are all the same. What is the recommended way to fix this? Just append the format to the title?

realRobotix:
Ok sure, but I still think most people just want the full version and having to specify that every time would be unnecessary. Should be ready to finally merge now.

dirkf:
And thanks again!

PR #32664, opened 2023-12-13
change regular expression

eukub:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

fixed regular expression for support unicode digits too

### DIFF
--- a/devscripts/gh-pages/update-copyright.py
+++ b/devscripts/gh-pages/update-copyright.py
@@ -19,7 +19,7 @@ from youtube_dl import compat_str
 year = compat_str(datetime.datetime.now().year)
 for fn in glob.glob('*.html*'):
     content = read_file(fn)
-    newc = re.sub(r'(?P<copyright>Copyright © 2011-)(?P<year>[0-9]{4})', 'Copyright © 2011-' + year, content)
+    newc = re.sub(r'(?P<copyright>Copyright © 2011-)(?P<year>\d{4})', 'Copyright © 2011-' + year, content)
     if content != newc:
         tmpFn = fn + '.part'
         write_file(tmpFn, newc)


dirkf:
What digits from Unicode that are not 0-9 do you expect to be found in a year value?

eukub:
> What digits from Unicode that are not 0-9 do you expect to be found in a year value?
For example if you get digits like ０１２３４５６７８９ or like ０１２３４５６７８９ instead of ASCII digits 0123456789 

The numbers from the first two sequences can be processed using \d

This links can be helpful: 
https://docs.python.org/3/library/re.html
https://en.wikipedia.org/wiki/Numerals_in_Unicode

dirkf:
Yes, but:
* this code is run to update the pages of the GH website, and so is rarely used (compared to some code in the YouTube extractor, say) and only by project maintainers
* the value being matched is the year value that was previously set in the same way, unless one of the said maintainers has weirdly typed a Unicode digit other than 0-9 in a manual update
* that value is a py2 `unicode` or py3 `str` containing the result of the Python `datetime` library call a couple of lines up, which is going to match `'[0-9]{4}'` for as long as we care.

I'm not saying that `\d` is wrong and `[0-9]` is better: just that the existing code covers the likely cases and I wouldn't have looked there for something to improve.

eukub:
> Yes, but:
> 
> * this code is run to update the pages of the GH website, and so is rarely used (compared to some code in the YouTube extractor, say) and only by project maintainers
> * the value being matched is the year value that was previously set in the same way, unless one of the said maintainers has weirdly typed a Unicode digit other than 0-9 in a manual update
> * that value is a py2 `unicode` or py3 `str` containing the result of the Python `datetime` library call a couple of lines up, which is going to match `'[0-9]{4}'` for as long as we care.
> 
> I'm not saying that `\d` is wrong and `[0-9]` is better: just that the existing code covers the likely cases and I wouldn't have looked there for something to improve.

If your code has less reasons to fail, that's better, isn't it?

dirkf:
In general, but arguably this code should fail since no weird Unicode digits should be in the scanned text, entirely under the control of the project. If that should happen, it should be investigated.

Whereas, extractor code has to deal with external web pages that might suddenly happen to contain whatever characters.

PR #32672, opened 2023-12-22
👷 use latest major version for actions/checkout

thomasmerz:

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

Use latest major version of [actions/checkout](https://github.com/actions/checkout) for GitHub Actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,7 +122,9 @@ jobs:
           run-tests-ext: sh
     steps:
     - name: Checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@master
+      with:
+        show-progress: false
     #-------- Python 3 -----
     - name: Set up supported Python ${{ matrix.python-version }}
       id: setup-python
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,7 +122,7 @@ jobs:
           run-tests-ext: sh
     steps:
     - name: Checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
     #-------- Python 3 -----
     - name: Set up supported Python ${{ matrix.python-version }}
       id: setup-python
@@ -433,7 +433,7 @@ jobs:
     name: Linter
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,6 +122,7 @@ jobs:
           run-tests-ext: sh
     steps:
     - name: Checkout
+      # v4+ of the action respect show-progress
       uses: actions/checkout@master
       with:
         show-progress: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -436,6 +436,7 @@ jobs:
     name: Linter
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


dirkf:
What would be useful is to go through the CI here, and in the dependent actions repos here, and in the nightly repo (separate PRs, obvs) changing all uses to v4 **and setting `progress: false` in the `with` parameter list (see the action changelog and PR 1067 describing this). If you, or anyone, does do so, pls announce any new PRs from the other repos here.

thomasmerz:
Hey @dirkf , are you interested in my PR or should I close it? Can I help in anyway with your CI or what will be done…? 🤔

dirkf:
I just love the pointless version dependency hell that is the GH actions ecosystem.

Apparently changing to v4 might clear some warnings about node.js versions. Why that needs new major version is far from clear. But we do get the new `--show-progress` option that we should set, as above. Instead we could set version `master`, or whatever syntax, if it exists, that means `v4` or later (is this field actually a git commit-ish ?)

Then we need to make the equivalent changes in the [nightly](https://github.com/ytdl-org/ytdl-nightly) repo.

.github/workflows/ci.yml
```diff
@@ -122,7 +122,7 @@ jobs:
           run-tests-ext: sh
     steps:
     - name: Checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
```
dirkf, line 125: 1. Avoid having to change the version again (maybe?).
2. Inhibit checkout progress spam.
```suggestion
      uses: actions/checkout@master
      with:
        show-progress: false
```

.github/workflows/ci.yml
```diff
@@ -433,7 +433,7 @@ jobs:
     name: Linter
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
```
dirkf, line 436:
```suggestion
      uses: actions/checkout@master
      with:
        show-progress: false
```

dirkf (changes requested):
What do you think?

.github/workflows/ci.yml
```diff
@@ -122,7 +122,9 @@ jobs:
           run-tests-ext: sh
     steps:
     - name: Checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@master
```
dirkf, line 125:
```suggestion
      # v4+ of the action respect show-progress
      uses: actions/checkout@master
```

.github/workflows/ci.yml
```diff
@@ -433,7 +436,9 @@ jobs:
     name: Linter
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+      uses: actions/checkout@master
```
dirkf, line 439:
```suggestion
    - name: Checkout
      uses: actions/checkout@master
```

dirkf:
That looks better! But I'm still undecided whether it's appropriate to use `@v4`, `@master`, or no version.

To expand on https://github.com/ytdl-org/youtube-dl/pull/32672#issuecomment-2150330366, in principle it should be enough to use `@v4`. However the action devs don't maintain a `v4` interface. Instead if they want to use a new version of, say, Node (an internal change that should not affect action functionality) they make the change in the latest release only, so propagating their technical debt to all users of the action. Those users then have to update the action version even if they are not using any new features of the action.

thomasmerz:
Hey @dirkf , are you still undecided about this? I'm looking forward to close my PR if it's of no use to this project… Please give feedback if I can close or if you want some changes.

PR #32752, opened 2024-03-23
[Youtube] Fix AttributeError no attribute '_YoutubeIE__ie_msg' if unable to decode n-parameter

gy-chen:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature


dirkf (approved):
Thanks for the catch. This is a bleeding chunk of code moved to `InfoExtractor`, as you've plainly found. Your fix is what should have been committed.

dirkf:
This (should-be) obvious bug isn't being detected by _flake8_ or even _pylint_, which does detect a less complicated case as `E1101: Instance of '{class}' has no '__private_meth' member (no-member)`.

PR #32762, opened 2024-04-03
[extractor/common] Fix crash in _report_ignoring_subs

salty-horse:

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

When subs are not available and `_report_ignoring_subs` is called, it crashes because it passes an argument to `bug_reports_message`, which accepts none.

I changed it to be appended to the message, like what's done in other places.
Now the message is like this:

> WARNING: [youtube] Ignoring subtitle tracks found in the DASH manifest; if any subtitle tracks are missing, ; please report this issue on https://yt-dl.org/bug . Make sure you are using the latest version; see  https://yt-dl.org/update  on how to update. Be sure to call youtube-dl with the --verbose flag and include its complete output.

The ", ;" is not correct but I don't know if you want to get rid of the semicolon. The important thing is that it doesn't crash, and the file can be downloaded.

### DIFF
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -1812,10 +1812,10 @@ class InfoExtractor(object):
         }
 
     def _report_ignoring_subs(self, name):
-        self.report_warning(bug_reports_message(
+        self.report_warning(
             'Ignoring subtitle tracks found in the {0} manifest; '
-            'if any subtitle tracks are missing,'.format(name)
-        ), only_once=True)
+            'if any subtitle tracks are missing, {1}'.format(name, bug_reports_message()),
+            only_once=True)
 
     def _extract_m3u8_formats(self, m3u8_url, video_id, ext=None,
                               entry_protocol='m3u8', preference=None,
--- a/youtube_dl/extractor/common.py
+++ b/youtube_dl/extractor/common.py
@@ -1812,10 +1812,10 @@ class InfoExtractor(object):
         }
 
     def _report_ignoring_subs(self, name):
-        self.report_warning(
+        self.report_warning(bug_reports_message(
             'Ignoring subtitle tracks found in the {0} manifest; '
-            'if any subtitle tracks are missing, {1}'.format(name, bug_reports_message()),
-            only_once=True)
+            'if any subtitle tracks are missing,'.format(name)
+        ), only_once=True)
 
     def _extract_m3u8_formats(self, m3u8_url, video_id, ext=None,
                               entry_protocol='m3u8', preference=None,
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -2376,10 +2376,7 @@ def bug_reports_message(before=';'):
         update_cmd = 'type  youtube-dl -U  to update'
     else:
         update_cmd = 'see  https://yt-dl.org/update  on how to update'
-    msg = '; please report this issue on https://yt-dl.org/bug .'
-    msg += ' Make sure you are using the latest version; %s.' % update_cmd
-    msg += ' Be sure to call youtube-dl with the --verbose flag and include its complete output.'
-    return msg
+    return '; please report this issue on https://yt-dl.org/bug .'
 
 
 class YoutubeDLError(Exception):
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -2375,8 +2375,20 @@ def bug_reports_message(before=';'):
     if ytdl_is_updateable():
         update_cmd = 'type  youtube-dl -U  to update'
     else:
-        update_cmd = 'see  https://yt-dl.org/update  on how to update'
-    return '; please report this issue on https://yt-dl.org/bug .'
+        update_cmd = 'see  https://github.com/ytdl-org/youtube-dl/#user-content-installation  on how to update'
+
+    msg = (
+        ' please report this issue on https://github.com/ytdl-org/youtube-dl/issues ,'
+        ' using the appropriate issue template.'
+        ' Make sure you are using the latest version; %s.'
+        ' Be sure to call youtube-dl with the --verbose option and include the complete output.'
+    ) % update_cmd
+
+    before = (before or '').rstrip()
+    if not before or before.endswith(('.', '!', '?')):
+        msg = msg[0].title() + msg[1:]
+
+    return (before + ' ' if before else '') + msg
 
 
 class YoutubeDLError(Exception):
--- a/youtube_dl/utils.py
+++ b/youtube_dl/utils.py
@@ -2371,7 +2371,7 @@ def make_HTTPS_handler(params, **kwargs):
         return YoutubeDLHTTPSHandler(params, context=context, **kwargs)
 
 
-def bug_reports_message():
+def bug_reports_message(before=';'):
     if ytdl_is_updateable():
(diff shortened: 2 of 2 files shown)


dirkf:
Thanks, good catch.

There was supposed to be a matching change to `utils.py`: I'd prefer to add that rather than change the code in `extractor/common.py`.

Would you like to revert your change and apply the missing change (adapted from yt-dlp) instead?

```diff
--- old/youtube-dl/youtube_dl/utils.py
+++ new/youtube-dl/youtube_dl/utils.py
@@ -2371,15 +2371,24 @@
         return YoutubeDLHTTPSHandler(params, context=context, **kwargs)
 
 
-def bug_reports_message():
+def bug_reports_message(before=';'):
     if ytdl_is_updateable():
         update_cmd = 'type  youtube-dl -U  to update'
     else:
-        update_cmd = 'see  https://yt-dl.org/update  on how to update'
-    msg = '; please report this issue on https://yt-dl.org/bug .'
-    msg += ' Make sure you are using the latest version; %s.' % update_cmd
-    msg += ' Be sure to call youtube-dl with the --verbose flag and include its complete output.'
-    return msg
+        update_cmd = 'see  https://github.com/ytdl-org/youtube-dl/#user-content-installation  on how to update'
+
+    msg = (
+        ' please report this issue on https://github.com/ytdl-org/youtube-dl/issues ,'
+        ' using the appropriate issue template.'
+        ' Make sure you are using the latest version; %s.'
+        ' Be sure to call youtube-dl with the --verbose option and include the complete output.'
+    ) % update_cmd
+
+    before = (before or '').rstrip()
+    if not before or before.endswith(('.', '!', '?')):
+        msg = msg[0].title() + msg[1:]
+
+    return (before + ' ' if before else '') + msg
 
 
 class YoutubeDLError(Exception):
```

salty-horse:
That makes much more sense!

I'm uncomfortable with copy-pasting someone else's code without credit - I'd much rather you commit it as originally intended.

youtube_dl/extractor/common.py
```diff
@@ -1812,10 +1812,10 @@ def _m3u8_meta_format(self, m3u8_url, ext=None, preference=None, m3u8_id=None):
         }
 
     def _report_ignoring_subs(self, name):
-        self.report_warning(bug_reports_message(
+        self.report_warning(
             'Ignoring subtitle tracks found in the {0} manifest; '
-            'if any subtitle tracks are missing,'.format(name)
-        ), only_once=True)
+            'if any subtitle tracks are missing, {1}'.format(name, bug_reports_message()),
+            only_once=True)
```
dirkf, line 1818: Revert and fix `utils.bug_reports_message()` (1):
```suggestion
        self.report_warning(bug_reports_message(
            'Ignoring subtitle tracks found in the {0} manifest; '
            'if any subtitle tracks are missing,'.format(name)
        ), only_once=True)
```

youtube_dl/utils.py
```diff
@@ -2371,15 +2371,12 @@ def set_alpn_protocols(ctx):
...
+def bug_reports_message(before=';'):
     if ytdl_is_updateable():
         update_cmd = 'type  youtube-dl -U  to update'
     else:
         update_cmd = 'see  https://yt-dl.org/update  on how to update'
-    msg = '; please report this issue on https://yt-dl.org/bug .'
-    msg += ' Make sure you are using the latest version; %s.' % update_cmd
-    msg += ' Be sure to call youtube-dl with the --verbose flag and include its complete output.'
-    return msg
+    return '; please report this issue on https://yt-dl.org/bug .'
```
dirkf, line 2379: Revert and fix `utils.bug_reports_message()` (2):
```suggestion
    else:
        update_cmd = 'see  https://github.com/ytdl-org/youtube-dl/#user-content-installation  on how to update'

    msg = (
        ' please report this issue on https://github.com/ytdl-org/youtube-dl/issues ,'
        ' using the appropriate issue template.'
        ' Make sure you are using the latest version; %s.'
        ' Be sure to call youtube-dl with the --verbose option and include the complete output.'
    ) % update_cmd

    before = (before or '').rstrip()
    if not before or before.endswith(('.', '!', '?')):
        msg = msg[0].title() + msg[1:]

    return (before + ' ' if before else '') + msg
```

dirkf (commented):
With my maintainer superpower, I (will) have added some commits into your PR branch to make my suggested change.

It would be a great help if you'd be able to pull the latest version of the branch and test that the issue is fixed (since I'm deep in the mind-numbing twaddle that fixing the TikTok extractor brings). 

Once it's OK, I can merge the PR as a single commit credited to us jointly with acknowledgments to yt-dlp, and then it can be in the next nightly build.

youtube_dl/extractor/common.py
dirkf, line 1818: Revert and fix `utils.bug_reports_message()` (1b):
```suggestion
            'if any subtitle tracks are missing,'.format(name)
        ), only_once=True)
```

youtube_dl/utils.py
```diff
@@ -2371,15 +2371,12 @@ def set_alpn_protocols(ctx):
         return YoutubeDLHTTPSHandler(params, context=context, **kwargs)
 
 
-def bug_reports_message():
+def bug_reports_message(before=';'):
     if ytdl_is_updateable():
         update_cmd = 'type  youtube-dl -U  to update'
     else:
         update_cmd = 'see  https://yt-dl.org/update  on how to update'
```
dirkf, line 2378: Revert and fix `utils.bug_reports_message()` (2a):
```suggestion
        update_cmd = 'see  https://github.com/ytdl-org/youtube-dl/#user-content-installation  on how to update'
```

salty-horse:
The video that caused the issues no longer does (probably because it was a live video on youtube, and after a while, I think they re-encode it in a different format.) But I forced the error, and it works fine.

My only suggestion is removing the whitespace before `please report this issue`, because it the error has a double space:
`if any subtitle tracks are missing,  please report this issue`

youtube_dl/utils.py
```diff
@@ -2371,15 +2371,24 @@ def set_alpn_protocols(ctx):
...
     else:
-        update_cmd = 'see  https://yt-dl.org/update  on how to update'
-    msg = '; please report this issue on https://yt-dl.org/bug .'
-    msg += ' Make sure you are using the latest version; %s.' % update_cmd
-    msg += ' Be sure to call youtube-dl with the --verbose flag and include its complete output.'
-    return msg
+        update_cmd = 'see  https://github.com/ytdl-org/youtube-dl/#user-content-installation  on how to update'
+
+    msg = (
+        ' please report this issue on https://github.com/ytdl-org/youtube-dl/issues ,'
```
dirkf, line 2381: Remove initial space that is added in l.2391 if needed.
```suggestion
        'please report this issue on https://github.com/ytdl-org/youtube-dl/issues ,'
```

dirkf:
Yes, the crash would only have been seen in this specific combination of circumstances:
* the selected format is a DASH manifest
* the manifest has subtitles
* the extractor calls the original yt-dl DASH extraction API `_extract_mpd_formats()`, rather than the newer yt-dlp API `_extract_mpd_formats_and_subtitles()` (this might be all yt-dl extractors at the moment).

Many thanks for your help!

PR #32780, opened 2024-05-01
indent json with --pretty-json

knighthat:

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [x] New feature

---

This switch `--pretty-json` prints result of `--dump-json` in a more human readable format. Applicable in storing video's information. 

**Reason**: I am making a wrapper called [youtube-dl-java](https://github.com/knighthat/youtube-dl-java). Loading 1 line of JSON with hundreds of thousands of characters takes a significantly long time due to how much memory is allocated for 1 single line.

This also makes it easier for users to inspect the return json since it's nicely indented.

### DIFF
--- a/youtube_dl/YoutubeDL.py
+++ b/youtube_dl/YoutubeDL.py
@@ -1914,7 +1914,10 @@ class YoutubeDL(object):
             self.to_stdout(formatSeconds(info_dict['duration']))
         print_mandatory('format')
         if self.params.get('forcejson', False):
-            self.to_stdout(json.dumps(self.sanitize_info(info_dict)))
+            if self.params.get('indent_json', False):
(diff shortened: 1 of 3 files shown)


dirkf:
Thanks for your proposal.

I don't think we should add a feature just to help your programming project. Instead, ask whether you are loading the JSON into your Java program efficiently, or why Python can handle files of this sort (eg, `--load-info-json`) without complaint. There seem to be quite a few libraries dedicated to doing this for Java.

Also, there are many external tools available for JS[ON] parsing and beautification, such as _jq_.

knighthat:
The main purpose of this feature is to let user save outputs to physical file. There are tools out there and I'm using it daily too but why do I have to install or write a script to parse it when python can just do the same? 

Btw, copying 1 line of string with hundreds of thousands of characters is a real pain.

PR #32786, opened 2024-05-16
Treat empty format selection string as "anything goes".

mk-pmb:

---

- [x] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Empty format selection string always leads to `ExtractorError: requested format not available`, even if there clearly are some. If the user has no criteria at all, i.e. anything goes, then `worst` obviously is good enough for them, but a lie is not.

I had considered giving them `best`, but for users who really don't care about media quality, we should optimize for low power consumption in all affected equipment.

### DIFF
--- a/test/test_YoutubeDL.py
+++ b/test/test_YoutubeDL.py
@@ -136,6 +136,11 @@ class TestFormatSelection(unittest.TestCase):
         ]
         info_dict = _make_result(formats)
 
+        ydl = YDL({'format': ''}) # no criteria => anything goes
+        ydl.process_ie_result(info_dict.copy())
+        downloaded = ydl.downloaded_info_dicts[0]
+        self.assertEqual(downloaded['format_id'], '35')
+
         ydl = YDL({'format': '20/47'})
         ydl.process_ie_result(info_dict.copy())
         downloaded = ydl.downloaded_info_dicts[0]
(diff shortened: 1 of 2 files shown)


dirkf:
But there's a default if the option isn't mentioned.

Is this an API issue?

mk-pmb:
Yes, it is an API issue. What I really wanted to do was submit another test case, but to express that other test case in a simple way, I need this patch here first. Otherwise I'd have to include a workaround for this issue here and explain the seeming lie.

mk-pmb:
What's the current obstacle?

dirkf:
Is there an issue for this? I think it needs some examples. 

Is this the point?
* the API requires you to specify a format selection (ie the default is to match no format)
* whereas the CLI provides a default, as documented.

If so, it's up to the API user to provide a format selection. We'd have to consider the impact of changing that behaviour. 

Arguably the exception raised should clarify when no format is found because no format selection was provided vs when there were actually no formats matching the specified selection.

Or, the API default should be to match all formats.

Or, a default of `worst` should probably be `worst/worstvideo+worstaudio`.

mk-pmb:
It's part of issue #30742. The test case that I wanted to get going was the one [in this addition to test_YoutubeDL.py](https://github.com/ytdl-org/youtube-dl/pull/30742/changes#diff-f1abc5b91afa326c01cbc20d98d5fe6e2fb38ea0e03bcc31d7f1a5b5ee23e0ea). I wanted to add a baseline test with no restrictions, and once that's accepted, add tests for my new shortside/longside restrictions. Having tests where addition of my new criteria are the only difference will <del>cleanly prove</del> <ins>give very strong indication</ins> whether their implementation is at fault, vs. random side effects from other code.

PR #32844, opened 2024-07-10
Output name of the program for --version

tansy:

---

- [ x ] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [ na ] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [ x] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [ x] Covered the code with tests (note that PRs without tests will be REJECTED)

- [ x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

-----

'The GNU standard --version option should direct the program to print information about its name, version' So added a name of program to version displaye along with shotr option \`-V'.

```
$ python youtube-dl.py -V
youtube-dl 2024.07.08
```

### DIFF
--- a/youtube_dl/options.py
+++ b/youtube_dl/options.py
@@ -118,7 +118,7 @@ def parseOpts(overrideArguments=None):
     fmt.format_option_strings = _format_option_string
 
     kw = {
-        'version': __version__,
+        'version': 'youtube-dl '+__version__,
         'formatter': fmt,
         'usage': '%prog [OPTIONS] URL [URL...]',
         'conflict_handler': 'resolve',
@@ -132,7 +132,7 @@ def parseOpts(overrideArguments=None):
         action='help',
         help='Print this help text and exit')
     general.add_option(
-        '--version',
+        '-V', '--version',
         action='version',
         help='Print program version and exit')
     general.add_option(


dirkf:
Someone should have pointed this out ca. 2010, before people wrote (I expect) lots of scripts that parse the output of the `--version` option as `yyyy.mm.dd`. Can I suggest, say, 
```
youtubedl() {
    case "$1" in
    -V|--version) printf "youtube-dl " ;;
    esac
    command youtube-dl "$@"
}
alias youtube-dl=youtubedl
```
Or something similar in a shell other than POSIX.

tansy:
> Someone should have pointed this out ca. 2010

Ca. 2010 i didn't know about this program. I'm sure that 95% of present users didn't know it either.

> before people wrote (I expect) lots of scripts

I'm certain there ate lots of Scots that use ytdl

> that parse the output of the `--version` option

But none of them, ok 99.(9)% of them don't paste version number.
What for?

>  as `yyyy.mm.dd`

As above. I bet here and now that no one parses version number.
What for?
The only reason to check version is to report a bug. But maybe that's just me.


> Can I suggest, say,

Sure.

> alias youtube-dl=youtubedl

`youtubedl "$@"` works better.

> Or something similar in a shell other than POSIX.

In non POSIX is not posixble.

--

I will leave it, if I can, and wait for objections or consent.

dirkf:
Fair enough. There are other ways in which the command-line syntax doesn't fully behave as GNU-ishly as one would expect, basically because of the Python command line parsing module used.

Maybe someone will find a way to fix this and then we could have a `--gnuishly-correct` option together with which `--version` could include the program name and possibly other information such as licensing and versions of dependencies (eg _ffmpeg_).

The same issue still applies in _yt-dlp_:
```console
$ ~/Downloads/yt-dlp-2024.07.02 --version
2024.07.02
$ 
```

tansy:
I try to reason it this way - it makes sense to announce program name when asking for a version. Cause version of what? It's especially true here, where date is used as a version. What do you think when you see this thing: \`2024-01-01', cause I see a date. What that date refers to is not known.
What do you think when you see this: \`gzip 1.13'? First think that comes to a mind is gzip, version 1.13.
On the other hand, I referred to [GNU coding standards](https://www.gnu.org/prep/standards/html_node/_002d_002dversion.html) as they are simple and concise. Other programs include all kinds od unnecessary rap to their version print. So this was the simplest and most logical of all.

> The same issue still applies in yt-dlp

Anecdotal evidence. Hundreds of other programs do include their name in version print.

> Maybe someone will find a way to fix this and then we could have a --gnuishly-correct option together with which --version could include the program name and possibly other information such as licensing and versions of dependencies (eg ffmpeg).

I quoted GNU as the option is \`--version' brings to mind GNU coding standards rather than any other 'code of conduct'.

If it was put in a function that would do it, would it be better?
I can do it so it would print \`program-name version' in basic option and, in verbose, it would add some more stuff. Maybe like this:

```
$ youtube-dl --version
youtube-dl 2011-11-11

$ youtube-dl --version --verbose
youtube-dl 2011-11-11
Copyright/left (C) 2011
This is free and unencumbered software released into the public domain.
For more information, please refer to <http://unlicense.org/>
There is NO WARRANTY, to the extent permitted by law.
```

and in help:

```
    -V, --version                        Print program version and exit. Verbose output prints also licence.
```

Where shall I put this function, in \`version.py' or \`utils.py' or somewhere else?
And what is \`action='store_true''? \`dest='update_self'' is, as I guess, the function to call.

tansy:
Just realised that it would also immediately show that user uses \`youtube-dl' and not \`yt-dlp'.

PR #32849, opened 2024-07-11
zip better release

tansy:

---

- [na] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [ x] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [ x] Covered the code with tests (note that PRs without tests will be REJECTED)

- [ ] Bug fix
- [ x ] Improvement
- [ ] New extractor
- [ ] New feature

---

It makes sense to compress release version the best possible as it is done only once and then decompressed many times over.
It also doesn't cost much - here, less than 0.25 s more during make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ youtube-dl: youtube_dl/*.py youtube_dl/*/*.py
 	done
 	touch -t 200001010101 zip/youtube_dl/*.py zip/youtube_dl/*/*.py
 	mv zip/youtube_dl/__main__.py zip/
-	cd zip ; zip -q ../youtube-dl youtube_dl/*.py youtube_dl/*/*.py __main__.py
+	cd zip ; zip -q -9 ../youtube-dl youtube_dl/*.py youtube_dl/*/*.py __main__.py
 	rm -rf zip
 	echo '#!$(PYTHON)' > youtube-dl
 	cat youtube-dl.zip >> youtube-dl


dirkf:
The default compression is `-6` which gives a 12% smaller than `-1` for 30% more `time` (against the current codebase).

The trade-off for `-9` is x2 CPU vs 7% better compression. This example shows why `-6` was set to be the default.

But as far as I can tell there's no reason why you couldn't use the higher, or any, compression level in your builds, since all Python versions that we support have a compatible `zlib` module.

tansy:
On ancient computer:

```
$ time make -f Makefile youtube-dl

real    0m2.048s
user    0m1.426s
sys     0m0.539s

$ time make -f Makefile-zip-9 youtube-dl

real    0m2.128s
user    0m1.781s
sys     0m0.274s
```
difference in time is, literally, blink of an eye.

Even decompression is faster. Slightly but it is.

Overall gain is small, but the cost is even smaller. It's imperceptible.

--
I'd like to take this opportunity to ask why is 'executable' (\`youtube-dl' pyexe) included in source package (youtube-dl.tar.gz)? It's already generated as executable and then included in source. I guess noone is using it as a source but still, the question remains.

PR #32876, opened 2024-07-23
[ci] Implement release workflow

### LINKED ISSUE #31585: New Release?
## Checklist

- [ ] I'm asking a question
- [ ] I've looked through the README and FAQ for similar questions
- [ ] I've searched the bugtracker for similar questions including closed ones

## Question

Preface: 
There have been a number of fixes done since the last release (2021-12-17) {over a years worth} The error messages refer using youtube-dl -U (which is already in place). All the errors I am experiencing are already closed issues, but no no release so youtube-dl -U won't fix it. [Issue #31583 ](https://github.com/ytdl-org/youtube-dl/issues/31583)

Question:
Can either the instructions be updated (and the error message refering to youtube-dl -U be dropped) to "install" via a github pull from Master

or

Can a new release be cut?

Thanks

Grub4K:

---

  - See [Grub4K/youtube-dl@`master`](https://github.com/Grub4K/youtube-dl/tree/master) for successful application of this PR, and the [2024.07.23](https://github.com/Grub4K/youtube-dl/releases/tag/2024.07.23) and [2024.07.23.1](https://github.com/Grub4K/youtube-dl/releases/tag/2024.07.23.1) for example releases, including changelog. Writing tests for CI results is hard

- [x] New feature

---

Implements an automated release workflow, similar to yt-dlp's (Rewrite of the nightly workflow as an automated `workflow_dispatch` action). Also removes the old, obsolete build scripts in favor of using the current workflow.

It supports:
- Arbitrary version tags
- Prereleases
- Automated changelog generation
  - This is more hacked than correct since the commit format is non normative.
    `make_changelog.py` can be backported if the yt-dlp commit format will be adapted
- Separate build workflow
  - Each step is separated so adding more targets is trivial

After this is merged, the only barrier to publishing a new release will be running the "Release" workflow.

Closes #31585


.github/workflows/build.yml
```diff
@@ -0,0 +1,170 @@
...
+      # Temporary workaround for Python 3.4/5 failures - May 2024
+      PIP_TRUSTED_HOST: "pypi.python.org pypi.org files.pythonhosted.org"
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python 3.4
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.4"
```
Grub4K, line 85: This is used bc thats what the old youtub-dl executable used. Mayhaps it would make sense to bump this eventually?

.github/workflows/build.yml
```diff
@@ -0,0 +1,170 @@
...
+            youtube-dl.tar.gz
+          compression-level: 0
+
+  windows32:
+    if: inputs.windows32
+    runs-on: windows-2022
+    env:
+      PYCRYPTO: pycrypto-2.6.1-cp34-none-win32
+      # Temporary workaround for Python 3.4/5 failures - May 2024
+      PIP_TRUSTED_HOST: "pypi.python.org pypi.org files.pythonhosted.org"
```
Grub4K, line 77: Coped over, but no idea what this means

devscripts/changelog.py
```diff
@@ -0,0 +1,100 @@
...
+    if not markdown:
+        return '* ' + short
+
+    return '* [%s](https://github.com/ytdl-org/youtube-dl/commit/%s)' % (short, sha)
+
+
+def generate_changelog(markdown):
+    most_recent_tag = run([
+        'git', 'tag', '--list', '--sort=-v:refname',
+        '????.??.??', '????.??.??.?',
```
Grub4K, line 42: Asumes no more than 9 patch release will be made. I think that is reasobable

Grub4K (commented):
I could not test the devscripts under 2.6, but also thats not required as far as the build process goes, so...

Additional targets can be added later to the `build.yml` if desired

.github/workflows/build.yml
dirkf: As long as 3.4 is supported in the codebase, it can/should stay at this value, because it means that the bundled .exe runs on Windows XP.

At one time there was a [graphic that explained this in Lesmiscore's PR workflow](https://github.com/ytdl-org/youtube-dl/pull/30644/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721
>    # reason to choose 3.4: https://media.discordapp.net/attachments/807245652072857613/942409077701619742/unknown.png

but owing to Discordance this has been discarded.

The workflow gets or prepares some antique wheels to provide crypto support for this Python version.

.github/workflows/build.yml
dirkf: The TLS parameters for the listed hosts have been upgraded so that you can't connect with an old _pip_ -- [except with this work-around](https://github.com/ytdl-org/youtube-dl/commit/a08f2b7e4567cdc50c0614ee0a4ffdff49b8b6e6), essentially a `--no-check-certificate` for those hosts.

Security measures as denial-of-service: a very relevant issue this week.

.github/workflows/build.yml
```diff
@@ -0,0 +1,170 @@
...
+            youtube-dl
+            youtube-dl.tar.gz
+          compression-level: 0
+
+  windows32:
+    if: inputs.windows32
+    runs-on: windows-2022
+    env:
+      PYCRYPTO: pycrypto-2.6.1-cp34-none-win32
+      # Temporary workaround for Python 3.4/5 failures - May 2024
```
dirkf, line 76:
```suggestion
      # Workaround for Python 3.4/5 PyPi certificate verification failures - May 2024
```

devscripts/changelog.py
```diff
@@ -0,0 +1,100 @@
...
+from __future__ import unicode_literals
+
+import os
+import subprocess
+import sys
+
+
+def run(args):
+    process = subprocess.Popen(args, stdout=subprocess.PIPE, universal_newlines=True)
+    return process.communicate()[0].strip()
```
dirkf, line 11: Should this be using `utils.process_communicate_or_kill()`?

.github/workflows/build.yml
Grub4K: The graphic does not explain anything. It just shows a binary running 3.4. If you desire to support 3.4 forever then i guess it will have to stay this way

devscripts/changelog.py
Grub4K: While that would work, yeah, it would require me importing `youtube_dl`. Since the process is short running and well behaving (never had problems with `git` binary before) i think its not needed

dirkf:
> I could not test the devscripts under 2.6, but also thats not required as far as the build process goes, so...

Exactly. Anyone who wants to run the scripts with 2.6 should be able to fix any problems.

> Additional targets can be added later to the `build.yml` if desired

Sure. But this is making the same ones as the nightly, isn't it, and that's all we need for now, except that maybe there should be two versions of the `youtube-dl` bundled POSIX builds, adding one with `#!/usr/bin/env python3`.

.github/workflows/build.yml
dirkf: Apparently Windows XP is quite popular in some parts of the world. As long as we're supporting 2.7, 3.4 is not much of a stretch.

Grub4K:
I think adding another target for this might not make sense. Many modern os have already switched to `python` pointing to python 3. Additionally, there is `python-is-python3` if need be.

If you disagree ill add the target though. Just let me know what to call it

dirkf:
> I think adding another target for this might not make sense. Many modern os have already switched to `python` pointing to python 3. Additionally, there is `python-is-python3` if need be.
> 
> If you disagree ill add the target though. Just let me know what to call it

It seems that macOS typically installs a `python3` only, with no option of a _python-is-python3_: those users are probably less likely to understand, or be amenable to, creating a symlink. And who knows what "security" feature Apple may introduce to confound this.

If only there were some way to formulate a `#!` command that would find both `python` and `python3` in some order. I haven't been able to find one that doesn't involve shipping and installing a further script, like this `python_any`:
```sh
#!/bin/sh
exec python3 "$@" || exec python "$@"
```

In all the systems where I've tried it, a _sed_ command works (preserving the executable permission if it was set):
```sh
sed -i -e '1 s/python$/python3/' youtube-dl
```

As POSIX-y systems are meant to hide any platform differences between "text" and "binary" files, maybe this is good enough, once the shipped default has been decided.

Otherwise, a separate target, say `youtube-dl-py3`, seems to be the path of least resistance.

Grub4K:
Done. See [2024.07.25](<https://github.com/Grub4K/youtube-dl/releases/tag/2024.07.25>) release for results

bashonly (approved):

Grub4K:
Side note, @dirkf, do you have access to the PyPI package?

Grub4K:
What is the holdup on this? Anything that can be done to get this merged?

Grub4K:
@dirkf is this PR being considered?

dirkf:
Definitely, but #32905 needs to be sorted before I can get back to it.

bashonly:
Isn't #32905 pretty much sorted now?

dirkf:
The YT extractor has been hacked for now but https://github.com/ytdl-org/youtube-dl/issues/33035#issuecomment-2621201479.

bashonly:
Why is that a prerequisite to a new release? The last release is completely broken. The master branch works. PO Token support in yt-dlp is mostly just future-proofing

gustawdaniel:
IMHO this should be merged because it is blocking for installation on mac now.

mk-pmb:
Is there something I can help to get this going?

Grub4K:
@mk-pmb I no longer think that @dirkf has any interest to do a proper release so I am closing this PR on basis of that.

PR #32931, opened 2024-09-26
Update kommunetv.py

hmlb-no:
Look at https://www.aventia.no/kommunetv/vare-kunder/

They are also in Sweeden.

Here is a problem tho.

https://stavanger.kommunetv.no/live/841

https://stavanger.kommunetv.no/archive/841

what if there is a live and a archive at the same time? how dose that work?

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

### DIFF
--- a/youtube_dl/extractor/kommunetv.py
+++ b/youtube_dl/extractor/kommunetv.py
@@ -6,7 +6,8 @@ from ..utils import update_url
 
 
 class KommunetvIE(InfoExtractor):
-    _VALID_URL = r'https://(\w+).kommunetv.no/archive/(?P<id>\w+)'
+    _VALID_URL = r'https?://\w+\.kommun(?:etv\.no|\.tv)/(?:archive|live)/(?P<id>\w+)'
+
     _TEST = {
         'url': 'https://oslo.kommunetv.no/archive/921',
         'md5': '5f102be308ee759be1e12b63d5da4bbc',


dirkf:
>Here is a problem tho.
>
>https://stavanger.kommunetv.no/live/841
>
>https://stavanger.kommunetv.no/archive/841
>
>what if there is a live and a archive at the same time? how dose that work?

If those are not the same thing (eg `.../live/nnnn` while live and `.../archive/nnnn` afterwards), the `id` could contain the last two elements of the path instead of the just the last element. Or some `id` that is unique across both path types may be hidden in the page, in which case that should be used.

youtube_dl/extractor/kommunetv.py
```diff
@@ -6,7 +6,8 @@
 
 
 class KommunetvIE(InfoExtractor):
-    _VALID_URL = r'https://(\w+).kommunetv.no/archive/(?P<id>\w+)'
+    _VALID_URL = r'https?://\w+\.kommun(?:etv\.no|\.tv)/(?:archive|live)/(?P<id>\w+)'
+
     _TEST = {
```
dirkf, line 11: Make this 
```py
    _TESTS = [{
...
    }]
```
and add tests for the various URL types.

hmlb-no:
> > Here is a problem tho.
> > https://stavanger.kommunetv.no/live/841
> > https://stavanger.kommunetv.no/archive/841
> > what if there is a live and a archive at the same time? how dose that work?
> 
> If those are not the same thing (eg `.../live/nnnn` while live and `.../archive/nnnn` afterwards), the `id` could contain the last two elements of the path instead of the just the last element. Or some `id` that is unique across both path types may be hidden in the page, in which case that should be used.


Yes they are different content. The central difference is that the 'live' ID is for ween they are live and then after they are live the get converted into a archive but with a different ID. But ween a there exist a archive and ween they are livestraming tehy are not downloading the live but instead youtube-dl downloads the archive woon.


For instance https://stavanger.kommunetv.no/live/841 gets converted into a new ID and that is https://stavanger.kommunetv.no/archive/1142

youtube_dl/extractor/kommunetv.py
hmlb-no: Yes, I have considered doing something like that, but I'm unsure how to obtain the MD5 values.

Alternatively, I could create a new function specifically for live streams.

dirkf:
live/841 corresponds to streamType=2&id=841 in the API;
archive/1142 corresponds to streamType=1&id=1142 in the API.

Maybe the `id` format should be `2_841`, `1_1142`?

If the live id gives 404 (no longer live), the page tries `https://stavanger.kommunetv.no/api/search/live/{live_id}/related` which, if it succeeds, gives the archive id at `[data][0]['id']`.

youtube_dl/extractor/kommunetv.py
dirkf: The MD5 values may not be critical, especially if the site is sending adaptive formats where the value could vary. Anyhow, just draft a test case with `url`, `id`, `ext`; the test harness will tell you what other values must be included. For optional items like `md5`, include a dummy value and the test harness will show the correct value.

youtube_dl/extractor/kommunetv.py
```diff
@@ -6,7 +6,8 @@
 
 
 class KommunetvIE(InfoExtractor):
-    _VALID_URL = r'https://(\w+).kommunetv.no/archive/(?P<id>\w+)'
+    _VALID_URL = r'https?://\w+\.kommun(?:etv\.no|\.tv)/(?:archive|live)/(?P<id>\w+)'
```
dirkf, line 9:
```suggestion
    _VALID_URL = r'(?P<where>https?://\w+\.kommun(?:etv\.no|\.tv))/(?P<what>archive|live)/(?P<id>\w+)'
```
Then use the `where` group in the API URL (maybe Stavanger.kommunetv.no/api has diferent data from Oslo.kommunetv.no/api ?). Use the `what` to generate the correct `streamType` for the API, using `._download_json_handle(..., expected_status=404)` for the live case and then chasing down the archive content if the live is gone.

PR #32957, opened 2024-10-22
Update README.md

AdrikaPanwar:
Hi, I found that in you readme there is no instructions to install yt-dlp using snap so I add it, I will be happy if you will review it and merge it

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,11 @@ Or with [MacPorts](https://www.macports.org/):
 
     sudo port install youtube-dl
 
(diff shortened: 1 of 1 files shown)


seproDev:
yt-dlp and youtube-dl are different projects

AdrikaPanwar:
Thanks for mentioning it, as I was totally confused because I saw that the yt-dlp repository was forked from there. So, I raised the pull request there. I will mention this on the yt-dlp repository. Thanks once again.

Issue #5704, opened 2015-05-15
Can't Download Video from CoreAnimator.com


aaronvegh:
The page for this site includes an apparently straight-forward Youtube embed, but youtube-dl says this is an unsupported URL. Here's the verbose output from the latest version of youtube-dl.

$ youtube-dl -g --verbose 'http://www.coreanimator.com/'
[debug] System config: []
[debug] User config: []
[debug] Command-line args: [u'-g', u'--verbose', u'http://www.coreanimator.com/']
[debug] Encodings: locale UTF-8, fs utf-8, out UTF-8, pref UTF-8
[debug] youtube-dl version 2015.05.10
[debug] Python version 2.7.3 - Darwin-14.3.0-x86_64-i386-64bit
[debug] exe versions: ffmpeg 2.5, ffprobe 2.5
[debug] Proxy map: {}
WARNING: Falling back on generic information extractor.
ERROR: Unsupported URL: http://www.coreanimator.com/
Traceback (most recent call last):
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/generic.py", line 937, in _real_extract
    doc = parse_xml(webpage)
  File "/usr/local/bin/youtube-dl/youtube_dl/utils.py", line 1558, in parse_xml
    tree = xml.etree.ElementTree.XML(s.encode('utf-8'), **kwargs)
  File "/usr/local/Cellar/python/2.7.3/Frameworks/Python.framework/Versions/2.7/lib/python2.7/xml/etree/ElementTree.py", line 1301, in XML
    parser.feed(text)
  File "/usr/local/Cellar/python/2.7.3/Frameworks/Python.framework/Versions/2.7/lib/python2.7/xml/etree/ElementTree.py", line 1643, in feed
    self._raiseerror(v)
  File "/usr/local/Cellar/python/2.7.3/Frameworks/Python.framework/Versions/2.7/lib/python2.7/xml/etree/ElementTree.py", line 1507, in _raiseerror
    raise err
ParseError: syntax error: line 1, column 0
Traceback (most recent call last):
  File "/usr/local/bin/youtube-dl/youtube_dl/YoutubeDL.py", line 650, in extract_info
    ie_result = ie.extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/common.py", line 273, in extract
    return self._real_extract(url)
  File "/usr/local/bin/youtube-dl/youtube_dl/extractor/generic.py", line 1467, in _real_extract
[...]

PR #32964, opened 2024-11-01
Update generic.py Improve error message for unsupported URLs

houssamallali:
Mention that this PR addresses issue [#5704](https://github.com/ytdl-org/youtube-dl/issues/5704) by enhancing the error message for unsupported URLs to make it more informative.


dirkf:
Thanks for your suggestion, but it seems to me that the proposed new text is just saying what the existing code does in more words?

Other readers may care to comment ...

yinshiyi:
agree with @dirkf .
Thanks for the effort! This might not have enough changes to warrant a PR—let's revisit if more updates are needed.

PR #33051, opened 2025-02-07
[Eporner] Check if video was deleted

Holek:

---

- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

This change gives user a friendly error message, instead of a stacktrace. I couldn't find examples for tests of deleted videos. All of them I found had `skip_video` flag set, so that's not what I wanted.

If you point me to an example that has this test, I'll gladly add it to the test matrix for this site.

### DIFF
--- a/youtube_dl/extractor/eporner.py
+++ b/youtube_dl/extractor/eporner.py
@@ -59,6 +59,11 @@ class EpornerIE(InfoExtractor):
 
         video_id = self._match_id(urlh.geturl())
 
+        deleted = self._html_search_regex(r'<strong>Video has been deleted</strong>', webpage, 'deleted', default=False, fatal=False, group=0)
+        if deleted:
+            raise ExtractorError(
+                'Video %s has been deleted' % video_id, expected=True)
+
         hash = self._search_regex(
             r'hash\s*[:=]\s*["\']([\da-f]{32})', webpage, 'hash')
 


youtube_dl/extractor/eporner.py
```diff
@@ -59,6 +59,11 @@ def _real_extract(self, url):
 
         video_id = self._match_id(urlh.geturl())
 
+        deleted = self._html_search_regex(r'<strong>Video has been deleted</strong>', webpage, 'deleted', default=False, fatal=False, group=0)
+        if deleted:
+            raise ExtractorError(
+                'Video %s has been deleted' % video_id, expected=True)
```
dirkf, line 65: Partly a matter of taste, but this arrangement could be extended to other error cases by adding to the `tuple` of error message patterns. It's better then to use the default search for group 1 -- `default=None` is all that's needed to silence the "not found" case; and the pattern can be relaxed in case the HTML gets split up (I didn't make each space a `\s+` which might be logical):
```suggestion
        missing = self._html_search_regex((
            r'<strong>\s*(Video has been deleted)\s*</strong>',
        ), webpage, 'missing', default=None)
        if missing:
            raise ExtractorError(
                'Video %s is not available: "%s"' % (video_id, missing), expected=True)
```

Holek: Makes sense to me.

PR #33056, opened 2025-02-13
[Edited] Fix minor bug in the main function

MayureshMore:
Changes made in branch: **MayureshMore:master**
[Edited] Fix minor bug in the main function


dirkf:
Use the template. Then you would have to explain what the PR offers other than removing the required 1st line and breaking the layout (excuse me if I overlooked an **actual** change that should be considered).

PR #33068, opened 2025-02-21
- Implemented _parse_date method to convert dateTimeString to a UNIX …

Asry18:
This PR adds functionality to parse the dateTimeString from events and return a valid timestamp for each event. The following changes have been made:

Implemented _parse_date method: This method parses the dateTimeString (in ISO 8601 format) and converts it into a UNIX timestamp. The timestamp is also adjusted to UTC before being returned.
Modified _real_extract method: The method now includes the parsed timestamp field in the returned event data.
This enhancement ensures that each event's timestamp is accurately parsed, providing consistent date/time information for further processing.

Additional Information:
The dateTimeString field is expected to be in ISO 8601 format (e.g., 2022-01-01T12:00:00-05:00).
The timestamp is now returned as part of the event data, which can be used for further processing or display.
This PR improves the extraction process by adding proper timestamp handling.

### DIFF
--- a/youtube_dl/extractor/stretchinternet.py
+++ b/youtube_dl/extractor/stretchinternet.py
@@ -1,10 +1,11 @@
 from __future__ import unicode_literals
+
 from .common import InfoExtractor
-from datetime import datetime, timezone
+from datetime import datetime
 try:
-    from zoneinfo import ZoneInfo  # Python 3.9+
+    from zoneinfo import ZoneInfo as compat_ZoneInfo  # Python 3.9+
 except ImportError:
-    from ..compat import compat_zoneinfo as ZoneInfo  # Fallback for older versions
+    compat_ZoneInfo = None  # Fallback: Do not handle alphabetic time zones
 
 from ..utils import unified_timestamp, parse_iso8601
 
@@ -13,6 +14,7 @@ class StretchInternetIE(InfoExtractor):
 
     def _real_extract(self, url):
         video_id = self._match_id(url)
+
         media_url = self._download_json(
             'https://core.stretchlive.com/trinity/event/tcg/' + video_id,
             video_id)[0]['media'][0]['url']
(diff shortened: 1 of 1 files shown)


dirkf:
_pytz_ cannot be included in the project because it's MIT licensed. Therefore a fallback would be needed, which would make _pytz_ irrelevant anyway.

Review `unified_timstamp()` and `parse_iso8601()` in utils.py which should address the case of numeric TZ (but perhaps both don't). If you find test-cases (try the ones from _yt-dlp_ first) that fail, please raise an issue with the details. Also, please raise an issue for the extractor problem that this PR was intended to address.

For alphabetic TZ we should compat the _zoneinfo_ module from Py3.9+, so that utils.py can say `from ..compat import compat_zoneinfo as zoneinfo` or similar.

Asry18:
I’ve updated the implementation to remove pytz and use zoneinfo (for Python 3.9+) with a compat_zoneinfo fallback for older versions.
Additionally, I’ve integrated unified_timestamp() and parse_iso8601() from utils.py to handle timestamp parsing where applicable. If these fail, the extractor falls back to manual ISO 8601 parsing while ensuring UTC conversion.

dirkf:
One approach is just to fall back to not handling alphabetic TZ if _zoneinfo_ is unavailable.

A better but harder one is to borrow and convert the _zoneinfo_ from Py3.9+ into a `youtube_dl.zoneinfo` module. If we only need the `ZoneInfo` class it can then be loaded in compat.py:
```py
try:
    from zoneinfo import ZoneInfo as compat_ZoneInfo
except ImportError:
    from .zoneinfo import _ZoneInfo as compat_ZoneInfo
```
Or the passthrough module mechanism from _yt-dlp_ could be used.
 
The _ZoneInfo_ class uses the system TZ data, which is less likely to succeed in older systems, or the TZ data from the PyPi _tzdata_ package.

The first problem here is that the _tzdata_ package is under the Apache Licence, which means we can't distribute it either. Users would have to install it themselves.

The second problem is that, although the package is supposedly compatible with Py2 (we care about 2.6 and 2.7), PyPi isn't, because it requires HTTPS features that don't exist in Py2.6, and so _pip_ access to PyPi fails. So Py2.6 users can't install with _pip_ in the normal way, although it's still possible to install from a copy of the package wheel stored locally or on a less fussy network server: perhaps it would be acceptable to proxy the wheel in a separate repo forked from the _tzdata_ repo.

Asry18:
Thanks for the feedback! I've removed pytz and replaced it with zoneinfo (Python 3.9+). If zoneinfo is unavailable, the code falls back to ignoring alphabetic time zones. Numeric time zones are still parsed correctly.

PR #33116, opened 2025-04-22
my first PR

arjunthazhath2001:

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -1196,7 +1196,8 @@ Don't capture id attribute name here since you can't use it for anything anyway.
 Correct:
 
 ```python
-r'(?:id|ID)=(?P<id>\d+)'
+r'(?i)\b id \s* = \s* ["\']?(?P<id>\d+)\b["\']?'
+
 ```
 
 Incorrect:


README.md
```diff
@@ -1196,7 +1196,8 @@ Don't capture id attribute name here since you can't use it for anything anyway.
 Correct:
 
 ```python
-r'(?:id|ID)=(?P<id>\d+)'
+r'(?i)\b id \s* = \s* ["\']?(?P<id>\d+)\b["\']?'
```
dirkf, line 1199: This pattern is an example. The context is clearly one where the target may be like `id=1234` or `ID=1234` and the teaching point is that the variable name shouldn't be captured when its value is.

Thus:
* changing the LHS of the assignment pattern destroys the teaching point, so no
* if the pattern were being used, `\b` followed by whitespace is pointless
* so is `\s*` surrounded by spaces, unless you actually wanted to match only a sequence of space, optional whitespace, space
* it would not be helpful to this example but a case where a capturing group need not be referenced in the code is when it's used for an internal match, as in matching a quoted string that may contain escaped quotes with eg `'''('|")(?P<string>(?:\\\1|(?!\1).)*\1'''`; the suggested value pattern here which doesn't do this matches, undesirably,  `123` in `'123"`, `'123 `
* the example assumes that the `id` or `ID` variable will have a digit value that is terminated as appropriate to the grammatical context (eg, optional whitespace followed by newline or `[,;}]` in JS): a production pattern might check for that, but `\b` isn't necessarily right.

README.md
```diff
@@ -1196,7 +1196,8 @@ Don't capture id attribute name here since you can't use it for anything anyway.
...
+
```
dirkf, line 1200: Unnecessary newline.

dirkf (commented):
Thanks for your work. As explained in the review comments, I won't be merging your proposal. 

Please remember in future to complete the template fully especially if the PR does not reference an issue where the change has already been discussed. Generally it's better to raise an issue before creating a PR, to identify whether there is a real problem and how it should be resolved if so.

PR #31981, opened 2023-04-01
[infomaniak] Add new extractor to resolve #31329

ishabahmed:
### I have:
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [x] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [x] Covered the code with tests (note that PRs without tests will be REJECTED)
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/).
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of my *pull request*?
- [ ] Bug fix
- [ ] Improvement
- [x] New extractor
- [ ] New feature

---

### Description of my *pull request* and other information

Added an extractor to support player.vod2.infomaniak.com as requested in #31329

PR #33234, opened 2025-12-15
[infomaniak] Add VOD2 embed extractor

### LINKED ISSUE #31329: Add support for player.vod2.infomaniak.com/
## Checklist

- [X] I'm reporting a new site support request
- [X] I've verified that I'm running youtube-dl version **2021.12.17**
- [X] I've checked that all provided URLs are alive and playable in a browser
- [X] I've checked that none of provided URLs violate any copyrights
- [X] I've searched the bugtracker for similar site support requests including closed ones

## Example URLs

- Single video: https://www.youtube.com/watch?v=BaW_jenozKc
- Single video: https://youtu.be/BaW_jenozKc
- Playlist: https://www.youtube.com/playlist?list=PL4lCao7KL_QFVb7Iudeipvc2BCavECqzc

## Description

youtube-dl -v https://player.vod2.infomaniak.com/embed/1jhvl2uqg6ywp

[debug] System config: []
[debug] User config: []
[debug] Custom config: []
[debug] Command-line args: ['-v', 'https://player.vod2.infomaniak.com/embed/1jhvl2uqg6ywp']
[debug] youtube-dl version 2021.12.17
[debug] Python version 3.8.10 (CPython) - Linux-5.4.0-131-generic-x86_64-with-glibc2.29
[debug] exe versions: ffmpeg 4.2.7, ffprobe 4.2.7, rtmpdump 2.4
[debug] Proxy map: {}
[generic] 1jhvl2uqg6ywp: Requesting header
[generic] 1jhvl2uqg6ywp: Downloading webpage
[generic] 1jhvl2uqg6ywp: Extracting information
ERROR: Unsupported URL: https://player.vod2.infomaniak.com/embed/1jhvl2uqg6ywp
  File "/usr/local/lib/python3.8/dist-packages/youtube_dl/YoutubeDL.py", line 815, in wrapper
    return func(self, *args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/youtube_dl/YoutubeDL.py", line 836, in __extract_info
    ie_result = ie.extract(url)
  File "/usr/local/lib/python3.8/dist-packages/youtube_dl/extractor/common.py", line 534, in extract
  File "/usr/local/lib/python3.8/dist-packages/youtube_dl/extractor/generic.py", line 3489, in _real_extract
    raise UnsupportedError(url)
youtube_dl.utils.UnsupportedError: Unsupported URL: https://player.vod2.infomaniak.com/embed/1jhvl2uqg6ywp

THANKS FOR YOUR WORK

hwang-cadent:
Fixes #31329

Add support for Infomaniak VOD2 embed URLs (player.vod2.infomaniak.com/embed/<id>) by introducing a new InfomaniakVod2IE extractor that reads the public share JSON and extracts HLS + direct MP4 sources. Fixes #31329 (example: https://player.vod2.infomaniak.com/embed/1jhvl2uqg6ywp).

Tests
python -m youtube_dl -v --no-check-certificate --skip-download https://player.vod2.infomaniak.com/embed/1jhvl2uqg6ywp
python -m youtube_dl --no-check-certificate --skip-download -F https://player.vod2.infomaniak.com/embed/1jhvl2uqg6ywp
export SSL_CERT_FILE="$(python -c 'import certifi; print(certifi.where())')"
python test/test_download.py TestDownload.test_InfomaniakVod2

Supersedes, closes #31981.

### DIFF
--- a/youtube_dl/extractor/infomaniak.py
+++ b/youtube_dl/extractor/infomaniak.py
@@ -78,7 +78,13 @@ class InfomaniakVod2IE(InfoExtractor):
                 'display_id': share_id,
             }, entries[0], rev=True)
 
+        missing_id = sum(1 for e in entries if not e.get('id'))
         entries = [e for e in entries if e.get('id')]
+        if missing_id:
+            self.report_warning(
+                'Share contains multiple media entries, but %d item(s) are missing an id and will be skipped. '
+                'If this looks incorrect, please report the URL.' % missing_id,
+                share_id, only_once=True)
         if not entries:
             # Keep the error message actionable for site support requests
             raise ExtractorError('Unable to find any media in share JSON', expected=True)
--- a/youtube_dl/extractor/infomaniak.py
+++ b/youtube_dl/extractor/infomaniak.py
@@ -1,0 +1,88 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    parse_duration,
+)
+
+
+class InfomaniakVod2IE(InfoExtractor):
+    IE_NAME = 'infomaniak:vod2'
+    _VALID_URL = r'https?://player\.vod2\.infomaniak\.com/embed/(?P<id>[0-9a-z]+)'
+    _TEST = {
+        'url': 'https://player.vod2.infomaniak.com/embed/1jhvl2uqg6ywp',
+        'info_dict': {
+            'id': '1jhvl2uqg6xis',
+            'display_id': '1jhvl2uqg6ywp',
+            'ext': 'mp4',
+            'title': 'Conférence à Dyo, octobre 2022',
+            'thumbnail': r're:^https?://.*\.(?:jpe?g|png)$',
+        },
+        'params': {
+            'skip_download': True,
+        },
+    }
+
+    def _real_extract(self, url):
+        share_id = self._match_id(url)
+        share = self._download_json(
+            'https://api.infomaniak.com/2/vod/res/shares/%s.json' % share_id,
+            share_id)
+        data = (share or {}).get('data') or {}
+        media = data.get('media') or []
+
+        entries = []
+        for m in media:
+            if not isinstance(m, dict):
+                continue
+            media_id = m.get('id') or share_id
+            title = m.get('title') or share_id
+            duration = parse_duration(m.get('duration'))
+            thumbnails = m.get('thumbnails') or {}
+            thumbnail = thumbnails.get('poster') or thumbnails.get('image')
+
+            source = m.get('source') or {}
+            formats = []
+
+            hls_url = source.get('hls')
+            if hls_url:
+                formats.extend(self._extract_m3u8_formats(
+                    hls_url, media_id, 'mp4', entry_protocol='m3u8_native',
+                    m3u8_id='hls', fatal=False))
+
+            mpd_url = source.get('dash')
+            if mpd_url:
+                formats.extend(self._extract_mpd_formats(
+                    mpd_url, media_id, mpd_id='dash', fatal=False))
+
+            best_url = source.get('best')
+            if best_url:
+                formats.append({
+                    'url': best_url,
+                    'format_id': 'http',
+                })
+
+            self._sort_formats(formats)
+
+            entries.append({
+                'id': media_id,
+                'display_id': share_id,
+                'title': title,
+                'duration': duration,
+                'thumbnail': thumbnail,
+                'formats': formats,
+                'webpage_url': url,
+            })
+
+        if not entries:
+            # Keep the error message actionable for site support requests
+            raise ExtractorError('Unable to find any media in share JSON', expected=True)
+
+        if len(entries) == 1:
+            return entries[0]
+
+        return self.playlist_result(entries, playlist_id=share_id)
+
+
--- a/youtube_dl/extractor/infomaniak.py
+++ b/youtube_dl/extractor/infomaniak.py
@@ -1,10 +1,17 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
+from ..compat import compat_str as str
+
 from .common import InfoExtractor
 from ..utils import (
     ExtractorError,
+    merge_dicts,
     parse_duration,
+    str_or_none,
+    T,
+    traverse_obj,
+    url_or_none,
 )
 
 
@@ -13,15 +20,16 @@ class InfomaniakVod2IE(InfoExtractor):
     _VALID_URL = r'https?://player\.vod2\.infomaniak\.com/embed/(?P<id>[0-9a-z]+)'
     _TEST = {
         'url': 'https://player.vod2.infomaniak.com/embed/1jhvl2uqg6ywp',
+        'md5': '08c3a89906b70a614fa0fdb057e8a22e',
         'info_dict': {
             'id': '1jhvl2uqg6xis',
             'display_id': '1jhvl2uqg6ywp',
             'ext': 'mp4',
             'title': 'Conférence à Dyo, octobre 2022',
-            'thumbnail': r're:^https?://.*\.(?:jpe?g|png)$',
+            'thumbnail': r're:https?://.+\.(?:jpe?g|png)$',
         },
         'params': {
-            'skip_download': True,
+            'format': 'http',
         },
     }
 
@@ -30,20 +38,11 @@ class InfomaniakVod2IE(InfoExtractor):
         share = self._download_json(
             'https://api.infomaniak.com/2/vod/res/shares/%s.json' % share_id,
             share_id)
-        data = (share or {}).get('data') or {}
-        media = data.get('media') or []
-
         entries = []
-        for m in media:
-            if not isinstance(m, dict):
-                continue
-            media_id = m.get('id') or share_id
-            title = m.get('title') or share_id
-            duration = parse_duration(m.get('duration'))
-            thumbnails = m.get('thumbnails') or {}
-            thumbnail = thumbnails.get('poster') or thumbnails.get('image')
-
-            source = m.get('source') or {}
+        for media in traverse_obj(share, (
+                'data', 'media', lambda _, v: v.get('source').get)):
+            media_id = media.get('id') or share_id
+            source = media['source']
             formats = []
 
             hls_url = source.get('hls')
@@ -62,26 +61,32 @@ class InfomaniakVod2IE(InfoExtractor):
                 formats.append({
                     'url': best_url,
                     'format_id': 'http',
+                    'preference': 10,
                 })
 
             self._sort_formats(formats)
 
-            entries.append({
-                'id': media_id,
-                'display_id': share_id,
-                'title': title,
-                'duration': duration,
-                'thumbnail': thumbnail,
-                'formats': formats,
-                'webpage_url': url,
-            })
+            entries.append(merge_dicts(
+                traverse_obj(media, {
+                    'id': ('id', T(str_or_none)),
+                    'title': ('title', T(str)),
+                    'duration': ('duration', T(parse_duration)),
+                    'thumbnail': ('thumbnails', ('poster', 'image'), T(url_or_none), any),
+                }), {
+                    'display_id': share_id,
+                    'formats': formats,
+                }))
 
+        if len(entries) == 1:
+            return merge_dicts({
+                'id': share_id,
+                'display_id': share_id,
+            }, entries[0], rev=True)
+
+        entries = [e for e in entries if e.get('id')]
         if not entries:
             # Keep the error message actionable for site support requests
             raise ExtractorError('Unable to find any media in share JSON', expected=True)
-
-        if len(entries) == 1:
-            return entries[0]
 
         return self.playlist_result(entries, playlist_id=share_id)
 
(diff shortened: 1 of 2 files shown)


hwang-cadent:
Hey @dirkf, please take a look at this PR if you get chances. Thanks!

youtube_dl/extractor/infomaniak.py
```diff
@@ -0,0 +1,88 @@
...
+    IE_NAME = 'infomaniak:vod2'
+    _VALID_URL = r'https?://player\.vod2\.infomaniak\.com/embed/(?P<id>[0-9a-z]+)'
+    _TEST = {
+        'url': 'https://player.vod2.infomaniak.com/embed/1jhvl2uqg6ywp',
+        'info_dict': {
+            'id': '1jhvl2uqg6xis',
+            'display_id': '1jhvl2uqg6ywp',
+            'ext': 'mp4',
+            'title': 'Conférence à Dyo, octobre 2022',
+            'thumbnail': r're:^https?://.*\.(?:jpe?g|png)$',
```
dirkf: `re:` already implies a match anchored at the start, and the path extension surely can't be immediately after the `//`.
```suggestion
            'thumbnail': r're:https?://.+\.(?:jpe?g|png)$',
```

youtube_dl/extractor/infomaniak.py
```diff
@@ -0,0 +1,88 @@
...
+        },
+        'params': {
+            'skip_download': True,
+        },
```
dirkf: There's a non-adaptive format, so we can have file download check (the test suite just gets an initial chunk, so the test won't be fetching the whole 4GB). If the `best`/`http` format isn't automatically the best (but it should be, I suggest: see below), you could put `'format': 'http',` in `params`.

```suggestion
        'md5': '08c3a89906b70a614fa0fdb057e8a22e',
        'info_dict': {
            'id': '1jhvl2uqg6xis',
            'display_id': '1jhvl2uqg6ywp',
            'ext': 'mp4',
            'title': 'Conférence à Dyo, octobre 2022',
            'thumbnail': r're:^https?://.*\.(?:jpe?g|png)$',
        },
```

youtube_dl/extractor/infomaniak.py
```diff
@@ -0,0 +1,88 @@
...
+        share = self._download_json(
+            'https://api.infomaniak.com/2/vod/res/shares/%s.json' % share_id,
+            share_id)
+        data = (share or {}).get('data') or {}
+        media = data.get('media') or []
+
+        entries = []
+        for m in media:
+            if not isinstance(m, dict):
+                continue
```
dirkf: More robust and concise, since `data` and `media` aren't used elsewhere:
```suggestion

        entries = []
        for m in traverse_obj(share, (
                'data', 'media', lambda _, v: v.get('source'))):
```

dirkf: More robust and concise, since `data` and `media` aren't being used elsewhere,
```suggestion
        entries = []
        for m in traverse_obj(share, (
                'data', 'media', lambda _, v: v.get('source').get)):
            media_id = m.get('id') or share_id
```
Then `m`, not a great variable name, could be renamed as `media`.

youtube_dl/extractor/infomaniak.py
```diff
@@ -0,0 +1,88 @@
...
+            media_id = m.get('id') or share_id
+            title = m.get('title') or share_id
+            duration = parse_duration(m.get('duration'))
+            thumbnails = m.get('thumbnails') or {}
+            thumbnail = thumbnails.get('poster') or thumbnails.get('image')
+
+            source = m.get('source') or {}
```
dirkf: 1. Either assume that `title` will be present, or specialise IE._generic_title() to make s/t like `Video {m['id']} from vod2.infomaniak.com`. The ID alone adds nothing, since it's not meaningful (not like `conference-a-dyo`, say).
2. Metadata extraction can be done in the return expression.
3. In the previous suggestion the list was filtered for items with a `source` `dict`.
```suggestion
            source = m['source']
```

youtube_dl/extractor/infomaniak.py
```diff
@@ -0,0 +1,88 @@
...
+            mpd_url = source.get('dash')
+            if mpd_url:
+                formats.extend(self._extract_mpd_formats(
+                    mpd_url, media_id, mpd_id='dash', fatal=False))
+
+            best_url = source.get('best')
+            if best_url:
+                formats.append({
+                    'url': best_url,
+                    'format_id': 'http',
```
dirkf, line 59: If this format is always the same as the best m3u8 format, it should be preferred:
```suggestion
                    'format_id': 'http',
                    'preference': 10,
```

youtube_dl/extractor/infomaniak.py
```diff
@@ -0,0 +1,88 @@
...
+
+            entries.append({
+                'id': media_id,
+                'display_id': share_id,
+                'title': title,
+                'duration': duration,
+                'thumbnail': thumbnail,
+                'formats': formats,
+                'webpage_url': url,
+            })
```
dirkf: 1. If it's possible for `id` to be absent, `share_id` could be used when there's only one item (as in these suggestions), or an index number could be appended (`{share_id}_{idx}`). Otherwise append `T(require('id'))` to the traversal tuple for `id`.
2. Similarly if `title` must always be present add `T(require('title'))` there. 
```suggestion
            entries.append(merge_dicts({
                'formats': formats,
            }, traverse_obj(m, {
                'id': ('id', T(str_or_none)),
                'title': ('title', T(str)),
                'duration': ('duration', T(parse_duration)),
                'thumbnail': ('thumbnails', ('poster', 'image'), T(url_or_none), any),
            })))
```

youtube_dl/extractor/infomaniak.py
```diff
@@ -0,0 +1,88 @@
...
+
+        if not entries:
+            # Keep the error message actionable for site support requests
+            raise ExtractorError('Unable to find any media in share JSON', expected=True)
+
+        if len(entries) == 1:
+            return entries[0]
```
dirkf: This implements using `share_id` as a default `id` if there's only one entry, and stripping out multiple entries with no `id`:
```suggestion
        if len(entries) == 1:
            return merge_dicts({
                'id': share_id,
                'display_id': share_id,
            }, entries[0], rev=True)

        entries = [e for e in entries if e.get('id')]
        if not entries:
            # Keep the error message actionable for site support requests
            raise ExtractorError('Unable to find any media in share JSON', expected=True)
```

youtube_dl/extractor/infomaniak.py
```diff
@@ -0,0 +1,88 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from .common import InfoExtractor
+from ..utils import (
+    ExtractorError,
+    parse_duration,
```
dirkf, line 9: Other suggestions require these imports: 
```suggestion
from ..compat import compat_str as str
from ..utils import (
    ExtractorError,
    merge_dicts,
    parse_duration,
    str_or_none,
    T,
    traverse_obj,
    url_or_none,
```

dirkf (changes requested):
Thanks for your work!

This site is even more tractable withe new (?) API. I've made a few suggestions based on my edited version that passes the d/l test and with the problem URL from the linked issue. Generally, it would be good to have some more test-cases, especially with multiple entries.

youtube_dl/extractor/infomaniak.py
hwang-cadent: Thanks! Updated the thumbnail regex to drop the redundant ^ and make it more realistic (no immediate // → extension). Now using: r're:https?://.+\.(?:jpe?g|png)$'.

youtube_dl/extractor/infomaniak.py
hwang-cadent: Good point. I added an md5 to _TEST and switched the test params to prefer the direct file ('format': 'http') so the test suite can do the lightweight chunk download without pulling the full video.

youtube_dl/extractor/infomaniak.py
hwang-cadent: Agreed. this is more cleaner. I refactored the parsing to use traverse_obj(...) for iterating media entries and renamed the loop variable to media for readability.

youtube_dl/extractor/infomaniak.py
hwang-cadent: Makes sense. I now treat source as required (the traversal filters for entries with a source dict) and simplified the code accordingly. Metadata extraction is also consolidated rather than being scattered across temporary variables.

youtube_dl/extractor/infomaniak.py
hwang-cadent: Yep. added preference: 10 to the http format so it wins over the HLS variants when both are present.

youtube_dl/extractor/infomaniak.py
hwang-cadent: Thanks for the suggested pattern. I adopted the merge_dicts(traverse_obj(...), {...}) approach for id/title/duration/thumbnail and used url_or_none/parse_duration via T(...) conversions.

youtube_dl/extractor/infomaniak.py
hwang-cadent: Good catch. For the single-entry case, I now default to id=share_id if needed, and for multi-entry playlists I filter out entries missing id before returning.

hwang-cadent:
> Thanks for your work!
> 
> This site is even more tractable withe new (?) API. I've made a few suggestions based on my edited version that passes the d/l test and with the problem URL from the linked issue. Generally, it would be good to have some more test-cases, especially with multiple entries.



Thanks!  I agree that adding a multi-entry (data.media length ≥ 2) test would be ideal.
I tried several approaches to locate a stable multi-entry share/embed ID, but so far all public samples I’ve found return exactly one media item:
Probed known Infomaniak share JSONs I already had (e.g. the original example and a second public share) — both had data.media length 1.
Tried finding embed/share IDs from a public catalog:
I inspected https://carac.tv/replay and found it loads replay items via https://api.m0g.ch/v1/replays/<page>/<page_size>.
I extracted ~480 unique player.vod2.infomaniak.com/embed/<id> values from that API and checked each corresponding https://api.infomaniak.com/2/vod/res/shares/<id>.json.
Result: no multi-entry shares found (all had data.media length 1).
Tried to discover IDs via page JS bundles (Nuxt chunks) and direct string searches; no multi-entry share IDs surfaced.
If you or someone can point me to (or confirm) a public embed/share URL where the share JSON contains multiple objects in data.media, I’m happy to add it as a _TESTS playlist case (e.g. with playlist_mincount: 2).

youtube_dl/extractor/infomaniak.py
```diff
@@ -0,0 +1,93 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+from ..compat import compat_str as str
+
+from .common import InfoExtractor
```
dirkf: Conventional order:
```suggestion
from .common import InfoExtractor
from ..compat import compat_str as str
```

youtube_dl/extractor/infomaniak.py
```diff
@@ -0,0 +1,93 @@
...
+            self._sort_formats(formats)
+
+            entries.append(merge_dicts(
+                traverse_obj(media, {
+                    'id': ('id', T(str_or_none)),
+                    'title': ('title', T(str)),
+                    'duration': ('duration', T(parse_duration)),
+                    'thumbnail': ('thumbnails', ('poster', 'image'), T(url_or_none), any),
+                }), {
+                    'display_id': share_id,
```
dirkf: Not needed because it's added if needed in l.83: 
```suggestion
```

youtube_dl/extractor/infomaniak.py
```diff
@@ -0,0 +1,93 @@
...
+        'md5': '08c3a89906b70a614fa0fdb057e8a22e',
+        'info_dict': {
+            'id': '1jhvl2uqg6xis',
+            'display_id': '1jhvl2uqg6ywp',
+            'ext': 'mp4',
+            'title': 'Conférence à Dyo, octobre 2022',
+            'thumbnail': r're:https?://.+\.(?:jpe?g|png)$',
+        },
+        'params': {
+            'format': 'http',
```
dirkf: Not strictly necessary now because of l.64.

dirkf (changes requested):
Maybe the answer for the multiple entry case is to generate a warning or even error  in that case and hope that the user will report it?

youtube_dl/extractor/infomaniak.py
hwang-cadent: Thanks. updated the import order to match the project’s convention

youtube_dl/extractor/infomaniak.py
hwang-cadent: Good call.  Removed it. Since the direct MP4 format now has preference: 10, it’s selected automatically, so forcing format=http is unnecessary.

youtube_dl/extractor/infomaniak.py
hwang-cadent: Agreed. Dropped display_id from individual entries. The single-entry return path already sets display_id appropriately.

hwang-cadent:
> Maybe the answer for the multiple entry case is to generate a warning or even error in that case and hope that the user will report it?

Thanks . I went with the warning approach for the multi-entry edge case. For multi-entry shares, if any items are missing an id (so they’d be skipped), the extractor now emits a one-time report_warning noting how many entries were skipped and asking the user to report the URL if it looks incorrect. I avoided hard-failing with an error since the remaining entries can still be usable, but this should help surface real multi-entry cases for follow-up/testing.

PR #33258, opened 2026-03-19
[extractor] Prioritize YouTube extractors to reduce cold dispatch latency

AsymptotaX:
### What is the purpose
- [ ] Bug fix
- [x ] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description 

This improves performance for the most common extractor case (YouTube).

Extractor dispatch in youtube_dl is order-sensitive in the non-lazy path.

Previously, YouTube extractors (YoutubeIE and related classes) were located
near the end of the extractor list, causing YouTube URLs to traverse a large
number of unrelated extractors before matching.

This change moves the YouTube extractor family closer to the top of the
dispatch order, reducing cold-start latency for one of the most common
URL types.

No behavior changes: matching semantics and extractor logic remain unchanged.

### Benchmark
Cold dispatch benchmark comparing the legacy YouTube placement against the new order:

| url | before | after | Δ ms | idx_b→idx_a |
|---|---:|---:|---:|---:|
| youtube/watch | 81.3 | 2.5 | +78.8 | 1233→1 |
| youtube/playlist | 76.5 | 0.3 | +76.1 | 1236→4 |
| youtu.be | 78.4 | 2.5 | +75.9 | 1233→1 |
| vimeo | 68.7 | 73.4 | -4.7 | 1112→1127 |
| soundcloud | 54.5 | 57.3 | -2.8 | 870→885 |
| bbc | 4.7 | 7.9 | -3.2 | 75→90 |

On this sample:
• YouTube URLs improve by about 76-79 ms on the cold path.
• Later non-YouTube matches regress slightly, by about 3-5 ms in the sampled cases.


dirkf:
_yt-dlp_ does a similar thing. It could be better (instead or also) to move a matched IE to the top of the extractor list, so that the next search in that invocation of _yt-dl_ will be fast if the same site is being hit. This could help with playlists and batches.

bashonly:
> *yt-dlp* does a similar thing.

@dirkf yt-dlp *used* to do this, until it was determined that it didn't make any difference because of how yt-dlp looks up its extractors and a different optimization was needed: https://github.com/yt-dlp/yt-dlp/commit/e2ea6bd6ab639f910b99e55add18856974ff4c3a

dirkf:
Isn't a better approach to have a class attribute like `_IE_PRIO = 1000` with getter `ie_prio()` and sort the extractors by that value when populating `extractor.__ALL_CLASSES`. Then we can, eg, promote all YT extractors by setting the base class value to 2000, or promote just individual classes, maybe not `YoutubeTruncatedURLIE`, or demote all unlisted IEs, or ..., but only by changing the IE module itself.

Maintainers must act against any [arms race](https://devblogs.microsoft.com/oldnewthing/20110310-00/?p=11253).

PR #33262, opened 2026-03-30
fix: pin 1 unpinned action

dagecko:
Re-submission of #33260. Had a problem with my fork and had to delete it, which closed the original PR. Apologies for the noise.

## Summary

This PR pins a GitHub Action to an immutable commit SHA instead of a mutable version tag.

## How to verify

Review the diff, the change is mechanical and preserves workflow behavior:
- **SHA pinning**: `action@v3` becomes `action@abc123 # v3`, original version preserved as comment
- No workflow logic, triggers, or permissions are modified

I've been researching CI/CD supply chain attack vectors and submitting fixes to affected repos. Based on that research I built a scanner called Runner Guard and open sourced it [here](https://github.com/Vigilant-LLC/runner-guard) so you can scan yourself if you want to. I'll be posting more advisories over the next few weeks [on Twitter](https://x.com/vigilance_one) if you want to stay in the loop.

If you have any questions, reach out. I'll be monitoring comms.

\- Chris (dagecko)


dirkf:
Thanks for your work.

SHA pinning may be a good practice when using external actions, but this action belongs to our GH org/repo.

Also, SHA pinning defeats semantic versioning. Asking for v1 should automatically get the latest v1. If the action no longer fulfils the interface of v1, it should be numbered v2.x, and then the action invocation needs to be reviewed.

bashonly:
> Asking for v1 should automatically get the latest v1. If the action no longer fulfils the interface of v1, it should be numbered v2.x, and then the action invocation needs to be reviewed.

This is a bug, not a feature, IMO. It allows for things like the [recent trivy attack](https://github.com/aquasecurity/trivy/security/advisories/GHSA-69fq-xp46-6x23).

dirkf:
With SHA pinning, to ensure the integrity of your product, you ought to QA all your external actions (etc) and their dependencies. And then do that again if you change any pinned version, in case a pinned dependency was changed by someone less careful (or worse). Do people really do that, even should tool support exist? 

SHA pinning makes this explicit. Or, as here, you can fork the action(s) and internalise the problem.

It's a fair point that semantic versioning for GH Actions, even when/if respected by action devs, is vulnerable, because the ecosystem lacks the "package management" layer that exists for, eg, Debian.

Generally, "Reflections on trusting trust".

dagecko:
@dirkf thanks for the thoughtful questions. Let me address each one.

On this being an internal action: you are right that the trust profile is different when you control the action. But the attack does not require someone to compromise your repo. It requires compromising a maintainer account on the upstream action's repo. That is a different org, different credentials. If that account gets compromised and malicious code gets pushed to the tag, every downstream workflow that references it by tag pulls the poisoned version automatically. That is exactly what happened with tj-actions last month.

On SHA pinning defeating semver: I understand the concern. Semver is convenient. But as @bashonly pointed out, mutable tags are exactly the mechanism attackers are exploiting. The Trivy attack worked because the attacker force-pushed 76 of 77 version tags to malicious commits. Every organization that trusted those tags ran credential-stealing malware. Cisco lost over 300 source code repositories as a direct result, including customer code from banks and US government agencies. Semver only works when you can trust the person holding the tag. That trust is being broken repeatedly right now.

On the QA burden: fair question. Do people really QA every pinned version? In practice, tools like Dependabot and Renovate handle this. They auto-update SHA pins when new versions are released and surface them as reviewable PRs. From our data across 28 merged repos, teams are running Dependabot on weekly or monthly schedules and none have reported it being a burden. The alternative is trusting that every upstream tag is safe forever, which is the assumption that keeps getting exploited.

The reason I started doing this a few weeks ago is that I watched this supply chain campaign escalate from small CI tools to security scanners to AI infrastructure to Axios last night, 100 million weekly downloads compromised with a full remote access trojan attributed to North Korean hackers by Google. I have mapped a 7 phase escalation pattern and we just entered Phase 4. I have been trying to get fixes in place across the ecosystem before the next phase hits.

I have opened 90+ of these PRs and 28 have been merged by projects including Next.js, Apache Superset, Microsoft Terminal, Keras, Svelte, and others. If you do not want to move forward with this one, just close it and no hard feelings.

Full research: https://www.vigilantdefense.com/research/github-top-50k-repos-cicd-security-scan

- Chris (dagecko)

PR #32725, opened 2024-02-21
[XFileShare] Update for 2024 with new InfoExtractor features 

dirkf:
<details><summary>Boilerplate: improvement/own code</summary>

## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [x] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [x] Covered the code with tests (note that PRs without tests will be REJECTED)
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/), except for fragments from yt-dlp where this or the below have already been asserted 
- [ ] I am not the original author of this code but it is in public domain or released under [Unlicense](http://unlicense.org/) (provide reliable evidence)

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

</details>

### Description of your *pull request* and other information

This PR updates the XFileShare extractor module in three steps:
1. adds a `_search_json()` method to `InfoExtractor` matching the yt-dlp method; it uses the error diagnostic to truncate extra text from the JSON string, which might be confused by a Python `json` library implementation that uses a different diagnostic template;

2. updates the common JWPlayer extraction to use `traverse_obj()` and `_search_json()`, to support extracting playlists in a `.load({**video1},{**video2}, ...)` fragment, and to support a `transform_source` parameter in `_extract_jwplayer_data()`;

3. updates the `XFileShareIE` extractor to make use of these features.

As a result, filemoon.sx can be supported as one of the known sites and its separate extractor is removed: resolves #32716.

Supersedes, closes #32452.

PR #33266, opened 2026-04-12
[darkibox] Add extractor

appfolder:
## Summary
- Add extractor for [darkibox.com](https://darkibox.com/), an XFileSharing-based video hosting platform

## How it works
1. Fetches embed page for title
2. POSTs to `/dl` with `op=embed&file_code=ID&auto=1`
3. Unpacks Dean Edwards packed JavaScript via `decode_packed_codes`
4. Extracts video URL from PlayerJS `file:` parameter
5. Handles HLS (m3u8), multi-quality MP4 (`[label]url`), and direct URLs

## Supported URLs
- `https://darkibox.com/FILECODE`
- `https://darkibox.com/d/FILECODE`
- `https://darkibox.com/embed-FILECODE.html`

## Test plan
- [x] Import verified
- [x] All URL patterns match correctly


dirkf:
Thanks for your work.

I think this site should be handled within the existing extractor as updated in #32725. In fact I had code to do this but now I see that the site is showing a 403 error pager for any URL that I try in the domain.

dirkf:
I should also thank your colleague Claude while pointing out that nice formatting does not excuse the PR submitter from completing the PR template, which ought to be just the sort of thing it should be able to help with.

PR #32988, opened 2024-12-07
Replace all 'Twitter' links with 'X'

alisina97:
## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

### Before submitting a *pull request* make sure you have:
- [x] [Searched](https://github.com/ytdl-org/youtube-dl/search?q=is%3Apr&type=Issues) the bugtracker for similar pull requests
- [x] Read [adding new extractor tutorial](https://github.com/ytdl-org/youtube-dl#adding-support-for-a-new-site)
- [x] Read [youtube-dl coding conventions](https://github.com/ytdl-org/youtube-dl#youtube-dl-coding-conventions) and adjusted the code to meet them
- [x] Covered the code with tests (note that PRs without tests will be REJECTED)
- [x] Checked the code with [flake8](https://pypi.python.org/pypi/flake8)

### In order to be accepted and merged into youtube-dl each piece of code must be in public domain or released under [Unlicense](http://unlicense.org/). Check one of the following options:
- [x] I am the original author of this code and I am willing to release it under [Unlicense](http://unlicense.org/)

### What is the purpose of your *pull request*?
- [ ] Bug fix
- [x] Improvement
- [ ] New extractor
- [ ] New feature

---

### Description of your *pull request* and other information

Replaced all "Twitter" links in the codebase and documentation with their "X" equivalents to align with the platform's rebranding. Only the links were updated, as other occurrences of "Twitter" in the code or filenames were left unchanged to avoid potential issues.

- Updated documentation and comments containing links to "Twitter."
- Verified all updated links function correctly without breaking existing workflows.

This change ensures the repository is consistent with the platform's new branding while maintaining stability for all other aspects of the project.

PR #33268, opened 2026-04-15
[twitter] Add support for x.com domains (Fixes #33267)

### LINKED ISSUE #33267: [Twitter/X] Rebranded x.com URLs are not supported
https://github.com/ytdl-org/youtube-dl/blob/6508688e88c83bb811653083db9351702cd39a6a/youtube_dl/extractor/twitter.py#L34

adityakrmishra:
Resolves #33267

### Description
Since Twitter's rebranding to X, millions of URLs now use the `x.com` domain. The `_BASE_REGEX` in the Twitter extractor was previously hardcoded to only match `twitter\.com`, causing all `x.com` links to fail before extraction could begin. 

This PR updates the base regular expression to support both domains simultaneously. 

### Technical Details
* Replaced the hardcoded domain literal with a non-capturing alternation group `(?:twitter|x)`.
* Subdomain handling (`www.`, `m.`, `mobile.`) remains unchanged and applies properly to both domains.
* This is a regression-safe, backward-compatible fix that slots directly into the existing structure without modifying downstream `_VALID_URL` consumers.

### Regex Coverage Matrix
| URL | Matches |
|---|---|
| `https://twitter.com/user/status/123` | ✅ |
| `https://mobile.twitter.com/user/status/123` | ✅ |
| `https://x.com/user/status/123` | ✅ |
| `https://mobile.x.com/user/status/123` | ✅ |

Supersedes, closes #32988.


MathivadhaniArul:
Looked into the failing Windows/3.4 job — the annotations are just cascade noise ("canceled because ... failed", "operation was canceled"), no real test error in there. Pretty sure this is a pre-existing flake with building Python 3.4 on Windows runners, not caused by this PR.

Can someone check if this job also fails on main? If so we can probably just re-run/ignore it.

Fix itself looks good btw, matches the coverage matrix. Only thing I'd add is explicit x.com/mobile.x.com test cases to _TESTS so this doesn't regress later.

PR #33288, opened 2026-06-30
Fix case-sensitivity bugs when handling uppercase file extensions

aoright:
This pull request fixes several case-sensitivity bugs across youtube-dl when handling uppercase file extensions (e.g. .MP4, .MP3, .WEBM). Specifically: 1. In utils.py: fixed determine_ext, _change_extension, and _UnsafeExtensionError.sanitize_extension to do case-insensitive comparisons on extensions. 2. In YoutubeDL.py: fixed compatible_formats and correct_ext (in process_info) to check extensions case-insensitively, preventing the creation of double-extension output files (e.g. filename.MP4.mp4). 3. In postprocessors: updated embedthumbnail.py and ffmpeg.py to match extension types case-insensitively, resolving crashes and postprocessing errors when running on files with uppercase extensions. All new and existing unit tests pass successfully.


dirkf:
Thanks for your work, but 

**complete the template**!

Edit the completed template into the PR rather than starting again.
As you don't reference any issue that reports any potential fixed potential bug, either create one or post cases here.

At the same time consider that some filesystems are case-sensitive and some not, and in each case whether the wrongly or unexpectedly cased extension is being incorrectly generated elsewhere in code or through user action.

Also, check the behaviour of _yt-dlp_ in each case and raise an issue [there](https://github.com/yt-dlp/yt-dlp/issues/new/choose) if appropriate (different behaviour will set a high bar for merging).

PR #33292, opened 2026-07-31
/usr/bin/env: ‘python’: No such file or directory

### LINKED ISSUE #30129: /usr/bin/env: ‘python’: No such file or directory
## Checklist

- [ ] I'm reporting a broken site support issue
- [x] I've verified that I'm running youtube-dl version **2021.06.06**
- [x] I've checked that all provided URLs are alive and playable in a browser
- [x] I've checked that all URLs and arguments with special characters are properly quoted or escaped
- [x] I've searched the bugtracker for similar bug reports including closed ones
- [x] I've read bugs section in FAQ

## Verbose log

```
yoones:~$ whereis youtube-dl
youtube-dl: /usr/local/bin/youtube-dl
yoones:~$ youtube-dl
/usr/bin/env: ‘python’: No such file or directory
```

## Description

OS: Debian 11.1

I have the following binaries installed:
- python2.7
- python3
- python3.9

The shebang of youtube-dl (`#!/usr/bin/env python`) does not work in this context since there is no executable file called `python`.

**Suggested solution:**

Have a shell script that acts as a launcher. This script could for instance embed a list of known python interpreters (python, python2.7, python3, ...) and use the first of the list that is present on the host.

xiaoyaoqilan:

This PR fixes the documentation issue reported in #30129: corrects `#!/usr/bin/env python` to `#!/usr/bin/env python3` in `devscripts/bash-completion.py`.

Fixes #30129

---

- [ ] Bug fix
- [ ] Improvement
- [ ] New extractor
- [ ] New feature

---

Fixes #30129


dirkf:
The proposal would not solve the issue. To do so requires changing the commands used to create the bundled `zipimport` POSIX executable in the `Makefile`, but hitherto no-one has worked out how. Also, even if changing the shebang for the script that makes the _bash_ completion rules might have solved the problem, the proposal would have failed with an installed _Python_ that is not called `python3`. And **you did not complete the template**.

However, I see that the Mozilla devs came up with a trick where a POSIX shell script is hidden in what appears to be a Python program. As the comment says:
>The beginning of this script is both valid POSIX shell and valid Python, such that the script starts with the shell and is reexecuted with the right Python.

And
>[A shell script can be embedded] inside a Python triple quote. This pattern is valid [POSIX] shell because `''':'`, `':'` and `:` are all equivalent, and `:` is a no-op.

Like this:
```
#!/bin/sh
''':'
python=python3
command -v "$python" > /dev/null || python=python
exec /usr/bin/env "$python" "$@"
'''
# Python program goes here, the zipimport of yt-dl, or eg:
print (sys.argv[1:])
```

